PR #53, opened 2013-12-19
Implemented InsertBatchWithIdentity

lvaleriu:
Implemented InsertBatchWithIdentity
I had to add again my Emit classes. You can replace the IL calls by lambda expressions if you want to.


jogibear9988:
@ili What about this pull req? Should we (I) rework it?

sdanyliv:
@ili, @jogibear9988, it is really needed in main repository?

jogibear9988:
i thought, also because igor looked to merge it. i think it was supported by bltoolkit?

ili:
I'v looked through this PR, and it is not clear for me how does it works and what the difference between `InsertBatchWithIdentity` and `InsertBatch`

Also it looks like it is implemented only for Oracle, and to do the things right we should implement it for all databases.

Also it is not clear what for `Emit` is used, it would be better to use expressions.

So there are too much questions about this PR, it would be cool if somebody would make more deep review, to make design either we do need or not.

And it needs to be reworked before merging.

jogibear9988:
@lvaleriu can you say smth about the discussed points?

lvaleriu:
Hello. Yes, these 2 things are useful methods for Oracle mainly. 
InsertBatchWithIdentity:  Is a BulkCopy, but before it initiates the PRIMARY KEY fields with a new Oracle Sequence value. For doing that we reserve n sequence values for the n values we want to bulk copy (BuildReserveSequenceValuesSql).
InsertBatch: not needed anymore since BulkCopy specific implementation for ORACLE exists now.
(Note that there is a specific implementation for Oracle BulkCopy when there is no transaction: [)](https://github.com/lvaleriu/linq2db/blob/a5973add01924fb22e6ba6355e70ef355caa4d38/Source/DataProvider/Oracle/OracleDataProvider.cs#L492)
Emit is used since it is/was faster (for the things i was doing at least) then expressions. But that can be changed if you want.

jogibear9988:
@ili or @sdanyliv have you suggestions for @lvaleriu what to do, so we get this merged?

sdanyliv:
For start we have to get rid of conflicts and Emit if it is possible to replace by Expressions.
Emit may not work for iOS. I'm not familiar with .Net Core but i know that iOS has limitation - you can not build code on the fly. It is OS restriction.

ili:
@lvaleriu it would be great if you can rewrite this PR with points made by @sdanyliv 

Also it would be good to make it part of `BulkCopyOptions`. `InsertWithIdentity` returns inserted Id so `InsertBatchWithIdentity` should do it too and it should do real bulk copy, and now default implementation makes simple inserts.

So I suppose it is better to add `BulkCopyOptions.RetrieveSequence` or smth. like this

lvaleriu:
I'll take the task of rewriting this. But not before the next week, since
i'm in vacation these days. I wont use emit anymore. I'll keep you informed
of the d'advancement.
Le jeu. 6 avr. 2017 à 06:11, Ilya Chudin <notifications@github.com> a
écrit :

ili:
@lvaleriu cool :)

lvaleriu:
I've finished with this. I'll pull a new request during the day.

sdanyliv:
@lvaleriu, just create another commit in this branch and pull data to Github.
And for future, create for each pull request separate branch in your own repository, after merging it can be deleted.

lvaleriu:
Yes, this request will come from a different feature branch. I'll delete it after the pull request gets accepted.

lvaleriu:
Pull request is ready: https://github.com/linq2db/linq2db/pull/635. I'll close this one too during the day.

ili:
#635 merged, closing this

PR #172, opened 2015-02-19
Added opportunity to set association without using id property

VitaliyFedorov:
Sample of usage placed in Tests/Model/TestEntity1.cs

### DIFF
--- a/Data/Create Scripts/PostgreSQL.sql
+++ b/Data/Create Scripts/PostgreSQL.sql
@@ -1,4 +1,4 @@
-DROP TABLE "Doctor"
+﻿DROP TABLE "Doctor"
 GO
 
 DROP TABLE "Patient"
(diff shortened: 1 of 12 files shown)


igor-tkachev:
What is this for?

ili:
IMHO it would be better to determine `ThisKey` & `OtherKey` from `PrimaryKey` fields

VitaliyFedorov:
Reasons:
1) It's a way to do POCO class smaller.
2) If you already have many POCO classes (migrate to another ORM or use multiple ORM offering different functionality), now you do not need to create additional fields.

forik:
It would be great if this pr will be merged as this functionality increases POCO integrity. Say one had fetched entity from database and then has modified ReferenceId but Reference still has old Id.

jogibear9988:
@ili: don't you think this is usefull? i don't think we can determine the keys from the primary keys, maybe i don't use the primary key  for the association!

ili:
I do not like adding new property. 
I think it would be better to add smth like this:
```cs
public class Child
{
    [Assosiation]
    public Parent Parent;
}
```

And truly speaking I think that complex models for linq2db is sucks, and it should be done like this:
```cs
public class Child 
{
    public int ChildId;
    public int ParentId;
}
public class Parent
{
    public int ParentId;
}

public static DataExtensions
{
    public static ITable<Parent> Parent(this Child child) {/*some magic*/}
    public static ITable<Child> Childs(this Parent child) {/*some magic*/}
}

db.Parent.Where(_ => _.ParentId == 101).SelectMany(_ => _.Childs());
```

In general this possible using `ExpressionAttribute` but I'm not good with it... 
And it would be great to update T4 models to have a possibility to generate "light models"...

jogibear9988:
i like when we only need to set association attribute. if doneso, we can use the primary keys, but ifsomene sets the values we use them.
But I think it also would be good if we had the possibility that the id field has no roperty in the dto

ili:
Use case unclear :(
Closing, will reopen on feedback

PR #177, opened 2015-03-04
Added support for loading F# record types.

VoiceOfWisdom:


VoiceOfWisdom:
Added support for loading F# record types. Does not support more complicated properties, like `Association`, but resolves basic loading of F# records.

@igor-tkachev, I am going to create tests for inserting and deleting F# records, would you like to me add anything to the documentation for F#? Your library having explicit support for F# would be pretty great, most of the 'micro' orms don't work with F#.

VoiceOfWisdom:
I have added support for loading nested F# records and added more tests for F# records. The support required pretty heavy refactoring to the table builder expression, but the new logic looks to be much more future proof.

@igor-tkachev I'd like to add some F# snippets to the docs, would that be a acceptable change?

VoiceOfWisdom:
What I'm thinking for the F# section in the docs:

# F\# Record Support

**Linq2 to DB** has first class for F# record objects:

```f#
Type Name =
{ First : string
  [<Nullable>]Middle : string
  Last : string }

[<Table("Person", IsColumnAttributeRequired=false)>]
[<Column("FirstName",  "Name.First")>]
[<Column("MiddleName", "Name.Middle")>]
[<Column("LastName",   "Name.Last")>]
Type Person =
{ [<Identity>]
  [<Column("PersonID", IsPrimaryKey=true)>]
  ID : int
  Age : int
  Name : Name }


let john = { 
    Age = 20
    Age = 20
    Name = { 
        First = "John"
        Middle = ""
        Last = "Smith" }}

db.Insert(john);

let johnLoaded = query {
    for p in db.GetTable<john>
    where (p.Name.First = "John")
    exactlyOne }

let johnUpdated = { johnLoaded with Age = 30 }
db.Update(johnUpdated)

```

PR #185, opened 2015-03-11
Again, setting a class access modifier to public

lvaleriu:
Thanks!


lvaleriu:
I've added a small change on the following issue: Issue with the DataType value when executing a Where Query

PR #188, opened 2015-03-13
Implemented method RebindParameters to rebind Output and InputOutput parameters after procedure call.

bahram-aliyev:


bahram-aliyev:
Can somebody check this fix, and if possible merge with main branch? I use linq2db as nuget package, it would be great to have a new version with this fix.

PR #231, opened 2015-06-17
Linq

amccorma:
add the iSeries data provider


ferbenor:
Hello Mr. Igor, i don't understand how to work the Storage Property of Attribute Association, i think that thist property save the data into variable without in for set sentence, but in the load of data the inviroment use the set method of my property for example, i want that from load of data, this set into variable directly without pass for set method of my property. By other case, i undestand that the storage property of attribute association onli set the value directly into variable when not especify the set method in the property.
You could explain to me please.

I check the method loadwith and not use the storage property, how can implement this?.

Trank you.

ferbenor:
Don't Work storage property on loadWith method for association attribute...

ferbenor:
How fix it in storage property for association attrubute.

PR #255, opened 2015-08-31
Npgsql 3.0+ hstore and linq2db

breakingthings:
Hi,

Npgsql added hstore support as of 3.0 previous hack of using string as parameter type for hstore doesn't work anymore. Added new DataType Dictionary - maybe other DB-s will also have keyvalue extensions in the future. Currently requires Column attribute to have DataType = DataType.Dictionary set to work.


breakingthings:
I added datetime handling for ngpsql 3.0 also will make new pull request

PR #258, opened 2015-09-16
Fixed problem with generic filter predicates

rechkalov:
Hi! We have a problem with some complex queries. I think that the cause in line 
```
if (me.Member == binding.Member)
```

When compared "me.Member" and "binding.Member" the first one can be interface preperty that is implemented by type with the second one. In this case comparision gives us "false" and then exception raised, but I think that it is possible to proceed.

Simplest solution is to replace "me.Member == binding.Member" with "me.Member.Name == binding.Member.Name", but it is not good idea, so I used InterfaceMapping to determine if property is interface implementation.

Without this fix test "WhenPredicateFactoryIsNotGeneric" passes, but "WhenPredicateFactoryIsGeneric" fails. With it both tests pass.

### DIFF
--- a/Source/Linq/Builder/SelectContext.cs
+++ b/Source/Linq/Builder/SelectContext.cs
@@ -1078,6 +1078,30 @@ namespace LinqToDB.Linq.Builder
 									GetMemberExpression(binding.Expression.Unwrap(), expression, level + 1);
 						}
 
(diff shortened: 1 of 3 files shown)


rechkalov:
Thank you and shame on me, I did not found "EqualsTo" method :)

Issue #260, opened 2015-09-17
InsertWithIdentity is not working for DB generated GUIDs with [PrimaryKey, Identity] attribute on MS SQL Server


inickvel:
Need to use sequential guids , generated by DB.

Example:
CREATE TABLE [dbo].TestTable (
    [Id]     UNIQUEIDENTIFIER DEFAULT (newsequentialid()) NOT NULL,
    [FooCol] VARCHAR (50)     DEFAULT ('BLABLABLA') NULL,
    PRIMARY KEY CLUSTERED ([Id] ASC)
);

PR #262, opened 2015-09-20
Support InsertWithIdentity for DB generated GUIDs with [PrimaryKey, Identity] attribute on MS SQL Server. #260

inickvel:


inickvel:
May be better replace SCOPE_IDENTITY() with OUTPUT for all types, not only GUIDs ?
SCOPE_IDENTITY also does not working with SEQUENCEs (MS SQL 2012).

PR #265, opened 2015-10-05
Npgsql 3.0 datetime and hstore support

breakingthings:
Added support to npsql 3.x datetime and new datatype Dictionary for hstore key-value pairs that are now supported by npgsql driver.


breakingthings:
Have you looked inside this request and is it possible to merge this? Or is anything more required?

Issue #268, opened 2015-10-14
InsertWithIdentity is not working for SEQUENCEs with [PrimaryKey, Identity] attribute on MS SQL Server 2012.


inickvel:
May be better replace SCOPE_IDENTITY() with OUTPUT for all types, not only for GUIDs ?
SCOPE_IDENTITY also does not working with SEQUENCEs (MS SQL 2012).

http://www.sqlbadpractices.com/how-not-to-retrieve-identity-value/

igor-tkachev:
Be my guest :)

PR #280, opened 2015-11-01
Retrieving identity value by using the OUTPUT clause. MSSQL Server since 2005 version. #268

inickvel:
Replace SCOPE_IDENTITY() with OUTPUT for all types, not only for GUIDs,
SCOPE_IDENTITY also does not working with SEQUENCEs (MS SQL 2012).

http://www.sqlbadpractices.com/how-not-to-retrieve-identity-value/


inickvel:
http://www.sqlbadpractices.com/how-not-to-retrieve-identity-value/

PR #299, opened 2015-11-18
issues:278

10der:
GetQuery/FindQuery: bad performance

### DIFF
--- a/Source/Linq/Query.cs
+++ b/Source/Linq/Query.cs
@@ -179,22 +179,22 @@ namespace LinqToDB.Linq
 
 		static Query<T> FindQuery(IDataContextInfo dataContextInfo, Expression expr)
 		{
-			Query<T> prev = null;
+            //Query<T> prev = null;
 			var      n    = 0;
 
 			for (var query = _first; query != null; query = query.Next)
 			{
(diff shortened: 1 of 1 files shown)


ili:
not clear about performance. for now frequently used queries ate on the top of cache list

10der:
Kk. I'll investigate this issue again.

ili:
Thanks! We are waiting for feedback! :):)

ili:
Gettiing back to this issue, may be we can do like this:

```cs
static Query<T> FindQuery(IDataContextInfo dataContextInfo, Expression expr)
{
	Query<T> prev = null;
	var      n    = 0;

	for (var query = _first; query != null; query = query.Next)
	{
		if (query.Compare(dataContextInfo.ContextID, dataContextInfo.MappingSchema, expr))
		{
			if (prev != null)
			{
				prev.Next  = query.Next;
				query.Next = _first;
				Interlocked.Exchange(_first, query); // HERE!
			}

			return query;
		}

		if (n++ >= CacheSize)
		{
			query.Next = null;
			return null;
		}

		prev = query;
	}

	return null;
}
```

MaceWindu:
closing in favor of #645

Issue #305, opened 2015-11-22
Cannot use InsertWithIdentity when a trigger is on the table (MS SQL Server).


inickvel:
http://stackoverflow.com/questions/13198476/cannot-use-update-with-output-clause-when-a-trigger-is-on-the-table

http://blogs.msdn.com/b/sqlprogrammability/archive/2008/07/11/update-with-output-clause-triggers-and-sqlmoreresults.aspx

inickvel:
Up. Pull request made about 3 month ago :)

PR #306, opened 2015-11-22
Fix. Cannot use InsertWithIdentity when a trigger is on the table (MS SQL Server). #305

inickvel:
http://stackoverflow.com/questions/13198476/cannot-use-update-with-output-clause-when-a-trigger-is-on-the-table

http://blogs.msdn.com/b/sqlprogrammability/archive/2008/07/11/update-with-output-clause-triggers-and-sqlmoreresults.aspx


inickvel:
Up. Pull request made about 3 month ago :)

PR #317, opened 2015-12-15
Possible fix #316

### LINKED ISSUE #316: Memory leak with Query<T>.ObjectOperation<T1>
I have detected a memory leak when using DataConnection with customized MappingSchema:

``` c#
var schema = new MappingSchema();
var builder = schema.GetFluentMappingBuilder();
builder.Entity<Foo>().HasPrimaryKey(x => x.Id);

while (true)
{
    using (var db = new DataConnection("Test"))
    {
        db.AddMappingSchema(schema);
        db.Insert(new Foo() { ... });
    }
}
```

The problem is that:
- MappingSchema instance is a part of key used in static readonly dictionary in Query<T>.ObjectOperation<T1> 
- MappingSchema is compared only by reference 
- method AddMappingSchema always creates a new instance of MappingSchema 

As a result for each new iteration cache is appended.

Maybe I am wrong, but looks like that ConfigurationID property of MappingSchema can be used in dictionary key instead of MappingSchema itself.

rechkalov:
Here is one possible fix for memory leak #316

### DIFF
--- a/Source/Linq/Query.cs
+++ b/Source/Linq/Query.cs
@@ -567,7 +567,7 @@ namespace LinqToDB.Linq
 
 			Query<int> ei;
 
-			var key = new { dataContextInfo.MappingSchema, dataContextInfo.ContextID };
+			var key = new { dataContextInfo.MappingSchema.ConfigurationID, dataContextInfo.ContextID };
 
 			if (!ObjectOperation<T>.Insert.TryGetValue(key, out ei))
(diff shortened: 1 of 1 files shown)


rechkalov:
Sorry, it is not solution, Query<T> object keeps a link to MappingSchema instance, it is identical to passed with dataContextInfo one, but not the same. 
This fix works in my case, but in common case troubles possible.

PR #357, opened 2016-04-22
Possible fix #356

### LINKED ISSUE #356: Syntax error in generated sql code
Hi! I've found case when generated sql is wrong, it uses unavailable alias in join condition.

Here is extrimly simplified example:

``` C#
public class Foo
{
    public long Id { get; set; }
    public string Value { get; set; }
}

public class Baz
{
    public long Key { get; set; }
}

...

var union = db.GetTable<Foo>()
    .Union(db.GetTable<Foo>())
    .Distinct();

var result = db.GetTable<Baz>()
    .SelectMany(x => union.Where(date => date.Id == x.Key).Select(z => new {x.Key, z.Value}));

result.Take(10).ToArray();
```

this linq code leads to the next sql code:

``` SQL
SELECT TOP (10)
    [t4].[Key] as [Key1],
    [t3].[Value11] as [Value111]
FROM
    [Baz] [t4]
        INNER JOIN (
            SELECT DISTINCT
                [date1].[Id] as [Id1],
                [date1].[Value1] as [Value11]
            FROM
                (
                    SELECT
                        [t1].[Id],
                        [t1].[Value] as [Value1]
                    FROM
                        [Foo] [t1]
                    UNION
                    SELECT
                        [t2].[Id],
                        [t2].[Value] as [Value1]
                    FROM
                        [Foo] [t2]
                ) [date1]
        ) [t3] ON [date1].[Id] = [t4].[Key]
```

where "date1" alias is unavailable in join condition, "t3.Id1" should be used there (used MS SQL 2014)

This has workaround. It is enough to add dummy "Select" after "Distinct":

``` C#
var union = db.GetTable<Foo>()
    .Union(db.GetTable<Foo>())
    .Distinct()
    .Select(date => new Foo { Id = date.Id, Value = date.Value });
```

and generated sql becomes correct:

``` SQL
SELECT TOP (10)
    [t4].[Key] as [Key1],
    [t3].[Value11] as [Value111]
FROM
    [Baz] [t4]
        INNER JOIN (
            SELECT DISTINCT
                [date1].[Id] as [Id1],
                [date1].[Value1] as [Value11]
            FROM
                (
                    SELECT
                        [t1].[Id],
                        [t1].[Value] as [Value1]
                    FROM
                        [Foo] [t1]
                    UNION
                    SELECT
                        [t2].[Id],
                        [t2].[Value] as [Value1]
                    FROM
                        [Foo] [t2]
                ) [date1]
        ) [t3] ON [t3].[Id1] = [t4].[Key]
```

rechkalov:
Many of "Builder" classes checks IsDistinct property, but "WhereBuilder" don't. 
May be this it is cause of #356, at least it fixes it (see test)


rechkalov:
Appveyor build errors looks like the same as in current master

jogibear9988:
Is this issue still open? Is igors added test realy for this issue? I don't see a distinct in the test!

rechkalov:
No, commit 5c6be4fc73511b110d1b6ba61ac089ca70d610b6 looks like not related to this problem. Second, #357 is a pull request, not an issue. Third, that commit was created earler than pull request #357. So, it is probably typo.
The problem is #356 and test name is ```Issue356Tests```

PS. Problem is really minor, because "distinct" after "union" is redundant.

rechkalov:
Test ```Issue356Tests``` still fails at current master.

jogibear9988:
Can we merge this fix?

ili:
I'll look it in next few days

rechkalov:
Thank you.

PR #367, opened 2016-05-11
Rebind command return parameters

mdvtj:
If a command requires a parameter with `ParameterDirection.ReturnValue`,
its value must also be rebound after execution.

Otherwise e.g. calling an Oracle PL/SQL stored procedure that returns a value can be called, but the result cannot be obtained.


ili:
actually merged from another pr

PR #381, opened 2016-06-07
Added support jsonb PostgreSQL type

IgorPikul:
I decided this problem using property "InlineParameters" and set it to "true". After this .net type string can save as json/jsonb type to database (PostgreSQL)
Thanks.


ili:
@IgorPikul many tests are failing because of treating all strings as `Json` (PostgreSQL expects correct `Json` syntax).
It would be great to have more cases to work on `Json` support

ili:
Closing, will reopen on feedback

PR #392, opened 2016-07-05
oracle persists empty strings as nulls, so to make searches predictab…

gusev-p:
…le we convert empty strings to nulls for search predicates

### DIFF
--- a/Source/DataProvider/Oracle/OracleSqlBuilder.cs
+++ b/Source/DataProvider/Oracle/OracleSqlBuilder.cs
@@ -53,6 +53,31 @@ namespace LinqToDB.DataProvider.Oracle
 			return base.GetIdentityExpression(table);
 		}
 
+        private static void ConvertEmptyStringToNullIfNeeded(ISqlExpression expr)
+        {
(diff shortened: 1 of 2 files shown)


ili:
I'v run your test without `OracleSqlBuilder` changes, so it fails for empty string value.
Not sure why should we treat `""` as `null`? why not to use `string.IsNullOrEmpty()`?

gusev-p:
Because for Oracle checking for equality with empty string in searches is useless in any case, it will always return false and is a true sign of a bug.

gusev-p:
It's unnatural and easy to forget to use string.IsNullOrEmpty, especially if app code always explicitly pass empty strings.

sdanyliv:
@ili, maybe it can be configurable on connection/global level?

gusev-p:
Our most painfull scenario is protecting against duplicates by some string key property:
* receive some object with string key from the outside
* select() with filter by equality on this property
* if nothing returned, save this object to db, return selected otherwise

Without this path the code above create duplicates when key property is empty.

ili:
I'll look forward what can we do...

PR #414, opened 2016-08-18
Nested unions support

genusP:
Adds brackets for UNION queries


jogibear9988:
@ili can you also merge this?

ili:
I'll plan it. And it would be great if there would be also the test...
For me it is unclear what is fixed, and as for me it is not a good idea to merge PR without test cover :)

jogibear9988:
Seems it could lead to a Problem with SQLite if a Query starts with brackets:

(
SELECT [t1].[ChannelID] FROM [WMS_GlobalTask] [t1]
WHERE [t1].[ChannelID] IS NOT NULL
AND [t1].[StorageShelfDestinationID] IS NULL
AND [t1].[Status] < 88
)
UNION
(
SELECT [t2].[DestinationChannelID] as [ChannelID]
FROM [WMS_OutfeedTransportOrder] [t2]
WHERE [t2].[DestinationChannelID] IS NOT NULL
AND [t2].[Status] < 88
)

this does not work in Sqlite...

jogibear9988:
I think it needs a test and a better fix!!

ili:
What code produces this query?

jogibear9988:
Mine or his?

      var ignoreChannels = (
                from gt in s.Query<GlobalTaskDTO>()
                where gt.ChannelID.HasValue &&
                      !gt.StorageShelfDestinationID.HasValue &&
                      gt.Status < GlobalTaskStatus.Finished
                select gt.ChannelID.Value
                ).Union(
                from oto in s.Query<OutfeedTransportOrderDTO>()
                where oto.DestinationChannelID.HasValue &&
                      oto.Status < OutfeedTransportOrderStatus.Finished
                select oto.DestinationChannelID.Value
                ).ToList();

jogibear9988:
Issue for this Pull Req: https://github.com/linq2db/linq2db/issues/415

jogibear9988:
Maybe we need a outer Select * if we add the brackets?

ili:
@jogibear9988 yes... it looks like in this case we should make subquery, and it should looks like
```sql
select a, b
from
( 
  (...)
  union all
  (...) 
) 
```

jogibear9988:
yes, think you're right. 
Can you fix this? I think im not familiar enough with the expression parsing process atm

ili:
not sure that my Kung Fu is good enough :)
It will take a time...

ili:
Fix should be done in parser, not in SQL builder.
#572

PR #416, opened 2016-08-20
Test for F# query builder issue with leftOuterJoin

evilbloodydemon:
Provided query fails with NotImplementedException


jogibear9988:
What is the problem here? Since appveyor runs without error

ili:
It fails...b there are the problem with join parsing

PR #466, opened 2016-11-16
Fluent Mapping -> a defined Property should be included by default

jogibear9988:


jogibear9988:
If I don't do this, when I use Fluent Mapping, I always have to add another Method after the Property call:

              .Property(e => e.ServerLimitCheckScripts).IsNullable(true)

or 

         .Property(e => e.ServerLimitCheckScripts).IsNullable(true)

when the Property is a complex type!

ili:
@jogibear9988 if you'll provide any test to reproduce the problem, and check the fix, it would be great! :)

jogibear9988:
will do so

jogibear9988:
what is the expected way? that i commit a failing test, then create a fix so that the test works? can we configure appveyor to run the unit tests?

ili:
@jogibear9988 test + fix IMHO is the best way :)

appveyor runs tests now (can see in checks for this PR) :)

jogibear9988:
checks pass, can this be merged?

ili:
I'm going to do it tomorrow.
Now @igor-tkachev is reviewing core support

PR #467, opened 2016-11-17
Disable Clear of oderbys

jogibear9988:
If you hav nested Queryables, your object may become a Iqueryable again so you cannot use ThenBy. So don't clear orderbys

### DIFF
--- a/Source/Common/Configuration.cs
+++ b/Source/Common/Configuration.cs
@@ -13,6 +13,7 @@ namespace LinqToDB.Common
 			public static bool IgnoreEmptyUpdate;
 			public static bool AllowMultipleQuery;
 			public static bool GenerateExpressionTest;
+			public static bool DoNotClearOrderBys;
 		}
 
 		public static class LinqService
--- a/Source/Linq/Builder/OrderByBuilder.cs
+++ b/Source/Linq/Builder/OrderByBuilder.cs
@@ -50,8 +50,8 @@ namespace LinqToDB.Linq.Builder
 
 			builder.ReplaceParent(order, sparent);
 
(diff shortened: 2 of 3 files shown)


ili:
@jogibear9988 test? :)

jogibear9988:
I could not test, because I could not run the tests!
If created a setting, so the default is not changed

jogibear9988:
test works ;-)

ili:
@jogibear9988 what would be if we do not clear `OrderBy` at all?

jogibear9988:
It will be a breaking change! And it is not the same as when you use linq2objects. There orderby does clear

jogibear9988:
But also when we use dynamic.linq, multiple orderings are easier if we don't clear

ili:
@jogibear9988 I'v got an idea, static parameter is not good in multithread context, may be we can solve the case with new extension method like `ThenOrOrderBy`?

jogibear9988:
we have more static parameters, then we need to replace them all!

i think about it.

ili:
@jogibear9988 I suppose this is a good idea ) for example we can do smth. like `db.GetTable<T>().WithMultipleQuery()...`
And in any way, if we can avoid adding new static parameter, I think we should avoid :)

jogibear9988:
I think we should use this as it is for now.

And I work on an extra Patch wich removes the Static Configuration, and refactor them as a Settings Class of the DataConnection wich can be change per dataConnection instance?

ili:
@jogibear9988 for  `IDataContext` i think is better as for `InlineParameters` done...

jogibear9988:
done???

ili:
I'm going to merge it little bit later in the evening :) no time to do it from workplace

PR #470, opened 2016-11-21
Postgres: Quoting of columns like it was in bltoolkit (only the reser…

jogibear9988:
…ved words).

Make SqlBuilder public, so quote Mode can be set !
(Should also not be a static setting) -> more work needed

### DIFF
--- a/Source/DataProvider/PostgreSQL/PostgreSQLIdentifierQuoteMode.cs
+++ b/Source/DataProvider/PostgreSQL/PostgreSQLIdentifierQuoteMode.cs
@@ -6,6 +6,7 @@ namespace LinqToDB.DataProvider.PostgreSQL
 	{
 		None,
 		Quote,
+		Needed,
 		Auto
 	}
 }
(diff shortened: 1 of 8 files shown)


ili:
@jogibear9988 really strange logic for me, also i do not understand why not to quote always by default (or quote makes names case sensitive?), but in any way if we are doing quotation configuration available we should make it easy to understand.

For now we do have:
* `None` - we do not use quoting at all (*OK*)
* `Auto` - not used (*???*)
* `Quote` - quote only if any is upper or spaces (*why not all?*)
* `QuoreReserved` - only reserved, but what about spaces?

My suggesting is:
* `None` - we do not use quoting at all 
* `Auto` - quote only if any is upper or spaces (*default as now*)
* `Quote` - quote everything
* `QuoreReserved` - quote only reserved & spaces

jogibear9988:
Yes, quoting makes names case sensitive!

I think your suggestion is good, I will change the Code that it follows your suggestion. (as the class was not yet public, it's a non breaking change if I keep the same default behavior!)

As for now, shoudl I leave the static Variable in the PostgreSQLSqlBuilder or should I add it to the COnfiguration class? (I've already created a Issue to remove the static Configuration Class -> V2)

ili:
I think for now we can leave static field, as done for other providers

jogibear9988:
I've now changed to the following:

- None - we do not use quoting at all (OK)
- Quote - Quote everything
- Needed - Reserved Words & Spaces
- Auto - same as Needed + Uppercase Column Names

PR #475, opened 2016-11-22
Bugfix #411

jogibear9988:
Fix for #411

PR #477, opened 2016-11-23
New Fix for #475 and Test

jogibear9988:


jogibear9988:
Hopefully the test works, could not run it on my machine

ili:
@jogibear9988 still not clear for me...
if we are using `Single` there should be only one row in result dataset. So using `Take(1).Single()` looks like a hack, why not to use `First()` if we are not sure about rows count and do not care about it?

ili:
Yap, I see "if it works for linq2objects it should for linq2db" :):)

jogibear9988:
thats what i mean!

jogibear9988:
I don't say that it's agood idea to write it so. but it's correct code, and should return the same result as linq2objects!

PR #478, opened 2016-11-23
Northwind Sqlite Database for Local Tests

jogibear9988:


ili:
Cool!
But may be it is better to include "NorthwindDataContextAttribute" into `NorthwindDataContextAttribute`?

I think this should work fine:
```
 		[AttributeUsage(AttributeTargets.Method)]
 		public class NorthwindDataContextAttribute : IncludeDataContextSourceAttribute
 		{
 			public NorthwindDataContextAttribute() : base("Northwind", "NorthwindSqlite")
 			{
  			}
  		}
```

jogibear9988:
I've now fixes your suggestions, and started all Northwind Tests also against Sqlite.

But as you see, 6 do not work: 

https://ci.appveyor.com/project/igor-tkachev/linq2db/build/1.7.5.147/tests

Should we disable, and create extra issues for them?

ili:
I think we should find out why they are failing and fix it if possible :)

jogibear9988:
I've now disabled a few SQL Server specific tests (FreeText,...)

jogibear9988:
but a few still do not work

jogibear9988:
VisualBasicTests.SearchCondition3 & SearchCondition4 do not work cause of DateTime Format. But the dat ein the sqlite database file seems correct!

jogibear9988:
We now have following not working Tests:

- AllNestedTest, ComplexAllTest -> SQL Query gives another count than the Linq2Objects Call. Maybe we can compare the Generated SQL from Sqlite to Sqlserver
- FirstOrDefaultEntitySet, NestedSingleOrDefaultTest -> They need Subquery's when used with Sqlite -> Why???
- SearchCondition3, SearchCondition4 -> they seem to work when I load the Objects from the Sqlite DB, so it has to do with, how is the date stored in Sqlite (cause sqlite has no datetime type)

jogibear9988:
for the 3rd error:

maybe we should wrap every datetime call in sqlite with the date function so generated sql looks like this:

SELECT
	[t1].[OrderID]
FROM
	[Orders] [t1]
WHERE
	date([t1].[OrderDate]) = date(@p1)

jogibear9988:
For the 3rd error, I've created a fix. But I don't like the Method Name: "WrapParmeter". 
@ili : do you have a better idea?

jogibear9988:
For the other 4 failing tests, can you show me the generated SQL for SqlServer?
I've no Northwind DB to test against

ili:
Great job! I'll take a look at weekend

ili:
I'v got a brief look on SQLite date time case... I'm really surprised. Only `==` fails, and `>` for ex. succeed...

So I do think yes, we should arrange date time parameters with appropriate function.
But only parameters:
SQLite has `datetime` columns, but it looks like to be a fake. For ex if I do  
```
Select * from linqdatatype
```
In SqliteStudio column header  for `DateTimeValue` would be `date(DateTimeValue)`...

jogibear9988:
Yes, but only 2 fail, because no comparisons against other datetime values are made. it will also work, when we use linq2db for save, because then the correct string is used for comparison

jogibear9988:
> and < do string comparisons. so they succeed, but maybe wring!

ili:
Yes, i think that for now we have very poor situation with `DateTime` comparison in SQLite... And if I'm not mistaken nowhere else in tests  `DateTime` comparison is checked...

For now it looks like that your solution with generating call to [DateTime function](https://www.sqlite.org/lang_datefunc.html) is the way... But may be we should look how it is done for [EF](https://www.nuget.org/packages/System.Data.SQLite.EF6/) or for [System.Data.Sqlite.Linq](https://www.nuget.org/packages/System.Data.SQLite.Linq/). It would be great if there are any profiler for SQLite.. If not may be [MiniProfiler](https://www.nuget.org/packages/MiniProfiler/) can help (I think we should see actual query executed...)

So i think we should:
* look how it is done in another systems
* do more tests
* try to wrap only parameters (it looks like it would be enough)
* try to use `datetime()` instead of `date()` (i'm afraid that `date()` will fail comparisons with day time)

jogibear9988:
No, if you only Wrap Parameters, my Test will fail!

I don't think the other Libraries will wrap the calls, because the only Problem is that the Date format in the column is different from the one wich is used when we safe with linq2db. If it's the same, the comparison will work. So my fix is for hand edited Sqlite Databases.

I think the change to datetime is a good idea!

Have you also looked at the other two issues? The AllowSubQuery maybe it's no bug, maybe sqlite is missing some features, but we can enable them and it will work!

But the other 2 tests? Maybe you can show me the SQL wich is generated for SqlServer?

ili:
Speaking about SQL Server you can find northwind db files in bltoolkit repo... 
No, I have not looked for other tests, less time, preparing big release on primary job :)

Wrapping fields in conditions is a point to research, because we can have a complex expression like `e.DateValue.Date == parameterDate` and so on...

jogibear9988:
i've  no sqlserver installed.

does it work on other sql providers to work with datetime parts?

ili:
I think yes... I'll check it on weekend.

jogibear9988:
i thin, most linq sql providers don't supporting this scenario 

http://stackoverflow.com/a/28857666/579623

jogibear9988:
okay, seems like ef is supporting it: http://stackoverflow.com/a/7320703/579623

does linq2db also support this already?

ili:
> i thin, most linq sql providers don't supporting this scenario
if they don't - we should! :)

> does linq2db also support this already?
you meaning `DateTime` converts? yes, linq2db supports this, https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Linq/DateTimeFunctionsTests.cs

There are tests covering `DateTime` convert options, but no tests for `DateTime` comparisons. I suppose all comparisons works fine for all data providers, except SQLite, and I'm going to check it now :)

ili:
I'v discussed DateTime issue with @igor-tkachev he suggested to make an override for this case in `SqlBuilder.BuildPredecate` so I'll come back to it little bit later

jogibear9988:
but can't we use this pull req as is right now? i think when you override build predicate, we still need a setting to enable the use of datetime func. and i would like to use northwind sqlite in other pull req

ili:
I'll look forward to merge it today or tomorrow

ili:
Actually merged in https://github.com/linq2db/linq2db/commit/9f6d92c7b13b3299bf22a5018c69d7f11638cb3d

PR #490, opened 2016-12-01
Northsqlite

ili:


jogibear9988:
do we now always use the datetime function? willthis not have a worse performance? i think we should make this optional!

ili:
Only in predicates. In another case we are not sure about result, I think that unpredictable resultset is much more worse then little performance gap...

PR #502, opened 2016-12-13
Fix Linq2DB with System.Data.Sqlite on Mono

jogibear9988:
a file URI only starts with "file://" the next slash is part of the path!

### DIFF
--- a/Source/Common/Tools.cs
+++ b/Source/Common/Tools.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
-
+using System.IO;
+using System.Reflection;
 using JetBrains.Annotations;
 
 namespace LinqToDB.Common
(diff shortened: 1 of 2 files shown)


ili:
in other [cases](https://github.com/linq2db/linq2db/search?utf8=%E2%9C%93&q=CodeBase.Replace&type=Code) we also do have `file:///`

jogibear9988:
okay, i'll have a look

jogibear9988:
we should fix like here http://stackoverflow.com/a/4107765/579623

ili:
@jogibear9988 i think yes, but there is one problem, for ex we can have a path:
`c:\myfolder\#subfolder`
it would be `file://c:\myfolder\#subfolder` (or smth like that) so `LocalPath` would return only `c:\myfolder\` (evetything beore `#`) because everything after `#` is treated like `Uri.Fragment`

So I'm commonly using this method:
```cs
		public static string GetFileFromUri(this Uri uri)
		{
			if(!uri.Scheme.Equals("file", StringComparison.OrdinalIgnoreCase))
				throw new ArgumentException(string.Format("{0} - is not a file Uri", uri));

			return Path.Combine(uri.PathAndQuery, uri.Fragment);
		}
```

jogibear9988:
Fix should work, don't know why build fails...

jogibear9988:
Can you also merge this?

PR #503, opened 2016-12-14
Every Database Provider has it own Reserved words.

jogibear9988:
Without the extra list for Postgres, our Querys do not work after switch from bltoolkit to linq2db


jogibear9988:
Lists are from Migrator.NET, wich has them from the Database Manuals

Source/SqlQuery/ReservedWords.cs
```diff
@@ -22,8 +22,38 @@ static ReservedWords()
...
+					ReservedWords.ReservedWordsDictionaryPostgres.Add(s, s);
+			}
+
+			name = assembly.GetManifestResourceNames().Single(_ => _.EndsWith("ReservedWordsOracle.txt"));
+#if NETFX_CORE
+			using (var stream = assembly.GetManifestResourceStream(name))
+#else
+			using (var stream = assembly.GetManifestResourceStream(name))
+#endif
+			using (var reader = new StreamReader(stream))
```
sdanyliv: Looks like preprocessor directives is not needed in this method.

ili:
Why not to put all reserved words in one source?

jogibear9988:
Because if you Quote a Reserved Word, it's Case Sensitive. And if you don't quote when creating the Column this could cause Problems. (It causes Problems in our code!). It works with BLToolkit, but not with Linq2DB any more!

ili:
I see, i'll merge it on weekend

Issue #344, opened 2016-02-28
Creation of Plugins for Alternative Providers


to11mtm:
I'm looking at using Linq2Db for a project involving an Oracle Back-End.

However, our development house uses the Devart Oracle Driver instead of ODP.NET. We'd rather not use the ODP Driver (Bitness issues, non-native code, etc.) unless necessary.

However, it looks like you can't really 'plug in' a provider by making a separate library. This seems primarily due to how the library is currently encapsulated (i.e. AssemblyResolver and other classes are private)

Are there any plans to make the library more extensible in this regard?

tb-mtg:
To assist in creating custom DataProviders, could the following were made public?

**Classes**
- LinqToDB.DataProvider.BasicBulkCopy
- LinqToDB.DataProvider.BasicMerge
- LinqToDB.DataProvider.BulkCopyReader
- LinqToDB.DataProvider.DataTools
- LinqToDB.DataProvider.MultipleRowsHelper
- LinqToDB.Linq.Builder (**all in this namespace**)

**Class constructors**
- LinqToDB.SqlQuery.SqlDataType(DataType dbType, Type type, int? length, int? precision, int? scale)

**Properties**
- LinqToDB.Data.BulkCopyRowsCopied.RowsCopied

**Methods**
- LinqToDB.LinqExtensions.GetContext

ili:
We'll manage it

jogibear9988:
Devart ODP was supported by Bltoolkit, don't we support it in linq2db now?

ili:
No. In general there is Managed Oracle Provider... But i think it is a good idea to make a possibility to write custom providers. We have #462 for ex. but we can't merge it, because we can't test it. I don't know where can we get  AS/400 server :(

ili:
@tb-mtg i'v checked what now is public

Are rest really needed, or we can close an issue?

RoyC-Retail-Assist:
The changes noted _will_ allow a custom provider to be written  ( https://github.com/LinqToDB4iSeries/Linq2DB4iSeries ) so  suggest that any further changes could be requested if any future provider writer needed them.

ili:
So i'm closing the issue :)

PR #509, opened 2016-12-19
Changes required to allow providers in external assemblies : Issue #344

RoyC-Retail-Assist:
Facilitating DB2.iSeries provider LinqToDB4iSeries


Source/Mapping/MappingSchema.cs
```diff
@@ -361,7 +361,7 @@ ConvertInfo.LambdaInfo GetConverter(Type from, Type to, bool create)
 				var li   = info.GetConvertInfo(@from, to);
 
 				if (li != null && (i == 0 || !li.IsSchemaSpecific))
-					return i == 0 ? li : new ConvertInfo.LambdaInfo(li.CheckNullLambda, li.Lambda, null, false);
+					return i == 0 ? li : new ConvertInfo.LambdaInfo(li.CheckNullLambda, li.CheckNullLambda, null, false);
```
ili: @RoyC-Retail-Assist why have you changed it back?

Source/Mapping/MappingSchema.cs
```diff
@@ -655,10 +655,20 @@ public T GetAttribute<T>(Type type, Func<T,string> configGetter, bool inherit =
 		public T GetAttribute<T>(MemberInfo memberInfo, Func<T,string> configGetter, bool inherit = true)
 			where T : Attribute
 		{
+			var specific = GetSpecificAttributes<T>(memberInfo);
```
ili: @RoyC-Retail-Assist what is it for?

Source/DataProvider/IDataProvider.cs
```diff
@@ -17,7 +17,7 @@ public interface IDataProvider
 		Type               DataReaderType        { get; }
 		MappingSchema      MappingSchema         { get; }
 		SqlProviderFlags   SqlProviderFlags      { get; }
-
+		string             DummyTableName        { get; }
```
ili: @RoyC-Retail-Assist i cant see in changes where does it used, are you sure it is needed in base interface & DataProvider implementation?

Source/DataProvider/MySql/MySqlSqlBuilder.cs
```diff
@@ -215,15 +215,17 @@ public override object Convert(object value, ConvertType convertType)
...
 		{
 			return base.BuildExpression(
 				expr,
 				buildTableName && SelectQuery.QueryType != QueryType.InsertOrUpdate,
 				checkParentheses,
 				alias,
 				ref addAlias,
-				throwExceptionIfTableNotFound);
+				throwExceptionIfTableNotFound,
+			    wrapWithCastIfParameter);
```
ili: @RoyC-Retail-Assist what this parameter is for?

Source/SqlProvider/ValueToSqlConverter.cs
```diff
@@ -147,7 +149,7 @@ public bool TryConvert(StringBuilder stringBuilder, object value)
 #endif
 				)
 			{
-				stringBuilder.Append("NULL");
+				stringBuilder.Append(ParameterValueExpression(null, "NULL"));
```
ili: @RoyC-Retail-Assist please, can you explain use case?

Source/SqlProvider/ValueToSqlConverter.cs
```diff
@@ -179,11 +181,7 @@ bool TryConvertImpl(StringBuilder stringBuilder, SqlDataType dataType, object va
 			{
 				switch (type.GetTypeCodeEx())
 				{
-#if NETSTANDARD
```
ili: @RoyC-Retail-Assist why have you removed it?

Source/SqlProvider/ValueToSqlConverter.cs
```diff
@@ -179,11 +181,7 @@ bool TryConvertImpl(StringBuilder stringBuilder, SqlDataType dataType, object va
 			{
 				switch (type.GetTypeCodeEx())
 				{
-#if NETSTANDARD
-					case (TypeCode)2       : stringBuilder.Append("NULL");   return true;
-#else
 					case TypeCode.DBNull   : stringBuilder.Append("NULL");   return true;
```
ili: @RoyC-Retail-Assist why here is not used `ParameterValueExpression(null, "NULL")`?

Source/Mapping/MappingSchema.cs
RoyC-Retail-Assist: It was an error when merging codebases.  I will correct it and create a new pull request.

Source/Mapping/MappingSchema.cs
RoyC-Retail-Assist: It is to allow a provider specific version of the class to return specific attributes for the various methods in Sql.cs rather than it be done by reflection.  I could have just changed public  T GetAttribute<T>(MemberInfo memberInfo, Func<T,string> configGetter, bool inherit = true) to virtual however having a specific method to override gives IMO a more natural extension point.

This allows the following in the derived class:

`		protected override T GetSpecificAttributes<T>(MemberInfo memberInfo)
		{
			if (typeof(T) == typeof(Sql.ExpressionAttribute))
			{
				switch (memberInfo.Name)
				{
					case "CharIndex":
						return (T)(object)new Sql.FunctionAttribute("Locate");

					case "Trim":
						if (memberInfo.ToString().EndsWith("(Char[])", StringComparison.CurrentCultureIgnoreCase))
						{
							return (T)(object)new Sql.ExpressionAttribute(DB2iSeriesFactory.ProviderName, "Strip({0}, B, {1})");
						}
						break;
					case "TrimLeft":
						if (memberInfo.ToString().EndsWith("(Char[])", StringComparison.CurrentCultureIgnoreCase) ||
							memberInfo.ToString().EndsWith("System.Nullable`1[System.Char])", StringComparison.CurrentCultureIgnoreCase))
						{
							return (T)(object)new Sql.ExpressionAttribute(DB2iSeriesFactory.ProviderName, "Strip({0}, L, {1})");
						}
						break;
					case "TrimRight":
						if (memberInfo.ToString().EndsWith("(Char[])", StringComparison.CurrentCultureIgnoreCase) ||
							memberInfo.ToString().EndsWith("System.Nullable`1[System.Char])", StringComparison.CurrentCultureIgnoreCase))
						{
							return (T)(object)new Sql.ExpressionAttribute(DB2iSeriesFactory.ProviderName, "Strip({0}, T, {1})");
						}
						break;
					case "Truncate":
						return (T)(object)new Sql.ExpressionAttribute(DB2iSeriesFactory.ProviderName, "Truncate({0}, 0)");
					case "DateAdd":
						return (T)(object)new Sql.DatePartAttribute(DB2iSeriesFactory.ProviderName, "{{1}} + {0}", Precedence.Additive, true, new[] { "{0} Year", "({0} * 3) Month", "{0} Month", "{0} Day", "{0} Day", "({0} * 7) Day", "{0} Day", "{0} Hour", "{0} Minute", "{0} Second", "({0} * 1000) Microsecond" }, 0, 1, 2);
					case "DatePart":
						return (T)(object)new Sql.DatePartAttribute(DB2iSeriesFactory.ProviderName, "{0}", false, new[] { null, null, null, null, null, null, "DayOfWeek", null, null, null, null }, 0, 1);
					case "TinyInt":
						return (T)(object)new Sql.ExpressionAttribute(DB2iSeriesFactory.ProviderName, "SmallInt") { ServerSideOnly = true };
					case "DefaultNChar":
					case "DefaultNVarChar":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Char") { ServerSideOnly = true };
					case "Substring":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Substr") { PreferServerSide = true };
					case "Atan2":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Atan2", 1, 0);
					case "Log":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Ln");
					case "Log10":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Log");
					case "NChar":
					case "NVarChar":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Char") { ServerSideOnly = true };
					case "Replicate":
						return (T)(object)new Sql.FunctionAttribute(DB2iSeriesFactory.ProviderName, "Repeat");
				}
			}

			return default(T);
		}
`

Source/DataProvider/IDataProvider.cs
RoyC-Retail-Assist: DB2 on the iSeries requires a FROM clause for each select statement and there is a dummy table on the iSeries that can be used.

i.e. where SQL Server is happy with select CURRENT_TIMESTAMP
the equivalent on the iSeries would be select CURRENT_TIMESTAMP from SYSIBM.SYSDUMMY1

The dummy database name could be held internally but exposing it through the interface allows it to be used in the tests eg:

`	[Test]
		public void EnumExecuteScalarTest()
		{
			using (var dbm = new DataConnection())
			{
				string fromClause = string.IsNullOrWhiteSpace(dbm.DataProvider.DummyTableName) ? "" : string.Format(" from {0}", dbm.DataProvider.DummyTableName);
				var gender = dbm.Execute<Gender>(string.Format("select 'M'{0}", fromClause));

				Assert.That(gender, Is.EqualTo(Gender.Male));
			}
		}`

I haven't modified the tests in LinqToDb as I have copied them into the the LinqToDB4iSeries solution however the changes could be uplifted if required.

I can see a benefit to this but am happy to remove it.

Source/DataProvider/MySql/MySqlSqlBuilder.cs
RoyC-Retail-Assist: DB2 on the iSeries does not like parameters parameters in the SELECT element directly.  It will, however, allow them if they are cast.

e.g. SELECT @p1 FROM X  is not allowed but SELECT CAST(@p1 AS CHAR) FROM X  is.

I added the extra parameter to the base method to allow this wrapping and the only place that the method was overridden is in the MySQL implementation therefore I added the extra parameter in the class.

Source/SqlProvider/ValueToSqlConverter.cs
RoyC-Retail-Assist: Error in merging.  I will correct it and create a new pull request.

Source/SqlProvider/ValueToSqlConverter.cs
RoyC-Retail-Assist: Error in merging. I will correct it and create a new pull request.

Source/Mapping/MappingSchema.cs
ili: Lets make  `GetAttribute<T>` - `virtual` so in derived type you can do the same:
```cs
public override T GetAttribute<T>(MemberInfo memberInfo, Func<T,string> configGetter, bool inherit = true) 
	where T : Attribute 
{ 
    if (if (typeof(T) == typeof(Sql.ExpressionAttribute))
    {
          // some code
    }
 
    return base.GetAttribute<T>(memberInfo, configGetter, inherit);

}
```

Source/SqlProvider/ValueToSqlConverter.cs
ili: I see... I think it also can be solved as with `wrapWithCastIfParameter`
Can you test solution with overriding `BasicSqlBuilder.BuildColumnExpression` may be it will solve both this cases, if not we'll use `ParameterValueExpression`

Source/Mapping/MappingSchema.cs
RoyC-Retail-Assist: Will do

Source/DataProvider/IDataProvider.cs
RoyC-Retail-Assist: I'll remove it.

Source/DataProvider/MySql/MySqlSqlBuilder.cs
RoyC-Retail-Assist: I tried this as you suggested but it fails 28 of the tests.  The $CONVERT$ outputs SQL functions e.g SmallInt(@p1) rather than CAST(@p1 as SmallInt) unfortunately the iSeries is very particular and will only accept a CAST in these cases.  

If we can use the ParameterValueExpression from your comment below I propose to leave the call in BasicSqlBuilder but add a new SqlProviderFlag.WrapParametersWithCast to control whether it gets called.  This means that the method signature doesn't change but the Provider could control whether the call is made:


```
if (SqlProviderFlags.WrapParametersWithCast)
    name = ValueToSqlConverter.ParameterValueExpression(new SqlDataType(parm.SystemType), (string)name);

```

Source/SqlProvider/ValueToSqlConverter.cs
RoyC-Retail-Assist: Please see my response to the wrapWithCastIfParameter  comments

Source/DataProvider/MySql/MySqlSqlBuilder.cs
RoyC-Retail-Assist: Sure.  Thanks for your time on this.  I'll push the other changes now so we will only have this to resolve.

ili: I'v done some tests, so what you can do:
```cs
		protected override void BuildColumnExpression(ISqlExpression expr, string alias, ref bool addAlias)
		{
			if (expr is SqlParameter || expr is SqlValue)
			{
				expr = new SqlFunction(expr.SystemType, "Convert", SqlDataType.GetDataType(expr.SystemType),
					expr);
			}

			base.BuildColumnExpression(expr, alias, ref addAlias);
		}
```

this should work either for parameters or for `null`s and values

ili:
In general it would be great to have iSeries server to run tests with it we can include it in default distribution...

I'v even tried to write to IBM...

RoyC-Retail-Assist:
I've removed the wrapWithCastIfParameter and DummyTableName changes.  I'm working on removing the ValueToSqlConverter.ParameterValueExpression function for dealing with NULL in the select clause.

There is a public iSeries at http://pub400.com/ but I've not managed to find out if it has ODBC running on it as I haven't managed to get my signup acknowledged yet.  I'll let you know if I get anywhere with it but I'm happy to keep the provider in a separate project and publish it in its own NuGet with Linq2Db as a dependency.  Somehow I don't see hundreds of people wanting this as it is a pretty niche product.

ili:
> I'm working on removing the ValueToSqlConverter.ParameterValueExpression function for dealing with NULL in the select clause.

can you provide a test case? In example i'v posted earlier I'v looked this case, it also should be covered...

RoyC-Retail-Assist:
I've removed the ParameterValueExpression in my code and it is just 10 of the ConcatUnionTests that are failing e.g. Concat87.  I'll work through this tomorrow and am hopeful of getting a resolution.

RoyC-Retail-Assist:
I've removed the ParameterValueExpression function and the GetSpecifedAttributes method.  I had to make a slight change to the ConcatUnionBuilder so that it added the type to any null value expressions so that the builder could do the correct cast.

ili:
Actually merged, but I'v reverted making `MappingSchema.GetAttribute` `virtual`, because it should delegates to `IMetadataReader`

jogibear9988:
and are the renamings from "_" to "underscore"correct?

ili:
don't think that it is critical one :)

RoyC-Retail-Assist:
IMetadataReader works well for me.  Thanks.

Do you have a rough idea when this version is likely to released on NuGet?

ili:
@RoyC-Retail-Assist i'll try my best to do it in next couple of weeks. 
you can always find prerelease builds on myget (they all are supposed to be stable, published only if all tests are succeed)

RoyC-Retail-Assist:
Thanks.  I have the pre-release version but wanted to create a release version of the DB iSeriesProvider package.  A few weeks would be great :o)

If there is anything that I can look into that will help with this please let me know.

PR #515, opened 2016-12-21
Add appropriate oracle type for datetimeoffset to SQL builder

ronnyek:
Added DateTimeOffset to oracle create test, and verified I could write a date in and read it back out the same. Ultimately didn't leave it in the test project.


ili:
Can you please restore the test?

PR #519, opened 2016-12-27
Adding tests for F# option types

jjj-herobase:
As promised here is a passing test on how to use option types.


ili:
@jjj-herobase thanks a lot! :)

PR #520, opened 2016-12-27
Proposal for adding fluent mapping api for associations

jjj-herobase:
Hi.

I like to keep my models free of mapping attributes, and my mapping free of stringy property names, which is easily done by using the fluent mapping api. Except for associations, which doesn't seem to enjoy any fluent support. To fix that I've been using the below extension method. It works great, but I thought I'd try and submit it as a proper addition to the fluent mapping, so here you go.

```cs
using System;

using LinqToDB.Mapping;

namespace Actimizer.LRM.Linq2DB.Mappers
{
    public static class MapperExtensions
    {
        //private static GetMemberName( Expression expr )
        //{
        //    ExpressionVisitor

        //}

        public static PropertyMappingBuilder<T> Association<T,S,ID1,ID2>( 
            this PropertyMappingBuilder<T> pmb, 
            Expression<Func<T, S>> prop, 
            Expression<Func<T, ID1>> thisKey, 
            Expression<Func<S, ID2>> otherKey )
        {
            var thisKeyName = ((MemberExpression)thisKey.Body).Member.Name;
            var otherKeyName = ((MemberExpression)otherKey.Body).Member.Name;

            var objProp = Expression.Lambda<Func<T, object>>( prop.Body, prop.Parameters );

            return pmb.Property( objProp ).IsNotColumn().HasAttribute( new AssociationAttribute { ThisKey = thisKeyName, OtherKey = otherKeyName } );
        }

        public static PropertyMappingBuilder<T> IsNotNullable<T>(
            this PropertyMappingBuilder<T> pmb )
        {
            return pmb.IsNullable( false );
        }
    }
}
```

### DIFF
--- a/Source/Mapping/EntityMappingBuilder.cs
+++ b/Source/Mapping/EntityMappingBuilder.cs
@@ -115,7 +115,20 @@ namespace LinqToDB.Mapping
 			return new PropertyMappingBuilder<T>(this, func);
 		}
 
-		public EntityMappingBuilder<T> HasPrimaryKey(Expression<Func<T,object>> func, int order = -1)
+        public PropertyMappingBuilder<T> Association<S, ID1, ID2>(
+            Expression<Func<T, S>> prop,
+            Expression<Func<T, ID1>> thisKey,
+            Expression<Func<S, ID2>> otherKey )
+        {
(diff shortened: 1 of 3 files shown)


jogibear9988:
i think we should also support strings in Associations method, if a db field is used wich does not have a property (don't know if this is supported atm)

PR #521, opened 2017-01-02
Fluent mapping Fixes and Addons

jogibear9988:
- Fix Spaces to Tabs
- Fix Association should not Change IsColumn Attribute
- Support Inheritance via Fluent mapping (In Fluent Mapping we have the Problem, that Inherited classes do not have the attributes of the parent class!)

### DIFF
--- a/Source/Mapping/EntityDescriptor.cs
+++ b/Source/Mapping/EntityDescriptor.cs
@@ -197,9 +197,14 @@ namespace LinqToDB.Mapping
 					};
 
 					var ed = _mappingSchema.GetEntityDescriptor(mapping.Type);
+					var columnsList =
+						this.Columns.Union(ed.Columns.Where(x => !this.Columns.Any(y => y.MemberName == x.MemberName))).ToList();
+					ed.Columns.Clear();
+					ed.Columns.AddRange(columnsList);
 
 					foreach (var column in ed.Columns)
 					{
+
 						if (Columns.All(f => f.MemberName != column.MemberName))
 							Columns.Add(column);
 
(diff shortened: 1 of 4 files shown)


jogibear9988:
mhmm... test do not work yet... i need to look

jjj-herobase:
* Sorry about the spaces. I've installed an editorconfig extension to avoid that in future contributions. 
* Can an association ever be a column? It's not that I mind either way, I just added it to avoid adding it to every single association property.

jogibear9988:
don't know, but with my new code, when you use association attribute, it does not set iscolumn property, but does also not reset.

Source/Mapping/PropertyMappingBuilder.cs
```diff
@@ -20,8 +20,6 @@ public class PropertyMappingBuilder<T>
 			_entity       = entity;
 			_memberGetter = memberGetter;
 			_memberInfo   = MemberHelper.MemberOf(memberGetter);
-
```
ili: Why this was removed?

Source/Mapping/EntityDescriptor.cs
```diff
@@ -198,6 +198,12 @@ internal void InitInheritanceMapping()
 
 					var ed = _mappingSchema.GetEntityDescriptor(mapping.Type);
 
+					foreach (var column in this.Columns)
```
ili: @jogibear9988 what does this for? it looks like this breaks [some tests](https://ci.appveyor.com/project/igor-tkachev/linq2db/build/1.7.6.274/tests) and without it everything is fine...

PR #532, opened 2017-01-25
Test & Fix for Issue 531

jogibear9988:


jogibear9988:
Is this fix ok so?

jogibear9988:
@ili is this fix ok so?

jogibear9988:
maybe Guid should also be added as supported type.

Source/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1889,7 +1896,18 @@ ISqlExpression GetParameter(Expression ex, MemberInfo member)
 				accessorExpression = enumMapExpr.GetBody(accessorExpression);
 			}
 
-			var expr = dataContext.MappingSchema.GetConvertExpression(type, typeof(DataParameter), createDefault: false);
+			LambdaExpression expr = null;
+			if (buildParameterType != BuildParameterType.InPredicate
```
ili: @jogibear9988 may be we should always ignore `ConvertExpression` when building `In`?  I mean if we do have something like `IEnumerable<MyString>` it should be converted only when we are iterating through array?

jogibear9988: Your 're right, I will change

PR #543, opened 2017-02-02
trying to fix test fail after #521

ili:
One test failed after merging #521

### DIFF
--- a/Source/Mapping/EntityDescriptor.cs
+++ b/Source/Mapping/EntityDescriptor.cs
@@ -198,11 +198,11 @@ namespace LinqToDB.Mapping
 
 					var ed = _mappingSchema.GetEntityDescriptor(mapping.Type);
 
-					foreach (var column in this.Columns)
-					{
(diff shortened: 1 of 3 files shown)


jogibear9988:
When you look at the appveyor result from my pull req: https://ci.appveyor.com/project/igor-tkachev/linq2db-8c54d/build/1.7.6.265/tests all test work

ili:
@jogibear9988 yes, Core configuration succeed.

Also all tests are succeed on my home and work machine. Only fails on AV... I do think this is because of different test order...

But in anyway, we do have the situation where some tests are failed. I think we should revert this change.

jogibear9988:
dont think we should revert, we should find the issue!

jogibear9988:
i'll look into it

ili:
All tests are succeed...
Please take a look, because i have no idea what for this block was introduced...

ili:
@jogibear9988 i'll merge this fix for now.

If this code block is really needed, please, make separate PR

Issue #461, opened 2016-11-07
SubQuery - The method or operation is not implemented.


integral-llc:
I have such a query, which works OK if I don't use string concatenation in the anonymous object `Customer`, if I add `l.FullName` + "C" I get an exception saying "The method or operation is not implemented."

```
            var q = from se in _dataContext.ScheduleEvent
                where se.Start.HasValue && se.Start.Value.Date == request.Date.Date
                from seas in _dataContext.ScheduleEventAttendees
                where seas.ScheduleEvent_Id == se.Id
                from sea in _dataContext.ScheduleEventAttendee
                where sea.Id == seas.ScheduleEventAttendee_Id
                from sepNull in (from sep in _dataContext.ScheduleEventProperties
                    where sep.ScheduleEvent_id == se.Id && (sep.Name == Scheduler.PropEventJobId || sep.Name == Scheduler.PropEventLeadId)
                    select sep).DefaultIfEmpty()
                from appTypeNull in (from appType in _dataContext.ScheduleEventAppointmentTypeHistory
                    where appType.ScheduleEvent_Id == se.Id
                          && appType.Id == (from appTypeMax in _dataContext.ScheduleEventAppointmentTypeHistory
                              where appTypeMax.ScheduleEvent_Id == se.Id
                              select appTypeMax.Id).Max()
                    select appType).DefaultIfEmpty()
                from jobTypeNull in (from sep in _dataContext.ScheduleEventProperties
                    where sep.ScheduleEvent_id == se.Id && (sep.Name == Scheduler.PropEventJobId || sep.Name == Scheduler.PropEventLeadId)
                    select new
                    {
                        JobId = sep.IntValue,
                        JobIdType = sep.Name == Scheduler.PropEventJobId ? JobQuoteJobIdType.JobQuote : JobQuoteJobIdType.Lead
                    }).DefaultIfEmpty()
                from fabNull in (from sep in _dataContext.ScheduleEventProperties
                    where sep.ScheduleEvent_id == se.Id && (sep.Name == Scheduler.PropEventJobId || sep.Name == Scheduler.PropEventLeadId)
                    select new
                    {
                        Fabricator = sep.Name == Scheduler.PropEventJobId
                            ? (from j in _dataContext.ClientJob
                                where j.Id == sep.IntValue
                                from quote in _dataContext.Quote
                                where quote.ClientJob_id == j.Id
                                from cb in _dataContext.CustomerBase
[...]

PR #544, opened 2017-02-03
Fix for Issue #461

sdanyliv:
@igor-tkachev, please look at this fix for issue #461.
I'm really worried about this.


ili:
Actually merged

PR #547, opened 2017-02-06
don't generate system table entity

YaungOu:


NetworKKnighT:
I also wanted to create a pull request on this issue.

I don't know it is a bug or feature, but Mysql.Data assembly does not filter views by database name and returns all views for all databases on current server. It leads to T4 Templates for Mysql does not work correctly. I suggest to use restrictions and the GetTables method in MySqlSchemaProvider class  will be:
```C#
protected override List<TableInfo> GetTables(DataConnection dataConnection)
{
	// restrictions: TABLE_CATALOG, TABLE_SCHEMA, TABLE_NAME
	var restrictions = string.IsNullOrEmpty(dataConnection.Connection.Database) ? null : new [] { null, dataConnection.Connection.Database };

	var tables = ((DbConnection)dataConnection.Connection).GetSchema("Tables");
	var views  = ((DbConnection)dataConnection.Connection).GetSchema("Views", restrictions);

	return
		(
			from t in tables.AsEnumerable()
			let catalog = t.Field<string>("TABLE_SCHEMA")
			let name    = t.Field<string>("TABLE_NAME")
			select new TableInfo
			{
				TableID         = catalog + ".." + name,
				CatalogName     = catalog,
				SchemaName      = "",
				TableName       = name,
				IsDefaultSchema = true,
				IsView          = false,
			}
		).Concat(
			from t in views.AsEnumerable()
			let catalog = t.Field<string>("TABLE_SCHEMA")
			let name    = t.Field<string>("TABLE_NAME")
			select new TableInfo
			{
				TableID         = catalog + ".." + name,
				CatalogName     = catalog,
				SchemaName      = "",
				TableName       = name,
				IsDefaultSchema = true,
				IsView          = true,
			}
		).ToList();
}
```

YaungOu:
@NetworKKnighT good idea

sdanyliv:
Why do not use
GetSchemaOptions.ExcludedSchemas?
It's from documentation https://github.com/linq2db/t4models

NetworKKnighT:
As I said, the Mysql.Data assembly returns all views for all databases on current server. What if I have dozens of databases? And the point is that the Mysql.Data has a different behavior for tables and views. For tables, it returns the data filtered by database name.

sdanyliv:
My comment is related to thi pull request. It looks unnecessary if it is needed only for T4 generation.

NetworKKnighT:
Using GetSchemaOptions.ExcludedSchemas does not work, because the database name is contained in CatalogName property rather than SchemaName.

ili:
Actually merged.
Now system tables should not be generated. Also views issue fixed, thanks to @NetworKKnighT :)
[GetSchemaOptions.IncludedCatalogs and ExcludedCatalogs](https://github.com/linq2db/linq2db/blob/3d57331ff7e45414e5cfbf7d0539ede6fd0119d7/Source/SchemaProvider/GetSchemaOptions.cs#L13) added, so now you can apply filtering also by `CatalogName` :)

PR #550, opened 2017-02-07
Possible fix for #498

sdanyliv:
@igor-tkachev, another fix that needs your support.
I have not found any source history, even in BLToolkit, why this part of code was introduced,

In #498 introduced Association with CanBeNull = false. After grouping join become not weak and duplicate is not removed later.

### DIFF
--- a/Source/Linq/Builder/GroupByBuilder.cs
+++ b/Source/Linq/Builder/GroupByBuilder.cs
@@ -80,16 +80,6 @@ namespace LinqToDB.Linq.Builder
 			foreach (var sql in groupSql)
 				sequence.SelectQuery.GroupBy.Expr(sql.Sql);
 
-			new QueryVisitor().Visit(sequence.SelectQuery.From, e =>
-			{
-				if (e.ElementType == QueryElementType.JoinedTable)
-				{
-					var jt = (SelectQuery.JoinedTable)e;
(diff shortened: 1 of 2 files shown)


jogibear9988:
What does the "Weak" Property on a JoinedTable mean?

sdanyliv:
As i suppose this notifies optmizator that if there is no dependencies on this join (projections, etc.) - it can be removed. This technique widely used for Associations.

jogibear9988:
But shouldn't the SQL Code Gen also work is a join is not weak, so #498 should also work without removing this code?

(Does not mean that that code should not be removed if it leads to doubled joins)

sdanyliv:
I don't clearly understand your comment. I've removed small piece of code that makes Weak INNER JOIN to become strong. Why only INNER and why it is introduced ni GroupByBuilder - that's is the question. Looking for the sample that can break my fix.

jogibear9988:
Why does your change make a join strong? You removed code wich sets weak to false!

sdanyliv:
Hm, my change removed weak to strong conversion. That is what i've written in previous comment.

jogibear9988:
seems igor removed the code (commented out), so this is fixed?

sdanyliv:
@ili, i think Igor already committed fix https://github.com/linq2db/linq2db/commit/cb853e67bc1613c3b8493ead2f240d486876f09a

Issue #75, opened 2014-04-30
Invalid column index.


jack128:
Unit test     

``` C#
   [Test, DataContextSource]
        public void InvalidColumnIndex(string context)
        {
            using (var db = GetDataContext(context))
            {
                var childs = db.Child.Select(c => new 
                {
                    Child = c,
                    HasChildren = db.Child.Any(c2 => c2.ParentID == c.ChildID)
                });
                childs =
                    from child in childs
                    join parent in db.Parent on child.Child.ParentID equals parent.ParentID
                    where parent.Value1 == 1
                    select child;
                Assert.That(childs.ToList().Count, Is.GreaterThan(0));
            }
        }
```

generated sql

``` sql
SELECT
    child.ParentID as ParentID1,
    child.ChildID as ChildID1 -- "HasChildren" is not selected in query
FROM
    (
        SELECT
            c.ParentID,
            c.ChildID,
            CASE WHEN EXISTS(
                SELECT
                    *
                FROM
                    Child c2
                WHERE
                    c2.ParentID = c.ChildID
            ) THEN 1 ELSE 0 END as c1
        FROM
            Child c
    ) child
        INNER JOIN Parent parent ON child.ParentID = parent.ParentID
WHERE
    parent.Value1 = 1
```

jogibear9988:
#479 is a test for this

jogibear9988:
maybe anyone has a clue how to fix?

jogibear9988:
close?

ili:
Yes :) 
and released in [1.7.6](https://github.com/linq2db/linq2db/releases/tag/Release.1.7.6)

PR #556, opened 2017-02-15
Possible fix for #75

sdanyliv:
Fix for issue #75

I've found that for sub-queries is not called function ConvertToParentIndex which registers such kind of columns in parent query.
@igor-tkachev, please review this pull request.


ili:
It fails for SqlCe on my machine. Working on it

Issue #553, opened 2017-02-13
Concurrency checking


RoyC-Retail-Assist:
I'm looking to write some optimistic concurrency functionality similar to Entity Framework or NHibernate but I need something that I can hook into whenever an Update or Delete is to take place that I can then choose to cancel the update if desired.  Alternatively a method to add in row version columns to the where clause and a check that rows have been modified.

Is there anything currently available in the source code that could be used ( I cannot see anything obvious)?

If not I'd be interested in discussing how this could be added preferably as some extensibility points that could be used to supply providers for alternative strategies.

sdanyliv:
Check my answer here https://github.com/linq2db/linq2db/issues/331#issuecomment-277987587

RoyC-Retail-Assist:
Looks promising.

RoyC-Retail-Assist:
I've had a look at this and, whilst I can add the extra condition I cannot see how to get the original value to compare it with as the row being updated doesn't seem to be available in the ProcessQuery method.

ili:
Yes, on this point you'll receive only SQL AST that would be translated to SQL. So only constants as values would be available, but not parameters. 

Let's discuss how are you seeing such changes.

ili:
For ex my sample of optimistic concurrency:

```cs
public class Entity
{
    public int Id;
    public int Version;
    public string SomeValue;
}

public void Update(IDataContext db, Entity obj)
{
    var r = db.GetTable<Entity>()
        .Where(_ => _.Id == obj.Id && _.Version == obj.Version)
        .Set(_ => _.Version,  _ => _.Version + 1)
        .Set(_ => _.SomeValue, obj.SomeValue)
        .Update();

    if (r == 0) throw new ConcurrencyException(obj);
}
```

To implement that I can modify SQL AST adding `&& _.Version == obj.Version` and `.Set(_ => _.Version,  _ => _.Version + 1)`

So i do not need to know particular values...

P.S. not sure that we can now add new parameter to `SelectQuery` but why not... also point is how to throw exception...

RoyC-Retail-Assist:
I'm trying to avoid the developer having to code the concurrency check if at all possible and I also want to cater for a Version variable that uses the DB specific RowVersion.

Something like (using SQLite for testing):

```

		[Table("TestTable")]
		public class TestTable
		{
			[Column(Name = "ID", IsPrimaryKey = true, IsIdentity = true)]
			public int ID { get; set; }

			[Column(Name = "Description")]
[...]

PR #566, opened 2017-02-22
Concurency changes WIP

RoyC-Retail-Assist:
Work in process code for comment re #553

### DIFF
--- a/Source/Data/DataConnection.Linq.cs
+++ b/Source/Data/DataConnection.Linq.cs
@@ -384,6 +384,12 @@ namespace LinqToDB.Data
 			return dp.GetSqlOptimizer;
 		}
 
+		public void QueryExecuted(int count, string action)
+		{
+			
+		}
+
+
 		Func<ISqlOptimizer> IDataContext.GetSqlOptimizer
 		{
 			get { return GetGetSqlOptimizer(DataProvider); }
--- a/Source/DataExtensions.cs
+++ b/Source/DataExtensions.cs
@@ -231,15 +231,15 @@ namespace LinqToDB
 
 		#region Update
 
-		public static int Update<T>([NotNull] this IDataContextInfo dataContextInfo, T obj)
+		public static int Update<T>([NotNull] this IDataContextInfo dataContextInfo, T obj, bool overrideConcurrency = false)
 		{
 			if (dataContextInfo == null) throw new ArgumentNullException("dataContextInfo");
-			return Query<T>.Update(dataContextInfo, obj);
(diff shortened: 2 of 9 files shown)


Source/IDataContext.cs
```diff
@@ -34,5 +34,7 @@ public interface IDataContext : IDisposable
 		IDataContext        Clone              (bool forNestedQuery);
 
 		event EventHandler  OnClosing;
+
+		void				QueryExecuted		(int count, string action);
```
sdanyliv: You do not need this check TraceConnection action

Source/DataExtensions.cs
```diff
@@ -231,15 +231,15 @@ public static object InsertWithIdentity<T>(this IDataContext dataContext, T obj)
 
 		#region Update
 
-		public static int Update<T>([NotNull] this IDataContextInfo dataContextInfo, T obj)
+		public static int Update<T>([NotNull] this IDataContextInfo dataContextInfo, T obj, bool overrideConcurrency = false)
```
sdanyliv: For what you have needed these overrides? Looks like overkill.

Source/Data/DataConnection.Linq.cs
```diff
@@ -384,6 +384,12 @@ static Func<ISqlOptimizer> GetGetSqlOptimizer(IDataProvider dp)
 			return dp.GetSqlOptimizer;
 		}
 
+		public void QueryExecuted(int count, string action)
```
sdanyliv: You do not need this. Check TraceConnection action. It is unnecessary to add this method.

Source/IDataContext.cs
RoyC-Retail-Assist: The concurrency checks only need to be done for a limited number of methods.  Using trace connection would require parsing of the SQL to determine whether or not the concurrency checks are required and the parsing would need to be DB specific to be successful.  I'm happy to explore any other suggestions as to how this could be achieved without underlying changes.

Source/DataExtensions.cs
RoyC-Retail-Assist: This provides the capability for the user to be able to force the update to succeed even if the record has changed in the DB ("user wins" scenario).  This is demonstrated in CheckUpdateFailWithOverride test.  Without the override the user would have to re-query the DB to get the latest version of the record and then reapply the change.

MaceWindu:
@RoyC-Retail-Assist what is the status of this PR? Do you plan to return to this feature? If so, could you update PR with latest changes.

PR #570, opened 2017-02-24
bugfix for 565

jogibear9988:


ili:
Changing `EntityMappingBuilder` is not a good idea. You see, it is generic method, and now it have a deal with hardcoded `ColumnAttribute`
I think change should be done in `PropertyMappingBuilder`  & `PropertyMappingBuilder.SetColumn(Action<ColumnAttribute> setColumn)`

1) change _memberGetter to get "right" (as you'v done in  `EntityMappingBuilder`
2) make some changes in `PropertyMappingBuilder.SetColumn(Action<ColumnAttribute> setColumn)`

jogibear9988:
i look to fix on monday.

jogibear9988:
maybe so the fix is better?

Issue #573, opened 2017-02-25
10 Not working Tests from ORM Battle


jogibear9988:
Orm Battle has 16 failing Tests with current Linq2DB (tested against Northwind Sqlite)

Not working Tests:

- AnyParameterizedTest 
- ComplexGroupingTest 
- ~~ComplexTakeSkipTest~~ 
- ComplexTest2 
- ComplexTest4 
- ComplexTest6 
- **~~ConcatTest~~** 
- **~~EntitySetCountTest~~** 
- GroupJoinTest 
- LeftJoinTest
- OrderByDistinctTest 
- **~~OrderByPredicateTest~~** 
- OrderBySelectManyTest 
- ~~SumTest~~
- **TypeCastIsChildConditionalTest**
- **~~WhereNullableParameterTest~~** (is fixed in pull req #574)

The bold Tests work in bltoolkit!

I've removed the source code links, cause i've added the tests to linq2db in #574

jogibear9988:
TypeCastAsTest, TypeCastIsChildTest and TypeCastOfTypeTest do not work, because Inheritance Mapping is Configured with true/false. In Sqlite true is stored as '1' but that seems not to work in inheritance mapping! If I change to use '1' it works. I think we should fix this.

jogibear9988:
@ili should I create extra issues for each bug?

ili:
It would better to create pull requests with fixies :):):)
I don't think new issues are needed

jogibear9988:
I think, I should be able to fix the inheritance issues in sqlite... for the rest... I'm not sure if I'm able to solve...

jogibear9988:
Can I add the tests from ormbattle to linq2db?

ili:
Yes

jogibear9988:
I've added a fix for WhereNullableParameterTest  to my Pull Req.

The rest, I could not fix....

jogibear9988:
sdanyliv: have you any idea about the non working tests? maybe you could help, seems like you now linq2db sql generation very well

jogibear9988:
what I don't understand is why sum test is not workin.

jogibear9988:
I've noe included the tests (all) into Linq2DB. (result: https://ci.appveyor.com/project/igor-tkachev/linq2db-8c54d/build/1.7.7.377/tests) 

Should i now disable the not working ones? I don't think, I'm able to fix them atm.

I'll look that the tests will run against all Database providers, when I've time. Atm, they only use Sqlite

[...]

PR #574, opened 2017-02-27
tests from ormbattle #573

jogibear9988:


jogibear9988:
why? i think it's good if all this would be tested, so we'll see if smth will break linq functionality

jogibear9988:
I think it's a good idea, to add all the tests. Look at test "WhereNullableParameterTest", it's broken cause of a change we did in sqlite with datetime, so we would not find out if we only add tests wich were broken.

But I also think, we should add the tests for all DB Providers, not only sqlite, but that can be a step 2

Source/DataProvider/SQLite/SQLiteSqlBuilder.cs
```diff
@@ -131,17 +131,22 @@ protected override void BuildPredicate(ISqlPredicate predicate)
 				var leftType  = exprExpr.Expr1.SystemType;
 				var rightType = exprExpr.Expr2.SystemType;
 
-				if (IsDateTime(leftType) || IsDateTime(rightType))
+				if ((IsDateTime(leftType) || IsDateTime(rightType)) && !((exprExpr.Expr1 is SqlValue && ((SqlValue)exprExpr.Expr1).Value == null) || (exprExpr.Expr2 is SqlValue && ((SqlValue)exprExpr.Expr2).Value == null)))
```
ili: @jogibear9988 it is better to check if type is `IValueContainer`, this way you will close two cases: `SqlValue` & `SqlParameter`

Source/DataProvider/SQLite/SQLiteSqlBuilder.cs
```diff
@@ -131,17 +131,22 @@ protected override void BuildPredicate(ISqlPredicate predicate)
 				var leftType  = exprExpr.Expr1.SystemType;
 				var rightType = exprExpr.Expr2.SystemType;
 
-				if (IsDateTime(leftType) || IsDateTime(rightType))
+				if ((IsDateTime(leftType) || IsDateTime(rightType)) && !((exprExpr.Expr1 is SqlValue && ((SqlValue)exprExpr.Expr1).Value == null) || (exprExpr.Expr2 is SqlValue && ((SqlValue)exprExpr.Expr2).Value == null)))
 				{
+					
+					if (leftType != null)
```
ili: @jogibear9988 `leftType` or `rightType` would be `null` for `null` `Value`?

Source/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -421,8 +421,13 @@ TransformInfo OptimizeExpressionImpl(Expression expr)
 							var isList = typeof(ICollection).IsAssignableFromEx(me.Member.DeclaringType);
 
 							if (!isList)
+								isList =
```
ili: @jogibear9988 what does it for? We do check the same below and in more general way. I think this check is not needed.

ili:
Speaking about OrmBattle tests, i'm agree with @sdanyliv - it is better to rewrite failing tests using our object model, and run them against **all databases** - for ex "WhereNullableParameterTest" fails because of bug in SQLite DateTime compassion, and we are not sure that it is only database where it fails.

So, my idea is to treat tests failing now as point to fix now, and check for all databases. In future may be we'll rewrite all tests using our model to run for all databases...

Or, may be we would migrate to https://github.com/lerocha/chinook-database instead of Northwind

Source/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -421,8 +421,13 @@ TransformInfo OptimizeExpressionImpl(Expression expr)
 							var isList = typeof(ICollection).IsAssignableFromEx(me.Member.DeclaringType);
 
 							if (!isList)
+								isList =
+									me.Member.DeclaringType.IsGenericType &&
```
ili: @jogibear9988 one more point - there are reflection extensions (in `ReflectionExtensions.cs`) they are written to be compatible for different frameworks (conditional compilation), so in future avoid using direct reflection functions, firstly check if there are one in extensions.

For ex if you want to chech if type is generic use `ReflectionExtensions.IsGenericTypeEx(this Type type)`.
Most methods are with the same name as reflection methods or properties, only having `Ex` postfix :)

Source/DataProvider/SQLite/SQLiteSqlBuilder.cs
jogibear9988: it is in a test from ormbattle

Source/DataProvider/SQLite/SQLiteSqlBuilder.cs
jogibear9988: don't know what you mean

jogibear9988:
i'll rewrite the tests to use our model, when the errors are fixed

Source/Linq/Builder/ExpressionBuilder.cs
jogibear9988: with my change test works, without not...

Source/Linq/Builder/ExpressionBuilder.cs
jogibear9988: NetStandart will have the same reflection methods in 2.0 as full framework. Du we then still need the extensions?

Source/Linq/Builder/ExpressionBuilder.cs
jogibear9988: if you think the check 2 lines lower works, it does not if your type is directly icollection<T>, which is the case here!

jogibear9988:
Some of the OrmBattle Test will not work on Sqlite on NetCore, see: https://github.com/aspnet/Microsoft.Data.Sqlite/issues/331

jogibear9988:
I've switched the OrmBattle Tests to our model

jogibear9988:
have you looked at my reworked fixes? how should I go on with this?

ili:
I'll take a look today :)

ili:
>Some of the OrmBattle Test will not work on Sqlite on NetCore, see: aspnet/Microsoft.Data.Sqlite#331

yes. they do say that it is because MS uses official SQLite build, and System.Data.SQLite - non official

That's why i'v introduced SQLiteMs configuration for tests, exclude failing tests.

Source/Linq/Builder/ExpressionBuilder.cs
ili: `me.Member.DeclaringType.GetInterfacesEx()` do not return `ICollection<>`?? that's strange...

ili:
>have you looked at my reworked fixes? how should I go on with this?

Yes! :)

jogibear9988:
yes?

i do mean, i'm not ableto fix the rest of the tests.. (the sqlite ones i will change to sqlitems). should i disable them ? so this could be merged? 

we could kerp the issue open, so we know we've work to do...

ili:
Some misunderstood. I mean that you should continue working on this.

Yes, we can disable some tests for .net core using conditional compilation.

Other failing test may be disabled for now

jogibear9988:
I've now added

     if (context != TestProvName.SQLiteMs)
			{

to the failing tests... but they still fail...

jogibear9988:
Okay, we only have "NorthwindSqlite" and "Northwind". But how can I check if "NorthwindSqlite" uses Microsoft.sqlite?

ili:
> Okay, we only have "NorthwindSqlite" and "Northwind". But how can I check if "NorthwindSqlite" uses Microsoft.sqlite?

New configuration `NorthwindSqliteMs` is needed. So for Core we'll use `NorthwindSqliteMs` instead of `NorthwindSqlite` in `UserDataProviders.Core.txt`

Then we'll need to make some rewrite of `NorthwindDataContextAttribute`, like this:

```cs
[AttributeUsage(AttributeTargets.Method)]
public class NorthwindDataContextAttribute : IncludeDataContextSourceAttribute
{
	public NorthwindDataContextAttribute(bool excludeSqlite, bool excludeSqliteMs = false) : base(
		excludeSqlite
		? new[] { "Northwind" }
		:  ( excludeSqliteMs  ? new[] { "Northwind", "NorthwindSqlite" } : new[] { "Northwind", "NorthwindSqlite",  "NorthwindSqliteMs"}))
	{
	}

	public NorthwindDataContextAttribute() : this(false)
	{
	}
}
``` 

and will mark excluding tests as `[NorthwindDataContext(false, true)]`

jogibear9988:
i will do so

jogibear9988:
the tests now work ;-)

jogibear9988:
can be merged....

jogibear9988:
but don't close #573, there are still failing tests

PR #578, opened 2017-03-01
Fixes for TravisCI

jogibear9988:
[Travis CI builds](https://travis-ci.org/jogibear9988/linq2db/) 

badge

Issue #504


Source/global.json
```diff
@@ -0,0 +1,3 @@
+{
+  "sdk": { "version": "1.0.0-preview2-1-003177" }
```
ili: @jogibear9988 are you sure this is needed? by default should be used latest installed also configured `dotnet: 1.0.0-preview2-003156`

jogibear9988: i don't need this on travis, but on windows! without it uses the newer(i have vs2017 installed), and i could not compile

ili:
About MySql, try to use MySql57 instead of MySql in *DataProviders.txt (it should solve problems with year).

Tests/Linq/TravisDataProviders.Osx.txt
```diff
@@ -1,8 +1,3 @@
-﻿--Access
---SqlCe
---SQLite
---NorthwindSqlite
-
-!MySql           * Server=localhost;Database=TestData;Uid=root;Pwd=;charset=utf8;                                            * MySql.Data.MySqlClient
-!PostgreSQL     * Server=localhost;Database=testdata;User Id=postgres;Password=;Pooling=true;MinPoolSize=10;MaxPoolSize=100 * Npgsql
+﻿
+PostgreSQL     * Server=localhost;Database=testdata;User Id=postgres;Password=;Pooling=true;MinPoolSize=10;MaxPoolSize=100 * Npgsql
```
ili: `!` symbol means that it is default configuration if it is not set, then would be used `SQLite` or `SQLiteMs` Default connection is widely used in tests, so if we do not run `SQLite` we need to redefine one. So you should use !PostgreSQL

ili:
Is it possible to make more complex travis matrix:

- linux
  - mono
  - core
- osx
  - mono
  - core

Now all tests are done in one buildlog, and if i'm not mistaken build fails after failing Core tests, it would be great if we can also see mono results...

jogibear9988:
don't know... but i think, at first we should check that it works... i think if we splitt to 4 tasks, the build takes much longer, because they are not run in parallel

ili:
OK, it looks like we do have `-1.79E+308d` problem... 
lets do one thing - 
* lets extract double tests with 1.7900000000000008E+308d in separate method and mark them as `WindowsOnly`
* in numeric tests lets try to make some cheat like `1.5E+308d` may be it would help...

ili:
@jogibear9988 what problems we do have now with travis build?

jogibear9988:
i'll look after the orbattle pull req works

ili:
F# tests are failing on appveyor. Need to update nunit package reference in project.

Issue #445, opened 2016-10-22
Max Pool Size limit Exception


kdkdkdkd85:
Hello.
When using linq2db the following error:

```
Timeout. The timeout period elapsed before managed to get a connection from the pool. Perhaps all of the connections in the pool are already in use and will reach maximum size of the pool.
   в System.Data.ProviderBase.DbConnectionFactory.TryGetConnection(DbConnection owningConnection, TaskCompletionSource'1 retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, DbConnectionInternal& connection)
   в System.Data.ProviderBase.DbConnectionInternal.TryOpenConnectionInternal(DbConnection outerConnection, DbConnectionFactory connectionFactory, TaskCompletionSource'1 retry, DbConnectionOptions userOptions)
   в System.Data.SqlClient.SqlConnection.TryOpenInner(TaskCompletionSource'1 retry)
   в System.Data.SqlClient.SqlConnection.TryOpen(TaskCompletionSource'1 retry)
   в System.Data.SqlClient.SqlConnection.Open()
   в LinqToDB.Data.DataConnection.get_Connection() в i:\linq2db\Source\Data\DataConnection.cs:строка 499
   в LinqToDB.Data.DataConnection.CreateCommand() в i:\linq2db\Source\Data\DataConnection.cs:строка 567
   в LinqToDB.DataProvider.DataProviderBase.InitCommand(DataConnection dataConnection, CommandType commandType, String commandText, DataParameter[] parameters) в i:\linq2db\Source\DataProvider\DataProviderBase.cs:строка 90
   в LinqToDB.Data.DataConnection.LinqToDB.IDataContext.ExecuteReader(Object query) в i:\linq2db\Source\Data\DataConnection.Linq.cs:строка 261
   в LinqToDB.Linq.Query'1.<RunQuery>d__11.MoveNext() в i:\linq2db\Source\Linq\Query.cs:строка 371
   в LinqToDB.Linq.Query'1.<Map>d__6a.MoveNext() в i:\linq2db\Source\Linq\Query.cs:строка 1177
```

I use linq2db as follows:
1. UnitOfWork pattern:

```
public class Linq2DbUow : IUnitOfWork {
    private readonly ShymkentEduDb _db;
    public Linq2DbUow() {
        _db = new ShymkentEduDb();
        LinqToDB.Common.Configuration.Linq.AllowMultipleQuery = true;
    }
    public void Dispose() {
[...]

PR #584, opened 2017-03-06
Deny execute queries on disposed connections to avoid connection leaks

ili:
#445

This to fail in error immediately when wrong using pattern is applied


MaceWindu:
Funny that couple of weeks we had this issue due to return of linq query from transaction scope and I even planned to implement protection from it in our application. But having it fixed in linq2db is even better.

For now new linq2db releases published without release notes. Maybe it makes sense to provide them with new releases? I expect to receive bug reports about broken applications when this fix approved and released. Good release notes will help users to understand what is going on.

ili:
>For now new linq2db releases published without release notes. Maybe it makes sense to provide them with new releases? I expect to receive bug reports about broken applications when this fix approved and released. Good release notes will help users to understand what is going on.

Certainly you are right. We need some mechanism to build release notes from issues and pull requests... I think there is some suggested by GitHub, and, shame on me, I do not know it :(

Source/Linq/DataContextInfo.cs
```diff
@@ -7,11 +7,11 @@ namespace LinqToDB.Linq
...
-		public DataContextInfo(IDataContext dataContext)
-		{
-			DataContext    = dataContext;
-			DisposeContext = false;
-		}
+		//public DataContextInfo(IDataContext dataContext)
+		//{
+		//	DataContext    = dataContext;
+		//	DisposeContext = false;
+		//}
```
sdanyliv: Why this constructor commented? Maybe just remove?

Source/SqlQuery/SelectQuery.cs
```diff
@@ -168,31 +168,82 @@ public   string  Alias
...
+				//	|| new QueryVisitor().Find(Expression, e =>
+				//		{
+				//			switch (e.ElementType)
+				//			{
+				//				case QueryElementType.Column: return ((Column)e).Expression.Equals(other.Expression);
+				//			}
+				//			return false;
+				//		}) != null;
+
+				//return found;
```
sdanyliv: Commented for history?

MaceWindu:
For release notes maybe we could look at those tools?
https://habrahabr.ru/company/eastbanctech/blog/323760/#cz

Source/ServiceModel/RemoteDataContextBase.cs
```diff
@@ -353,6 +361,8 @@ IDataReader IDataContext.ExecuteReader(object query)
 
 		public void ReleaseQuery(object query)
 		{
+			ThrowOnDisposed();
```
sdanyliv: It is good place for throwing?

ili: The Idea was to throw exception on calling `IDataContext` methods (almost all methods).

PR #587, opened 2017-03-09
Hints support for TOP closure

ili:
* Access
* Sql Server


ili:
@sdanyliv this is unsupported scenario. I'v added exception throwing & documentation to make it clear

Issue #82, opened 2014-05-20
Wrong SQL generated with Grouping and Count! 


jogibear9988:
Same issue as in BlToolkit

https://github.com/igor-tkachev/bltoolkit/issues/312

jogibear9988:
Do you hav any News for this?

igor-tkachev:
I am still working on it?

jogibear9988:
News?

jogibear9988:
I've this Expression:

```
   var query = from m in s.Query<MaterialDTO>()
            join ir in s.Query<InventoryResourceDTO>() on m.Id equals ir.MaterialID into irc
            select new MaterialInfoDTO
            {
                Material = m,
                CountResources = irc.Count(),
                CountInventory = irc.Sum(x => x.Quantity)
            };
```

this works, I get the COUNT and SUM

but when I try to do a where after the query:

```
    query = query.Where(x => x.CountResources > 0);
```

wrong SQL is generated! (Tested with SQLite and MSSQL)

If I write like this, it works:

```
    var query = from m in s.Query<MaterialDTO>()
            select new MaterialInfoDTO
            {
                Material = m,
                CountResources = s.Query<InventoryResourceDTO>().Count(x => x.MaterialID == m.Id),
                CountInventory = s.Query<InventoryResourceDTO>().Where(x => x.MaterialID == m.Id).Sum(x => x.Quantity)
            };

        query = query.Where(x => x.CountResources > 0);
```

jogibear9988:
I'll look to create a test

sdanyliv:
@jogibear9988, it is reproducible for now?

jogibear9988:
i'll look this week

jogibear9988:
Yes, it's still failing: https://github.com/linq2db/linq2db/pull/604

PR #604, opened 2017-03-21
Test for issue #82

jogibear9988:


jogibear9988:
It seems that test is only crashing for NorthwindSqlite, but I could not see if it was also ran against "Northwind". So maybe it's only a sqlite error, maybe also one we could not solve? maybe no support from the DB Engine?

ili:
"Northwind" configuration does not run on CI i think we should configure it for Appveyor

jogibear9988:
Does Northwind Database exist in AppVeyor? (don't think so). Are there anywhere import Scripts for Northwind?

jogibear9988:
should I create a import from the sql script from here: https://northwinddatabase.codeplex.com/releases/view/71634 ?

jogibear9988:
@ili @sdanyliv I've now added Northwind from https://northwinddatabase.codeplex.com, but as you see, it's not only a sqlite problem. query does also not work on mssql

do anyone of you know how to fix the freetext errors?

ili:
Moved to #610

Issue #192, opened 2015-03-24
Data type conversion failure inside a query (version 1.7.0.1)


Lexey:
Suppose we have the following table:

```cs
[Table(Name = "csp_domain")]
public class Domain
{
    /// <summary>Domain name</summary>
    [Column(Name = "domainname", Length = 50), NotNull]
    public string DomainName { get; set; }

    [Column(Name = "verified", DataType = DataType.Char), NotNull]
    public bool Verified { get; set; }
}
```

And the following types mapping (bool is stored as Y/N in the DB):

```cs
var mapping = LinqToDB.Mapping.MappingSchema.Default;

mapping.SetConvertExpression<bool, DataParameter>(_ => DataParameter.Char(null, _ ? 'Y' : 'N'));
mapping.SetConvertExpression<char, bool>(_ => _ == 'Y');
```

Inserts work fine with this mapping. But, the following fails with "Npgsql.NpgsqlException : ERROR: 42883: operator does not exist: character = boolean"

```cs
return !db.GetTable<Domain>()
    .Any(_ => _.DomainName == domain && !_.Verified);
```

However, the following workaround works:

```cs
var dummy = false;
return !db.GetTable<Domain>()
    .Any(_ => _.DomainName == domain && _.Verified == dummy);
```

The same happens here:

```cs
// fails
db.GetTable<Domain.Domain>().Update(_ => _.DomainName.ToLower() == domain
    , _ => new Domain.Domain { Verified = true });

// works
var dummy = true;
db.GetTable<Domain.Domain>().Update(_ => _.DomainName.ToLower() == domain
    , _ => new Domain.Domain { Verified = dummy });
```

Lexey:
The problem is even more funny with Guid? <-> varchar mapping.

```
	/// <summary>Parent subscription id</summary>
	[Column(Name = "parentid", DataType = DataType.VarChar), Nullable]
	public Guid? ParentId { get; set; }
```

```
	mapping.SetConvertExpression<Guid, DataParameter>(_ => DataParameter.VarChar(null, _.ToString()));
	mapping.SetConvertExpression<string, Guid>(_ => Guid.Parse(_));
```

```
// fails with "Npgsql.NpgsqlException : ERROR: 42883: operator does not exist: character varying = uuid"
	var children = db.GetTable<Subscription>().Where(_ => _.ParentId == id).Select(_ => _.Id).ToList();

//works
[...]

PR #607, opened 2017-03-24
Fix for #192

ili:
subject


Source/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1094,6 +1094,11 @@ static bool IsQueryMember(Expression expr)
 		Expression _lastExpr1;
 		bool       _lastResult1;
 
+		bool HasConverterToDataPartameter(Type type)
```
sdanyliv: Typo here, HasConverterToDataPartamete -> HasConverterToDataParameter

ili:
Done, function inlined, some unneeded code removed

Issue #271, opened 2015-10-19
sql query generated using nvarchar instead varchar


cfuentealba:
We have sql server database where target table has float and varchar fields as PK, with its indexes...
When we query the table the sql statement generated by linq2db uses nvarchar comparison instead of varchar which causes serious performance issues as it does not use existing indexes.

example:
```sql
select * from dbo.table t
where t.field2_varchar=N'value' and
t.field2_varchar2= N'value2' and
t.field3_float=value 
```
Is there a way to force the query generation to use varchar comparison like this?
```sql
select * from dbo.table t
where t.field2_varchar='value' and
t.field2_varchar2= 'value2' and
t.field3_float=value
```

MaceWindu:
Explicit type specification for such fields should help:
[Column(DataType = DataType.Char)]

sdanyliv:
@cfuentealba, proposed solution helped?

PR #608, opened 2017-03-25
Fix for #271

ili:
In general this implementation is hack.
It would be better to make `DataType`  required parameter for `SqlValue` but it would break tons of code. And only for char\nchar in predicate. I think is does not worth.

If we would face with same situation in another query part or with other type we'll make total refactoring.


ili:
@jogibear9988  I don't think this affect performance, it's just couple of if statements.

@sdanyliv ok

ili:
@sdanyliv comment done

PR #610, opened 2017-03-26
Fix for #82

ili:
This is overriding of #604


Source/Linq/Builder/AggregationBuilder.cs
```diff
@@ -21,7 +21,7 @@ protected override bool CanBuildMethodCall(ExpressionBuilder builder, MethodCall
 
 		protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
 		{
-			var sequence = builder.BuildSequence(new BuildInfo(buildInfo, methodCall.Arguments[0]));
+			var sequence = builder.BuildSequence(new BuildInfo(buildInfo, methodCall.Arguments[0]) { CreateSubQuery = true });
```
jogibear9988: Can we, or should we, later check in Query Optimizer if SubQuery really is needed?

ili:
> Have you seen, that in my Pull request, I also added Northwind for Appveyor?

No, i didn't. I'll merge this changes too :)

Source/Linq/Builder/AggregationBuilder.cs
ili: really, can't say for sure...

PR #614, opened 2017-03-30
Bugfix for issue #612

AK107:
Bugfix for DateTimeOffset in Orace: nanoseconds was
improperly initialized from milliseconds; added full DateTimeOffset
precision (100 ns) instead of millisecond, like native DateTime.

### DIFF
--- a/Source/DataProvider/Oracle/OracleDataProvider.cs
+++ b/Source/DataProvider/Oracle/OracleDataProvider.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace LinqToDB.DataProvider.Oracle
 {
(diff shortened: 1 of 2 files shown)


sdanyliv:
Do not close. Just commit another change in that brach and **Sync** or **Push**.

Source/DataProvider/Oracle/OracleDataProvider.cs
```diff
@@ -257,15 +258,22 @@ protected override void OnConnectionTypeCreated(Type connectionType)
 				MappingSchema.AddScalarType(_oracleXmlStream, GetNullValue(_oracleXmlStream), true, DataType.Xml); // ?
 		}
 
+	    static int GetDateTimeOffsetNanoseconds(DateTimeOffset value)
+	    {
+            var tmp = new DateTimeOffset(value.Year, value.Month, value.Day, value.Hour, value.Minute, value.Second, value.Offset);
```
jogibear9988: Dont you need the miliseconds also in the constructor?

jogibear9988: okay, sorry, i see im wrong! the nanoseconds include the miliseconds

Issue #421, opened 2016-08-26
LinqExtentions Insert Blob: wrong sql


Igonik:
SQL (Oracle):

```sql
create table PERSON
(
  id    NUMBER not null,
  photo BLOB not null
)
```

C#:

```cs
    public class Person
    {
        [PrimaryKey]
        public int Id { get; set; }

        [Column]
        public byte[] Photo { get; set; }
    }

    class Program
    {
        static void Main(string[] args)
        {
            DataConnection.TurnTraceSwitchOn();
            DataConnection.OnTrace += info =>
            {
                if (info.BeforeExecute) return;
                Debug.WriteLine(info.SqlText);
            };


            using (var dataConnection = new DataConnection("default"))
            {
                                // **Success**
                dataConnection.GetTable<Person>().Insert(() => new Person() {Id = 1, Photo = new byte[] {1, 2, 3}});

                                // ** Wrong SQL: ORA-00917: missing comma ** 
                dataConnection.Into(dataConnection.GetTable<Person>())
                    .Value(p => p.Id, 1)
                    .Value(p => p.Photo, new byte[] {1, 2, 3})
                    .Insert();
            }
        }
    }
```

Generated SQL (from tracer):

```sql
--**Success**
DECLARE @p1 Blob -- Object
SET     @p1 = System.Byte[]

INSERT INTO test_Person
(
    Id,
    Photo
)
VALUES
(
    1,
    :p1
)

--** Wrong SQL: ORA-00917: missing comma ** 
INSERT INTO test_Person
(
    Id,
    Photo
)
VALUES
(
    1,
    System.Byte[]
)
```

If comment followng lines in the OracleSqlOptimizer class:

```
            new QueryVisitor().Visit(selectQuery.Select, element =>
            {
                if (element.ElementType == QueryElementType.SqlParameter)
                    ((SqlParameter)element).IsQueryParameter = false;
            });
```

All work fine: right sql is generated in the both cases.

PR #619, opened 2017-04-02
Fix for #421

ili:


ili:
@sdanyliv in general we do not need update test. Changes were done with parameters and this code is same. Also i'v done a review of all places where `SqlParameter.IsQueryParameter` is set to `false` and applied changes there. So it should be OK

sdanyliv:
@ili, possible, but when additional changes in code will appear - it sould be good if such testes will fail.

ili:
@sdanyliv additional tests for update are added :)

sdanyliv:
Great ;)

Issue #446, opened 2016-10-26
Fluent ....HasTableName(..) works only first time (version 1.0.7.4)


mr-clx:
Seems like changing mapping TableName with fluent mapping works only once per process instance.

Example:
I have a set of tables with same data models. 

```
        // Data model class (for different tables):
        public class SomeModel
        {
            [Column("id")]
            public int Id { get; set; }
            [Column("field")]
            public string Field { get; set; }
        }

        // Testing mapping same class with different tables (in different instances of DbConnection!)
        public void Test()
        {
            var confName = "Test";

            // map to table1
            using (var db = new LinqToDB.Data.DataConnection(confName))
            {
                db.MappingSchema.GetFluentMappingBuilder().Entity<ModelBase>().HasTableName("table1");
                var q = db.GetTable<ModelBase>(); // = sql=... from "table1"...
                var t = q.ToArray();  // <--- mapped to "table1" 
            }
            // map to table2
            using (var db = new LinqToDB.Data.DataConnection(confName))
            {
                db.MappingSchema.GetFluentMappingBuilder().Entity<ModelBase>().HasTableName("table2");
                var q = db.GetTable<ModelBase>(); // = sql=... from "table1"... !!
                var t = q.ToArray();  // <--- still mapped to "table1", not "table2"  !!!
            }
        }

```

I am using linq2db version 1.0.7.4
Maybe I've missed something?

sdanyliv:
Why you do not use db.GetTable<ModelName>.TableName('table1').Select... ?

mr-clx:
Thank you for this solution!
It will work for me.

But anyway, why GetFluentMappingBuilder....  does now work as expecting ?

sdanyliv:
Sorry, never tried that ;) 
But remember that there are some method to reset mapping caches. Not sure.

mr-clx:
Ok. It's ok to set table name for queries (... TableName()...)
But how can I set table name for update/insert operations?
i.e.  .... Update(MyObject)

springjazzy:
Still doesn't work with updates/deletes since Query is cached in LinqToDB.Linq.QueryRunner.Update<T> ->_queryChache dictionary. If I map entity to one table, update it, then to another all subsequent updates will be made into first mapped entity.

sdanyliv:
We will cover these Update and Inserts in this issue https://github.com/linq2db/linq2db/issues/1026

sdanyliv:
[...]

Issue #615, opened 2017-03-31
Usage of linq2db with differnet MappingSchemas


jogibear9988:
I use linq2db in a Business Apllication.

In this I have different DataConnections, with different mapped DTO's and different Mapping Schemas.

I want to know, how should I use linq2db correctly? In the Parts of the Code where I need DB Access, I create a DataConnection, and dispose when I've finished the DB Access. But what is with the MappingSchema? I think the creation of this is expensive, so I should keep it? How do I Setup a DataConnection with a existing MappingSchema?

jogibear9988:
At the moment I use 
     dataConnection.AddMappingSchema(_mappingSchema)
but it looks like this is also a expensiv call, or am I wrong?

ili:
This all depends on what you really need.
For example, you can configure:
* `MappinfSchema.Default` - this will affect all connections & all data providers
* or you can configure data provider specific mapping schema, for example: `AccessTools.GetDataProvider().MappingSchema` - this will affect all Access connections
* and you can specify per query MappingSchema with `DataConnection.AddMappingSchema()` (here we should make a point: `DataConnection.AddMappingSchema()` uses composition approach, so configuration from default schemas would be used if not overridden in your instance`

Now about performance. MappingSchema caches mapping related data to avoid using reflection. So if it is possible, if you use `DataConnection.AddMappingSchema()` it would be better to use static MappingSchema instance then you would have no performance gap (certainly only after second query, for first query caches would be builded).

jogibear9988:
Means, I have a problem, when I use linq2db with different databases of the same type, for example mssql, because i could not use one static mapping schema (for example they use different converters), but i could not cache dataconnection.
Maybe we need a mappingschema.compile? and a dataconnection constructor accepting a mapping schema?

ili:
Not really clear. May be you can provide sample, demonstrating the problem&

jogibear9988:
I mean, I've a Programm wich needs to connect to different MSSql Databases every few seconds. I use Converters in my mappings wich are incompatible, means i create a mappingschema for each database. To read data, I create a DataConnection, but I have to add the mappingschema every time to the DataConnection, so every time it consts performance, cause I can not say to the DataConnection use this already compiled Mappingschema.

sdanyliv:
@jogibear9988, it means taht you have the same DataModel but different DataConnection descedants?

jogibear9988:
No, I've not the same data Model. I've different! But Converters are Global

[...]

PR #625, opened 2017-04-10
Fix for #446

ili:
Resetting EntityDescriptor cache if EntityMappingBuilder is created


ili:
@jogibear9988 yes, this affect performance, because we need to drop cache for `EntityDescriptor`, but what we can do else? From another side building `EntityDescriptor` takes milliseconds.

If this mean I would recommend to use several static `MappingSchema`s for different mapping rules.

jogibear9988:
yes, but with the static mapping rules you hit this issue https://github.com/linq2db/linq2db/issues/615

Issue #202, opened 2015-04-15
Fluent Mapping: inheritance of Associations


akonogov:
The object of Organization contains Jur. Object MyOrganization is inherited from Organization. Jur is connected to the table OrganizationJur 1 to 1 on an id.

```cs
 MappingSchema.Default.GetFluentMappingBuilder()
                .Entity<MyOrganization>()
                .HasTableName("Clients")
                .Property(p => p.Id)
                .IsPrimaryKey()
                .Property(p => p.Jur)
                .HasAttribute(new AssociationAttribute() { ThisKey = "Id", OtherKey = "Id", CanBeNull = true })
                ...

                .Entity<OrganizationJur>()
                .HasTableName("ClientsJur")
                .Property(p => p.Id)
                .IsPrimaryKey()
```

Everything works while I don't address to object of Jur in the table Clients.

```
dc.Organizations.FirstOrDefault(o => o.Address != null);
```

```
An exception of type 'System.NullReferenceException' occurred in Account.Repository.Sql.dll but was not handled in user code
```

Having connected your project to my project and debugging it I noticed an error:
MyOrganization object mapped by me has 15 fields and 0 associations
![](http://dl1.joxi.net/drive/0003/2332/252188/150415/3d33093aaf.jpg)
And the object Orgadization has 3 associations though it wasn't described anywhere.
![](http://dl1.joxi.net/drive/0003/2332/252188/150415/2841410f80.jpg)
Respectively he looks for associationsin MyOrganization object, doesn't find and returns an error.

PR #628, opened 2017-04-11
Fix for #202

ili:


Source/Metadata/FluentMetadataReader.cs
```diff
@@ -30,7 +31,32 @@ public T[] GetAttributes<T>(MemberInfo memberInfo, bool inherit = true)
...
+			if (got)
+				return attrs.OfType<T>().ToArray();
+
+			if (inherit == false)
+				return Array<T>.Empty;
+
+			var parent    = memberInfo.DeclaringType;
+			var reflected = memberInfo.ReflectedTypeEx();
+
+			if (parent == null || parent == reflected || !parent.IsSameOrParentOf(reflected))
```
sdanyliv: Maybe here loop is needed. For collecting attributes for more nested classes.

ili: May be... first implementation was with loop, then I thought "what for? memberInfo.DeclaringType is always type where member is declared", but I think you are right, I should do loop, because we can have ClassA: ClassB: ClassC and attribute can be applied to ClassB

PR #630, opened 2017-04-12
fix for #615

jogibear9988:


ili:
@jogibear9988 this is **very** dangerous solution, you can't just do:
```cs
var ms = new MappingSchema();
using (var db = new DataConnection(ms))
{
// some code here
}
```

you should do:
```cs 
var dataProvider = GetDataProvider();
var ms = new MappingSchema(dataProvider.MappingSchema);
using (var db = new DataConnection(ms))
{
// some code here
}
```

This because each DataProvider's MappingSchema contains specific mapping settings.

I think we should consider another solution.

jogibear9988:
I know, I'll do so! But how could I use 2 different Mapping Schemas without Perf. implications without setting it in the constructor?
If I don't do like this, I need to use the "AddMapping..." call wich merges every time I'll do and costs Performance!

ili:
Performance issue is only goes to getting EntityDescriptors... May be we can make internal mapping schema constructor taking dictionary with descriptors.
I think it would be better solution.

But! It is not really safe too.

From another side, have you measured gap you are getting? May be it does not worth? Really it should matter only if you are making great number of such calls:

```Cs
var ms = new MappingSchema();
using (var db = new DataConnection(ms))
{
// some code here
}
```

jogibear9988:
I do a big number of such calls. Not the new mapping schema, but the DataConnection, is created every db access in my code, and id I could not set it in the constructor, the merge will be called every time!

jogibear9988:
Any News to this?

jogibear9988:
And maybe we can create some Method, so a Mapping Schema is compiled (or initialized), so I can do it, before any of my Code Accesses a DTO, so I get the same Performance on each call!

Issue #307, opened 2015-11-23
Protected or Private Construcotors of Entitys supported?


jogibear9988:
Doe linq2db support protected or private constructors of DB Entitys?

jogibear9988:
I think not atm.

Maybe someone can add the "feature" Label

ili:
Is it really needed?

ili:
As for me it is a bad idea to use private constructors... Does for ex EF support it, and how it looks like?

sdanyliv:
We can not support this. Our reader uses Expression Tree, which do not support such reflection tricks.

ili:
Then closing

jogibear9988:
Can I not use a Private Constructor in Expression Trees? When I get the constructor via reflection, and use https://msdn.microsoft.com/en-us/library/bb301982(v=vs.110).aspx

jogibear9988:
We support this already !!! see: https://github.com/linq2db/linq2db/pull/631

PR #631, opened 2017-04-12
test for #307

jogibear9988:


Tests/Linq/UserTests/Issue307Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+	using Model;
+
+	[TestFixture]
+	public class Issue307Tests : TestBase
+	{
+
+		[Table(Name = "Person")]
+		public class Entity307
+		{
+			private Entity307()
```
sdanyliv: @jogibear9988, check that this constructor is called. Put here some informative.

jogibear9988:
Why should it not be called? We have no other!

sdanyliv:
Just for reinsurrance. I'm not near computer, but i think your sample is not clear.

sdanyliv:
@jogibear9988, yes it works, my fault.

PR #634, opened 2017-04-15
Column equals performance regression

MaceWindu:
During development environment setup I've discovered that following test: Linq.JoinTests.StackOverflow.SqlServer.2008 test (JoinTests.StackOverflow)
has a poor performance. It countains 100 iterations with 20 iterations taking 9 seconds to execute and each iteration adds 100% to this time resulting in about 345 000 000 000 000 000 years estimation.
This PR contains two fixes:
1. Removal of IndexOf calls in SelectQuery.SelectClause class for Columns collection, as it is not required and only leads to excessive calls to Columns.Equals. This reduced aforementioned test execution time by 50%.
2. Removal of Expression.Equals call from Columns.Equals that was added by this commit https://github.com/linq2db/linq2db/commit/6622ed36c4c19ceb9ca1492e9f03926942069e05. @ili could you confirm that this code was added only as optimization and could be removed? Also I think if it was perfomance fix - we need a test that demonstrate it. Looks like it related to Issue508Test set of tests but they run faster with this PR.

After fix:
- StackOverflow test runs only 1 second
- both fixes deducted other tests execution time by several seconds in total each
- no new failed tests spotted

BTW, I don't know why this test marked to run only for SqlServer2008 provider, but looks like this is why this regression wasn't spotted sooner as AppVeyor configured only for SqlServer2012.


ili:
I do know about this issue, irl it is fixed in #584... But I'm expecting Igor's review.

MaceWindu:
Good, I've removed my "fix" for Equals so now this PR contains only IndexOf calls removal and also disable broken test temporary.

ili:
I'v created #638 to fix this problem without waiting #584 review

PR #638, opened 2017-04-21
this fixes Column.Equals performance

ili:
if we do have really big query calling QuetyVisitor.Find() takes too much time.
And IRL we do need only to compare the most low level expressions for two columns


ili:
>Looks like these changes is also part of #584 PR

yes, you are right, this is performance critical so I'v done separate fix

PR #640, opened 2017-04-23
#615 perfomance fix

### LINKED ISSUE #615: Usage of linq2db with differnet MappingSchemas

ili:
#615 Moving EntityDescriptors cache to MappingSchemaInfo to avoid performance gap on using many MappingSchemas

### DIFF
--- a/Source/Mapping/ColumnDescriptor.cs
+++ b/Source/Mapping/ColumnDescriptor.cs
@@ -16,7 +16,6 @@ namespace LinqToDB.Mapping
 	{
 		public ColumnDescriptor(MappingSchema mappingSchema, ColumnAttribute columnAttribute, MemberAccessor memberAccessor)
 		{
-			MappingSchema  = mappingSchema;
 			MemberAccessor = memberAccessor;
 			MemberInfo     = memberAccessor.MemberInfo;
 
@@ -118,7 +117,6 @@ namespace LinqToDB.Mapping
 			}
 		}
 
-		public MappingSchema  MappingSchema   { get; private set; }
 		public MemberAccessor MemberAccessor  { get; private set; }
 		public MemberInfo     MemberInfo      { get; private set; }
 		public MemberInfo     StorageInfo     { get; private set; }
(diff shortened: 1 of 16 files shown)


Source/Mapping/MappingSchemaInfo.cs
```diff
@@ -239,5 +239,42 @@ public void SetDefaultFromEnumType(Type enumType, Type defaultFromType)
...
+		}
+
+		/// <summary>
+		/// Enumerate types for cached <see cref="EntityDescriptor"/>s
+		/// </summary>
+		/// <seealso cref="GetEntityDescriptor"/>
+		/// <returns><see cref="IEnumerable{T}"/></returns>
+		public IEnumerable<Type> GetEntites()
+		{
+			return _entityDescriptors.Keys;
```
sdanyliv: Better return list here. It is thread safe.

Source/Mapping/ColumnDescriptor.cs
```diff
@@ -16,7 +16,6 @@ public class ColumnDescriptor
 	{
 		public ColumnDescriptor(MappingSchema mappingSchema, ColumnAttribute columnAttribute, MemberAccessor memberAccessor)
```
sdanyliv: Something wrong here. MappingSchema used for initialization.

ili: Yes, only for initialization

sdanyliv: Hipothetic situation, other mapping schema changes DbType and DataType for ColumnDescriptor - it will not broke something?

ili: Yes, but this change does not apply this, without this change the situation is the same.

sdanyliv:
@jogibear9988, please confirm that this solution works for you.

jogibear9988:
I think it should, looks good.
In my production application I could only test if this is merged (because if I only replace the linq2db dll's it will not run)

jogibear9988:
merge?

jogibear9988:
tested and works!

PR #641, opened 2017-04-23
Fixes for tests

MaceWindu:
Some fixed for unstable tests, detected during setup of dev-environment. Only for databases I was able to run for now and only for full .NET framework.
- specified collation explicitly for legacy text/ntext columns in MSSQL script, otherwise it will fail if default server collation has _SC flag and not overriden on database level
- load MSSQL spatial types libraries explicitly for spatial types tests
- removed hardcoded database names and replaced with database name loading in test itself where it is used by tests to make tests independent from database name
- updated MySql tests to ignore fractional seconds for versions prior to 5.6.4 for DATETIME column in expected data
- fixed Oracle local timezone test to use proper offset for any local timezone (that existed at that time)
- updated assert for XmType Oracle column to work for all providers
- added test file for Oracle BFILE column to repository
- fixed Firebird database script to instert non-ASCII test data propery as Firebird doesn't support non-ascii literals
- reenabled StackOverflow test as fix was already merged into master

Untested due to issues with providers:
- Firebird 3.0 (only 2.5) - looks like repository contains old provider without 3.0 suport
- Oracle Native - provider in repository too old to attach to latest Oracle XE
- DB2 & Informix - providers fail

Just not tested:
- SQL Azure
- SAP HANA - don't have time to play with VMWare for now as SAP HANA Express image conversion to Hyper-V produce unbootable image
- ASE - wasn't able to connect to server from provider (?)

Not fixed:
- DataProvider.SqlServerTests.CreateAlltypes.SqlServer.2012/DataProvider.SqlServerTests.CreateAlltypes.SqlServer.2014 tests to work on database with _SC collation flag - requires collation support in mapping schema and CreateTable method 

I plan to work on issues next and will return to tests later including .net core framework testing and dev-environment setup document.


sdanyliv:
@MaceWindu, imported SapHana VM to VirtualBox. Looks like it works.

PR #664, opened 2017-05-18
Fix for missing dictionary datatype

breakingthings:
Latest changes break Sql functions using dicitionary fields (postgresql hstore).


sdanyliv:
@breakingthings, please use leading tabs in your editor.

Issue #673, opened 2017-05-21
Regression with BinaryAggregateExpression


MaceWindu:
@sdanyliv , looks like another issue like #667. After merge with master one of my tests started to fail with following error:
```
Test Name:	Merge.MergeTests.Issue200InUpdate.SqlServer.2014
Test FullName:	Tests.Merge.MergeTests.Issue200InUpdate("SqlServer.2014")
Test Source:	C:\GitHub\linq2db\Tests\Linq\Update\MergeTests.Issues.cs : line 231
Test Outcome:	Failed
Test Duration:	0:00:00,01

Result StackTrace:	
at lambda_method(Closure , Expression , Object[] )
   at LinqToDB.Linq.Query`1.SetParameters(Expression expr, Object[] parameters, Int32 idx) in C:\GitHub\linq2db\Source\Linq\Query.cs:line 412
   at LinqToDB.Linq.Builder.ContextParser.Context.<>c__DisplayClass3_0`1.<BuildQuery>b__0() in C:\GitHub\linq2db\Source\Linq\Builder\ContextParser.cs:line 49
   at LinqToDB.DataProvider.BasicMergeBuilder`2.GeneratePredicateByTargetAndSource(Expression`1 predicate) in C:\GitHub\linq2db\Source\DataProvider\BasicMergeBuilder.cs:line 176
   at LinqToDB.DataProvider.BasicMergeBuilder`2.GenerateMatch() in C:\GitHub\linq2db\Source\DataProvider\BasicMergeBuilder.cs:line 478
   at LinqToDB.DataProvider.BasicMergeBuilder`2.GenerateCommand() in C:\GitHub\linq2db\Source\DataProvider\BasicMergeBuilder.cs:line 538
   at LinqToDB.DataProvider.BasicMergeBuilder`2.BuildCommand() in C:\GitHub\linq2db\Source\DataProvider\BasicMergeBuilder.cs:line 892
   at LinqToDB.DataProvider.DataProviderBase.Merge[TTarget,TSource](DataConnection dataConnection, IMerge`2 merge) in C:\GitHub\linq2db\Source\DataProvider\DataProviderBase.cs:line 449
   at LinqToDB.Data.MergeExtensions.Merge[TTarget,TSource](IMerge`2 merge) in C:\GitHub\linq2db\Source\Data\Merge.cs:line 504
[...]

PR #675, opened 2017-05-22
Fixed regression with BinaryAggregateExpression when working with parameters

sdanyliv:
Fixes issue #673


ili:
No test to reproduce #673?

sdanyliv:
It's something new from @MaceWindu, really reproduced only using his branch. I've not foud simple way to extract test.

ili:
@MaceWindu i'll merge this PR, and it will be great if you'll be able to write unit test later

MaceWindu:
Will do

MaceWindu:
Sorry, looks like there will be no test for it.

When I tried to reproduce it without Merge, I found what caused this issue:  ContextParser.Context.SetParameters method was passing Builder.Expression (with BinaryAggregateExpression) to SetParameters instead of query.Expression. Changing it fixed problem too.

Because this class used only by old and new Merge API and I fixed it - I don't see another way to reproduce it anymore.

Anyway, I think it was a good improvement.

ili:
OK, let it be so )

Issue #676, opened 2017-05-22
Wrong SQL in Join Statement


jogibear9988:
See following joins:
```cs
		[Test, NorthwindDataContext]
		public void Test(string context)
		{
			using (var db = new NorthwindDB(context))
			{
				DataConnection.TurnTraceSwitchOn(TraceLevel.Verbose);
				DataConnection.WriteTraceLine += (a, b) => { System.Diagnostics.Debug.WriteLine(a + " -> " + b); };
				var jj = from o in db.Order
					join c in db.Customer on o.CustomerID equals c.CustomerID into cg
					from c in cg.DefaultIfEmpty().Take(1)
					select new {o, c};

				var res = jj.ToList();

				var jj2 = from o in db.Order
					join c in db.Customer.Take(1) on o.CustomerID equals c.CustomerID into cg
					from c in cg.DefaultIfEmpty()
					select new { o, c };

				var res2 = jj2.ToList();
			}
		}
```
Both Querys create the same SQL, wich I think is not correct. Don't know if the first one is possible in SQL.

Generated SQL:

```sql
	DataConnection: -- NorthwindSqlite SQLite
	SELECT
		[o].[OrderID],
		[o].[CustomerID],
		[o].[EmployeeID],
		[o].[OrderDate],
		[o].[RequiredDate],
		[o].[ShippedDate],
		[o].[ShipVia],
		[o].[Freight],
		[o].[ShipName],
		[o].[ShipAddress],
		[o].[ShipCity],
		[o].[ShipRegion],
		[o].[ShipPostalCode],
		[o].[ShipCountry],
		[t1].[CustomerID1] as [CustomerID11],
		[t1].[CompanyName] as [CompanyName1],
		[t1].[ContactName] as [ContactName1],
		[t1].[ContactTitle] as [ContactTitle1],
		[t1].[Address] as [Address1],
		[t1].[City] as [City1],
		[t1].[Region] as [Region1],
		[t1].[PostalCode] as [PostalCode1],
		[t1].[Country] as [Country1],
		[t1].[Phone] as [Phone1],
		[t1].[Fax] as [Fax1]
	FROM
		[Orders] [o]
			LEFT JOIN (
				SELECT
					[cg].[CustomerID] as [CustomerID1],
					[cg].[CompanyName],
					[cg].[ContactName],
					[cg].[ContactTitle],
					[cg].[Address],
					[cg].[City],
					[cg].[Region],
					[cg].[PostalCode],
[...]

PR #684, opened 2017-05-25
test for #676

jogibear9988:


jogibear9988:
Look at results lres and res with Linq2DB Customer is null with linq2objects it works

ili:
actually implemented in #701

Issue #241, opened 2015-07-24
ArgumentException on dynamic .Contans (LIKE) and GroupBy  


georgi-atanassov:
### How to reproduce

Model:

>   [Table("t1")]
>   public partial class t1
>   {
>       [Column,     NotNull ] public string text { get; set; } // varchar(255)
>   }

and the following code 

``` c#
string str = "str to find";
var q = from x in db.t1 select x;
if(shouldFindText)
   q = q.Where(_ => _.text.Contains(str));
var t1g = q.GroupBy(_ => _).ToDictionary(_ => _.Key, _ => _.ToList());
```

The exception is thrown as follows:

> System.ArgumentException was caught
>  HResult=-2147024809
>  Message=An item with the same key has already been added.
>  Source=mscorlib
>  StackTrace:
>       at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
>       at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
>       at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
>       at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
>       at LinqToDB.Linq.Builder.GroupByBuilder.GroupByContext.GroupByHelper`3.GetGrouping(GroupByContext context) in \linq2db-master\Source\Linq\Builder\GroupByBuilder.cs:line 213
>       at LinqToDB.Linq.Builder.GroupByBuilder.GroupByContext.BuildGrouping() in \linq2db-master\Source\Linq\Builder\GroupByBuilder.cs:line 328
>       at LinqToDB.Linq.Builder.GroupByBuilder.GroupByContext.BuildExpression(Expression expression, Int32 level) in \linq2db-master\Source\Linq\Builder\GroupByBuilder.cs:line 338
>       at LinqToDB.Linq.Builder.SequenceContextBase.BuildQuery[T](Query`1 query, ParameterExpression queryParameter) in \linq2db-master\Source\Linq\Builder\SequenceContextBase.cs:line 37
> ...
>  InnerException: 
### RCA

[...]

PR #690, opened 2017-05-28
Test for #241

jogibear9988:


sdanyliv:
@ili, tests for 676 should be removed.  They have extended and moved to JoinToLimitedTests.

Issue #693, opened 2017-05-29
Nullable Enum searching for wrong converter


jogibear9988:
If I have a Nullable Enum in my DTO object, I can use following converters (for save):

```cs
            MappingSchema.Default.SetConverter<Key?, string>((obj) =>
            {
                if (obj != null)
                    return obj.ToString();
                return null;
            });
            MappingSchema.Default.SetConverter<Key?, DataParameter>((obj) =>
            {
                if (obj != null)
                    return new DataParameter {Value = obj.ToString()};
                return new DataParameter { Value = DBNull.Value };
            });
```
but to load from the Database the following is not enough:
```cs
            MappingSchema.Default.SetConverter<string, Key?>((txt) =>
            {
                if (string.IsNullOrEmpty(txt))
                    return null;
                return (Key)Enum.Parse(typeof(Key), txt, true);
            });
```
I also need 
```cs
            MappingSchema.Default.SetConverter<string, Key>((txt) =>
            {
                return (Key)Enum.Parse(typeof(Key), txt, true);
            });
```
I think the Nullable Converter should be used!

sdanyliv:
@jogibear9988, can you handle it yourself?

jogibear9988:
i'll take a look

jogibear9988:
https://github.com/linq2db/linq2db/pull/694

ili:
#694 merged

PR #694, opened 2017-05-29
fix issue with nullable enum #693

jogibear9988:


Source/Expressions/ConvertFromDataReaderExpression.cs
```diff
@@ -70,7 +70,7 @@ public Expression Reduce(IDataReader dataReader)
 				if (mapType != ex.Type)
 				{
 					// Use only defined convert
-					var econv = mappingSchema.GetConvertExpression(ex.Type, toType, false, false);
+					var econv = mappingSchema.GetConvertExpression(ex.Type, type, false, false);
```
sdanyliv: Maybe we have to use variant with **toType** when variant with **type** returns null?

jogibear9988:
i also thought about this, but in the other direction we also only search for the nullable type!

PR #697, opened 2017-05-31
Fix #695

### LINKED ISSUE #695: SqlCe schema provider PK duplications causes error on FK generation
Caused by recent change 88c0e54b4dcf95bd5c3951bbc2834568300cb252

These changes made it possible to retrieve FKs correctly but the PKs are duplicated. It is a really old piece of code and was no problem as previously the FKs were just an empty list.

Inside `SchemaProviderBase.GetSchema()` FKs are joined onto PKs and later processed, but PKs contain normal index columns as well as real PKs...and thus causes error on otherColumn selection as `.Single()` fails on multiple results:
`otherColumn = (from c in otherTable.Columns where c.ColumnName == fk.OtherColumn select c).Single();`

This is a huge problem for us as we use SqlCe for integration testing and now the whole pipeline is broken.

ili:
Work in process,
Tests & branch for #695


Tests/Linq/UserTests/Issue445Tests.cs
```diff
@@ -90,6 +90,21 @@ public void ObjectDisposedException1(string context)
 			});
 		}
 
+		[Test, IssueContextSourceAttribute]
+		private IEnumerable<Person> GetPersonsFromDisposed3(string context)
```
MaceWindu: nunit do not support private tests

PR #717, opened 2017-06-06
Experimental non recursive enumerator

sdanyliv:
@igor-tkachev, I've added experimental non recursive enumerator. And I've implemented alternative EqualsTo method for expressions.

No idea how to test performance regression in that case.

@detoxhby, could you please check performance regression on your test environment?


sdanyliv:
Non relevant, so closing.

PR #720, opened 2017-06-07
implement regex

jogibear9988:


Source/Sql/Sql.cs
```diff
@@ -1065,6 +1065,17 @@ public static bool FreeText(object table, string text)
...
 
+		[Sql.Expression(ProviderName.MySql, "{0} REGEXP {1}")]
+		[Sql.Expression(ProviderName.SQLite, "{0} REGEXP {1}")]
+		[Sql.Expression(ProviderName.PostgreSQL, "{0} SIMILAR TO {1}")]
+		[Sql.Expression(ProviderName.Oracle, "REGEXP_LIKE ({0}, {1})")]
+		[Sql.Expression(ProviderName.SapHana, "{0} LIKE_REGEXPR {1}")]
+		[Sql.Expression(ProviderName.SqlServer, "dbo.RegexMatch({0}, {1})")]
+		public static bool RegExp(string data, string regex)
+		{
+			return new Regex(regex).IsMatch(data);
```
sdanyliv: I think it should be ServerSide method, otherwise patterns must match.

Tests/Linq/UserTests/Issue698Tests.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+	[TestFixture]
+	public class Issue698Tests : TestBase
+	{
+		[Test]
+		public void Test(string context)
+		{
+			using (var db = new NorthwindDB(context))
+			{
+				var zz =
```
sdanyliv: Test should be rewritten and moved to Test with Sql.Like test.
Also you have to test how it work in combination with other predicates.

Tests/Linq/UserTests/Issue698Tests.cs
```diff
@@ -0,0 +1,29 @@
...
+using NUnit.Framework;
+
+using Tests.Model;
+
+namespace Tests.UserTests
+{
+	[TestFixture]
+	public class Issue698Tests : TestBase
+	{
+		[Test]
```
sdanyliv: Include supported DataContexts

Source/Sql/Sql.cs
jogibear9988: Thats what I'm trying to support.

ili: Yes, in this implementation it should throw `NotImplementedException` Also XML doc required where should be written that `regex` should be written within database dialect.

sdanyliv: I think efforts do not worth expectations.

ili:
Different databases has different regex dialect, am I right? Or one standard is used?

jogibear9988:
I'll continue work on this, but I will not have time in the next weeks...

Issue #723, opened 2017-06-09
CreateTable vs Oracle issues


Lexey:
There are two issues with the *CreateTable<T>* method.
1) The 4th parameter name *ownerName* is confusing. It should be named *schemaName*.
2) When an explicit schema name is set via the FluentMappingBuilder or via the ownerName parameter it is being respected only for the CREATE TABLE statement. But related CREATE SEQUENCE and CREATE TRIGGER statements have no schema specifications.

A table definition:
```c#
    [Table("FactorList")]
    public class FactorList
    {
        /// <summary>Id</summary>
        [Column, PrimaryKey, Identity, NotNull]
        public int Id { get; set; }
        /// <summary>Name</summary>
        [Column(Length = 255), Nullable]
        public string Name { get; set; }

        #region Associations
        [Association(ThisKey = nameof(Id), OtherKey = nameof(FactorListFactor.FactorListId), CanBeNull = true
            , IsBackReference = true)]
        [NotForeignKey]
        public IEnumerable<FactorListFactor> Factors { get; set; }
        #endregion Associations
    }
```

The create call (1st version):
```c#
MappingSchema.GetFluentMappingBuilder().Entity<FactorList>().HasSchemaName(schemaName);
CreateTable<FactorList>();
```

The create call (2nd version):
```c#
CreateTable<FactorList>(null, null, schemaName);
```

Resulting SQL:

```sql
CREATE TABLE MySchema.FactorList
(
	Id   Int            NOT NULL,
	Name VarChar2(255)      NULL,

	CONSTRAINT PK_FactorList PRIMARY KEY (Id)
)

CREATE SEQUENCE SIDENTITY_FactorList

CREATE OR REPLACE TRIGGER  TIDENTITY_FactorList
BEFORE INSERT ON FactorList FOR EACH ROW
BEGIN
	SELECT SIDENTITY_FactorList.NEXTVAL INTO :NEW.Id FROM dual;
END;
```

### Environment details
linq2db version: *1.8.2*  
Database Server: *Oracle 11g*
Database Provider: *Oracle.Managed*  
Operating system: *Windows 7 SP1*  
Framework version: *.NET Framework 4.6.2*

ili:
@Lexey PR would be great! )))

Lexey:
:) May be later. At this moment I have no time to dig into the 2nd issue.

PR #724, opened 2017-06-10
Fix for #723

ili:
Fix for #723 
Firebird creates trigger too, but it does not support schemas.


Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -1838,5 +1838,32 @@ public void Issue539(string context)
...
+		public void Issue723(string context)
+		{
+			using (var db = new TestDataConnection(context))
+			{
+				try
+				{
+					db.Execute("GRANT CREATE ANY TRIGGER TO TestUser");
+					db.Execute("CREATE USER Issue723Schema IDENTIFIED BY password");
+					db.CreateTable<Issue723Table>(schemaName: "Issue723Schema");
+				}
```
MaceWindu: Could you add assert that LastQuery contains proper name?

ili: You mean `Assert.That(db.LastQuery.Contains("Issue723Schema.Issue723Table"))`?

MaceWindu: Yes, or by quering schema information from server

Lexey:
It looks like it fixes the missing schema for the CREATE TRIGGER. But, as it has been said in #723 the same problem exists for the CREATE SEQUENCE.

ili:
@Lexey `CREATE SEQUENCE` does not refers table.. or you do mean we should change sequence name?

Lexey:
@ili Yes, I mean that the sequence name should the prefixed with the schema name.

Source/DataExtensions.cs
```diff
@@ -266,33 +266,33 @@ public static int Delete<T>([NotNull] this IDataContext dataContext, T obj)
 		public static ITable<T> CreateTable<T>([NotNull] this IDataContextInfo dataContextInfo,
 			string         tableName       = null,
 			string         databaseName    = null,
-			string         ownerName       = null,
+			string         schemaName      = null,
```
sdanyliv: Why you have removed owner name?
[Owner and Schma](http://www.sqlteam.com/article/understanding-the-difference-between-owners-and-schemas-in-sql-server)

Lexey: I have requested this rename. It is really the schema name. ownerName just confuses a potential user of the method.

ili:
@Lexey @sdanyliv @MaceWindu can anybody check last changes with oracle? I do have strange problems with DBHost VM (it does not see network) so I can't do it myself :(

MaceWindu:
@ili I will check it today

MaceWindu:
ok, managed to test it finally. I think this is not what @Lexey  expected:
```sql
CREATE SEQUENCE SIDENTITY_Issue723SchemaIssue723Table
Message  : ORA-00972: identifier is too long
```
I think it should be something like
```sql
CREATE SEQUENCE Issue723Schema.SIDENTITY_Issue723Table
```

BTW, I've updated test a bit to work more stable:
```
[Test, OracleDataContext]
		public void Issue723Test(string context)
		{
			using (var db = new TestDataConnection(context))
			{
				var currentUser = db.Execute<string>("SELECT user FROM dual");
				db.Execute("GRANT CREATE ANY TRIGGER TO " + currentUser);

				db.Execute("CREATE USER Issue723Schema IDENTIFIED BY password");

				try
				{
					var tableSpace = db.Execute<string>("SELECT default_tablespace FROM sys.dba_users WHERE username = 'ISSUE723SCHEMA'");
					db.Execute("ALTER USER Issue723Schema quota unlimited on {0}".Args(tableSpace));

					db.CreateTable<Issue723Table>(schemaName: "Issue723Schema");

					try
					{
						for (var i = 1; i < 3; i++)
						{
							var id = Convert.ToInt32(db.InsertWithIdentity(new Issue723Table() { StringValue = i.ToString() }));
							Assert.AreEqual(i, id);
						}
						Assert.That(db.LastQuery.Contains("Issue723Schema.Issue723Table"));
					}
					finally
					{
						db.DropTable<Issue723Table>(schemaName: "Issue723Schema");
					}
				}
				finally
				{
					db.Execute("DROP USER Issue723Schema CASCADE");
				}
			}
		}
```

Lexey:
@MaceWindu, you are right. That is not I have expected. :)
Even after an obvious bugfix of GetFullYabelName
```c#
return table.Owner != null ? table.Owner + '.' + table.PhysicalName : table.PhysicalName;
```
it produces wrong results. 
Really the name generation code should be like:
```c#
				var table = SelectQuery.CreateTable.Table;
				StringBuilder
					.Append("DROP TRIGGER ");
				if (table.Owner != null) StringBuilder.Append(table.Owner).Append('.');
				StringBuilder
					.Append("TIDENTITY_")
					.Append(table.PhysicalName)
					.AppendLine();
```

ili:
@Lexey & @MaceWindu what about now? :)

Source/DataProvider/Oracle/OracleSqlBuilder.cs
```diff
@@ -288,6 +288,11 @@ public override int CommandCount(SelectQuery selectQuery)
 			return base.CommandCount(selectQuery);
 		}
 
+		private static string GetFullTableName(SqlTable table)
```
MaceWindu: Looks like this method is not used anymore

Source/DataProvider/Oracle/OracleSqlBuilder.cs
```diff
@@ -297,7 +302,9 @@ protected override void BuildDropTableStatement()
 			else
 			{
 				StringBuilder
-					.Append("DROP TRIGGER TIDENTITY_")
+					.Append("DROP TRIGGER ")
+					.Append(SelectQuery.CreateTable.Table.Owner)
+					.Append(".TIDENTITY_")
```
MaceWindu: It looks like without owner, FQN will start with dot. Will Oracle handle it properly?
I think we should add second test that perfroms same actions without schema name

Source/DataProvider/Oracle/OracleSqlBuilder.cs
```diff
@@ -310,7 +317,9 @@ protected override void BuildCommand(int commandNumber)
 				if (commandNumber == 1)
 				{
 					StringBuilder
-						.Append("DROP SEQUENCE SIDENTITY_")
+						.Append("DROP SEQUENCE ")
+						.Append(SelectQuery.CreateTable.Table.Owner)
+						.Append(".SIDENTITY_")
```
MaceWindu: same question about schema-less case

Source/DataProvider/Oracle/OracleSqlBuilder.cs
```diff
@@ -322,19 +331,26 @@ protected override void BuildCommand(int commandNumber)
...
 						.Append(SelectQuery.CreateTable.Table.PhysicalName)
 						.AppendLine();
 				}
 				else
 				{
 					StringBuilder
-						.AppendFormat("CREATE OR REPLACE TRIGGER  TIDENTITY_{0}", SelectQuery.CreateTable.Table.PhysicalName)
-						.AppendLine  ()
-						.AppendFormat("BEFORE INSERT ON {0} FOR EACH ROW", SelectQuery.CreateTable.Table.PhysicalName)
+						.AppendFormat("CREATE OR REPLACE TRIGGER {0}.TIDENTITY_{1}", SelectQuery.CreateTable.Table.Owner, SelectQuery.CreateTable.Table.PhysicalName)
```
MaceWindu: same question about schema-less case

Source/DataProvider/Oracle/OracleSqlBuilder.cs
```diff
@@ -322,19 +331,26 @@ protected override void BuildCommand(int commandNumber)
...
+						.AppendFormat("BEFORE INSERT ON ");
+
+					BuildPhysicalTable(SelectQuery.CreateTable.Table, null);
+
+					StringBuilder
+						.AppendLine(" FOR EACH ROW")
 						.AppendLine  ()
 						.AppendLine  ("BEGIN")
-						.AppendFormat("\tSELECT SIDENTITY_{1}.NEXTVAL INTO :NEW.{0} FROM dual;", _identityField.PhysicalName, SelectQuery.CreateTable.Table.PhysicalName)
+						.AppendFormat("\tSELECT {2}.SIDENTITY_{1}.NEXTVAL INTO :NEW.{0} FROM dual;", _identityField.PhysicalName, SelectQuery.CreateTable.Table.PhysicalName, SelectQuery.CreateTable.Table.Owner)
```
MaceWindu: same question about schema-less case

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -1838,5 +1838,62 @@ public void Issue539(string context)
...
+				db.Execute("CREATE USER Issue723Schema IDENTIFIED BY password");
+
+				try
+				{
+
+					var tableSpace = db.Execute<string>("SELECT default_tablespace FROM sys.dba_users WHERE username = 'ISSUE723SCHEMA'");
+					db.Execute("ALTER USER Issue723Schema quota unlimited on {0}".Args(tableSpace));
+
+					db.CreateTable<Issue723Table>(schemaName: "Issue723Schema");
+					Assert.That(db.LastQuery.Contains("Issue723Schema.Issue723Table"));
```
MaceWindu: move assert to try, so drop table will be called even if assert failed

Source/DataProvider/Oracle/OracleSqlBuilder.cs
Lexey: AFAIR, it will fail in this case

ili:
@Lexey can you make a review please? :)

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -1839,6 +1839,73 @@ public void Issue539(string context)
...
+
+				try
+				{
+
+					var tableSpace = db.Execute<string>("SELECT default_tablespace FROM sys.dba_users WHERE username = 'ISSUE723SCHEMA'");
+					db.Execute("ALTER USER Issue723Schema quota unlimited on {0}".Args(tableSpace));
+					try
+					{
+
+						db.CreateTable<Issue723Table>(schemaName: "Issue723Schema");
```
MaceWindu: Let's move CreateTable before ```try``` so it will not try to delete it in finally when create failed

PR #726, opened 2017-06-10
Nullability fixes in enum mapping

MaceWindu:


jogibear9988:
what does not work?

MaceWindu:
Check added tests (you can try run them without fixes and see results) - there were several issues with nullable enum fields and null in mappings:
1. duplicate mapping values exception when enum mapped to nullable value type (e.g. int? ) and have 0 and null in mapped values
2. default(type) instead of default(type?) used as default value for nullable enum mapped to nullable value type
3. in general if enum mapped to value type had mapping to null - it produced exception

With this change mapping to value type with null involved behaves same as mapping to reference type (e.g. string)

ili:
why 1.9.0? you suppose it is not ready to merge?

Source/Common/ConvertBuilder.cs
```diff
@@ -627,7 +638,7 @@ select a
 					?? mappingSchema.GetDefaultFromEnumType(typeof(Enum))
 					?? Enum.GetUnderlyingType(type);
 
-			if (enumType.IsNullable() && !defaultType.IsClassEx() && !defaultType.IsNullable())
+			if ((enumType.IsNullable() || fields.Any(attrs => attrs.Count != 0 && attrs[0].Value == null)) && !defaultType.IsClassEx() && !defaultType.IsNullable())
```
sdanyliv: Can you describe this change?

MaceWindu:
I assigned only urgent defects and already approved fixes to 1.8.3. If we will have anything else ready to release with it - I have no objection too.

Source/Common/ConvertBuilder.cs
MaceWindu: good question... I think I need to add more tests to see if it introduce change of behavior for some mapping cases.

As for this change it fixes following cases:

enum->db mapping for following case:
```cs
enum Enum
{
// explicit mapping to null
    [MapValue(null)] Value
}

// non-nullable field
Enum field;

// old behavior - mapped to default(int) = 0
// new behavior - mapped to default(int?) = null
```

and for db-> enum mapping it fixes exception for this case:

```cs
enum Enum
{
    [MapValue(0)] Value1,
    [MapValue(null)] Value2
}
```
```
Mapping ambiguity. MapValue(0) attribute is defined for both 'Enum.Value1' and 'Enum.Value2'.
```

MaceWindu:
Ok, I've reviewed enum mapping functionality once more and don't see issues with this PR.
As a bonus - added xml comments to MapValueAttribute as a part of our documentation efforts

PR #727, opened 2017-06-10
fix CHAR(X)/NCHAR(X) padding trimming

MaceWindu:
This PR addresses issue with padding trimming when we read data into char or string field. Current implementation uses string.TrimEnd, which trims about 20 different characters, but we need to trim only space (0x20).
For mapping to char field current implementation uses same trimming logic as for mapping to string which could result in empty string if database returns trimmable character. It results in '\0' character in mapped field.

Remaining issues:
- LinqService trimming tests disabled as they fail for providers with trimming logic, because right now we don't use it in LinqService. This is a known issue and will not be fixed in this PR.

### DIFF
--- a/Source/DataProvider/Access/AccessDataProvider.cs
+++ b/Source/DataProvider/Access/AccessDataProvider.cs
@@ -34,7 +34,7 @@ namespace LinqToDB.DataProvider.Access
 			SqlProviderFlags.IsInnerJoinAsCrossSupported = false;
 
 			SetCharField("DBTYPE_WCHAR", (r,i) => r.GetString(i).TrimEnd(' '));
-			SetCharFieldToType<char>("DBTYPE_WCHAR", (r, i) => AccessTools.GetChar(r, i));
+			SetCharFieldToType<char>("DBTYPE_WCHAR", (r, i) => DataTools.GetChar(r, i));
 
 			SetProviderField<IDataReader,TimeSpan,DateTime>((r,i) => r.GetDateTime(i) - new DateTime(1899, 12, 30));
 			SetProviderField<IDataReader,DateTime,DateTime>((r,i) => GetDateTime(r, i));
--- a/Source/DataProvider/Access/AccessDataProvider.cs
+++ b/Source/DataProvider/Access/AccessDataProvider.cs
@@ -33,7 +33,8 @@ namespace LinqToDB.DataProvider.Access
 			SqlProviderFlags.IsCrossJoinSupported        = false;
 			SqlProviderFlags.IsInnerJoinAsCrossSupported = false;
 
-			SetCharField("DBTYPE_WCHAR", (r,i) => r.GetString(i).TrimEnd());
+			SetCharField("DBTYPE_WCHAR", (r,i) => r.GetString(i).TrimEnd(' '));
+			SetCharFieldToType<char>("DBTYPE_WCHAR", (r, i) => AccessTools.GetChar(r, i));
 
 			SetProviderField<IDataReader,TimeSpan,DateTime>((r,i) => r.GetDateTime(i) - new DateTime(1899, 12, 30));
 			SetProviderField<IDataReader,DateTime,DateTime>((r,i) => GetDateTime(r, i));
--- a/Tests/Linq/Linq/CharTypesTests.cs
+++ b/Tests/Linq/Linq/CharTypesTests.cs
@@ -239,7 +239,7 @@ namespace Tests.Linq
 			new CharTestTable()
 		};
 
-		[DataContextSource(false	)]
+		[DataContextSource(false)]
 		public void CharTrimming(string context)
 		{
 			using (var db = GetDataContext(context))
--- a/Tests/Linq/Linq/CharTypesTests.cs
+++ b/Tests/Linq/Linq/CharTypesTests.cs
@@ -190,6 +190,8 @@ namespace Tests.Linq
 				|| context == ProviderName.PostgreSQL + ".LinqService"
 				|| context == ProviderName.DB2
 				|| context == ProviderName.DB2 + ".LinqService"
+				|| context == ProviderName.SQLite
+				|| context == ProviderName.SQLite + ".LinqService"
 				|| context == ProviderName.SqlCe
 				|| context == ProviderName.SqlCe + ".LinqService"
 				|| context == ProviderName.SapHana
--- a/Tests/Linq/Linq/CharTypesTests.cs
+++ b/Tests/Linq/Linq/CharTypesTests.cs
@@ -24,23 +24,23 @@ namespace Tests.Linq
 			[Column("id", Configuration = ProviderName.PostgreSQL)]
 			public int Id;
 
-			[Column(Configuration = ProviderName.SqlCe, IsColumn = false)]
-			[Column(Configuration = ProviderName.DB2, IsColumn = false)]
-			[Column(Configuration = ProviderName.PostgreSQL, IsColumn = false)]
-			[Column(Configuration = ProviderName.MySql, IsColumn = false)]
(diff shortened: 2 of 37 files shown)


Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+			[Column("id", Configuration = ProviderName.PostgreSQL)]
+			public int Id;
+
+			[Column(Configuration = ProviderName.SqlCe, IsColumn = false)]
+			[Column(Configuration = ProviderName.DB2, IsColumn = false)]
+			[Column(Configuration = ProviderName.PostgreSQL, IsColumn = false)]
+			[Column(Configuration = ProviderName.MySql, IsColumn = false)]
+			[Column(Configuration = TestProvName.MySql57, IsColumn = false)]
+			[Column(Configuration = TestProvName.MariaDB, IsColumn = false)]
+			[Column("char20DataType")]
```
ili: table formatting

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+
+			[Column("ncharDataType")]
+			[Column("nchar20DataType", Configuration = ProviderName.SapHana)]
+			[Column("CHAR20DATATYPE", Configuration = ProviderName.DB2)]
+			[Column("char20datatype", Configuration = ProviderName.PostgreSQL)]
+			[Column("char20DataType", Configuration = ProviderName.MySql)]
+			[Column("char20DataType", Configuration = TestProvName.MySql57)]
+			[Column("char20DataType", Configuration = TestProvName.MariaDB)]
+			[Column(Configuration = ProviderName.Firebird, IsColumn = false)]
+			public string NString;
```
ili: table formatting

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+			public int Id;
+
+			[Column(Configuration = ProviderName.SqlCe, IsColumn = false)]
+			[Column(Configuration = ProviderName.DB2, IsColumn = false)]
+			[Column(Configuration = ProviderName.PostgreSQL, IsColumn = false)]
+			[Column(Configuration = ProviderName.MySql, IsColumn = false)]
+			[Column(Configuration = TestProvName.MySql57, IsColumn = false)]
+			[Column(Configuration = TestProvName.MariaDB, IsColumn = false)]
+			[Column("char20DataType")]
+			public char? Char;
```
ili: table formatting

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+
+			[Column("ncharDataType")]
+			[Column("nchar20DataType", Configuration = ProviderName.SapHana)]
+			[Column("CHAR20DATATYPE", Configuration = ProviderName.DB2)]
+			[Column("char20datatype", Configuration = ProviderName.PostgreSQL)]
+			[Column("char20DataType", Configuration = ProviderName.MySql)]
+			[Column("char20DataType", Configuration = TestProvName.MySql57)]
+			[Column("char20DataType", Configuration = TestProvName.MariaDB)]
+			[Column(Configuration = ProviderName.Firebird, IsColumn = false)]
+			public char? NChar;
```
ili: table formatting

Tests/Linq/Tests.csproj
```diff
@@ -43,7 +43,7 @@
     <DefineConstants>TRACE;DEBUG;FW4;MOBILE1</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>3</WarningLevel>
-    <PlatformTarget>x86</PlatformTarget>
+    <PlatformTarget>AnyCPU</PlatformTarget>
```
ili: Does it right? Previously some data clients (DB2 if i'm not mistaken) needed x86

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+			// filter out null-character test cases for servers/providers without support
+			if (context == ProviderName.PostgreSQL
+				|| context == ProviderName.PostgreSQL + ".LinqService"
+				|| context == ProviderName.DB2
+				|| context == ProviderName.DB2 + ".LinqService"
+				|| context == ProviderName.SqlCe
+				|| context == ProviderName.SqlCe + ".LinqService"
+				|| context == ProviderName.SapHana
+				|| context == ProviderName.SapHana + ".LinqService")
+				return CharTestData.Where(_ => _.NChar != '\0').ToArray();
```
ili: table formatting

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+				|| context == ProviderName.DB2 + ".LinqService"
+				|| context == ProviderName.SQLite
+				|| context == ProviderName.SQLite + ".LinqService"
+				|| context == TestProvName.SQLiteMs
+				|| context == TestProvName.SQLiteMs + ".LinqService"
+				|| context == ProviderName.SqlCe
+				|| context == ProviderName.SqlCe + ".LinqService"
+				|| context == ProviderName.SapHana
+				|| context == ProviderName.SapHana + ".LinqService")
+				return StringTestData.Where(_ => !(_.NString ?? string.Empty).Contains("\0")).ToArray();
```
ili: table formatting

Tests/Linq/Linq/CharTypesTests.cs
```diff
@@ -0,0 +1,324 @@
...
+				|| context == ProviderName.DB2 + ".LinqService"
+				|| context == ProviderName.PostgreSQL
+				|| context == ProviderName.PostgreSQL + ".LinqService"
+				|| context == ProviderName.MySql
+				|| context == ProviderName.MySql + ".LinqService"
+				|| context == TestProvName.MySql57
+				|| context == TestProvName.MySql57 + ".LinqService"
+				|| context == TestProvName.MariaDB
+				|| context == TestProvName.MariaDB + ".LinqService";
+		}
```
ili: table formatting

Tests/Linq/Tests.csproj
MaceWindu: In my case I have clients that work only in x86 mode: Access, DB2 and clients that work only in x64 mode: SAP HANA and Informix. With this change I can switch between x86 and x64 test runners when I need to run tests against all databases.

x86 in general doesn't make sense for our test project as it doesn't have hard dependency on x86 environment - processor architecture must be set on test runner level.

MaceWindu:
@ili I've addressed your code review comments

Source/DataProvider/Access/AccessDataProvider.cs
```diff
@@ -34,7 +34,7 @@ protected AccessDataProvider(string name, MappingSchema mappingSchema)
 			SqlProviderFlags.IsInnerJoinAsCrossSupported = false;
 
 			SetCharField("DBTYPE_WCHAR", (r,i) => r.GetString(i).TrimEnd(' '));
-			SetCharFieldToType<char>("DBTYPE_WCHAR", (r, i) => AccessTools.GetChar(r, i));
+			SetCharFieldToType<char>("DBTYPE_WCHAR", (r, i) => DataTools.GetChar(r, i));
```
ili: may be  `SetCharFieldToType<char>("DBTYPE_WCHAR", DataTools.GetChar)` to avoid extra method call... (from another side because this is one line method it should be inlined by csc compiler in release build...)

MaceWindu: those methods set expression<func> expressions, so I don't think it will work :)

Issue #733, opened 2017-06-16
Intercepting DB Exceptions


RoyC-Retail-Assist:
We would like to be able to catch database exceptions and handle them as necessary.  We would also like to be database agnostic where possible however each provider returns its own Exception type meaning that we would have to supply catch statements appropriate for each type of database we are supporting.

I'm looking for a way of intercepting DB exceptions before they actually get re-thrown so that the user can supply their own exceptions generated from a single place.

I think that this would need to be in the DataConnection ExecuteNonQuery and ExecuteReader methods.  If these methods were made virtual then a user could simply inherit DataConnection and override these methods wrapping their own try/catch block around the base call and then throwing their own exception as required.

I can supply a PR if this is agreeable.

jogibear9988:
I'd like this idea! In bltoolkit we had implemented a retry for some exceptions !

jogibear9988:
Is this solved with the RetryPolicy?

RoyC-Retail-Assist:
Yes.  Retry Policy works great thanks.

PR #735, opened 2017-06-17
Exception interception issue #733

RoyC-Retail-Assist:
For issue #733 

Added virtual method that is called before a database exception is finally thrown to allow users to replace the database specific exception with a more generic one.

I can't find a mechanism to force a Exception for the ExecuteScalar method in the tests so suggestions would be welcome.

### DIFF
--- a/Source/Data/DataConnection.cs
+++ b/Source/Data/DataConnection.cs
@@ -291,14 +291,6 @@ namespace LinqToDB.Data
 		}
 
 		public static Action<string,string> WriteTraceLine = (message, displayName) => Debug.WriteLine(message, displayName);
-
-		static Func<Exception, Exception> defaultExceptionIntercept = (original) => original;
-		private Func<Exception,Exception> _exceptionIntercept = defaultExceptionIntercept ;
-		public Func<Exception, Exception> ExceptionIntercept
-		{
-			get { return _exceptionIntercept; }
-			set { _exceptionIntercept = value ?? defaultExceptionIntercept; }
-		}
 		#endregion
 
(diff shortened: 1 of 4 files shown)


MaceWindu:
Could you update your PR to use C# 5? Unfortunatelly we don't support C# 6+ in our codebase right now

jogibear9988:
Can we change the Code so that also a retry is possible? For Example if "ExceptionIntercept" returns null?

ili:
Interception is really needed or #736 is enough?

RoyC-Retail-Assist:
#733 was to allow interception of Exceptions to allow re-throwing a different, database agnostic exception.  I only added in the retry code after @jogibear9988 made the suggestion as it wasn't much effort to include.

I can rework this to fit in with the #736 code changes without too much effort if it is likely to be pulled.

jogibear9988:
Can't you also use the RetryPolicy to change the Exceptions? If you put a Try/Catch inside it, and throw a new Exception?

RoyC-Retail-Assist:
I can wrap the Execute???Intenal methods introduced in #736 with the try/catch blocks in this PR and it should then allow the retry policy to run and the allow the interception of any resultant exception.

I'll just return the original exception if the user returns a null :o)

@ili Is this approach likely to be pulled?

RoyC-Retail-Assist:
I've had a closer look at the RetryPolicy and I can, indeed, utilise it to be able to catch and then re-throw the exceptions.

I'll close this PR and then create a new one with a sample unit test that will demonstrate a solution to #733

PR #736, opened 2017-06-18
Retry policy infrastructure

andrewvk:
Base contract for retry policy strategy and db access wrapping

### DIFF
--- a/Source/Common/Configuration.cs
+++ b/Source/Common/Configuration.cs
@@ -6,6 +6,7 @@ namespace LinqToDB.Common
 	{
 		public static bool IsStructIsScalarType = true;
 		public static bool AvoidSpecificDataProviderAPI;
+		public static IRetryPolicy RetryPolicy;
 
 		public static class Linq
 		{
--- a/Source/Common/Tools.cs
+++ b/Source/Common/Tools.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.IO;
 using System.Reflection;
+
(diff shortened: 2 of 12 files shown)


jogibear9988:
Is this to https://github.com/linq2db/linq2db/pull/735

andrewvk:
No, it's just a coincidence. I prepare to move old azure project from EF (+ MS TransientFaultHandling) to linq2db, and retry policy support is critical.

sdanyliv:
How it should work with transactions?

andrewvk:
This version just ignore it, do all retries inside current TX (like EF Core or early variants for classic EF does). Later I will port a helper form EF Core to make retry of all transaction content or create something more implicit.

sdanyliv:
Great. I saw EF implementation, that's why i'm asking ;)

jogibear9988:
but if we include this, we can ignore #735?

andrewvk:
I think yes, ignore it. Returning null for retrying is a highly non-obvious logic.

ili:
@andrewvk for now retry is done only for queries, we do have 'DataConnection.InitConnection` method, it opens connection - the question is - should it be retried or not?

andrewvk:
Fix it in master

jogibear9988:
Can I use the same DbRetryPolicy (derived from RetryPolicyBase) for multiple DataConnections ? If I do so, is then maybe the retry counter wrong?

Issue #732, opened 2017-06-14
Invalid merge statement generated for InsertOrUpdate with Oracle


Lexey:
I have the following table defined:

```c#
    [Table("Currency")]
    public class Currency
    {
        /// <summary>The length of the CurrencyCode</summary>
        public const int CurrencyCodeLength = 3;


        /// <summary>Currency code</summary>
        /// <example>RUB</example>
        /// <example>USD</example>
        [Column(DataType = DataType.Char, Length = CurrencyCodeLength), PrimaryKey, LinqToDB.Mapping.NotNull]
        public string Code { get; set; }
    }
```
And I am trying to perform the following operation on it:
```c#
                dataAccess.GetTable<DbCurrency>()
                        .InsertOrUpdate(() => new DbCurrency { Code = currencyCode }, _ => new DbCurrency {});
```
So, it is just an "insert if not exists".

The following sql statement is being generated:
```sql
MERGE INTO pfe.Currency t1
USING (SELECT :currencyCode AS Code FROM SYS.DUAL) s ON
(
	t1.Code = s.Code
)
WHEN MATCHED THEN
	UPDATE 
	SET

WHEN NOT MATCHED THEN
	INSERT
	(
		Code
	)
	VALUES
	(
		:currencyCode
	)
```
And it is obviously invalid.

### Environment details
linq2db version: *1.8.2*  
Database Server: *Oracle 11g*  
Database Provider: *Oracle.Managed*  
Operating system: *Windows 7*  
Framework version: *.NET Framework 4.6.2*

ili:
@MaceWindu can you look at this?

MaceWindu:
This is duplicate of https://github.com/linq2db/linq2db/issues/100.
I will try to address it this week

Issue #100, opened 2014-07-25
Wrong SQL generated on InsertOrUpdate with empty onDuplicateKeySetter


ghost:
Some C# code what calls InsertOrUpdate:

``` csharp
                    dataContext.BudgetCategories.InsertOrUpdate(
                        () => new DataModel.BudgetCategory
                        {
                            BudgetId = budgetId,
                            CategoryId = categoryId,
                        },
                        (x) => new DataModel.BudgetCategory {
                        });
```

This code generates SQL:

``` sql
MERGE INTO [Pfm].[BudgetCategories] [t1]
USING (SELECT @budgetId AS [BudgetId], @categoryId AS [CategoryId]) [s] ON
(
        [t1].[BudgetId] = [s].[BudgetId] AND
        [t1].[CategoryId] = [s].[CategoryId]
)
WHEN MATCHED THEN
        UPDATE
        SET


WHEN NOT MATCHED THEN
        INSERT
        (
                [BudgetId],
                [CategoryId]
        )
        VALUES
        (
                @budgetId,
                @categoryId
        );
```

Section 'WHEN MATCHED THEN' should be completely omittied in this case. (Same for 'WHEN NOT MATCHED THEN').

MaceWindu:
@sdanyliv could you assign it to me? I've started to work on merge operation support improvements

sdanyliv:
Go on, i can't assign issue, because you are not meber of linq2db team. I hope we will fix that soon ;)

PR #746, opened 2017-06-22
fix for #100 and #732

MaceWindu:
Fixes call of InsertOrUpdate without updateable fields. See #100, #732 
Implements insert if not exists operation


Tests/Linq/Update/UpdateTests.cs
```diff
@@ -738,7 +738,10 @@ public void UpdateTop(string context)
 			ProviderName.SqlServer2000,
 			ProviderName.SapHana,
 			ProviderName.MySql,
-			TestProvName.MySql57
+			TestProvName.MySql57,
```
ili: it looks like easier to define supported providers :)

Issue #753, opened 2017-06-27
Association with constant value


genusP:
In table Assignments exist two columns: TargetName, TargetId. This is columns contains target table name and id.
Could I create association property, for entity Request, when TargetName field always contains value 'bda.Requests'?

ili:
You mean that on association you always need to generate query like:
```sql
select *
from Requests t
inner join Assignments a on t.Id = a.TargetId  and a.TargetName = 'bda.Requests'
```

I think no, you would need to deal with this manually, but you can use some helper:
```cs
public static IQuerable<Assignments> RequestAssignments(this IDataContext db) => db.GetTable<Assignments>().Where(_ => _.TargetName == "bda.Requests");

var data = from r in db.GetTable<Requests>()
           from a in db.RequestAssignments()
           select a;
```

genusP:
I use OData library and Association used as Navigation property.
I fork library and add extend handling associations:
```C#
class SomeClass{
  ....
  [Association(ThisKey ="TargetId, TargetName", OtherKey ="Id, const:bda.Requests", Relationship = Relationship.ManyToOne)]
  public Class1 Request { get; set; }
  ....
}
```
If necessary I can make a PR

ili:
@genusP do it! )

ili:
Another question is, should be this constraint applied if I do:
```cs
var data = from r in db.GetTable<Requests>()
           join a in db.RequestAssignments() on r.Id equals a.TargetId
           select a;
```

you see, in this case association would not be used, query would be:
```sql
select *
from Requests t
inner join Assignments a on t.Id = a.TargetId 
```

to cover both cases we do need to introduce this behavior using query optimizer like [here](https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Samples/ConcurrencyCheckTests.cs) but adding check to join condition

ili:
```cs
public class Request
{
    [Assosiation(ExpressionPredicate = "RequestAssignmentPredicate")]
    public IEnumerable<Assignment> Assignments { get; set; }

    static Expression<Func<Request, Assignment, bool>> RequestAssignmentPredicate => 
        (r, a) => r.Id == a.TargetId && a.TargetName == "bda.Requests";
}
```

igor-tkachev:
Try this:

```c#
public class Request
{
    [ExpressionMethod("AssignmentsImpl")]
    public IEnumerable<Assignment> Assignments { get; set; }

[...]

PR #757, opened 2017-06-29
Add support constant as association key

genusP:
See #753


sdanyliv:
I think algorithm should support string constants, that may contain spaces and comas.

sdanyliv:
@genusP, don't do anyting according this feature, maybe we have better idea how to implement such functionality in more generic way.

genusP:
Well I'll wait

ili:
We have discussed this issue and more generic way would be:
```cs
[Assosiation(ExpressionPredicate = "ThisOtherPredicate")]
public IEnumerable<OtherClass> Other { get; set; }

static Expression<Func<ThisClass, OtherClass, bool>> ThisOtherPredicate => 
    (t, o) => t.OtherId == o.Id && o.Target == "ThisClass";

```

So I'm closing this PR for now, this would be implemented in 1.9.0.

PR #758, opened 2017-07-02
[WIP] async implementation

igor-tkachev:


ferbenor:
Hi my friends, how i can use ToListAsync on FW4?

ili:
If i'm not mistaken you can, but also shoud reference [Microsoft.Bcl.Async](https://www.nuget.org/packages/Microsoft.Bcl.Async/)

ili:
From my own practice it is not a good idea. I'v faced with some strange bugs and application terminating using BCL in FW4 project

Issue #773, opened 2017-07-24
SQLite FTS table Match support


peterzt:
Greetings,

Can you please tell how to code support for Match operator (Full text search in SQLite)

Example query:

select f1,f2 from t3 where t3 match 'someText*'

The above performs match against all fields. 

Thanks in advance,
Peter

sdanyliv:
I do not see possibility in current implementation will handle this.

peterzt:
I have managed to do it through some sort of hack

` public static class Exts
    {
        [Sql.Expression("DataTableFTS MATCH {1}", ServerSideOnly = true, IsPredicate = true)]
        public static bool MatchFts<T>(this T x, T val) where T : IComparable<T>
        {
            throw new NotSupportedException();
        }
    }
...
                var _list = _dbData.Where(arg => arg.Col1.MatchFts("Joe*")).ToList();
...
`

This resulted in needed sql query

sdanyliv:
Yes, it is possible,  but table aliases may have change, so it's why i do not accept this solution.

peterzt:
I see but keep in mind that SQLite doesn't support alias in this case, i.e. you can't write
select * from TableFTS [t1] where [t1] match 'Joe*'
It must always be table name
select * from TableFTS [t1] where TableFTS  match 'Joe*'

SQLite FTS4.

sdanyliv:
Can you check this variant?
```sql
select * from TableFTS [t1] where [t1].TableFTS match 'Joe*'
```

peterzt:
Yes, this works fine.

sdanyliv:
According to specification Sqlite's tables has hidden field named as table, so it's why it works.
Looking for better solution in your case.

sdanyliv:
There is solution that works with current sources:
```c#
	public static class SqlLite
	{
		class MatchBuilder : Sql.IExtensionCallBuilder
		{
			public void Build(Sql.ISqExtensionBuilder builder)
			{
				var field = builder.GetExpression("src") as SqlField;
				if (field == null)
					throw new InvalidOperationException("Can not get table");

				var sqlTable = (SqlTable) field.Table;
				var newField = new SqlField
				{
					Name  = sqlTable.PhysicalName,
					Table = sqlTable
				};

				builder.AddParameter("table_field", newField);
			}
		}

		[Sql.Extension("{table_field} matches {match}", BuilderType = typeof(MatchBuilder), IsPredicate = true)]
[...]

PR #777, opened 2017-07-27
Extension updates

sdanyliv:
Added possibility to use table as an expression. 
Added possibility to examine SelectQuery during Extension building. It may be useful for searching TableSource.


MaceWindu:
related issue: #773

PR #779, opened 2017-07-30
[Documentation] DataConnection

MaceWindu:
Xml documentation for DataConnection class public surface

### DIFF
--- a/Source/Data/DataConnection.cs
+++ b/Source/Data/DataConnection.cs
@@ -61,7 +61,7 @@ namespace LinqToDB.Data
 		/// Creates database connection object that uses provided connection configuration.
 		/// </summary>
 		/// <param name="configurationString">Name of database connection configuration to use with this connection.
-		/// In case of null, configuration from <see cref="DefaultConfiguration"/> property will be used.</param>
+		/// In case of <c>null</c>, configuration from <see cref="DefaultConfiguration"/> property will be used.</param>
 		public DataConnection(string configurationString)
 		{
 			InitConfig();
@@ -262,7 +262,7 @@ namespace LinqToDB.Data
 
 		private int? _id;
 		/// <summary>
-		/// Unique identifier of mapping schema and connection combination. Shared by connections with the same mapping schema and connection configuration or connection string.
+		/// For internal use only.
 		/// </summary>
 		public  int   ID
 		{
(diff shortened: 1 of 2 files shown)


MaceWindu:
[ci skip]

Source/Data/DataConnection.cs
```diff
@@ -397,6 +515,12 @@ static DataConnection()
 		static readonly List<Func<IConnectionStringSettings,string,IDataProvider>> _providerDetectors =
 			new List<Func<IConnectionStringSettings,string,IDataProvider>>();
 
+		/// <summary>
+		/// Registers database provider factory method.
+		/// Factory accepts connection string settings and connection string. Could return null, of cannot create provider
```
sdanyliv: Typo here 'if' not 'of'

Source/Data/DataConnection.cs
```diff
@@ -170,17 +238,32 @@ public DataConnection(string configurationString)
...
 #endif
 
 		static readonly ConcurrentDictionary<string,int> _configurationIDs;
 		static int _maxID;
 
 		private int? _id;
+		/// <summary>
+		/// Unique identifier of mapping schema and connection combination. Shared by connections with the same mapping schema and connection configuration or connection string.
+		/// </summary>
 		public  int   ID
```
igor-tkachev: This property is for internal use only. We can hide it and make it internal.

Source/Data/DataConnection.cs
```diff
@@ -683,19 +856,29 @@ internal void InitCommand(CommandType commandType, string sql, DataParameter[] p
...
 		{
 			get { return _command ?? (_command = CreateCommand()); }
 			set { _command = value; }
 		}
 
+		/// <summary>
+		/// Creates new command object, associated with current connection/transaction.
+		/// </summary>
+		/// <returns>Command object.</returns>
 		public IDbCommand CreateCommand()
```
igor-tkachev: Not sure this method should be public.

Source/Data/DataConnection.cs
```diff
@@ -709,6 +892,9 @@ public IDbCommand CreateCommand()
 			return command;
 		}
 
+		/// <summary>
+		/// Disposes current command object, associated with connection.
+		/// </summary>
 		public void DisposeCommand()
```
igor-tkachev: And this one.

Source/Data/DataConnection.cs
jogibear9988: if we do so, we should increase major version, it's maybe a breaking change

MaceWindu:
- fixed typo
- marked internal API methods with "Internal api" comment for now
- examples will be provided with next PR

PR #780, opened 2017-08-01
[WIP] Added DatePart and DateAdd replacement samples.

sdanyliv:
It is sample replacement for current Sql.DatePart and Sql.DateAdd functions which implementation is sparse across SqlOptimizer's code for specific provider.

**Planned**
- [x] Replace current implementation with new functions
- [x] Remove obsolete code from SqlOptimizers
- [x] Add DateDiff functions


MaceWindu:
still on radar?

sdanyliv:
Will finsh it when found time. Not too much work left here. Just needs to ensure that this works for every provider.

MaceWindu:
postgresql doesn't like new implementation :smile:

sdanyliv:
Will check.

Tests/Linq/Linq/DateTimeFunctionsTests.cs
```diff
@@ -178,6 +178,29 @@ public void DatePartSecond([DataSources] string context)
 					from t in db.Types select Sql.AsSql(Sql.DatePart(Sql.DateParts.Millisecond, t.DateTimeValue)));
 		}
 
+		[Test]
+		public void DatepartDynamicynamic(
```
MaceWindu: test name

MaceWindu:
We decided just document `week` datepart behavior for each provider for now

Issue #264, opened 2015-10-04
Group by error


Svancara:
Hello! Thanks for your hard work! 
I think I have found an error.
I am using MySql.

I have this LINQ:
```cs
var TABLE3 = from syst in C2SYSTEMS group syst by new { Y = syst.Added.Year };
```

where "Added" is DateTIme:
```cs
    public class c2ex_publicsystems
    {
        public c2ex_publicsystems();

        [Column]
        [NotNull]
        public DateTime Added { get; set; }

...snip....
    }
```

linq2db generates a wrong SQL command:
```sql
SELECT
    `t1`.`Added`
FROM
    `c2ex_publicsystems` `t1`
GROUP BY
    Extract(Year from `t1`.`Added`),
    `t1`.`Added`
```
Sincerely,
Bob

Svancara:
Is this a solution? It works for me.

BasicSqlBuilder.cs:

protected virtual void BuildGroupByClause()
		{
			if (SelectQuery.GroupBy.Items.Count == 0)
				return;

var items = SelectQuery.GroupBy.Items.Where(i => !(i is SqlValue || i is SqlParameter)).ToList();

**// Bob: If there are SqlExpressions and the last one is SqlField, then remove SqlField:
            if (items.Count > 1 && items[items.Count - 2] is LinqToDB.SqlQuery.SqlExpression && items[items.Count - 1] is LinqToDB.SqlQuery.SqlField)
            {
                items.RemoveAt(items.Count - 1);
            }**

sdanyliv:
@Svancara, I've  prepared correct fix for that issue #660

MaceWindu:
Reopen, see #782

Issue #781, opened 2017-08-01
GroupBy() followed by Count() tries to perform count on client


MaceWindu:
At least following linq patterns generate `select *` query (which fails, because usually not all table fields included into group by) instead of `select count(*):
```c#
query.GroupBy(...).Count();
query.GroupBy(...).LongCount();
query.GroupBy(...).Having(...).Count();
query.GroupBy(...).Having(...).LongCount();
```

### Environment details
linq2db version: *1.8.3*  
Database Server: *SQL Server 2016*  
Operating system: *Windows 10*  
Framework version: *.NET Framework 4.6.1*

MaceWindu:
BTW, for workaround, add `Select(...)` call before `Count`

MaceWindu:
Looks like this happens only if you group not by plain field, but by some expression. Will add tests soon

ili:
fixed via #775

Preservejohan3:
> BTW, for workaround, add `Select(...)` call before `Count`





https://eddyencouragement.github.io/return-value36/

PR #782, opened 2017-08-01
Fix for #781

MaceWindu:
Tests for #781 and andother case of #264


MaceWindu:
Note what query generated for workaround for #264 :)

sdanyliv:
Fixed issue #782 
Still not fixed #264 - need help from @igor-tkachev

ili:
I'v extracted #264 into #790

Issue #784, opened 2017-08-04
SQLite schema generation error when Foreign Key targets table only


Corey-M:
I have an SQLite3 database file from a vendor application that is causing schema generation to fail with an exception when generating the schema:

```
System.InvalidOperationException occurred
  HResult=0x80131509
  Message=Sequence contains no elements
  Source=System.Core
  StackTrace:
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source)
   at LinqToDB.SchemaProvider.SchemaProviderBase.GetSchema(DataConnection dataConnection, GetSchemaOptions options) in M:\Dev\GitHub\linq2db\Source\SchemaProvider\SchemaProviderBase.cs:line 156

```

I downloaded the linq2db code to track down the cause and eventually tracked down the source of the issue as the SQL used to generate the table.  A minimal example would be:

```sql
CREATE TABLE `Parent` (
  `KeyField` INTEGER NOT NULL,
  `NonKeyField` TEXT NOT NULL,
  PRIMARY KEY(`KeyField`)
);

CREATE TABLE `Child`
(
  `ParentKey` INTEGER NOT NULL,
  `ChildKey` INTEGER NOT NULL,
  `SomeData` TEXT NOT NULL,
  PRIMARY KEY (`ParentKey`, `ChildKey`),
  FOREIGN KEY (`ParentKey`) REFERENCES Parent ON DELETE CASCADE
);
```

SQLite supports this and correctly cascades deletions.

When generating a schema for the database the foreign key list has an empty string in `FK_TO_COLUMN` which then causes the exception in the lookup of `otherColumn` on line 157.

### Steps to reproduce
Create tables using the above create statements.

The problem can be seen immediately in LINQPad using the LinqToDB driver.  Refresh the database after creating the tables and you will receive a schema error.

For C# create a project, add Nuget packages: System.Data.SQLite.Core, linq2db

Put the following code into the program.cs and run.

```cs
using System;
using LinqToDB.Data;
using LinqToDB.DataProvider.SQLite;

namespace SQLiteErrorTest
{
    public class TestDB : DataConnection
    {
        public TestDB()
            : base(new SQLiteDataProvider(), @"DataSource=M:\Dev\TestData.db")
        { }
    }

    public class Program
    {
        public static void Main()
        {
            var provider = new SQLiteDataProvider();
            var schemaProvider = provider.GetSchemaProvider();
            var schema = schemaProvider.GetSchema(new TestDB());
        }
    }
[...]

PR #785, opened 2017-08-04
Fix for #784

Corey-M:
Fixes issue #784 

For foreign keys with no target columns, selects the matching Primary Key column by table and ordinal.

Tested against example given in issue, also against original problem database that caused the issue, works as expected.


ili:
@Corey-M thanks for your contribution!
It would be great to add tables demonstrating the issue to `SQLite.sql` script to test the fix

Corey-M:
@ili The issue I submitted has the full details including tables that exhibit the problem.  I'll see if I can add that into the `SQLite.sql` script.

Corey-M:
@ili Added some items to the bottom of the `SQLite.sql` file to create tables that will test the issue.  First commit was bad, reverted and actually tested the create scripts this time.

PR #787, opened 2017-08-06
XML Documentation Part 2

MaceWindu:
Next batch of XML documentation comments.

Notes:
 - I've decided to postpone work on examples and concentrate on comments first and return to them on second pass. Otherwise it will slow down documentation process a lot.
 - there are two uncommented maxBatchSize parameters in DataConnectionExtensions - I will get back to them later when I will write comments for BulkCopy functionality as it requires some research on what those parameters mean for different providers and BulkCopy modes.
- OptimizeJoins configuration option lacks details on what type of optimizations performed. I ask @sdanyliv  to fill it in.
 - I've made small change to RetryPolicy.UseDefaultPolicy https://github.com/linq2db/linq2db/pull/787/files#diff-e9463d73885c65dc8372c43028deff49R183 - please check if it is correct change


Source/Common/Configuration.cs
```diff
@@ -43,70 +111,108 @@ public static class Linq
...
 
 			/// <summary>
 			/// The default number of retry attempts.
+			/// Default value: <c>5</c>.
 			/// </summary>
 			public static int DefaultMaxRetryCount = 5;
 
 			/// <summary>
 			/// The default maximum time delay between retries, must be nonnegative.
+			/// /// Default value: 30 seconds.
```
sdanyliv: Duplicated comment here

MaceWindu: fixed

PR #793, opened 2017-08-10
Set SkipOnInsert and SkipOnUpdate for Computed columns for SqlServer

avtc:
That will also help `linq2db.t4models` to generate column attributes for computed columns:
`[Column(SkipOnInsert = true, SkipOnUpdate = true)]`


ili:
We'll look forward to implement this feature. 
For now you can [do it manually in your Model.tt](https://github.com/linq2db/t4models#customizing-generation-process)

avtc:
I know, that i can do it manually.
That pull request will make it automatically for SqlServer provider.

PR #812, opened 2017-08-19
UseBinaryAggregateExpression rework

MaceWindu:


MaceWindu:
@detoxhby I hope it will solve your issue

MaceWindu:
Found another issue with this functionality - query cache comparison uses initial expression and still fails

detoxhby:
Thanks, tested with 7e05a48 :white_check_mark:

detoxhby:
Have to correct myself, I forgot that I've removed the `LinqToDB.Common.Configuration.Linq.UseBinaryAggregateExpression = true` for previous tests and ran without it. Setting it true again fails the same way. :no_entry: (tested with 683cedf)

MaceWindu:
I have good news.
I've enabled this option for all tests and now it reproduces pretty stable on two tests (TypesTests.Bool1 and  Issue271Tests.Test2). This means I can debug and fix it.

MaceWindu:
@detoxhby could you test it once more?

detoxhby:
@MaceWindu just did it, something broke. Random errors from not inserting data to even missing FKs :/

MaceWindu:
@detoxhby now it should work

detoxhby:
@MaceWindu seems good! ✅ 

Even though the commit diff looks terrifying ... was it really "just" that? How was it working for other providers (like SqlServer) and struggle only on SQLCe if there is no provider-specific changes here?
Don't get me wrong, I'm just curious and trying to understand better.

MaceWindu:
Well, it wasn't SqlCe specific issue. For queries with parameters we generate methods to extract parameter value from expression. They were generated to work with initial expression, but when this optimization applied - expression structure changes and parameter extractor receives expression with different structure and that is why you get all those invalid cast expressions: when method expected element of one type - he received completely different one.

Why it was broken only for SqlCe in your case - I cannot say for sure. It suspect one of those:
- expression wasn't optimized
- optimized expression still had structure, compatible with parameter extraction method

detoxhby:
@MaceWindu thanks for the explanation and fix, keep up the good work!

Source/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -334,7 +331,30 @@ Expression AggregateExpression(Expression expression)
 
 							if (items.Count > 2)
```
sdanyliv: I thin we can use here Count > 3

PR #686, opened 2017-05-27
New Merge API

MaceWindu:
DB2 iSeries/DB2 zOS not tested due to their public unavailability. Should work. Or not.

Fixes:
- #200 
- #234 
- #334 
- #335 
- #336 
- #337 
- #438 (partially, only for providers with native merge support)
- #145 
- #668 

Other fixes:
- DateTimeOffset with negative offset saved properly for Oracle's TIMESTAMP WITH TIME ZONE
- Various issues with literals generation

Code tasks:
- [x] Providers: SQL Server
- [x] Providers: DB2 LUW
- [x] Providers: Informix
- [x] Providers: SAP HANA (v2)
- [x] Providers: Firebird 2.1-2.5
- [x] Providers: Firebird 3
- [x] Providers: Oracle
- [x] Providers: Sybase
- [x] Custom update for providers with "alternative update" (DB2, Oracle, Firebird, Informix, SAP HANA)
- [x] Generate aliases for source columns
- [x] Add tests for different data types (in progress)
- [x] Restore old API (remove Obsolete attribute)
- [x] Refactor API naming/structure
- [ ] testing using DB2 zOS
- [ ] testing using DB2 iSeries using Linq2DB4iSeries provider (needs collaboration with @RoyC-Retail-Assist)
- [x] add tests for new API methods
- [x] implement match using keys

Documentation tasks:
- [x] XML Documentation on new API methods
- [x] [Merge documentation root page](https://github.com/MaceWindu/linq2db/wiki/Merge-API)
- [x] [old API migration guide](https://github.com/MaceWindu/linq2db/wiki/Merge-API-:-Migration-from-Previous-API)
- [x] [Document support level by provider wiki](https://github.com/MaceWindu/linq2db/wiki/Merge-API-:-Background-Information-and-Providers-Support)
- [x] [API description wiki](https://github.com/MaceWindu/linq2db/wiki/Merge-API-:-Description)
- [ ] Move wiki pages to main repository from fork

PR #814, opened 2017-08-20
Add CanCombineParameters support in merge-based InsertOrUpdate

MaceWindu:
Post-mortem for review.
Looks like CanCombineParameters flag was never supported for merge-based InsertOrUpdate query and we didn't spotted it before due to:
- only Sybase use this flag in default distribution
- Sybase doesn't use merge-based query for InsertOrUpdate
- for iSeries it wasn't spotted before due to a bug with duplicate parameters, which I fixed here https://github.com/linq2db/linq2db/pull/686#pullrequestreview-40652778

### DIFF
--- a/Source/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/SqlProvider/BasicSqlBuilder.cs
@@ -514,9 +514,37 @@ namespace LinqToDB.SqlProvider
 
 			AppendIndent().Append("USING (SELECT ");
 
+			var insertIdx = 0;
 			for (var i = 0; i < keys.Count; i++)
 			{
-				BuildExpression(keys[i].Expression, false, false);
+				var keyExpression = keys[i].Expression;
+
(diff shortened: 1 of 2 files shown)


MaceWindu:
@RoyC-Retail-Assist could you test it?

Source/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -514,9 +514,37 @@ protected virtual void BuildInsertOrUpdateQueryAsMerge(string fromDummyTable)
...
+				{
+					keyExpression = new QueryVisitor().Convert(keyExpression, e =>
+					{
+						switch (e.ElementType)
+						{
+							case QueryElementType.SqlParameter:
+								{
+									var p = (SqlParameter)e;
+
+									if (p.IsQueryParameter)
```
sdanyliv: Not sure, but i think dictionary can be created outside lambda.

PR #815, opened 2017-08-20
Fix for CanCombineParameters flag not being honoured

RoyC-Retail-Assist:
Re #814 - change that fixes issue for iSeries DB2 provider.  It will also apply to the Sybase provider but I do not have a Sybase setup to test against.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ linq2db.sln.ide/
 *.lock.json
 /api
 /linq2db.github.io
+/.axoCover
--- a/Source/SqlQuery/SelectQuery.cs
+++ b/Source/SqlQuery/SelectQuery.cs
@@ -22,11 +22,11 @@ namespace LinqToDB.SqlQuery
 		{
 			SourceID = Interlocked.Increment(ref SourceIDCounter);
 
-			_select = new SelectClause(this);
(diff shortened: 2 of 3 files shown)


RoyC-Retail-Assist:
Done

MaceWindu:
@RoyC-Retail-Assist could you provide more information on this PR? Do we need it or #814 solved your issue already? Because right now we don't plan to accept it as it doesn't look like a right place for a fix. If #814 didn't solved your issue - we need to take a look at this one once more.

RoyC-Retail-Assist:
#814 still didn't add the parameter if it was already in the parameters collection.  iSeries DB2 uses positional parameters (even though they look like they are named parameter) so if there are 4 @ signs in the query it requires 4 parameters.

v 1.8.3 provided the required parameters #814 doesn't fix the issue and this PR is a modification to #814 to add in the parameter if the CanCombineParameters flag is false.  It would appear that this would only affect the Sybase provider as that is tho only other provider that has the flag set to false.

If there is a better place for this that's great but this PR addresses the iSeries DB2 issues and is simply a small modification on #814.

RoyC-Retail-Assist:
Issue resolved by #831

Issue #820, opened 2017-08-23
Access to Nullable<T>.Value could fail query parameters extraction


MaceWindu:
Code like that:
```c#
void Test(int? param)
{
db.Table.Where(r => param != null && r.property == param.Value).ToList();
}
```
could fail if nullable parameter is null on parameter value extraction from expression with following exception:
```
System.InvalidOperationException: Nullable object must have a value.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at lambda_method(Closure , Expression , Object[] )
   at LinqToDB.Linq.QueryRunner.SetParameters(Query query, IDataContext dataContext, Expression expression, Object[] parameters, Int32 queryNumber)
   at LinqToDB.Linq.QueryRunnerBase.SetCommand(Boolean clearQueryHints)
   at LinqToDB.Data.DataConnection.QueryRunner.ExecuteReader()
```

Will add tests a bit later

### Environment details
linq2db version: *1.9.0-prerelease*

PR #821, opened 2017-08-23
Tests for #820

MaceWindu:
Tests for #820


ili:
@MaceWindu please review the fix

PR #831, opened 2017-08-27
CanCombineParameters fix

MaceWindu:
This time it should work

### DIFF
--- a/Source/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/SqlProvider/BasicSqlBuilder.cs
@@ -514,37 +514,26 @@ namespace LinqToDB.SqlProvider
 
 			AppendIndent().Append("USING (SELECT ");
 
-			var insertIdx = 0;
+			if (!SqlProviderFlags.CanCombineParameters)
+			{
+				SelectQuery.Parameters.Clear();
+
+				for (var i = 0; i < keys.Count; i++)
+					ExtractParameters(keys[i].Expression);
+
+				foreach (var expr in SelectQuery.Update.Items)
+					ExtractParameters(expr.Expression);
+
+				foreach (var expr in SelectQuery.Insert.Items)
+					ExtractParameters(expr.Expression);
+
+				if (SelectQuery.Parameters.Count > 0)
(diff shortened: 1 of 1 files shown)


MaceWindu:
@RoyC-Retail-Assist could you test this fix?

RoyC-Retail-Assist:
@MaceWindu thanks for your help with this. 

I can test this but the iSeries provider already overrides this method so I will have to manually merge the changes to test.  I'm not in work today but will try to find time to test it as I don't want to hold things up.  

Out of curiosity,  why is this a better fix than PR #815?  I prefer to offer solutions rather than just problems  so any help in understanding the code better is appreciated.

MaceWindu:
#815 'fix' parameters in place, where it is not yet known that query will be generated as merge statement so there is no information on order of expressions and parameters in generated query. E.g. for merge we will duplicate update key expressions and it means we need to duplicate parameters too for iSeries.

#815 fix works by coincidence and I will not be surprised if it will break on composite keys or on later changes in some other code.

RoyC-Retail-Assist:
Thanks for your response.  It makes sense when it is explained :o)

This fix will resolve the issue. 

Can the change be put into a protected method that I can use from my override of BuildInsertOrUpdateQueryAsMerge. Alternatively can you make ExtractParameters protected. 

		protected void ExtractParametersIfCannotCombine()
		{
			if (!SqlProviderFlags.CanCombineParameters)
			{
				var keys = SelectQuery.Update.Keys;

				SelectQuery.Parameters.Clear();

				for (var i = 0; i < keys.Count; i++)
					ExtractParameters(keys[i].Expression);

				foreach (var expr in SelectQuery.Update.Items)
					ExtractParameters(expr.Expression);

				foreach (var expr in SelectQuery.Insert.Items)
					ExtractParameters(expr.Expression);

				if (SelectQuery.Parameters.Count > 0)
					SelectQuery.IsParameterDependent = true;
			}
		}

MaceWindu:
Great! I've extracted it to separate method

RoyC-Retail-Assist:
Thanks.  I'll make my changes to use this once it has been merged.

PR #833, opened 2017-09-01
fixed extension associations

ili:
Fixed
```cs
public void ExtensionTest3(string context) 
{ 
	{ 
		AreEqual( 
		   Child.Select(_ => new { p = _.Parent   }).Select(_ => _.p.ParentID), 
		db.Child.Select(_ => new { p = _.Parent() }).Select(_ => _.p.ParentID)); 
	} 
} 
``

### DIFF
--- a/Source/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -1061,7 +1061,7 @@ namespace LinqToDB.Linq.Builder
 							if (CountBuilder.MethodNames.Contains(e.Method.Name) || e.IsAggregate(MappingSchema))
 								result = IsQueryMember(e.Arguments[0]);
 						}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Source/Linq/Builder/ExpressionBuilder.SqlBuilder.cs [hunk 1]
```diff
@@ -1061,7 +1061,7 @@ bool IsServerSideOnly(Expression expr)
 							if (CountBuilder.MethodNames.Contains(e.Method.Name) || e.IsAggregate(MappingSchema))
 								result = IsQueryMember(e.Arguments[0]);
 						}
-						else if (e.IsAggregate(MappingSchema))
+						else if (e.IsAggregate(MappingSchema) || e.IsAssociation(MappingSchema))
```
sdanyliv: I think this check should be placed on top of `case ExpressionType.Call`, also it may be needed for properties.

PR #852, opened 2017-09-12
Support for custom expression postprocessing

n9:
This small change adds support for a custom expression postprocessor. (Postprocessor is not configured statically but by particular DataContext.)

### DIFF
--- a/Source/Linq/Query.cs
+++ b/Source/Linq/Query.cs
@@ -21,6 +21,11 @@ namespace LinqToDB.Linq
 	using SqlQuery;
 	using SqlProvider;
 
+	public interface IExpressionPostprocessor
+	{
+		Expression PostprocessExpression(Expression expression);
+	}
+
 	abstract class Query
 	{
 		public Func<IDataContextEx,Expression,object[],object> GetElement;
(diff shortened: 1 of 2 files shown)


n9:
It looks like that C# 7 is not yet supported.

sdanyliv:
@n9, for which purpose this was introduced?

n9:
Expression postprocessing is useful in case you need tweak the expression before is translated into the native query language (e.g. SQL): for instance, processing application specific annotations.

sdanyliv:
I more looks like preprocessing before parsing. BTW, please correct syntax, we currently supports only C# 6.
For me it's ok, we will check.

igor-tkachev:
```c#
public interface IExpressionPostprocessor
{
	Expression PostprocessExpression(Expression expression);
}
```

???

n9:
@igor-tkachev I have updated the API according to your suggestion.
@sdanyliv I have changed the code to C# 6. And set the lang version in csproj to 6.

n9:
Is this PR ok? Or should I change it somehow?

sdanyliv:
We will merge this PR soon.

PR #866, opened 2017-10-02
BulkInsert Return sequence using MySql (with restrictions: works only with 1 int…

zxsanny:
… field as primary key in table)

I made several tests, please review

### DIFF
--- a/Source/DataProvider/MySql/MySqlDataProvider.cs
+++ b/Source/DataProvider/MySql/MySqlDataProvider.cs
@@ -104,30 +104,28 @@ namespace LinqToDB.DataProvider.MySql
 		public override BulkCopyRowsCopied BulkCopy<T>(
 			[JetBrains.Annotations.NotNull] DataConnection dataConnection, BulkCopyOptions options, IEnumerable<T> source)
 		{
-            if (source == null)
-                throw new ArgumentException("Source is null!", "source");
-            if ( options.RetrieveSequence)
-            {
-                var supportedFileTypes = new[] { typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(short), typeof(ushort) };
+			if (source == null)
+				throw new ArgumentException("Source is null!", "source");
+			if ( options.RetrieveSequence)
+			{
+				var supportedFileTypes	= new[] { typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(short), typeof(ushort) };
+				var entityDescriptor	= dataConnection.MappingSchema.GetEntityDescriptor(typeof(T));
+				var columns				= entityDescriptor.Columns.Where(x => x.IsIdentity).ToList();
 
-                var entityDescriptor = dataConnection.MappingSchema.GetEntityDescriptor(typeof(T));
-                var columns          = entityDescriptor.Columns.Where(x => x.IsIdentity).ToList();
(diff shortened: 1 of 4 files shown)


Source/DataProvider/MySql/MySqlDataProvider.cs
```diff
@@ -103,7 +104,26 @@ protected override void SetParameterType(IDbDataParameter parameter, DataType da
...
 		{
-			return new MySqlBulkCopy().BulkCopy(
+            if (options.RetrieveSequence)
+            {
+                var supportedFileTypes = new[] { typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(short), typeof(ushort) };
+
+                var entityDescriptor = dataConnection.MappingSchema.GetEntityDescriptor(typeof(T));
+                var columns          = entityDescriptor.Columns.Where(x => x.IsPrimaryKey).ToList();
+
+                if (columns.Count == 1 && supportedFileTypes.Contains(columns.First().MemberType))
```
sdanyliv: I think we have to throw exception otherwise.

Source/DataProvider/MySql/MySqlDataProvider.cs
```diff
@@ -103,7 +104,26 @@ protected override void SetParameterType(IDbDataParameter parameter, DataType da
...
+                {
+                    var c = columns.First();
+                    var lastId = dataConnection.Query<int>($"SELECT {c.ColumnName} FROM {entityDescriptor.TableName} ORDER BY {c.ColumnName} DESC LIMIT 1").FirstOrDefault();
+                    lastId = lastId + source.Count(); //Insert the max id first in order to prevent inserting collisions on inserting from another thread (on big batches)
```
sdanyliv: Source enumerable have to be converted to list. Otherwise - multiple enumerations.

Source/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -513,8 +513,8 @@ class MultipleQueryHelper<TRet> : IMultipleQueryHelper
 			}
 		}
 
-		static readonly MethodInfo _whereMethodInfo =
-			MemberHelper.MethodOf(() => LinqExtensions.Where<int,int,object>(null,null)).GetGenericMethodDefinition();
+        static readonly MethodInfo _whereMethodInfo =
+			MemberHelper.MethodOf(() => WhereExtension.Where2<int,int,object>(null,null)).GetGenericMethodDefinition();
```
sdanyliv: I can't handle why these changes introduced.

zxsanny:
Please move then this approach: Source/DataProvider/Oracle/OracleDataProvider.cs BulkCopy<T>(...)
also to LinqToDB.Tools. I was inspired mostly by this when I wrote same for MySQL (because we need identities after bulk insert, and we're not using GUIDs)
Thanks!

Source/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
zxsanny: This method was conflicted with 'Where(x => x.IsPrimaryKey)'

sdanyliv:
> But still we need .Where(x => x.IsIdentity). Or maybe here exists another approach?
I've tried to remove your changes with Where - all works fine.

PR #635, opened 2017-04-19
Implement InsertBatchWithIdentity feature using BulkCopy

lvaleriu:
Added BulkCopyOptions.RetrieveSequence property.
When RetrieveSequence is true, before doing bulkCopy, we will reserve N oracle sequence values and set them to the N objects in list.
After bulkcopy, the id's of these objects will still be present.

Issue #487, opened 2016-11-30
SqlParameter with the good DataType


CITnDev:
Hello,

I had an issue with the CommandParameter data type. If I set the DataType property of the column attribute, the CommandParameter is correctly set only if there is no jointure.
To fix the case with the Jointure, I change the code below in BasicSqlOptimizer.cs file, ConvertPredicate method, line 786.

I replaced this code :

```csharp
if (expr.Expr1 is SqlField && expr.Expr2 is SqlParameter)
{
    if (((SqlParameter)expr.Expr2).DataType == DataType.Undefined)
        ((SqlParameter)expr.Expr2).DataType = ((SqlField)expr.Expr1).DataType
}
else if (expr.Expr2 is SqlField && expr.Expr1 is SqlParameter)
{
    if (((SqlParameter)expr.Expr1).DataType == DataType.Undefined)
        ((SqlParameter)expr.Expr1).DataType = ((SqlField)expr.Expr2).DataType;
}
```
by

```csharp
if (expr.Expr2 is SqlParameter)
{
	var innerExpr = expr.Expr1;
	while (innerExpr != null && innerExpr is SelectQuery.Column)
		innerExpr = ((SelectQuery.Column) innerExpr).Expression;

	if (innerExpr != null && innerExpr is SqlField)
		((SqlParameter) expr.Expr2).DataType = ((SqlField) innerExpr).DataType;
}
else if (expr.Expr1 is SqlParameter)
{
	var innerExpr = expr.Expr2;

	while (innerExpr != null && innerExpr is SelectQuery.Column)
		innerExpr = ((SelectQuery.Column) innerExpr).Expression;

	if (innerExpr != null && innerExpr is SqlField)
		((SqlParameter) expr.Expr1).DataType = ((SqlField) innerExpr).DataType;
}
```

The code was tested on Oracle (with OracleManaged provider).

ili:
Can you please provide test, demonstrating the problem?

CITnDev:
Ok I prepare a pull request with a sample

CITnDev:
To explain quickly why there is a loop, each time I add a jointure in the query, the SqlField property is deeper.

PR #878, opened 2017-10-13
Additional Oracle Bulkcopy Insert Method to improve performance

laurentKTR:
Change type of UseAlternativeBulkCopy from bool to enum => Update Oracle Unit test

### DIFF
--- a/Source/DataProvider/Oracle/OracleBulkCopy.cs
+++ b/Source/DataProvider/Oracle/OracleBulkCopy.cs
@@ -121,9 +121,15 @@ namespace LinqToDB.DataProvider.Oracle
 		protected override BulkCopyRowsCopied MultipleRowsCopy<T>(
 			DataConnection dataConnection, BulkCopyOptions options, IEnumerable<T> source)
 		{
-			return OracleTools.UseAlternativeBulkCopy
-				? MultipleRowsCopy2(dataConnection, options, source)
-				: MultipleRowsCopy1(dataConnection, options, source);
+			switch (OracleTools.UseAlternativeBulkCopy)
+			{
+				case AlternativeBulkCopy.InsertInto:
+					return MultipleRowsCopy2(dataConnection, options, source);
+				case AlternativeBulkCopy.InsertDual:
+					return MultipleRowsCopy3(dataConnection, options, source);
+				default:
+					return MultipleRowsCopy1(dataConnection, options, source);
+			}
 		}
 
 		BulkCopyRowsCopied MultipleRowsCopy1<T>(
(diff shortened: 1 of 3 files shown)


sdanyliv:
@laurentKTR, before reviewing. Have you checked performance of Merge?
https://github.com/linq2db/linq2db/wiki/Merge-API

laurentKTR:
We are actually using merge syntax mainly to update data on a table that contains 300 millions lines per year but we didn"t give a try for insert only operations => i will have a look 
We are adding 1 million lines per day in this table using several methods : sqlLoader and bulkcopy

We have migrated from BLToolkit to Linq2db and noticed that the performance has decreased greatly 
the difference was the insert syntax used, it's the reason why i have added this "bltoolkit" syntax

laurentKTR:
ok so i have performed some tests:
Using the method "Addtional Bulkcopy Insert Method " : 50 s to insert 10 000 lines
Using the merge, i have cancelled the operations after 6 mins for 6000 lines

This is the syntax i have used for merge:
dataContext.GetTable<T>().Merge().Using(elems).On((p,j) => false).InsertWhenNotMatched().Merge()
which generate a sql query : ON ( 1 = 0 ) => fast as it doesn't compare any primary key

sdanyliv:
Great, we will review this PR soon. Possibly it can be deivered with the next release.

laurentKTR:
I have noticed that you probably have a "column type" problem in merge command

I have declared my date property with DataType = DataType.Date
and the generated sql use TO_TIMESTAMP instead of TO_DATE
=> in oracle it's a killing performance convertion

it was a bug already fixed in a previous issue #487  for basic operations ( insert/update/select) : 
Looking in bulkcopy code you are using it dataConnection.MappingSchema.GetDataType(column.MemberType).DataType
but i didn't find trace of this method in BasicMergeBuilder

=> it might be why the performance is so bad

laurentKTR:
hi

I had a look at Merge and it looks like for oracle you are not using parameters but plain value
so the problem seats in OracleMappingSchema.ConvertDateTimeToSql
I'm not sure how to proceed : open another pull request or continue in this one even if the subject is not anymore bulkcopy but merge?

sdanyliv:
I think new PR would be great.

laurentKTR:
ok

laurentKTR:
I have kept the same behaviour by assigning the default value to the existing default method 
because i was not sure if you prefer to : 
keep the same functionality and user must toggle something to have benefit of "new feature" 
or
assign the "new feature" by default

sdanyliv:
Looks like you are trying to emulate OracleBulkCopy from Oracle.DataAccess.dll.
So you have to vote ;)
https://apex.oracle.com/pls/apex/f?p=18357:39:5439210162954::NO::P39_ID:27881

laurentKTR:
voted :-) 
the native bulkcopy is already in linq2db but i don't use it as it doesn't support transaction 
https://github.com/linq2db/linq2db/pull/635

PR #879, opened 2017-10-16
OracleMappingSchema  ConvertDateTimeToSql

laurentKTR:
http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1841

image

### DIFF
--- a/Source/DataProvider/Oracle/OracleMappingSchema.cs
+++ b/Source/DataProvider/Oracle/OracleMappingSchema.cs
@@ -96,9 +96,9 @@ namespace LinqToDB.DataProvider.Oracle
 		static void ConvertDateTimeToSql(StringBuilder stringBuilder, SqlDataType dataType, DateTime value)
 		{
 			var format =
-				dataType.DataType == DataType.DateTime ?
-					"TO_DATE('{0:yyyy-MM-dd HH:mm:ss}', 'YYYY-MM-DD HH24:MI:SS')" :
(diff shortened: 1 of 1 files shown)


sdanyliv:
I can not see improvements here. Same behavior.
Looks like this one https://github.com/linq2db/linq2db/pull/879/files#diff-160cc6cc4d8dc621a46fd656966987aeR105
may override your change

laurentKTR:
https://github.com/linq2db/linq2db/pull/879/files#diff-160cc6cc4d8dc621a46fd656966987aeR105
is my change !

Do we agree on the array posted in 1st message ?

In the current code only DateTime is "cast" as TO_DATE, that means DataType.Date, DataType.Time, ..
are casted to TO_TIMESTAMP which is weird

Actually there are inconsistencies if you declare your column c#Datetime as DataType.Datetime 
db.InlineParameters = true/false
=> parameter is type of TimeSpan
=> by value is type of To_date

To fix this we have force all datetime columns to be of type DataType.Date
but when i have launched Merge, i was surprised to see again a TO_TIMESTAMP

sdanyliv:
I see.

laurentKTR:
Can you have a look at protected override void BuildDataType(SqlDataType type, bool createDbType)
in OracleSqlBuilder

switch (type.DataType)
{
	case DataType.DateTime       : StringBuilder.Append("timestamp");      

there are a lot of switch and it seems that upon them DateTime is not consistent

Issue #851, opened 2017-09-12
Merge fails if strings are used (Firebird)


Franki1986:
DB: Firebird 2.1.3 (dialect 1, charset ISO8859_1)
Linq2Db: 1.9.0

I am using a simple InsertWhenNotMatched Merge. Everything works fine if I merge tables where no strings are used (in firebird I use VARCHARs).

For every VARCHAR a sql statement is created with "CAST(_utf8 x'47D' AS VARCHAR(2))" 
where the number after the 'x' changes and the size of VARCHAR also... 
The weired thing is I have no VARCHARs with this size.

```
[Table("CUSTOMER")]
	public partial class Customer
	{
		[Column("ID"),                        PrimaryKey,  NotNull] public int    Id { get; set; } 
		[Column(),                                     NotNull] public string Name { get; set; } 
		[Column(),                                     NotNull] public string FirstName { get; set; } 
		[Column(),                           Nullable         ] public string Company { get; set; } 
		[Column(),                                     NotNull] public string ZipCode { get; set; }
		[Column(),                                     NotNull] public string Town { get; set; } 
	}
var multiCust = new List<Customer>()
         {
            new Customer() {Id = 405, Name = "Max", FirstName = "Mad", Town = "stut", ZipCode = "55845", Company = "My"}
         };
        try
         {
            using (var dbContext = new MyDb())
            {
               dbContext.GetTable<Customer>().Merge()
                        .Using(multiCust)
                        .OnTargetKey()
                        .InsertWhenNotMatched()
                        .Merge();
            }
         }
         catch (Exception e)
         {
            Console.WriteLine(e);            
         }
```

I get following in the log file:

_MERGE INTO CUSTOMER Target
USING
(
	SELECT 405, CAST(_utf8 x'4D6178' AS VARCHAR(3)), CAST(_utf8 x'4D6164' AS VARCHAR(3)), CAST(_utf8 x'4D79' AS VARCHAR(2)), CAST(_utf8 x'3535383435' AS VARCHAR(5)), CAST(_utf8 x'73747574' AS VARCHAR(4)) FROM rdb$database
) Source (c0, c1, c2, c3, c4, c5)
ON (Target.ID = Source.c0)

WHEN NOT MATCHED THEN
INSERT
(
	ID,
	Name,
	FirstName,
	Company,
	ZipCode,
	Town
)
VALUES
(
	Source.c0,
	Source.c1,
	Source.c2,
	Source.c3,
	Source.c4,
[...]

PR #881, opened 2017-10-19
Fix for #851

sdanyliv:
* Trying to do not encode string for Firebird if not needed. 
* Added FirebirdConfiguration.SupportsLiteralEncoding for older versions of Firebird.

### DIFF
--- a/Source/DataProvider/Firebird/FirebirdConfiguration.cs
+++ b/Source/DataProvider/Firebird/FirebirdConfiguration.cs
@@ -19,8 +19,8 @@ namespace LinqToDB.DataProvider.Firebird
 		}
 
 		/// <summary>
-		/// Specifies that Firebird supports literals encoding. Availiable from version 2.5.
+		/// Specifies that Firebird supports literal encoding. Availiable from version 2.5.
 		/// </summary>
-		public static bool SupportsLiteralEncoding = true;
+		public static bool IsLiteralEncodingSupported = true;
 	}
 }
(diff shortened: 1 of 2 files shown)


Source/DataProvider/Firebird/FirebirdConfiguration.cs
```diff
@@ -17,5 +17,10 @@ public static bool QuoteIdentifiers
 				FirebirdSqlBuilder.IdentifierQuoteMode = value ? FirebirdIdentifierQuoteMode.Quote : FirebirdIdentifierQuoteMode.None;
 			}
 		}
+
+		/// <summary>
+		/// Specifies that Firebird supports literals encoding. Availiable from version 2.5.
+		/// </summary>
+		public static bool SupportsLiteralEncoding = true;
```
igor-tkachev: It's better to call boolean flags like IsLiteralEncodingSupported.

Source/DataProvider/Firebird/FirebirdMappingSchema.cs
```diff
@@ -16,15 +18,41 @@ protected FirebirdMappingSchema(string configuration) : base(configuration)
...
+			SetValueToSqlConverter(typeof(char)  , (sb, dt, v) => ConvertCharToSql  (sb, (char)v));
 		}
 
 		static void ConvertStringToSql(StringBuilder stringBuilder, string value)
 		{
 			if (value == string.Empty)
 				stringBuilder.Append("''");
 			else
-				MakeUtf8Literal(stringBuilder, Encoding.UTF8.GetBytes(value));
+				if (FirebirdConfiguration.SupportsLiteralEncoding && value.Any(NeedsEncoding))
```
igor-tkachev: Not sure about 'Any' performance. I would use IndexOfAny instead.

PR #889, opened 2017-10-24
Added SqlEvaluateAttribute for controlling query caching.

sdanyliv:


sdanyliv:
Already fixed in #903

PR #898, opened 2017-10-31
Added Query.ClearCaches method

sdanyliv:
It may help in clearing all Query Caches.
See #894 .


Source/Linq/Query.cs
```diff
@@ -124,6 +124,24 @@ public object GetConvertedEnum(Type valueType, object value)
 		}
 
 		#endregion
+
+		#region Cache Support
+
+		internal static readonly ConcurrentBag<Action> CacheCleaners = new ConcurrentBag<Action>();
+
+		/// <summary>
+		/// Clears query caches for all typed queries
```
MaceWindu: dot missing

PR #920, opened 2017-11-21
Allow PostgreSQLBulkCopy to keep identities

evilbloodydemon:
No special mode is required in Postgres (https://www.postgresql.org/docs/9.6/static/datatype-numeric.html#DATATYPE-SERIAL), so BulkCopy can just insert data as is.


MaceWindu:
Superseeded by https://github.com/linq2db/linq2db/pull/1037

PR #935, opened 2017-11-25
Provider-specific BulkCopy for PostgreSQL

evilbloodydemon:
References:
PostgreSQL COPY documentation - https://www.postgresql.org/docs/current/static/sql-copy.html
COPY support in Npgsql - http://www.npgsql.org/doc/copy.html

It would be better to explicitly specify field type (second parameter in writer method), but I didn't figure out how to get NpgsqlDbType from field description.

### DIFF
--- a/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
@@ -1,5 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
+using LinqToDB.Extensions;
+using LinqToDB.SqlProvider;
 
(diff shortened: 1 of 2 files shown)


ili:
@evilbloodydemon cool!
It would also be great to adopt tests to test both specific and default behavior

evilbloodydemon:
@ili ok, I'm going to add AllTypes tests, similar to SqlServer ones

MaceWindu:
@evilbloodydemon what is the status of this PR? Did you have time to work on tests?

evilbloodydemon:
@MaceWindu nope, I haven't had time to work on tests. 
this solution is not reliable and requires more work - we should specify NpgsqlDbType for every field, which we can't just guess from its .net type. this should be derived from schema somehow. 
another approach is to use 'Text COPY' - then we just need to escape values properly.

MaceWindu:
Continued in #1091

PR #948, opened 2017-12-11
PostgreSQL 9.5 UPSERT support

evilbloodydemon:
INSERT ... ON CONFLICT clause (UPSERT) support for InsertOrUpdate, see https://www.postgresql.org/docs/9.5/static/sql-insert.html#SQL-ON-CONFLICT

It contains questionable code which overrides and restores table alias. But there is no existing facility to create an alias in INSERT clause and I wanted to make as little changes to the core as possible.

### DIFF
--- a/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs
@@ -16,7 +16,7 @@ namespace LinqToDB.DataProvider.PostgreSQL
 	{
 		public PostgreSQLDataProvider(PostgreSQLVersion version = PostgreSQLVersion.v92)
 			: this(
-				version == PostgreSQLVersion.v92 ? ProviderName.PostgreSQL92 : ProviderName.PostgreSQL93,
+				GetProviderName(version),
 				new PostgreSQLMappingSchema(),
 				version)
 		{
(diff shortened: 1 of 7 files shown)


MaceWindu:
Merged to continue in local branch

Alex-ttt:
Hello @MaceWindu!
I'm trying to use `InsertOrUpdate` extension method to insert as a bulk my data records. My table has unique index by two columns. If data record is not unique it should be updated or inserted otherwise.

The problem following: I want to "InsertOrUpdate" collection of records but not one query per record. Is it possible to use `InsertOrUpdate` for these proposes?

Something like this:

`
myDataSet.ToLinqToDBTable().InsertOrUpdate(records, ..., ..., ... )
`

sdanyliv:
@Alex-ttt, do you need  `InsertOrUpdate` for collections for PostgreSQL or other database?

MaceWindu:
nope, `InsertOrUpdate` API works only with single record and we currently don't have bulk operations support for postgresql with update. There are plans to add additional `InsertOrUpdate` APIs for multiple records, but not in near future.

You can use BulkCopy to instert your data to temporary table and then apply it to target table with two instert and update queries..

Alex-ttt:
@sdanyliv PostgreSQL. In case of MSSQL `Merge` methods do something similar

MaceWindu:
Don't have your expectations too high, but `MERGE` could be a thing in pgsql 15 https://commitfest.postgresql.org/36/3408/

sdanyliv:
Well, you can use temporary tables for such task.

Schematically, not checked by compiler:
```cs
using var db = context.CreateLinqToDBConnection();

using var temp = db.CreateTempTable(records);

var recordsToUpdate = 
	from d in myDataSet.ToLinqToDBTable()
	from t in temp.InnerJoin(t => t.SomeKey == d.SomeKey)
	select new { d, t }

recordsToUpdate
	.Set(u => u.d.Field1, u => u.t.Field1)
	.Set(u => u.d.Field2, u => u.t.Field3)
	.Set(u => u.d.Field3, u => u.t.Field3)
	.Update()

var recordsToInsert = 
	from t in temp
	from d in myDataSet.LeftJoin(d => t.SomeKey == d.SomeKey)
	where d == null
	select t;

recordsToInsert.Insert(myDataSet.ToLinqToDBTable(), x => x)
```

Alex-ttt:
@sdanyliv @MaceWindu 
Thank you!

Issue #744, opened 2017-06-22
Dynamic columns


boris-turk:
Hi everybody,

let me first say that the linq2db library is a wonderful piece of software, very flexible and powerful. Keep up the great work you are doing!

Now to my problem. I work for a company which has their product on the marked since quite some time. We have to be backwards-compatible with certain core features it provides. One of them is the possibility to define so-called dynamic columns in the DB. Each table (and the associated POCO class) can have additional columns not know in advance. Somebody already explained this issue (in short) in the following ticket:

   https://github.com/linq2db/linq2db/issues/507

The second suggestion from Ili (using dictionary) probably won't work, because this approach can only be used if you are filling several values into one column. But in my case there is one DB column for one value in the dictionary.

I also tried with emit and already prepared a working prototype, but then I discovered the following:

   https://stackoverflow.com/questions/2503645/reflect-emit-dynamic-type-memory-blowup
   https://goo.gl/nvCnnz

I only did some measurements for the first problem mentioned on stackoverflow. A realistic scenario in my case would be 200 dynamically generated classes, each of them having 20 additional properties. When I do that, the CLR allocates additional 10 Mb of memory, quite a lot. The problem is that the user of the system might add/remove/change columns at runtime, without closing the application. For each change (even if it is a tiny one) I have to recreate all 200 dynamically generated classes. One might think that it is OK to recreate just the one that changed, but that one could be referenced by other dynamically generated classes (through Association attribute) and it is not trivial to detect these associations (not to mention that the association chains could be long). Unloading the dynamically generated assembly is most probably not an option, because the dynamic classes might still exist somewhere in memory.

Long story short: do you have an idea how hard would it be to add this kind of feature to linq2db? Maybe it goes against the philosophy of linq2db or you are under time pressure and need to do other important things first. In that case I am willing to invest some of my time and try to come up with a solution, even if it means that I will have my own linq2db build.

It would be necessary to fix at least these things:

 * create the appropriate SELECT statements with additional fields
 * create the INSERT, UPDATE statements with additional fields and parameters
 * fill the POCO instance with additional values
 * make sure that LoadWith also takes the additional values from related entity

You have a good feeling about the library itself, so I am asking you if it would be possible to do such a change in a reasonable amount of time (e.g. one week)? If yes, which are the core classes that need to be fixed?

Thank you in advance for any kind of feedback. Best regards from hot Slovenia :)

Boris Turk

ili:
hi @boris-turk thanks for your good words.

[...]

Issue #507, opened 2016-12-18
Mapping of dynamic columns


Sp1rit:
Hi,

I need to map dynamic columns, that can be added by users at runtime (and start with a prefix), to a collection like a dictionary while selecting the data.

Is there any way to realize something like that in linq2db? Any extension points or built in functionality for that?

For Example:
````
class User {
    int Id;
    string Name;
    Dictionary<string, object> AdditionalFields;
}

**Comment by ili** [comment]
You can emit new type with needed columns

**Comment by ili** [comment]
Another way is to store `Dictionary<string, object> AdditionalFields` like serialized object using 
```cs 

MappingSchema.Default.SetConverter<Dictionary<string, object>, string>(...);
MappingSchema.Default.SetConverter<string, Dictionary<string, object>>(...);
MappingSchema.Default.SetConverter<Dictionary<string, object>, DataParameter>(...);
MappingSchema.Default.SetConverter<DataParameter, Dictionary<string, object>>(...);

```

like [here](https://github.com/linq2db/linq2db/blob/4369a08839b835cf92fe933af63590962ab80bab/Tests/Linq/UserTests/ConverterInsertTests.cs#L55)

ili:
Closing issue

PR #964, opened 2017-12-29
[WIP] Dynamic columns

toneb:
## Background

Our project has multitenant model in which customers can add columns to existing entities, or create completly new entities. These can then be used through the application in the same way as built-in entities/columns. Also, most of the core functionalities are trully generic/dynamic (like listings, forms, etc.). We have implemented custom LINQ provider very similar in design to linq2db, which supports our core scenarios, but is very limited in terms of extensibility and query translation. We're now looking to replace our custom provider with something more mature.

We tried (and are using to some extent) linq2db in combination with runtime emiting of new types for dynamic entities and columns. But that approach has some rough corners in scenarios where we want to use strongly typed models in code, which can be extended with dynamic columns. Thus, we propose to include support for dynamic columns to linq2db (as already requested in #744 and #507).

## Proposal

We propose to add a new method Sql.Property which allows to access arbitrary column/association on given entity (whether the column/association is dynamic or non-dynamic):
```C#
public static Property<T>(object entity, string propertyName)
```

### Use in queries

Accessing column "FirstName" on Person:
```C#
db.GetTable<Person>()
    .Where(x => Sql.Property<string>(x, "FirstName") == "John")
    .ToList()
```

Accessing column "Diagnosis" on associated member Person.Patient:
```C#
db.GetTable<Person>()
    .Where(x => Sql.Property<string>(x.Patient, "Diagnosis") != null)
    .ToList()
```

Combining the above examples
```C#
db.GetTable<Person>()
    .Where(x => Sql.Property<string>(Sql.Property<Patient>(x, "Patient"), "Diagnosis") != null)
    .ToList()
```

The type in Sql.Property is somewhat optional, it helps with writing typesafe code, but we could also write the above query as:
```C#
db.GetTable<Person>()
    .Where(x => Sql.Property<object>(Sql.Property<object>(x, "Patient"), "Diagnosis") != null != null)
    .ToList()
```
Which gives a great amount of flexibility in writing dynamic queries. Think about dynamic order-by, or filter:
```C#
var filteringColumn = "FirstName";
var sortingColumn = "LastName";
var filter = "John";

db.GetTable<Person>()
    .Where(p => Sql.Property<object>(p, filteringColumn) == filter)
    .OrderBy(p => Sql.Property<object>(p, sortingColumn))
    .ToList();
```

Obviously, the Sql.Property can be used in any of the query methods:
```C#
(from p in db.Person
    join pa in db.Patient on Sql.Property<string>(p, "FirstName") equals Sql.Property<string>(pa, "Diagnosis")
    select p)
    .LoadWith(x => Sql.Property<Patient>(x, "Patient"))
    .GroupBy(x => Sql.Property<string>(Sql.Property<Patient>(x, "Patient"), "Diagnosis"))
    .Select(p => new {p.Key, Count = p.Count()})
    .ToList();
```

### Use in insert/update

Sql.Property can also be used in set operators:
```C#
db
    .GetTable<Person>()
    .Value(c => Sql.Property<int>(c, "ID"), () => id)
    .Value(c => Sql.Property<string>(c, "FirstName"), () => "John1")
    .Value(c => Sql.Property<string>(c, "LastName"), () => "Test1")
    .Value(c => Sql.Property<Gender>(c, "Gender"), () => Gender.Male)
    .Insert();

db.GetTable<Person>()
    .Where(c => Sql.Property<int>(c, "ID") == id)
    .Set(c => Sql.Property<string>(c, "FirstName"), () => "Johnny1")
    .Update();
```

### Use in metadata configuration

```C#
var ms = new MappingSchema();

ms.GetFluentMappingBuilder()
    .Entity<Person>().HasTableName("Person")
    .HasPrimaryKey(x => Sql.Property<int>(x, "ID"))
    .Property(x => Sql.Property<string>(x, "FirstName")).IsNullable(false)
    .Property(x => Sql.Property<string>(x, "LastName")).IsNullable(false)
    .Property(x => Sql.Property<string>(x, "MiddleName"))
    .Association(x => Sql.Property<Patient>(x, "Patient"), x => Sql.Property<int>(x, "ID"), x => x.PersonID);
```

When the member used in Sql.Property is not defined on class, then it's added to entity definition as dynamic column (only in fluent mapping builder scenario).

### Storing dynamic columns/associations data

The class CAN HAVE a "DynamicColumnsStore" property, which has to be assignable from Dictionary type. For example:
```C#
[DynamicColumnsStore]
public IDictionary<string, object> ExtendedProperties { get; set; }
```

This property is required only when we want to have dynamic columns/associations accessible in materialized entity. "Sql.Property" can be used in queries with or without the dynamic columns store property defined. 

## Implementation

The proposed approach has minimal impact on public API of linq2db (only one additional public API method - Sql.Property). 

Dynamic properties are represented as new type "DynamicColumnInfo" inheriting from "PropertyInfo". This enables implementation of dynamic columns with minimal changes to linq2db code. Note that this approach does not work in NETSTANDARD1_6, since it doesn't have required APIs.

Included are unit tests covering querying, updating, and metadata configuration scenarios. I will also prepare documentation chapter if this proposal gets approved.

### DIFF
--- a/Source/LinqToDB/Expressions/MemberHelper.cs
+++ b/Source/LinqToDB/Expressions/MemberHelper.cs
@@ -11,8 +11,6 @@ namespace LinqToDB.Expressions
 
 	public static class MemberHelper
 	{
-		private static readonly MemberInfo SQLPropertyMethod = MethodOf(() => Sql.Property<string>(null, null)).GetGenericMethodDefinition();
-
 		/// <summary>
 		/// Gets the member information from given lambda expression. <seealso cref="GetMemberInfo(System.Linq.Expressions.Expression)"/>
 		/// </summary>
@@ -47,8 +45,7 @@ namespace LinqToDB.Expressions
 			if (expr.NodeType == ExpressionType.New)
 				return ((NewExpression)expr).Constructor;
 
-			if (expr is MethodCallExpression methodCall && methodCall.Method.IsGenericMethod &&
-			    methodCall.Method.GetGenericMethodDefinition() == SQLPropertyMethod)
+			if (expr is MethodCallExpression methodCall && methodCall.Method.IsSqlPropertyMethodEx())
 			{
 				// validate expression and get member name
 				var arg1 = methodCall.Arguments[0].NodeType == ExpressionType.Convert
--- a/Source/LinqToDB/Expressions/MemberHelper.cs
+++ b/Source/LinqToDB/Expressions/MemberHelper.cs
@@ -12,11 +12,11 @@ namespace LinqToDB.Expressions
 	public static class MemberHelper
 	{
 		/// <summary>
-		/// Gets the member information from given lambda expression. <seealso cref="GetMemberInfo(System.Linq.Expressions.Expression)"/>
+		/// Gets the member information from given lambda expression. <seealso cref="GetMemberInfo(Expression)" />
 		/// </summary>
 		/// <param name="func">The lambda expression.</param>
 		/// <returns></returns>
-		/// <exception cref="ArgumentException">Only simple, non-navigational, member names are supported in this context (e.g.: x => Sql.Property(x, \"SomeProperty\")).</exception>
+		/// <exception cref="ArgumentException">Only simple, non-navigational, member names are supported in this context (e.g.: x =&gt; Sql.Property(x, \"SomeProperty\")).</exception>
 		public static MemberInfo GetMemberInfo(LambdaExpression func)
 		{
 			return GetMemberInfo(func.Body);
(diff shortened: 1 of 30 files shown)


toneb:
@igor-tkachev 

> Sql.Property<string>(x, "FirstName") can be replaced/aliased with x.SqlColumn<string>("FirstName")

Sure, I don't mind how it's named. (we're talking about a rename here right?, I couldn't find an existing method named SqlColumn in the solution?)

> I think we could also have [DynamicColumnNames] and static list to keep all the extended property names.

It could be as additional feature, but shouldn't be required as implementation detail. For instance, we have multiple tenants, each with it's own extended columns, but running on the same web-server/process, so there is a possibility of conflicts ...

Source/LinqToDB/Expressions/MemberHelper.cs
```diff
@@ -1,27 +1,82 @@
...
+#else
+				throw new NotSupportedException("Dynamic columns are not supported on .NET Standard 1.6.");
+#endif
+			}
 
 			return
-				ex is MemberExpression     ? ((MemberExpression)    ex).Member :
-				ex is MethodCallExpression ? ((MethodCallExpression)ex).Method :
-				                 (MemberInfo)((NewExpression)       ex).Constructor;
+				expr is MemberExpression     ? ((MemberExpression)    expr).Member :
```
jogibear9988: If you changed here, avoid double cast, use result from "is"

MaceWindu:
I'm merging it into a separate branch to resolve conflicts

Source/LinqToDB/Expressions/MemberHelper.cs
MaceWindu: I will fix it. I think we migrated to latest C# version after this PR was created

MaceWindu:
@toneb , I've merged latest master to it and pushed as new PR #1073. Could you create documentation for this feature?

boris-turk:
Thank you, thank you, thank you!! You guys rock! This is the only thing which I missed in linq2db till now. Big thanks also to @toneb for the great suggestion about how to extend the API. Nice and elegant.

guillaume86:
@toneb fantastic work. I'm in the exact same situation as the one you explained in your original post (custom columns / custom tables by tenant).

Do you still emit runtime types for the custom tables?

sdanyliv:
@apollo82, please open separate issue for this to track conversation.

PR #984, opened 2018-01-25
fix mapping of simple types with subproperties (fixes #983)

### LINKED ISSUE #983: Mapping of SubProperties via FluentMapping does not work if subproperties are a simple Type (like int)
This does not work, because C# then adds a Convert Expression to the Expression Tree, wich is not handeled in our Code!

A fix follows in 5 minutes.

But I could not create a UnitTest, cause I have to fix it in Release 1.10, and I could not run the tests with VS2017!

jogibear9988:


jogibear9988:
see test:

		[Test, DataContextSource]
		public void SelectComplex3(string context)
		{
			var ms = new MappingSchema();
			var b  = ms.GetFluentMappingBuilder();

			b
				.Entity<ComplexPerson3>()        .HasTableName ("Person")
				.Property(_ => _.ID)             .HasColumnName("PersonID")
				.Property(_ => _.Name.FirstName) .HasColumnName("FirstName")
				.Property(_ => _.Name.LastName)  .HasColumnName("LastName")
				.Property(_ => _.Name.MiddleName).HasColumnName("MiddleName");

It works, because all subproperties are of type "string", it does not work for "int"

jogibear9988:
Is the Pull Req against "release" branch correct? Should I also create a fix against "master" or will the release be merged into master?

Source/Mapping/PropertyMappingBuilder.cs
```diff
@@ -119,7 +119,12 @@ PropertyMappingBuilder<T> SetColumn(Action<ColumnAttribute> setColumn)
 		{
 			var getter     = _memberGetter;
 			var memberName = null as string;
-			var me         = _memberGetter.Body as MemberExpression;
+			var body       = _memberGetter.Body;
```
sdanyliv: use _memberGetter.Body.Unwarp() here

jogibear9988:
Next time I will do so!

jogibear9988:
@sdanyliv is the patch now ok?
@ili or @sdanyliv or @igor-tkachev can this be merged?
will we create a new bugfix release?

sdanyliv:
Merged by hand into version1 - it is right branch for your fix.
Will check when we can release 1.10.X version.

Issue #924, opened 2017-11-22
Read out Metadata of all mapped Types...


jogibear9988:
I still have the Problem, I need to get all Metadata of all Mapped types to extracte Table and Column Names for a Reporting Tool.

I tried to use GetEntites() of MappingSchema wich was introduced 
 closing this commit: https://github.com/linq2db/linq2db/issues/637
But it returns a empty list of Types, and after it the mapping schema is not yet ready filled with the entities...

I use Fluent mapping!

sdanyliv:
@jogibear9988, PR will be appreciated. You know many things in that area.

jogibear9988:
i look... only thought maybe I did smth. wrong

jogibear9988:
I've looked at the Code. It would be good if the Mapping Schema could return all the mapped Types. For this, every MetaDataReader should return a List of it's types. Should I add a GetMappedTypes function to IMetaDataReader and MappingSchema (but that would be a breaking change, because I had to modify a Interface)?

jogibear9988:
Another Problem on my COde was, I used GetFluentMappingBuilder() call for each type, so I had a huge tree of MetaDataReaders...

jogibear9988:
@sdanyliv as now version2 branch is merged, i should i work on this? will a change of the interfaces be accepted?

sdanyliv:
Please use current master branch. I will review your PR.

jogibear9988:
I'll look to work on this this week

jogibear9988:
is fixed in #1587

MaceWindu:
#1587 merged

PR #985, opened 2018-01-27
Proposal for implementing #924

jogibear9988:


jogibear9988:
is fixed in #1587 see https://github.com/linq2db/linq2db/pull/1587/commits/9c40dd5bf501c27a260bbb93f9c68561ecbe2c0d#diff-d98d72d4eeed6068720f7e6b185cb707R1503

PR #1000, opened 2018-02-07
Implemented automatic INNER Join to LEFT Join converter for Firebird.

### LINKED ISSUE #982: Wrong plan for inner join (Firebird internal issue)
Hello, 

It looks like firebird has an internal bug regarding inner join plans. 
If you make an inner join on big tables you will get a natural plan for the inner joined table, this can cause big performance problems.
Example can be found here :

http://tracker.firebirdsql.org/browse/CORE-3707

This is not a linq2db issue, but due to the fact that this issue won't be fixed soon, can we create a generic between solution?
Maybe something like adding a 'where primariy key is not null' to the hole sql, where primary key is of the non inner joined table.

sdanyliv:
Fixes #982


MaceWindu:
@Franki1986 this fix is ready and we need your confirmation that it solves your issue before we can merge it. Otherwise it doesn't make sense to include it.
Could you build this branch, enable fix by
```cs
FirebirdConfiguration.ConvertInnerJoinsToLeftJoins = true
```
and verify that it work for you.
Or just test query that converts inner join to left join using following logic:
- match condition goes to where
- join's match condition set to 1=1

Using your sample query:
```sql
select s.* from STUDENTS s
join PERSONS p on p.Id = s.PersonId
where p.AGE > 20;
```
will be converted to
```sql
select s.* from STUDENTS s
join PERSONS p on 1=1
where p.AGE > 20 and p.Id = s.PersonId;
```

Franki1986:
Tomorrow I am back at the pc, then I will test.

Source/LinqToDB/DataProvider/Firebird/FirebirdMergeBuilder.cs
```diff
@@ -81,5 +83,37 @@ protected override bool SupportsParametersInSource
...
 				base.AddSourceValue(valueConverter, column, columnType, value, isFirstRow);
 		}
+
+		protected override void BuildMatch()
+		{
+			var old = FirebirdConfiguration.DisableConvertInnerJoinsToLeftJoins.Value;
+
+			try
+			{
+				FirebirdConfiguration.DisableConvertInnerJoinsToLeftJoins.Value = true;
```
sdanyliv: What is it? Looks like a hack.

MaceWindu: For merge we generate fragments of query, like join condition, and inner join converson breaks it. I've added this internal flag to disable this conversion during merge query generation for now. it will not be required later when merge will be migrated to AST

Franki1986:
Ok,

tested it with two tables:

Table A (students): 500.000 records
Table B (persons)  : 7.500     records

```
select s.* from STUDENTS s
left join PERSONS p on 1=1
where p.AGE > 20 and p.Id = s.PersonId;
```

looks like this query is still producing a natural plan for join... 

Also the null check is not considered from firebird:

```
select s.* from STUDENTS s
left join PERSONS p on 1=1
where p.AGE > 20 and p.Id = s.PersonId and s.Id is not null;
```

Only thing that worked is checking the referenced ID with some value:

```
select s.* from STUDENTS s
left join PERSONS p on p.Id = s.PersonId where s.Id > -1;
```

I don't know if checking the associated key against a value can be made in a generic way?!

MaceWindu:
ok, I will try to reproduce this case locally and test different approaches

MaceWindu:
@Franki1986 
tried several options - check for null or moving join condition indeed doesn't affect plan, but adding comparison condition over key field fixes query even with inner join.
I will try to implement your proposal in following way:
1. applied to all joins (as I see natural plan for all joins: inner, left, right, cross)
2. join type not changed
3. allways true condition over key fieds added to WHERE

MaceWindu:
Also I think it makes sense to limit supported key field types to ranged types (numbers, dates, need to check what types allowed by firebird here), as I see you also have an issue with condition for string types :)

Franki1986:
Sorry for the delayed problem discovery! Could have been found out earlyer from my side..

MaceWindu:
@Franki1986 this could work. Will check later, but not sure if it will improve performance or make it worse

MaceWindu:
Looks like char_length doesn't help and I'm not surprised, as we apply function to each record

Franki1986:
What about comparing a string to empty string? 
`where s.StringId >= ''`
should always succeed.

Franki1986:
Had two sleepless night :-) but I think this should work:

```
switch (k.DataType)
							{
								case DataType.Decimal:
								case DataType.Money:
								case DataType.SmallMoney:
									maxSizeDecimal = decimal.MaxValue;
									break;
								case DataType.SByte:
								case DataType.Byte:
									maxSizeLong = System.Byte.MaxValue;
									break;
								case DataType.Char:
								case DataType.NChar:
									maxSizeLong = System.Char.MaxValue;
									break;
								case DataType.Double:
									maxSizeDouble = System.Double.MaxValue;
									break;
								case DataType.Int32:
									maxSizeLong = System.Int32.MaxValue;
									break;
								case DataType.Int16:
									maxSizeLong = System.Int16.MaxValue;
									break;
								case DataType.Int64:
									maxSizeLong = System.Int64.MaxValue;
									break;
								case DataType.UInt32:
									maxSizeLong = System.UInt32.MaxValue;
									break;
								case DataType.UInt16:
									maxSizeLong = System.UInt16.MaxValue;
									break;
								case DataType.UInt64:
									maxSizeUlong = System.UInt64.MaxValue;
									break;
								case DataType.VarChar:
								case DataType.NVarChar:
									maxStringExpression = "";
									break;
								case DataType.Date:
									maxDateTime = new System.DateTime(9999, 09, 09);
									break;
								case DataType.DateTime:
								case DataType.DateTime2:
								case DataType.DateTimeOffset:
								case DataType.Time:
									maxDateTime = new System.DateTime(9999, 09, 09, 09, 09, 09);
									break;
							}
							SqlSearchCondition cond = null;
							if (maxStringExpression != null)
							{
								cond = new SqlSearchCondition();
								cond.Expr(sqlExpressions.First()).GreaterOrEqual.Expr(new SqlExpression(maxStringExpression));
							}
							else if (maxDateTime != null)
							{
								cond = new SqlSearchCondition();
								cond.Expr(sqlExpressions.First()).Less.Expr(new SqlExpression(typeof(System.DateTime), maxDateTime.Value.ToString("O")));
							}
							else
							{
								maxStringExpression = maxSizeLong?.ToString() ?? maxSizeDecimal?.ToString() ?? maxSizeDouble?.ToString() ?? maxSizeUlong?.ToString();
								if (maxStringExpression != null)
								{
									cond = new SqlSearchCondition();
									cond.Expr(sqlExpressions.First()).Less.Expr(new SqlExpression(maxStringExpression));
								}
							}
							if (cond != null)
							{
								q.Select.Where.ConcatSearchCondition(cond);
							}
```
For strings I am testing that they are greater that empty string. For DateTime I am testing that they are less than '09.09.9999 09:09:99', this is a valid DateTime but will never be used..

MaceWindu:
I will check it later this week, but could you provide some information on what kind of query (and amount of data) this issue could be reproduced with noticable performance impact.
I want to be sure that this fix not just change plan, but also that it helps with original issue (low performance)

MaceWindu:
I've tried to reproduce it again (including case from FB Jira issue), but still cannot see real performance boost and only see changes in plan and slight performance degradation due to extra conditions.

In this situation I'm strictly opposed to include this into linq2db at this stage as I'm not sure it will add any value. For now I took different approach to help you with this issue:
1. I modified code you provided to not fail on any of our existing tests
2. made firebird sql optimizer public to allow it's subclassing
3. moved changes to a test with subclassed optimizer

This will allow you to use this optimization and modify it without making changes to linq2db and if you will see real performance improvements, we can review it again later.

If everybody aggrees with this approach, I propose to merge this PR

Franki1986:
I am sorry for not giving you a complete test scenario, these weeks I am in hospital so haven't my pc here. 
I think the suggestion with making the optimizer public is indeed the best solution, so anyone who has a database constellation like mine can handle this problem. Due to the fact that this should only be a workaround until there is a fix from firebird side. So I think this is the best solution. When I am ready with my complete workaround I will post it here. 
Thanks a lot for giving such an excellent support to your product!

MaceWindu:
Don't worry and get better :)
If we will be able to see real effect from this fix, we will include it as an option into library.

Issue #170, opened 2015-02-16
Left join over association field returns default value instead of null for non-nullable fields.


MaceWindu:
db.Table.Select(_ => new { Value =(int?) _.LeftJoinAssociation.IntegerField }).ToList();

Linq query above always return default value for non-nullable fields, selected from left-join associations, even if result type hint (cast) specified. Generated SQL is correct and returns NULL values.

http://rsdn.ru/forum/prj.rfd/5954281.flat#5954281

MaceWindu:
Thanks for quick fix.
I have regression with it currently. Looks like cast used too aggressivly now.
I have text field mapped to enumeration, and in my Select statement I used it like that:

    db.Table.Select(_ => new { (byte)_.EnumField })

(I use byte, because my enum iherits from byte).

Before that fix, cast performed after text-to-enum mapping, but now I think it skips that step and tries to cast string to byte.

    System.FormatException: Input string was not in a correct format.
       at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
       at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
       at System.Byte.Parse(String s, NumberStyles style, NumberFormatInfo info)
       at lambda_method(Closure , IDataReader )
       at LinqToDB.Expressions.ConvertFromDataReaderExpression.ColumnReader.GetValue(IDataReader dataReader) in c:\GitHub\linq2db\linq2db\Source\Expressions\ConvertFromDataReaderExpression.cs:line 128
       at lambda_method(Closure , QueryContext , IDataContext , IDataReader , Expression , Object[] )
       at LinqToDB.Linq.Query`1.<Map>d__6a.MoveNext() in c:\GitHub\linq2db\linq2db\Source\Linq\Query.cs:line 1181

PR #1001, opened 2018-02-07
Fix for #170 when you select data from Many association

MaceWindu:


MaceWindu:
do not merge, will make another PR with tests a bit later

Issue #996, opened 2018-02-06
LoadWith used with Nullable<int> foreign key throws exception => "Inheritance mapping is not defined for discriminator value '' "


dubbeglas:
LoadWith used with an int works fine but the usage of a Nullable<int> foreign key leads to the exception
**Inheritance mapping is not defined for discriminator value '' in the 'LinqToDbTest.A' hierarchy**
```
Exception message:
LinqToDB.Linq.LinqException was unhandled
  HResult=-2146233088
  Message=Inheritance mapping is not defined for discriminator value '' in the 'LinqToDbTest.A' hierarchy.
  Source=linq2db
  StackTrace:
       bei LinqToDB.Linq.Builder.TableBuilder.TableContext.DefaultInheritanceMappingException(Object value, Type type)
       bei lambda_method(Closure , IQueryRunner , IDataReader )
       bei LinqToDB.Linq.QueryRunner.Mapper`1.Map(IQueryRunner queryRunner, IDataReader dataReader)
       bei LinqToDB.Linq.QueryRunner.<ExecuteQuery>d__8`1.MoveNext()
       bei System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
       bei System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
       bei LinqToDbTest.LinqToDbIssue.LoadTest() in c:\users\ckreikler\documents\visual studio 2015\Projects\LinqToDbTest\LinqToDbTest\TestClass.cs:Zeile 107.
       bei LinqToDbTest.LinqToDbIssue..ctor() in c:\users\ckreikler\documents\visual studio 2015\Projects\LinqToDbTest\LinqToDbTest\TestClass.cs:Zeile 58.
       bei LinqToDbTest.Program.Main(String[] args) in c:\users\ckreikler\documents\visual studio 2015\Projects\LinqToDbTest\LinqToDbTest\Program.cs:Zeile 13.
       bei System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       bei System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
       bei Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
       bei System.Threading.ThreadHelper.ThreadStart_Context(Object state)
[...]

Issue #994, opened 2018-02-05
Load base class entity and "LoadWith" ref properties of derived classes 


dubbeglas:
Given the following scenario mapped with Inheritance (Attributes missing and not necessary to understand)

```c#
    public class Eye
    {
        public string Xy { get; set; }
    }

    public class Animal
    {
        public string Name { get; set; }

        public string Discriminator { get; set; }
    }

    public class Dog : Animal
    {
        public Eye Bla { get; set; }
    }
```

Loading all "Animal"s gives me "Animal"s and "Dog"s.
```c#
db.GetTable<Animal>().ToList();

```

To load the ref property Bla in Dog I expected something like

```c#
db.GetTable<Animal>().LoadWith<Dog>(x => x.Bla)
```

How can I "LoadWith" the Bla property?

sdanyliv:
I need to check sources if it is possible. Usually i do not use inheritance when working with database. DB is not OOP ;)

dubbeglas:
It was not my idea... ;-)

jogibear9988:
@sdanyliv @ili @igor-tkachev please review

PR #1002, opened 2018-02-08
Test for #994 #996 und #1006

### LINKED ISSUE #966: Tests CharIndex1 and CharIndex2 supported by Firebird
linq2db version: *master*  

Tests\Linq\Linq\StringFunctionTests.cs
- CharIndex1
- CharIndex2

These tests supported by Firebird data provider.

------ 1
DataConnection: -- Firebird
SELECT
	t1.FirstName,
	t1.PersonID,
	t1.LastName,
	t1.MiddleName,
	t1.Gender
FROM
	Person t1
WHERE
	Position('oh', t1.FirstName) = 2 AND t1.PersonID = 1

DataConnection: Query Execution Time: 00:00:00.0280006

DataConnection: Total Execution Time: 00:00:00.3205403. Rows Count: 1.

------ 2
DataConnection: -- Firebird
SELECT
	t1.FirstName,
	t1.PersonID,
	t1.LastName,
	t1.MiddleName,
	t1.Gender
FROM
	Person t1
WHERE
	Position('p', t1.LastName, 2) = 3 AND t1.PersonID = 1

DataConnection: Query Execution Time: 00:00:00.0230004

DataConnection: Total Execution Time: 00:00:00.3068433. Rows Count: 1.

jogibear9988:
This adds Tests for #994 and #996 and a fix for #994 
#996 still needs to be fixed correctly

--
edited by @MaceWindu

List of remaining tasks
- [x] create isolated test for incorrect table name generated when #994 and #996 tests run in parallel and fix it
- [ ] extract #924 changes to separate PR
- [ ] extract #994 changes to separate PR
- [ ] extract #996 changes to separate PR
- [ ] extract #1006 changes to separate PR
- [ ] test for flag enums

List of remaining questions
- [ ] if we map an enum to a string column, should we use ToString() as default? If not it should be configurable or the GetDefaultMappingFromEnumType is nearly useless
- [ ] how should we get the converter in ExpressionBuilder.SqlBuilder, and should we cache the expression?
- [ ] to fix #966 I'll check in TableBuilder.TableContext.cs if tableContext is AssociatedTableContext and then i remove the exception throwing, is this enough? wrong values in the column then also throw no exception

### DIFF
--- a/Source/LinqToDB/Extensions/MappingExtensions.cs
+++ b/Source/LinqToDB/Extensions/MappingExtensions.cs
@@ -27,7 +27,7 @@ namespace LinqToDB.Extensions
 				{
 
 					var type = Converter.GetDefaultMappingFromEnumType(mappingSchema, systemType);
-					if (type == typeof(string))
+					if (Configuration.UseEnumValueNameForStringColumns && type == typeof(string))
 						return new SqlValue(type, value.ToString());
 
 					return new SqlValue(type, Converter.ChangeType(value, type, mappingSchema));
--- a/Source/LinqToDB/Extensions/MappingExtensions.cs
+++ b/Source/LinqToDB/Extensions/MappingExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using LinqToDB.Data;
 
 namespace LinqToDB.Extensions
 {
@@ -24,7 +25,10 @@ namespace LinqToDB.Extensions
 			{
 				if (value != null || systemType == underlyingType)
 				{
+
 					var type = Converter.GetDefaultMappingFromEnumType(mappingSchema, systemType);
+					if (type == typeof(string))
+						return new SqlValue(type, value.ToString());
 
 					return new SqlValue(type, Converter.ChangeType(value, type, mappingSchema));
 				}
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -1668,19 +1668,31 @@ namespace LinqToDB.Linq.Builder
 						var origValue = Enum.Parse(type, name, false);
 
 						if (!dic.TryGetValue(origValue, out var mapValue))
-							mapValue = ((ConstantExpression)value).Value;
+							mapValue = origValue; //((ConstantExpression)value).Value;
 
 						ISqlExpression l, r;
+
+						SqlValue sqlvalue;
+						var ce = MappingSchema.GetConverter(type, typeof(DataParameter), false);
+						if (ce != null)
+						{
+							mapValue = ((DataParameter) ce.Lambda.Compile().DynamicInvoke(origValue)).Value;
+							sqlvalue = new SqlValue(mapValue);
+						}
+						else
+						{
+							sqlvalue = MappingSchema.GetSqlValue(type, mapValue);
+						}
 
 						if (left.NodeType == ExpressionType.Convert)
 						{
 							l = ConvertToSql(context, operand);
-							r = MappingSchema.GetSqlValue(type, mapValue);
+							r = sqlvalue;
 						}
 						else
 						{
 							r = ConvertToSql(context, operand);
-							l = MappingSchema.GetSqlValue(type, mapValue);
+							l = sqlvalue;
 						}
 
 						return Convert(context, new SqlPredicate.ExprExpr(l, op, r));
@@ -1964,14 +1976,7 @@ namespace LinqToDB.Linq.Builder
 			BuildParameterType  buildParameterType = BuildParameterType.Default)
 		{
 			var type        = accessorExpression.Type;
-			var defaultType = Converter.GetDefaultMappingFromEnumType(dataContext.MappingSchema, type);
-
-			if (defaultType != null)
-			{
-				var enumMapExpr = dataContext.MappingSchema.GetConvertExpression(type, defaultType);
-				accessorExpression = enumMapExpr.GetBody(accessorExpression);
-			}
-
+			
 			LambdaExpression expr = null;
 			if (buildParameterType != BuildParameterType.InPredicate)
 				expr = dataContext.MappingSchema.GetConvertExpression(type, typeof(DataParameter), createDefault: false);
@@ -1982,6 +1987,16 @@ namespace LinqToDB.Linq.Builder
 
 				accessorExpression         = Expression.PropertyOrField(body, "Value");
 				dataTypeAccessorExpression = Expression.PropertyOrField(body, "DataType");
+			}
+			else
+			{
+				var defaultType = Converter.GetDefaultMappingFromEnumType(dataContext.MappingSchema, type);
+
+				if (defaultType != null)
+				{
+					var enumMapExpr = dataContext.MappingSchema.GetConvertExpression(type, defaultType);
+					accessorExpression = enumMapExpr.GetBody(accessorExpression);
+				}
 			}
 
 			// see #820
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -1676,6 +1676,10 @@ namespace LinqToDB.Linq.Builder
 						var ce = MappingSchema.GetConverter(type, typeof(DataParameter), false);
 						if (ce != null)
 						{
+							var p = Expression.Parameter(typeof(object));
+							var a = Expression.Lambda<Func<object, DataParameter>>(Expression.Invoke(ce.Lambda, Expression.Convert(p, type)), p);
+							var tr = a.Compile();
+							var tt = tr(origValue);
 							mapValue = ((DataParameter) ce.Lambda.Compile().DynamicInvoke(origValue)).Value;
 							sqlvalue = new SqlValue(mapValue);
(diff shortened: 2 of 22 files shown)


Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -468,6 +468,8 @@ select ConvertToParentIndex(_indexes[f].Index, null)
 								Expression.Constant(dindex)),
 							Expression.Constant(ObjectType)),
 						ObjectType);
+
+					//expr = Expression.Constant(null, ObjectType);
```
jogibear9988: This will fix #996, but then we do not have any exception on wrong discriminators

Source/LinqToDB/Linq/Builder/LoadWithBuilder.cs
```diff
@@ -14,7 +14,7 @@ class LoadWithBuilder : MethodCallBuilder
 	{
 		protected override bool CanBuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
 		{
-			return methodCall.IsQueryable("LoadWith");
+			return methodCall.IsQueryable("LoadWith") || methodCall.IsQueryable("LoadWithDerived");
```
jogibear9988: should we switch to use nameof(smth) here?

jogibear9988:
i still need help for the correct fix, so we only throw exception for discriminator null when no association is used

sdanyliv:
@jogibear9988, sorry i’m on vacation. Maybe someone else can help you.

jogibear9988:
i‘ve now added a change, so ony when calledfrom a association it‘s no error! need a review

Source/LinqToDB/Linq/Builder/LoadWithBuilder.cs
```diff
@@ -109,6 +109,17 @@ static IEnumerable<MemberInfo> GetAssociations(ExpressionBuilder builder, Expres
 							goto default;
 						}
 
+					case ExpressionType.Convert:
+						{
+							if (expression is UnaryExpression)
```
sdanyliv: Looks like here you can convert without type check.

jogibear9988:
Now there is a Bug left with ComponentMapping... I've to look whats wrong...

jogibear9988:
@sdanyliv can you review again? Now it should fix multiple Problems with Component Mapping & Inheritance Mapping

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1668,19 +1668,31 @@ ISqlPredicate ConvertEnumConversion(IBuildContext context, Expression left, SqlP
...
-							mapValue = ((ConstantExpression)value).Value;
+							mapValue = origValue; //((ConstantExpression)value).Value;
 
 						ISqlExpression l, r;
 
+						SqlValue sqlvalue;
+						var ce = MappingSchema.GetConverter(type, typeof(DataParameter), false);
+						if (ce != null)
+						{
+							mapValue = ((DataParameter) ce.Lambda.Compile().DynamicInvoke(origValue)).Value;
```
jogibear9988: @ili @sdanyliv or @igor-tkachev 
have you a better way for me to convert the value? I think this will not be very performant

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1983,6 +1988,16 @@ ISqlExpression GetParameter(Expression ex, MemberInfo member)
 				accessorExpression         = Expression.PropertyOrField(body, "Value");
 				dataTypeAccessorExpression = Expression.PropertyOrField(body, "DataType");
 			}
+			else
+			{
+				var defaultType = Converter.GetDefaultMappingFromEnumType(dataContext.MappingSchema, type);
```
jogibear9988: I moved the DefaultMappingFromEnumType down here, I think if a special Converter is defined, it should be used with higher priority

Source/LinqToDB/Extensions/MappingExtensions.cs
```diff
@@ -24,7 +25,10 @@ public static SqlValue GetSqlValue(this MappingSchema mappingSchema, Type system
 			{
 				if (value != null || systemType == underlyingType)
 				{
+
 					var type = Converter.GetDefaultMappingFromEnumType(mappingSchema, systemType);
+					if (type == typeof(string))
+						return new SqlValue(type, value.ToString());
```
jogibear9988: i think for strings the default for enums should be "ToString" insteady of the value as string

Source/LinqToDB/Linq/Builder/LoadWithBuilder.cs
jogibear9988: is fixed

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1668,19 +1668,35 @@ ISqlPredicate ConvertEnumConversion(IBuildContext context, Expression left, SqlP
 						var origValue = Enum.Parse(type, name, false);
 
 						if (!dic.TryGetValue(origValue, out var mapValue))
-							mapValue = ((ConstantExpression)value).Value;
+							mapValue = origValue; //((ConstantExpression)value).Value;
 
 						ISqlExpression l, r;
 
+						SqlValue sqlvalue;
+						var ce = MappingSchema.GetConverter(type, typeof(DataParameter), false);
```
jogibear9988: For me there are 2 questions left:

- Is this a good way to get the converter ?
- Should we cache the Compiled Expression in a Field in the object returned by MappingSchema.GetConverter

jogibear9988: I think these need to be answered of one of the core developers

sdanyliv:
@jogibear9988. I've fixed Travis build. There are strange bugs in your branch.

jogibear9988:
habe you also looked at #1019 about missing fields?

jogibear9988:
@sdanyliv I don't know why I have ths strange bugs atm. but when you look at #1019, there are tables missing in the test database. and I've checked it with sqlite browser, they are really missing!

MaceWindu:
I'm looking into it. Will refactor some stuff first like moving DDL operations to scripts

MaceWindu:
One problem that I see with this PR - it contains code at least for four unrelated issues: #924, #994, #996 and #1006.
We can continue work in this stream but I think we will integrate it using separate PRs for each issue.

MaceWindu:
SQL logic error or missing database "no such table: B"

this error occur only if you run both 994 and 996 tests. I suspect it is related to creation and drop of the table with the same name Test in tests and will dissapear if we move DDL to scripts.
But before that I wan't to find out why we generate incorrect table name in this case, because it is definitly our bug

MaceWindu:
PostgreSQL error: column t1.char20DataType does not exist

this column defined twice in mapping and marked as IsColumn = false for one of column. Still generated query contains select for two columns, where one is escaped.
Still need to identify what change broke it

MaceWindu:
*Instances of abstract classes cannot be created*

Inheritance tests broken by changes in this pr. Generate mapper tries to create instance of base abstract class

MaceWindu:
inheritance tests failure: *Expected and result lists are different. Length:   Expected: 3  But was:  5*

another regression - now test query returns also other types with the same discriminator. Need to check it more detaily on what behavior required here

MaceWindu:
MySQL: *Unknown column 'FieldDateTime2' in 'field list'*
SQLIte: *table TestMerge1 has no column named FieldDateTime2*
same as PostgreSQL issue - bogus SQL generated for column marked as IsColumn = false

MaceWindu:
@jogibear9988 , what I propose to.
I will continue work in this stream for now and will try to identify exact reasons for broken tests and you will create separate PRs for each issue meantime. This will allow us to close issues that are ready and concentrate on remaining issues.

jogibear9988:
I know that I‘ve mixed up patches for multiple issues. If the tests are fixed, and my questions are answerd (for example if it‘s the correct way to get the converter, and if the expression should be cached), I can fix that, if it‘s required.
I will look why the mapper tries to create a instance of the abstract class

MaceWindu:
narrowed down issues to this:
- inheritance tests broken by changes to ExpressionBuilder.SqlBuilder.cs
- IsColumn = false broken by changes to EntityDescriptor.cs
- sqlite table issue - broken by changes to query runner

MaceWindu:
Update for SQLite table issue:
MappingSchema.GetEntityDescriptor() access first schema for entity descriptor, but in this case needed type located in Schema[2]. This looks wrong to me that some methods of MappingSchema enumerate schemas, and some just take first one.
To me it looks like existing bug in MappingSchema

jogibear9988:
@MaceWindu 
can you help me with some of the bugs?

changes in entitydescriptor are needed, component mapping is broken in current code. look at issue #1005 the tests do not cover that the data is read.

changes in ExpressionBuilder.SqlBuilder.cs are needed, because atm. at this place the converters for a column are ignored

jogibear9988:
If everything is working i will create new commits with only the fixes and a description for what they are.

jogibear9988:
i've added a few questions at the top

jogibear9988:
move all code to new pull requests

Issue #1026, opened 2018-03-03
InsertWithIdentity..., Update and Delete with tablename, schemaname, databasename overrides (like Insert)


beercsaba:
DataExtensions.Insert has a variant where the tablename, schemaname and databasename can be overridden.
InsertWithIdentity..., Update and Delete should also have a variant like this.

Implementation: 
- CreateQuery method in QueryRunner.Insert class has tableName, schemaName and databaseName parameters and uses them correctly.
- CreateQuery method in QueryRunner.insertWithIdentity...,  QueryRunner.Update or QueryRunner.Delete has no parameters like these...

sdanyliv:
From start check this issue:
https://github.com/linq2db/linq2db/issues/987
If you still need new API, it will be good to propose exact interface for these methods and unit tests.

beercsaba:
Hi @sdanyliv!
workarounds described in #987 could help, but a more elegant way would be far better :)
And this elegant way is already implemented in DataExtensions.Insert, but the very same thing is not implemented in DataExtensions.Update, Delete and InsertWith...Identity variants.
There is nothing to invent, maybe exactly the same code would fit (or with minimal obvious modifications) in these other extension methods to make them consistent.

in code:
QueryRunner.Insert.CreateQuery(..) method has these tableName, databaseName, schemaName parameters and contains these rows:
`if (tableName    != null) sqlTable.PhysicalName = tableName;`
`if (databaseName != null) sqlTable.Database     = databaseName;`
`if (schemaName   != null) sqlTable.Schema       = schemaName;`

but QueryRunner.Update.CreateQuery(...) does not, it uses simply the names defined in its sqlTable instance (which comes from attributes, i think)

so IMHO, if these 3 parameters and these 3 lines of code would be added into every QueryRunner.xxx.CreateQuery method, then
- it would be elegant enough
- it would be consistent in itself

What do You think?

(or should i implement it and create a pull request?)

sdanyliv:
@beercsaba, PR is always wellcome ;)

beercsaba:
short and quite essential answer  :)

sdanyliv:
From start just post API interface.

sdanyliv:
Currently we are under high load, and help is very appreciated ;)

sdanyliv:
There is sample API, shown on Update operation (same for others) :

**First variant**
```cs
public static int Update<T>([NotNull] this IDataContext dataContext, T obj, 
[...]

PR #1042, opened 2018-03-17
Issue #1026

beercsaba:
implementation of Issue #1026

### DIFF
--- a/Source/LinqToDB/DataExtensions.cs
+++ b/Source/LinqToDB/DataExtensions.cs
@@ -1,10 +1,12 @@
-﻿using JetBrains.Annotations;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading;
+using System.Threading.Tasks;
+
+using JetBrains.Annotations;
 
 namespace LinqToDB
 {
@@ -96,9 +98,9 @@ namespace LinqToDB
 		/// <param name="query">The query expression to be compiled.</param>
 		/// <typeparam name="TDc">Type of data context parameter, passed to compiled query.</typeparam>
 		/// <typeparam name="TResult">Query result type.</typeparam>
-		public static Func<TDc, TResult> Compile<TDc, TResult>(
+		public static Func<TDc,TResult> Compile<TDc,TResult>(
 			[NotNull] this IDataContext dataContext,
-			[NotNull] Expression<Func<TDc, TResult>> query)
+			[NotNull] Expression<Func<TDc,TResult>> query)
 			where TDc : IDataContext
 		{
 			return CompiledQuery.Compile(query);
@@ -115,9 +117,9 @@ namespace LinqToDB
 		/// <typeparam name="TDc">Type of data context parameter, passed to compiled query.</typeparam>
 		/// <typeparam name="TArg1">Type of parameter for compiled query.</typeparam>
 		/// <typeparam name="TResult">Query result type.</typeparam>
-		public static Func<TDc, TArg1, TResult> Compile<TDc, TArg1, TResult>(
+		public static Func<TDc,TArg1,TResult> Compile<TDc,TArg1,TResult>(
 			[NotNull] this IDataContext dataContext,
-			[NotNull] Expression<Func<TDc, TArg1, TResult>> query)
+			[NotNull] Expression<Func<TDc,TArg1,TResult>> query)
 			where TDc : IDataContext
 		{
 			return CompiledQuery.Compile(query);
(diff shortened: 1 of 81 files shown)


Source/LinqToDB/DataExtensions.cs
```diff
@@ -462,12 +503,12 @@ public static int Delete<T>([NotNull] this IDataContext dataContext, T obj)
 		/// </param>
 		/// <returns>Created table as queryable source.</returns>
 		public static ITable<T> CreateTable<T>([NotNull] this IDataContext dataContext,
-			string          tableName       = null,
-			string          databaseName    = null,
-			string          schemaName      = null,
-			string          statementHeader = null,
-			string          statementFooter = null,
-			DefaultNullable defaultNullable  = DefaultNullable.None)
+			string tableName = null,
```
sdanyliv: Please return table formatting back. This comment is related to whole file.

beercsaba: I saw this issue. my VS made this formatting :(
Do you know any way to set the tabular formatting in VS?

sdanyliv: We just undo VS fromatting. And trying to disable, formatting on paste, after semicolon.
Igor made this formatting by hands to better read and understand code.

sdanyliv: Before commit we ensure that we do not change non-related parts.

Source/LinqToDB/Linq/QueryRunner.Insert.cs
```diff
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Concurrent;
+﻿using System.Collections.Concurrent;
```
sdanyliv: Looks like this file do not need changes

Source/LinqToDB/Linq/QueryRunner.InsertOrReplace.cs
```diff
@@ -14,18 +13,23 @@ static partial class QueryRunner
...
-			static readonly ConcurrentDictionary<object,Query<int>> _queryChache = new ConcurrentDictionary<object,Query<int>>();
+			static readonly ConcurrentDictionary<object, Query<int>> _queryCache = new ConcurrentDictionary<object, Query<int>>();
 
-			static Query<int> CreateQuery(IDataContext dataContext)
+			static Query<int> CreateQuery(IDataContext dataContext, string tableName = null, string databaseName = null, string schemaName = null)
 			{
 				var fieldDic = new Dictionary<SqlField,ParameterAccessor>();
 				var sqlTable = new SqlTable<T>(dataContext.MappingSchema);
+
+				if (tableName		!= null)	sqlTable.PhysicalName	= tableName;
```
sdanyliv: Fromatting rules, are: line starts from Tab, then formatting by spaces. You can see online that formatting is broken.

sdanyliv:
If you doing rollbacks by hands, do’t do that i will fix everything tomorrow.
Next time be careful with such kind of commits.
Thanks!

Source/LinqToDB/Linq/QueryRunner.Insert.cs
beercsaba: i've fixed a typo (in these QueryRunner.* classes): _queryChache should be spelled queryCache

beercsaba:
Yes, i'm doing it by hands.  I don't want to break your rules.
There are some small differences remained, but maybe (i hope) they are acceptable.

imho: 
from some point of view It would be better to wrap this Collection<Entity ->  (tableName, schemaName, databaseName) triplet> mapping into a CustomNaming class and make it a field of the DataContext class, which can be set in constructor. This would restrict to one-to-one mapping, but make method calls shorter (no need of these 3 parameters), so code will be easier to read. 
What do you think?
I have no time for working on this now and the next 2 months, so if you agree with my opinion, i could work on that later.

Source/LinqToDB/DataExtensions.cs
```diff
@@ -49,10 +47,10 @@ public static ITable<T> GetTable<T>(this IDataContext dataContext)
 		[LinqTunnel]
 		[Pure]
 		public static ITable<T> GetTable<T>(
-			this IDataContext dataContext,
-			object instance,
-			[NotNull] MethodInfo methodInfo,
-			[NotNull] params object[] parameters)
+						this IDataContext	dataContext,
```
sdanyliv: Undo formatting. Leave changed only your methods.

beercsaba: done

Source/LinqToDB/DataExtensions.cs
```diff
@@ -98,9 +96,9 @@ public static ITable<T> GetTable<T>(this IDataContext dataContext)
 		/// <param name="query">The query expression to be compiled.</param>
 		/// <typeparam name="TDc">Type of data context parameter, passed to compiled query.</typeparam>
 		/// <typeparam name="TResult">Query result type.</typeparam>
-		public static Func<TDc,TResult> Compile<TDc,TResult>(
+		public static Func<TDc, TResult> Compile<TDc, TResult>(
```
sdanyliv: Spaces also revert, its’ specific Igor’s style, no spaces between generic types.
For me is better to revert whole file and just copy paste your changes from previous version.

beercsaba:
I think i'm done with formatting, which takes almost as much time as the implementation.
i think i've fixed some formatting inconsistencies in these files. If you do not agree, i could revert these fixes.
i've renamed field _queryChache to _queryCache in there QueryRunner.... classes, i think this is a typo. If you do not agree, i could revert these too.
@sdanyliv what do you think?

sdanyliv:
@beercsaba, i'm picking your PR and will made corrections. Thanks for your help.

sdanyliv:
I have opened new PR #1045.

PR #1054, opened 2018-03-25
fix execute cmd outside of script folder

croban:
According to documentation:
[https://github.com/linq2db/linq2db/blob/538a1f2a2c632f3a56206559d1f5dad5a189c284/CONTRIBUTING.md#building](url)
Two batch files can be run to build the project but if Build.cmd or Compiled.cmd are executed not from within script folder, there is error
```
C:\Git\linq2db>.\Source\LinqToDB\Compile.cmd
MSBUILD : error MSB1009: Project file does not exist.
Switch: linq2db.sln
```


igor-tkachev:
Should we just close this PR?

PR #1055, opened 2018-03-25
Fix execute cmd outside of script folder (done right)

croban:
According to documentation:
[https://github.com/linq2db/linq2db/blob/538a1f2a2c632f3a56206559d1f5dad5a189c284/CONTRIBUTING.md#building](url)
Two batch files can be run to build the project but if Build.cmd or Compiled.cmd are executed not from within script folder, there is error
```
C:\Git\linq2db>.\Source\LinqToDB\Compile.cmd
MSBUILD : error MSB1009: Project file does not exist.
Switch: linq2db.sln
```


sdanyliv:
Looks like you have closed wrong PR. And you do not need to close them. We have to merge first.

croban:
it is late :( and i have deleted the wrong pr, sorry, 
I have restore it again again...

sdanyliv:
No problem. Reopened.

MaceWindu:
could I propose to fix it differently?
instead of adding environment var, just add `cd /d "%~dp0"` as a first line?

PR #1056, opened 2018-03-26
Proposal for Fix GetCteTable

croban:
If following method is called:
```
return this.GetCte<CmsDictionary>(employeeHierarchy =>
			{
				return
				(
					from f in query
					select f
				).Concat( 
					from f in query
					join x in employeeHierarchy on f.Id equals x.Id
					select f
				);
			});
```

there is exception because 
SqlCteTable(mapping, Cte) calls base constructor of SqlTable, inside of base constructor body is referenced property "Name" which is overriden inside upper class (SqlCteTable)
Name property returns Name of CTE which is initializes after base init is finished.

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -289,7 +289,8 @@ namespace LinqToDB.SqlProvider
 				AppendIndent();
 				StringBuilder
 					.Append(cte.Name)
-					.Append(" (");
+					.Append(" ");
+					//.Append(" (");
 
 				var firstField = true;
 				foreach (var field in cte.Fields.Values)
@@ -300,7 +301,7 @@ namespace LinqToDB.SqlProvider
 					StringBuilder.Append(Convert(field.PhysicalName, ConvertType.NameToQueryField));
 				}
 
-				StringBuilder.AppendLine(")");
+				//StringBuilder.AppendLine(")");
 				Indent--;
 				AppendIndent();
 				StringBuilder.AppendLine("AS");
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -287,10 +287,9 @@ namespace LinqToDB.SqlProvider
 				first = false;
 
 				AppendIndent();
-				StringBuilder
-					.Append(cte.Name)
-					.Append(" ");
-					//.Append(" (");
+				StringBuilder.Append(cte.Name);
+				if (cte.Fields.Count > 0) 
+					StringBuilder.Append(" (");
 
 				var firstField = true;
 				foreach (var field in cte.Fields.Values)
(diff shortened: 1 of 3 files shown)


Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -289,7 +289,8 @@ protected virtual void BuildWithClause(SqlWithClause with)
 				AppendIndent();
 				StringBuilder
 					.Append(cte.Name)
-					.Append(" (");
+					.Append(" ");
```
sdanyliv: Why you have removed brackets?

Source/LinqToDB/SqlQuery/SqlTable.cs
```diff
@@ -58,7 +58,7 @@ public SqlTable()
 
 		#region Init from type
 
-		public SqlTable([JetBrains.Annotations.NotNull] MappingSchema mappingSchema, Type objectType) : this()
+		public SqlTable([JetBrains.Annotations.NotNull] MappingSchema mappingSchema, Type objectType, string physicalName=null) : this()
```
sdanyliv: Use our formatting guidelines. There should be spaces.

Source/LinqToDB/SqlQuery/SqlTable.cs
```diff
@@ -68,7 +68,7 @@ public SqlTable([JetBrains.Annotations.NotNull] MappingSchema mappingSchema, Typ
 			Schema       = ed.SchemaName;
 			Name         = ed.TableName;
 			ObjectType   = objectType;
-			PhysicalName = Name;
+			PhysicalName = physicalName??Name;
```
sdanyliv: Also missed spaces.

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
croban: With a brackets I get following SQL exception:
System.Data.SqlClient.SqlException : Incorrect syntax near ')'. 
executed sql is:
WITH
	employeeHierarchy  ()  <-- brackets
AS
(
	SELECT
		[t1].[pk],
		[t1].[id],
		[t1].[parent],
		[t1].[key] as [key1]
	FROM
		[AmnestyAt]..[cmsDictionary] [t1]
)

SELECT....

It can be that this fix collided with some other "WITH" statement that needs brackets.

Source/LinqToDB/SqlQuery/SqlTable.cs
croban: i don't know why normally resharper do that for me...

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
sdanyliv: There should be field names. Not ready yet (

croban: okay, I have added check if cte.fields are empty

MaceWindu:
@igor-tkachev FYI

PR #1065, opened 2018-04-01
Inheritance fixes

jogibear9988:


jogibear9988:
@sdanyliv @MaceWindu also done, if anyone can review?

jogibear9988:
this one? ;-)

Tests/Model/ParentChild.cs
```diff
@@ -507,7 +507,6 @@ public class ParentInheritance12 : ParentInheritanceBase2
 	#region Inheritance3
 
 	[Table(Name="Parent")]
-	[InheritanceMapping(Code = null, Type = typeof(ParentInheritanceBase3))]
```
sdanyliv: Why you have removed this mapping from test?

jogibear9988:
cause it‘s invalid!
base class is abstract, and so no instance of it can be created!
we have to remove the abstract or the mapping!

Source/LinqToDB/Expressions/ConvertFromDataReaderExpression.cs
```diff
@@ -151,7 +151,22 @@ public object GetValue(IDataReader dataReader)
...
 				}
 
-				return func(dataReader);
+				try
+				{
+					return func(dataReader);
+				}
+				catch (FormatException ex)
+				{
+					throw new FormatException("FormatException on Column: " + dataReader.GetName(_columnIndex), ex);
```
MaceWindu: Why do we need 3 cases here? I think one rethrow is enough

jogibear9988: I wanted to keep the original exception type!
What type of exception should I throw?

MaceWindu: Well, I don't think it makes sense to mimic original exception. We just can throw LinqToDbException like "mapping of column {name} value failed, see inner exception for details"

PR #1066, opened 2018-04-01
Mapping fixes - fixes #1005

### LINKED ISSUE #1005: Component Mapping Broken!
Look in the Tests "SelectTests.SelectComplex1 - 3", the Middle and Second Name are always "null". They are not selected from the database

jogibear9988:


Source/LinqToDB/Metadata/MetadataReader.cs
```diff
@@ -20,10 +23,16 @@ public class MetadataReader : IMetadataReader
...
+			_readers = new List<IMetadataReader>(readers) ?? throw new ArgumentNullException(nameof(readers));
 		}
 
-		readonly IMetadataReader[] _readers;
+		IList<IMetadataReader> _readers;
+
+		internal void AddReader(IMetadataReader reader)
+		{
+			// creation of new list is cheaper than lock on each method call
+			_readers = new[] { reader }.Concat(_readers).ToList();
```
sdanyliv: So, why not Array here?

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -130,21 +130,27 @@ void Init(MappingSchema mappingSchema)
 					continue;
 				}
 
-				var ca = mappingSchema.GetAttribute<ColumnAttribute>(TypeAccessor.Type, member.MemberInfo, attr => attr.Configuration);
+				var cas = mappingSchema.GetAttributes<ColumnAttribute>(TypeAccessor.Type, member.MemberInfo, attr => attr.Configuration);
 
-				if (ca != null)
+				if (cas != null && cas.Length > 0)
```
sdanyliv: Just for clarification, why several column attributes are needed?

Source/LinqToDB/Metadata/MetadataReader.cs
jogibear9988: @MaceWindu can you look? or should I change?

Source/LinqToDB/Mapping/EntityDescriptor.cs
jogibear9988: look here: https://github.com/linq2db/linq2db/blob/b02ddd52cb496c7a1252fd49befa57d92f3eb75b/Tests/Model/ComplexPerson.cs#L11

jogibear9988: The tests SelectComplex1-3 never returned the correct db value

jogibear9988:
@sdanyliv 
I now see my error.... The Column Attribute is on the class not the column.

The issue is, for ComplexTest3, the FluentMetadatareader does return the ColumnAttribute not for the class it returns them for the property...

I will fix that...

jogibear9988:
Atm I've no clue how to fix. Should a mapping for Complex Properties be on the class or the Property?

Source/LinqToDB/Metadata/MetadataReader.cs
jogibear9988: https://stackoverflow.com/a/16323412

sdanyliv:
Never use this functionality, i have to look at this closer.

jogibear9988:
I've now fixed it (hopefully). I now Add FluentMappings (if they specify SubProperties) on the class instead of the Property (as done with attribute mappings)

Source/LinqToDB/Metadata/MetadataReader.cs
MaceWindu: Yes, I forgot to change it back to array... no list needed here

jogibear9988:
@MaceWindu @sdanyliv I've fixed the ToList into ToArray. Think it's done, can anyone approve?

jogibear9988:
@MaceWindu this one? ;-)

PR #1068, opened 2018-04-01
Fix #792 for sybase and some other providers

### LINKED ISSUE #792: Linq2db execute Stored procedure when I connect to sybase ASE
Using linq2db with linqpad. Over 150 stored procedure on the database where executed when I tried to connect to the first time and it corrupted our data. 

When we took out the execute permission for stored procedure and tried to reconnect we got error 
10330 You do not have the permission to execute this command. We got 3 error 2812 Store procedure % not found . Specify owner object name.

It seems that when we try to connect to our envireonment linq2db try to execute store proc.

MaceWindu:
PR content:
- Adds tests for #792 for all providers
- Fix DB2/Oracle provider issue when procedures not loaded without tables
- Adds guard exception for Sybase/MySQL for to avoid database corruption (#792)

Following providers not tested as we don't have procedure schema load for them yet (nobody asked?):
- Informix
- PostgreSQL

InvalidOperationException from DB2/MSSQL provider inside of transaction is not catched and rethrown. User just need to remove transaction if he tried to read schema from it.


Source/LinqToDB/DataProvider/DB2/DB2LUWSchemaProvider.cs
```diff
@@ -49,24 +49,30 @@ protected override List<TableInfo> GetTables(DataConnection dataConnection)
 				where
 					new[] { "TABLE", "VIEW" }.Contains(t.Field<string>("TABLE_TYPE"))
 				let catalog = dataConnection.Connection.Database
-				let schema  = t.Field<string>("TABLE_SCHEMA")
-				let name    = t.Field<string>("TABLE_NAME")
-				let system  = t.Field<string>("TABLE_TYPE") == "SYSTEM TABLE"
-				where IncludedSchemas.Count != 0 || ExcludedSchemas.Count != 0 || schema == CurrenSchema
+				let schema = t.Field<string>("TABLE_SCHEMA")
```
sdanyliv: Loosing formatting )

MaceWindu: Will fix...

sdanyliv: There is also same formatting issue in files below.

MaceWindu:
This PR is ready, but I would like to ask someone to run those new tests against dbhost VM as it have other versions of rdbms and could uncover some new issues

PR #1070, opened 2018-04-01
fixes Problems with Insert, Update and Delete when using Inheritance - fixes #1017

### LINKED ISSUE #1017: Problem with Update when using inheritance
If you look at my Demo Data Model in Pull Req 

https://github.com/linq2db/linq2db/blob/42888f042764cb5f2f1aa141b43313e093b4f4eb/Tests/Linq/UserTests/Issue994Tests.cs

If you will do this:

		using (var db = new TestDataConnection())
		{
			var dog = db.GetTable<Dog>().First();
			db.Update(dog);
			db.Update((Animal)dog);
		}

the first Update will work, the second one not!

jogibear9988:
…(fixes #1017)


sdanyliv:
No unit tests?

jogibear9988:
I have some I'm my huge commit https://github.com/linq2db/linq2db/pull/1002/

But they also depend on other Fixes, so I will move them when the rest is fixed

PR #1073, opened 2018-04-02
Dynamic columns. Fix #744, fix #507

### LINKED ISSUE #507: Mapping of dynamic columns

### LINKED ISSUE #744: Dynamic columns

MaceWindu:
This is basically PR #964, updated to latest master.
Still need to fix some unstable tests, but in general looks like everything works


MaceWindu:
This is ready. Only changes from original PR:
- fixed regression with UpdateBuilder
- fixed new tests

PR #1077, opened 2018-04-02
regexp support [WIP] [DRAFT]

jogibear9988:


jogibear9988:
That was, or is my plan! It‘s not yet finished. I‘m looking that it‘s workin for each Database in the same way (if it‘s possible)

MaceWindu:
This feature needs wiki documentation article with per-provider support information:
- is it supported or not
- what server functionality is behind it
- other limitations or provider-specific notes.

Also I don't think it should support Regex.IsMatch, because server behavior, especialy for some providers could have nothing in common with .net regex behavior. It it better to have only explicit Sql.RegExp extension.

Also we shouldn't use 3rd party (SQLNET) functionality for basic API. If anyone wants to have support for this library, it could be done as a separate class with extensions or even separate extensions project.

igor-tkachev:
If you need to convert any .net method/property to SQL, please see how we convert hundreds of standard functions [here](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/Expressions.cs).

You can also add method to SQL mapping right from your application 

```c#
LinqToDB.Linq.Expressions.MapMember((string s1, string s2) => Regex.IsMatch(s1, s2), n => Sql.RegEx(s1, s2));
```

jogibear9988:
I'm thinking of creating a addon dll cause, sql server and sqlite also need a add-on for regex support!

what do you think about this?

But a Register Method for the "MapMeber" call would be needed. Could we create a official syntax for this? I would use something like "Linq2Db.RegExAddon.Register()"

MaceWindu:
@jogibear9988 FYI, take a look at FTS extensions implementation #1649 
If you are still interested in implementing this feature, I think it is a proper way to do it - implement provider-specific extensions in Sql.Ext.DBNAME() namespace

jods4:
@MaceWindu not sure if here's the best place to discuss this but... 

>  I think it is a proper way to do it - implement provider-specific extensions in Sql.Ext.DBNAME() namespace

Is it the best way?
It's clean but usage is quite verbose.

In the middle of a large complex query, best readability is achieved with short, simple function names. 
`Sql.Ext.SqlServer().` is 20 chars, not short and to the point (even less so when function name is `ContainsPropertyWithLanguage`) and it has no intrinsic meaning in the context of my query. 
Now imagine I'm using Sql Server extensions four or five times in the same query... it adds up quickly and compares badly to a raw SQL string.

I think many projects only use a single DB provider, so having repeated `.SqlServer()` all over the code is just noise for a mainstream case.

Some ideas:

1. Why not put the extensions methods directly `ISqlExtension` but inside a `SqlServer` namespace?
So if I do: 
``` 
```
Then `Sql.Ext.Contains()` lights up. That simplifies the queries a good bit.

2. Put the extensions on a different static class. `SqlExt.Contains()` (no dot -- or find a better name).
Again this would be in a `SqlServer` namespace so it only lights up if you have a matching using.
If you use multiple providers in the same file, you can alias classes with
`using MsSql = LinqToDB.DataProvider.SqlServer.SqlExt`.

On cool thing with real static members is that you can do a static using on the class.
```csharp
using static LinqToDB.DataProvider.SqlServer.SqlExt;

var query = from u in db.Users 
            where Contains(u.Address, "Virginia")
            select u;
```
This allows direct function calls. It is very short and sweet, like the real SQL code.

MaceWindu:
Closing stale PRs. Reopen new one if you still plan to work on this feature

PR #1081, opened 2018-04-04
possibility to modify entityDescriptor

jogibear9988:


llehn:
It would be awesome to have that!

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -27,28 +27,61 @@ public EntityDescriptor(MappingSchema mappingSchema, Type type)
 
 			Init(mappingSchema);
 			InitInheritanceMapping(mappingSchema);
+
+			var cb = mappingSchema.EntityDescriptorCreatedCallback;
+			if (cb != null)
+			{
+				mappingSchema.EntityDescriptorCreatedCallback?.Invoke(mappingSchema, this);
```
sdanyliv: Move that functionality to place where entity descriptor is instantiated.

sdanyliv: Remove if ()
Just mappingSchema.EntityDescriptorCreatedCallback?.Invoke(mappingSchema, this) should be be enough.

PR #1082, opened 2018-04-05
Add new options in T4 templates.

hardhub:
Update the following templates:

**T4Model.ttinclude**
```C#  
  UsingsBlankLines = 1; // Count of blank lines after "usings" block.
  DefaultAccessModifier = AccessModifier.Public; // Desired access modifier for each class.
```

**DataModel.ttinclude**
```C#
  ParameterlessConstructor = true; // Generate or not parameterless constructor of database context class.
```
**LinqToDB.ttinclude**

Regex to extract `<table>` and `<target>` to use in column name
Example: `@"FK_(?<table>.+)_(?<target>.+)Id";`
```C#
string ReferenceMainPattern = null; // Default
```
Regex to extract <target> and <column> to use in column name
Example: `@"FK_(?<target>.+)_(?<column>.+)Id_BackReference";`
Instead of `IEnumerable<Comment> FKCommentAuthorIds` it will generate `IEnumerable<Comment> AuthorComments`
```C#
string ReferenceManyPattern = null; // Default
```
**LinqToDB.SqlServer.ttinclude**

Add System.Linq namespace to SqlServer script because it is required by FreeText generated code.

### DIFF
--- a/Source/LinqToDB.Templates/DataModel.ttinclude
+++ b/Source/LinqToDB.Templates/DataModel.ttinclude
@@ -35,6 +35,8 @@ bool     GenerateAssociationExtensions = false;
 bool     ReplaceSimilarTables          = true;
 bool     IncludeDefaultSchema          = true;
 
+bool     ParameterlessConstructor     = true;
+
 Class    DataContextObject;
 
 bool PluralizeClassNames                 = false;
--- a/Source/LinqToDB.Templates/LinqToDB.SqlServer.ttinclude
+++ b/Source/LinqToDB.Templates/LinqToDB.SqlServer.ttinclude
@@ -16,8 +16,9 @@ void DoGenerateSqlServerFreeText()
 {
 	if (!GenerateSqlServerFreeText)
 		return;
-
+	
 	Model.Usings.Add("System.Collections.Generic");
+	Model.Usings.Add("System.Linq");
 	Model.Usings.Add("System.Linq.Expressions");
 	Model.Usings.Add("System.Reflection");
 	Model.Usings.Add("LinqToDB");
--- a/Source/LinqToDB.Templates/LinqToDB.ttinclude
+++ b/Source/LinqToDB.Templates/LinqToDB.ttinclude
@@ -1,8 +1,9 @@
-<#@ assembly name="System.Data"        #>
-<#@ import namespace="System.Data"     #>
-<#@ import namespace="LinqToDB.Data"   #>
-<#@ import namespace="System.Text" #>
-<#@ include file="DataModel.ttinclude" #>
+<#@ assembly name="System.Data"                       #>
+<#@ import namespace="System.Data"                    #>
+<#@ import namespace="LinqToDB.Data"                  #>
+<#@ import namespace="System.Text"                    #>
+<#@ import namespace="System.Text.RegularExpressions" #>
+<#@ include file="DataModel.ttinclude"                #>
 <#
 	if (BaseDataContextClass == null)
 		BaseDataContextClass = "LinqToDB.Data.DataConnection";
@@ -24,19 +25,25 @@ bool?  GenerateScaleProperty     = null;
 bool   GenerateDbTypes           = false;
 bool   GenerateSchemaAsType      = false;
 bool   GenerateViews             = true;
+string ReferenceMainPattern		 = null;
+string ReferenceManyPattern		 = null;
 string SchemaNameSuffix          = "Schema";
 string SchemaDataContextTypeName = "DataContext";
 
 Dictionary<string,string> SchemaNameMapping = new Dictionary<string,string>();
 
-Func<string,string,IEnumerable<Method>> GetConstructors =  (conf, name) => GetConstructorsImpl(conf, name);
+Func<bool,string,string,IEnumerable<Method>> GetConstructors =  (parameterless, conf, name) => GetConstructorsImpl(parameterless, conf, name);
 
-static IEnumerable<Method> GetConstructorsImpl(string defaultConfiguration, string name)
+static IEnumerable<Method> GetConstructorsImpl(bool parameterlessConstructor, string defaultConfiguration, string name)
 {
-	if (defaultConfiguration == null)
-		yield return new Method(null, name);
-	else
-		yield return new Method(null, name) { AfterSignature = { ": base(\"" + defaultConfiguration + "\")" } };
+	if (parameterlessConstructor == true)
+    {
+		if (defaultConfiguration == null)
+			yield return new Method(null, name);
+		else
+			yield return new Method(null, name) { AfterSignature = { ": base(\"" + defaultConfiguration + "\")" } };
+    }
+
 	yield return new Method(null, name, new[] { "string configuration" }) { AfterSignature = { ": base(configuration)" } };
 }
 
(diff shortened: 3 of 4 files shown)


sdanyliv:
`UsingsBlankLines` - i don't understand for which purpose this was introduced. Code is not passed code review? :) Next is spaces between parameters? I've exaggerate problem, but it's out of scope of this project.  

`ParameterlessConstructor` - why just do not disable constructors generation? `GenerateConstructors = false`, and create constructors by yourself in other file?

`ReferenceMainPattern`, `ReferenceManyPattern` - bad idea.
Introduce one callback GetAssociation and put all needed data in callback. In this case you can process/filter out references and set all needed properties.

hardhub:
> Code is not passed code review? :) 
Exactly, automatic style checks )))

> ParameterlessConstructor - why just do not disable constructors generation? GenerateConstructors = false, and create constructors by yourself in other file?
The main issue if something is changed in next release I will have to manually copy this code in my custom constructors. In case of breaking change I will just change my code (after getting compilation errors). In case of my constructors it will most likely will say nothing but it will not work as designed.  
Also I can suggest to set access modifier for constructors. It will allow to fully control instantiation but with all native constructors generated.

>ReferenceMainPattern, ReferenceManyPattern - bad idea.
>Introduce one callback GetAssociation and put all needed data in callback. In this case you can >process/filter out references and set all needed properties.
Maybe bad idea.. Let's guess we need to change it to GetAssociation callback. Is it implemented behavior and I need to provide the callback function? Or I need to implement this behavior? I never see it before...

Is DefaultAccessModifier OK?
Maybe we have to split access modifier of each element (model, constructiors, etc.)

Let's assume we will finally solve all questions.
Should it be different pull request? Or will you pick something from this one.
I do not have a lot experience in open-source projects. How is it usually going at Linq2db?

sdanyliv:
>Exactly, automatic style checks )))

You should disable style check from files that contain `<auto-generated>` header. It's should be standard behavior for style check tools.

>The main issue if something is changed in next release I will have to manually copy this code in my custom constructors. In case of breaking change I will just change my code (after getting compilation errors). In case of my constructors it will most likely will say nothing but it will not work as designed.
Also I can suggest to set access modifier for constructors. It will allow to fully control instantiation but with all native constructors generated.

I did not get idea. If i have such problems i'm just disabling constructor generations. Class is partial, so i can define everything that i need in other file.

>Maybe bad idea.. Let's guess we need to change it to GetAssociation callback. Is it implemented behavior and I need to provide the callback function? Or I need to implement this behavior? I never see it before...

It will be something new. What i don't like in your approach - that it is very limited in functionality. In my variant - you have full C# power.

>Is DefaultAccessModifier OK?
Maybe we have to split access modifier of each element (model, constructiors, etc.)

For me it's superfluous also. You can enumerate tables before generation and set everything that you need for members even additional attributes.

hardhub:
> I did not get idea. If i have such problems i'm just disabling constructor generations. Class is partial, so i can define everything that i need in other file.

I believe it is not the same. In my case I need to control only my part of instantiation (my own data used only in partial or derived class). In your case I have to control how context is created at all. So implementing my own constructor in theory (especially in case of constantly growing framework) can make wrong initialization of context (in other words not my code should be responsible for basic initialization of context).

>It will be something new. What i don't like in your approach - that it is very limited in functionality. In my variant - you have full C# power.

I agree your idea is better from point of view of flexibility... and I solved only my specific needs (but maybe somebody else needs that as well). So I am ready to think how it is better to implement considering your idea. But for now I see one issue. Callback means that I will have access only to one association (currently processing). But to generate correct names I need to have whole collection of associations (considering duplicates). GetAssociation becomes GetAssociations... So I would to know what you think about the issue before I do any code work on it.

>For me it's superfluous also. You can enumerate tables before generation and set everything that you need for members even additional attributes.

2 questions...
1) Will it be also new feature or I can enumerate it in current version?
2) What is about access modifier for context class itself.. I think it is not something related to table enumeration. But as you can see it in my PR a context is generated with specified default access modifier (in my project it must be internal as well as models).

>Just commit new changes in your branch. And DO NOT do that from your master branch. For each PR create new branch in your fork.

OK, I will re-fork repository and create new branch.

sdanyliv:
>I believe it is not the same. In my case I need to control only my part of instantiation (my own data used only in partial or derived class). In your case I have to control how context is created at all. So implementing my own constructor in theory (especially in case of constantly growing framework) can make wrong initialization of context (in other words not my code should be responsible for basic initialization of context).

Still can not understand your point. Small sample will be great. You can create all needed constructors and hide dangerous methods in other file that placed near generated in the same library.

>But to generate correct names I need to have whole collection of associations (considering duplicates). GetAssociation becomes GetAssociations... So I would to know what you think about the issue before I do any code work on it.

User have access to global variable `Tables`, he can analyse duplicates.

>1. Will it be also new feature or I can enumerate it in current version?

It is already present possibility [Global variables](https://github.com/linq2db/linq2db/tree/master/Source/LinqToDB.Templates#useful-members-and-data-structures)

> 2. What is about access modifier for context class itself.. I think it is not something related to table enumeration. But as you can see it in my PR a context is generated with specified default access modifier (in my project it must be internal as well as models).

In T4 template, table is a class that will be generated [here](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB.Templates/DataModel.ttinclude#L670).

hardhub:
> Still can not understand your point. Small sample will be great.

Imagine that the framework is evolving. The developers of framework are responsible for the functionality of the basic constructors. If I create my own constructors (duplicating the behavior of original), then later such constructors will lose relevance (you can implement specific initialization code). Therefore for today I would like to exclude the constructor without parameters or it will be good to have ability to set access modifier of constructors. I need original default constructors to avoid custom initialization of basic functionality. And that should look like:
```C#
internal partial class MyContext
{
      public MyContext(MyOptions options)
           : this(options.GetConfiguration())
      {
             // Do MyOptions specific initialization.
      }
}
```

I should not take care about implementation of _this(string configuration)_ constructor. It is framework constructor and MUST BE generated. But I want to avoid parameterless constructor because in that case I cannot demand arguments required for construction of a context instance.

Maybe best choice is to create something like Tables but for Constructors. So I will be able to enumerate all and make each one for example _private_ (so it is generated, but available only inside context itself - for other constructors).

> User have access to global variable Tables, he can analyse duplicates.

OK, I will check.

> In T4 template, table is a class that will be generated here.

I think it is not the answer for my question. I need to change access modifer for context class and table classes. Tables are available in in Tables global variable. At least it is what you said. But what is about context class itself? How can I change access modifier for context?

sdanyliv:
About constructors. Create base class `BaseMyDataConnection` with new super constructor. Set this class in T4 template as base class for your DataConnection.
You should have access to everything, every class in model.

hardhub:
>About constructors. Create base class BaseMyDataConnection with new super constructor. Set this class in >T4 template as base class for your DataConnection.
>You should have access to everything, every class in model.
I told about "no way" to limit access to native constructors. You suggested to omit constructor generation at all. And then I wrote my previous post. So I am not sure you understand me correctly.

sdanyliv:
Code for constructors is ok. My bad.

igor-tkachev:
If you do not like default constructor implementation, you can disable it by setting GenerateConstructors to false, and then implement your own version in partial class.

MaceWindu:
Could we close it in favor of separate PRs per-feature?

MaceWindu:
Closing stale PRs. If anything rised here is still relevant, please open separate issue for each change. I plan to work on new scaffolding framework soon, so it will be addressed here.

PR #1087, opened 2018-04-07
fix 1057

jogibear9988:


jogibear9988:
@sdanyliv sorry, I accidentally committed to master.

it's possible to block master for me?

PR #1090, opened 2018-04-08
PostgreSQL 9.5 UPSERT support (continuation of #948)

MaceWindu:
Original PR: #948
Remaining known issues:
- [x] need to add table alias to "INSERT INTO table AS alias" part
- [x] for some reason tests try to load value from non-existing sequences


MaceWindu:
This is ready

PR #1091, opened 2018-04-09
PostgreSQL native bulk copy

MaceWindu:
Continuation of #935 

Remaining tasks:
- [x] make types test work (add non-null values and fix all discovered issues)
- [x] update BulkCopy wiki article with PostgreSQL changes: https://github.com/linq2db/linq2db/wiki/Bulk-Copy
- [ ] update BulkCopy wiki article with documentation of more BulkOptions properties
- [x] retest all databases as some changes done to default mappings
- [x] fix code formatting :-/


Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs
```diff
@@ -291,13 +393,149 @@ public override CommandBehavior GetCommandBehavior(CommandBehavior commandBehavi
...
+					dbType = "double precision";
+					break;
+				case "varchar":
+					dbType = "character varying";
+					break;
+				case "varbit":
+					dbType = "bit varying";
+					break;
+			}
+
```
sdanyliv: Not sure but regular expression looks more suitable here, especially with code below.

MaceWindu: I would prefere to not use regexps in library code :)

jogibear9988: I think a regex will be less performant

MaceWindu:
This one is ready for review.

Short overview of changes:
- native bulk copy for postgresql with test for all types (also should work with npgsql 4.0 when it will be released, as it contains breaking changes to this functionality)
- updated AllTypes test table names to reduce number of postgresql-specific test mappings
- added missing and updated existing postgresql types support (different versions of provider have different sets of types and some of them were missing/outdated)
- updated type name generation for some postgresql types to generate proper type name

PR #1093, opened 2018-04-10
Rename all Issue Tests

jogibear9988:
Fix sorting in Visual Studio


jogibear9988:
Why?

Some Tests already Start with „0“ in Name

jogibear9988:
any more suggestions? I don't like the random order of the classes in VS

sdanyliv:
Actually i’m just afraid of loosing easy way to access history. If no one other has obligations, we can do that. Wait for others and resolve conflicts.

jogibear9988:
closing this. 

as it's not needed, and sq @sdanyliv sais, history access is then harder

PR #1098, opened 2018-04-12
Fix incorrect processing order in query visitor

MaceWindu:
Fixes issue for insert from select query when select contains nullable parameter and first call of query is done with null value. It leads to failures on next calls with non-null parameters.

Except insert statement fix, also fixed order for update/insertOrUpdate statements and tried to create tests update/insertOrUpdate, but update case generates bogus sql, and for insertOrUpdate don't see how it is possible to create such query.


MaceWindu:
do not merge yet. looks like Update doesn't like change of visiting order

Issue #1083, opened 2018-04-05
How to use "dynamic queries" with Dynamic columns #964 feature?


drekoo:
Hi,
it seems that 
```c#
var filteringColumn = "FirstName";
var sortingColumn = "LastName";
var filter = "John";

db.GetTable<Person>()
    .Where(p => Sql.Property<object>(p, filteringColumn) == filter)
    .OrderBy(p => Sql.Property<object>(p, sortingColumn))
    .ToList();
```
from current #964 feature implementation is actualy not "fully implemented". "Only constant strings are alowed for member name in Sql.Property expressions.". **Will this be implemented in the future**.

I have entity

```c#
[Table("PRODUCT")]
public partial class Product
{
...
 public double PRICE1 { get; set; }
 public double PRICE2 { get; set; }
 public double PRICE3 { get; set; }
 public double PRICE4 { get; set; }
 public double PRICE5 { get; set; }
...
}

//I want to query with

string whichPriceName = "PRICE" + settings.PriceID.ToString();

from p in dbContext.Db.GetTable<Product>()
                       select new ProductDTO
                       {
                           ID = p.ID,
                           PRICE = Sql.Property<double>(p, whichPriceName )
                       };

//Possible (current) solution to my problem is using expression method in Product Entity

 [ExpressionMethod("WhichPriceExpression")]
        public double WhichPrice(int priceID)
        {
            //local
            switch (priceID)
            {
                case 5:
                    return PRICE5;
                case 4:
                    return PRICE4;
                case 3:
                    return PRICE3;
                case 2:
                    return PRICE2;
                default:
                    return PRICE1;
            }

        }

        public static Expression<Func<Product, int, double>> WhichPriceExpression()
        {
            return (product, index) => index == 5 ?
            product.PRICE5: index == 4 ?
            product.PRICE4: index == 3 ?
            product.PRICE3: index == 2 ?
            product.PRICE2: product.PRICE1;
        }

          from p in dbContext.Db.GetTable<Product>()
                       select new ProductDTO
                       {
                           ID = p.ID,
                           PRICE = p.WhichPrice(settings.PriceID)
                       };

```
The expression method is actually translated to SQL CASE statement which is not nice, if there are a lot of PRICE{X} fields...

**Is technically possible in Linq2DB to have expression method with parameter....**

```c#

 public static Expression<Func<Product, double>> WhichPriceExpression(int priceID)
        {
            switch (priceID)
            {
                case 5:
[...]

PR #1102, opened 2018-04-15
Added possibility to use dynamic columns in queries without required storage.

sdanyliv:
For now dynamic columns are allowed to use in queries even they are not specified in storage or class.
Fixes issue #1083


MaceWindu:
Looks ok, but not all related tests passed

sdanyliv:
Will check. Looks like this feature is a way how to write dynamic SQL if you really need that.

Issue #1107, opened 2018-04-20
Error inserting Object with DateTime into SqlServer when Time is not Set


jogibear9988:
I got a Exception when trying to insert a DateTime Field into SqlServer via BulkCopy when Time of the DateTime Object is not Set.

Linq2Db Create following value '2017-02-02' wich does not work in SQL

PR #1108, opened 2018-04-20
Bugfix for #1107

jogibear9988:


sdanyliv:
I know it is boring, but we need unit tests for that.

jogibear9988:
i‘ll look

jogibear9988:
@sdanyliv I saw, that I accidentally (again) pushed to master instead of my branch ;-(
Should I revert it? (I've added a test for the issue)

jogibear9988:
and one more problem is, the test did not fail on appveyor, but did on my sqlserver db.

maybe it's cause of sqlserver version, or db collation?

sdanyliv:
Let’s fix it here. And merge again. Same branch.
Which exception do you have?

jogibear9988:
it‘s fixed with my changes but without my changes the test does not fail in mssql on appveyor

sdanyliv:
You have to investigate. It can be problem for other users.

sdanyliv:
I’ve missed your previous issue. So everythyng is ok. Thanks.

PR #1122, opened 2018-04-29
Added possibility to handle third party query provider for common extensions.

sdanyliv:
Needs idea verification.


MaceWindu:
I'm not sure that all-or-nothing approach is good. It will mean that anyone who wants to replace one method or group of methods will need to implement all of them.

sdanyliv:
Problem that we have collision with EF. To do not break existing users code we have to do that. Pity but these methods still not standardized.

PR #1127, opened 2018-05-03
bugfix when inheritance is used #1128

jogibear9988:


jogibear9988:
Are we finished with this? For me looks good.

ili:
I think yes. Dmitry would merge this into beta5

PR #1131, opened 2018-05-04
Added OnTraceConnection property for tracing DataContext's connection.

sdanyliv:
Needed for proper tracing of linq2db queries.


sdanyliv:
I hope it will be delivered with 2.0. Nothing critical here.

PR #1136, opened 2018-05-07
bugfix sqlserver bulkkcopy with retry, fixes #1135

### LINKED ISSUE #1135: Bulkcopy does not work cause of retryingdbconnection
Linq2db BulkCopy does not work correctly with the retrypolicy, cause this is creating a RetryingDbCOnnection, and then this cast: https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs#L28
will fail

jogibear9988:


MaceWindu:
Would make sense to review our code to find all such places

PR #1137, opened 2018-05-07
switch unittests to use localtable

jogibear9988:


jogibear9988:
does anyone know why travisCi is failing?

PR #1139, opened 2018-05-09
better logging for sqlserver bulk copy

jogibear9988:


sdanyliv:
@jogibear9988, do not merge yet. Let others to decide.

jogibear9988:
Why what= Why let others decide, or why this change?

jogibear9988:
Why my change? Because i've errors in the Mapping ("caseing") but it worked without BulkCopy. Now if I have an error, I see the Column Names tried

igor-tkachev:
I like the idea to provide more details. The question is why lambda?

jogibear9988:
cause i thought to build the string obly when logging is enabled (micro perf. optimation)

PR #1145, opened 2018-05-11
Release 2.0 - Make SqlExtensions public

RoyC-Retail-Assist:
I use the SqlExtensions class in the DB2 iSeries provider but it needs to be explicitly scoped public.


sdanyliv:
Already done in other commit 92c514d43b80fd68872d30c97ff05dfd7aebe988

PR #1154, opened 2018-05-17
Added SqlProviderFlags.CustomFlags

RoyC-Retail-Assist:
The DB2 iSeries provider has options as to how it handles Guids as they are not available natively in DB2 (handle them as a string or a blob).  This setting needs to be available in the SqlBuilder.  I used a constructor overload to set this value when creating the SqlBuilder in the provider but the LinqService directly instantiates the SqlBuilder therefore in the LinqService the option is not available.

This change simply adds a CustomFlags list into  SqlProviderFlags allowing me to set my custom options which are then applied when the Linq service creates the SqlBuilder.


MaceWindu:
sorry, closed unintentionally. will merge asap

PR #629, opened 2017-04-11
Fix cant load Sap.Data.Hana.dll and Diference between HANA Server and HANA Client

FabioNascimento:
Fix for #239

PR #1157, opened 2018-05-19
Fix load of Sap.Data.Hana provider from GAC. Fix #239

### LINKED ISSUE #239: Hana Linq Query failed - cant load Sap.Data.Hana.dll
Started looking at HANA for a project and downloaded Linq2db. Set up a query string, managed to create a small test Context, opened a connection all fine.

When trying to execute a LINQ query, I get a crash suggesting that 'Sap.Data.Hana' could not be found.

Although my Provider IS 'Sap.Data.Hana' and the Connection class can open a connection, my assembly is actually called 'Sap.Data.Hana.v4.5', straight from the Hana client install.

I guess that something is trying to instantiate a type by looking for the assembly name - either the object type needs to be grabbed from the connection or the assembly name similarly grabbed from the connection or providers table or something.

Apologies, my code is not available to me right now, I'm just hoping that somebody just knows what it is. I may be able to post example later when I pick up my laptop.

MaceWindu:
This is a fix from #629, updated to recent codebase.
Thanks @FabioNascimento to fix it and sorry for it taking so long to merge.

Note that this fix will be applied only to projects that target full .net framework, as it uses API, which will be avalaible only in .net core 2.1: https://github.com/dotnet/corefx/issues/4571
From other side, it doesn't matter, as .net core don't have this issue due to absense of GAC


Source/LinqToDB/DataProvider/SapHana/SapHanaDataProvider.cs
```diff
@@ -46,6 +47,45 @@ protected SapHanaDataProvider(string name, MappingSchema mappingSchema)
...
+			}
+		}
+
+		protected override Type GetConnectionType()
+		{
+			if (_connectionType == null)
+				lock (SyncRoot)
+					if (_connectionType == null)
+					{
+						var db = DbProviderFactories.GetFactory("Sap.Data.Hana").CreateConnection();
```
sdanyliv: Can GetFactory return null?

MaceWindu: It throws exception if it cannot find factory. Will check if it needs changes

MaceWindu:
fixed type load to first try base approach without exception and then use dbproviderfactory

PR #1177, opened 2018-06-06
Improved caching detection for SqlQueryDependentAttribute.

### LINKED ISSUE #973: Is there a way to parameterize Contains query?
Currently linq query with Contains method generates sql operator "in" with inlined parameters.

For example:

```
[Test]
		public void __Linq2Db_InlineParameters_For_Contains()
		{
			var ids = new int[] { 1, 2, 3 };

			using (var db = GetConnection())
			{
				var query = db.Users.Where(x => ids.Contains(x.Id));

				Console.WriteLine(query);
			}
		}
```

generates:

```
SELECT
	[t1].[Id],
	[t1].[Name],
	[t1].[IsActive],
	[t1].[AppointmentId]
FROM
	[Users] [t1]
WHERE
	[t1].[Id] IN (1, 2, 3)
```

Is there a way to generate something like [t1].[Id] IN (@p0, @p1, @p2)?

### Environment details
linq2db version: 1.10.1
Database Server: SqlServer 2016 
Database Provider: SqlServer.2012
Operating system: Windows 10
Framework version: .NET Framework 4.6.2

sdanyliv:
Should fix #973

### DIFF
--- a/Source/LinqToDB/Expressions/InternalExtensions.cs
+++ b/Source/LinqToDB/Expressions/InternalExtensions.cs
@@ -394,10 +394,24 @@ namespace LinqToDB.Expressions
 			var parameters = expr1.Method.GetParameters();
 			for (var i = 0; i < expr1.Arguments.Count; i++)
 			{
-				if (parameters[i].GetCustomAttributes(typeof(SqlQueryDependentAttribute), false).Any())
+#if !NETSTANDARD1_6
+				var dependedAttributes = parameters[i].GetCustomAttributes(typeof(SqlQueryDependentAttribute), false);
+#else
+				var dependedAttributes = parameters[i].GetCustomAttributes(typeof(SqlQueryDependentAttribute), false)
+					.OfType<SqlQueryDependentAttribute>()
+					.ToArray();
+#endif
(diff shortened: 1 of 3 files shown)


Source/LinqToDB/Expressions/InternalExtensions.cs
```diff
@@ -394,10 +394,24 @@ static bool EqualsToX(MethodCallExpression expr1, MethodCallExpression expr2, Eq
 			var parameters = expr1.Method.GetParameters();
 			for (var i = 0; i < expr1.Arguments.Count; i++)
 			{
-				if (parameters[i].GetCustomAttributes(typeof(SqlQueryDependentAttribute), false).Any())
+#if !NETSTANDARD1_6
+				var dependedAttributes = parameters[i].GetCustomAttributes(typeof(SqlQueryDependentAttribute), false);
```
MaceWindu: 1. `depended` -> `dependent`
2. don't we already have cross-platform GetCustomAttributes wrapper? Maybe it makes sense to add it then

sdanyliv: It should be fast place, so i decided to do not play with additional calls.

PR #1184, opened 2018-06-10
PostgreSQL functions schema provider support

### LINKED ISSUE #1162: Failed to generate postgres DataModel with all functions from db schemas
Hello. The RSDN topic http://rsdn.org/forum/prj.rfd/6989413.1 says that i must ask questions here, but i can't see here any button like "Ask question", so i will create the issue.

How i can get generated data model  from postgres db with all stored procedures, i.e. "functions", as they named in postgres? 

I'm using linq2db.Core 1.10.2. Here is code inside t4 template:
```
GetSchemaOptions.ExcludedSchemas = new[] { "Link" }; 
NamespaceName = "MyApi.Persistence.DataModel";
GenerateFindExtensions = false;
LoadPostgreSQLMetadata("localhost", "15432", "db", "postgres", "password");
GenerateModel();
```

After this, i got the file named "PostgreSQL.Core.generated.cs" with all needed tables, but i can't find any function. No one. Nothing. All functions is located in "public" schema and available for reading by user "postgres".

In sources, in file "LinqToDB.Templates\README.md" i can't find an options to enable or disable wrappers generation for stored procedues and other routines. 

I found the issue https://github.com/linq2db/linq2db/issues/792 with glorious advice "...when you create LINQ to DB Connection, uncheck "Include Stored Procedures and Functions" option.". I downloaded and explored sources of LinqPad, found delightful mapping of this option to property "IncludeRoutines" and more to come - mappting of that property to next - "ExcludeRoutines"! But i can't find a link between this "option" and any of options of t4 generator.

So, most of all DB projects is legacy. If one want to use linq2db with them, he must be able to EASY get all DB schema into his code and play with it. Without all these code digging like mine experience.

### LINKED ISSUE #1191: PostgreSQL: T4 generates invalid code if table/view with same name exists in different schemas
I tested the T4 data model generation with the sample PostgreSQL database from here: https://github.com/NorfolkDataSci/adventure-works-postgres

Unfortunately the generated code is invalid because it contains classes with the same name and/or properties in the same class with the same name. 
The result are for example the following errors (just two of many):
> CS0102	The type 'AdventureworksDB' already contains a definition for 'Crs'
> CS0102	The type 'Sr' already contains a definition for 'Modifieddate'

See this gist for the generated model: https://gist.github.com/zer/209c2eaa2077654ff9bfbf9daac36cd2

### Steps to reproduce
Generate data model from the database from here: https://github.com/NorfolkDataSci/adventure-works-postgres

OR 

the short version:
1. Create new Database
2. Create schemas and tables:
`CREATE SCHEMA s1;`
`CREATE SCHEMA s2;`
`CREATE TABLE s1.t1 (c1 int);`
`CREATE TABLE s2.t1 (c1 int);`
3. Generate data model from the newly created database
4. Result should look something like this: https://gist.github.com/zer/209c2eaa2077654ff9bfbf9daac36cd2#file-simple-models-generated-cs

### Environment details
linq2db version: *2.0.0*  
linq2db.t4models: *1.10.0*
Database Server: *PostgreSQL 10.4*  
Database Provider: *Npgsql 3.2.5*  
Framework version: *.NET Core 2.0*

### LINKED ISSUE #1205: T4 generates invalid code for foreign key
I'm trying to create a database model using the T4 generator and got the invalid code for the foreign key.

### Foreign key:
```sql
ALTER TABLE "secret_specification_production" ADD CONSTRAINT "FKsecret_spe188741" FOREIGN KEY ("object_id") REFERENCES "production_object_type" ("id");
```
### Creates this code:

```c#
[Association(ThisKey="ObjectId", OtherKey="Id", CanBeNull=false, Relationship=Relationship.ManyToOne, KeyName="FKsecret_spe188741", BackReferenceName="FKsecretspes")]
		public ProductionObjectType object { get; set; }
```

The reserved word "object" is used for the property name.

### Environment details
linq2db version: *2.1.0*
Database Server: *PostgreSQL 10.4*
Database Provider: *Npgsql 3.2.5.0*

MaceWindu:
Fix #1162, #1205, #1191

What included in this PR:
- functions/procedures schema loading for postgresql
- functions/procedures generation by T4 for postgresql
- aggregate functions generation by T4
- fix #1205 name generation using c# keyword for association properties and procedure parameters
- fix #1191 table and procedures generation with duplicate names)

What is not supported by this PR (feel free to create separate issue):
- postgresql functions with dynamic results like json_to_record/json_to_recordset (we load them in schema but don't generate properly in T4)
- procedures in pgsql 11 - this could work, but not tested as pgsql v11 is not released yet

Important implementation notes

Because prior to v11 postgresql didn't have procedures, it supports some unusual function features:
- void functions. Because we need to be able to call such functions in C# as functions, they will have `object func(...)` signature
- out parameters in functions. Postgresql return then as a single return value: scalar for one parameter and record for multiple out parameters. We don't generate out/ref parameters for functions and map results to scalar value or class in case of multiple out parameters:

One out parameter:
```cs
// one out parameter of text type
string somefunction(...inputs...)
```
Multiple out parameters (two):
```cs
FunctionResult Function(...inputs...)

class FunctionResult
{
    public int? out_param;
    public string another_out_param;
}

// this method is called from generated data context constructor and if you disabled constructors generation, you need to call it from your custom constructor
public void InitMappingSchema()
{
    MappingSchema.SetConvertExpression<object[], FunctionResult>(tuple => new FunctionResult() { out_param= (int?)tuple[0], another_out_param= (string)tuple[1] });
}
```


MaceWindu:
ok, looks like now I get all needed information from pg. now we just need to generate usable model from it.

Source/LinqToDB.Templates/DataModel.ttinclude
```diff
@@ -436,6 +437,17 @@ void LoadMetadata(DataConnection dataConnection)
 			if (col.MemberName == t.TypeName)
 				col.MemberName += "_Column";
 		}
+
+		foreach (var fk in t.ForeignKeys.Values)
+		{
+			if (KeyWords.Contains(fk.MemberName))
+				fk.MemberName = "@" + fk.MemberName;
```
igor-tkachev: ConvertToCompilable on the next line will replace '@' with '_'.

Source/LinqToDB.Templates/DataModel.ttinclude
```diff
@@ -469,6 +481,15 @@ void LoadMetadata(DataConnection dataConnection)
 		}
 	}
 
+	foreach (var proc in Procedures.Values)
+	{
+		foreach (var param in proc.ProcParameters)
+		{
+			if (KeyWords.Contains(param.ParameterName))
+				param.ParameterName = ConvertToCompilable("@" + param.ParameterName, true);
```
igor-tkachev: KeyWord.Contains should be included in ConvertCompatible.

Source/LinqToDB.Templates/DataModel.ttinclude
MaceWindu: No, ConvertToCompilable skips '@' on replace. But I agree that this functionality needs  improvement. Still I would prefer it to be done as separate PR.

MaceWindu:
I've partially fixed duplicate function/class name generation, but it still could be broken in some cases. To fix it properly, we need to update our code-generation AST (`ITree`) to store types as links to ITree elements instead of strings - currently it makes rename operation very fragile.

I would like to work on this later (probably for next or next+1 release)

PR #1199, opened 2018-06-20
Updating netcore assembly reference to nuget reference

ronnyek:
Updated package reference to oracles nuget package, which should be easier to maintain, updated filter on the CreateTests for running tests in netcoreapp2.0 against oracle. 

I was seeing test failures in netcoreapp2.0 before and after this change that seemed to be related to running NUNIT tests as netcore2.0. Otherwise all oracle tests are passing and didn't see new failures as a result to changing this reference.


MaceWindu:
Sorry, don't have time to test it before today's release, so it will go into next one.

MaceWindu:
BTW, you can check my sybase core provider PR for details, maybe it will help you with nuget/tests https://github.com/linq2db/linq2db/pull/1159

MaceWindu:
I'm merging it into separate branch for testing. Don't want to break master before release...

Issue #1202, opened 2018-06-21
NullReferenceException with self generated Expression Tree


Fruchuxs:
I get a NullReferenceException with a self generated Select expression tree. I also generate some Null-Checks but I dont know (after some hours of debugging), which part of my expression tree triggers the exception. Can you handle this case and add a meaningful exception message for better debug purposes?

```
Unhandled Exception: NullReferenceException
  System.NullReferenceException: Object reference not set to an instance of an object.
   at LinqToDB.Linq.Builder.SelectContext.BuildExpression(Expression expression, Int32 level, Boolean enforceServerSide)
   at LinqToDB.Linq.Builder.SelectContext.BuildExpression(Expression expression, Int32 level, Boolean enforceServerSide)
   at LinqToDB.Linq.Builder.ExpressionBuilder.TransformExpression(IBuildContext context, Expression expr, Boolean enforceServerSide)
   at LinqToDB.Linq.Builder.ExpressionBuilder.<>c__DisplayClass72_0.<BuildExpression>b__0(Expression expr)
   at LinqToDB.Expressions.Extensions.Transform(Expression expr, Func`2 func)
   at LinqToDB.Expressions.Extensions.<>c__DisplayClass34_1.<Transform>b__1(MemberBinding b)
   at LinqToDB.Expressions.Extensions.Transform2[T](ICollection`1 source, Func`2 func)
   at LinqToDB.Expressions.Extensions.Transform(Expression expr, Func`2 func)
   at LinqToDB.Expressions.Extensions.<>c__DisplayClass34_1.<Transform>b__1(MemberBinding b)
   at LinqToDB.Expressions.Extensions.Transform2[T](ICollection`1 source, Func`2 func)
   at LinqToDB.Expressions.Extensions.Transform(Expression expr, Func`2 func)
   at LinqToDB.Linq.Builder.SelectContext.BuildExpression(Expression expression, Int32 level, Boolean enforceServerSide)
   at LinqToDB.Linq.Builder.SelectContext.BuildQuery[T](Query`1 query, ParameterExpression queryParameter)
   at LinqToDB.Linq.Builder.FirstSingleBuilder.FirstSingleContext.BuildQuery[T](Query`1 query, ParameterExpression queryParameter)
   at LinqToDB.Linq.Builder.ExpressionBuilder.Build[T]()
   at LinqToDB.Linq.Query`1.CreateQuery(IDataContext dataContext, Expression expr)
[...]

PR #1208, opened 2018-06-26
Implemented Null Propagation optimization.

sdanyliv:
@igor-tkachev, please verify idea.
It may help in situations with #1202, and OData providers. Usually they also add check for null in projection.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -2786,6 +2786,45 @@ namespace LinqToDB.Linq.Builder
 			return sqlExpression;
 		}
 
+		bool IsNullConstant(Expression expr)
+		{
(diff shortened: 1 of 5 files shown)


Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -2786,6 +2786,45 @@ internal ISqlExpression ConvertSearchCondition(IBuildContext context, ISqlExpres
...
+		Expression RemoveNullPropagation(Expression expr)
+		{
+			switch (expr.NodeType)
+			{
+				case ExpressionType.Conditional:
+					var conditional = (ConditionalExpression)expr;
+					if (conditional.Test.NodeType == ExpressionType.NotEqual)
+					{
+						var binary = (BinaryExpression)conditional.Test;
+						if (IsNullConstant(binary.Right))
```
MaceWindu: Looks like it will not handle C-style reverse conditions like `null ==/!= smth`

PR #1233, opened 2018-07-21
Extracting T4 renderer into separate class

ili:
* Renderer is extracted from model
* basic TypeScript renderer is implemented


MaceWindu:
move to 2.4.0?

PR #1239, opened 2018-07-24
[WIP] Fix nullable keys handling in merge-based methods

### LINKED ISSUE #1238: InsertOrUpdate does not handle correctly nullable columns
The SQL `MERGE` statement generated by linq2db for SQL Server does not handle correctly nullable columns and `null` values in the match condition. It may be the case that this also applies to other database engines but I have only tested with SQL Server so far.

The following code is generated for SQL Server:

```SQL
MERGE INTO [dbo].[Data] [t1]
USING (SELECT @dbId1 AS [Id], @name1 AS [Name]) [s] ON
(
	[t1].[Id] = [s].[Id] AND
	[t1].[Name] = [s].[Name]
)
...
```

The above code fails to find a match in case the value of the `Name` column and the `@name1` argument are both `null`.

The correct code should be

```SQL
MERGE INTO [dbo].[Data] [t1]
USING (SELECT @dbId1 AS [Id], @name1 AS [Name]) [s] ON
(
	[t1].[Id] = [s].[Id] AND
	([t1].[Name] = [s].[Name] OR ([t1].[Name] IS NULL AND [s].[Name] IS NULL))
)
...
```

The `IS NULL` check should be added only if the column (in this case `Name`) is nullable.

This issue results in a "duplicate key exception" when the column is a part of a unique index.

### Environment details
linq2db version: linq2db.core 1.10.2
Database Server: Default
Database Provider: Default
Operating system: Windows 10
Framework version: .NET Standard 2.0.3

MaceWindu:
Fix #1238

### DIFF
--- a/Tests/Linq/UserTests/Issue1238Tests.cs
+++ b/Tests/Linq/UserTests/Issue1238Tests.cs
@@ -1,0 +1,166 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+using NUnit.Framework;
+
+namespace Tests.UserTests
+{
+	using LinqToDB;
+	using LinqToDB.Mapping;
+	using Tests.Model;
+
+	[TestFixture]
+	public class Issue1238Tests : TestBase
+	{
+		[Table("InheritanceParent")]
+		public class TestTable
+		{
+			[Column("InheritanceParentId"), PrimaryKey]           public int    Key1;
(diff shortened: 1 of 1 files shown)


MaceWindu:
3 new tests doesn't work on DB2 yet, will try to fix soon

MaceWindu:
One DB2 test fails, but it is unrelated issue and we can fix it later if we will not find good solution right now

PR #1244, opened 2018-07-25
Fix oracle identifiers quotation detection

### LINKED ISSUE #1243: Incorrect sql script if table/column name set with quotes in oracle
Hi.
Seems some problem with oracle sql generation - sql script is invalid if table/column name set with quotes.

Steps.

Some table in Oracle db. Table name and column name sets inside quotation-marks aka case sensitive names.

Script for creating table.

_Sql_

```
CREATE TABLE {some_schema}."TestTable" (
  "FirstField"  NVARCHAR2(50),
  "SecondField" NVARCHAR2(50)
)
```

In way to get easy quoting names creats some custom attrs:

```
 public sealed class TableNameAttribute : TableAttribute
    {
        public TableNameAttribute(string tableName)
        {
            Name = Config.IfQuote? '"' + tableName + '"' : tableName;
        }
    }
```
	
and 
	
```
	 public class ColumnNameAttribute : ColumnAttribute
    {
        public ColumnNameAttribute(string columnName)
        {
            Name = Config.IfQuote? '"' + columnName + '"' : columnName;
        }
    }
```
	

> **Config.IfQuote**   

Just global const (true, false).	

_And next c# model like this_

```
[TableName("JustTable")]
    public class SimpleTable
    {
        [ColumnName("FirstField")]
        public string FirstField { get; set; }
        [ColumnName("SecondField")]
        public string SecondField {get; set; }
    }
```

Simple database context:

```
 public class OracleContext : DataConnection
    {
        public OracleContext() : base("OracleConnection") { }
        public ITable<SimpleTable> Simples => GetTable<SimpleTable>();
    }
```
	

> **OracleConnection**

  Connection from app.config. Its direct connection via Oracle.ManagedDataAccess.

First try.

Set IfQuote to false. It's mean that table and column names without quotes.
 
Starts this code:
```
using (var dbContext = new OracleContext())
            {
                var querry = dbContext.Simples.SqlText;
                var data = dbContext.Simples.ToList();
            }
```
			
In querry variable sqlText:

> SELECT
> 	t1.FirstField,
> 	t1.SecondField
> FROM
> 	JustTable t1

	
This incorrect. When try get data gets **OracleException ORA-00942**.
	
Second try.

Set IfQuote to true. It's mean that table and column names with quotes 

Starts this code:
```
using (var dbContext = new OracleContext())
            {
                var querry = dbContext.Simples.SqlText;
                var data = dbContext.Simples.ToList();
            }
```
In querry variable sqlText:	

> SELECT
> 	t1.""FirstField"" as FirstField,
> 	t1.""SecondField"" as SecondField
> FROM
> 	""JustTable"" t1

This incorrect. Double quoting.

### Environment details
linq2db version: *2.1.0*
Database Server: *Oracle Database 11g*
Database Provider: *Oracle.ManagedDataAccess.Client*
Framework version: *.NET Framework 4.5*

MaceWindu:
Fix #1243

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilder.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilder.cs
@@ -313,8 +313,9 @@ namespace LinqToDB.DataProvider.Oracle
 			// Also we should allow only uppercase letters:
 			// "Nonquoted identifiers are not case sensitive. Oracle interprets them as uppercase"
 			return !IsReserved(name) &&
-				(name[0] >= 'A' && name[0] <= 'Z') &&
+				((OracleTools.DontEscapeLowercaseIdentifiers && name[0] >= 'a' && name[0] <= 'z') || (name[0] >= 'A' && name[0] <= 'Z')) &&
 				name.All(c =>
+					(OracleTools.DontEscapeLowercaseIdentifiers && c >= 'a' && c <= 'z') ||
 					(c >= 'A' && c <= 'Z') ||
 					(c >= '0' && c <= '9') ||
 					c == '$' ||
--- a/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilder.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilder.cs
@@ -310,10 +310,11 @@ namespace LinqToDB.DataProvider.Oracle
 			// https://docs.oracle.com/cd/B28359_01/server.111/b28286/sql_elements008.htm#SQLRF00223
 			// TODO: "Nonquoted identifiers can contain only alphanumeric characters from your database character set"
 			// now we check only for latin letters
+			// Also we should allow only uppercase letters:
+			// "Nonquoted identifiers are not case sensitive. Oracle interprets them as uppercase"
 			return !IsReserved(name) &&
-				((name[0] >= 'a' && name[0] <= 'z') || (name[0] >= 'A' && name[0] <= 'Z')) &&
+				(name[0] >= 'A' && name[0] <= 'Z') &&
 				name.All(c =>
-					(c >= 'a' && c <= 'z') ||
 					(c >= 'A' && c <= 'Z') ||
 					(c >= '0' && c <= '9') ||
 					c == '$' ||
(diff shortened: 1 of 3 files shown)


MaceWindu:
Right now it broke all our tests, because we need to specify identifiers with proper casing for Oracle in many test mappings

MaceWindu:
Ok, we will add 
```cs
OracleTools.DontEscapeLowercaseIdentifiers = true;
```
option to control escaping behavior to not break it for existing users.
To enable proper escaping option must be set to `false`.
Next releases will switch it to false and remove it at some point.

laurentKTR:
Any news on the subject ?
Changing DontEscapeLowercaseIdentifiers from true or false
still put "" around table name which make an oracle exception
ORA-00942: table or view does not exist

sdanyliv:
@laurentKTR, which table name do you have?

laurentKTR:
In fact it's not a table but a view , do it behaves differently ?
i'm getting select id, name from "schema.viewname" tr

sdanyliv:
Sorry, i can not check for now, but looks like you have to separate schema from table name. There are appropriate properties Schema in `TableAttribute` or you can write
```cs
db.GetTable<View>().SchemaName(“schema”).ToList();
```

laurentKTR:
Thanks you very much !!
this object was the only one not to use
[Table(Schema = "xxx", Name = "yyyy")] syntax

No idea why we had written 
[Table(Name = "xxx.yyyy")] syntax

laurentKTR:
Hello
i'm facing new issues with left join queries
Both tables use the Schema and name properties
When doing a standard join 

                    from i in db.GetTable<DataImport>()
                    join m in db.GetTable<DataMapping>() on i.DeclaredProduct equals m.DeclaredProduct
                    where i.Activation != 50 

=> it's working ( i had to set LinqToDB.Common.Configuration.Linq.CompareNullsAsValues = false; to avoid null value checking)

When doing a left join 

          from i in db.GetTable<DataImport>()
          join m in db.GetTable<DataMapping>() on i.DeclaredProduct equals m.DeclaredProduct into majointure
          where i.IdData == null
          from m in majointure.DefaultIfEmpty()

=> Table not found for 't7.t5.IdMapping'.'
the db.lastQuery is null

This left query works with linq2db v2.1 and then fails with next versions 

Thanks
i'm trying to migrate from version 1.10.2.0

sdanyliv:
@laurentKTR, could you try to use exact fields for join? Possible we have regression,
And please post your model with associations.

laurentKTR:
yes sorry #1716

PR #1255, opened 2018-07-29
update linq2db.t4models package publishing

### LINKED ISSUE #1251: Update linq2db.t4models package publishing for v2.x
We still didn't started to publish updated versions of this package for linq2db 2.x

MaceWindu:
Fix #1251

I preserved init.ps1 for now. Do we really need it for anything?


ili:
It is also needed to add build.props as for other packages with specifying MSBuild variable containing path to *.ttinclude files. This is needed to make includes in *.tt templates

MaceWindu:
@ili review

PR #1256, opened 2018-07-29
Add SourceLink support

### LINKED ISSUE #1217: Enable Sourcelink
https://github.com/dotnet/designs/blob/master/accepted/diagnostics/source-link.md#why-source-link

MaceWindu:
Fix #1217

### DIFF
--- a/NuGet/linq2db.nuspec
+++ b/NuGet/linq2db.nuspec
@@ -15,6 +15,7 @@
 		<tags>
 			linq linq2db LinqToDB ORM database DB SQL SqlServer Access SqlCe SqlServerCe MySql Firebird SQLite Sybase Oracle ODP PostgreSQL DB2 Informix SapHana
 		</tags>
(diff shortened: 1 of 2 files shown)


jogibear9988:
Any Idea why it does not work for net core?

jogibear9988:
Maybe PrivateAssets="All" is needed? see: https://github.com/dotnet/sourcelink

jogibear9988:
Note that .NET Core SDK 2.1.300 or newer is required for SourceLink to work. If building via desktop msbuild (as opposed to dotnet build) you'll need version 15.7.

MaceWindu:
Ok, as I suspected, it is due to some strange caching issues with nuget, when you use same package version for testing. I will fix travis build and merge this commit to test it once more with myget packages to check that appveyor tooling produce binaries with sourcelink support

MaceWindu:
Looks like travis have this issue: https://github.com/travis-ci/travis-ci/issues/9877
but because we don't publish with travis, we just need to disable sourcelink for it

Issue #1261, opened 2018-08-02
.All() generates incomplete WHERE


evilbloodydemon:
.All() method generates incomplete WHERE section for complex conditions on nullable fields - only first part of a condition is used.

### Steps to reproduce
See pull request #1262 

### Environment details
linq2db version: *master branch*
Database Server: *all*

sdanyliv:
Fixed in https://github.com/linq2db/linq2db/commit/7898b574b7a0f0d61eb6b40ee511a81e980e9e50

PR #1262, opened 2018-08-02
Test for issue #1261 - .All() generates incomplete WHERE

evilbloodydemon:


evilbloodydemon:
@jogibear9988 why is that? they're just plain booleans

jogibear9988:
sorry, you‘re right... i thought they return a list... silly

jogibear9988:
Why does appveyor not fail?

MaceWindu:
Because test disabled by [ActiveIssue] attribute

evilbloodydemon:
@jogibear9988 because of ActiveIssue attribute

jogibear9988:
ahhhh, I've not seen this...

Issue #1219, opened 2018-07-04
Conditional Mapping


RoyC-Retail-Assist:
I have dates stored in the database as DECIMAL(8,0) yyyyMMdd format and times as DECIMAL(6,0) hhmmss.

I'd like to be able to use these as DateTime however there are other DateTime values that are actually stored as DateTime on the database..
```cs
public class Test
{
	public int Id { get; set; }

	[Column(DataType = LinqToDB.DataType.Decimal, Length = 8, Precision = 0)]
	public DateTime Date { get; set; }
}
```

For reading data I can set a converter.
```cs
MappingSchema.Default.SetConverter<decimal, DateTime>(val =>
{
	if (val >= 10000101 && val < 99991231)
	{
		if (DateTime.TryParseExact(val.ToString(), "yyyyMMdd", null, System.Globalization.DateTimeStyles.None, out var retval))
			return retval;
	}

	if (val > 0 && val < 240000)
	{
		if (DateTime.TryParseExact(val.ToString(), "hhmmss", null, System.Globalization.DateTimeStyles.None, out var retval))
			return retval;
	}

	return DateTime.MinValue;
});
```

I can set a converter for writing data but this will apply to any DateTime > DataParameter:
```cs
MappingSchema.Default.SetConverter<DateTime, DataParameter>((obj) =>
{
	return new DataParameter { DataType = DataType.Decimal, Value = GetDecimalDate(obj) };
});
```

Is there any way of having a converter that is conditional upon _something_ so that any users of the class don't have to do any special to write the data? Alternatively, is there a better way of achieving this?

Hopefully these are dumb questions 😇 

### Environment details
linq2db version: 2.1.0
Database Server: Any
Database Provider: Any
Operating system: Windows 10
Framework version: .NET Framework 4.6.2

Franki1986:
Why don't add a custom attribute so that you can evaluate when decimal should be used?

RoyC-Retail-Assist:
The Column attribute has enough details to use but the Converter doesn't, as far as I can see, have any reference to the target column so the attribute cannot be accessed.  I'm guessing that there might be something further up the expression tree that can be utilised but haven't been able to find it yet hence my question.

sdanyliv:
@RoyC-Retail-Assist, how do you plan to use such classes with data fields? I assume only on client side.
Maybe better to extend your T4 template and generate additional readonly properties that reads data in appropriate way.

RoyC-Retail-Assist:
[...]

PR #1263, opened 2018-08-02
[WIP] Implemented flexible mechanism for supporting custom types.

sdanyliv:
Originally i have encountered problem when tried to use custom range types with `tstzrange`. Threre is only one type which implements timestamp range `NpgsqlRange<DateTime>` in Npgsql and there is no way to tell current engine that this particular convertation should be `timestam with range`. Analysis showed that we have to improve mapping logic.

This PR also help in the following case: #1219
- [x] Fix tests


jogibear9988:
As this changes a public API:

public virtual void SetParameter(IDbDataParameter parameter, string name, DataType dataType, object value)		 		

public virtual void SetParameter(IDbDataParameter parameter, string name, DbDataType dataType, object value)

also major should be increased when this is merged

MaceWindu:
We decided to include this pr in 2.x despite minor breaking changes for provider implementors as it contains very important changes we need for multiple features.

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs
```diff
@@ -441,7 +447,20 @@ protected override void SetParameterType(IDbDataParameter parameter, DataType da
...
 				case DataType.BinaryJson : if (_setJsonb     != null) _setJsonb(parameter);     else base.SetParameterType(parameter, dataType); break;
-				default                  : base.SetParameterType(parameter, dataType); break;
+				default :     
+				{
+					if (!string.IsNullOrEmpty(dataType.DbType))
+					{
+						var nativeType = GetNativeType(dataType.DbType);
+						if (nativeType != null)
+						{
+							_setNativeParameterType(parameter, nativeType);
```
MaceWindu: https://github.com/linq2db/linq2db/pull/1447 regression

MaceWindu: fixed

PR #1273, opened 2018-08-10
MERGE target hints support

### LINKED ISSUE #1272: Add merge target table hints support
Right now we don't generate target table hints for query like this:
```cs
db.TargetTable.With("HOLDLOCK")....
```
and it should be fixed

### Environment details
linq2db version: *2.2.0*

MaceWindu:
Fix #1272

SQL Server supports hints on target table, this PR adds support for hints generation.

~Because merge uses different ordering of hint and table alias, you cannot use `With` method, which will put hint after alias and need to use `WithTableExpression` method to specify proper ordering:~
```cs
// {0} is a table
// {1} is an alias
// for merge target alias must go after hint
table.WithTableExpression("{0} WITH (HOLDLOCK) {1}")
```

TODO:
- ~[ ] `UsingTarget` method produce bad sql. We should remove hint from target table, when it used as source~

Edit:
code above shouldn't be used for merge-specific hints, see below

### DIFF
--- a/Source/LinqToDB/DataProvider/BasicMergeBuilder.cs
+++ b/Source/LinqToDB/DataProvider/BasicMergeBuilder.cs
@@ -298,7 +298,7 @@ namespace LinqToDB.DataProvider
 
 		private void BuildSource()
 		{
-			Command.AppendLine("USING");
+			Command.AppendLine(" USING");
 
 			if (Merge.QueryableSource != null && SupportsSourceSubQuery)
 				BuildSourceSubQuery(Merge.QueryableSource);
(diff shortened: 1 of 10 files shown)


MaceWindu:
Taking into account 
- #1274 
- target hint having custom syntax
- table hints doesn't work well with `UsingTarget` API

I will add merge hint parameter of merge operation that user could pass to `Merge`/`MergeInto` methods.
This will allow us to generate hints in proper places easily for all databases that support merge-specific hints.

MaceWindu:
Ok, what we have. I've added overloads to `Merge`/`MergeInto` methods to accept string hint. Those overloads will accept only non-null values and execution of merge will throw exception if provider doesn't support hints.

### What providers supported (only those have hints according to documentation)

#### Sql Server
Hint applied to merge target (OPTIONS clause still not supported). Example:
```cs
db.TargetTable
    .Merge("HOLDLOCK")
    .UsingTarget()
    .OnTargetKey()
    .UpdateWhenMatched()
    .Merge();
```
```sql
MERGE INTO [TargetTable] WITH(HOLDLOCK) [Target] ....
```

#### Oracle
```cs
db.TargetTable
    .Merge("append")
    .UsingTarget()
    .OnTargetKey()
    .UpdateWhenMatched()
    .Merge();
```
```sql
MERGE /*+ append */ INTO TestMerge1 Target ...
```

#### Informix
```cs
db.TargetTable
    .Merge("AVOID_STMT_CACHE")
    .UsingTarget()
    .OnTargetKey()
    .UpdateWhenMatched()
    .Merge();
```
```sql
MERGE {+ AVOID_STMT_CACHE } INTO TestMerge1 Target ...
```

PR #1282, opened 2018-08-14
fixes #1281 support inherited attributes

### LINKED ISSUE #1281: Sealed classes
I see that some of the classes are marked with the `sealed` keyword while others are not.

From extensibility point of view, sealing classes causes problems.

Is there a specific reason for sealing classes or would you mind removing the seals or accepting a PR?

Non-sealed: `LinqToDB.Mapping.TableAttribute`, `LinqToDB.Mapping.IdentityAttribute`,  `LinqToDB.Mapping.NullableAttribute`

Sealed: `LinqToDB.Mapping.PrimaryKeyAttribute`

jogibear9988:


Source/LinqToDB/Metadata/SystemDataLinqAttributeReader.cs
```diff
@@ -14,7 +14,7 @@ public class SystemDataLinqAttributeReader : IMetadataReader
 		public T[] GetAttributes<T>(Type type, bool inherit)
 			where T : Attribute
 		{
-			if (typeof(T) == typeof(TableAttribute))
+			if (typeof(TableAttribute).IsAssignableFrom(typeof(T)))
```
sdanyliv: I think changes in this file not needed. linq2db will request exact types here.

jogibear9988: Your right. 
Didn't see that it was not about Linq2db attributes

PR #1286, opened 2018-08-19
[DO_NOT_MERGE] Various fixes to SQLite provider and WCF services

MaceWindu:
~SQLite released netstandard2.0-compatible version several days ago. I propose to update our linq2db.sqlite package dependency to this version.~ upgrade failed, but a lot of other issues found and fixed

Issues to handle here:
- [x] TestNumerics tests fail to roundtrip values for both MS and SQLITE providers using literals (works for parameters) - fixed by using assert with delta
- [ ] netcore projects fail to load System.Data.Sqlite due to https://system.data.sqlite.org/index.html/tktview?name=d292f2e23d - downgraded dependency to 1.0.108
- [x] Incorrect package for MS provider referenced by linq2db.sqlite.ms (doesn't belong to this PR, so if we will not ship it, this fix should be moved to master)
- [x] default value converter should prefer SomeType.Parse(string, IFormatProvider) to SomeType.Parse(string)
- [x] updated system.data.sqlite dependency from 1.0.107 to 1.0.108. 1.0.109.1 is unuasable right now
- [x] Explicitly set invariant culture for System.Data.Sqlite queries as it use current culture to convert numbers to text
- [x] Fixed issues with use of current culture to serialize/deserialize nullable value in WCF service
- [x] Fixed WCF service to use provider's execute scope. Fixes WCF tests for System.Data.Sqlite and Informix


MaceWindu:
Filled ticket for sqlite provider: https://system.data.sqlite.org/index.html/tktview?name=fb9e4b3087

ili:
If i'm not mistaken the same with MS provider. As for me this is not critical at all

MaceWindu:
Nope, this test pass for ms provider

MaceWindu:
Rechecked it with 2.x version of MS provider - you are right. Looks like it is a change in sqlite between 3.22 and 3.24

MaceWindu:
In this case I will update tests to take it into account

Source/LinqToDB/Common/ConvertBuilder.cs
```diff
@@ -110,7 +113,14 @@ static Expression GetParse(Type from, Type to, Expression p)
 		{
 			if (from == typeof(string))
 			{
-				var mi = to.GetMethodEx("Parse", from);
+				var mi = to.GetMethodEx("Parse", from, typeof(IFormatProvider));
```
MaceWindu: Need this change as for 2.1.0 sqlite.ms sqlite provider use parse for decimal due to this bug: https://github.com/aspnet/Microsoft.Data.Sqlite/issues/544 and on my locale it parse "1.0" as 10

jogibear9988:
Are we sure that always "." was used as a decimal separator?

sdanyliv:
@MaceWindu, your sqlite ticket closed as designed. So we have to think that this is acceptable.

MaceWindu:
Yep, I seen it yesteday. They are right, different parse algorithms produce different values when parse floats in non-roundtrip format.
I think real problem here is that we even generate such literal. I thought I changed it to use roundtrip format when was working on merge year ago, but looks like sqlite escaped this change for some reason.

Source/LinqToDB/DataProvider/InvariantCultureRegion.cs
```diff
@@ -4,16 +4,16 @@
 
 namespace LinqToDB.DataProvider.Informix
 {
-	internal class InformixCultureFixRegion : IDisposable
+	internal class InvariantCultureRegion : IDisposable
```
sdanyliv: Move `InvariantCultureRegion` to other namespace.

MaceWindu:
Looks like SQLite culture could be configured by connection string parameter `Flags=BindInvariantText`, so it makes sense to postpone those changes to sqlite provider/string parser and rethink it once more.

I will extract WCF service changes into separate PR.

MaceWindu:
Dismissed

PR #1291, opened 2018-08-27
Fix char parameters for microsoft.data.sqlite

### LINKED ISSUE #1279: Char mapping
```c#
	[TestFixture]
	public class CharValueTests : TestBase
	{
		class Table1
		{
			[PrimaryKey(1)]
			[Identity] public int Id { get; set; }

			public Char CharFld { get; set; }
		}

		[Test, DataContextSource(false)]
		public void Test(string context)
		{
			using (var db = new TestDataConnection())
			{
				db.DropTable<Table1>(throwExceptionIfNotExists: false);
				db.CreateTable<Table1>();

				var val = 'P';

				db.Insert(new Table1 { CharFld = val });

				var result = db.GetTable<Table1>().First().CharFld;

				Assert.AreEqual(val, result);

				db.DropTable<Table1>(throwExceptionIfNotExists: false);
			}
		}
	}
```

When run for SQLite.MS result is 56 ('8') not 80 ('P').  I've tested other values and it looks like the return value is the first digit of the value of the char rather than the char itself e.g. for 'z' 122 the result is '1', for char.MinValue 0 '\0' the result is 48 '0'

SQLite.Classic works fine so it would suggest that this is something that the underlying DataConnection is returning differently.

I'm getting the same problem with the DB2 iSeries provider (which is what prompted me to do further investigation) so this not confined to SQLite.MS.

### Environment details
linq2db version: 2.2.0
Database Server: N/A
Database Provider: SQLite.MS, DB2 iSeries and possibly others
Operating system: Windows 10
Framework version: .NET Framework 4.6.2

MaceWindu:
Fix #1279

Confirm that only microsoft's sqlite provider returns wrong value (both 1.x and 2.x)

### DIFF
--- a/Source/LinqToDB/DataProvider/SQLite/SQLiteDataProvider.cs
+++ b/Source/LinqToDB/DataProvider/SQLite/SQLiteDataProvider.cs
@@ -107,8 +107,7 @@ namespace LinqToDB.DataProvider.SQLite
 
 		public override void SetParameter(IDbDataParameter parameter, string name, DataType dataType, object value)
 		{
-			if (Name == ProviderName.SQLiteMS && value is char
-				&& (dataType == DataType.Char || dataType == DataType.NChar))
+			if (Name == ProviderName.SQLiteMS && value is char)
 			{
 				value = value.ToString();
 			}
--- a/Source/LinqToDB/DataProvider/SQLite/SQLiteDataProvider.cs
+++ b/Source/LinqToDB/DataProvider/SQLite/SQLiteDataProvider.cs
@@ -107,6 +107,12 @@ namespace LinqToDB.DataProvider.SQLite
 
 		public override void SetParameter(IDbDataParameter parameter, string name, DataType dataType, object value)
 		{
+			if (Name == ProviderName.SQLiteMS && value is char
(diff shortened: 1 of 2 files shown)


MaceWindu:
Looks like it is related to https://github.com/aspnet/Microsoft.Data.Sqlite/issues/296

MaceWindu:
Yep, we read char type from reader as string and then return first character. For some reason microsoft provider writes character code as string to database so instead of "P" it writes down "80" which is read back as "8".
To fix it we should probably change char parameter for sqlite.ms to string parameter

MaceWindu:
I've applied the same fix we already use for mysql. @RoyC-Retail-Assist, I think you need to do the same for iSeries provider.

MaceWindu:
Post-mortem:
MIcrosoft did strange decision to treat char as something more than char(1) on database/provider level which is not done by any other database/provider I aware of. It results in a lot of problems if user use different types (char vs string) for read and write same data.
In my fix I made sure we will not pass char as char Microsoft.Data.Sqlite provider, so it will not mess with it.

MaceWindu:
Microsoft will fix it in 3.0 https://github.com/aspnet/EntityFrameworkCore/issues/15020

Issue #1289, opened 2018-08-23
Improve SQL Server version detection


sdanyliv:
We have to improve SQL Server version detection. Nothing special here but when compatibility level is not recognized we have to use previous version detection algorithm based on SQL Server version.

Original issue here: https://github.com/linq2db/linq2db.EntityFrameworkCore/issues/7#issuecomment-415319151

PR #1293, opened 2018-08-27
Added additional Server version check for SQL Server.

sdanyliv:
Fixes issue #1289

### DIFF
--- a/Source/LinqToDB/DataProvider/SqlServer/SqlServerTools.cs
+++ b/Source/LinqToDB/DataProvider/SqlServer/SqlServerTools.cs
@@ -91,7 +91,7 @@ namespace LinqToDB.DataProvider.SqlServer
 
 								if (int.TryParse(conn.ServerVersion.Split('.')[0], out var version))
 								{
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/DataProvider/SqlServer/SqlServerTools.cs
```diff
@@ -91,21 +91,34 @@ static IDataProvider ProviderDetector(IConnectionStringSettings css, string conn
...
+											case  8 : return _sqlServerDataProvider2000;
+											case  9 : return _sqlServerDataProvider2005;
+											case 10 : return _sqlServerDataProvider2008;
+											case 11 : return _sqlServerDataProvider2012;
+											case 12 : return _sqlServerDataProvider2012;
+											default :
+												if (version > 12)
+													return _sqlServerDataProvider2012;
+												return _sqlServerDataProvider2008;
 										}
```
MaceWindu: We should write version and compatibility level to Trace log for debug purposes

Issue #1303, opened 2018-09-04
Wrong conversion of byte[] identifiers for MySQL


hardhub:
For MS SQL Server provider I will get correct values in WHERE clause.  
But in MySQL it looks like just ToString() called by mistake.  

```sql
--  MySql
SELECT
	`t1`.`Id`,
	`t1`.`Name`,
	`t1`.`Surname`
FROM
	`Account` `t1`
WHERE
	`t1`.`Id` IN (System.Byte[], System.Byte[], System.Byte[], System.Byte[])

```

It should be like this:

```sql
SELECT *
FROM Account
WHERE Id  IN (0x0164576A362F652476256346B2626501, 0x64743E66556712i71894244624623402, 0xD1923467267FDA742326E63234263423)
```

### Environment details
MySql 8.0
Linq2db: from master (2.4.0)

sdanyliv:
@hardhub, can you prepare PR with fix? Seems that this not so difficult task.

hardhub:
I guess such lines should be added in MySqlMappingSchema:

SetValueToSqlConverter(typeof(byte[]),         (sb,dt,v) => ConvertBinaryToSql        (sb, (byte[])v));
SetValueToSqlConverter(typeof(Binary),         (sb,dt,v) => ConvertBinaryToSql        (sb, ((Binary)v).ToArray()));

I will check and update it if all is correct.

sdanyliv:
@hardhub, thanks for fix. Also could you provide simple test that reproduces problem?
`CreateaLocalTable`, insert record, retrieve record, compare binary values. There are many similar tests in code.

hardhub:
> @hardhub, thanks for fix. Also could you provide simple test that reproduces problem?
CreateaLocalTable, insert record, retrieve record, compare binary values. There are many similar tests in code.

I think I cannot do that because I do not know what are requirements for tests in Linq2Db repository.  
I need to invest some time in understanding of source code and existing tests.  
Unfortunately that is not something doable in nearest future.  
If anybody can create test for this specific issue it will help me a lot in the future.

sdanyliv:
@hardhub, tests are needed and usually they have more code than fix :)
It's easy. Follow these steps:

## Initial - only once
1. Make copy of file `\linq2db\Tests\DataProviders.json`  -> `\linq2db\Tests\UserDataProviders.json` - this file mentioned in git ignore so it will not boring you when switching between branches.
[...]

PR #1304, opened 2018-09-04
Fix wrong conversion of byte[] identifiers for MySQL (#1303).

hardhub:


MaceWindu:
I've added test and following fixes:
- support binary fields in CreateTable for Firebird using BLOB type
- add missing literals support for System.Data.Linq.Binary for SAP HANA
- add missing literals support for byte[] and System.Data.Linq.Binary for Firebird, DB2, Sybase, SQL CE, SQLite, Access, PostgreSQL (MySQL support already added by @hardhub ), Oracle (using `HEXTORAW` function)

What I want to fix before merge:
- for some reason `InlineParameters` ignored by `byte[]` parameters but affect `Binary` parameters

Issue #1305, opened 2018-09-04
Add Order field for ColumnAttribute to allow greater control for code-first


Corey-M:
I propose that we modify `LinqToDB.Mapping.ColumnAttribute` to add an `Order` property to allow specification of column order for code-first scenarios.

I've been working on building a versioned record system for one of my projects.  I have a dozen tables that all start with the same fields:

    RecordID INT PRIMARY KEY,
    EffectiveStart DATETIME NOT NULL,
    EffectiveEnd DATETIME NULL,
    Key UNIQUEIDENTIFIER NOT NULL

In C# I have something like this:

    public abstract class VersionedRecord
    {
        // SQLite ROWID proxy
        [Column(IsPrimaryKey = true, SkipOnInsert = true)]
        public int RecordID { get; set; }

        [Column]
        public DateTime EffectiveStart { get; set; }

        [Column]
        public DateTime? EffectiveEnd { get; set; }

        [Column]
        public Guid Key { get; set; }
    }

    [Table("Accounts")]
    public class Account : VersionedRecord
    {
        [Column, NotNull]
        public string Code { get; set; }

        [Column, NotNull]
        public string Name { get; set; }
    }

When I create the table using `context.CreateTable<Account>();` the fields from the most-derived class are declared first.  In the case above the SQL statement looks like this:

	--  SQLite.Classic SQLite
	CREATE TABLE [Accounts]
	(
		[Name]           NVarChar(255)     NULL,
		[Code]           NVarChar(255)     NULL,
		[RecordID]       INTEGER       NOT NULL,
		[EffectiveStart] DateTime2     NOT NULL,
		[EffectiveEnd]   DateTime2         NULL,
		[Key]            Guid          NOT NULL,
	
		CONSTRAINT [PK_Accounts] PRIMARY KEY ([RecordID])
	)

(If I were creating the table myself I'd put the `Name` and `Code` fields after the `Key` field.)

What I would like is a way to specify the ordering of the columns independent of their position in the type(s) where they are defined.

In [`System.ComponentModel.DataAnnotations.Schema.ColumnAttribute`][1] there is an `Order` property which lets you define the order of columns.

Adding the `Order` property to `LinqToDB.Mapping.ColumnAttribute` would allow the manual specification of column ordering.  Columns with unset order values would come after the columns whose order is set.  Columns with conflicting order would be sorted as per current ordering.

### Environment details
linq2db version: *2.3.0*
Database Server: *N/A*
Database Provider: *All*
[...]

PR #1308, opened 2018-09-05
Add column order to ColumnAttribute and related code (#1305)

Corey-M:
Added `Order` field to `ColumnAttribute` and supporting code to change the ordering of columns in the table when created by `DataExtensions.CreateTable<T>(...)`.  Only change that interlocks with existing code is the `OrderBy` call in `SqlTable`.

Resolves issue #1305.


Corey-M:
I tested this manually with the POCOs from the issue.  Before changes the generated SQL is:

	--  SQLite.Classic SQLite
	CREATE TABLE [Accounts]
	(
		[Name]           NVarChar(255)     NULL,
		[Code]           NVarChar(255)     NULL,
		[RecordID]       INTEGER       NOT NULL,
		[EffectiveStart] DateTime2     NOT NULL,
		[EffectiveEnd]   DateTime2         NULL,
		[Key]            Guid          NOT NULL,
	
		CONSTRAINT [PK_Accounts] PRIMARY KEY ([RecordID])
	)

After changes the generated SQL is:

	--  SQLite.Classic SQLite
	CREATE TABLE [Accounts]
	(
		[RecordID]       INTEGER       NOT NULL,
		[EffectiveStart] DateTime2     NOT NULL,
		[EffectiveEnd]   DateTime2         NULL,
		[Key]            Guid          NOT NULL,
		[Name]           NVarChar(255)     NULL,
		[Code]           NVarChar(255)     NULL,
	
		CONSTRAINT [PK_Accounts] PRIMARY KEY ([RecordID])
	)

Source/LinqToDB/Mapping/ColumnAttribute.cs
```diff
@@ -311,5 +312,22 @@ public int   Scale
...
 		public string CreateFormat { get; set; }
+
+		private int? _order;
+		/// <summary>
+		/// Specifies the order of the field in table creation.
+		/// Default value: 999, ensure unspecified order puts columns after low numbered orders
+		/// </summary>
+		public int Order
+		{
+			get => _order ?? 999;
```
sdanyliv: Convert to auto property `int?`. Only during sorting apply defaults.

Source/LinqToDB/SqlQuery/SqlTable.cs
```diff
@@ -70,7 +70,7 @@ public SqlTable([JetBrains.Annotations.NotNull] MappingSchema mappingSchema, Typ
 			ObjectType   = objectType;
 			PhysicalName = physicalName ?? Name;
 
-			foreach (var column in ed.Columns)
+			foreach (var column in ed.Columns.OrderBy(_ => _.Order))
```
sdanyliv: Apply default here.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -289,6 +290,12 @@ string IColumnChangeDescriptor.ColumnName
 		/// </summary>
 		public string         CreateFormat    { get; private set; }
 
+		/// <summary>
+		/// Sort order for column list.
+		/// Default value: 999, ensure unspecified order puts columns after low numbered orders
+		/// </summary>
```
sdanyliv: Remove magic number

Source/LinqToDB/Mapping/ColumnAttribute.cs
```diff
@@ -311,5 +312,22 @@ public int   Scale
 		/// - {3} - identity specification.
 		/// </summary>
 		public string CreateFormat { get; set; }
+
+		private int? _order;
+		/// <summary>
+		/// Specifies the order of the field in table creation.
+		/// Default value: 999, ensure unspecified order puts columns after low numbered orders
```
MaceWindu: nit: xmldoc comments should have closing dot (.)

Source/LinqToDB/Mapping/ColumnAttribute.cs
```diff
@@ -311,5 +312,22 @@ public int   Scale
...
+		/// </summary>
+		public int Order
+		{
+			get => _order ?? 999;
+			set => _order = value;
+		}
+
+		/// <summary>
+		/// Returns <c>true</c> if <see cref="Order"/> was configured for current attribute
```
MaceWindu: dot missing

Source/LinqToDB/Mapping/PropertyMappingBuilder.cs
```diff
@@ -309,5 +309,15 @@ public PropertyMappingBuilder<T> HasScale(int scale)
 		{
 			return SetColumn(a => a.Scale = scale);
 		}
+
+		/// <summary>
+		/// Sets the Order of the database column.
+		/// </summary>
+		/// <param name="order">Column order</param>
```
MaceWindu: dots missing

Corey-M:
@MaceWindu If we use `int.MaxValue` then the unordered columns will always be at the end.  I changed it to a lower number specifically so that the user could manually order columns after the unordered set if they chose.  I'll think on how best to work this.

And put dots on my sentences :)

Source/LinqToDB/Mapping/ColumnAttribute.cs
Corey-M: Can't use `int?` or any nullable type as an attribute property.  Have changed to straight `int` in latest version.

Source/LinqToDB/SqlQuery/SqlTable.cs
Corey-M: Magic numbers removed in latest version.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
Corey-M: Magic numbers removed in latest version.

Source/LinqToDB/Mapping/ColumnAttribute.cs
Corey-M: Fixed in latest version.

Source/LinqToDB/Mapping/ColumnAttribute.cs
Corey-M: Fixed in latest version.

Source/LinqToDB/Mapping/PropertyMappingBuilder.cs
Corey-M: Fixed in latest version.

Corey-M:
In order to support ordering fields at the end of the table as well as the start I've changed the code so that negative order values sort at the end.  This lets me do things like:

	public abstract class VersionedRecord
	{
		[Column(IsPrimaryKey = true, SkipOnInsert = true, Order = 1)]
		public int RecordID { get; set; }
		[Column(Order = 2)]
		public DateTime EffectiveStart { get; set; }
		[Column(Order = 3)]
		public DateTime? EffectiveEnd { get; set; }
		[Column(Order = 4)]
		public Guid Key { get; set; }
		[Column(Order = -1)]
		public int AuditID { get; set; }
	}

Now when I derive from that class my records will have `AuditID` as the last field.

Corey-M:
Added test.  Creates table from derived class with specification of column order in base class and tests that the order is as specified.  Test allows any order for columns with no specified order, in case that changes in future.

Source/LinqToDB/SqlQuery/SqlTable.cs
```diff
@@ -70,7 +70,9 @@ public SqlTable([JetBrains.Annotations.NotNull] MappingSchema mappingSchema, Typ
 			ObjectType   = objectType;
 			PhysicalName = physicalName ?? Name;
 
-			foreach (var column in ed.Columns)
+			// Order columns by the Order field.  Positive first then negative.
+			var columns = ed.Columns.OrderBy(_ => _.Order >= 0 ? 0 : 1).ThenBy(_ => _.Order);
```
sdanyliv: I see here the following algorithm for sorting, which place ordered columns at the first and then columns without order but with saving their original position:
```cs
    var columns = ed.Columns.Where(c => c.Order != null).OrderBy(c => c.Order.Value);
    columns = columns.Concat(ed.Columns.Where(c => c.Order == null));
```

Tests/Linq/UserTests/Issue1305Tests.cs
```diff
@@ -0,0 +1,84 @@
...
+		/// Confirm that tables creation uses the <see cref="ColumnAttribute.Order"/> field correctly.
+		/// </summary>
+		/// <param name="context">Configuration string for test context.</param>
+		[Test, DataContextSource(false, "SQLite.MS")]
+		public void Test1(string context)
+		{
+			using (var db = new DataConnection(context))
+			{
+				// Force recreate table
+				db.DropTable<ColumnOrderTest>(throwExceptionIfNotExists: false);
```
sdanyliv: Here you can use our new approach, which do exactly the same but wit less code and automatic cleanup during table creation.
https://github.com/linq2db/linq2db/blob/master/Tests/Tests.Playground/TestTemplate.cs#L22

Source/LinqToDB/Mapping/ColumnAttribute.cs
```diff
@@ -311,5 +312,14 @@ public int   Scale
...
 		public string CreateFormat { get; set; }
+
+		/// <summary>
+		/// Specifies the order of the field in table creation.
+		/// Positive values first, then unspecified, then negative values.
+		/// </summary>
+		/// <remarks>
+		/// Ordering performed in <see cref="SqlTable.SqlTable(MappingSchema, Type, string)"/> constructor.
+		/// </remarks>
+		public int Order { get; set; } = int.MaxValue;
```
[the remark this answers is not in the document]
Corey-M: I can't.  You cannot have a Nullable property on an attribute.  They aren't supported, unless you never want to set the value.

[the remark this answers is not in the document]
Corey-M: Here's the reason why we can't use `int?` as a property on an attribute:

https://stackoverflow.com/a/20385488/920669

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -289,6 +290,12 @@ string IColumnChangeDescriptor.ColumnName
 		/// </summary>
 		public string         CreateFormat    { get; private set; }
 
+		/// <summary>
+		/// Sort order for column list.
+		/// Positive values first, then unspecified, then negative values.
+		/// </summary>
+		public int            Order           { get; private set; }
```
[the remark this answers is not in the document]
Corey-M: If it were possible to use `int?` for the `Order` property on the `Column` attribute then yes, this would also be `int?`.  Or if I add `HasOrder()` etc.

Source/LinqToDB/SqlQuery/SqlTable.cs
sdanyliv: @Corey-M, in my variant i’ve just tried to keep order the same aa before your change. No ordering in my case should be faster than just sorting by the same values which may change order if someone from Core library will cange sorting algorithm.

And about nullls, yes agree, just keep nullable variable as we do in other properties and return -1 if it is is not set.

Corey-M: I think you're missing the point that I have 3 grouping requirements:

1. Columns specified to sort at the start of the list
2. Columns with no specified order to go in the middle
3. Columns specified to sort at the end of the list

I don't mind what the actual ordering is inside the unspecified group - if I care I can specify them - so changes in the ordering of properties at the core level don't make much difference to me.  I just want to make sure that my primary key(s) and so on sort at the start and audit/change/whatever fields appear at the end.

So... with an `int` field with default `int.MaxValue` the three categories sort as:

1. Positive order values
2. Unspecified
3. Negative order values

The fastest way to do it in code is an unchecked cast from `int` to `uint`:

    ed.Columns.OrderBy(_ => unchecked((uint)_.Order))

Slightly slower but a little more obvious:

    ed.Columns.OrderBy(_ => _.Order == int.MaxValue ? 1 : _.Order >= 0 ? 0 : 1).ThenBy(_ => _.Order)

And the slowest option:

    ed.Columns.Where(_ => _.Order >=0 && _.Order < int.MaxValue).OrderBy(_ => _.Order)
    .Concat(ed.Columns.Where(_ => _.Order == int.MaxValue))
    .Concat(ed.Columns.Where(_ => _.Order < 0).OrderBy(_ => _.Order))

But honestly, we're talking about an operation that happens once in a while on a collection of items that is likely to be quite small.  If it's happening frequently on thousands of columns then there is probably something horribly wrong happening.  Speed isn't the most important factor here.

Corey-M:
@sdanyliv Could you check the replies to your change requests and let me know which you would prefer?  I favour the `int` version but I'll do the `HasOrder()` and so on if you so think it's the better option.  I can do the sorting regardless.

On performance... saving a few microseconds on the sort is probably not an important aspect compared to making the code readable and maintainable.  I'm happy to do a multi-pass concat to get each group (positive, unspecified, negative) into the right sequence.  I like the brevity of the unchecked cast but I'm not wedded to it.

@MaceWindu your feedback would also be appreciated.

Corey-M:
@sdanyliv @MaceWindu I know you guys are probably busy, but I'd like to get this sorted.  Could you please cast an eye over my questions and give me some direction?

sdanyliv:
@MaceWindu, can you handle this? I think we can include this PR in the upcoming release.

MaceWindu:
Will do

Source/LinqToDB/Mapping/ColumnAttribute.cs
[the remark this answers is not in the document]
igor-tkachev: We can always use nullable type inside that attribute.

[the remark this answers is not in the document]
Corey-M: @igor-tkachev I started out that way, then realized that I hadn't allowed for sorting columns to the end.  When sdanyliv and MaceWindu pointed out that the magic number constant I used while trying to address the issue was poor form - and I agreed - using `Int32.MaxValue` and negative `Order` values made sense.  If you want I'll switch to null values internally.  It's not a big change, and I'm happy to go with the existing format.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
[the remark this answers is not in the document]
Corey-M: This is part of the other comments and will be resolved on those.

Source/LinqToDB/SqlQuery/SqlTable.cs
Corey-M: Resolved by latest commit from MaceWindu.

PR #1332, opened 2018-10-01
use assembly-qualified name for data types as Type.GetType() fails to…

### LINKED ISSUE #1331: PostgreSQL T4 model generator use object type for some database types
Original issue reported here #1320 in comments

### Environment details
linq2db version: *2.3.0*

MaceWindu:
… locate provider-specific types without assembly name

Fix #1331


MaceWindu:
I would add that for specific case, menthioned in original issue, npgsql type will be used instead of DateTime/object as before.

PR #1333, opened 2018-10-01
[WIP] Fix: Convert Firebird UInt32 to Int32 when dialect is 1

### LINKED ISSUE #1318: Firebird int comparison is cast to BigInt although dialiect 1
LinqToDB-Version : 2.3.0
Firebird-Version: 2.1.3
SQL-Dialect: 1

Hello,

I run a simple SQL that searches for a specific number:

UInt32 number = 50;
using (var context = new DbContext())
               {
                  var p = context.Persons.FirstOrDefault(p1 => p1.Age == number)
               }

In  the resulting sql the Age property is cast to a BigInt. 
The mapped property is UInt16.

The problem is that BigInts are not supported by SQL-Dialect 1.

MaceWindu:
Fix #1318
- New configuration to set a dialect in FirebirdConfiguration
- If the dialect is set to 1 then convert UInt32 to Int32

### DIFF
--- a/Source/LinqToDB/DataProvider/Firebird/FirebirdConfiguration.cs
+++ b/Source/LinqToDB/DataProvider/Firebird/FirebirdConfiguration.cs
@@ -4,6 +4,20 @@ namespace LinqToDB.DataProvider.Firebird
 {
 	public static class FirebirdConfiguration
 	{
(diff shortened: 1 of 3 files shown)


MaceWindu:
continued in https://github.com/linq2db/linq2db/pull/1334

PR #1334, opened 2018-10-01
[DO_NOT_MERGE][WIP] Fix: Convert Firebird UInt32 to Int32 when dialect is 1

### LINKED ISSUE #1318: Firebird int comparison is cast to BigInt although dialiect 1

MaceWindu:
This is re-publish of #1333 as I don't have permissions to push to that PR.

Fix #1318
- New configuration to set a dialect in FirebirdConfiguration
- If the dialect is set to 1 then convert UInt32 to Int32


MaceWindu:
@Franki1986 I've changed the way we get information about dialect to avoid new public-surface members so it will be easier for us to refactor it later.

Could you test that this PR correctly works for you when dialect1 set?

Franki1986:
Sure , will check this tomorrow

MaceWindu:
Looks like this approach will not work with our current architecture, because:
- data provider called before we know connection string
- data provider could be shared between different connections

I think I will revert it to use your approach for now with static configuration option. At least it will work

MaceWindu:
I suspect it will not make into next release. I started testing it and it doesn't work very well.

Franki1986:
Is there skmething I can test explicit?The tests against my db worked well..

MaceWindu:
I will submit my tests soon. Currently I see following issues:
- it is not correct to convert uint to int as it will convert incorrectly values in (int.MaxValue, uint.MaxValue] range. Unsafe conversion produce negative value which is wrong. E.g. you cannot use it in comparison operators. For this situation we can use same trick linq2db use for ulong type - map it to decimal type.
- current solution lacks proper literal generation (also it doesn't work in general for ulong according to my tests)
- Current implementation doesn't allow you to work with different dialects connections using same provider. Taking into account that dialect1 is very special case, we can tolerate it for now and just tell people to not mix dialects in single application for now

Taking into account that all this should be implemented, tested and reviewed - I pretty sceptical about it in 2.4 release.

MaceWindu:
Code you linked performs conversion that could lead to overflow. If you really need such unsafe conversions we cannot add them to linq2db and could only provide missing extension points to configure linq2db to work like that.

Or using C# unit/long/ulong as decimal in database will cover your needs?

MaceWindu:
Ok, played around with mapping schema and provider and I think for your case it will be enough to configure linq2db like below

uint to int unsafe mapping:
```cs
mappingSchema.SetDataType(typeof(uint), DataType.Int32);
mappingSchema.SetValueToSqlConverter(typeof(uint), (sb, dt, v) => sb.Append(unchecked((int)(uint)v).ToString(NumberFormatInfo.InvariantInfo)));

Expression<Func<IDataReader, int, uint>> expr = (r, i) => unchecked((uint)r.GetInt32(i));
provider.ReaderExpressions[new ReaderInfo { FieldType = typeof(int), ToType = typeof(uint) }] = expr;
```

Similar options you need to set for long and ulong types.

Please check if it will work for you.

Franki1986:
Thanks for your awesome work! Looking back one year I had to decide between ef core and linq2db, till now I don't regret anything! I had to work for a long time with Nhibernate and it was often frustrating, because there were often an unresolvable problem. With the linq2db team every problem is handled like it is a private problem and there is always a way for a solution, no matter how rare a constelation is.. Had to say these words!

MaceWindu:
Superseeded by #2693

PR #1335, opened 2018-10-03
Release 2.4.0

MaceWindu:
@linq2db/testers we are planning to release 2.4.0 tomorrow. Would be nice if you can test it

Release notes: https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap

Remaining tasks:
- [x] run full test suite


MaceWindu:
Waiting for CI and releasing.
Last commit fixed:
- issue with informix provider trying to generate binay literal for `byte[]`/`Binary` types. As I didn't found any information about such literals it will always use parameter for those types
- fixed serveral tests
- disabled bunch of SAP HANA tests that were failing for a long time as they test something we don't support for HANA provider

Issue #1313, opened 2018-09-07
Error "An item with the same key has already been added." during query execution


panuckov-nick:
Prerequisites: you have object **A** with property  **Prop** in one namespace and you have object **B** that is inherited from object **A** and hiding property **Prop** by using new keyword in property notation.
```cs
public class A {
    public int Prop {get; set;}
}

public class B : A {
    public new long Prop {get; set;}
}
```
When you are using object **B** as a result of the query you getting error, in my case property ***Prop** even was not used in the query - it was just a part of used object **B**.

Reason to have such strange behavior is step by step migration from one set of business object to other that is located in different namespace.

Exception message: "An item with the same key has already been added."
Stack trace:
```
 at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at LinqToDB.Linq.Builder.ExpressionBuilder.ProcessProjection(Dictionary`2 members, Expression expression) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\ExpressionBuilder.SqlBuilder.cs:line 2923
   at LinqToDB.Linq.Builder.SelectContext..ctor(IBuildContext parent, LambdaExpression lambda, IBuildContext[] sequences) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\SelectContext.cs:line 51
   at LinqToDB.Linq.Builder.SelectBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\SelectBuilder.cs:line 50
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\ExpressionBuilder.cs:line 175
   at LinqToDB.Linq.Builder.FirstSingleBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\FirstSingleBuilder.cs:line 25
[...]

PR #1336, opened 2018-10-05
Address #1313 - Duplicate members when field/property replaced with 'new'

Corey-M:
Done as new `GetPublicInstanceValueMembers` extension.

Existing `GetPublicInstanceMembersEx` extension not modified.  Can likely be removed, doesn't appear to be used elsewhere.


Corey-M:
Had some problems getting the tests to work at all.  Might be a problem with the `Mono.Cecil` package being included in `Tests` project.  Removed the nuget package and tests worked again.  Didn't commit that change to the PR though.  Will raise a separate issue for this.

Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -121,6 +121,41 @@ public static MemberInfo[] GetPublicInstanceMembersEx(this Type type)
...
+
+				orderby h.ord, m.ord
+
+				group m.mem by m.mem.Name into grp
+				select grp.First();
+
+			return members.ToArray();
+		}
+
+		public static Type[] GetTypeHierarchy(this Type type)
```
sdanyliv: Simplify this function to IEnumerable, less allocations:
```cs
		public static IEnumerable<Type> GetTypeHierarchy(this Type type)
		{
			var current = type;
			do
			{
				yield return current;
				current = current.BaseTypeEx();
			} while (current != null && current != typeof(object));
		}
```

Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -121,6 +121,41 @@ public static MemberInfo[] GetPublicInstanceMembersEx(this Type type)
 			return type.GetMembers(BindingFlags.Instance | BindingFlags.Public);
 		}
 
+		public static MemberInfo[] GetPublicInstanceValueMembers(this Type type)
```
sdanyliv: There is another variant with simplified algorithm and reduced allocation.
```cs
		public static MemberInfo[] GetPublicInstanceValueMembers(this Type type)
		{
			var result       = new LinkedList<MemberInfo>();
			var alreadyAdded = new HashSet<string>();

			foreach (var t in GetTypeHierarchy(type))
			{
				var members = t.GetMembers(BindingFlags.Instance | BindingFlags.Public | BindingFlags.DeclaredOnly)
					.Where(m => m.MemberType == MemberTypes.Field ||
					            m is PropertyInfo pi && pi.GetIndexParameters().Length == 0);

				foreach (var m in members)
				{
					if (alreadyAdded.Contains(m.Name))
						continue;
					result.AddFirst(m);
					alreadyAdded.Add(m.Name);
				}
			}

			return result.ToArray();
		}
```

Source/LinqToDB/Extensions/ReflectionExtensions.cs
Corey-M: Of course there's a BaseTypeEx.  I didn't even look :)

Corey-M: That was about the third iteration.  I had a couple of other methods that were using the same iterator, but they were superfluous.  Your code looks OK.  I'd prefer to have the test at the top to handle cases where `type` is null or `typeof(object)`.

What do you mean by less allocations though?

sdanyliv: Ok, moving `while` up, is ok.
Less allocations because we do not create array - we do not need it.

Source/LinqToDB/Extensions/ReflectionExtensions.cs
Corey-M: Won't work unfortunately.  If you replace a property with a new field in a derived class your code will return the property from the base instead of the new field.  Have a look at the test I wrote to confirm all the possible combinations of changing value type (int to long) and between field and property.

Also have a look at what you get back from `GetMembers(typeof(I1313_Derived))`.  All the properties are at the top, all the fields are at the bottom.  I think it's because they essentially get each type of member in separate operations and concat the results.

Which is why I needed to grab the class hierarchy so that I could order first by the class's position in the hierarchy.  That way even if we replace a property with a field we get the right output.

I don't see an easy way to preserve all of the ordering that I need without using the indexed selects.  I could maybe cut down on the anonymous types if we could use `ValueTuple`s.  I could almost certainly drop the one on the members, just don't like to assume that ordering will be preserved through all of the operations.

Give me a bit to re-work it, see if I can find a more obvious way to do this.  Shame though... I *really* like LINQ 😁

Corey-M: Oh, heck, I was wrong.  I missed the outer loop in your code that does the whole thing one class at a time in the hierarchy.  That would in fact work.

Now that we have a couple of options - 3 including the one I was just finishing off - it might be a good idea to do some profiling and figure out what we can do to maximize the speed.  I'll throw the various methods through some high-iteration testing to see how slow they are in various scenarios.

Perhaps the best place to start would be a shortcut test for objects that are derived from `System.Object` and just filter the `GetMembers` call without the additional window dressing.  Start out the same for any class that's deeper than that, and only do the full thing when we need to because we've found a duplicate.

I'll play around and see what I can figure out.

sdanyliv: Your sollution has a lot of allocations, soring, grouping. That‘s why i‘m warning you. Linq2db still fastest ORM because of controlling core functionality to be effective anyway.

If you also test my solution, try different approaches. Store in List instead of LinkedList, append items and reverse list at the end.

Corey-M: I wasn't certain that it needed to be reversed since the current output isn't.  When it comes to derived classes in general it's probably nicer to have the base class fields & properties appear first, but getting true declared order is probably not going to be possible.

After testing on a variety of different classes and situations, the situation is still a little slightly unclear.

* The code I committed with a single LINQ expression at best takes twice the time, so it can go in the trash right up front.

* Using a list and reversing it at the end slows the process down a little.  If we want to reverse the list then that's not the way to go.  It's faster to use a `LinkedList<>` for both cases.

* Using an iterator to get the hierarchy is slower than using a for loop.  

* For flat classes (derived from `object` or `ValueType`) your method lags (by ~20% for simple cases) vs testing the base type and returning the filtered list of members.

* For derived classes it is marginally faster to use your code than my revised version by about 15% for a derived class, reducing the deeper you go.

The final two points are of interest.  If we are commonly using flat POCO types - which is certainly the normal case for me - then the fastest solution for those is to use what amounts to the old format.  I have a hybrid that works well for flat POCOs and only a little slower for derived classes.

Latest commit has new code.  It does not reverse the order, both because it is more consistent with current output and because it takes longer to do in the case of non-derived classes.

Corey-M:
@sdanyliv Lots of testing, have posted a new version.  Big discussion on your old comment thread, but basically this is the fastest I could find for flat classes and for deep hierarchies (4+ levels).  It's a bit slower for single-derived, but only by ~15%.  Worst case scenario (from the ones I tested) it's a few microseconds slower on a single derived class.  Shouldn't be any slower than the pre-patch version for flat classes.

I didn't go with reversed ordering since it didn't seem to be useful at the time.  It reverses both the class and member order, so it wasn't going to do what I wanted it to.  All you have to do is change `.AddLast` to `.AddFirst`.

Corey-M:
@sdanyliv I've looked and can't see any further improvements.  I might have missed something in testing.

sdanyliv:
@Corey-M, your code looks OK for me. I'll run benchmark tests later. Thanks for your work!

Issue #1350, opened 2018-10-17
Is there a way to prevent a value from being converted into a SqlParameter?


matthewjones555:
I'm trying to work around an issue in SqlCe 3.0, where you cannot use a SQL parameter with a db function, such as COALESCE, or ISNULL.

Is there a way of telling LinqToDb to not parameterize a specific value?

I've been digging into the source, but was wondering if there was a built in way of doing this, before I look at making a fork.

### Steps to reproduce
Use a query that includes a coalesce on a SQL CE database.

```c#
var query = from c in context.MyTable
            where (c.StartDate ?? SqlDateTime.MinValue.Value) <= DateTime.Now
            select c;
```

This produces the following SQL

```SQL
-- MyContext SqlCe
DECLARE @p1 DateTime
SET     @p1 = '1753-01-01'
DECLARE @p2 DateTime
SET     @p2 = '2018-10-17 10:07:02.353'

SELECT
	[t1].[Id],
	[t1].[FirstName],
	[t1].[LastName],
	[t1].[StartDate]
FROM
	[MyTable] [t1]
WHERE
	Coalesce([t1].[StartDate], @p1) <= @p2
```

Here's the exception that running the query produces.

```
Exception message: System.Data.SqlServerCe.SqlCeException: 'The specified argument value for the function is not valid. [ Argument # = 2,Name of function(if known) = coalesce ]'
Stack trace:
   at System.Data.SqlServerCe.SqlCeCommand.CompileQueryPlan()
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteCommand(CommandBehavior behavior, String method, ResultSetOptions options)
   at System.Data.SqlServerCe.SqlCeCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.System.Data.IDbCommand.ExecuteReader(CommandBehavior behavior)
   at LinqToDB.Data.DataConnection.ExecuteReader(CommandBehavior commandBehavior)
   at LinqToDB.Data.DataConnection.ExecuteReader()
   at LinqToDB.Data.DataConnection.QueryRunner.ExecuteReader()
   at LinqToDB.Linq.QueryRunner.<ExecuteQuery>d__9`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
```

[...]

PR #1352, opened 2018-10-18
Use inline function parameters for SQL CE, for issue #1350

matthewjones555:
I've created this pull request in relation to the following issue.

https://github.com/linq2db/linq2db/issues/1350

SQL CE 3.0 doesn't support parameters as function arguments, so I've forced them to be used inline.

I'm not sure if I've done this in the correct place, or in line with the development practices that the maintenance team use, so any feedback will be welcome.  I'm happy to make any changes you guys would like.

Thanks,

Matt


matthewjones555:
@sdanyliv, just to clarify, this issue is for SqlCe 3.0, but doesn't seem to be a limitation of 4.0.  I have no idea about 3.5, so I'm going to run tests with it.  The documentation for 3.0 is poor, and I can't seem to dig up where I found the information from.  I think it was a reply from Microsoft buried in a Microsoft forum thread.  I will continue to search and see if I can dig it up.  

It's clearly unnecessary to inline all function parameters, so is there a way to check which version of CE we're currently using at runtime, from withing the SqlCeSqlOptimizer?  That way I can only switch on the fix if necessary.  It'd be useful if this feature didn't have to touch too much other stuff.

I'm currently looking at writing unit tests, but the tests currently seem to use CE 4.0.  I'm evaluating how easy it would be to add a 3.0 and 3.5 database to test.

As a worst case scenario, I'm happy to continue to work from a fork, but it would be better to help improve things for any other poor souls who are still stuck on 3.0!

sdanyliv:
@matthewjones555, in unit tests you just need to check that parameters are corrected.
There is `GetSelectQuery()` extensions for `IQueryable` which gives access to generated AST in testing environment.

Since we still have not designed good approach for version auto-detection and connection specific settings, i propose to add new static class `SqlCeConfiguration` with static property `InlineFunctionParameters`.  You can use as sample `FirebirdConfiguration`.

matthewjones555:
@sdanyliv, thanks for the pointers.  I will take a look.

It'd be nice to make it do a version check and enable the setting automatically, that'd give people the best user experience with it, but I suppose a manual setting is better than nothing.  I'll take a look after I've done the unit tests.

sdanyliv:
@matthewjones555, there is update. I have investigates how caching in this case works, so it is not needed for testing. What is manadatory is to provide static configuration, yes it is not so good but until design is not finished it is better than nothing.
Also (i’m also still learning some places), there is already implemented functionality which can inline parameters anywhere in linq query. So this query should work without your fix:
```cs
var query = from c in context.MyTable
            where (c.StartDate ?? Sql.ToSql(SqlDateTime.MinValue.Value)) <= DateTime.Now
            select c;
```
It works because `Sql.ToSql()` forces inlining parameters.

MaceWindu:
I will finalize this pr - will add option and some tests.
As @sdanyliv already said - unfortunatelly behavior like that should be configured using options, because current architecture doesn't allow us to detect db version automatically without unwanted side-effects. Hope it will be fixed in v3.

MaceWindu:
As you can see from tests, option have no effect for already cached queries

Issue #1355, opened 2018-10-19
Tracing DataReader timings?


yallie:
Hello,

I'm planning to update my project from BLToolkit to the latest version of Linq2db. 
My project relies on some BLToolkit internals for tracing SQL queries.
Apparently they are not relevant for Linq2db :)

I've implemented custom DataReader class to measure query execution timings:
* from ICommand.ExecuteReader to IDataReader.Read → query execution time
* from IDataReader.Read to IDataReader.Dispose → data fetching time

I used to use custom data providers with their own MappingSchema classes and
override CreateDataReaderMapper to wrap the original DataReader with my wrapper.
That worked just fine for my BLToolkit-based application.

What can I do in Linq2db to achieve similar functionality?
I see that DataConnection has this handy OnTraceConnection callback.
That's cool, but I don't seem to be able to track DataReader's lifetime with that.
Can I plug into the data provider pipeline to wrap the original DataReader with my own class?
Or is there any better option?

Thanks :)

sdanyliv:
@yallie, `OnTraceConnection` gives all needed information

>from ICommand.ExecuteReader to IDataReader.Read → query execution time

TraceInfoStep.AfterExecute get info.ExecutionTime

>from IDataReader.Read to IDataReader.Dispose → data fetching time

TraceInfoStep.Completed get info.ExecutionTime

>but I don't seem to be able to track DataReader's lifetime with that.

Do you still needs injecting in this functionality?

yallie:
Oh, that sounds great!
If TraceInfoStep.Completed gives the fetch timings, that's all I need :)
The only reason I would wrap the DataReader is to measure these timings.

But looking at the sources, I see that Completed step is only invoked by QueryRunner.
So I assume, it won't be triggered when I read the DataReader myself, right?
I don't seem to be getting the Completed step when doing something like that:

```c#
using (var db = new DataConnection())
{
	DataConnection.TurnTraceSwitchOn();
	db.OnTraceConnection = (e) =>
	{
		Console.WriteLine("Step: {0}, Time: {1}", e.TraceInfoStep, e.ExecutionTime);
	};

	using (var reader = db.ExecuteReader("select * from my_table"))
	{
		var query = reader.Query<MyTable>(); // mapped to my_table
		foreach (var record in query)
		{
			Console.WriteLine($"Record: {record}");
		}
	}
}
```

Console log:

```
Step: BeforeExecute, Time:
Step: AfterExecute, Time: 00:00:00.0189886
[...]

PR #1358, opened 2018-10-20
DataReader reporting TraceInfoStep.Completed step

yallie:
Issue #1355.

Added tracing to the DataReader wrapper's Dispose method.
Looks like the project didn't have any tests related to the tracing, so I've added a fixture.


yallie:
@sdanyliv Thanks for approving my patch!
Anything else I can do to help merging this PR?

jogibear9988:
The DateTime Type and Now Value should stay?

jogibear9988:
It could also produce wrong results when dailight saveing switch happens between the call

MaceWindu:
Hmm, right. Makes sense to calculate duration more safely.

MaceWindu:
PRs are welcome :)

MaceWindu:
#1367

PR #1360, opened 2018-10-22
Improving Subquery optimization.

sdanyliv:
Added new IQueryable extensions.

- `AsSubQuery()` - forces Optimizer do not remove subquery during optimization.
- `HasUniqueKey()` - registers unique keys for tables or even subqueries. It should cause more aggressive join optimization. For example it is useful if we have `LEFT JOIN` to subquery and we know that result of this subquery always distinct by specific keys and in join we have used all of these keys. If in projection we do not use columns of this subquery - subquery will be removed because it do not change count of resulting records.

Improved JoinOptimizer to handle `group by` and `distinct`  subqueries

To be done:
- [x]  Review merging keys during optimization
- [x]  Review visitors


Tests/Tests.Playground/Tests.Playground.csproj
```diff
@@ -70,9 +70,17 @@
     <Reference Include="IBM.Data.Informix">
       <HintPath>..\..\Redist\IBM\IBM.Data.Informix.dll</HintPath>
     </Reference>
+    <Reference Include="Sap.Data.Hana.v4.5">
+      <HintPath>..\..\Redist\SAPHana\Sap.Data.Hana.v4.5.dll</HintPath>
+    </Reference>
     <Reference Include="Sybase.AdoNet45.AseClient">
       <HintPath>..\..\Redist\Sybase\Sybase.AdoNet45.AseClient.dll</HintPath>
     </Reference>
+    <Reference Include="Oracle.DataAccess, Version=4.122.1.0, Culture=neutral, PublicKeyToken=89b483f429c47342, processorArchitecture=x86">
```
MaceWindu: why it was added back by merge??

PR #1367, opened 2018-10-25
Fix query execution time calculation

MaceWindu:
Fixes query execution time calculation for tracing to be not affected by DST shifts

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.Async.cs
+++ b/Source/LinqToDB/Data/DataConnection.Async.cs
@@ -68,7 +68,7 @@ namespace LinqToDB.Data
 
 			try
 			{
-				var now = DateTime.Now;
+				var now = DateTime.UtcNow;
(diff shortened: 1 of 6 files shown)


jogibear9988:
Why not tickcount?

https://stackoverflow.com/questions/4075525/why-are-datetime-now-datetime-utcnow-so-slow-expensive

yallie:
I'm sorry I blindly copied that DateTime code when adding tracing to the DataReader.
My BLToolkit-based applications use Stopwatch for the DataReader tracing events.
Stopwatch gives much more accurate results than DateTime-based arithmetic.

>Why not tickcount?

I believe Environment.TickCount is the same hi-res tick counter used by Stopwatch internally?

>I don't mind to have tick count, but don't want to waste my time on such microoptimizations :)

Can I change it myself then?
That's not about performance actually, that's about the accuracy :)
Also, I'd like to add the StartTime property to the TraceInfo event.

MaceWindu:
superseeded by #1368

PR #1368, opened 2018-10-25
Using Stopwatch to trace execution times

yallie:
* Using Stopwatch to measure intervals
* Added TraceInfo.StartTime property

### DIFF
--- a/Source/LinqToDB/Data/TraceInfo.cs
+++ b/Source/LinqToDB/Data/TraceInfo.cs
@@ -19,6 +19,7 @@ namespace LinqToDB.Data
 		public TraceLevel     TraceLevel       { get; set; }
 		public DataConnection DataConnection   { get; set; }
 		public IDbCommand     Command          { get; set; }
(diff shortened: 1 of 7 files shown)


Source/LinqToDB/Data/TraceInfo.cs
```diff
@@ -19,6 +19,7 @@ public TraceInfo(TraceInfoStep traceInfoStep)
 		public TraceLevel     TraceLevel       { get; set; }
 		public DataConnection DataConnection   { get; set; }
 		public IDbCommand     Command          { get; set; }
+		public DateTime?      StartTime        { get; set; }
```
MaceWindu: We need a comment to tell people that this is UTC time

sdanyliv: @yallie, it would be nice if you help us to write XML Doc for this class.

yallie: Sure, no problem!

PR #1372, opened 2018-10-28
Add DataConnection.OnConnectionOpened event

### LINKED ISSUE #1113: Cannot access when i use  conn.ChangePassword("string") method.
When I connect at sqlite database, throw exception "(26) : file is not a database".

I tried Sqlitebrowser's Set Encryption and this code.

```cs
SQLiteConnection conn = new SQLiteConnection("Data Source=database.sql;Version=3;");
conn.Open();
conn.ChangePassword("pw");
conn.Close();
```

```cs
SQLiteConnection conn = new SQLiteConnection("Data Source=database.sql;Version=3;Password=pw;");
conn.Open();
conn.ChangePassword((string)null);
conn.Close();
```

My environment.
Use ILinqToDBSettings

```cs
public class ConnectionSetting : ILinqToDBSettings
{
    private const string Provider = "SQLite";

    public IEnumerable<IDataProviderSettings> DataProviders { get { yield break; } }
    public string DefaultConfiguration => Provider;
    public string DefaultDataProvider => Provider;
    public IEnumerable<IConnectionStringSettings> ConnectionStrings
    {
        get
        {
            yield return new ConnectionStringSettings()
            {
                Name = "db",
                ProviderName = Provider,
                ConnectionString = "Data Source=database.sql;Version=3;Password=pw;"
            };
        }
    }
}
```

So. I want use password. 
How can I?

Thanks.

MaceWindu:
Fixes #1113

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.Async.cs
+++ b/Source/LinqToDB/Data/DataConnection.Async.cs
@@ -31,8 +31,7 @@ namespace LinqToDB.Data
 						await ((DbConnection)_connection).OpenAsync(cancellationToken);
 
 					_closeConnection = true;
-
-					OnConnectionOpened?.Invoke(this, _connection);
+					await OnConnectionAsyncOpened?.Invoke(this, _connection, cancellationToken);
 				}
 				catch (Exception ex)
 				{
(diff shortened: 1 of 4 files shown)


sdanyliv:
@MaceWindu, if you open connection async, event should be async and part with crypto keys also should be async. Code duplication but it how async works.

PR #1374, opened 2018-10-29
[ci skip] fix link to Analytic file

hahn-kev:
previously it was linking directly to a line number, but that line didn't make much sense to link to anymore, might want to fix that


MaceWindu:
Thanks!

PR #1384, opened 2018-11-04
fix ODP.NET GAC discovery

MaceWindu:
- Add support for ODP.NET client discovery in GAC (similar to [this issue with SAP HANA](https://github.com/linq2db/linq2db/issues/239))
- Remove local copies of SAP HANA and ODP.NET assemblies from repository

This change required for LinqPad to properly locate ODP.NET provider and resolve issue for users when our provider assembly doesn't match version of installed native part of provider.


MaceWindu:
BTW, similar fix could/should be done for Sybase.ASE and probably DB2 LUW providers (v2.6?)

PR #1388, opened 2018-11-06
Added FromSql method for supporting Raw SQL queries in linq queries.

sdanyliv:
Implemented long waited `FromSql` method which supports Raw SQL inside linq queries.

To be done:
- [x] Correct parameters naming
- [x] Code polishing
- [x] Apply ident for raw query


Source/LinqToDB/LinqToDB.csproj
```diff
@@ -75,6 +75,9 @@
 		<Reference Include="System.Runtime.Serialization" />
 
 		<Compile Remove="Metadata\SystemComponentModelDataAnnotationsSchemaAttributeReader.cs" />
+	</ItemGroup>
+
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net40' OR '$(TargetFramework)' == 'net45' OR '$(TargetFramework)' == 'net451' ">
```
MaceWindu: I think we can cleanup net40/net451. it was pre-2.x artifact

sdanyliv: Will do, thanks.

PR #1389, opened 2018-11-07
Fixes issue #1365. Added additional methods for simplifying mapping.

### LINKED ISSUE #1365: Inferring columns based on interface mappings
This is a follow on from #1099 which I wrote last year concerning queries which were created using interface entities.

## Overview - my goal
We build bespoke web applications, many of which need a shopping cart.  As such, we have abstracted the common functionality into a different assembly.  This common functionality includes data queries such as `GetCart()`, `GetCartItems()` etc.  

Because we do not know the precise database layout of each individual client app that will use this library, our data queries operate over _interfaces_ - not _classes_.  For example, we have an interface:

``` c#
public interface ICartItem {
    int CartItemID {get;set;}
    string Name {get;set;}
}
```

Our data interface (nothing to do with L2DB yet) is:

```c#
public interface ICartRepository : IRepository
{
	IQueryable<ICartItem> CartItem { get; }
}
```

And our query is:

```c#
public List<TCartItem> GetCartItem(int cartItemID)
{
    using (var cxt = Dependency.Resolve<ICartRepository>()) {
        return cxt.CartItem.FirstOrDefault(x => x.CartItemID == cartItemID);
    }
}
```

So far, so good.  Now we get to the actual client implementation:

```c#
public class OrderItem : ICartItem{
    public int CartItemID {get;set;} // ICartItem
    // Other properties...
}
```

If the `CartItemID` property is the same as a mapped database name...

```c#
fluent.Entity<OrderItem>().Property(x => x.CartItemID).HasColumnName("CartItemID");
```

...everything works fine.

But this is not usually the case.  Rather, the client database should be at liberty to create _their own database structure_, which we then map to our common interface like this:

```c#
public class OrderItem : ICartItem{
    public int OrderItemID {get;set;}
    public int CartItemID {get { return this.OrderItemID; } set { this.OrderItemID = value; } // ICartItem
    // Other properties...
}
```
and
```c#
fluent.Entity<OrderItem>().Property(x => x.OrderItemID).HasColumnName("OrderItemID");
```

In this case, when we try the above query, (`cxt.CartItem.FirstOrDefault(x => x.CartItemID == cartItemID)`), I get an error indicating that `CartItemID` does not belong to the table.

Okay, well not ideal, but I kind of understand why L2DB wouldn't support this.  So, as an interim, I try to map the column to the OrderItemID column:
```c#
fluent.Entity<OrderItem>().Property(x => x.CartItemID).HasColumnName("OrderItemID");
```

Unfortunately, this throws a different error because now _two_ columns are being mapped to the same column name.

Can anybody think of a way I can keep my dream alive?

## System specs
Windows 10
L2DB v 2.4.0
ASP.Net v4.6.1

Thanks guys - much appreciated (great product by the way)

sdanyliv:
Fixes #1365

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
@@ -693,27 +693,34 @@ namespace LinqToDB.Linq.Builder
 
 			if (attr != null)
 			{
-				Expression expr;
+				if (attr.Expression != null)
+					return attr.Expression;
 
-				if (mi is MethodInfo method && method.IsGenericMethod)
+				if (!string.IsNullOrEmpty(attr.MethodName))
 				{
-					var args  = method.GetGenericArguments();
-					var names = args.Select(t => (object)t.Name).ToArray();
-					var name  = string.Format(attr.MethodName, names);
+					Expression expr;
 
-					expr = Expression.Call(
-						mi.DeclaringType,
-						name,
-						name != attr.MethodName ? Array<Type>.Empty : args);
+					if (mi is MethodInfo method && method.IsGenericMethod)
+					{
+						var args  = method.GetGenericArguments();
+						var names = args.Select(t => (object)t.Name).ToArray();
+						var name  = string.Format(attr.MethodName, names);
+
+						expr = Expression.Call(
+							mi.DeclaringType,
+							name,
+							name != attr.MethodName ? Array<Type>.Empty : args);
+					}
+					else
+					{
+						expr = Expression.Call(mi.DeclaringType, attr.MethodName, Array<Type>.Empty);
+					}
+
+					var call = Expression.Lambda<Func<LambdaExpression>>(Expression.Convert(expr,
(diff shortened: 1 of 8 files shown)


Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -693,27 +693,34 @@ LambdaExpression ConvertMethodExpression(Type type, MemberInfo mi)
 
 			if (attr != null)
 			{
-				Expression expr;
+				if (attr.Expression != null)
+					return attr.Expression;
 
-				if (mi is MethodInfo method && method.IsGenericMethod)
+				if (!string.IsNullOrEmpty(attr.MethodName))
```
MaceWindu: is it possible to fail this condition? I see that attribute will have at least one property set: Expresion or MethodName

sdanyliv: Just double check ;)

PR #1390, opened 2018-11-08
Don't call sequence on Insert method

laurentKTR:
Don't call sequence on Insert command. Do this only when calling InsertWithIdentity method.


MaceWindu:
I'm closing this PR as it breaks existing API. If you still have issue with this functionality - please open issue to discuss it.

laurentKTR:
OK , i have done that because in linq2db v1.xx Insert doesn't call Sequence Value, InsertWithIdentity does it
so for me i was a breaking change between v1 and v2

PR #1393, opened 2018-11-10
[WIP] Fix multiple T4/SchemaProvider code generation issues

### LINKED ISSUE #1383: Create shared model generation framework to use with T4 and Linqpad
Right now we have almost the same model generation logic (with some variations) for T4 model generator and Linqpad model builder. We should merge them into single model generation framework and use in both places.

Update:
Right now I see it following way.

Create datamodel generation project (separate nuget) and use it for:
- T4 templates
- Linqpad
- .net core scaffolding tool (like ef.core has already)
- maybe full framework scaffolding tool
- allow users to use it directly

Also we need to add some kind of metadata file (xml file will be fine) to store mappings between database objects and generated names for corresponding code members (classes, methods, properties). It should be used during generation process to generate same names as before.
It will solve serious usability issue we have right now when any fix/change to T4 leads to issues for users when they run T4 template again and it starts generate names differently.

MaceWindu:
Fixes #1383, required for https://github.com/linq2db/linq2db.LINQPad/issues/16

- [x] fixes strange situation when schema provider returns invalid C# identifiers in model and we need to fix them in T4 and LinqPad integration. Now we have method to convert name to valid C# identifier in linq2db and use it for schema names generation
- [x] adds method to properly generate string literals and use it T4
- [x] add more naming tests for T4 templates
- [x] [T4] fix procedure generation for procedures only with output parameters
- [x] removed ConvertToCompilable delegate from T4, because we don't use it anymore with this change

Remaining tasks:
- [ ] I want to try and implement automated tests for T4 templates
- [ ] Still lacks more high-level generation logic for #1384 to share between T4 and LinqPad
- [ ] I think we should remove C# names from schema API as it is out-of-place here

### DIFF
--- a/Data/Create Scripts/SqlServer.sql
+++ b/Data/Create Scripts/SqlServer.sql
@@ -963,3 +963,62 @@ EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'Column descrip
 EXEC sys.sp_addextendedproperty @name=N'MS_Description', @value=N'Index description' , @level0type=N'SCHEMA',@level0name=N'dbo', @level1type=N'TABLE',@level1name=N'Issue1144', @level2type=N'INDEX',@level2name=N'PK_Issue1144'
 
 GO
+
+-- T4/LINQPad naming test-cases
+IF EXISTS (SELECT * FROM sys.objects WHERE type = 'P' AND name = 'proc@t4')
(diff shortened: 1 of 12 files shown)


MaceWindu:
Not actual, remaining issues will be handled by #1383 feature

PR #1405, opened 2018-11-15
[WIP] Improved SQL generation by setting correct aliases.

sdanyliv:
Improved Aliases for SQL queries. 

- Queries become readable and and needs almost nothing for sharing between others.
- Aliases for Associations are now valuable (configurable)
- Added `AliasName` for `AssociationAttribute`.

Sample query before fix:
```sql
SELECT
	[p].[FirstName] as [FirstName1],
	[p].[PersonID] as [PersonID1],
	[p].[LastName] as [LastName1],
	[p].[MiddleName] as [MiddleName1],
	[p].[Gender] as [Gender1]
FROM
	(
		SELECT
			[t1].[PersonID] + 1 as [c1],
			[t1].[FirstName],
			[t1].[PersonID],
			[t1].[LastName],
			[t1].[MiddleName],
			[t1].[Gender]
		FROM
			[Person] [t1]
		WHERE
			[t1].[PersonID] = 1
	) [p]
WHERE
	[p].[c1] + 1 = 3 AND [p].[c1] = 2
```
After fix
```sql
SELECT
	[p_2].[FirstName],
	[p_2].[PersonID],
	[p_2].[LastName],
	[p_2].[MiddleName],
	[p_2].[Gender]
FROM
	(
		SELECT
			[p_1].[PersonID] + 1 as [ID],
			[p_1].[FirstName],
			[p_1].[PersonID],
			[p_1].[LastName],
			[p_1].[MiddleName],
			[p_1].[Gender]
		FROM
			[Person] [p_1]
		WHERE
			[p_1].[PersonID] = 1
	) [p_2]
WHERE
	[p_2].[ID] + 1 = 3 AND [p_2].[ID] = 2
```

To be done:

- [x] Fix tests.
- [x] Show query samples with improvements.
- [ ] Check T4 generation because of changes in `MakeUniqueNames` function.


Source/LinqToDB/Common/Configuration.cs
```diff
@@ -266,5 +266,76 @@ public static bool UseDefaultPolicy
...
+			/// FROM
+			///	   [Child] [child]
+			///       LEFT JOIN [Parent] [A_Parent] ON ([child].[ParentID] = [A_Parent].[ParentID])
+			/// </code>
+			/// </example>
+			/// <remarks>
+			/// Set this value to <c>null</c> to disable special alias generation queries.
+			/// </remarks>
+			/// </summary>
+			public static string AssociationAlias { get; set; } = "A_{0}";
```
igor-tkachev:
```suggestion
			public static string AssociationAlias { get; set; } = "a_{0}";
```

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -266,5 +266,76 @@ public static bool UseDefaultPolicy
...
+			/// Otherwise alias will be removed
+			/// <code>
+			/// SELECT
+			///	   [child].[ChildID]
+			/// FROM
+			///	   [Child] [child]
+			/// </code>
+			/// </example>
+			/// </summary>
+			public static bool GenerateFinalAliases { get; set; } = true;
```
igor-tkachev:
```suggestion
			public static bool GenerateFinalAliases { get; set; } = false;
```

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -1050,7 +1050,7 @@ public void SetAlias(string alias)
 				if (alias == null)
 					return;
 
-				if (alias.Contains('<'))
+				if (!alias.Contains('<'))
```
igor-tkachev: Is it was a bug?

Source/LinqToDB/SqlQuery/SqlStatement.cs
```diff
@@ -296,22 +298,22 @@ public virtual IEnumerable<IQueryElement> EnumClauses()
...
-		IDictionary<string,object> _aliases;
+		HashSet<string> _aliases;
 
 		public void RemoveAlias(string alias)
 		{
 			if (_aliases != null)
 			{
 				alias = alias.ToUpper();
-				if (_aliases.ContainsKey(alias))
+				if (_aliases.Contains(alias))
```
igor-tkachev:
```suggestion

```
You do not need to check out if you use HashSet.

Source/LinqToDB/SqlQuery/SqlStatement.cs
```diff
@@ -354,89 +356,111 @@ internal void SetAliases()
...
+							break;
+						}
 					case QueryElementType.SqlParameter:
 						{
 							var p = (SqlParameter)expr;
 
 							if (p.IsQueryParameter)
 							{
-								if (!objs.ContainsKey(expr))
+								if (!paramsVisited.Contains(expr))
```
igor-tkachev:
```suggestion
								if (!paramsVisited.Add(expr))
```

Source/LinqToDB/SqlQuery/SqlStatement.cs
```diff
@@ -354,89 +356,111 @@ internal void SetAliases()
...
 								{
-									objs.Add(expr, expr);
-									p.Name = GetAlias(p.Name, "p");
+									paramsVisited.Add(p);
```
igor-tkachev:
```suggestion
```

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
sdanyliv: Strange but it will not allow aliasing. For me it looks like a bug because here we are trying do not use anonymous type name but in wrong way.

MaceWindu:
T4 will be fixed as part of #1433

Fruchuxs:
And how I can set it back to the old `t1` beavior to fix my tests?

Fruchuxs:
Checked out and it seems, that the first alias wins.

`queryable.Where(x => x.Count > 5).Where(y => y.Id == 5).Select(v => v.Id);` the alias is x. Seems to work fine with yoda syntax, but looks volatile for the method syntax.

sdanyliv:
> And how I can set it back to the old `t1` behavior to fix my tests?

It will be a problem even in future. It is bad to check generated SQL in tests. Especially when we apply additional optimizations.

Fruchuxs:
It's just an integration test to evaluate the lib on top we built our application. Small changes are not a problem, but this change was hard to understand in the method-syntax and is rather a configuration thing as a optimization for me.

sdanyliv:
We have implemented this feature to simplify understanding queries. Feel free to open issue when we can improve aliasing.  Samples are wellcome.
And yes implementation is closer to linq query syntax.

PR #1410, opened 2018-11-18
Better DbProviderFactories support

### LINKED ISSUE #1385: Add DbProviderFactories support to more providers
This will help to locate installed provider assembly for providers that support provider factory and not referenced directly from user's project.

Right now we use DbProviderFactories to locate provider assembly for SAP HANA provider and starting from 2.5.1 release we will use it for ODP.NET too.

Still there are several providers that could benefit from it:
- DB2
- Informix
- SAP/Sybase ASE

### Environment details
linq2db version: *2.5.0*

MaceWindu:
Fixes #1385
- [x] Move DbProviderFactories support code from SAP/Oracle providers to base DynamicDataProviderBase class
- [x] Enable DbProviderFactories for DB2
- [x] Enable DbProviderFactories for Sybase ASE
- [x] Enable DbProviderFactories for Informix
- [x] Maybe enable DbProviderFactories for SQL CE

Note that it is supported only for desktop framework for now, because we don't build netcoreapp2.1/netstandard2.1 targets yet.

### DIFF
--- a/NuGet/DB2/LinqToDB.DB2.Tools.ttinclude
+++ b/NuGet/DB2/LinqToDB.DB2.Tools.ttinclude
@@ -1,3 +1,2 @@
-<#@ assembly name="$(LinqToDBT4DB2ToolsDirectory)IBM.Data.DB2.dll" #>
 <#@ include file="LinqToDB.Tools.ttinclude" #>
 <#@ include file="LinqToDB.DB2.ttinclude" #>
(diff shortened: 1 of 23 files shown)


MaceWindu:
I left DB2/Informix assemblies in repository, because test classes use them. And switching to GAC references will break test project build for people without those providers installed. So I just removed local copy to test output folder for those.

Issue #1412, opened 2018-11-19
Query wich creates wrong mapping in 2.5 (worked in 2.3)


jogibear9988:
I've the following query:

            var qry = from res in s.Query<WmsLoadCarrierDTO>(showArchivedItems)
                      join type in s.GetTable<WmsResourceTypeDTO>() on res.TypeID equals type.Id into tpList
                      from tp in tpList.DefaultIfEmpty()
                      join r in s.GetTable<WmsResourcePointDTO>() on res.ResourcePointID equals r.Id into pointList
                      from rp in pointList.DefaultIfEmpty()
                      join rss in s.GetTable<RefResourceStorageShelfDTO>() on res.Id equals rss.ResourceID into rssn
                      from rss in rssn.DefaultIfEmpty()
                      join ss in s.GetTable<StorageShelfDTO>() on rss.StorageShelfID equals ss.Id into ssn
                      from ss in ssn.DefaultIfEmpty()
                      join resv in s.GetTable<RefOutfeedTransportOrderResourceDTO>() on res.Id equals resv.ResourceID into reslist
                      select new WmsResourceCombinedDTO() { LoadCarrier = res, ResourceType = tp, ResourcePoint = rp, StorageShelf = ss, Reservations = reslist.Count() };

            return qry;

wich creates the following Mapping Expression 

	.Lambda #Lambda1<System.Func`3[LinqToDB.Linq.IQueryRunner,System.Data.IDataReader,MCC.WMS.ServiceInterfaces.DTO.WmsResourceCombinedDTO]>(
		LinqToDB.Linq.IQueryRunner $qr,
		System.Data.IDataReader $dr) {
		.Invoke (.Lambda #Lambda2<System.Func`6[LinqToDB.Linq.IQueryRunner,LinqToDB.IDataContext,System.Data.IDataReader,System.Linq.Expressions.Expression,System.Object[],MCC.WMS.ServiceInterfaces.DTO.WmsResourceCombinedDTO]>)(
			$qr,
			$qr.DataContext,
			$dr,
			$qr.Expression,
			$qr.Parameters)
	}

	.Lambda #Lambda2<System.Func`6[LinqToDB.Linq.IQueryRunner,LinqToDB.IDataContext,System.Data.IDataReader,System.Linq.Expressions.Expression,System.Object[],MCC.WMS.ServiceInterfaces.DTO.WmsResourceCombinedDTO]>(
		LinqToDB.Linq.IQueryRunner $qr,
[...]

PR #1419, opened 2018-11-20
Fix for #1412. Used first sequence for building expression.

sdanyliv:
I hope this is last commit in this area.
@jogibear9988, please test fix on your sources before next patch release.


jogibear9988:
@sdanyliv @MaceWindu Tested with our sources, code now works again

PR #1424, opened 2018-11-22
Fix for #1423. Added additional key - fields order and their type.

### LINKED ISSUE #1423: Caching of the mapping expression for stored procedure or sql-query
There's a stored procedure:

```
CREATE PROCEDURE TestProc 
	@ReturnFullRow bit = 1
AS
BEGIN
	IF @ReturnFullRow = 1 
	BEGIN
		SELECT 
			1 as Code,
			'123' as Value1,
			'456' as Value2 
	END
	ELSE 
                 SELECT 
                        -12345 as Code
END
GO
```

As you can see, it returns full row or partial row.

A class for mapping:

```
internal class SomeSpResult
{
        public int Code { get; set; }
        public string Value1 { get; set; }
        public string Value2 { get; set; }
}

```
Then I use QueryProc extension to get a partial row result:

` var someSpResults1 = dataConnection.QueryProc<SomeSpResult>("TestProc", new DataParameter("ReturnFullRow", false)).ToList();`

Output:
Code = -12345; Value1 = null; Value2 = null;

Use QueryProc to get full row:

` var someSpResults2 = dataConnection.QueryProc<SomeSpResult>("TestProc").ToList();`

Output:
Code = 1; Value1 = null; Value2 = null;

Expected:
Code = 1; Value1 = "123"; Value2 = "456";

Linq2db cached expression by Mapping Type and query text during the first execution. 
And if the first result was partial row, we continue to get partial-mapped classes, although the procedure returns full row.

sdanyliv:
Fix #1423.


Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -1065,9 +1073,24 @@ public static void ClearObjectReaderCache()
...
+					.Append(';');
+			}
+
+			return sb.ToString();
+		}
+
 		static Func<IDataReader,T> GetObjectReader<T>(DataConnection dataConnection, IDataReader dataReader, string sql)
 		{
-			var key = new QueryKey(typeof(T), dataConnection.ID, sql);
+			var key = new QueryKey(typeof(T), dataConnection.ID, sql, GetFieldsKey(dataReader));
```
igor-tkachev: It can cause some performance issues. So, this should be parameterized. By default it should be applied to sprocs and not applied for regular queries. Also we can add global settings.

sdanyliv: We had hidden problems here not even with stored procs. In test `MySqlTests.TestChar` was bug with mapping because we pass different type of parameters.
It was also bug with `MySqlProvider` but we have to consider that SQL is the same but result type is different.

sdanyliv:
@igor-tkachev, i have limited key generation to only StoredProcs.

Issue #1441, opened 2018-12-04
StackOverflowException on query


epetrukhin:
Following code throws StackOverflowException:

### Steps to reproduce
```c#
using System.Data.SqlClient;
using System.Linq;
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.DataProvider;
using LinqToDB.DataProvider.SqlServer;
using LinqToDB.Mapping;

namespace ConsoleApp1
{
    internal static class Program
    {
        [Table("#Authors")]
        internal sealed class Author
        {
            [PrimaryKey, Column(CanBeNull = false)]
            public int Id { get; set; }

            [Column(CanBeNull = false, DataType = DataType.NVarChar, Length = 100)]
            public string Name { get; set; }
        }

        [Table("#Books")]
        internal sealed class Book
        {
            [PrimaryKey, Column(CanBeNull = false)]
            public int Id { get; set; }

            [Column(CanBeNull = false)]
            public int AuthorId { get; set; }

            [Column(CanBeNull = false, DataType = DataType.NVarChar, Length = 100)]
            public string Title { get; set; }
        }

        private static readonly IDataProvider DataProvider = new SqlServerDataProvider("Test", SqlServerVersion.v2012);

        private static void Main()
        {
            var connectionString = new SqlConnectionStringBuilder
            {
                DataSource         = ".",
                InitialCatalog     = "Test",
                IntegratedSecurity = true
            }.ToString();

            using (var db = new DataConnection(DataProvider, connectionString))
            using (var authorsTable = db.CreateTempTable<Author>())
            using (var booksTable = db.CreateTempTable<Book>())
            {
                IQueryable<Author> authors = authorsTable;

                var onlyWithBooks = true; // Some filter logic

                if (onlyWithBooks)
                {
                    // This causes StackOverflowException
                    authors =
                        from book in booksTable
                        from author in authors.InnerJoin(author => author.Id == book.AuthorId)
                        select author;

                    /* This works fine
                    authors =
                        from author in authors
                        from book in booksTable.InnerJoin(book => book.AuthorId == author.Id)
                        select author;
                    */
                }

                var authorsList = authors.ToList();

                // Consume authors list
            }
        }
    }
}
```

### Environment details
linq2db version: *2.5.2*
Database Server: *Sql Server 2016*
Operating system: *Windows 10*
Framework version: *.NET Framework 4.7.2*

sdanyliv:
[...]

Issue #386, opened 2016-06-18
'l => l.Tags' cannot be converted to SQL. when Join with FreeTextTable 


JCKodel:
This works fine: (joining an existing query to a FreeTextTable, using inner join)

```cs
return from q in query
   join f in db.FreeTextTable<Link, Guid>(l => l.Tags, authenticatedUserTags) on q.Id equals f.Key
   orderby f.Rank descending
   select q;
```

But this don't: (only changed inner join to left join)

```cs
return from q in query
   from f in db.FreeTextTable<Link, Guid>(l => l.Tags, authenticatedUserTags).Where(f => f.Key == q.Id).DefaultIfEmpty()
   orderby f.Rank descending
   select q;
```

Exception:

```
[LinqException: 'l => l.Tags' cannot be converted to SQL.]
   LinqToDB.Linq.Builder.ExpressionBuilder.ConvertToSql(IBuildContext context, Expression expression, Boolean unwrap) in C:\CodeArt\Submodules\linq2db\Source\Linq\Builder\ExpressionBuilder.SqlBuilder.cs:988
   LinqToDB.Linq.Builder.<>c__DisplayClass28_0.<.ctor>b__0(Expression a) in C:\CodeArt\Submodules\linq2db\Source\Linq\Builder\TableBuilder.cs:190
   System.Linq.WhereSelectEnumerableIterator`2.MoveNext() +223
   System.Linq.Buffer`1..ctor(IEnumerable`1 source) +153
   System.Linq.Enumerable.ToArray(IEnumerable`1 source) +106
   LinqToDB.DataProvider.SqlServer.FreeTextTableExpressionAttribute.SetTable(MappingSchema mappingSchema, SqlTable table, MemberInfo member, IEnumerable`1 expArgs, IEnumerable`1 sqlArgs) in C:\CodeArt\Submodules\linq2db\Source\DataProvider\SqlServer\FreeTextTableExpressionAttribute.cs:28

```

But using string, works fine: (removed lambda)

```cs
return from q in query
   from f in db.FreeTextTable<Link, Guid>("Tags", authenticatedUserTags).Where(f => f.Key == q.Id).DefaultIfEmpty()
   orderby f.Rank descending
   select q;
```

ili:
@JCKodel please provide your classes & FreeTextTable method definitions

MaceWindu:
We don't plan to fix this issue as this attribute now obsoleted and will be removed in future releases.
[...]

PR #1442, opened 2018-12-04
Added test for reproducing #1441.

sdanyliv:
Looks like it is Roslyn related bug. They have to evaluate variable value before `SelectMany` call generation.

#1441


MaceWindu:
maybe update to activeissue and merge? also formatting is strange here

MaceWindu:
actually I think it is possible to fix, as it fails because we trigger same expression visiting from expression visiter. I tried to add visited nodes tracking, but it didn't helped, because we use new visitor each time

MaceWindu:
I'm feeling myself stupid :smile: . Actually it looks like the same issue as in #386 where SelectMany call generated with Constant expression parameter instead of Quote. This time I will try to create report for roslyn. If not fixed, at least we will hear reasoning behind this behavior.

MaceWindu:
Test moved to #2170

PR #1452, opened 2018-12-06
Fix postgresql provider "character varying" type bug

vanbukin:
`_npgsqlTypeMap` don't contains `varchar varying` type. 
So I've replace it by `character varying`, that exists [there](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs#L155) (and [in PostgreSQL docs](https://www.postgresql.org/docs/11/datatype-character.html)). Also I've added some additional check (for `character varying` with specified length).


MaceWindu:
omg, thanks. Not sure how it happened

PR #1459, opened 2018-12-10
Table-valued parameters support

### LINKED ISSUE #332: Feature Request: Structured data type parameters support for SQL Server
It would be nice for DataParameter class to support Structured type for SQL Server 2008+.
I see it defined in LinqToDB.DataType enum, but commented.

### LINKED ISSUE #1034: A way to pass sqlparameter with unsupported type to query
Is there a way to pass sqlparameter with type SqlDbType.Structured to a query?
I tried to pass it as an object to DataParameter but got an exception No mapping exists from object type System.Data.SqlClient.SqlParameter to a known managed provider native type.

### LINKED ISSUE #1451: T4 templates. Stored procedure is ignored if "table type" parameter present
I have stored procedure in MS SQL 2016 with user-defined table type parameter.
It looks like this:
``` sql
CREATE TYPE MyTableType AS TABLE
(
	MyTableId INT
)

CREATE PROC MyStoredProc (
	@Ids MyTableType READONLY
)
AS 
BEGIN
    SELECT 1 AS Result
END
```
This proc is ignored by t4 generator because of this condition:
https://github.com/linq2db/linq2db/blob/cdf43a43ce404b8746078c7aa51e3246c2fa94e8/Source/LinqToDB/SchemaProvider/SchemaProviderBase.cs#L258

I've managed to call procedure with this code:
``` csharp
var dataTable = new DataTable();
dataTable.Columns.Add("MyTableId", typeof(int));
dataTable.Rows.Add(1);
List<object> res = this.QueryProc<object>("MyStoredProc", new DataParameter("@Ids", dataTable)).ToList();
```

Is there any chance to support this scenerio in t4 templates?

### Environment details
linq2db version: *2.5.4*
Database Server: *Sql Server 2016*

MaceWindu:
Fix #332, fix #1034, fix #1451

Remaining tasks:
- [ ] fix tableexpression test
- [x] implement schema/T4 support for #1451

Issues and limitations:
- doesn't work with linq service as values doesn't survive serialization
- `DbDataReader`-typed parameters doesn't work, because linq2db prerocess them to lists (?)
- Parameters must be passed as `DataParameter` with `DbType` set for queries, as we need to pass `DbType` to sql server
- Supported parameter value types: `DataTable`, `IEnumerable< SqlDataRecord >` . Also it should be possible to use values that implement `IConvertible` with conversion to `IEnumerable< DbDataRecord >`, but it is not tested


MaceWindu:
Ready for review.
Expected changes:
- tableexpression data parameter support will be fixed @sdanyliv 
- dbtype changes will be replaced with similar changes from #1263 after both PRs merged

PR #1474, opened 2018-12-18
Fixes query cache problems. 

### LINKED ISSUE #1471: How to clear query cache or force a rebuild after FluentMappingBuilder change
I think I need to clear the query cache to fix problem that arises when I use a FluentMappingBuilder to change the mapping in between calls.  For the first call, it always follows the FluentMappingBuilder that I am holding in the MappingSchema that is contained in the data context, but every time after that, it uses the cached query even if the FluentMappingBuilder is changed.

I have confirmed that the data context and its mapping schema inside the query are exactly as they should be.  I can get any of the various correct mappings to execute as long as it is the first call.

I have an IExpressionQuery, but it does not expose a ClearCaches method.  Its internal member Info is a Query<T>, and I could use reflection to get to that ClearCaches method, but I do not like to sneak into internal members.

The cache is quite long lived (lifetime of process?).  Even if you dispose of the DataConnection, and create a new one, it will still find and use that cached query.  Is there a better way I should be clearing the query cache?  Alternatively, is there a way force a new query to be created?

linq2db version: 2.5.4
Database Server: IBM.Data.DB2.iSeries 7.3
Database Provider: LinqToDB.DataProvider.DB2iSeries 2.1.0.0
Operating system: Windows 7
Framework version: .NET Framework 4.5

sdanyliv:
Fixes #1471
QueryRunner for now respects `Configuration.Linq.DisableQueryCache` option.


Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -66,6 +69,9 @@ public MappingSchema(string configuration/* ??? */)
 		/// <param name="schemas">Base mapping schemas.</param>
 		public MappingSchema(string configuration, params MappingSchema[] schemas)
 		{
+			if (configuration.IsNullOrEmpty() && (schemas == null || schemas.Length == 0))
+				configuration = Interlocked.Increment(ref _configurationCounter).ToString();
```
MaceWindu: maybe add some prefix like 'auto_'

sdanyliv: Will do.

PR #1481, opened 2018-12-21
Add comment, sequenceNameAttribute (if different with autogeneration)

KrolikROJER:
Hi,
Can you see my template for generate postgres dto-model, pls? It work for my db with 8th schemas

Added new settings:

```
        public bool IsUpdateSequenceNameAttribute;//Add SequenceNameAttribute if name of sequence is diffrent by autogeneration algoritm
	public bool IsUpdateTableComments;//Add table comment
	public bool IsUpdateColumnComments;//Add column comment
	public bool IsUpdateDateTimeType;//Change all pSql dateTime type to System.DateTime for DTO-classes
	public bool IsUpdateViewIsNotNull;//Change property IsNullable to false for all DTO-view-object
	public bool Is_Id_FirstColumnInView;//Set IsPrimaryKey = true for First property in VDTO-view-object
	public Func<string, string, bool> CanChangeNotNull = (table, column) => true; //Function to accept change nullableAttribute to notnullAttribute
```

Example for use CanChangeNotNull

```
        var nullableColumnView = new Dictionary<string, string[]>
			  {
				  {"tableName", new[]{"column_name"}},
				  {"tableName2", new[]{"column_name_id", "column_name_2"}}
			  };
	CanChangeNotNull = (tableName, columnName) =>
					   {
						   if (nullableColumnView.TryGetValue(tableName, out var clm) &&
							   clm.Contains(columnName))
						   {
							   return false;
						   }

						   return true;
					   };
```


KrolikROJER:
Add support 
`ORDER BY .... DESC NULLS LAST`
to configure:
```
LinqToDB.DataProvider.PostgreSQL.PostgreSQLConfiguration.DescNullsLast = true;
```

MaceWindu:
Thanks for your PR. We cannot accept it in current form as it mixes multiple unrelated features and also queries schema information directly from T4 (instead Schema API should be extended to return mising data).

To help us integrating it, can you split it into several PRs with short description for each feature?

KrolikROJER:
Ok, i'll try

MaceWindu:
We decided to choose another approach to fix sequence names support for T4: https://github.com/linq2db/linq2db/pull/1576#issuecomment-459968037

Please confirm that it will work for you (it should). For other functionality please create separate PRs

MaceWindu:
Closing inactive PRs.

`NULLS LAST` feature request tracked here #2068
For sequence name attribute added new feature request #2766

PR #1491, opened 2018-12-26
Used Microsoft's MemoryCache implementation for purging cached queries.

sdanyliv:
Given implementation should purge items from cache if they are not used more than 1 hour (configurable).


MaceWindu:
already merged as part of #1587

PR #1518, opened 2019-01-09
Asp.net integration

### LINKED ISSUE #1514: Is there easy way to set ConnectionString and ProviderName by json file in Asp.Net Core 2.2?
Is there any way to set `ConnectionString` and `ProviderName` from Json config and easy bind it to DataBaseConnection, Factory, Context and so on?

hahn-kev:
This should allow users to easily configure linq2db for Asp.NET core

In Setup.cs
```C#
var connectionString; //from json or config
var provider;
services.AddLinqToDbContext<IAppContext, AppContext>((provider, options) => {
    options.UseConnectionString(provider, connectionString)
});
```

Now anywhere in your app you can inject `AppContext`. By Default It's injected with Request scope, but you can adjust that when calling Add

One caveat is that you need to declare a constructor that looks like this if you have more than one context
```C#
public AppContext1(LinqToDbConnectionOptions<AppContext1> options) : base(options) { }
```
Otherwise you might get the wrong options injected

In order to get this to work I had to add a reference to `Microsoft.Extensions.DependencyInjection.Abstractions` the version should be checked for compatibility. 

closes #1514

### DIFF
--- a/Source/LinqToDB/Configuration/ServiceConfiguration.cs
+++ b/Source/LinqToDB/Configuration/ServiceConfiguration.cs
@@ -16,6 +16,13 @@ namespace LinqToDB.Configuration
 #if NETSTANDARD2_0
 	public static class ServiceConfiguration
 	{
+		public static IServiceCollection AddLinqToDb(
+			this IServiceCollection serviceCollection,
+			Action<IServiceProvider, LinqToDbConnectionOptions> configure,
+			ServiceLifetime lifetime = ServiceLifetime.Scoped)
+		{
+			return AddLinqToDbContext<DataConnection>(serviceCollection, configure, lifetime);
+		}
 		public static IServiceCollection AddLinqToDbContext<TContext>(
 			this IServiceCollection serviceCollection,
 			Action<IServiceProvider, LinqToDbConnectionOptions> configure,
--- a/Source/LinqToDB/Configuration/ServiceConfiguration.cs
+++ b/Source/LinqToDB/Configuration/ServiceConfiguration.cs
@@ -1,0 +1,212 @@
+using System;
+using System.Data;
+using System.Reflection;
+using LinqToDB.Data;
+using LinqToDB.DataProvider;
+using LinqToDB.Extensions;
+using LinqToDB.Mapping;
+
+#if NETSTANDARD2_0
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+#endif
+
+namespace LinqToDB.Configuration
+{
+#if NETSTANDARD2_0
+	public static class ServiceConfiguration
+	{
+		public static IServiceCollection AddLinqToDbContext<TContext>(
+			this IServiceCollection serviceCollection,
+			Action<IServiceProvider, LinqToDbConnectionOptions> configure,
+			ServiceLifetime lifetime = ServiceLifetime.Scoped) where TContext : IDataContext
+		{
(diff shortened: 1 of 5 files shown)


hahn-kev:
We could do it like that, but the connection string also needs to be provided. 
In my experience no one hard codes the connection string like in the example from Npgsql.
What I did for a case like that is more like this:
```C#
services.AddLinqToDb((provider, options) => {
    options.UsePostgreSQL(connectionString);
});
```

If I'm understanding you correctly then we already have an options pattern setup. For example you might do this:
```C#
services.AddLinqToDb((provider, options) => {
    options.UsePostgreSQL(connectionString);
    options.UseMappingSchema(new MyCustomMappingScheme());

    options.UseDataProvider(new MyDataProvider());
// or
    options.UseDataProvider(provider.GetService<MyDataProvider>());
});
```

I'm not sure what the use cases are for injecting settings into a dataprovider, but we could setup something to help configure those too.

If you're talking about also using an options pattern like this as a replacement for the settings currently in `LinqToDB.Common.Configuration` then that'll take a lot more work, but I don't see why this couldn't be first steps to doing that. Would this be a breaking change compared to the current version with those options being static?

sdanyliv:
We have to think about options, it is a pain and there are several issues with static options. 

What i see from your current implementation - we can do that as an additional package without changing current library. For example `linq2db.AspNetCore`.

Let me think today later and we will discuss about this internally.

MaceWindu:
Looked briefly at this PR and have some comments:
- did you use different line ending/tabsVSspaces settings? Github shows too much changes and it is hard to review. we have .editorconfig file with tabls/lineend options set
- I don't see how it is possible to specify sql dialect version for providers that support it: sql server, db2, postgresql
- also some providers have additional options. I think I have all of them documented here: https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Configuration/IDataProviderSettings.cs#L51. Actually it contains dialect support for sql server/db2, so only pgsql is configured differently, but it makes sense to just add support for it to pgsql

hahn-kev:
@MaceWindu Thanks for the review. 

It looks like both of those files were committed with CRLF line ends, but the files were commited as LF by git using the autocrlf setting (generally recommended to commit with LF). If that's what you want then it should be enforced with both `.editorconfig` and a `.gitattributes` file otherwise most peoples git setups will probably try to convert CRLF ends to LF. 

If you wanted to change the dialect version then we could add that as either an overload to call like this:
```C#
options.UseSqlServer("connection string", SqlServerVersion.v2012);
```
This would work by switching the provider name being used. Or what they could do right now is this
```C#
options.UseConnectionString(new SqlServerDataProvider("my name", SqlServerVersion.v2012), "my connection string");
```
That would give us a lot more options.

The only thing we wouldn't be able to do would be share the same `DataProviders` between multiple different Connections. I'm not sure what the use cases are though, so we could design it to do that if needed.

Looking at the settings I don't see how we could set those per Provider as it just effects which provider gets created, which we could already solve as mentioned above. What I've already done could be even better seeing as how we can add extra settings to the overloads, instead of passing settings by string value like we are with `IEnumerable<NamedValue>`

MaceWindu:
I see. Can you update .gitattributes and fix line endings?

We discussed it a bit and I think best course of actions will be implementing it as separate library (with own nuget), which will allow us to avoid extra dependencies in linq2db and support it in netfx 4.6.1+ without extra target for linq2db itself

sdanyliv:
Superseded by #1814

PR #1520, opened 2019-01-10
Failed when reuse BulkCopyOptions.

igor-tkachev:
v 2.1.0 works fine. 
v 2.6.2 throws an exception when reuse BulkCopyOptions as TableName is cached.


MaceWindu:
As I can see, this change https://github.com/linq2db/linq2db/commit/8e0a3ef261085736ab7edb784b03cecdcf268d5e made TableName non-null on TableT

Which broke this logic
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Data/DataConnectionExtensions.cs#L1439

MaceWindu:
This change https://github.com/linq2db/linq2db/commit/8e0a3ef261085736ab7edb784b03cecdcf268d5e  in v2.2 made [this](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Data/DataConnectionExtensions.cs#L1439) code setting options.TableName on first call

MaceWindu:
I think BulkCopyOptions passed to linq2db should be used as read-only and cloned instead of updated

igor-tkachev:
> I think BulkCopyOptions passed to linq2db should be used as read-only and cloned instead of updated (or just change it to struct from class)

Right. It should not be modified inside BulkCopy methods.

igor-tkachev:
> Found 3 places where we still alter options, passed by user:

Thanks.

Issue #1523, opened 2019-01-10
LinqToDBException for QueryExpressionMethod with hint With("NOLOCK")


arwyl:
Entity:
```
public class SomeEntity
		{
			public int Id { get; set; }

			[Association(QueryExpressionMethod = nameof(OtherImpl), CanBeNull = true)]
			public SomeOtherEntity Other { get; set; }

			private static Expression<Func<SomeEntity, IDataContext, IQueryable<SomeOtherEntity>>> OtherImpl()
			{
				return (e, db) => db.FromSql<SomeOtherEntity>($"dbo.fn_SomeFunction({e.Id})").Take(1);
			}
		}
```

usage:
```
[Test]
		public void Test()
		{
			using (var db = GetDataConnection())
			{
				var q = db.GetTable<SomeEntity>().TableName("SomeTable").With("NOLOCK").Where(x => x.Id == 123);

				Console.WriteLine(q.ToString());
				q.ToArray();
			}
		}
```

throws an Exception "LinqToDB.LinqToDBException : Invalid association query. It is not possible to inline query. Can not find owner table"

arwyl:
@sdanyliv I found out that there was another reason for the bug - With("NOLOCK") hint
will update the text

sdanyliv:
@arwyl, i think i have fixed everything. Could you please check?

arwyl:
@sdanyliv sorry, didn't see your comment. Is the fix on myget already?

arwyl:
@sdanyliv weird.. Everything works in linq2db project tests, but when I copy your test to me and reference dll from linq2db project, I get en exception 
Message: System.InvalidOperationException : variable 'e' of type 'SomeEntity' referenced from scope '', but it is not defined

sdanyliv:
Strange. Please find time to reproduce. Maybe wrong version of `linq2db.dll` is used.

arwyl:
@sdanyliv I redid everything, and the bug went away. Everything is fine now

MaceWindu:
Closed by #1527

PR #1527, opened 2019-01-11
Fix for issue #1523. Improved custom Associations generation.

sdanyliv:
Fix for issue #1523

### DIFF
--- a/Tests/Base/Tools/ComparerBuilder.cs
+++ b/Tests/Base/Tools/ComparerBuilder.cs
@@ -4,7 +4,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
-
+using JetBrains.Annotations;
 using LinqToDB.Expressions;
 using LinqToDB.Extensions;
 using LinqToDB.Linq;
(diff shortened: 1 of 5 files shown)


Tests/Linq/Linq/QueryableAssociationTests.cs
```diff
@@ -222,6 +247,43 @@ select new
 			}
 		}
 
+		static void CreateFunction(DataConnection dc, string tableName)
```
MaceWindu: maybe better add table-less function to sql script and return fixed data set?

Tests/Base/Tools/ComparerBuilder.cs
```diff
@@ -214,4 +214,12 @@ private static Expression RemoveCastToObject(Expression expression)
 			return ((UnaryExpression)expression).Operand;
 		}
 	}
+
+	public static class ComparerBuilder
+	{
+		public static IEqualityComparer<T> GetEqualityComparer<T>([NoEnumeration] IEnumerable<T> enumerable)
```
MaceWindu: :+1:

Tests/Linq/Linq/QueryableAssociationTests.cs
sdanyliv: I think it is better to test in real case scenario.

PR #1562, opened 2019-01-29
Implemented functions for generate CONCAT_WS and STRING_AGG function calls.

### LINKED ISSUE #1559: Linq2Db and string.join()
I am using Linq2db in my big query with subqueries.
At the one place inside it, I want to use string.Join():

    ...
    FullPath = string.Join(" -> ", GetPathQuery(db, c.Id).Select(pi => pi.Name))
    ...
    
But I have received an exception:

> LinqException: 'Join(" -> ", value(RI.DAL.Categories.AdminCategoryPreviewDAL).GetPathQuery(value(RI.DAL.Categories.AdminCategoryPreviewDAL+<>c__DisplayClass4_0).db, c.Id).Select(pi => pi.Name))' cannot be converted to SQL.

I use Postgre SQL and it has the `concat_ws` function which is perfect for me. So I try to use it:

    [Sql.Expression("concat_ws({1}, {0})")]
    public static string JoinAsString(this IQueryable<string> query, string separator)
    {
        return string.Join(separator, query);
    }
    
    ...
    FullPath = GetPathQuery(db, c.Id).Select(pi => pi.Name).JoinAsString(" -> ")
    ...

But I was failed with the same exception.
So how to make it work?

**Environment details**
linq2db version: 2.6.3
Database Server: PostgreSQL 10
Database Provider: standard provider "PostgreSQL.9.5"
Operating system: Windows 10
Framework version: .Net core 2.1

sdanyliv:
Also fixes #1559


MaceWindu:
Aggregation:
- GROUP_CONCAT user by MySQL, SQLITE
- Oracle use LISTAGG : https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions089.htm#SQLRF30030

Concat:
- CONCAT_WS supported by oracle
- missing: sqlite

MaceWindu:
yep, of course

MaceWindu:
fixes included to 2.6.4, new functions will be added in 2.7.0 (we need to improve support for different databases)

MaceWindu:
tested with all databases and applied following fixes:
- enabled tests for all supported databases
- fixed GROUP_CONCAT for mysql
- fixed LISTAGG for oracle to use  WITHIN GROUP(ORDER BY ROWNUM) (not valid without WITHIN clause)
- removed CONCAT_WS support for oracle - not sure why I added it to list of supported databases before
- removed CONCAT_WS sap hana support too

not tested yet:
- sql server 2017 - need to install test database

Also I think it makes sense to implement CONCAT for other unsupported databases in similar way we did here for sqlite

MaceWindu:
SQL Server test for concat fails with:
Message: System.Data.SqlClient.SqlException : The concat_ws function requires 3 to 254 arguments.

MaceWindu:
From other side - users just shouldn't call concat in such cases

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -964,17 +964,23 @@ public T GetAttribute<T>(Type type, MemberInfo memberInfo, bool inherit = true)
...
 			foreach (var c in ConfigurationList)
+			{
 				foreach (var a in attrs)
 					if (configGetter(a) == c)
 						list.Add(a);
+				if (exactForConfiguration && list.Count > 0)
+					break;
+			}
 
 			return list.Concat(attrs.Where(a => string.IsNullOrEmpty(configGetter(a)))).ToArray();
```
MaceWindu: this line still add default configuration attributes even if list is not empty and exact flag set

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -988,17 +994,23 @@ public T[] GetAttributes<T>(Type type, Func<T,string> configGetter, bool inherit
...
```
MaceWindu: same here

PR #1563, opened 2019-01-29
Fix for issue #1560. Improved Join Optimization.

### LINKED ISSUE #1560: Incorrect cross apply for QueryExpressionMethod
Entities
```cs
	public class SomeTableType
	{
	}

	[Table("FewNumberEntity")]
	public class FewNumberEntity
	{
		[Column]
		public int Id { get; set; }
		[Column]
		public int UserId { get; set; }
	}

	[Table("LargeNumberEntity")]
	public class LargeNumberEntity
	{
		[Column]
		public int Id { get; set; }

		[Association(QueryExpressionMethod = nameof(GetSomeValue), CanBeNull = false)]
		public SomeTableType SomeValue { get; set; }

		private static Expression<Func<LargeNumberEntity, IDataContext, IQueryable<SomeTableType>>> GetSomeValue()
		{
			return (t, db) => db.FromSql<SomeTableType>($@"SELECT 
	COUNT(*) as [Value] 
FROM 
	[dbo].[SomeTable] [st] WITH(NOLOCK) 
WHERE 
	[st].[LargeNumberEntityId]={t.Id}");
		}
	}
```

Usage
```cs
[Test]
public void Test()
{
	using (var db = GetDataConnection())
	{
		var tasksQuery = db.GetTable<LargeNumberEntity>();

		var filtered = db.GetTable<FewNumberEntity>().Where(x => x.UserId == 123);

		var final = filtered.InnerJoin(tasksQuery, (x, y) => x.Id == y.Id, (x, y) => y);

		var q = final.Select(x => new
		{
			x.Id,
			x.SomeValue.Value
		});

		Console.WriteLine(q.ToString());
	}
}
```

SQL:
```sql
SELECT
	[y].[Id],
	[a_SomeValue].[Value]
FROM
	[FewNumberEntity] [x]
		INNER JOIN [LargeNumberEntity] [y]
			CROSS APPLY (
				SELECT 

					COUNT(*) as [Value] 

				FROM 

					[dbo].[SomeTable] [st] WITH(NOLOCK) 

				WHERE 

					[st].[LargeNumberEntityId]=[y].[Id]
			) [a_SomeValue]
		ON [x].[Id] = [y].[Id]
WHERE
	[x].[UserId] = 123
```

As we can see, there is join Table cross apply () on ...
and if there are many records in Table, all of them are cross applied

why is the query built this way and not like this:
join Table on ...
cross apply ()
?

this way few records from FewNumberEntity will be joined to LargeNumberEntity and the result (also few records) will be cross applied

### Environment details
linq2db version: *2.6.4*
Operating system: *E.g. Windows 10*

sdanyliv:
Fixes #1560

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -2345,6 +2345,8 @@ namespace LinqToDB.SqlProvider
 			if (text.IsNullOrEmpty())
 				return text;
 
+			text = text.Replace("\r", "");
+
 	        var strArray = text.Split('\n');
 	        var sb = new StringBuilder();
 			for (var i = 0; i < strArray.Length; i++)
--- a/Source/LinqToDB/SqlProvider/JoinOptimizer.cs
+++ b/Source/LinqToDB/SqlProvider/JoinOptimizer.cs
@@ -45,7 +45,7 @@ namespace LinqToDB.SqlProvider
 					for (var si = 0; si < j.Table.Joins.Count; si++)
 					{
 						var sj = j.Table.Joins[si];
-						if ((sj.JoinType == JoinType.Inner || sj.JoinType == JoinType.Left)
+						if ((sj.JoinType == JoinType.Inner || sj.JoinType == JoinType.Left || sj.JoinType == JoinType.CrossApply || sj.JoinType == JoinType.OuterApply)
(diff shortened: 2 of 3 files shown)


sdanyliv:
Unfortunately NO. Join Optimizator introduced flattening for joins which perform faster. But this functionality can be separated from Join Optimizator if it is needed.

PR #1576, opened 2019-02-01
generate postgres sequencename

lobster2012-user:
Hi!
Now sequences are used with standard names. (xxx_seq)

I fixed it for postgres.
1. Added property SequenceName (columnXXX objects)
2. Generating [SequenceName] attribute for postgress from table schema


Source/LinqToDB/SchemaProvider/ColumnSchema.cs
```diff
@@ -22,6 +22,10 @@ public class ColumnSchema
 		/// <summary>
 		/// Gets flag indicating that it is identity column.
```
MaceWindu: wrong xml-doc

MaceWindu:
taking into account #1579 looks like we don't need T4 support for sequences (maybe only in schema api itself)

Issue #1579, opened 2019-02-02
Question. Postgres. Default nextval


lobster2012-user:
Hi!

1. Why is the default value not used for a column with identity type? (checking...)
2. Why is `currval` being called, although you can use the `returning id`?

### Environment details
Database Provider: *Npgsql*

MaceWindu:
I would say - probably because it was implemented like that from start and nobody asked before.

Need to check pgsql documentation, but I suspect default value will not work when:
- it is not configured in database
- user want to use values from other sequence

lobster2012-user:
@MaceWindu 
We can use `default` instead of `nextval` (Which is the default). 
If user wants to use another sequence, then he can override the SequenceNameAttr


In the current implementation, the default sequence with a non-standard name will not be used. Replacing the above would partially solve this problem.
There will be a problem with the generation of the database from the generated model, for this you will still have to extract the sequence name with the schema.

MaceWindu:
makes sense

PR #1580, opened 2019-02-02
Add sequence schema name to query

lobster2012-user:
1.+ Added schema name to **SequenceNameAttribute**
2.+ **Postgres** : using sequence schema name for query 

#1576 
#1579 

```csharp
    [Table(Schema = "public", Name = "test")]
    public partial class Test
    {
        [Column("id"), Identity, SequenceName("custom_seq", Schema = "custom_schema")]
        public long Id { get; set; } // bigint
    }

```

```csharp
     var mapping = new MappingSchema();
     mapping.GetFluentMappingBuilder()
                   .Entity<DataModels.Test>()
                   .Property(e => e.Id)
                   .HasAttribute(new SequenceNameAttribute("custom_seq")
                   {
                        Schema = "custom_schema"
                   });
     db.AddMappingSchema(mapping);                    
```


lobster2012-user:
>     * please revert formatting changes
> 
>     * there is another place in PostgreSQLSqlBuilder.cs where we generate sequence name for nextval

Other providers ?

MaceWindu:
> 
> 
> > ```
> > * please revert formatting changes
> > 
> > * there is another place in PostgreSQLSqlBuilder.cs where we generate sequence name for nextval
> > ```
> 
> Other providers ?

no, I mean GetIdentityExpression method in PostgreSQLSqlBuilder

lobster2012-user:
> no, I mean GetIdentityExpression method in PostgreSQLSqlBuilder


```csharp

                public override ISqlExpression GetIdentityExpression(SqlTable table)
		{
			if (!table.SequenceAttributes.IsNullOrEmpty())
			{
				var attr = GetSequenceNameAttribute(table, false);

				if (attr != null)
				{
					var name = Convert(attr.SequenceName, ConvertType.NameToQueryTable).ToString();
					var database = GetTableDatabaseName(table);
					var schema = attr.Schema != null
						? Convert(attr.Schema, ConvertType.NameToSchema).ToString()
						: GetTableSchemaName(table);

					var sb = BuildTableName(new StringBuilder(), database, schema, name);

					return new SqlExpression($"nextval('{sb}')", Precedence.Primary);
				}
			}

			return base.GetIdentityExpression(table);
		}
```
Sorry, I apologize, I do not see any problems.

MaceWindu:
My fault, dodn't noticed you fixed it too

lobster2012-user:
@MaceWindu 
Check, please.

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLSqlBuilder.cs
```diff
@@ -257,9 +261,11 @@ public override ISqlExpression GetIdentityExpression(SqlTable table)
 
 				if (attr != null)
 				{
-					var name     = Convert(attr.SequenceName, ConvertType.NameToQueryTable).ToString();
+					var name = Convert(attr.SequenceName, ConvertType.NameToQueryTable).ToString();
 					var database = GetTableDatabaseName(table);
-					var schema   = GetTableSchemaName  (table);
+					var schema = attr.Schema != null
+						? Convert(attr.Schema, ConvertType.NameToSchema).ToString()
+						: GetTableSchemaName(table);
```
MaceWindu: can you get rid of '' here too (`return new SqlExpression($"nextval('{sb}')", Precedence.Primary);`)

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLSqlBuilder.cs
```diff
@@ -42,24 +42,28 @@ protected override void BuildCommand(SqlStatement statement, int commandNumber)
 			{
 				var into = insertClause.Into;
 				var attr = GetSequenceNameAttribute(into, false);
-				var name =
+				var sequenceName =
 					attr != null
 						? attr.SequenceName
 						: Convert(
```
MaceWindu: this double Convert call in case of missing attribute looks strange, can you remove this one. It only works, because we don't escape identifiers starting from ", which could change later and break it.

MaceWindu:
Very nice. added to extra change requests.
Also if would be nice to move sequence name build to a separate method to not duplicate this logic, but if it is too hard to understand how to do it, I can do it myself later.

lobster2012-user:
> 
> 
> Very nice. added two extra change requests.
> Also if would be nice to move sequence name build to a separate method to not duplicate this logic, but if it is too hard to understand how to do it, I can do it myself later.

#1581 
A side effect of this update is the removal of a duplicate.

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLSqlBuilder.cs
lobster2012-user: You're right, but
#1581
A side effect of this update is the removal of this code.

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLSqlBuilder.cs
lobster2012-user: ``` $"nextval"```  still remains

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLSqlBuilder.cs
MaceWindu: I'm blind...

lobster2012-user:
@MaceWindu 
What about "nextval" ?
```csharp
var sb = new StringBuilder();
sb.Append("nextval(");
ValueToSqlConverter.Convert(sb, BuildTableName(new StringBuilder(), database, schema, name).ToStrng());
sb.Append(")");
return new SqlExpression(sb.ToString(), Precedence.Primary);
```


```sql

INSERT INTO "public".test
(
        id,
        q
)
VALUES
(
        nextval('"public".test_id_seq'),
        :Q
)
```

lobster2012-user:
>     * add field with nextval if attribute specified
> 
> 
> you can add some tests to ensure this behavior
Updated.

MaceWindu:
Can you merge latest master here?

lobster2012-user:
> 
> 
> Can you merge latest master here?

@MaceWindu 
Done.

PR #1581, opened 2019-02-02
currval -> returning id

lobster2012-user:
#1576
#1579

Provider: Postgres
Replace **currval -> returning id**

```sql
BeforeExecute
-- testdbDB PostgreSQL.9.5 PostgreSQL
DECLARE @Q Bigint -- Int64
SET     @Q = 10

INSERT INTO "public".test
(
        q
)
VALUES
(
        :Q
)
RETURNING
        id
```

```csharp
var key = await db.InsertWithInt64IdentityAsync(new DataModels.Test { Q = 10});
```


MaceWindu:
will not work with postgresql prior to 8.2, but nobody should use something ancient :)

PR #1587, opened 2019-02-04
Fix for #1585. Improved EntityDescriptor caching.

### LINKED ISSUE #446: Fluent ....HasTableName(..) works only first time (version 1.0.7.4)

### LINKED ISSUE #1585: Performance Issue with current Version (GetEntityDescriptor)
We use linq2db with our own MappingSchema wich we use when we create a DataConnection as a Paremeter.

It seems that the GetEntityDescriptor call always creates a new EntityDescriptor, wich takes most of our time. Descriptors should be cached in MappingSchemaInfo, but maybe that is also always recreated?

See Our perf. Profile:

image
image

jogibear9988:
Fixes #1585
Fixes #446


sdanyliv:
I have prepared fix but it needs merging https://github.com/linq2db/linq2db/pull/1491 first.
For comparing you can review this commit: https://github.com/linq2db/linq2db/pull/1587/commits/9c40dd5bf501c27a260bbb93f9c68561ecbe2c0d

jogibear9988:
For me the fix looks good (if it works ;-) ), but I don't understand why we don't use a Dictionary

sdanyliv:
Because after several days of server work, this dictionary will grab a lot of memory. 
We need some mechanism to purge not needed data. In this case if `EntityDescriptor` is not used 1 hour, it will be purged from cache.

jogibear9988:
But EntityDescriptors normaly will not increase, or am I wrong? Only the DataTypes for wich I created a Mapping. Ore are there other Types wich are stored in the Cache?

sdanyliv:
We are trying to cover many cases. There are users who change MappingSchema dynamically very often and we need to be sure that not needed descriptors are purged.

jogibear9988:
thanks for the info and fix

Source/LinqToDB/Metadata/FluentMetadataReader.cs
```diff
@@ -87,5 +87,10 @@ public void AddAttribute(MemberInfo memberInfo, Attribute attribute)
 		/// <inheritdoc cref="IMetadataReader.GetDynamicColumns"/>
 		public MemberInfo[] GetDynamicColumns(Type type)
 			=> _dynamicColumns.TryGetValue(type, out var dynamicColumns) ? dynamicColumns.Keys.ToArray() : new MemberInfo[0];
+
+		public Type[] GetRegisteredTypes()
```
MaceWindu: xml-doc missing

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1461,32 +1459,64 @@ public StringComparer ColumnNameComparer
...
+		/// <returns>
+		/// Mapping types.
+		/// </returns>
+		public Type[] GetDefinedTypes()
+		{
+			return Schemas.SelectMany(s => s.GetRegisteredTypes()).ToArray();
+		}
+
+		/// <summary>
+		/// Clears EntityDescriptor Cache.
```
MaceWindu: `Cache` -> `cache`

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1461,32 +1459,64 @@ public StringComparer ColumnNameComparer
...
 		/// <param name="type">Mapped type.</param>
 		/// <returns>Mapping descriptor.</returns>
 		public EntityDescriptor GetEntityDescriptor(Type type)
 		{
-			return Schemas[0].GetEntityDescriptor(this, type);
+			var key = new { Type = type, ConfigurationID };
+			var ed = EntityDescriptorsCache.GetOrCreate(key,
+				o =>
+				{
+					o.SlidingExpiration = Configuration.Linq.CacheSlidingExpiration;
```
jogibear9988: Shouldn't we create a extra Timeout For the entity descriptors?

sdanyliv: Maybe, you can introduce it if you feel uncomfortable with current settings. My opinion it should be enough.

jogibear9988:
is there anything wrong, why the tests fail?

jogibear9988:
performance is much better now !

MaceWindu:
Superseeds #1491? Any objections for merge?

jogibear9988:
@sdanyliv could be merged?

PR #1598, opened 2019-02-08
New Attribute for skipping specific values of a column on an insert/update

Franki1986:
New Attribute for skipping specific values of a column on an insert


sdanyliv:
@Franki1986, one hint - never create PR from your master. In this case you stick with one PR forever.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,41 @@
...
+		public SkipValuesOnInsertAttribute(params object[] values)
+		{
+			var valuesToSkip = values;
+			if (valuesToSkip == null)
+			{
+				_valuesToSkip = new HashSet<object>(){null};
+			}
+			else if(valuesToSkip.Length > 0)
+			{
+				_valuesToSkip = new HashSet<object>();
```
sdanyliv: `HashSet` has constructor which accept IEnumerable. It will remove duplicates.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -238,6 +245,14 @@ string IColumnChangeDescriptor.ColumnName
 		/// </summary>
 		public bool           SkipOnInsert    { get; private set; }
 
+		/// <summary>
+		/// Gets whether a column is insertable if the value is not representated.
+		/// This values will affect only insert operations with implicit columns specification like
+		/// <see cref="DataExtensions.Insert{T}(IDataContext, T, string, string, string)"/>
+		/// method and will be ignored when user explicitly specifies value for this column.
+		/// </summary>
+		public IEnumerable<object> SkipValuesOnInsert { get; private set; }
```
sdanyliv: It should be `HashSet`

Source/LinqToDB/Linq/QueryRunner.InsertWithIdentity.cs
```diff
@@ -43,7 +45,19 @@ static Query<object> CreateQuery(IDataContext dataContext, string tableName, str
 					if (field.Value.IsInsertable)
 					{
 						var param = GetParameter(type, dataContext, field.Value);
-
+						if (field.Value.SkipValuesOnInsert != null && field.Value.SkipValuesOnInsert.Any() && param.Expression is MemberExpression mExpr)
```
sdanyliv: Use here `field.Value.ColumnDescriptor.SkipValuesOnInsert` (in other places also)

Source/LinqToDB/Linq/QueryRunner.Insert.cs
```diff
@@ -67,9 +81,9 @@ public static int Query(IDataContext dataContext, T obj, string tableName, strin
 				var type = GetType<T>(obj, dataContext);
 				var key  = new { dataContext.MappingSchema.ConfigurationID, dataContext.ContextID, tableName, databaseName, schemaName, type };
 				var ei   = Common.Configuration.Linq.DisableQueryCache
-					? CreateQuery(dataContext, tableName, databaseName, schemaName, type)
+					? CreateQuery(dataContext, obj, tableName, databaseName, schemaName, type)
```
sdanyliv: Remember in EntityDescriptor that type has columns which may skip values during insert and DO NOT CACHE queries for this type.

Source/LinqToDB/Linq/QueryRunner.InsertWithIdentity.cs
```diff
@@ -43,7 +45,19 @@ static Query<object> CreateQuery(IDataContext dataContext, string tableName, str
 					if (field.Value.IsInsertable)
 					{
 						var param = GetParameter(type, dataContext, field.Value);
-
+						if (field.Value.SkipValuesOnInsert != null && field.Value.SkipValuesOnInsert.Any() && param.Expression is MemberExpression mExpr)
+						{
+							if ((mExpr.Member is PropertyInfo info && info.CanRead) || mExpr.Member is FieldInfo)
```
sdanyliv: Use `field.Value.ColumnDescriptor.GetValue(...)` for getting data.

Source/LinqToDB/Linq/QueryRunner.Insert.cs
Franki1986: I am not sure what to do here ? Only disable globaly the cache ?

sdanyliv: Add property for EntityDescriptor, `DoNotCacheObjectInsertQueries` and set to `true` if type has at least one SkipValuesOnInsert property

Franki1986:
Will handle this monday morning...

sdanyliv:
Also create tests, without tests PR is incomplete.

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -81,6 +81,15 @@ string IEntityChangeDescriptor.DatabaseName
...
 		/// </summary>
 		public bool IsColumnAttributeRequired { get; private set; }
+      
+		/// <summary>
+      /// Gets a value indicating whether insert queries should be cached based on
+      /// <see cref="SkipValuesOnInsertAttribute"/>.
+      /// </summary>
+		public bool DoNotCacheObjectInsertQueries
+		{
+			get => Columns != null && Columns.Any(c => c.SkipValuesOnInsert != null && c.SkipValuesOnInsert.Any());
```
sdanyliv: Calculate once during Init.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -131,6 +132,12 @@ public ColumnDescriptor(MappingSchema mappingSchema, ColumnAttribute columnAttri
 						DataType = a.DataType.Value;
 				}
 			}
+			
+			var skipInsertForValuesAttr = mappingSchema.GetAttribute<SkipValuesOnInsertAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo);
+			if (skipInsertForValuesAttr != null)
+			{
+				SkipValuesOnInsert = skipInsertForValuesAttr.Values;
```
sdanyliv: Do not assign if there is no values. Leave null. It will reduce checks later.

sdanyliv:
1. Add test with `null`.
2. Add FluentMappingBuilder support.

Franki1986:
I added a skip null test localy, but is there any to test that the inserted data is the right? I mean if skip null and the default value is null it is hard to handle if the test is successfull. If I have a local db I would create a nullable column with a default value.

Source/LinqToDB/Linq/QueryRunner.InsertOrReplace.cs
```diff
@@ -53,15 +54,26 @@ static Query<int> CreateQuery(IDataContext dataContext, string tableName, string
 				{
 					if (field.IsInsertable)
 					{
-						if (!supported || !fieldDic.TryGetValue(field, out param))
+						var notSupportedOrNotFound = !supported || !fieldDic.TryGetValue(field, out param);
+						if (notSupportedOrNotFound)
 						{
 							param = GetParameter(type, dataContext, field);
+						}
+						if (field.ColumnDescriptor.SkipValuesOnInsert != null && field.ColumnDescriptor.SkipValuesOnInsert.Any())
```
sdanyliv: Just check for null.

Source/LinqToDB/Linq/QueryRunner.InsertWithIdentity.cs
```diff
@@ -43,7 +45,14 @@ static Query<object> CreateQuery(IDataContext dataContext, string tableName, str
 					if (field.Value.IsInsertable)
 					{
 						var param = GetParameter(type, dataContext, field.Value);
-
+						if (field.Value.ColumnDescriptor.SkipValuesOnInsert != null && field.Value.ColumnDescriptor.SkipValuesOnInsert.Any())
```
sdanyliv: Just check for null.

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -226,6 +232,9 @@ void Init(MappingSchema mappingSchema)
 			foreach (var attr in typeColumnAttrs.Concat(attrs))
 				if (attr.IsColumn)
 					SetColumn(attr, mappingSchema);
+
+			// Check if values could be skipped during insert => disable cache...
+			DoNotCacheObjectInsertQueries = Columns != null && Columns.Any(c => c.SkipValuesOnInsert != null && c.SkipValuesOnInsert.Any());
```
sdanyliv: Just check for null.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace LinqToDB.Mapping
+{
+   /// <summary>  Attribute for skipping specific values to be inserted. </summary>
```
sdanyliv: Correct formatting.

Source/LinqToDB/Mapping/EntityMappingBuilder.cs
```diff
@@ -428,6 +428,36 @@ public EntityMappingBuilder<T> Ignore(Expression<Func<T,object>> func, int order
 				a => a.Configuration);
 		}
 
+		/// <summary>
+		/// Adds option for skipping values on insert for the given property current entity.
+		/// </summary>
+		/// <param name="func">Column member getter expression.</param>
+		/// <param name="values">Values that should be skipped during insert.</param>
+		/// <returns>Returns current fluent entity mapping builder.</returns>
+		public EntityMappingBuilder<T> HasSkipValuesOnInsert(Expression<Func<T, object>> func, params object[] values)
```
sdanyliv: Just suggestion, maybe here better have to have generic method?
`public EntityMappingBuilder<T> HasSkipValuesOnInsert<TV>(Expression<Func<T, TV>> func, params TV[] values)`

Source/LinqToDB/Linq/QueryRunner.Insert.cs
```diff
@@ -1,8 +1,11 @@
 ﻿using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using LinqToDB.Mapping;
```
sdanyliv: Move this using inside Namespace, like `using SqlQuery;`

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		/// <summary>
+		/// Gets or sets mapping schema configuration name, for which this attribute should be taken into account.
+		/// <see cref="ProviderName"/> for standard names.
+		/// Attributes with <c>null</c> or empty string <see cref="Configuration"/> value applied to all configurations (if no attribute found for current configuration).
+		/// </summary>
+		public string Configuration { get; set; }
+
+		public HashSet<object> Values
```
sdanyliv: It can be converted to auto property.

Source/LinqToDB/Mapping/EntityMappingBuilder.cs
Franki1986: I thought the same here, but then I have to cast it to 

`Expression<Func<T, object>> `

for the method

`SetAttribute`

or create a new method 

`internal EntityMappingBuilder<T> SetAttribute<TA, TV>(Expression<Func<T, TV>>`

Waht is ok for you?

Source/LinqToDB/Linq/QueryRunner.Insert.cs
Franki1986: Removed unnecessary using.

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -238,6 +245,14 @@ string IColumnChangeDescriptor.ColumnName
 		/// </summary>
 		public bool           SkipOnInsert    { get; private set; }
 
+		/// <summary>
+		/// Gets whether a column is insertable if the value is not representated.
+		/// This values will affect only insert operations with implicit columns specification like
+		/// <see cref="DataExtensions.Insert{T}(IDataContext, T, string, string, string)"/>
+		/// method and will be ignored when user explicitly specifies value for this column.
+		/// </summary>
+		public HashSet<object> SkipValuesOnInsert { get; private set; }
```
MaceWindu: Maybe something like:
```
Gets collection of values for which column should be skipped on non-LINQ insert operation like DataExtensions.Insert, DataExtensions.InsertWithIdentity or DataExtensions.InsertOrUpdate. Returns <c>null</c> if skip is not configured.
```

Source/LinqToDB/Mapping/EntityMappingBuilder.cs
```diff
@@ -428,6 +428,36 @@ public EntityMappingBuilder<T> Ignore(Expression<Func<T,object>> func, int order
 				a => a.Configuration);
 		}
 
+		/// <summary>
+		/// Adds option for skipping values on insert for the given property current entity.
```
MaceWindu: I think `for column on current entity` will sound more like eng[l|r]ish

Source/LinqToDB/Mapping/EntityMappingBuilder.cs
```diff
@@ -428,6 +428,36 @@ public EntityMappingBuilder<T> Ignore(Expression<Func<T,object>> func, int order
...
+		{
+			return SetAttribute(func,
+			                    true,
+			                    _ => new SkipValuesOnInsertAttribute(values) { Configuration = Configuration },
+			                    (_, a) => { },
+			                    a => a.Configuration);
+		}
+
+		/// <summary>
+		/// Adds option for skipping values on insert for the given property current entity.
```
MaceWindu: same `for column on current entity`

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace LinqToDB.Mapping
+{
+	/// <summary>
+	/// Attribute for skipping specific values to be inserted.
```
MaceWindu: `Attribute for skipping specific values on insert.`

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+	/// </summary>
+	[AttributeUsage(
+		AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Class | AttributeTargets.Interface,
+		AllowMultiple = false, Inherited = true)]
+	public class SkipValuesOnInsertAttribute : Attribute
+	{
+      /// <summary>  Default constructor.
```
MaceWindu: formatting looks somewhat off on both constructors

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+
+		/// <summary>
+		/// Gets or sets mapping schema configuration name, for which this attribute should be taken into account.
+		/// <see cref="ProviderName"/> for standard names.
+		/// Attributes with <c>null</c> or empty string <see cref="Configuration"/> value applied to all configurations (if no attribute found for current configuration).
+		/// </summary>
+		public string Configuration { get; set; }
+
+		/// <summary>
+		/// Returns the values that should be skipped during insert.
```
MaceWindu: `Gets collection with values to skip on insert.`

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+		AllowMultiple = false, Inherited = true)]
+	public class SkipValuesOnInsertAttribute : Attribute
+	{
+      /// <summary>  Default constructor. 
+      /// </summary>
+		public SkipValuesOnInsertAttribute() { }
+
+      /// <summary>  Constructor. 
+      /// </summary>
+      /// <param name="values">  The values that should be skipped during insert.
```
MaceWindu: `Values to skip on insert operations.`

Tests/Linq/UserTests/SkipValuesOnInsertTest.cs
```diff
@@ -0,0 +1,205 @@
...
+using System.Linq;
+using LinqToDB.Mapping;
+using NUnit.Framework;
+
+using LinqToDB;
+using LinqToDB.Common;
+
+namespace Tests.UserTests
+{
+	public class SkipValuesOnInsertTest: TestBase
```
MaceWindu: Can you add extra test for column with enum type, mapped to string?

Franki1986: Good point here! I think the way I would expect should not work...

Let's say you have something like this:

```
                public class Person
		{
			public enum GenderType
			{
				[MapValue(null)]
				Undefined,
				[MapValue("Male")]
				Male,
				[MapValue("Female")]
				Female
			}

			[Column("Id"), PrimaryKey]
			public Int32 Id { get; set; }
			[Column("Name")]
			public String Name { get; set; }
			[Column("Age")]
			public Int32? Age { get; set; }
			[Column("Gender"), SkipValuesOnInsert(GenderType.Female)]
			public GenderType? Gender { get; set; }
		}
```

skip value is an enum, the comparisson between the resulting value (string) end the enum does not work, I should do another check, maybe something like this:

```
var value = field.Value.ColumnDescriptor.GetValue(dataContext.MappingSchema, obj);
if (field.Value.ColumnDescriptor.SkipValuesOnInsert.Any(
		v => v == value || (field.Value.ColumnDescriptor.MemberType.GetTypeInfo().IsEnum && v?.ToString() == value?.ToString())))
	{
		continue;
	}
```

Is there any other case where the values that are inserted have another type?

MaceWindu: ToString will not work, because enum value could be mapped to arbitrary string using MapValue attribute. What I wanted to say by this request is that we should use original field value for comparison, not value, converted for db

MaceWindu: Looking at `ColumnDescriptor.GetValue` I think you need to add similar method that use only `MemberAccessor.GetterExpression`

Franki1986: `var value = field.Value.ColumnDescriptor.MemberAccessor.GetValue(obj);`

is this not enough?

MaceWindu: or this. it's basically alias for getter

Franki1986:
@MaceWindu ... maybe it will be better if skip check moved ot attribute's virtual Skip method, so ppl could use inheritance to implement dynamic/custom filtering?...

I am not sure what to do here?
Something like :

```
public bool ShouldSkip(params object[] value)
{
    return Values?.Contains(value) ?? false;
}

```

in the SkipValuesOnInsertAttribute class?
If so I would do that additionally..

MaceWindu:
maybe something like
```
public bool Skip(object value)
{
    return Values.Contains(value);
}
```
and on `ColumnDescriptor` something like that, assigned with `attr.Skip`
```
Func<object, bool> SkipValueOnInsert;
```

MaceWindu:
BTW, custom skip logic is the only way to use this attribute for types, not supported by attributes, without fluent mapper.

Franki1986:
Correct me if I am wrong, but using the SkipValueOnInsert delegate only works if the Attribute is decorated at the column (with fluent or static mapping)?!

MaceWindu:
> Correct me if I am wrong, but using the SkipValueOnInsert delegate only works if the Attribute is decorated at the column (with fluent or static mapping)?!

I don't see how it can work on something else. For some reason you have it allowed for class and interface https://github.com/linq2db/linq2db/pull/1598/files#diff-444d75b7c22a09f81450cdab5e89308fR10 but I don't see any code that supports it on types. I think you need to remove `AttributeTargets.Class | AttributeTargets.Interface`

> If you like I could implement it too..
I would prefer to have it. It will make feature more complete

Source/LinqToDB/Linq/QueryRunner.Insert.cs
```diff
@@ -36,10 +36,18 @@ static Query<int> CreateQuery(IDataContext dataContext, string tableName, string
 
 				foreach (var field in sqlTable.Fields)
 				{
-					if (field.Value.IsInsertable)
+					var skippedValues = field.Value.ColumnDescriptor.SkipValuesOnInsert;
+					if (field.Value.IsInsertable || skippedValues!= null)
```
MaceWindu: nit: add space before !=

Source/LinqToDB/Linq/QueryRunner.Insert.cs
```diff
@@ -36,10 +36,18 @@ static Query<int> CreateQuery(IDataContext dataContext, string tableName, string
...
 				{
-					if (field.Value.IsInsertable)
+					var skippedValues = field.Value.ColumnDescriptor.SkipValuesOnInsert;
+					if (field.Value.IsInsertable || skippedValues!= null)
 					{
+						if (skippedValues != null)
+						{
+							var value = field.Value.ColumnDescriptor.MemberAccessor.Getter(obj);
+							if (field.Value.ColumnDescriptor.SkipValuesOnInsert.Contains(value))
+							{
```
MaceWindu: we should use `skippedValues` variable if we introduced it

Source/LinqToDB/Linq/QueryRunner.InsertOrReplace.cs
```diff
@@ -127,10 +135,11 @@ public static int Query(IDataContext dataContext, T obj, string tableName, strin
 					return 0;
```
MaceWindu: update branch not updated

Source/LinqToDB/Linq/QueryRunner.Update.cs
```diff
@@ -85,11 +97,12 @@ public static int Query(IDataContext dataContext, T obj, string tableName, strin
 					return 0;
 
 				var type = GetType<T>(obj, dataContext);
+				var entityDescriptor = dataContext.MappingSchema.GetEntityDescriptor(obj.GetType());
 				var key  = new { dataContext.MappingSchema.ConfigurationID, dataContext.ContextID, tableName, schemaName, databaseName, type };
-				var ei   = Configuration.Linq.DisableQueryCache
-					? CreateQuery(dataContext, tableName, databaseName, schemaName, type)
+				var ei   = Configuration.Linq.DisableQueryCache || entityDescriptor.DoNotCacheObjectInsertQueries
```
MaceWindu: shouldn't it be `DoNotCacheObjectUpdateQueries` instead of `DoNotCacheObjectInsertQueries`?

Source/LinqToDB/Linq/QueryRunner.Update.cs
```diff
@@ -101,11 +114,12 @@ public static int Query(IDataContext dataContext, T obj, string tableName, strin
...
```
MaceWindu: same here about `DoNotCacheObjectInsertQueries`

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -131,6 +132,19 @@ public ColumnDescriptor(MappingSchema mappingSchema, ColumnAttribute columnAttri
 						DataType = a.DataType.Value;
 				}
 			}
+
+			var skipInsertForValuesAttr = mappingSchema.GetAttribute<SkipValuesOnInsertAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo, attr => attr.Configuration);
+			if (skipInsertForValuesAttr?.Values.Count > 0)
+			{
+				SkipValuesOnInsert = skipInsertForValuesAttr.Values;
```
MaceWindu: why do we need this property if we have check delegate?

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -131,6 +132,19 @@ public ColumnDescriptor(MappingSchema mappingSchema, ColumnAttribute columnAttri
...
+			var skipInsertForValuesAttr = mappingSchema.GetAttribute<SkipValuesOnInsertAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo, attr => attr.Configuration);
+			if (skipInsertForValuesAttr?.Values.Count > 0)
+			{
+				SkipValuesOnInsert = skipInsertForValuesAttr.Values;
+				SkipValueOnInsert = skipInsertForValuesAttr.Skip;
+			}
+			var skipUpdateForValuesAttr = mappingSchema.GetAttribute<SkipValuesOnUpdateAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo, attr => attr.Configuration);
+			if (skipUpdateForValuesAttr?.Values.Count > 0)
+			{
+				SkipValuesOnUpdate = skipUpdateForValuesAttr.Values;
```
MaceWindu: same question here

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -238,6 +252,34 @@ string IColumnChangeDescriptor.ColumnName
...
+		/// Gets collection of values for which column should be skipped.
+		/// This values will affect only insert operations with implicit columns specification like
+		/// <see cref="DataExtensions.Insert{T}(IDataContext, T, string, string, string)"/>
+		/// method and will be ignored when user explicitly specifies value for this column.
+		/// </summary>
+		/// <returns>The collection of values for which column should be skipped or <c>null</c> if skip is not configured.</returns>
+		public HashSet<object> SkipValuesOnInsert { get; private set; }
+
+		/// <summary>  
+      /// Delegate for checking values to be skipped during insert.
```
MaceWindu: formatting is off

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -226,6 +232,10 @@ void Init(MappingSchema mappingSchema)
 			foreach (var attr in typeColumnAttrs.Concat(attrs))
 				if (attr.IsColumn)
 					SetColumn(attr, mappingSchema);
+
+			// Check if values could be skipped during insert/update => disable cache...
+			DoNotCacheObjectInsertQueries = Columns != null && (Columns.Any(c => c.SkipValuesOnInsert != null && c.SkipValuesOnInsert.Any()) ||
+			                                                    Columns.Any(c => c.SkipValuesOnUpdate != null && c.SkipValuesOnUpdate.Any()));
```
MaceWindu: we should differentiate insert/update caching, otherwise update or insert will disable caching for second operation too

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+		/// Attributes with <c>null</c> or empty string <see cref="Configuration"/> value applied to all configurations (if no attribute found for current configuration).
+		/// </summary>
+		public string Configuration { get; set; }
+
+		/// <summary>
+		/// Gets collection with values to skip on insert.
+		/// </summary>
+		public HashSet<object> Values { get; }
+
+      /// <summary>  Check if the passed value should be skipped on insert. </summary>
```
MaceWindu: - formatting
- `Checks ..`
- `The value` -> `Value`

same for update attribute

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,53 @@
...
+      /// <param name="value">   The value that is checked for inserting. </param>
+      /// <returns>  True if value should be skipped on insert. </returns>
```
MaceWindu: `Returns <c>true</c> if ...`
same for update attribute

Source/LinqToDB/Mapping/SkipValuesOnUpdate.cs
```diff
@@ -0,0 +1,53 @@
...
+		/// </summary>
+		public string Configuration { get; set; }
+
+		/// <summary>
+		/// Gets collection with values to skip on update.
+		/// </summary>
+		public HashSet<object> Values { get; }
+
+		/// <summary>  Check if the passed value should be skipped on update. </summary>
+		/// <param name="value">   The value that is checked for updateing. </param>
```
MaceWindu: `for update`

Tests/Linq/UserTests/SkipValuesOnUpdateTest.cs
```diff
@@ -0,0 +1,259 @@
...
+					r = db.GetTable<TestTableFluent>().FirstOrDefault(t => t.Id == 1);
+					Assert.IsNotNull(r);
+					Assert.AreEqual(r.Age, 18);
+					Assert.AreEqual(r.Name, "Jack");
+				}
+			}
+		}
+
+		[Test]
+		public void TestSkipWithEnum([DataSources] string context)
```
MaceWindu: whould be nice to have mixed insert/update skip test

sdanyliv:
Thinking about another design.
```cs
[Flags]
public enum Modification
{
    Insert = 0x1,
    Update = 0x2
}

public class BaseSkipAttribute : Attribute
{
    public abstract bool ShouldSkipValue(object obj, ColumnDescripotr descriptor);
    public abstract Modification Affects { get; }
}
```
And implement `SkipForUpdateAttribute` and `SkipForInsertAttribute`. In this case user can implement it's own attribute and extend logic.

MaceWindu:
image

Franki1986:
I see :-)

Franki1986:
> Thinking about another design.
> 
> ```cs
> [Flags]
> public enum Modification
> {
>     Insert = 0x1,
>     Update = 0x2
> }
> 
> public class BaseSkipAttribute : Attribute
> {
>     public abstract bool ShouldSkipValue(object obj, ColumnDescripotr descriptor);
>     public abstract Modification Affects { get; }
> }
> ```
> 
> And implement `SkipForUpdateAttribute` and `SkipForInsertAttribute`. In this case user can implement it's own attribute and extend logic.

I like this idea, but there are a a few changes to make, could take a little..

sdanyliv:
> I like this idea, but there are a a few changes to make, could take a little..

Sure, there are many changes :)

Franki1986:
@sdanyliv : Why do we need ColumnDescriptor as Attribut-Method parameter?

sdanyliv:
We will add to ColumnDescriptor also `GetMemberValue`, so you will have access to prop/field value without direct `TypeAccessor` usage.

Franki1986:
Ok,

Only to be sure in which direction this should go...

ColumnDescriptor.cs:

... constructor

```
var skipValueAttributes = mappingSchema.GetAttributes<BaseSkipAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo, attr => attr.Configuration);
if (skipValueAttributes != null && skipValueAttributes.Length > 0)
{
	SkipValueAttributes = skipValueAttributes;
}
```

... property:

`private BaseSkipAttribute[] SkipValueAttributes { get; set; }`


.. public method:

```
public virtual bool ShouldSkip(object obj, SkipModification flags)
{
	if (SkipValueAttributes == null || SkipValueAttributes.Length <= 0)
	{
		return false;
	}
			
	for (var i = 0; i < SkipValueAttributes.Length; i++)
	{
		if((SkipValueAttributes[i].Affects & flags) != 0 && SkipValueAttributes[i].ShouldSkip(obj, this))
		{
			return true;
		}
	}
	return false;
}
```

BaseSkipAttribute...:

```
namespace LinqToDB.Mapping
{
	[Flags]
	public enum SkipModification
	{
		Insert = 0x1,
		Update = 0x2
	}
	[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = true)]
	public abstract class BaseSkipAttribute : Attribute
	{
		public abstract bool ShouldSkip(object obj, ColumnDescriptor descriptor);
		public abstract SkipModification Affects { get; }

		/// <summary>
		/// Gets or sets mapping schema configuration name, for which this attribute should be taken into account.
		/// <see cref="ProviderName"/> for standard names.
		/// Attributes with <c>null</c> or empty string <see cref="Configuration"/> value applied to all configurations (if no attribute found for current configuration).
		/// </summary>
		public abstract string Configuration { get; set; }
	}
}

```

sdanyliv:
Add descriptor. Rest is fine.
```cs
 public abstract bool ShouldSkipValue(object obj, EntityDescriptor entityDescriptor, ColumnDescripotr descriptor);
```

Franki1986:
Ok, I am close to it.
Tests for custom skipping logic works fine (like this feature!!).
Monday morning I will push all changes.

Franki1986:
Some of the tests did not pass here with a few providers. But I don't know what could be the problem. My local tests worked, but I have not all providers included...

Franki1986:
@sdanyliv / @MaceWindu 

Do you see some reasons to not pass this in the automatic tests?

sdanyliv:
Not yet. We do not see problems, but looks like it should be already fixed in other PR which is not already merged. Don’t worry.

Source/LinqToDB/Mapping/SkipBaseAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace LinqToDB.Mapping
+{
+	/// <summary>
+	/// Flags for specifying skip modifications used for Attributes based on <see cref="SkipBaseAttribute"/>.
+	/// </summary>
+	[Flags]
+	public enum SkipModification
+	{
+      /// <summary>
```
sdanyliv: Add 
```cs
None = 0x0,
```

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace LinqToDB.Mapping
+{
+	/// <summary>
+	/// Attribute for skipping specific values on insert.
+	/// </summary>
+	public class SkipValuesOnInsertAttribute : SkipBaseAttribute
```
sdanyliv: Create also another abstract class SkipValuesByList and move almost all to this class. There are too many same functionality.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+	{
+		public SkipValuesOnInsertAttribute()
```
sdanyliv: Remove this constructor.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+		/// <param name="entityDescriptor">The entity descriptor.</param>
+		/// <param name="columnDescriptor">The column descriptor.</param>
+		/// <returns><c>true</c> if object should be skipped for the operation.</returns>
+		public override bool ShouldSkip(object obj, EntityDescriptor entityDescriptor, ColumnDescriptor columnDescriptor)
+		{
+			// todo: replace MemberAccessor.Getter() with GetMemberValue
+			return Values?.Contains(columnDescriptor.MemberAccessor.Getter(obj)) ?? false;
+		}
+
+		public override SkipModification Affects { get; }
```
sdanyliv: Rewrite to
```cs
public override SkipModification Affects => SkipModification.Insert;
```

MaceWindu:
any news? :smile:

Franki1986:
Yes, tomorrow ;-) I had some new projects on my table and could not push my commits.
Will do my best tomorrow morning.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
Franki1986: Hope this is what you intended?!

Franki1986:
Do you have still any change requests? If so today I could invest some time... Or could we finalize it?

MaceWindu:
Will review it soon. Probably if there will be only small issues - will fix them myself as I want to look into failing tests issue.

Source/LinqToDB/Linq/QueryRunner.InsertOrReplace.cs
```diff
@@ -130,7 +130,7 @@ public static int Query(IDataContext dataContext, T obj, string tableName, strin
 				var type = GetType<T>(obj, dataContext);
 				var entityDescriptor = dataContext.MappingSchema.GetEntityDescriptor(obj.GetType());
 				var key  = new { dataContext.MappingSchema.ConfigurationID, dataContext.ContextID, tableName, schema, databaseName, type };
-				var ei   = Common.Configuration.Linq.DisableQueryCache || entityDescriptor.DoNotCacheObjectInsertQueries
+				var ei   = Common.Configuration.Linq.DisableQueryCache || (entityDescriptor.SkipModificationFlags & SkipModification.Insert) != 0
```
MaceWindu: Shouldn't we check both insert and update flags here (and below)?
We need tests for cache to ensure that it works properly for all queries and combination of flags.

Source/LinqToDB/Mapping/SkipValuesOnInsertAttribute.cs
```diff
@@ -1,64 +1,21 @@
...
-			else if(valuesToSkip.Length > 0)
-			{
-				Values = new HashSet<object>(values);
-			}
-		}
+		public SkipValuesOnInsertAttribute(params object[] values) : base(values){ }
 
 		/// <summary>
-		/// Gets collection with values to skip on insert.
+		/// Operations that affects skipping a value.
```
MaceWindu: `Operations, affected by value skipping.` sounds better to me

MaceWindu:
Could you add tests for caching and fix insertorupdate? Will review more detaily a bit later, but looks fine to me

Franki1986:
I am not sure what explicitly to test for caching?! I have made a test class for the scenario, but except that I am testing the methods, there is nothing special. A query cache check for a specific key could help...

sdanyliv:
For testing cache, just add to test method `[Values(1, 2, 1)] int value` and check that we have successfully updated data for 2 and skipped 1

Franki1986:
I hope this is what you've intended..

MaceWindu:
Made some changes, tests still fail -will check them tomorrow

MaceWindu:
Found it: https://github.com/linq2db/linq2db/pull/1598/files#diff-cea4ff6fdde15005fab530311e33e309R201

Franki1986:
the nullable int ?

sdanyliv:
@Franki1986, you have changed mapping schema for whole provider, so other tests fail.

Franki1986:
Is it possible to make a null skip test for these providers?

sdanyliv:
Just create new mapping schema and add to DataConnection.

MaceWindu:
Yes, I already did it - you need to pass new schema instance in test and confige it for each test.

MaceWindu:
I broke SkipValuesOnXXX(null) case. Will fix it soon. Don't really like `params object[]` parameters - they are hard to made usable...

MaceWindu:
BTW, I removed fluentmapper overloads with ISet parameters- if you need them back - you need to create tests, because they were passed to attribute as new object[] { HashSet<object> } which is not what you wanted :)

Franki1986:
Thanks for finalizing!!! I think for the end result this is the better way.. We have almost reached over 100 conversation messages :smile:

MaceWindu:
You can work on hashSet support if you want it to be over 100 messages :smile:

Issue #1601, opened 2019-02-12
support linq let reuse?


GonzoSpire:
from http://rsdn.org/forum/prj.rfd/7227580.1 

There is linq query
```cs
from cla in P2P_CALCULATED_CESSION_LOAN_ACCRUE
let cessionStartPlus2 = cla.CessionStart.AddDays(2)
let x = P2P_CALCULATED_CESSION_LOAN_ACCRUE.Sum(y => y.InvestorInterest)
select new
{
    Y1 = x < 0 ? 9 : x + 8,
    Y2 = Math.Round(x + x)
};
```
Which is converted in the following un-optimal query
```sql
SELECT
    CASE
        WHEN (
            SELECT
                Sum([t1].[InvestorInterest])
            FROM
                [dbo].[P2P_CALCULATED_CESSION_LOAN_ACCRUE] [t1]
        ) < 0
            THEN 9
        ELSE (
            SELECT
                Sum([t1].[InvestorInterest])
            FROM
                [dbo].[P2P_CALCULATED_CESSION_LOAN_ACCRUE] [t1]
        ) + 8
    END as [c2],
    (
        SELECT
            Sum([t2].[InvestorInterest])
        FROM
            [dbo].[P2P_CALCULATED_CESSION_LOAN_ACCRUE] [t2]
    ) as [c4]
FROM
    [dbo].[P2P_CALCULATED_CESSION_LOAN_ACCRUE] [t3]

```
But should generate something like this:
```sql
SELECT 
  CASE WHEN t1.x < 0 THEN 9 ELSE t1.x + 8 END as [c2],
    Round(t1.x  + t1.x, 2) as [c4]
FROM (
  SELECT Sum([InvestorInterest]) as x 
     FROM [dbo].[P2P_CALCULATED_CESSION_LOAN_ACCRUE]) [t1]
)
```

MaceWindu:
Actually was fixed long time ago in 2.7.1

PR #1602, opened 2019-02-12
Test for issue #1601. Unoptimal SQL when suing let in linq query.

GonzoSpire:
tests reproduced issue #1601


MaceWindu:
mergin tests

PR #1612, opened 2019-02-19
fix#1513  Add constr for DataContext(providerName, connectionString)

GonzoSpire:
and Tests

### DIFF
--- a/Source/LinqToDB/DataContext.cs
+++ b/Source/LinqToDB/DataContext.cs
@@ -67,10 +67,10 @@ namespace LinqToDB
 			if (providerName == null) throw new ArgumentNullException(nameof(providerName));
 			if (connectionString == null) throw new ArgumentNullException(nameof(connectionString));
 			var dataProvider = DataConnection.GetDataProvider(providerName, connectionString);
-			DataProvider = dataProvider ?? throw new LinqToDBException($"DataProvider '{providerName}' not found.");
-			ContextID = DataProvider.Name;
+			DataProvider     = dataProvider ?? throw new LinqToDBException($"DataProvider '{providerName}' not found.");
+			ContextID        = DataProvider.Name;
 			ConnectionString = connectionString;
-			MappingSchema = DataProvider.MappingSchema;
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/DataContext.cs
```diff
@@ -57,6 +57,24 @@ public DataContext([JetBrains.Annotations.NotNull] IDataProvider dataProvider, [
...
+		/// <summary>
+		/// Creates data context using specified database provider and connection string.
+		/// </summary>
+		/// <param name="providerName">Name of database provider to use with this connection. <see cref="ProviderName"/> class for list of providers.</param>
+		/// <param name="connectionString">Database connection string to use for connection with database.</param>
+		public DataContext(
+			[JetBrains.Annotations.NotNull] string providerName,
+			[JetBrains.Annotations.NotNull] string connectionString)
+		{
+			if (providerName == null) throw new ArgumentNullException(nameof(providerName));
```
sdanyliv: Follow table formatting rules.

Tests/Linq/Linq/DataContextTests.cs
```diff
@@ -71,5 +71,23 @@ public class DataContextTests : TestBase
...
+
+		[Test]
+		public void issue1513Test([DataSources(false)]string context)
+		{
+			using (var db = (TestDataConnection)GetDataContext(context))
+			{
+				var _db = new DataContext(db.DataProvider.Name, db.ConnectionString);
+
+				Assert.AreEqual(db.DataProvider.Name, _db.DataProvider.Name);
+				Assert.AreEqual(db.ConnectionString, _db.ConnectionString);
```
sdanyliv: Better to check fails here by putting wrong provider name and correct connection string, or wrong connection string with real provider name.

GonzoSpire: wrong connection string does not cause exceptions, and should?

sdanyliv: Just select first record of Child

PR #1616, opened 2019-02-22
Made ITable<T> immutable when changing DatabaseName, SchemaName or TableName.

### LINKED ISSUE #1274: ITable<T> extension methods should return new ITable instance instead of modified instance, passed to them
We have some extension methods (need to identify all such methods), that we can apply to ITable<T> instance. User will expect that calling such method on table object will not modify initial object, but return new modified copy.

E.g. of unexpected behavior:
```c#
var table = db.GetTable<SomeTable>();

// here With will add hint to table instance instead of returning new instance
table.With(hint).Select().ToList();

// here we will generate query with table having hint from select query!
table.Set(r => r.Field, someValue).Update();
```

### Environment details
linq2db version: *2.2.0*

sdanyliv:
Fixed dangerous mutation for after calling `LinqExtensions.DatabaseName`, `LinqExtensions.SchemaName` and `LinqExtensions.TableName`.

Fixes #1274


MaceWindu:
Fixes #1274

PR #1617, opened 2019-02-22
server-side function for current timestamp in UTC

### LINKED ISSUE #1577: [Question] How to use UTC time on the server
There is a need to use server-side timestamps for distributed locking and consistency. Difference in the clocks of the machines accessing the database is not tolerable so the best approach is to use the clock of the database machine.

I am aware that there are `Sql.GetDate()` and `Sql.CurrentTimestamp`, however, these properties return the local server time. Is there a way to access the UTC server time (e.g. `getutcdate()` on SQL Server) ?

### Environment details
linq2db version: *2.6.3*
Database Server: *SQL Server 2014*
Framework version: *.NET Core 2.2*

GonzoSpire:
fix #1577 
implement server-side function for current timestamp in UTC format some providers: SqlServer, SQLite, PostgreSQL, MySQL
and tests


Tests/Linq/Linq/DateTimeFunctionsTests.cs
```diff
@@ -32,6 +32,18 @@ public void CurrentTimestamp([DataSources] string context)
...
 		}
 
+		[Test]
+		public void CurrentTimestampUtc(
+			[DataSources(ProviderName.Access, ProviderName.DB2, ProviderName.Firebird, ProviderName.Informix, ProviderName.SqlCe, ProviderName.Oracle)]
+			string context)
+		{
+			using (var db = GetDataContext(context))
+			{
+				var q = from p in db.Person where p.ID == 1 select new { Now = Sql.CurrentTimestampUtc };
```
sdanyliv: I thought that here we need to check that returned DateTime is UTC.

Tests/Linq/Linq/DateTimeFunctionsTests.cs
```diff
@@ -32,6 +32,21 @@ public void CurrentTimestamp([DataSources] string context)
...
+				var res = q.ToList().First().Now;
+				
+				Assert.AreEqual(DateTime.UtcNow.Date, res.Date);
```
MaceWindu: I suspect Assert has proximity-comparison or it could be easily implemented. Those asserts could fail when called around XX:00:00

MaceWindu:
I will check it it later and update with more providers if they support such function

MaceWindu:
Added support for remaining providers. Only providers without support are:
- sql ce
- access
- firebird (required function will be in firebird 4)

for those providers value will be embedded as parameter/literal.



Also tests discovered issue with utc datetime parameter value sent incorrectly over wcf

PR #1618, opened 2019-02-22
Fix for #1546. Fixed problem with projection of nullable values.

### LINKED ISSUE #1546: Selecting properties from nullable associations via expression method in dynamic linq returns default value but not null
Entities
```
[Table()]
public class SomeEntity
{
	[Column()]
	public int Id { get; set; }
	[Column()]
	public int? OtherId { get; set; }

	[Association(ThisKey = "OtherId", OtherKey = "Id", CanBeNull = true)]
	public SomeOtherEntity Other { get; set; }

	[ExpressionMethod("GetThroughIsActual", IsColumn = false)]
	public bool? ThroughIsActual { get; set; }

	private static Expression<Func<SomeEntity, bool?>> GetThroughIsActual()
	{
		return t => Sql.ToNullable(t.Other.IsActual);
	}
}

[Table()]
public class SomeOtherEntity
{
	[Column()]
	public int Id { get; set; }
	[Column()]
	[Column]
	public bool IsActual { get; set; }
}
```
Usage
```
using (var db = new DataConnection(new ValhallaDataProvider(), ConnectionStringUtils.GetTCContextConnectionString(), schema))
{
	var qq = db.GetTable<SomeEntity>().Take(20).Select("new(Id, OtherId, ThroughIsActual)");
	var res = qq.Cast<object>().ToList();

	res.FirstOrDefault().GetType().GetProperties().ToList().ForEach(x => Console.WriteLine($"{x.PropertyType} {x.Name}"));

	Console.WriteLine(res.FirstOrDefault().ToJson();
}
```

Output:
```
System.Int32 Id
System.Nullable`1[System.Int32] OtherId
System.Nullable`1[System.Boolean] ThroughIsActual

  {
    "$type": "DynamicClass1, DynamicClasses",
    "id": 6822,
    "otherId": null,
    "throughIsActual": false
  },
```

As we can see, nullable field Other.IsActual is resolved as false instead of null

sdanyliv:
Long standing problem. It's a time to fix that.
@igor-tkachev, maybe there is much better place to correct reader.

Fix #1546


MaceWindu:
Don't merge yet. I have complex query case which doesn't work with cast to nullable (simple queries work) - will add it tomorrow to see if it is fixed too

MaceWindu:
Added two more failing tests

MaceWindu:
hmm, tests fail

Issue #1555, opened 2019-01-23
SqlServer nvarchar(max) AddScalarType


Kalina9001:
Hi!
Linq2db generate invalid SQL for user type (registered with AddScalarType function) columns with db type  'nvarchar(max)' ( Parameter Length is 4000 )
User Type:
```c#
    [Serializable]
    public class UserType: IConvertible
    {
        public string StringProperty{ get; set; }

        public int IntProperty{ get; set; }

        public UserType()
        {
        }

        public UserType(string stringProperty, int intProperty)
        {
            this.StringProperty = stringProperty;
            this.IntProperty    = intProperty;
        }

        public static UserType CreateFromString(string xml)
        {
            if(string.IsNullOrEmpty(xml))
            {
                return null;
            }

            var reader = new XmlSerializer(typeof(UserType));
            using(var sr = new StringReader(xml))
            {
               
                return (UserType)reader.Deserialize(sr);;
            }
        }

        public static string ConvertToString(UserType item)
        {
            if(item == null)
            {
                return null;
            }

            var writer = new XmlSerializer(typeof(UserType));

            var sb = new StringBuilder();
            using(var sw = new StringWriter(sb, CultureInfo.InvariantCulture))
            {
                using(var xw = XmlWriter.Create(sw, new XmlWriterSettings(){OmitXmlDeclaration = true,NewLineHandling =NewLineHandling.Entitize,Indent =false}))
                {
                    var ns = new XmlSerializerNamespaces();
                    ns.Add(string.Empty, string.Empty);

                    writer.Serialize(xw, item, ns);
                    return sb.ToString();
                }
            }
        }

      

        string IConvertible.ToString(IFormatProvider provider)
        {
            var str = ConvertToString(this);

            return str;
        }

        object IConvertible.ToType(Type conversionType, IFormatProvider provider)
        {
            if(conversionType == typeof(string))
            {
                var str = ConvertToString(this);

                return str;
            }

            return Convert.ChangeType(this, conversionType, provider);;
        }

          TypeCode IConvertible.GetTypeCode()
        {
            return TypeCode.Object;
        }

        bool IConvertible.ToBoolean(IFormatProvider provider)
        {
            throw new NotImplementedException();
        }

        char IConvertible.ToChar(IFormatProvider provider)
        {
            throw new NotImplementedException();
        }

        sbyte IConvertible.ToSByte(IFormatProvider provider)
        {
            throw new NotImplementedException();
        }

[...]

PR #1628, opened 2019-03-02
Parameter size support improvements

### LINKED ISSUE #1578: Optimize sql plans caching with varbinary size
Same issue as #989 _Optimize sql plans caching with varchar\nvarchar size_ but for `varbinary(max)`.

### Environment details
linq2db version: *2.6.3*
Database Server: *SQL Server 2014*
Database Provider: *SQL Server 2012*
Framework version: *.NET Core 2.2*

MaceWindu:
Fixes #1578 

This pr:
- adds length information to DbDataType object and pass it around
- [MSSQL] fixes varbinary parameter size for query to better hit sql query plan cache on subsequential calls with variable-sized parameters

[MSSQL] varbinary, varchar, nvarchar parameter sizing logic is following with this PR:
- if parameter used against column with length set in mapping and both column length and value length do not exceed max supported length (4000/8000 based on type) - use length from mapping
- otherwise if parameter value length does not exceed max supported length (4000/8000 based on type) - use max supported length
- otherwise use unsized parameter

Some details on how parameter size is important for sql server:
https://blogs.msdn.microsoft.com/psssql/2010/10/05/query-performance-and-plan-cache-issues-when-parameter-length-not-specified-correctly/

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -3150,18 +3150,18 @@ namespace LinqToDB.SqlProvider
 				{
 					switch (parameter.DbType)
 					{
-						//case DbType.AnsiString           :
-						//case DbType.AnsiStringFixedLength:
-						//case DbType.String               :
-						//case DbType.StringFixedLength    :
-						//	{
-						//		var value = parameter.Value as string;
+						case DbType.AnsiString           :
+						case DbType.AnsiStringFixedLength:
+						case DbType.String               :
+						case DbType.StringFixedLength    :
+							{
+								var value = parameter.Value as string;
 
-						//		if (!string.IsNullOrEmpty(value))
-						//			sb.Append('(').Append(value.Length).Append(')');
+								if (!string.IsNullOrEmpty(value))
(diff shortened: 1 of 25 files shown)


MaceWindu:
I plan to add fix for #1555 here, as it is closely-related issue

MaceWindu:
I've added fix for #1555 : now linq2db will not make wrong assumptions about parameter size if it's type is not known. Alternatively, we can call IConvertible methods here to get actual value, but it could affect performance, so I would recommend to setup DataParameter convertions instead.

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3150,18 +3150,18 @@ protected virtual void PrintParameterType(StringBuilder sb, IDbDataParameter par
...
-						case DbType.StringFixedLength    :
-							{
-								var value = parameter.Value as string;
-
-								if (!string.IsNullOrEmpty(value))
-									sb.Append('(').Append(value.Length).Append(')');
-
-								break;
-							}
+						//case DbType.AnsiString           :
```
MaceWindu: I've commented size-from-value code, as it doesn't match actual size, sent in `parameter.Size` and just misleading. Could revert it if needed

PR #1639, opened 2019-03-12
issue 1554 enum updating fixed.

zxsanny:
Issue can be reproduced when enum wrapped in object:
object x = SomeEnum.SomeValue;
entity.Set(x => x.EnumProp, x).Update();

### DIFF
--- a/Source/LinqToDB/Expressions/InternalExtensions.cs
+++ b/Source/LinqToDB/Expressions/InternalExtensions.cs
@@ -797,8 +797,13 @@ namespace LinqToDB.Expressions
 
 						if (!ue.Operand.Type.IsEnumEx())
 							return ue.Operand.Unwrap();
-
-						break;
+						else
+						{
+							var ext = ex;
+							while (ext.NodeType == ExpressionType.Convert || ext.NodeType == ExpressionType.ConvertChecked)
+								ext = ((UnaryExpression)ext).Operand;
+							return ext;
+						}
 					}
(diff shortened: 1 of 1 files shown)


MaceWindu:
Can you add tests?

Source/LinqToDB/Expressions/InternalExtensions.cs
```diff
@@ -797,8 +797,13 @@ public static Expression Unwrap(this Expression ex)
 
 						if (!ue.Operand.Type.IsEnumEx())
```
sdanyliv: This one prevents enums to be unwrapped. So, just removing this check should make it work. But i suspect that this fix will break a lot of tests.

zxsanny: @MaceWindu Sure, a bit later
@sdanyliv I'm not sure about removal. My first question was 'why this condition was added?' But I can't figure out this from history, and it was added a long time ago. So I've added else statement

sdanyliv: @zxsanny, add tests and we will see if it is appropriate place to fix. Insert tests also needed. 
Also check this variant `entity.Set(x => x.EnumProp, () => v).Update();`
Blame shows that this code were introduced 7 years ago. So, no one will help in understanding why it is written.

MaceWindu:
Superseeded by #1794

PR #1641, opened 2019-03-13
Tests for reproduce #1202

GonzoSpire:
#1202


MaceWindu:
@sdanyliv  could you check this? looks like reverse null check handling is broken

sdanyliv:
@MaceWindu, will handle this when it become critical.

Issue #1024, opened 2018-02-27
Firebird: should use true/false literals for boolean fields instead of 1/0


gryzunov:
linq2db generates following SQL

``` sql
-- Firebird
SELECT
	t1.ID,
	t1.Name,
	t1.IsDeleted
FROM
	Product t1
WHERE
	1 <> t1.IsDeleted
```
and it results to exception:
```
FirebirdSql.Data.FirebirdClient.FbException
  HResult=0x80004005
  Message=conversion error from string "1"
  Source=FirebirdSql.Data.FirebirdClient
  StackTrace:
   at FirebirdSql.Data.FirebirdClient.FbCommand.Fetch() in C:\Users\Jiri\Documents\devel\NETProvider\working\Provider\src\FirebirdSql.Data.FirebirdClient\FirebirdClient\FbCommand.cs:line 651
   at FirebirdSql.Data.FirebirdClient.FbDataReader.Read() in C:\Users\Jiri\Documents\devel\NETProvider\working\Provider\src\FirebirdSql.Data.FirebirdClient\FirebirdClient\FbDataReader.cs:line 207
   at LinqToDB.Linq.QueryRunner.<ExecuteQuery>d__2b`1.MoveNext() in c:\projects\linq2db\Source\Linq\QueryRunner.cs:line 329
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at ConsoleApp1.Program.Main(String[] args) in c:\users\bambr\documents\visual studio 2017\Projects\ConsoleApp1\ConsoleApp1\Program.cs:line 27
```

### DDL
``` sql
CREATE TABLE PRODUCT (
    ID         INTEGER NOT NULL,
    NAME       VARCHAR(64) NOT NULL,
    ISDELETED  BOOLEAN NOT NULL
);
```

### Code example

```c#
using System.Linq;
using LinqToDB.Data;
using LinqToDB.Mapping;

namespace ConsoleApp1
{
    [Table]
    public class Product
    {
        [PrimaryKey]
        public int ID { get; set; }

        [Column(CanBeNull = false, Length = 64)]
        public string Name { get; set; }

        [Column(CanBeNull = false)]
        public bool IsDeleted { get; set; }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var conn = new DataConnection())
            {
[...]

PR #1646, opened 2019-03-14
Add boolean type support for Firebird

Franki1986:
With the global configuration in FirebirdConfiguration.SupportBooleanType you can control the boolean type Firebird will use, so that a column has real boolean values and not 1 or 0.

This is only a global configuration, so that every boolean value is treated like boolean.

### DIFF
--- a/Source/LinqToDB/DataProvider/Firebird/FirebirdConfiguration.cs
+++ b/Source/LinqToDB/DataProvider/Firebird/FirebirdConfiguration.cs
@@ -22,5 +22,10 @@ namespace LinqToDB.DataProvider.Firebird
 		/// Specifies that Firebird supports literal encoding. Availiable from version 2.5.
 		/// </summary>
 		public static bool IsLiteralEncodingSupported = true;
+
+		/// <summary>
+		/// Specifies that Firebird supports boolean types. Availiable from version 3.0.
+		/// </summary>
+		public static bool SupportBooleanType { get; set; }
 	}
 }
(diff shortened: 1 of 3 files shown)


Franki1986:
Hey guys, what about what I mentioned in the Issue [1024#](https://github.com/linq2db/linq2db/issues/1024#issuecomment-472830127)

MaceWindu:
I would prefer if we can stick to version-specific provider names (like "Firebird.3.0") as we do for postgresql/sqlserver. This will allow us to actually test this change. Global option will break tests against FB <3

Franki1986:
This situation is a bit tricky, because you can use Firebird > Version 3.0 and want to use Smallint to map boolean.

Usaga:
Can this situation be solved by mapping desired type of column?

MaceWindu:
Yes, you can configure mappings yourself. This issue is more about which mapping we should use by default  based on Firebird version.

Usaga:
I think that BOOLEAN should be used by default as it normal way to represent true\false values. It would be very strange to have work-around way (char with '1' & '0') as default and to have explicit configuration (mapping) to force Linq2Db to do it right way (boolean).

Yes, this will break backward compatibility and will force users to migrate schema or explicitly configure mappings to use 'char'. But default behaviour, with BOOLEAN for true\false, will be sane way.

IMHO.

Franki1986:
What are the issues when it comes to auto  recognition for firebird server version. So by default we could check the Version and use true boolean values for versions >= 3.0.
But I think it is important to make it configurable.

MaceWindu:
Superseeded by #2693

PR #1682, opened 2019-04-13
Run each target platform testrun in parallel for appveyor

MaceWindu:
trying to speed-up test run


MaceWindu:
probably will fail due to shared test db

MaceWindu:
This PR tries to reduce appveryor run times, which recently vary from 50+ minutes to 1 hour timeout failure using following changes:
- all 3 target platforms (netfx, core1, core2) run in parallel on their own databases (to avoid dataconflicts)
- nunit integration with tests tab disabled as it takes a lot of time and tests page anyways unusable
- test result xmls attached as artifacts on successfull run
- test runners output attached as log text files to artifacts always. You can check them to see which test failed. This change also fixes main appveyor log trimming due to a lot of logged data from tests

image

MaceWindu:
Example of successful run: https://ci.appveyor.com/project/igor-tkachev/linq2db/builds/23885395/artifacts
Tests failure run: https://ci.appveyor.com/project/igor-tkachev/linq2db/builds/23881429/artifacts

PR #1692, opened 2019-04-20
Get rid of static provider parameter setters

MaceWindu:
We store custom parameter setters in static variable,

Having them as static doesn't give us anything, because each new instance of provider instance always overrides them.

From other side having them static leads to issues when different providers used for same database. E.g. we already removed statics for this reason from sybase provider. Now it gives issues when we run tests with both npgsql3/npgsql4 providers.

### DIFF
--- a/Tests/Linq/TestProviders/Npgsql4PostgreSQLDataProvider.cs
+++ b/Tests/Linq/TestProviders/Npgsql4PostgreSQLDataProvider.cs
@@ -48,7 +48,7 @@ namespace Tests
 #endif
 		}
 
-		private Type _dataReaderType;
+		private  Type _dataReaderType;
 		volatile Type _connectionType;
 
 		private readonly Assembly _assembly;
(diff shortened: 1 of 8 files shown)


Tests/Linq/TestProviders/Npgsql4PostgreSQLDataProvider.cs
```diff
@@ -110,30 +110,30 @@ protected override void OnConnectionTypeCreated(Type connectionType)
...
+			_setMoney     = GetSetParameter(connectionType    , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Money");
 			_setVarBinary = GetSetParameter(connectionType, "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Bytea");
-			_setBoolean = GetSetParameter(connectionType  , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Boolean");
-			_setXml = GetSetParameter(connectionType      , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Xml");
-			_setText = GetSetParameter(connectionType     , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Text");
-			_setBit = GetSetParameter(connectionType      , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Bit");
-			_setHstore = GetSetParameter(connectionType   , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Hstore");
-			_setJson = GetSetParameter(connectionType     , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Json");
-			_setJsonb = GetSetParameter(connectionType    , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Jsonb");
+			_setBoolean   = GetSetParameter(connectionType  , "NpgsqlParameter", "NpgsqlDbType", NpgsqlDbType, "Boolean");
```
sdanyliv: Formatting

PR #1693, opened 2019-04-20
Fix creation of table with identity trigger for oracle native provider

MaceWindu:
Set BindByName property only for commands with parameters for native provider to workaround issue when :NEW token treated by provider as parameter


MaceWindu:
@sdanyliv , @igor-tkachev  ping

PR #1695, opened 2019-04-21
Dynamic store configuration improvements

### LINKED ISSUE #1477: Store dynamic columns outside of current entity
Hello,
is it possible to store the dynamic columns to another dictionary <string, object> that doesn't exist in the entity itself? Our problem is that we have to use a given datastructure but have tables, that differ for each tenant. So we cannot extend the model class with a dictionary and the attribute [DynamicColumnsStore]. Or would it be possible to throw an event when a column ist encountered, that doesnt exist in the given type?

For example:

```cs
namespace OracleTest
{
    public class Customer
    {
        public string Guid { get; set; }
    }

    class Test1
    {

        public  Dictionary<string, object> MetaData { get; set; }
        public string ConnectionString { get; set; }

        public  List<Customer> GetCustomer(string name)
        {
            using (var db = new DataConnection(ProviderName.OracleManaged, ConnectionString))
            {
                Dictionary<string, string> metaDataDict = new Dictionary<string, string>();

                MappingSchema ms = MappingSchema.Default;
                var mb = db.MappingSchema.GetFluentMappingBuilder();
                mb.HasAttribute<Customer>(new TableAttribute("Customer"));
                mb.HasAttribute<Customer>(new ColumnAttribute("name", "name"));
                mb.HasAttribute<Customer>(new ColumnAttribute("guid", "guid"));
                mb.Entity<Customer>().Property(x => Sql.Property<string>(x, "AdditionalColumn"));

                var query = db.GetTable<Customer>().Where(p => p.Name == name);

                db.OnDynamicColumnEvent += OnDynamicColumnHandler;

                return query.ToList(); ;
            }

        }

        private  void OnDynamicColumnHandler(string columnName, object value)
        {
            MetaData.Add(columnName, value);
        }
    }
}
```
Thank you very much.

### LINKED ISSUE #1521: TypeAccessor<T> doesn't respect IMetadataReader
`TypeAccessor<T>` is used in relation to dynamic columns store, however, it doesn't respect custom `IMetadataReader` implementations and tries to find the member using the default `DynamicColumnsStoreAttribute` attribute.

[This is the exact line causing the issue].(https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Reflection/TypeAccessorT.cs#L94)

This matters when I have my own mapping attributes, due to wrapping the library, and the entity has custom columns store defined using a custom attribute rather than `DynamicColumnsStoreAttribute` attribute.

Do you think the `TypeAccessor<T>` should also respect a custom `IMetadataReader`?

MaceWindu:
Fix #1521, fix #1477

- respect `DynamicColumnsStoreAttribute.Configuration` property (before it was using first dictionary with such attribute)
- allow custom storage setter/getter methods using `DynamicColumnAccessorAttribute` mapping entity attribute instead of dictionary
- fluent mapper: new method `DynamicColumnsStore` to configure store dictionary
- fluent mapper: new method `DynamicPropertyAccessors` to configure store getters/setters
- fix type accessor to use current mapping schema instead of default for default values

### DIFF
--- a/Source/LinqToDB/Expressions/MappingExpressionsExtensions.cs
+++ b/Source/LinqToDB/Expressions/MappingExpressionsExtensions.cs
@@ -18,36 +18,37 @@ namespace LinqToDB.Extensions
 			if (members.Length > 1)
 				throw new LinqToDBException($"Ambiguous members '{memberName}' for type '{type.Name}' has been found");
 
-			if (members[0] is PropertyInfo propInfo)
+			switch (members[0])
 			{
-				var value = propInfo.GetValue(null, null);
-				if (value == null)
-					return null;
+				case PropertyInfo propInfo:
+					{
+						var value = propInfo.GetValue(null, null);
+						if (value == null)
+							return null;
 
-				if (value is TExpression expression)
-					return expression;
+						if (value is TExpression expression)
+							return expression;
 
-				throw new LinqToDBException($"Property '{memberName}' for type '{type.Name}' should return expression");
+						throw new LinqToDBException($"Property '{memberName}' for type '{type.Name}' should return expression");
+					}
+				case MethodInfo method:
+					{
+						if (method.GetParameters().Length > 0)
+							throw new LinqToDBException($"Method '{memberName}' for type '{type.Name}' should have no parameters");
+
+						var value = method.Invoke(null, Array<object>.Empty);
+						if (value == null)
+							return null;
+
+						if (value is TExpression expression)
+							return expression;
+
+						throw new LinqToDBException($"Method '{memberName}' for type '{type.Name}' should return expression");
+					}
+				default:
+					throw new LinqToDBException(
+						$"Member '{memberName}' for type '{type.Name}' should be static property or method");
 			}
-			else
-			{
-				if (members[0] is MethodInfo method)
-				{
-					if (method.GetParameters().Length > 0)
-						throw new LinqToDBException($"Method '{memberName}' for type '{type.Name}' should have no parameters");
-					var value = method.Invoke(null, Array<object>.Empty);
-					if (value == null)
-						return null;
-
-					if (value is TExpression expression)
-						return expression;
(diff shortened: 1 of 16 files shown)


Source/LinqToDB/Expressions/MappingExpressionsExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+			if (members.Length == 0)
+				throw new LinqToDBException($"Static member '{memberName}' for type '{type.Name}' not found");
+
+			if (members.Length > 1)
+				throw new LinqToDBException($"Ambiguous members '{memberName}' for type '{type.Name}' has been found");
+
+			var propInfo = members[0] as PropertyInfo;
+
+			if (propInfo != null)
```
sdanyliv: Better to use pattern matching here

Source/LinqToDB/Expressions/MappingExpressionsExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+				if (value is TExpression expression)
+					return expression;
+
+				throw new LinqToDBException($"Property '{memberName}' for type '{type.Name}' should return expression");
+			}
+			else
+			{
+				var method = members[0] as MethodInfo;
+				if (method != null)
```
sdanyliv: And here

Source/LinqToDB/Reflection/MemberAccessor.cs
```diff
@@ -336,6 +306,13 @@ void SetExpressions()
 			}
 		}
 
+		static MethodInfo _throwOnDynamicStoreMissingMethod = typeof(MemberAccessor).GetMethod(nameof(ThrowOnDynamicStoreMissing), BindingFlags.Static | BindingFlags.NonPublic);
```
sdanyliv: MethodOf should be better here.

PR #1699, opened 2019-04-24
Extension point adds possibility to change implementation of DbCommand methods ExecuteScalar, ExecuteNonQuery, ExecureReader

avtc:
Extension point adds possibility to change implementation of `DbCommand` methods `ExecuteScalar`, `ExecuteNonQuery`, `ExecureReader` and their Async equivalents.

One of possible use cases: put commands into queue and initiate them in special separate thread, to overcome lock contention in `TimerQueue.Timer`

See: https://github.com/avtc/timer.noLock

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.Async.cs
+++ b/Source/LinqToDB/Data/DataConnection.Async.cs
@@ -4,6 +4,7 @@ using System.Data.Common;
 using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
+using LinqToDB.Data.DbCommandProcessor;
 
 namespace LinqToDB.Data
 {
@@ -213,7 +214,7 @@ namespace LinqToDB.Data
 		internal async Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
 		{
 			if (TraceSwitch.Level == TraceLevel.Off || OnTraceConnection == null)
-				return await ((DbCommand)Command).ExecuteNonQueryAsync(cancellationToken);
+				return await ((DbCommand)Command).ExecuteNonQueryExtAsync(cancellationToken);
 
 			var now = DateTime.UtcNow;
 			var sw  = Stopwatch.StartNew();
@@ -232,7 +233,7 @@ namespace LinqToDB.Data
 
 			try
 			{
-				var ret = await ((DbCommand)Command).ExecuteNonQueryAsync(cancellationToken);
+				var ret = await ((DbCommand)Command).ExecuteNonQueryExtAsync(cancellationToken);
 
 				if (TraceSwitch.TraceInfo)
 				{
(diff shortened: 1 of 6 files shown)


avtc:
> Unfortunately I can not accept this PR, because I don't see advantages. linq2db already has `RetryPolicy` in `DataConnection` and you can use this technique to repeat commands.

This PR is not about Retry. 
It is about lock contention overcome.
That PR does not provide lock contention overcome, but provide extension point to allow such prevention, without affecting `linq2db` logic.

Lock contention happen inside of `TimerQueue.Timer` when `DbCommand` executed, somewhere inside of `TdsParserStateObject`.
And can be observed on multi-cpu, multi-core, especially SkyLake servers.

Prevention of lock contention with added extension point implemented there: https://github.com/avtc/timer.noLock

sdanyliv:
Maybe I'm loosing something... And looks like it is something new for us.
Looking into our code and see that we already have `RetryingDbConnection`, and `RetryingDbCommand` (with possible extensions that you need)
Yes, creating `RetryingDbConnection` is hardcoded but it can be easily extended.

BTW, we have to analyze that, new API is always headache.

Also please post all related links with problem that you are trying to solve.
Thanks.

avtc:
1. Lock contention details.

image

Lock contention point can be seen on image above. 
Contention is a result of multiple threads trying to enter `lock()` simultaneously. It is caused by improper implementation of `TimerQueueTimer`, where `lock()` used for every case - `Fire`, `Add`, `Change` etc. `TimerQueueTimer` is used by `TdsParserStateObject` deep inside of `SqlCommand`, and also by `Timer`, `Task.Delay`, `CancellationTokenSource` Timeout, `HttpClient` Timeout, etc. And cannot be changed for existing .NetFramework and .NetCore versions.

On multicore cpu with lots of threads this leads to 100% CPU usage sometimes under high load, where lock contention takes 99% of that usage. While before lock contention take place the CPU usage is about 5-10%.

2. Proposed way to overcome lock contention.

The proposed way to overcome the contention is to initiate `dbCommands` one by one in separate thread, so no (or much less) contention will take place. That was made with help of `TaskCompletionSource` and `ConcurrentQueue` there: https://github.com/avtc/timer.noLock/blob/master/dbCmd.noLock/DbCommandQueueProcessor.cs

For easy usage of single thread for all `DbCommands` in application, there are extensions exist https://github.com/avtc/timer.noLock/blob/master/dbCmd.noLock/DbCommandQueueExtensions.cs

To enable lock contention overcome for `linq2db`, just add to start of application:
`LinqToDB.Data.DbCommandProcessor.DbCommandProcessorExtensions.Instance = new linq2db.dbCmd.noLock.Linq2DbCommandQueueProcessor();`

4. Again about this PR.

The aim of this PR is to add extension point around every call to `DbCommand` in `linq2db`, so they can also be initiated in the same single thread as all other `DbCommands` in application.

There is nothing related with `Retry` in this PR, except adding extension points for `DbCommand` calls everywhere in `linq2db`, including `RetryingDbCommand`.

By default `linq2db` will use `DbCommandDefaultProcessor`, where execution path is same as before this PR.

Nothing broken with adding that extension point.

avtc:
I checked `RetryingDbCommand`, and it does not fit.

MaceWindu:
Sounds quite reasonable. Can you create a test?

avtc:
> Sounds quite reasonable. Can you create a test?

A test for extension point? I think there are enough tests in `linq2db` that already cover the code of executing `DbCommands`.

MaceWindu:
Nevermind :smile:

PR #1714, opened 2019-05-12
Add a template for implementing IEquatable interface

jossik:
Sometimes it is necessary to be able to compare classes of business objects.
I propose one of the ways to implement this functionality.

At first there was an idea to implement only two overloaded methods: `Equals` and `GetHashCode`. To do this, I would have to go through all the properties of each class.

But then I discovered the `СomparerBuilder` class with its method `GetEqualityComparer`, which already does all the work.

If there are any comments or suggestions for improving the implementation, I will be glad to know about them.


sdanyliv:
@jossik, in my opinion it is bad idea to make class as an EqualityComparer by itself. It should be separate class.
Better to generate Equals and GetHashCode, maybe by using CompareBuilder.
Also you can extend partial Class definition with property GenerateEquals (it is possible in T4 and we already use this technique). In this case you will be able to control which classes needs equality overrides.

Source/LinqToDB.Templates/Equatable.ttinclude
```diff
@@ -0,0 +1,101 @@
...
+		cl.SetTree();
+	}
+}
+
+partial class Class
+{
+	public bool IsEquatable;
+}
+
+class EquatableClass : Class
```
sdanyliv: I think this additional class is not needed. You can always set IsEquatable for the any class.

Source/LinqToDB.Templates/LinqToDB.ttinclude
```diff
@@ -269,6 +269,8 @@ void GenerateTypesFromMetadata()
 				aProp.Comment.AddRange(comments);
 		}
 
+		SetClassValue(t, "IsEquatable", true);
```
sdanyliv: Remove this.

Source/LinqToDB.Templates/T4Model.ttinclude
```diff
@@ -1523,6 +1523,14 @@ string ToCamelCase(string name)
 	return name.Substring(0, n).ToLower() + name.Substring(n);
 }
 
+event Action<Class,string,object> SetClassValueAction;
```
sdanyliv: Remove this also. You can always modify your classes after loading  model.

jossik:
Thank you for your answer.
I listened to your advice and replaced IEqualityComparer with IEquatable.
Also added an extended Class property: IsEquatable.

As for associations, I agree.
It will be more correct to compare only scalar properties.
Need to think about it.

Source/LinqToDB.Templates/Equatable.ttinclude
```diff
@@ -0,0 +1,101 @@
...
+									},
+
+								new Method(() => "bool", "Equals",
+									new Func<string>[]
+									{
+										() => $"object obj",
+									},
+									() => new[]
+									{
+										$"if (!(obj is {className}))", "{", "\treturn false;", "}", string.Empty,
```
sdanyliv: Use pattern matching and remove unnecessary parentheses.
```cs
if (!(obj is ClassName other))
    return false;
...
```

Source/LinqToDB.Templates/Equatable.ttinclude
```diff
@@ -0,0 +1,101 @@
...
+				});
+			}
+		}
+
+		cl.SetTree();
+	}
+}
+
+partial class Class
+{
```
sdanyliv: Initialize this property from static option, for example GenerateEquatable.

Source/LinqToDB.Templates/Equatable.ttinclude
jossik: I added this just for using in tests, like EditableProperty of EditableObject.ttinclude.

sdanyliv:
Looks good. But we have to invent how to omit associations and maybe compare only PKs if they are exists.

jossik:
Thanks for the review!
Due to lack of time, I can not fully return to this topic.
I will try to do everything that you wrote.
Not very clear about the new EqualityPropertiesFilter option. Where should it be?

By the way, the test is already there: EquatableTest.tt. Or is it not what you need?

MaceWindu:
- `EqualityPropertiesFilter` - it should be a delegate-typed option to set custom properties filter, so users can implement own logic how to select equality members. E.g. see https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB.Templates/LinqToDB.ttinclude#L15 (in or case it should be `=> true`) 

Using it user can implement something like that for equality on primary key fields (pseudocode):
```cs
EqualityPropertiesFilter = (type, member) => 
{
    // don't generate equality for non-table classes (e.g. data manager class)
    if (type is not entity) return false;
   // return true only for pk members
    return member is primary key;
```

- regarding additional test. First, your test check only limited set of positionve cases and second, it will be usefull to show example how to use your template with data model generation templates.


- another point - I just understood that you don't check that target class is static class, which cannot have interface implementation

MaceWindu:
Please inform me when you are done, so I can wrap this up. We plan to release 2.8 on 25 of July, so better to do it 1-2 weeks before release

jossik:
> Please inform me when you are done, so I can wrap this up. We plan to release 2.8 on 25 of July, so better to do it 1-2 weeks before release

Got it.

jossik:
I correctly understood the task?

The EqualityPropertiesFilter helps to select the necessary properties for comparison in the template, to generate them and use them in the method like: ComparerBuilder.GetEqualityComparer (p => p.Field1, p => p.Field2, etc.)?

MaceWindu:
Right!

jossik:
Done! Everything is ok?

Tests/T4.Model/EquatableTest.tt
```diff
@@ -0,0 +1,21 @@
...
+<#@ include file="..\..\Source\LinqToDB.Templates\LinqToDB.SQLite.ttinclude"        #>
+<#@ include file="..\..\Source\LinqToDB.Templates\Equatable.ttinclude"              #>
+<#
+	NamespaceName = "SQLiteDataContext";
+
+	GenerateAssociationExtensions = true;
+
+	var solutionsPath = Host.ResolveAssemblyReference("$(SolutionDir)");
+	LoadSQLiteMetadata(solutionsPath + @"\Data\", "TestData.sqlite");
+
```
MaceWindu: Please add use of options: comparer field name and custom filter to make sure they are used. E.g. for filter you can additionally filter out some properties with specific name.

Source/LinqToDB.Templates/Equatable.ttinclude
```diff
@@ -0,0 +1,115 @@
...
+										$"return EqualityComparer.GetHashCode(this);"
+									})
+									{
+										IsOverride = true
+									},
+
+								new Method(() => "bool", "Equals",
+									new Func<string>[]
+									{
+										() => $"object obj",
```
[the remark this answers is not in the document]
MaceWindu: Looks like you changed wrong line. Instead of Equals parameter declaration, you edited `if` condition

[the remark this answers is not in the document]
jossik: > Looks like you changed wrong line. Instead of Equals parameter declaration, you edited `if` condition

Omg.

jossik:
> Cool! Will merge it today or tomorrow

Thank you!

PR #1722, opened 2019-05-16
Fix #1721 - CreateTable generation for DateTime2 with Precision.

### LINKED ISSUE #1721: CreateTable: DateTime2 with Precision incorrectly generated
Given a record type with any DateTime2 field with a precision value, as generated by the T4 templates when `GenerateDataTypes` is enabled, the type cannot be used to create a table or a temporary table in any database due to an error in the SQL generation for the type.

### Steps to reproduce

Simple example (reformatted from T4 template output):

    public class DateTime2Test
    {
        [Column(DataType = DataType.DateTime2, Precision = 7), NotNull]
        public DateTime LastUpdate { get; set; } // datetime2(7)
    }

    public static class Test
    {
        public static void TestCreate(IDataContext ctx)
        {
            using (var tmp = ctx.CreateTempTable<DateTime2Test>("tmp"))
            {
            }
        }
    }

Attempting to create this table via `CreateTable` or `CreateTempTable` results in the following exception:

    System.Data.SqlClient.SqlException: 'Incorrect syntax near ')'.'

(The stack trace for this is long, boring and irrelevant.)

The generated query is as follows:

    CREATE TABLE [tmp]
    (
        [LastUpdate] DateTime2(7,) NOT NULL
    )

The bad type specifier is generated by `LinqToDB.SqlProvider.BasicSqlBuilder` method [`BuildDataType`][1] and the behavior is not overridden.  By default `BuildDataType` assumes that having a precision is only the case when there is a size as well.  This does not work for DateTime2 types.

Fix should be a simple override  in `LinqToDB.DataProvider.SqlServer.SqlServer2008Builder` like this (not tested):

    protected override void BuildDataType(SqlDataType type, bool createDbType)
    {
        switch (type.DataType)
        {
            case DataType.DateTime2:
                if (type.Precision > 0)
                {
                    StringBuilder.Append(type.DataType).Append('(').Append(type.Precision).Append(')');
                    return;
                }
                break;
        }

        base.BuildDataType(type, createDbType);
    }

(This is laid out approximately the same as a couple of the other overrides for consistency.)

### Environment details
linq2db version: *2.7.0*
Database Server: *Microsoft Sql Server 2017 (14.0.2002)*
Database Provider: *Native?*

[1]: https://github.com/linq2db/linq2db/blob/bb54f89c5752fcf93d85b9b056b02c4e14265cb4/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs#L2506-2536

Corey-M:
DateTime2 type first available in MS SQL 2008.  Added override to `SqlServer2008SqlBuilder` to correctly generate type string for DateTime2 with Precision.

Test add but not tested, fix confirmed in external program.  Please check this test.

### DIFF
--- a/Tests/Linq/UserTests/Issue1721Tests.cs
+++ b/Tests/Linq/UserTests/Issue1721Tests.cs
@@ -1,0 +1,32 @@
+﻿using LinqToDB;
+using LinqToDB.Mapping;
+using NUnit.Framework;
+using System;
+
+namespace Tests.UserTests
+{
+	[TestFixture]
+	public class Issue1721Tests : TestBase
+	{
+		public class I1721Model
+		{
+			[Column(DataType = DataType.DateTime2, Precision = 7), NotNull]
+			public DateTime BadField { get; set; }
+		}
+
+		[ActiveIssue(1721, Configuration = ProviderName.SqlServer2008)]
+		[Test]
+		public void Issue1721Test([IncludeDataSources(TestProvName.AllSqlServer2008Plus)] string context)
+		{
+			using (var db = GetDataContext(context))
+			{
+				Assert.DoesNotThrow(() =>
+				{
+					using (var temp = db.CreateTempTable<I1721Model>("Issue1721"))
+					{ }
+				}, 
+				"CreateTempTable with `DateTime2(7)` field.");
+			}
+		}
+	}
+}
--- a/Tests/Linq/UserTests/Issue1721Tests.cs
+++ b/Tests/Linq/UserTests/Issue1721Tests.cs
@@ -1,7 +1,11 @@
 ﻿using LinqToDB;
+using LinqToDB.Data;
 using LinqToDB.Mapping;
 using NUnit.Framework;
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace Tests.UserTests
 {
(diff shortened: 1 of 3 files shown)


Corey-M:
Fixes #1721

Corey-M:
OK, I made a bad assumption without checking, assumed that the various `SqlServerXXXXSqlBuilder` versions built off each other.  I'll shift the fix into the root `SqlServerSqlBuilder` class.

Tests/Linq/UserTests/Issue1721Tests.cs
```diff
@@ -0,0 +1,32 @@
...
+	[TestFixture]
+	public class Issue1721Tests : TestBase
+	{
+		public class I1721Model
+		{
+			[Column(DataType = DataType.DateTime2, Precision = 7), NotNull]
+			public DateTime BadField { get; set; }
+		}
+
+		[ActiveIssue(1721, Configuration = ProviderName.SqlServer2008)]
```
sdanyliv: ActiveIssue is not needed if bug fixed.

Corey-M: I was copying from another test.  Will fix.

Corey-M:
Hopefully this is the last change.  It's not super important but not specifying the precision defaults to 7 for `DateTimeOffset`, `DateTime2` and `Time` types if you don't specify.  This should properly handle all the things.

Tests/Linq/UserTests/Issue1721Tests.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		[Test]
+		public void Issue1721Test([IncludeDataSources(TestProvName.AllSqlServer2008Plus)] string context)
+		{
+			using (var db = GetDataContext(context))
+			{
+				Assert.DoesNotThrow(() =>
+				{
+					using (var temp = db.CreateTempTable<I1721Model>("Issue1721"))
+					{ }
```
sdanyliv: I think we have to check here that `LastQuery` contains DataType(Precision) text for every type.

Corey-M: You're right.  And while working on this I found an error that the test runner didn't throw.

Corey-M:
Previous code was falling through to the default code when `Precision` is specified as the default value of `7` for all types.  Fixed that and added testing of generated SQL to ensure that the right values are created.

PR #1756, opened 2019-06-11
[WIP] Feature Eager Loading

sdanyliv:
Long waiting feature, so we are trying to do our best.

**Tasks to do:**
- [x] Improve keys identification for projections
- [x] Associations
- [x] ToArray, ToDictionary, ect. for details in projection
- [x] LoadWith support
- [ ] Compiled queries support
- [x] Complete tests
- [x] Code cleanup
- [x] Fix null check warnings
- [x] Add LoadWith and ThenLoad extensions for IQueryable
- [x] Add Transaction for multiple queries execution.


jogibear9988:
Is also a BackReference to a Master Planed?

Like this:

    class Master
      List<Child> Children
    class Child
     Master Master

And Maybe a Child could have more than one Property with Type of "Master" wich would be referenced by different Id Properties?
I don't know, is smth. like this possible in EF or EFCore?

sdanyliv:
Yes, any kind of associations.

jogibear9988:
Another question...

In Redmine we have a Table with this properties:

        public long Id { get; set; }

        public int TagId { get; set; }

        public int TaggableId { get; set; }

        public string TaggableType { get; set; }

        public DateTime CreatedAt { get; set; }

In this Table the TaggableId References a "Issue" or a "Project" or a few other Entitys depending on a Value inside of "TaggableType". Would it be Possible that we create  Associations to Different Types of Associated DTOs, so if I use a Property "Issue" a Where with "TaggableType == "Issue" is added...

jogibear9988:
Oh wait, I think I could solve this with inheritance...
Hopefully this would also work with Eager loading

detoxhby:
Ran tests against this branch with latest successfully build ( https://ci.appveyor.com/project/igor-tkachev/linq2db/builds/30362878 ).

Everything looking fine, but couldn't really find specific cases where the eager loading would kick in. Group joins are executed separately no matter what, LoadWith()-s produce left joins as always with 1:N associations.

What kind of context/query type should I search for where the new eager loading would have effect?

sdanyliv:
@detoxhby, we replaced N+1 query problem with 1 + 1 queries.
var parentWithChildren = db.Parent.LoadWith(p => p.Children).ToArray();
New code will run only two queries. One for Children and second for Parents.
But current code (master) will run one query for Parents and N queries for every parent to get children.

detoxhby:
> @detoxhby, we replaced N+1 query problem with 1 + 1 queries.
> var parentWithChildren = db.Parent.LoadWith(p => p.Children).ToArray();
> New code will run only two queries. One for Children and second for Parents.
> But current code (master) will run one query for Parents and N queries for every parent to get children.

@sdanyliv, I see, thanks, will check again.

Is group join based N+1 problem target of the current scope?

For ex.:
```
from p in Parent
join c in Child on p.Id equals c.ParentId into children
select new ParentDTO()
{
  X = ...,
  Children = children.Select(x => ChildDTO()
  {
    X = ...
  })
}
```

Logically, it is the same as using LoadWith, but for dynamic system is really hard to enforce the usage of that helper. It would be really benefical to apply the same handler to these kind of constructs if it's possible to detect and formalize them in the proper way.

sdanyliv:
Yes, this kind of queries also. It should run only two queries.
Expand EagerLoadingTests.cs and you will be surprised.
https://github.com/linq2db/linq2db/pull/1756/files#diff-7c00e582a5324bd5cc02f46930233549R200

jogibear9988:
will 2 Loadwith the also produce 2 querys, or will it then be 3?

sdanyliv:
Yes it will produce more queries if you select different child records. It should work in any deep: Children which have DataItems, which have Operations, etc.
```cs
from p in Parent
select new ParentDTO()
{
  X = ...,
  Children = db.Children.Where(c => c.ParentId == p.Id).Select(x => ChildDTO()
  {
     DataItems = db.DataItems.Where(di => di.ChildId == c.Id).Select(di => new DataItemDTO{
     {
       Operation = db.Operation.Where(o => o.DataItemId == di.Id).Select(o => new OperationDTO{
         ...
     })
    }
   )
  })
}
```
Even better, you do not need group join here.

sdanyliv:
Anyway, looks like you did not run tests for this specific branch. You have to checkout and build it yourself by VS 2017+

jogibear9988:
Wouldn't it be better to do it one query, or is this not possible? (Or to complex?)

sdanyliv:
Yes it is possible, BUT.
There are two ways:

1. EF 6 doing that in singe result set. Usually bad performance and server memory usage.
2. EF Core. Creates singe query with several selects and read multiple result sets. It works only for database servers which supports that.

I prefer second one, but we have to do some refactoring before that. Currently we read records by separate requests to database.

jogibear9988:
> Yes it is possible, BUT.
> There are two ways:
> 
> 1. EF 6 doing that in singe result set. Usually bad performance and server memory usage because of Cartesian explosion in queries.
> 2. EF Core 2.1. Creates singe query with several selects and read multiple result sets. It works only for database servers which support that. Otherwise it runs several queries in separately.
> 3. EF Core 3.0 - works as EF 6 in this case. They have did regression again.
> 
> I prefer second one, but we have to do some refactoring before that. Currently we read records by separate requests to database.

I've read now a little bit on this in the EF core issues...
maybe we could support both modes? So user can decide wich one to use? (don't know how hard would it be to add support)

sdanyliv:
> I've read now a little bit on this in the EF core issues...
> maybe we could support both modes? So user can decide wich one to use? (don't know how hard would it be to add support)

It is possible but needs a lot of efforts. Maybe after implementing Defer execution.
Anyway we can do another quirk - generate union of resultsets in one sql and then on the client side separate needed information.
something like this:
```cs
var composed = 
           query1.Select(q1 => new { Rs = 1,  Q1 = q1,   Q2 = null, Q3 = null })
   .Concat(query2.Select(q2 => new { Rs = 2,  Q1 = null, Q2 = q2,   Q3 = null }))
   .Concat(query3.Select(q3 => new { Rs = 3,  Q1 = null, Q2 = null, Q3 = q3   }));
```
I have a lot of ideas, everything needs time for design and implementation.

to11mtm:
Question: does it make sense to have an option for Default transaction behavior if we are doing multiple queries/round trips? I'm thinking of issues like [EF Core #9014](https://github.com/dotnet/efcore/issues/9014) which gives a pretty good example of how multi-queries could impact some usages negatively if you don't have a transaction set.

sdanyliv:
@to11mtm, good catch, we definitely have to expose such configuration and by default start transaction if it is not started yet. But this will not work for RemoteDataContext. I have plans to refactor remoting to send whole expression tree to the server. It will simplify using of linq2db in Blazor applications.

MaceWindu:
наркоман чтоле?

MaceWindu:
[x] SAP HANA tests passed

yarecky1:
@sdanyliv Missing?
>     * [ ]  Compiled queries support

sdanyliv:
Just decided, if it is needed - wait for issue. Otherwise 3.0 version will never happen ;)

PR #1797, opened 2019-07-06
Fix issue #1622

### LINKED ISSUE #1622: ValueToSqlConverter for enums affects simple integers
Hi!
When i define custom value to sql converters for an enum, this enum is treated like its underlying value type (e.g., int) and primitive values of that underlying type are affected. For instance, I define some custom expression which compares some column value with custom enum value: 

```c#
[Table]
public class Entity
{
    [Identity, PrimaryKey]
    public int Id { get; set; }
    [Column]
    public string SomeText { get; set; }
}

public enum SomeEnum
{
    Value1, Value2
}

[Sql.Expression("{0} = {1}", InlineParameters = true, ServerSideOnly = true, IsPredicate = true)]
public static bool SomeComparison(string column, SomeEnum value) => throw new InvalidOperationException();
```
Here I set custom mapping for that enum into SQL: 
```c#
 _db.MappingSchema.SetValueToSqlConverter(typeof(SomeEnum),
                (sb, dt, v) =>
                {
                    sb.Append("'").Append(((SomeEnum)v).ToString()).Append("_suffix'");
                });
```
And finally I try to execute two queries:

```c#
var res = _db.GetTable<Entity>().Where(e => SomeComparison(e.SomeText, SomeEnum.Value1)).ToArray();
var res2 = _db.GetTable<Entity>().Where(e => e.Id == 1).ToArray();
```

The first one is executed as expected:
```sql
SELECT
	[e].[Id],
	[e].[SomeText]
FROM
	[Entity] [e]
WHERE
	[e].[SomeText] = 'Value1_suffix'
```
But, unfortunalely, converter tries to map int to its corresponding enum value and affects the second query too:
```sql
SELECT
	[e].[Id],
	[e].[SomeText]
FROM
	[Entity] [e]
WHERE
	[e].[Id] = 'Value2_suffix'
```

I've made workaround and defined container struct for enum and defined mapping  for that container: 
```c#
public struct SomeEnumContainerStruct
{
    public SomeEnum SomeEnum { get; set; }
}

[Sql.Expression("{0} = {1}", InlineParameters = true, ServerSideOnly = true, IsPredicate = true)]
public static bool SomeComparison(string column, SomeEnumContainerStruct value) => throw new InvalidOperationException();

_db.MappingSchema.SetValueToSqlConverter(typeof(SomeEnumContainerStruct),
                (sb, dt, v) =>
                {
                    sb.Append("'").Append(((SomeEnumContainerStruct)v).SomeEnum.ToString()).Append("_suffix'");
                });
```
This works as expected. By the way, container class (not struct) does not work (converter just does not fire). Is this by design?

Thanks!
### Environment details
linq2db version: _2.6.4_
Database Server: _Sql Server 13.0.1601.5_
Database Provider: _Sql.Data.SqlClient 4.6.27110.4_
Operating system: _Windows 10 Pro 10.0.17134_
Framework version: _.NET Core 2.2.102_

MaceWindu:
Fixes #1622
- fix enums support in ValueToSqlConverter
- adds support for custom ValueToSql conversions on sql builder level
- backport Sql.DateParts enum fixes from 3.0 branch

Known issues:
- fix doesn't work for remote context due to lack of custom mapping schema support on server-side. Already fixed in 3.0 branch

### DIFF
--- a/Source/LinqToDB.Templates/DataModel.ttinclude
+++ b/Source/LinqToDB.Templates/DataModel.ttinclude
@@ -99,7 +99,7 @@ HashSet<string> KeyWords = new HashSet<string>
 
 void LoadServerMetadata(DataConnection dataConnection)
 {
-	SqlBuilder = dataConnection.DataProvider.CreateSqlBuilder();
+	SqlBuilder = dataConnection.DataProvider.CreateSqlBuilder(dataConnection.MappingSchema);
 
 	var sp = dataConnection.DataProvider.GetSchemaProvider();
 	var db = sp.GetSchema(dataConnection, GetSchemaOptions);
--- a/Source/LinqToDB/Data/DataConnection.Linq.cs
+++ b/Source/LinqToDB/Data/DataConnection.Linq.cs
@@ -79,12 +79,7 @@ namespace LinqToDB.Data
 
 		string IDataContext.ContextID => DataProvider.Name;
 
-		static Func<ISqlBuilder> GetCreateSqlProvider(IDataProvider dp)
-		{
-			return dp.CreateSqlBuilder;
-		}
-
-		Func<ISqlBuilder> IDataContext.CreateSqlProvider => GetCreateSqlProvider(DataProvider);
+		Func<ISqlBuilder> IDataContext.CreateSqlProvider => () => DataProvider.CreateSqlBuilder(MappingSchema);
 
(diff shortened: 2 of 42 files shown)


MaceWindu:
Whoops, forgot to alter t4 templates for CreateSqlBuilder calls, will fix soon

PR #1800, opened 2019-07-06
Condition optimizations

### LINKED ISSUE #1767: Strange WHERE predicate begin generated when check for NULL
Hello.

I have met strange SQL generated when adding null-check in WHERE clause:

```c#
var veh = con.Vehicles.FirstOrDefault(x => x.StyleId != null && x.StyleId != -1000);
```

I get this SQL:

```sql
--  SqlServer.2008
SELECT TOP (1)
        [x].[VehicleId],
        [x].[LV_autoStyleId]
FROM
        [Vehicle] [x]
WHERE
        [x].[LV_autoStyleId] IS NOT NULL AND ([x].[LV_autoStyleId] IS NULL OR [x].[LV_autoStyleId] <> -1000)
```
`LV_autoStyleId` is checked against NULL and NOT NULL simultaneously. Is it by design or I am  doing something wrong?

This doesn't bother me much, just curious if it is normal.

I saw it in 2.6.4 and still see this in 2.7.4 (had to update just to check this is bug and was fixed).

MaceWindu:
Trying to add more condition optimizations to fix #1767 
Already implemented several optimizations, but main one to fix actual issue, not yet implemented.
Maybe will get back to it later...

### DIFF
--- a/Source/LinqToDB/SqlQuery/ISqlPredicate.cs
+++ b/Source/LinqToDB/SqlQuery/ISqlPredicate.cs
@@ -2,7 +2,7 @@
 
 namespace LinqToDB.SqlQuery
 {
-	public interface ISqlPredicate : IQueryElement, ISqlExpressionWalkable, ICloneableElement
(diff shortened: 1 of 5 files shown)


MaceWindu:
Not today

Issue #1791, opened 2019-07-03
[Idea] Package(s) with DB-specific functionality


MaceWindu:
Right now linq2db contains a lot of definitions of SOME db-specific functions, variables, etc. And miss event more of them.
Problems with this approach:
- people often are not aware that linq2db already have mapping for some function and create own mapping
- people often are not aware that linq2db already have mapping for some function and need to create own mappingneed to call some function, but we don't provide mapping for it out of box, so again they need to define it in their code (e.g. see #1790)
- people not aware how to create such mappings properly due to lack of proper documentation
- linq2db codebase became polluted with a lot of mappings

I propose to create separate package or packages which will provide access to db-specific functionality. Pros:
- no more function mappings in main project codebase
- it will be easier to discover existing mappings
- it will be easier to create PR with new mapping
- it will be easier to write mappings by looking at code of this new project compared to main linq2db codebase digging

I propose to release initial version of it with 3.0 release.

Open questions:
- [ ] should it be single package or multiple packages - one per db
- [ ] how navigation path to specific function/property should look. e.g. `Functions.SomeFunction()` or `MySql.SomeFunction()` or `MySql.Functions.SomeFunction()` or ...

PR #1803, opened 2019-07-07
improve RESPECT NULLS window function token support across databases

### LINKED ISSUE #1732: error with lag function
I am getting incorrect syntax near respect when using lag function.

```cs
var posi =
                    from p in db.GetTable<Position_c>()
                    where p.Msg_Dt == yesterday
                    select new vehicle_positions
                    {
                        Hral_veh_id = p.Hral_Veh_Id,
                        prev_veh_id = Sql.Ext.Lag(p.Hral_Veh_Id, Sql.Nulls.Respect, 1, null).Over().OrderBy(p.Msg_Dt).ThenBy(p.Hral_Veh_Id).ThenBy(p.Msg_Tm).ToValue(),

                    };
```

MaceWindu:
Fix #1732 
- [x] don't generate RESPECT NULLS as it is default option
- [x] test that it doesn't break other functions, that use this option: LEAD, FIRST_VALUE, LAST_VALUE, NTH_VALUE

### DIFF
--- a/Source/LinqToDB/Sql/Sql.Analytic.cs
+++ b/Source/LinqToDB/Sql/Sql.Analytic.cs
@@ -107,16 +107,15 @@ namespace LinqToDB
 		{
 			switch (nulls)
 			{
-				case Sql.Nulls.None :
-				case Sql.Nulls.Respect :
+				case Sql.Nulls.None   :
+				case Sql.Nulls.Respect:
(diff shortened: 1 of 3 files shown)


MaceWindu:
Tests show that we should tune syntax for some databases. Probably we can address it after #1791 implemented

PR #1804, opened 2019-07-07
Force subquery context for SelectManyBuilder left-side select query

### LINKED ISSUE #1799: Incorrect sql generated for complex query with window functions
https://github.com/linq2db/linq2db/issues/1763#issuecomment-502387772

Awesome, works like a charm.
Thank you very much. 
Didn't expect such quick resolution. For most other open-source libraries I would need to explore source code before got an advice (if ever). 

> Simple tip: if you rewriting SQL, separate it by parts, starting from subqueries.

Thanks. 
However with your awesome library it seems I will not need to use plain `SQL` at all, love this idea.

## Exploration
But I can't sleep until find out the root cause. Can you help me by elaborating your solution a little bit? Actually, it's not urgent of course, ok to do it later. 

So I explored your solution. 

Replaced `Sql.DateDiff` because the same errors appeared even when called `LAG` for pure `int` column, like `processID` (sorry, totally forgot to wrote about that in my previous messages, maybe tired). Your solution still worked.

Then removed
```cs
query = query.Where(q => q.Diff > 0 && q.Diff <= 5);
```
failed with
```cs
Unhandled Exception: System.Data.SqlClient.SqlException: 
Column 'Process_IndexedEvents.eventTime' is invalid in the 
select list because it is not contained in either an 
aggregate function or the GROUP BY clause.

Windowed functions cannot be used in the context of 
another windowed function or aggregate.
```
Ok, probably that, it's warm.

Moved `where` clause into second `query`, still worked.

Removed `joins`, still worked.

Current state
```cs
var query =
    from x in db.GetTable<Process_IndexedEvents>()
    select new
    {
        User = x.eventUser,
        Proc = x.processID,
        Diff = (x.eventTime - Sql.Ext
                                 .Lag(x.eventTime, Sql.Nulls.None)
                                 .Over()
                                 .PartitionBy(x.eventUser, x.processID)
                                 .OrderBy(x.eventTime)
                                 .ToValue()).TotalMinutes,
    };

var finalQuery = from q in query
                 where q.Diff > 0 && q.Diff <= 5
                 group q by new {q.User,q.Proc}
                 into g
                 select new
                 {
                     g.Key.User,
                     g.Key.Proc,
                     Sum = g.Sum(e => e.Diff) / 60.0
                 };
```
So can you explain what is the magic of `where` clause? I thought that root cause was related to `Windowed functions cannot be used in the context of another windowed function or aggregate`. I need to know, otherwise will return with a new query after few days.

_Originally posted by @Brains in https://github.com/linq2db/linq2db/issues/1763#issuecomment-502391032_

MaceWindu:
Fixes #1799

### DIFF
--- a/Source/LinqToDB/SqlQuery/SqlSelectClause.cs
+++ b/Source/LinqToDB/SqlQuery/SqlSelectClause.cs
@@ -240,11 +240,6 @@ namespace LinqToDB.SqlQuery
 
 		#region Distinct
 
-		public SqlSelectClause Distinct
-		{
-			get { IsDistinct = true; return this; }
-		}
-
 		public bool IsDistinct { get; set; }
 
 		#endregion
--- a/Source/LinqToDB/Linq/Builder/SelectContext.cs
+++ b/Source/LinqToDB/Linq/Builder/SelectContext.cs
@@ -173,26 +173,18 @@ namespace LinqToDB.Linq.Builder
 										case ExpressionType.New        :
 										case ExpressionType.MemberInit :
 											{
-												return memberExpression.Transform(e =>
+												var resultExpression = memberExpression.Transform(e =>
 												{
 													if (!ReferenceEquals(e, memberExpression))
 													{
 														switch (e.NodeType)
 														{
 															case ExpressionType.MemberAccess :
-																var sequence = GetSequence(memberExpression, 0);
-
-																if (sequence != null &&
-																	!sequence.IsExpression(e, 0, RequestFor.Object).Result &&
-																	!sequence.IsExpression(e, 0, RequestFor.Field). Result)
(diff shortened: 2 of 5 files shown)


Source/LinqToDB/SqlQuery/SqlSelectClause.cs
```diff
@@ -240,11 +240,6 @@ int AddOrFindColumn(SqlColumn col)
 
 		#region Distinct
 
-		public SqlSelectClause Distinct
```
sdanyliv: Why this removed?

MaceWindu: - not used
- changes AST state when query viewed in debugger

PR #1810, opened 2019-07-13
[MSSQL] Add DateTimeOffset members mapping

### LINKED ISSUE #1666: DateTimeOffset.Date is translated incorrectly
The `DateTimeOffset.Date` property in .NET returns the date component with time set to `00:00:00`. However, when used in a Linq2DB query, it does nothing and just returns the original date.

Here's a code in LinqPad:

```c#
void Main()
{
	using (var ctx = new Context())
	{
		var start = new DateTime(2019, 1, 1);
		var txns = ctx.Trans.Where(x => x.TransactionDate > start);
		new
		{
			Expected = txns.ToList() // forces IEnumerable
			               .GroupBy(d => d.TransactionDate.Date)
                                       .Select(x => new { Key = x.Key, Count = x.Count() }),
						   
			Actual = txns.GroupBy(d => d.TransactionDate.Date)
			             .Select(x => new { Key = x.Key, Count = x.Count() })
		}.Dump();
	}
}

class Context: LinqToDB.Data.DataConnection
{
	public Context(): base("SqlServer", "data source=.;integrated security=True") { }
	public LinqToDB.ITable<Transaction> Trans => GetTable<Transaction>();
}

[Table("Transactions")]
class Transaction
{
	[PrimaryKey] public long TransactionId {get; set;}
	[Column]     public DateTimeOffset TransactionDate {get; set;}
}
```

Here's the result:

image

### Environment details
linq2db version: *2.6.4*
Database Server: *SQL Server 14.0.1000.169*

MaceWindu:
Fixes #1666 

- Adds mapping to SQL for most of DateTimeOffset type members:
  - Year
  - Month
  - DayOfyear
  - Day
  - DayOfWeek
  - Hour
  - Minute
  - Second
  - Millisecond
  - Date
  - TimeOfDay
  - AddYears()
  - AddMonths()
  - AddDays()
  - AddHours()
  - AddMinutes()
  - AddSeconds()
  - AddMilliseconds()
- fixes SQL, generated for Add_interval_ methods for Oracle to preserve source value type

Those mappings should be be used only for SQL Server 2008+. For other databases it could work in some cases, but in general it is not supported (and hardly ever be supported) due to:
- lack of datetime type with offset on database side
- different logic implemented in .net for DateTimeOffset type and in database for DataTimeOffset-like type


Source/LinqToDB/Sql/Copy.DateTimeOffset.cs
```diff
@@ -0,0 +1,83 @@
+﻿using System;
```
sdanyliv: Bad name of the file

MaceWindu: busted...

PR #1819, opened 2019-07-22
Version 3.0 refactorings

MaceWindu:
Done:
- [x] solutions and projects cleaned and updated to VS2019/C#8 (without nullable types yet)
- [x] removed Travis/AppvVeyour integration
- [x] removed npgsql4 test provider and move existing postgress tests to npgsql4 as part of npgsql3 removal
- [x] removed build support for netcore 1.1/netstandard 1.6
- [x] enable nullable reference types for T4 test projects (including templates generation mode)
- [x] update build scripts (*.cmd)
- [x] Enable Azure DevOps build
- [x] Enable Azure DevOps testing in general
- [x] update nuspec files
- [x] fix nuget build warnings
- [x] Enable publish to [Azure DevOps feed](https://dev.azure.com/linq2db/linq2db/_packaging?_a=feed&feed=linq2db). Replaces myget feed, used by pre-3.0 (still myget could be re-added if needed)
- [x] Enable Nuget.org nuget publishing (will be tested when we release first beta to nuget.org)
- [x] cleanup/review/fix conditional build directives (like `#if NETSTANDARD1_6`)
- [x] create scripts to run tests from command line
- [x] Enabled CreateDatabase test for baseline database even when provider is disabled to ensure we have proper baseline data

To do:
- [ ] review/update nuget dependencies
- [ ] Enable more test providers (to be on par with old AppVeyor/Travis integrations)
- [ ] fix projects build warnings

Next tasks out of scope of this PR, that will be addressed later:
- support for features from netcoreapp2.1+/netstandard2.1
- nullable reference types everywhere
- enable more target databases for CI testing
- [maybe] enable missing XML-doc compiler warning and swith to warning as error mode to not regress here

After this task is done, we probably can release 3.0.0-beta1 release


MaceWindu:
marking it as ready for review, as it doesn't have blocking tasks left - we just need to enable more test providers for azure and it could be done by separate commits

Source/LinqToDB/Linq/Builder/EnumerableContext.cs
```diff
@@ -16,6 +16,7 @@ class EnumerableContext : IBuildContext
 
 #if DEBUG
 		public string                _sqlQueryText { get; }
+		public string                Path          { get; }
```
sdanyliv: This property should have implementation. Check other contexts.

MaceWindu:
Merging it. Will continue with remaining test providers in new PR

PR #1834, opened 2019-07-31
Feature multi result set support

sami016:
Feature adds support for stored procedures that return multiple result sets.
This is important to support in some cases due to the inefficiency of making multiple queries (especially when ids need to be passed to subsequent calls).
As far as I can tell, there is no other way to achieve this using linq2db.

**Attribute approach**

A stored procedure is capable of returning multiple results sets in a given order.
Using attributes we can specify these indexes on a result object like so:

```cs
    [MultipleResultSets]
    public class MySearchResultObject
    {
        [ResultSetIndex(0)]
        public IEnumerable<Person> People { get; set; }

        [ResultSetIndex(1)]
        public IEnumerable<Product> OwnedProducts { get; set; }

        [ResultSetIndex(2)]
        public IEnumerable<Group> GroupMemberships { get; set; }

        [ResultSetIndex(3)]
        public IEnumerable<Person> Friends { get; set; }
    }
```
We can then execute a new extension method in code that will fetch results for this:

```
var result = db.QueryProcMulti<MySearchResultObject>(
    "my_stored_procedure_name",
    new DataParameter("name", name)
);
```

Each property in the MySearchResultObject object marked with the ResultSetIndex attribute will have its contents serialized using the item type of the property.

**Changes**

Introduces a new section into CommandInfo.cs for handling multiple result sets.
This uses the NextResult method of the data reader to iterate through the data sets.
If a result set index is not defined on the result object, it will be ignored.

```cs
                T ReadMultipleResultSets<T>(IDataReader rd)
		{
			// Dictionary mapping the result query index to the type it should be reading.
			var resultSetIndexProperties = typeof(T).GetProperties()
				.Where(x => x.GetCustomAttribute<ResultSetIndexAttribute>() != null)
				.ToDictionary(
					x => (uint)x.GetCustomAttribute<ResultSetIndexAttribute>().Index,
					x => x
				);
			var readEnumeratorGeneric = typeof(CommandInfo).GetMethod("ReadEnumeratorAsList", BindingFlags.NonPublic | BindingFlags.Instance);

			uint resultIndex = 0;
			var result = (T)Activator.CreateInstance(typeof(T));
			do
			{
				// Only process the field if we're reading it into a property.
				if (resultSetIndexProperties.ContainsKey(resultIndex))
				{
					var property = resultSetIndexProperties[resultIndex];
					var itemType = property.PropertyType.GetItemType();
					var resultSetEnumerable = readEnumeratorGeneric.MakeGenericMethod(new Type[] { itemType })
						.Invoke(this, new object[] { rd });
					property.SetValue(result, resultSetEnumerable);
				}
				resultIndex++;
			} while (rd.NextResult());
			return result;
		}
```

Using a generic method to read the set of objects for each property based on the item type.
Note: I've made this output as a list enumerable and not a lazy-loaded yield enumerable since lazy loading will not work due to the NextResults call. I can see no way around this other than not lazy-load the items.

```cs
IEnumerable<T> ReadEnumeratorAsList<T>(IDataReader rd)
		{
			var results = new List<T>();
			if (rd.Read())
			{
				var additionalKey = GetCommandAdditionalKey(rd);
				var objectReader = GetObjectReader<T>(DataConnection, rd, DataConnection.Command.CommandText, additionalKey);
				var isFaulted = false;

				do
				{
					T result;

					try
					{
						result = objectReader(rd);
					}
					catch (InvalidCastException)
					{
						if (isFaulted)
							throw;

						isFaulted = true;
						objectReader = GetObjectReader2<T>(DataConnection, rd, DataConnection.Command.CommandText, additionalKey);
						result = objectReader(rd);
					}

					results.Add(result);

				} while (rd.Read());

			}
			return results;
		}
```
This would be invoked via an extension method in the typical way:

```cs
                public static T QueryProcMulti<T>(this DataConnection connection, string sql, params 
                     DataParameter[] parameters)
		{
			return new CommandInfo(connection, sql, parameters).QueryProcMulti<T>();
		}
```


sdanyliv:
Interesting idea, but realization is not complete. I think we will accept this PR but with a lot of changes.
Let's clarify some parts:

1. Same result sets can be returned by just raw SQL.
2. Do we need `ResultSetIndex`? Or maybe it is better also support case when it is not set.
3. How about one record results, or just scalar value (int, string, double, etc.)
4. Also needed supporting of array properties.
5. Async version is not implemented - we can cover that.
6. No Unit tests.

If you want to finalize your PR we can help in these steps.

sami016:
> Interesting idea, but realization is not complete. I think we will accept this PR but with a lot of changes.
> Let's clarify some parts:
> 
> 1. Same result sets can be returned by just raw SQL.
> 2. Do we need `ResultSetIndex`? Or maybe it is better also support case when it is not set.
> 3. How about one record results, or just scalar value (int, string, double, etc.)
> 4. Also needed supporting of array properties.
> 5. Async version is not implemented - we can cover that.
> 6. No Unit tests.
> 
> If you want to finalize your PR we can help in these steps.

I think ResultSetIndex is required, since you'd need to know the type being serialized when reading any given particular result set, and all we have to go off is the ordering they are provided in. I'm open to other ideas on this if anyone can think of a better way.

Good point regarding scalar values. There should be a case to handle this whenever the property type isn't an enumerable/list/array.

I can add in support for array values too, and take a look at getting unit tests in.

sdanyliv:
Check my old comment how to work with unit tests:
https://github.com/linq2db/linq2db/issues/1303#issuecomment-425113822
It might be outdated but scenario is similar.
In `Database/Create Scripts` folder you can find scripts for particular database.

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -452,6 +453,112 @@ public IEnumerable<T> QueryProc<T>(T template)
...
+		/// <summary>
+		/// Executes command using <see cref="StoredProcedure"/> command type and returns a result containing multiple result sets.
+		/// </summary>
+		/// <typeparam name="T">Result set type.</typeparam>
+		/// <returns>Returns result.</returns>
+		public T QueryProcMulti<T>()
+		{
+			CommandType = CommandType.StoredProcedure;
+			// Check whether attribute has been applied.
+			if (typeof(T).GetCustomAttributeEx<MultipleResultSetsAttribute>(false) == null)
```
sdanyliv: Replace with `DataConnection.MappingSchema.GetAttribute`

Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -307,6 +307,21 @@ public static object[] GetCustomAttributesEx(this Type type, bool inherit)
 #endif
 		}
 
+		public static T GetCustomAttributeEx<T>(this Type type, bool inherit)
```
sdanyliv: Not needed. Use MappingSchema.

sami016:
> Check my old comment how to work with unit tests:
> [#1303 (comment)](https://github.com/linq2db/linq2db/issues/1303#issuecomment-425113822)
> It might be outdated but scenario is similar.
> In `Database/Create Scripts` folder you can find scripts for particular database.

I've been trying to get this working for a while after following the steps you provided.
Visual studio (tried 2017+resharper, 2019+without resharper) doesn't seem to be recognising any method with a [DataSource] argument (it runs anything without arguments just fine).

```[31/07/2019 15:44:19 Warning] No test matches the given testcase filter `FullyQualifiedName=Tests.Data.ProcedureMultipleResultTests.MultiTest` in C:\develop\git\3rdParty\linq2db\Tests\Linq\bin\Debug\net46\linq2db.Tests.dll```

Is there anything I need to install (nuget packages, etc?) to get this to work?

**Comment by sdanyliv** [comment]
@sami016, i suggest to open linq2db.Playground.sln - it is small project and tests run fast.
Try to remove UserDataProviders.json, maybe it is broken.
Try to run tests in debug, maybe initialization failed.

**Comment by sami016** [comment]
> @sami016, i suggest to open linq2db.Playground.sln - it is small project and tests run fast.
> Try to remove UserDataProviders.json, maybe it is broken.
> Try to run tests in debug, maybe initialization failed.

Thanks, I managed to get this to work in the end.

I've added a support for multiple results for standard sql queries:

```
		public static T QueryMulti<T>(this DataConnection connection, string sql, params DataParameter[] parameters)
		{
			return new CommandInfo(connection, sql, parameters).QueryMulti<T>();
		}
```

In addition, I've also now added support for array types. 

Your suggested changes have been implemented, and unit tests have been written to cover the basic cases of this working.

File: Source/LinqToDB/Extensions/ReflectionExtensions.cs [hunk 3]
```diff
@@ -880,6 +880,15 @@ public static Type GetListItemType(this Type listType)
 			return typeof(object);
 		}
 
+		public static bool IsGenericEnumerableType(this Type type)
+		{
+			if (type.IsGenericTypeEx())
+				foreach (var aType in type.GetGenericArgumentsEx())
```
sdanyliv: Replace `foreach` with `if (typeof(IEnumerable<>) .IsSameOrParentOf(type))` you do not have to check every generic arguments.

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -452,6 +453,166 @@ public IEnumerable<T> QueryProc<T>(T template)
...
+
+		/// <summary>
+		/// Executes command and returns a result containing multiple result sets.
+		/// </summary>
+		/// <typeparam name="T">Result set type.</typeparam>
+		/// <returns>Returns result.</returns>
+		public T QueryMulti<T>()
+		{
+			// Check whether attribute has been applied.
+			if (DataConnection.MappingSchema.GetAttribute<MultipleResultSetsAttribute>(typeof(T)) == null)
```
sdanyliv: I do not see benefits in this attribute. Better to remove this check and whole attribute class.

Tests/Linq/Data/ProcedureMultipleResultTests.cs
```diff
@@ -0,0 +1,64 @@
...
+using LinqToDB.Mapping;
+using NUnit.Framework;
+using LinqToDB.Data;
+using System.Collections.Generic;
+using Tests.Model;
+
+namespace Tests.Data
+{
+	[TestFixture]
+	public class ProcedureMultipleResultTests : TestBase
```
sdanyliv: Concat tests in one file.

Tests/Linq/Data/ProcedureMultipleResultTests.cs
```diff
@@ -0,0 +1,64 @@
...
+	{
+		[Table]
+		[MultipleResultSetsAttribute]
+		class MultipleResultExample
+		{
+			[ResultSetIndex(0)] public IList<int> MatchingPersonIds { get; set; }
+			[ResultSetIndex(1)] public IEnumerable<Person> MatchingPersons { get; set; }
+			[ResultSetIndex(2)] public IEnumerable<Patient> MatchingPatients { get; set; }
+			[ResultSetIndex(3)] public bool DoctorFound { get; set; }
+			[ResultSetIndex(4)] public Person[] MatchingPersons2 { get; set; }
```
sdanyliv: Add test for single entity
```cs
public Person MatchingPerson { get; set; }
```

Tests/Linq/Data/ProcedureMultipleResultTests.cs
```diff
@@ -0,0 +1,64 @@
...
+		{
+			using (var db = new DataConnection(context))
+			{
+				var res = db.QueryProcMulti<MultipleResultExample>(
+					"PersonSearch",
+					new DataParameter("nameFilter", "Jane")
+				);
+
+				Assert.IsFalse(res.DoctorFound);
+				Assert.AreEqual(res.MatchingPersonIds.Count(), 1);
```
sdanyliv: Also need to check how fields are mapped.

sami016: Any idea how to handle fields?

I'm happy to add support for fields in the case that no properties are found.

Mixing fields and properties might be troublesome, since GetProperties() and GetFields() won't give an accurate ordering in the case of implicit result ordering?

sdanyliv: Sorry, i mean properties. Just properties. We will correct your mapper in next commit step.

sami016: I've checked the returned object fields in the latest check in, if that's what you mean. Please reopen if I've misunderstood.

sdanyliv:
@sami016, thanks for your work, i'll take care about additional changes by myself. I have ideas how to simplify code and i have to add async support.

sdanyliv:
Superseded by #1844 
@sami016, i have refactored a lot and added async support.

PR #1835, opened 2019-07-31
Implementing EXCEPT, INTERSECT support in Linq Queries.

sdanyliv:
Implemented missing EXCEPT, INTERSECT SQL Operators. 
Also added new functions `ExceptAll`, `IntersectAll` and as a bonus `UnionAll` - it is similar to `Concat `but name is closer to SQL variant.

**Tasks to do:**

- [x] Review optimizators
- [x] More tests
- [x] Run full test suite


Franki1986:
Should there be a version with join for those (Firebird, Access) who do not support except and intersect?

sdanyliv:
What can I say. I can support emulation in version 3.0 if there will be clear algorithm.

sdanyliv:
@Franki1986, emulation added. Don't worry.

PR #1858, opened 2019-08-09
Enable nullable reference types

MaceWindu:
- enables nullable reference types for all projects except tests
- add #enable disable opt-out to all files that give nullability warnings or needs review
- added couple of nullable annotations
- moved WPF T4 templates tests to T4 test project and removed WPF test project


MaceWindu:
Let's set starting point for NRT support

PR #1872, opened 2019-08-24
[Azure DevOps] Add DB2/Informix/SAP HANA support

### LINKED ISSUE #1870: Better IBM.Data.DB2.Core package support
List of tasks:
- support this package for Informix provider
- add t4 nugets for win/osx/linux versions of provider
- add testing on azure pipelines against db2 and informix for all environments

MaceWindu:
Hopefully last part of Azure pipelines testing support work. We have only 3 databases left not covered:
- DB2
- Informix
- SAP HANA 2

Remaining tasks will be (could be handled later at any time):
- refactoring of pipelines definitions to use templates to reuse build definitions
- add missing test database/target/os runs
- add more test pipelines to run tests for specific db/target/os
- migrate to private linq2db docker hub to have less depndency on 3rd parties and better optimized images for our needs
- use docker compose

Contains:
- [API Change] added trace level parameter to DataConnection.WriteTraceLine
- append test outputs always for error logs
- append test outputs always for failed tests
- Informix 14 testing using DB2.Core provider (v12 still not configured)
- DB2.Core provider support for Informix. Fix #1870
- DB2 LUW testing using DB2.Core provider
- DB2/Informix Core provider nugets (requires PackageReference support, otherwise DB2.Core reference will not be added automatically)


MaceWindu:
ok, looks like we have DB2 and IFX14 running
IFX12 still goin banana and SAP HANA image is useless, because there is only windows client and linux docker image, so they will never met each other in current Azure environment

MaceWindu:
requesting review for informix provider changes to support DB2.Core client

MaceWindu:
We can actually ignore IFX12 for now as we have v14 working anyway and only difference is CTE support in v14 (still will try a bit more to make it work).

Which will mean that we miss only SAP HANA testing in azure  in general.

MaceWindu:
Actually it makes sense to give a shot for SAP ODBC provider on linux, we already have ODBC implemented for HANA, but not use it tests

MaceWindu:
Also it makes sense to remove separate test project for DB2.Core provider we have and enable type tests in main test project

MaceWindu:
update:
- update dependencies
- retarget tests to use net core 2.1 instead of 2.0 (2.0 is not compatible with 16.3 test framework)
- fix nuget errors and add package icon
- add html reports support to local test.cmd

Remaining tasks for this pr:
- check and if it works, switch to docker 2.1 if it fixed for osx
- fix nuget.exe warnings for oracle.managed package

PR #1881, opened 2019-09-06
Fixed found problems with CTE and HasUniqueKeysBuilder.

sdanyliv:
Solves problems found during analysis issue #1875

### DIFF
--- a/Source/LinqToDB/Sql/Sql.Analytic.cs
+++ b/Source/LinqToDB/Sql/Sql.Analytic.cs
@@ -484,26 +484,26 @@ namespace LinqToDB
 		#region Count
 
 		[Sql.Extension("COUNT({expr})", IsAggregate = true, ChainPrecedence = 0)]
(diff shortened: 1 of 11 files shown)


Source/LinqToDB/Sql/Sql.Analytic.cs
```diff
@@ -518,7 +518,7 @@ public static IAggregateFunctionSelfContained<long> Count(this Sql.ISqlExtension
 		}
 
 		[Sql.Extension("COUNT({expr})", TokenName = FunctionToken, ChainPrecedence = 1, IsAggregate = true)]
-		public static IAggregateFunctionSelfContained<T> Count<T>(this Sql.ISqlExtension ext, [ExprParameter] T expr)
+		public static IAggregateFunctionSelfContained<long> Count<T>(this Sql.ISqlExtension ext, [ExprParameter] T expr)
```
MaceWindu: shouldn't it be int? and additional long method for COUNT_BIG?

MaceWindu: looks like it is more tricky, different databases use different approches with count(), better to handle it as separate task

PR #1893, opened 2019-09-17
[WIP] Added parameter value conversion logic.

### LINKED ISSUE #1888: [Question] type of output DataParameter.Value
I'm not sure if this is the expected behavior and/or how I should be doing the following.

I'm calling into an Oracle stored procedure that has Output parameters.

Here's an example:
```csharp
var p1 = new DataParameter { Name = "p1", Direction = ParameterDirection.Output, DataType = DataType.Int32 };
var p2 = new DataParameter { Name = "p2", Direction = ParameterDirection.Output, DataType = DataType.VarChar, Size = 400 };

db.ExecProc("package.proc_with_output", p1, p2);

var value1 = (int)p1.Value; // yes
var value2 = (string)p2.Value; // no, Value is OracleString
```

The `DataParameter` values are properly populated with what the procedure actually returned.

But the catch is that `p1.Value` is a C# type (`Int32`) rather than provider-specific (`OracleNumber`).
On the other hand, `p2.Value` is _not_ a C# type (`String`) but rather _is_ provider-specific `OracleString`.

Is that expected? Is there an easy way to get C# values (`null` or `string`) from an  output/return value `DataParameter`?

This would be convenient to use and also useful, as linq2db hides provider-specific implementation and dialect by offering a db-neutral API (i.e. you can switch providers).

For comparison, with Dapper you can do:
```csharp
var parameters = new DynamicParameters(); // etc. 
// after execution:
var value2 = parameters.Get<string>("p2");
```

sdanyliv:
Fixes #1888 

Question: should we use here `Configuration.AvoidSpecificDataProviderAPI` setting?


MaceWindu:
> 
> 
> Fixes #1888
> 
> Question: should we use here `Configuration.AvoidSpecificDataProviderAPI` setting?

I think yes. For some reason we don't use it in some providers, probably just overlooked.

Can you add outputs asserts to test to make sure it works?

MaceWindu:
I've started to rework miniprovider/AvoidSpecificDataProviderAPI support as part of work on #1948, as it is getting out of control

MaceWindu:
integrated into #1961

PR #1901, opened 2019-09-26
Fix for #1900. Added IsDynamic property to filter dynamically created fields.

### LINKED ISSUE #1900: Filter dynamic column with ExpressionMethod throws Exception
LinQToDb: 2.7.4
Firebird: 2.1

Hi,

is it possible to use dynamic columns with ExpressionMethods?
I have the following:

```
[Table("PERSON")]
public class Person{
    [COLUMN("AGE")]
    public int Age {get; set;}
    [COLUMN("NAME")]
    public string Name {get; set;}

    [ExpressionMethod(nameOf(GetTypeExpr), IsColumn = true)]
    public int Type {get; set;}

    public static Expression<Func<Person, int>> GetTypeExpr(){
        return p => (Sql.Property<int>(p, "OPTIONS")& 0xF);
    }
}
```

The table looks like this:

Tablename PERSON,
Columns: AGE (int), NAME (int), OPTIONS (string)
      

The real context looks complexer...

This works:

```
using(var db = new MyDb()){
    var result = db.GetTable<Person>().ToList();
}

```
result has all Types set with the right values.
But if I put in a Where filter, it throws an ArgumentOutOfRangeException (TableContext.cs line 335)

```
using(var db = new MyDb()){
    var result = db.GetTable<Person>().Where(p => p.Type == 4).ToList();
}
```

sdanyliv:
Fixes #1900

### DIFF
--- a/Source/LinqToDB/ServiceModel/LinqServiceSerializer.cs
+++ b/Source/LinqToDB/ServiceModel/LinqServiceSerializer.cs
@@ -676,6 +676,7 @@ namespace LinqToDB.ServiceModel
 							Append(elem.IsIdentity);
 							Append(elem.IsUpdatable);
 							Append(elem.IsInsertable);
+							Append(elem.IsDynamic);
 							Append((int)elem.DataType);
 							Append(elem.DbType);
 							Append(elem.Length);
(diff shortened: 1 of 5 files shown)


Source/LinqToDB/ServiceModel/LinqServiceSerializer.cs
```diff
@@ -1352,8 +1354,9 @@ bool Parse()
 								IsPrimaryKey    = isPrimaryKey,
 								PrimaryKeyOrder = primaryKeyOrder,
 								IsIdentity      = isIdentity,
-								IsInsertable    = isInsertable,
 								IsUpdatable     = isUpdatable,
```
MaceWindu: o_O

PR #1905, opened 2019-09-28
[t4] procedure return value support

### LINKED ISSUE #1897: [Request, T4 template] Support for ParameterDirection.ReturnValue
Current version of l2db template [has no support](https://github.com/linq2db/linq2db/blob/30e0cbf0dd7b3a3dad38855e21d672e5f3bd00b9/Source/LinqToDB.Templates/LinqToDB.ttinclude#L887-L896) for return parameters. 
Can you add an option like
```cs
Procedures["SomeProc"].UsesReturn = true;
```
?

**Use case:** we have a legacy MS SQL db with a lot of SPs that use `RETURN someValue;`. 

We've workarounded it with c# helper methods that wrap `QueryProc()` calls but having such feature will be a real time-saver:)

### Environment details
linq2db version: *2.9.0*
Database Server: *MS SQL 2017*
Database Provider: *System.Data.Sql*
Framework version: *.NET Core 3*

MaceWindu:
Adds support for return value parameter generation for procedures by T4.

Fixes #1897

### DIFF
--- a/Data/Create Scripts/SqlServer.sql
+++ b/Data/Create Scripts/SqlServer.sql
@@ -1116,4 +1116,14 @@ BEGIN
 END
 GO
 
+IF EXISTS (SELECT * FROM sys.objects WHERE type = 'P' AND name = 'Issue1897')
+BEGIN DROP Procedure Issue1897 END
+GO
 
+CREATE PROCEDURE dbo.Issue1897
+AS
+BEGIN
+	RETURN 4
+END
+
+GO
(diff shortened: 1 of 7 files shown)


MaceWindu:
Added `AddReturnParameter(procName, paramName = "@return")` helper to manually enable generation of return parameter for procedure.
Looks like sql server procedures always have return value parameter, even if it is not used explicitly, so we cannot tell if it should be generated or not.

MaceWindu:
@ig-sinicyn, can you check if this solution will work for you?

ig-sinicyn:
Checked on real project, no issues so far:)
Thanks a lot!

Issue #593, opened 2017-03-11
AccessSchemaProvider.GetForeignKeys returns empty list


Dutchdre:
Last week I was stumbling through the process of auto generating my model based on text templates with an underlying Microsoft Access database. Surprisingly the GetForeignKeys method of the AccessSchemaProvider is not implemented and returns an empty list.

ili:
OleDb do not provide schema info about foreign keys.

The way i'v found is to make query to `MSysRelationships` but wen i'm making select from tests I do have an error `Record(s) cannot be read; no read permission on 'MSysRelationships'. `

PR #1907, opened 2019-09-29
[schema] add option to disable FK schema information load

MaceWindu:
This change required to implement workaround for https://github.com/linq2db/linq2db.LINQPad/issues/23 where JET engine crashes with AV when we try to load foreign key information


MaceWindu:
Pity we cannot just read MSysRelationships table... https://github.com/linq2db/linq2db/issues/593

PR #1920, opened 2019-10-06
Netcore providers support part2

### LINKED ISSUE #1914: Need ProviderName for MS Access
Hello 

I try to use Linq2DB for MsAccess. I try it with MSSql, it works fine.

Now I Try to set the cionnectionstring for MSAccess (see below) I get an exception
"DataProvider is not provided for configuration"

Could you help me ? Thanks Andreas
```cs
new ConnectionStringSettings
{
   Name = "Access",
   ProviderName = "Access",
   ConnectionString = @"Data Source=c:\Test\AtgCor32.mdb;Locale Identifier=1033;Jet OLEDB:Engine Type=5;Persist Security Info=True"
};
```

MaceWindu:
- [x] MS Access: add .net core (win-only) suppport
- [x] SQL CE: add .net core (win-only) suppport

Fix #1914


MaceWindu:
In general Access provider works for core with ACE provider, but many tests with explicit transaction commit/rollback fail due to COM errors.

Could be just local issue due to install of JET x64 provider. Will check later

MaceWindu:
almost no changes to support SQL CE - works out of box, just enabled `DbProviderFactories` support for easier setup

Also see https://github.com/dotnet/corefx/issues/33897 for non-factory approach

With factory you just need to add this line to your application https://github.com/linq2db/linq2db/pull/1920/files#diff-c69cd55216d563674b5eff86b0bc43f4R94

MaceWindu:
Ok, Access oledb provider is a PITA.
- Microsoft.Jet.OLEDB.4.0 exists only as x86 library
- Microsoft.ACE.OLEDB.12.0 exists for both x86 and x64 but looks like if you install both versions, they start to conflict with each other
- also I have x64 version of something called Microsoft.ACE.OLEDB.15.0, need to try it too

Another problem is that we run our tests using JET right now, so we don't check ACE compatibility regulary...

Issue #1915, opened 2019-10-03
Implementing custom IDataContext and the public APIs being sufficient


davidglassborow:
Related to #1912, I'm trying to use linq2db where I can provide the actual data myself through other means, similar to [LINQ over WCF](http://blog.linq2db.com/2016/03/linq-over-wcf.html)

The [Public API ](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/IDataContext.cs#L17) has [IDataContext.GetQueryRunner](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/IDataContext.cs#L104) that returns a [IQueryRunner](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/IQueryRunner.cs#L9) which looks like it should do the job of what I need.

I looked at the source code for the [RemoteDataContextBase](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/ServiceModel/RemoteDataContextBase.QueryRunner.cs#L13) which is the base class for the WCF / ServiceModel functionality, and implements IQueryRunner, to see how to use this interface.

Looking at how it's implemented, and trying to replicate the logic of getting the SQL / Parameters so I can then handle getting the data myself, I found a couple of issues around the APIs not being sufficient / relying on internal access.

The perfect example is in [QueryRunner.ExecuteNonQuery()](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/ServiceModel/RemoteDataContextBase.QueryRunner.cs#L126) where it accesses an [Internal only API](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/Query.cs#L29) to get the details of the exact query to run, implemented as [IQueryContext](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/IQueryContext.cs).

I've managed to bypass this restriction using Reflection, but wouldn't want to do this in production obviously:
```csharp
// Work around because Query.Queries, a List<QueryInfo>, is internal
private IQueryContext queryContext()
{
    var fields = query.GetType().GetFields(BindingFlags.Instance | BindingFlags.NonPublic);
    var l = fields.Single(a => a.Name == "Queries").GetValue(query);
    var casted = l as System.Collections.IList;
    return casted[QueryNumber] as IQueryContext;	
}
```

I also had problems where the parameters didn't appear to being set, and were always null.

---

[...]

PR #1926, opened 2019-10-10
Example code, showing issue because of internal access #1915

davidglassborow:
Details for #1915

I've tried to show the issue, but because the tests have internal access, they have access to implementation details that a user's code would not have

### DIFF
--- a/ApiTest/ApiTest.csproj
+++ b/ApiTest/ApiTest.csproj
@@ -1,0 +1,62 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{ED853FDC-C869-4E35-8012-7E91D06166F6}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>ApiTest</RootNamespace>
+    <AssemblyName>ApiTest</AssemblyName>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <Deterministic>true</Deterministic>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\Tests\Base\DataProvider\RemoteData.cs">
+      <Link>RemoteData.cs</Link>
+    </Compile>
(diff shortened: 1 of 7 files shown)


davidglassborow:
I've added a test project, ApiTest, that shows the issue by being unable to compile

davidglassborow:
Sorry it was all a bit rushed last night.

I've added a file in Tests/Base/ApiTestDataContext.cs which attempts to implement a IDataContext and IQueryRunner.  I can get it to compile inside the tests.

I've also added a simple console project which references Linq2DB, and the ApiTestDataContext.cs file, which won't compile because of the need for internal access for getting SQL text, etc.

sdanyliv:
I propose to write your RemoteDataContext in our Playground project. Closer to finish, we will decide which API needs to be exposed.
Also you can change linq2db by yourself in your branch - and it will be visible for us which API needs opening.

davidglassborow:
So it's in both the playground project (and compiling ok), as well as an external project (which won't compile because of internal APIs).  

It already is a proof of concept that can correctly get the optimised SQL, and parameters, so that it can run them remotely, or locally with caching, etc.

I've only tried to implement `int IQueryRunner.ExecuteNonQuery()` as a base test.

Is there something else you need me to do ?

sdanyliv:
Moment, you just need optimized SQL and prepared parameters? I can do that in our source code without implementing another DataContext.

davidglassborow:
Yes, I need a way of having DataContext that delegates the actually running of SQL commands to something else, just like the WCF implementation you already have.

To do that I need the optimised SQL (so I'm happy for the local data context to know what the end target database is so the SQL is correct, in my case this is SQL Server).

Once I have the SQL and the parameters, I'll be using some of my own code that handles caching, throttling, etc.

It looks like the public interface of IQueryRunner should be sufficient to do that, but trying to implement it has lead me to believe that it is not possible to implement this interface outside of the core code base.

Hope that makes sense.

davidglassborow:
I'm using SQL Server.  The implementation will be in my own code base (which is largely F#), which will either be cached local calls, or remote calls over websockets.

davidglassborow:
Essentially I just need an extension API so I can write something similar to your WCF implementation, without needed access to all the internals.

davidglassborow:
@sdanyliv do you need anything else from me to proceed ?

sdanyliv:
Looking into, sorry for delay.

davidglassborow:
no worries, just shout if there is anything I can do to help 👍

sdanyliv:
Why you have not derived `ApiTestDataContext` from `RemoteDataContextBase`?

sdanyliv:
Or even better i can create additional SqlRemoteDataContextBase which generate SQL and prepare parameters, and rest is just send SQL to remote endpoint.

davidglassborow:
That sounds ideal.  I'd expect the SqlRemoteDataContextBase would need to have the DataProvider specified somehow so correctly optimised SQL is generated, right ?

sdanyliv:
Ask yourself about the following:

- Who should control configuration (connection string, DataProvider) Client, or Server?
- How Server should know about ConnectionString
- When query is generated on the Client - client should know about concrete DataProvider
- It is safe to send SQL via REST? If it is intercepted you can have a lot of problems :) Especially `DROP DATABASE`

davidglassborow:
So in my case:

- The server knows the connection string, the client needs to know the Data Provider
- Yes its all safe

Another case I have might be another way to think of this: to provide a cache where all data is cached for e.g. 5 minutes:

- linq2db passes my code the SQL and parameters it want's be to run
- I check the cache, if not present I create my own connection, command, set the parameters and execute the reader, loading it into a `DataTable` and storing it in the cache
- I then use `DataTable.CreateDataReader` and return the reader to linq2db to map into objects
- Next time the same request comes in, I already have the DataTable ready

Obviously the cache would be keyed against SQL and parameters.

davidglassborow:
In my case I already have a cache that does much more complicated behaviour - we have 1000s of users so we also do throttling of concurrent queries, complex cache invalidation behaviour etc.

I'm looking for an extension point in linq2db that essentially:

- I tell it what DataProvider I want it to generate SQL for
- It calls my code when it needs SQL executing with the SQL text + parameter definitions, and I return a DataReader

This would allow my caching, throttling scenarios, as well as remote data calls over websockets, etc.

sdanyliv:
> I tell it what DataProvider I want it to generate SQL for

Something like this?
```cs
var query = from t in db.GetTable<SomeTable>();

query.GenerateSql(dataProvider, out var sql, out var parameters);

// we can reuse provider information if RemoteDataContext is properly configured
query.GenerateSql(out var sql, out var parameters);
```

> It calls my code when it needs SQL executing with the SQL text + parameter definitions, and I return a DataReader

Are you talking about current linq2db implementation, or with your RemoteContext?

davidglassborow:
That is the sort of thing, although that only provides sql and parameters, I would need a way of returning the data so linq2db can do the mapping.

I'm happy to implement the public IQueryData interface in underlying code, but currently it doesn't provide enough access to actually implement it.  So if this interface provides the raw underlying SQL and parameters fill it correctly, I can implement it.

sdanyliv:
Oh, i messed up :)

> I'm happy to implement the public IQueryData interface in underlying code, but currently it doesn't provide enough access to actually implement it. So if this interface provides the raw underlying SQL and parameters fill it correctly, I can implement it.

It is releated to RemoteContext? If not you can intercept `ExecuteReader` by yourself on `DataConnection` level or by providing your realization of `IDbCommandProcessor` for `DbCommandProcessorExtensions.Instance`

davidglassborow:
Ah that looks like it might do the job !  Both could work for me 👍 

I'll close this and reopen if I have any further problems.

As an aside, it might be worth considering making some of the interfaces like IQueryRunner be internal only if they are not meant to be implemented by callers.

Thanks for all the help, appreciate it !

davidglassborow:
One issue is the DataConnection.ExecuteReader override only gives an IDbCommand (which has opaque Parameters as objects), where as IDbCommandProcessor provides DbCommand (which does have proper parameter objects with names etc).

Can the DataConnection overrides provide DbCommand rather than IDbCommand ?

sdanyliv:
I do not see any objectives in casting `IDbCommand` to `DbCommand`

davidglassborow:
I can do that, it's just that it's not type safe which I personally prefer not to do.  Is there any reason for the different between the 2 overload points ?

sdanyliv:
I see also mess here
https://github.com/linq2db/linq2db/blob/5e26fa1693208b92b0791dff445ab7b42e79fb1b/Source/LinqToDB/Data/DbCommandProcessor/DbCommandProcessorExtensions.cs

We will review this for 3.0.
Cast to (DbCommand)cmd and runtime will throw exception if casting is not allowed.

davidglassborow:
Will do.  I understand your pain, IDbCommand and related interfaces don't seem well designed, I always end up needed DbCommand to get proper functionality.

PR #1930, opened 2019-10-13
Trimming ColumnInfo.DataType & ColumnType

ili:
Schema provider can read type with spaces

Ex: SQLite can return "varchar " (this depends on database...) and corresponding model property is generated with `object` system type, instead of `string`.


Source/LinqToDB/SchemaProvider/ColumnInfo.cs
```diff
@@ -6,12 +6,15 @@ namespace LinqToDB.SchemaProvider
...
+		private string _dataType;
+		private string _columnType;
+
 		public string TableID;
 		public string Name;
 		public bool   IsNullable;
 		public int    Ordinal;
-		public string DataType;
-		public string ColumnType;
+		public string DataType      { get => _dataType;   set => _dataType   = value.Trim(); }
```
sdanyliv: Bad place for this fix.

ili: Fixes all possible cases :)
Another solution is to make trim from schema providers, doable, but this looks better for me

Issue #1899, opened 2019-09-26
Oracle "LONG" column value not selected in join


DonKonfetti:
Hi!

I've got a query that yields a wrong result for a value from an Oracle LONG column under certain conditions:

```c#
var query = from k in _db.F2cfgtree
    join v in _db.F2cfgvalues on k.IdF2cfgtree equals v.IdF2cfgtree
    where k.IdF2mandant.Equals(_mandantId) && k.Hkey == keyType && (k.Name ?? "<null>").Equals(keyValue) && k.Key.ToUpper().Equals(key.ToUpper()) && v.Name.ToUpper().Equals(name.ToUpper())
    select v.ValueL;
var values = await query.Take(2).ToArrayAsync();
```

This always yields an empty string although the actual VALUE_L column contains the string "LOOONG". Above query is generating the following SQL statement:

```SQL
SELECT
	v.VALUE_L
FROM
	F2cfgtree k_1
		INNER JOIN F2cfgvalues v ON k_1.ID_F2CFGTREE = v.ID_F2CFGTREE
WHERE
	ROWNUM <= 2 AND k_1.ID_F2MANDANT = 0 AND
	k_1.HKEY = 1 AND
	Nvl(k_1.NAME, '<null>') = '<null>' AND
	Upper(k_1.KEY) = Upper('AAA\BBB\CCC') AND
	Upper(v.NAME) = Upper('Text-Long');
```

This query, executed with SQL Developer, yields the correct result for VALUE_L.

The VALUE_L column mapping looks like this:

```c#
[Column("VALUE_L",        DbType="LONG",          DataType=DataType.Text,     Length=0),    Nullable]
public string ValueL { get; set; } // LONG
```

When I select all rows of the F2cfgvalues table with the following query, the correct VALUE_L values are being returned for all rows.

```c#
var query = from r in _db.F2cfgvalues select r;
var allRows = await query.ToArrayAsync();
```

So, the issue only appears with my first query. Any idea what I am doing wrong?

### Mapping

Full mapping of all involved tables is this:

```c#
[NormalizedIdColumn("IdF2cfgtree")]
[Table("F2cfgtree")]
public partial class F2cfgtree : VasDbTableWithRidVid, IHasInsertUpdateInfo
{
    [Column("ID_F2CFGTREE", DbType="NUMBER (12,0)", DataType=DataType.Decimal, Length=22, Precision=12, Scale=0, SkipOnUpdate=true), PrimaryKey, NotNull]
[...]

PR #1931, opened 2019-10-14
Fix for #1927. Implemented support of LONG and LONG RAW datatypes in Oracle.

### LINKED ISSUE #1927: Trying to insert Oracle LONG string value with length > 4000 fails
Tyring to insert a string value with length greater than 4000 into a LONG column fails with ORA-22835 (Buffer too small for CLOB to CHAR or BLOB to RAW conversion).

### Steps to reproduce

I can reproduce this issue using OracleConnection/OracleCommand:

```c#
OracleConnection con = new OracleConnection(<connection string>);
con.Open();

var query = "INSERT into TBL_WITH_LONG_COLUMN (LONG_COLUMN)  VALUES (:1)";
​
var cmd = new OracleCommand(query, con);
cmd.CommandType = CommandType.Text;
​
OracleParameter param = cmd.Parameters.Add(":1");
param.Direction = ParameterDirection.Input;
​param.Value = new String('A', 10000);
​
cmd.ExecuteNonQuery();
con.Dispose();
```

This fails with aforementioned Oracle error, but works after explicitely specifying the command parameter type being OracleDbType.Long:

```c#
OracleParameter param = cmd.Parameters.Add(":1", OracleDbType.Long);
```

So, is there a way to tell Linq2DB to explicitely specify the parameter type for LONG column values?

### Environment details

linq2db version: 2.9.0
Database Server: Oracle 12.1.0
Database Provider: Oracle.ManagedDataAccess.Core 2.19.31
Operating system: Windows 10
.NET Framework: .NET Core 2.2

sdanyliv:
Fixes #1927 and #1899

### DIFF
--- a/Data/Create Scripts/Oracle.sql
+++ b/Data/Create Scripts/Oracle.sql
@@ -748,6 +748,7 @@ CREATE TABLE AllTypes
 	binaryDataType           blob                           NULL,
 	bfileDataType            bfile                          NULL,
 	guidDataType             raw(16)                        NULL,
+	longDataType             long                           NULL,
 
 	uriDataType              UriType                        NULL,
 	xmlDataType              XmlType                        NULL
@@ -800,6 +801,7 @@ INSERT INTO AllTypes
 	binaryDataType,
 	bfileDataType,
 	guidDataType,
+	longDataType,
 
 	uriDataType,
 	xmlDataType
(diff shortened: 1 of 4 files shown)


MaceWindu:
reveted change to InitialLOBFetchSize, as it is not needed and leads to unexpected side-effects on empty blobs

Issue #672, opened 2017-05-21
GroupBy - not supported exception


integral-llc:
Given such a LINQ query 
```cs
IQueryable<GFDDataModels.Stone> stones;
stones = from s in _dataContext.Stone where s.Enabled == true select s;

stones = from s in stones
              where !s.Name.StartsWith("level - ") && s.ImageFullUrl.Length > 0
              group s by s.Name
              into sG
              select sG.First();

var list = stones.ToList();
```
it fails with the exception.
`"The method or operation is not implemented."`

Stack Trace:
```
   at LinqToDB.Linq.Builder.GroupByBuilder.GroupByContext.ConvertToSql(Expression expression, Int32 level, ConvertFlags flags) in c:\projects\linq2db\Source\Linq\Builder\GroupByBuilder.cs:line 520
   at LinqToDB.Linq.Builder.SelectContext.<>c__DisplayClass21.<ConvertToSql>b__13(IBuildContext ctx, Expression ex, Int32 l) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 284
   at LinqToDB.Linq.Builder.SelectContext.ProcessScalar[T](Expression expression, Int32 level, Func`4 action, Func`1 defaultAction) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 890
   at LinqToDB.Linq.Builder.SelectContext.ConvertToSql(Expression expression, Int32 level, ConvertFlags flags) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 281
   at LinqToDB.Linq.Builder.SelectContext.<>c__DisplayClass21.<ConvertToSql>b__13(IBuildContext ctx, Expression ex, Int32 l) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 284
   at LinqToDB.Linq.Builder.SelectContext.ProcessScalar[T](Expression expression, Int32 level, Func`4 action, Func`1 defaultAction) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 917
   at LinqToDB.Linq.Builder.SelectContext.ConvertToSql(Expression expression, Int32 level, ConvertFlags flags) in c:\projects\linq2db\Source\Linq\Builder\SelectContext.cs:line 281
[...]

PR #1945, opened 2019-10-22
Tests for issue 672.

GonzoSpire:
Tests for issue #672 passed.


MaceWindu:
Actually, issue is not completely fixed. There is no exception anymore, but each group loaded by separate query

Issue #1979, opened 2019-11-15
Problem with Association and Derived Class (and also without)


jogibear9988:
Describe your issue.

My Query in My Pull reqest does not work.

```
Exception message:
LinqToDB.LinqToDBException : Second parameter of expression predicate should be 'Tagging'
```

if I change the 

			public List<TaggingIssue> Tagging { get; set; }

to

			public List<Tagging> Tagging { get; set; }

I get the following exception:
    System.InvalidCastException : Das Objekt des Typs "System.Linq.Expressions.MethodCallExpressionN" kann nicht in Typ "System.Linq.Expressions.MemberExpression" umgewandelt werden.

### Steps to reproduce
Look at my Pull Request

jogibear9988:
see https://github.com/linq2db/linq2db/pull/1980

jogibear9988:
Is my mapping for the Association wrong?

jogibear9988:
@sdanyliv could you help me with this?

MaceWindu:
Association expressions had incorrect type. Also it is possible that changes in rc0 regarding associations also helped

PR #1980, opened 2019-11-15
Test for #1979

jogibear9988:


sdanyliv:
What are you trying to achieve here?
Getting records from table `issues` and use association for table `taggings`? I can not get whole idea of this example.

jogibear9988:
I expect it to create a query like this...

	select users.lastname, tags.name from issues
	left join users on issues.assigned_to_id = users.id
	left join (select * from taggings where taggable_type = 'Issue' ) as ab on ab.taggable_id = issues.id
	left join tags on tags.id = ab.tag_id
	where tags.name = 'Visu'

jogibear9988:
@sdanyliv is this possible? or what did i do wrong?

sdanyliv:
@jogibear9988, i don't think so. 
Try to write real query using linq, without associations maybe it may help to understand how to do that properly.

jogibear9988:
I did this...

But I think it's not the same...

        var query = from i in da.Issues
                    join t in (da.Taggings.Where(x => x.TaggableType == "Issue")) on i.Id equals t.TaggableId into tj
                    from t in tj.DefaultIfEmpty()
                    join tg in da.Tags on t.TagId equals tg.Id into tgj
                    from tg in tgj.DefaultIfEmpty()
                    join u in da.Users on i.AssignedToId equals u.Id into uj
                    from u in uj.DefaultIfEmpty()
                    where tg.Name == "Visu"
                    select new { Issue = i, User = u };

jogibear9988:
Cause I only want each issue once, and so it could be returned more often.

sdanyliv:
Something like this?
```cs
 var query = 
    from i in da.Issues
    from u in da.Users.LeftJoin(u => u.AssignedToId == u.Id)
    from ab in (da.Taggings.Where(x => x.TaggableType == "Issue")).LeftJoin(ab => ab.TaggableId == i.Id)
    from tg in da.Tags.LeftJoin(tg =>tg.Id == ab.TagId) 
    where tg.Name == "Visu"
    select new { Issue = i, User = u };
```

One note, if you have filter `tg.Name == "Visu"`, you do not need LeftJoin

jogibear9988:
And is this possible with my associations?

jogibear9988:
@sdanyliv I've added the query I expect...

MaceWindu:
- reduced test code to contain only required details
- added linq-based test to demonstrate expectations
- fixed couple of isses with association test, but it still fails, I suspect something missing in handling of nested association contexts

@jogibear9988 in future, try to simplify tests to show linq2db issue, not your application issue :smile:

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -1420,7 +1395,7 @@ select om
 					isNew = true;
 				}
 
-				if (tableAssociation == null && levelExpression.NodeType == ExpressionType.MemberAccess && objectMapper.Associations.Count > 0 || inheritance.Count > 0)
+				if (tableAssociation == null && levelExpression.NodeType == ExpressionType.MemberAccess && (objectMapper.Associations.Count > 0 || inheritance.Count > 0))
```
MaceWindu: Looks like github failed to build diff properly, I've changed only this line in this file

MaceWindu:
@jogibear9988 , I've fixed incorrectly-typed `QueryExpressionMethod`'s (starting from RC0 it gives decsriptive error message) and now tests work (probably due to changes in associations in rc0)

@sdanyliv could you check association-based query - is it expected generation? It generates APLLY query (so it doesn't work with sqlite now)

PR #1996, opened 2019-12-07
#1925 and #776 Repro Test and fix

### LINKED ISSUE #776: Don't escape my Escape Character
A manually attached LIKE clause `Sql.Like(f.FIELDNAME, "slot[_]__")` will be translated to `[f].[FIELDNAME] LIKE N'slot[[]_]__'`. What I want is match contents like "slot_01", "slot_02" and so on. I can't match with a `Sql.Like(f.FIELDNAME, "slot%")` or `f.FIELDNAME.StartsWith("slot_")` ...

Is there any option to disable the auto-escaping?

Edit:
Provider is Microsoft SqlServer.

Thanks.

### LINKED ISSUE #1925: Access LIKE wildcards not supported?
Access supports different wildcard characters with LIKE:
https://support.office.com/en-us/f1/topic/examples-of-wildcard-characters-939e153f-bd30-47e4-a763-61897c87b3f4

Specifically looking for the digit wildcard '#'.

I tried these two:
```C#
Projects.Select(p => p.ProjectManagerLookup).Where(p => Sql.Like(p, "#%"))
//and
Projects.Select(p => p.ProjectManagerLookup).Where(p => p.StartsWith("#"))
```
Which produces this SQL:
```SQL
SELECT
	[p].[Project Manager Lookup]
FROM
	[Projects] [p]
WHERE
	[p].[Project Manager Lookup] LIKE '#%'
```

but finds no results. However, if I copy the generated SQL and paste it into an Access query, it'll run and find the results.

`Regex.IsMatch` gives the "Cannot be converted to SQL" error, understandably.

#### Environment details
linq2db version: 2.9.1 with LINQPad 5.4
Database Provider: Access

to11mtm:
Fix #776
Fix #1925

### DIFF
--- a/Source/LinqToDB/DataProvider/DataTools.cs
+++ b/Source/LinqToDB/DataProvider/DataTools.cs
@@ -8,54 +8,46 @@ namespace LinqToDB.DataProvider
 	public class DataTools
 	{
 		/// <summary>
-		/// use in place of .Replace("[", "[[]")
+		/// Improved version of <c>Replace("[", "[[]")</c> code, used before.
 		/// </summary>
-		/// <param name="str"></param>
-		/// <returns></returns>
 		public static string EscapeUnterminatedBracket(string str)
 		{
 			if (str == null)
-			{
 				return str;
-			}
+
 			var nextIndex = str.IndexOf('[');
 			if (nextIndex < 0)
-			{
 				return str;
-			}
+
 			var lastIndex = 0;
-			StringBuilder newStr = new StringBuilder(str.Length + 10);
+			var newStr = new StringBuilder(str.Length + 10);
 
 			while (nextIndex >= 0)
 			{
 				if (nextIndex != 0)
-				{
 					newStr.Append(str.Substring(lastIndex, nextIndex - lastIndex));
-				}
 
 				lastIndex = nextIndex;
 
 				var closeBracket = str.IndexOf(']', nextIndex + 1);
-				nextIndex = str.IndexOf('[', nextIndex + 1);
+				nextIndex        = str.IndexOf('[', nextIndex + 1);
+
 				if ((closeBracket > 0 && (closeBracket < nextIndex || nextIndex < 0))
 				    ||
 				    (closeBracket - lastIndex == 2 && closeBracket - nextIndex == 1))
 				{
 					if (nextIndex < 0)
-					{
 						newStr.Append("[");
-					}
 				}
 				else
-				{
 					newStr.Append("[[]");
-				}
 
 			}
 
 			newStr.Append(str.Substring(lastIndex + 1));
 			return newStr.ToString();
 		}
+
 		static readonly char[] _escapes = { '\x0', '\'' };
 
 		public static void ConvertStringToSql(
--- a/Source/LinqToDB/DataProvider/DataTools.cs
+++ b/Source/LinqToDB/DataProvider/DataTools.cs
@@ -30,13 +30,16 @@ namespace LinqToDB.DataProvider
 			{
 				if (nextIndex != 0)
 				{
-					newStr.Append(str.Substring(lastIndex, nextIndex-lastIndex));
+					newStr.Append(str.Substring(lastIndex, nextIndex - lastIndex));
 				}
+
 				lastIndex = nextIndex;
 
 				var closeBracket = str.IndexOf(']', nextIndex + 1);
 				nextIndex = str.IndexOf('[', nextIndex + 1);
-				if (closeBracket > 0 && (closeBracket < nextIndex || nextIndex < 0))
+				if ((closeBracket > 0 && (closeBracket < nextIndex || nextIndex < 0))
+				    ||
+				    (closeBracket - lastIndex == 2 && closeBracket - nextIndex == 1))
 				{
 					if (nextIndex < 0)
 					{
@@ -47,8 +50,9 @@ namespace LinqToDB.DataProvider
 				{
 					newStr.Append("[[]");
 				}
-				
+
 			}
+
 			newStr.Append(str.Substring(lastIndex + 1));
 			return newStr.ToString();
 		}
--- a/Source/LinqToDB/DataProvider/DataTools.cs
+++ b/Source/LinqToDB/DataProvider/DataTools.cs
@@ -30,20 +30,24 @@ namespace LinqToDB.DataProvider
 			{
 				if (nextIndex != 0)
 				{
-					newStr.Append(str.Substring(lastIndex, nextIndex));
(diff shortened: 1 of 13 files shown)


to11mtm:
This is also related to #776 .

Still need to write full tests around the new code but idea is that we will only escape bracket if we can't find a terminating bracket in the 'right place'. I'm also trying to not be careless with allocations here.

Source/LinqToDB/DataProvider/DataTools.cs
```diff
@@ -7,6 +7,55 @@ namespace LinqToDB.DataProvider
...
+					newStr.Append(str.Substring(lastIndex, nextIndex - lastIndex));
+				}
+
+				lastIndex = nextIndex;
+
+				var closeBracket = str.IndexOf(']', nextIndex + 1);
+				nextIndex = str.IndexOf('[', nextIndex + 1);
+				if ((closeBracket > 0 && (closeBracket < nextIndex || nextIndex < 0))
+				    ||
+				    (closeBracket - lastIndex == 2 && closeBracket - nextIndex == 1))
```
to11mtm: If we don't want to keep `[[]` as-is when we run into it, we can take this line out.

Source/LinqToDB/DataProvider/DataTools.cs
```diff
@@ -7,6 +7,55 @@ namespace LinqToDB.DataProvider
 {
 	public class DataTools
 	{
+		/// <summary>
+		/// use in place of .Replace("[", "[[]")
+		/// </summary>
+		/// <param name="str"></param>
+		/// <returns></returns>
+		public static string EscapeUnterminatedBracket(string str)
```
to11mtm: If there is a better place for this please let me know. I put it here since that is where we have some other string escaping code.

Tests/Tests.Playground/Issue1925Tests.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using LinqToDB;
+using LinqToDB.Mapping;
+using NUnit.Framework;
+
+namespace Tests.Playground
+{
+	[TestFixture]
+	public class TestAccess1925 : TestBase
```
to11mtm: Should I move this test out of here and create copies for SQLServer / Sybase / Access? (Since this change impacts those three DBs)

MaceWindu: No, just add additional providers to test data sources or more tests, if they use different test-cases

MaceWindu:
Guys, I'm not sure it is a move in right direction. I think situation, where we mess with `like` pattern, passed by user is a bug itself - we shouldn't alter it at all.

We only should perform escaping, when we convert string's Contains/StartsWith/EndsWith method calls to SQL `LIKE`.

I think proper fix should be:
- make sure we don't mess with pattern if `Sql.Like` called
- make sure that string's search methods escape searched string properly for cases, mentioned in addressed issues

MaceWindu:
We also have same issue reported before: #776
So We can close it with this pr too.

Source/LinqToDB/DataProvider/DataTools.cs
MaceWindu: that's fine (if we still need this method)

to11mtm:
@MaceWindu  I put some thought to your comments but wanted to be sure I understood intent.

> We only should perform escaping, when we convert string's Contains/StartsWith/EndsWith method calls to SQL LIKE.

> -  make sure we don't mess with pattern if Sql.Like called
> -  make sure that string's search methods escape searched string properly for cases, mentioned in addressed issues

Does that mean that we should never escape bad brackets on `Sql.Like` input? see below for examples.

```
Assert.AreEqual(0, table.Where(r => r.Value.Contains("[0")).ToList().Count());
//generated: 	[r].[Value] LIKE '%[[]0%'

Assert.AreEqual(1, table.Where(r => r.Value.Contains("[0-9]")).ToList().Count());
//generated: 	[r].[Value] LIKE '%[0-9]%'

Assert.AreEqual(0, table.Where(r => Sql.Like(r.Value, "[0")).ToList().Count());
//generated: 	[r].[Value] LIKE '[0'
//throws: System.Data.OleDb.OleDbException : Invalid pattern string

Assert.AreEqual(1, table.Where(r => Sql.Like(r.Value, "[0-9]")).ToList().Count());
//generated: 	[r].[Value] LIKE '[0-9]'
```

If that is the way we want to go, I have a local commit that does this and can push to this branch. 
Do note this will be a minor change to the LinqServiceSerializer : We have to add a boolean flag to `SqlPredicate.Like` to know whether the generated sequence was generated from a LIKE predicate or a Contains/StartsWith/EndsWith method call.

I'm not sure it makes sense to do in `ExpressionBuilder.SqlBuilder` since the need for escaping is (To the best my my knowledge) only needed for Access/MSSQL/Sybase.

Please let me know if I should push this change up for review.

MaceWindu:
Yes, that's correct - if user pass invalid template to `Like` - he should get error.

to11mtm:
@sdanyliv I took a look at #2005 

I think logic in `ExpressionBuilder.SqlBuilder.ConvertLikePredicate(IBuildContext context, MethodCallExpression expression, string start, string end)` will have to be pushed into the BasicSQLBuilder because we handle it with code like this:

```
new SqlPredicate.Like(o, false, new SqlValue(start + value + end), null, false):
```

and the Firebird `STARTING WITH` does not want the tokens.

While I can do that in this PR, I think it will make size of change a decent bit larger (because we will have to move parts of `ExpressionBuilder.SqlBuilder` into `BasicSQLBuilder`).

I can start to look into it if we think it is better to do this all in one change instead of separate.

sdanyliv:
It was my previous idea how to fix Like problems. 
There are three string predicates
StartsWith,
Contains
EndsWith

And they can be processed differently on different database engines.
For most cases it is generating Like predicate with proper escaping, but as you can see Firebird’s optimizer do not analyze like expression and proceed with table scan.

So i propose to create SqlStingPredicate (or something like this) with enumeration (StartsWith, Contains, EndsWith) and create appropriate SqlBuilder functions for generating this predicate.

Does it makes sense?

MaceWindu:
I think we can merge it to 2.9.5 as fix for existing issue and implement new predicate as separate improvement. Any objections?

to11mtm:
> I think we can merge it to 2.9.5 as fix for existing issue and implement new predicate as separate improvement. Any objections?

@MaceWindu I'm good with that, I won't have more time to look into fixing #2005 until January.

MaceWindu:
I will add #2005 here

PR #1998, opened 2019-12-09
[fix] Nullable parameter in conversion operator

IgorBuchelnikov:
If conversion operator contains Nullable<T> parameter:

> System.InvalidOperationException: The operands for operator 'Convert' do not match the parameters of method 'op_Explicit'.


MaceWindu:
I've requested db-related test functionality removal, but want to note that:
- if we add new test objects, they should be added to db creation scripts
- it is not recommended to create new global test tables/columns without good reason

Better approach for such tests is to create local table in test itself, using CreateLocalTable test api with type-scoped mappings: [example](https://github.com/linq2db/linq2db/blob/master/Tests/Linq/UserTests/Issue1869Tests.cs#L129)

IgorBuchelnikov:
Done. Writing test faced to problems related with inserting test row:
1) For MS Access data provider the only way to define conversion from CustomMoneyType to decimal is make CustomMoneyType implements IConvertible. That interface does not provide conversion to Nullable<decimal> (only to decimal). Moreover conversion to decimal is perfomed throuth ToString(IFormatProvider provider) member. I did not find a way to insert null value. Seems returning null from ToString(IFormatProvider provider) is treated as exclusion of parameter (not as null parameter value). 
2) For other providers conversion from CustomMoneyType to decimal behaviour is diffent. Some providers call ToDecimal(IFormatProvider provider) or GetTypeCode(). May be there is a way to insert null for somу providers.

MaceWindu:
Rounding issue possibly cames from CreateTable generating type name without decimal part size set. This could be fixed by specifying [Scale](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Mapping/ColumnAttribute.cs#L293) for Column attribute.

Tests/Linq/Common/ConvertTests.cs
```diff
@@ -461,5 +463,20 @@ public void ConvertFromNullableEnum2()
 			Assert.AreEqual(0,    ConvertTo<int>. From((Enum15?)null));
 			Assert.AreEqual(null, ConvertTo<int?>.From((Enum15?)null));
 		}
+	
+		[Test]
+		public void NullableParameterInOperatorConvert([IncludeDataSources(ProviderName.Access)] string context)
+		{
+			using (var db = GetDataContext(context))
```
MaceWindu: Why do you test against database? Convert class could be tested without db involved (which will make things easier, as you don't need to struggle with Access)

Tests/Model/TestDataConnection.cs
```diff
@@ -65,6 +65,7 @@ void TestDataConnection_OnClosing(object sender, EventArgs e)
 		public ITable<GrandChild1>            GrandChild1            => GetTable<GrandChild1>();
 		public ITable<LinqDataTypes>          Types                  => GetTable<LinqDataTypes>();
 		public ITable<LinqDataTypes2>         Types2                 => GetTable<LinqDataTypes2>();
+		public ITable<LinqDataTypes3>         Types3                 => GetTable<LinqDataTypes3>();
```
MaceWindu: Don't add new members here. Better to use GetTable call in test itself.

Tests/Model/TestServiceModelDataContext.cs
```diff
@@ -44,6 +44,7 @@ public class TestServiceModelDataContext : ServiceModelDataContext, ITestDataCon
 		public ITable<GrandChild1>            GrandChild1            { get { return this.GetTable<GrandChild1>();            } }
 		public ITable<LinqDataTypes>          Types                  { get { return this.GetTable<LinqDataTypes>();          } }
 		public ITable<LinqDataTypes2>         Types2                 { get { return this.GetTable<LinqDataTypes2>();         } }
+		public ITable<LinqDataTypes3>         Types3                 { get { return this.GetTable<LinqDataTypes3>();         } }
```
MaceWindu: Same here, don't extend generic test infrastructure with one-test specific functionality

Tests/Model/LinqDataTypes.cs
```diff
@@ -136,4 +136,102 @@ public override string ToString()
 			return string.Format("{{{0,2}, {1,7}, {2:O}, {3,5}, {4}, {5}, '{6}'}}", ID, MoneyValue, DateTimeValue, BoolValue, GuidValue, SmallIntValue, StringValue);
 		}
 	}
+
+	[Table("LinqDataTypes")]
```
MaceWindu: This class should belong to class with test

Tests/Linq/Common/ConvertTests.cs
IgorBuchelnikov: I tested according [examle ](https://github.com/linq2db/linq2db/blob/master/Tests/Linq/UserTests/Issue1869Tests.cs#L129) you [provided](https://github.com/linq2db/linq2db/pull/1998#issuecomment-563307005)

Tests/Model/TestDataConnection.cs
IgorBuchelnikov: done

Tests/Model/TestServiceModelDataContext.cs
IgorBuchelnikov: done

Tests/Model/LinqDataTypes.cs
IgorBuchelnikov: done

sdanyliv:
@IgorBuchelnikov, what are you trying to achieve here? It can be easily done by configuring MappingSchema.

IgorBuchelnikov:
[> Rounding issue possibly cames from CreateTable generating type name without decimal part size set. This could be fixed by specifying [Scale](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Mapping/ColumnAttribute.cs#L293) for Column attribute.](https://github.com/linq2db/linq2db/pull/1998#issuecomment-564038422)

Setting of Scale attribute does't work.

Tests/Linq/Common/ConvertTests.cs
MaceWindu: `Convert<decimal?, CustomMoneyType>.From(1.11m)` should work as test code I think

IgorBuchelnikov:
https://github.com/linq2db/linq2db/pull/1998#issuecomment-564067594

I fix linq2db internal error. ConvertBuilder could not construct expression for the test provided.

> System.InvalidOperationException: The operands for operator 'Convert' do not match the parameters of method 'op_Explicit'.

IgorBuchelnikov:
https://github.com/linq2db/linq2db/pull/1998#issuecomment-564067594

Please, show me how to fix it by configuring MappingSchema

sdanyliv:
Not tested, but something like this:
```cs
var ms = new MappingSchema();

ms.SetConvertExpression<decimal, CustomMoneyType>(d => new CustomMoneyType(d));

ms.SetConverter<CustomMoneyType, DataParameter>(c => 
{
     return new DataParameter(c.Amount)
})

var dc = new DataConnection(ms);
```

sdanyliv:
Ok, i see.

PR #2019, opened 2019-12-28
DbContext with linq2sql classes

alexey-tyulkin:
DbContext can work with linq2sql  autogenerated classes from dbml via SystemDataLinqAttributeReader. I wrote some bug fix for SystemDataLinqAttributeReader and
some tests.

PS I do not have VS 2017 Enterprise so do not run Build.cmd - my apologies for this.


sdanyliv:
Thanks for the fix!
BTW, you do not need VS Enterprise version, Community edition is enough.
Current master branch builds on VS 2017 and VS 2019. Starting from version 3.0 it can be built only on VS 2019

PR #2059, opened 2020-01-26
Release 2.9.6

MaceWindu:
- run tests
    - [x] netfx
    - [x] netcore1
    - [x] netcore2

@linq2db/testers we plan to release on thursday (30/01)

Right now release miss minor T4 fix #2057 (will be merged bit later) and we have two CTE-related issues open, which also will be fixed later if additional details will be provided by reporter in time

[release notes](https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#release-296)


detoxhby:
Cannot find nighty/rc packages' feed for 2.x branch on MyGet... has something changed? Can we still have a direct feed for them or should I instead download the artifacts from appveyor and use a local repo?

detoxhby:
I've found some strange tests in different projects upgrading from 2.7.4. The result are affected (and the same) with 2.9.5 as well, so it's most likely not version specific to this release.

For now, I had time only to debug the main part of the problem, but it seems like prior to 2.7.5, it has done some horribly optimistic group-join optimization on nested query. From current data, it seems that the **whole** group join that was applied on an inline subquery (which also had a join in it) was completelly cut out, even though the result of that query would affect the main query's join and therefore the result set.

Something like this (just for clarification):
```
from x in tableWithSomeLoadWiths
join y in
    (from xx in PivotTable1
     join yy in SomeDataTable on xx.Id equals yy.OtherId
     select new { Id = xx.XId, Data = yy.SpecData }) on x.Id equals y.Id into yyJoinResults
where x.Id == @outerParameterId
select new { ... , Y = yyJoinResults.Select(x => x.Data)) }
```
(the group join join part was completelly optimized out prior to 2.7.5; but generates correct SQL now)

Long story sort, it seems like the new versions has the correct behavior and the tests are incorrect but were working due to previous versions join behavior. Will report back with the details before thursday.

sdanyliv:
Glad to hear that. 
Also looks like you use LoadWith extremely. Can you run tests against this branch? https://github.com/linq2db/linq2db/pull/1756
I’ve almost done with Eager Loading, but it may cause regressions.

MaceWindu:
Not sure yet why myget deploy broken. Looks like we didn't publish anything new there for 6 monthes...

detoxhby:
✅ Fixed the tests/queries to follow the new behavior, seems okay now.

It is really hard to believe that previously the whole join was missing. Our developers was lazy to realize a `group by` would be necessary for these cases where the group join would affact the main result set ... but as most _"I don't care if it works"_ type of developers, they have just gone with it. Sad story.

Issue #2064, opened 2020-01-28
Exception when debugging query in VS2017


RoyChase:
This looks to be the same issue as #1795 but that wasn't really resolved.

If, using VS2019 targeting .Net Framwork I put a breakpoint on an IQueryable then stepping over it gives a TypeInitializationException.  The same code running in VS2017 works fine.

If targeting .Net Core it works fine.

Without the breakpoint it works fine in all cases.

I have created the simplest demo code that I can https://github.com/RoyChase/DebugIssue.git which contains a .Net Framework console app and a .Net Core console app.

I agree that it look sot be something that MS has done with VS2019 but I'm hoping that you might be able to trace the issue deeper into the Linq2DB source code to be able to isolate it further and perhaps even make a fix.  Maybe System.Diagnostics.Debugger.NotifyOfCrossThreadDependency may be of use?

### Environment details
linq2db version: 2.9.5
Database Server:  Any
Database Provider: Any
Operating system: Windows 10
.NET Framework: 4.7.1

sdanyliv:
Which VS 2019 version do you use?

MaceWindu:
Yep, for @sdanyliv it fails in same place with same error as in referenced exception. For me it works fine, so basically there is something that changes debugger behavior, but we cannot identify what.

So we will change code, that triggers exception as workaround.

MaceWindu:
Closed by #2066

RoyChase:
Thanks for your, as usual, brilliant response.

Issue #1795, opened 2019-07-05
SqlText threw an exception of type 'System.TypeInitializationException'


BLIZZARD123:
Hello 

I'm using this simple statement to select some records from my db.
```c#
public static List<Rate> Get(TradingDB db, string id, ENUM_TIMEFRAME timeframe, DateTime fromDate)
        {
            try
            {
               var query = from row in db.Rates
                           where row.MarketId == id && row.Timeframe == timeframe.ToString() && row.TimeframeDate >= fromDate
                           orderby row.TimeframeDate descending
                           select row;

                return query.ToList();
            }
            catch (Exception e)
            {
                Logger.Print(ENUM_LOGGER.LOGGER_SQL, Helper.GetMethod(MethodBase.GetCurrentMethod()), e.Message);
                return new List<Rate>();
            }
        }
```

This statement have been working for month, but since some weeks?? it's not working anylonger.

IMO there is something wrong with the datetime 

```
Exception message:  '((LinqToDB.Linq.ExpressionQuery<DataModels.Rate>)query).SqlText' threw an exception of type 'System.TypeInitializationException' | string {System.TypeInitializationException}


Stack trace:    bei LinqToDB.Linq.Query`1.GetQuery(IDataContext dataContext, Expression& expr) in C:\projects\linq2db\Source\LinqToDB\Linq\Query.cs:Zeile 219.
   bei LinqToDB.Linq.ExpressionQuery`1.GetQuery(Expression& expression, Boolean cache) in C:\projects\linq2db\Source\LinqToDB\Linq\ExpressionQuery.cs:Zeile 84.
   bei LinqToDB.Linq.ExpressionQuery`1.get_SqlText() in C:\projects\linq2db\Source\LinqToDB\Linq\ExpressionQuery.cs:Zeile 53.
```

### Steps to reproduce

```c#
[Table(Schema="dbo", Name="rates")]
	public partial class Rate
	{
		[Column("market_id"),      PrimaryKey(1), NotNull] public string   MarketId      { get; set; } // nvarchar(80)
		[Column("timeframe"),      PrimaryKey(2), NotNull] public string   Timeframe     { get; set; } // nvarchar(13)
		[Column("timeframe_date"), PrimaryKey(3), NotNull] public DateTime TimeframeDate { get; set; } // smalldatetime
		[Column("open_time"),                     NotNull] public DateTime OpenTime      { get; set; } // smalldatetime
[...]

PR #2066, opened 2020-01-29
Fix debugger failures on some configurations

MaceWindu:
Removed uses of `ConcurrentBag<T>`, which (due to use of `ThreadLocal<T>`) doesn't work well with some debugger configurations (see #2064, #1795)

This alone will fix debugger issue, but I plan to see if we can improve related caching functionality.


MaceWindu:
Actually don't see what we can improve in caching right now - current code doesn't look like it needs urgent changes

PR #2087, opened 2020-02-10
support cross/outer apply and skip/take in oracle V12c

jogibear9988:


Source/LinqToDB/DataProvider/Oracle/OracleDataProvider.cs
```diff
@@ -14,27 +14,40 @@ namespace LinqToDB.DataProvider.Oracle
...
+			Version = version;
+
 			//SqlProviderFlags.IsCountSubQuerySupported        = false;
 			SqlProviderFlags.IsIdentityParameterRequired       = true;
 			SqlProviderFlags.IsCommonTableExpressionsSupported = true;
 			SqlProviderFlags.IsSubQueryOrderBySupported        = true;
 			SqlProviderFlags.IsDistinctOrderBySupported        = false;
 			SqlProviderFlags.IsUpdateFromSupported             = false;
 
+			if (version != OracleVersion.v11)
```
sdanyliv: Replace with
```cs
version >= OracleVersion.v12
```

Source/LinqToDB/DataProvider/Oracle/OracleVersion.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace LinqToDB.DataProvider.Oracle
+{
+	public enum OracleVersion
+	{
+		v11,
+		v12,
+		latest
```
sdanyliv: replace with `Latest`

sdanyliv:
Still needed Version Detector, as it is done for PostgresSQL and SQL Server
```sql
select VERSION from PRODUCT_COMPONENT_VERSION
where PRODUCT like 'PL/SQL%';
```

jogibear9988:
@MaceWindu 


> Still needed Version Detector, as it is done for PostgresSQL and SQL Server
> 
> ```sql
> select VERSION from PRODUCT_COMPONENT_VERSION
> where PRODUCT like 'PL/SQL%';
> ```

I've implemented something, but untested at the moment. Need to look when I find time to test!

Source/LinqToDB/DataProvider/Oracle/OracleDataProvider.cs
```diff
@@ -14,27 +14,40 @@ namespace LinqToDB.DataProvider.Oracle
...
 			SqlProviderFlags.MaxInListValuesCount              = 1000;
 
 			SetCharField            ("Char",  (r, i) => r.GetString(i).TrimEnd(' '));
 			SetCharField            ("NChar", (r, i) => r.GetString(i).TrimEnd(' '));
 			SetCharFieldToType<char>("Char",  (r, i) => DataTools.GetChar(r, i));
 			SetCharFieldToType<char>("NChar", (r, i) => DataTools.GetChar(r, i));
 
-			_sqlOptimizer = new OracleSqlOptimizer(SqlProviderFlags);
+			if (version == OracleVersion.v11)
+				_sqlOptimizer = new OracleSqlOptimizer(SqlProviderFlags);
```
MaceWindu: We should use 11 in provider classes for v11

Source/LinqToDB/DataProvider/Oracle/OracleVersion.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace LinqToDB.DataProvider.Oracle
+{
+	public enum OracleVersion
+	{
+		v11 = 11,
+		v12 = 12,
+		v18 = 18,
```
MaceWindu: We shouldn't add versions, we don't use

Source/LinqToDB/DataProvider/Oracle/OracleDataProvider.cs
jogibear9988: Don't know what you mean here...

MaceWindu: OracleSqlOptimizer -> Oracle11SqlOptimizer

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlBuilder.cs
```diff
@@ -97,6 +97,12 @@ protected override void BuildPredicate(ISqlPredicate predicate)
 			base.BuildPredicate(predicate);
 		}
 
+		protected bool BasiSqlBuilderBuildWhere(SelectQuery selectQuery)
```
sdanyliv: Why it is needed?

jogibear9988: So I can call the Base Function from V12 provider. So if smth is added to base V12 also has included this.

jogibear9988:
@MaceWindu @sdanyliv 
I think I'm finished with this feature.
should I rebase the commits into one? 
And is something left I should do?

sdanyliv:
@jogibear9988, i see regression at least in one test
TruncateIdentityNoResetTest

jogibear9988:
I don't understand what I should have changed so this test will fail. Cause the Code for oracle 11 should be the same.
And the test works on MacOs with V11... strange..

MaceWindu:
TruncateIdentityNoResetTest IIRC is unstable test, which fails from time to time waiting somebody to lok at it :)

jogibear9988:
maybe you can review my code. for me there is no change wich can cause it to fail.

sdanyliv:
Doing that right now. Failing test is really strange and contain long standing bug.

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlBuilder.cs
sdanyliv: Remove that it do nothing.

jogibear9988: I only put it there to use it in v12 provider to access the method from the base class. But I will remove it.

PR #2090, opened 2020-02-12
Added some overloads accepting sql parameters as object

### LINKED ISSUE #2089: Add extention method for queryproc to simplify parameter specification
Somebody already guessed this in some issue's comments, but it's still not implemented..

I want to query stored proc's using anonymous type as a parameter storage:

Db.QueryProc<MyType>("myStoredProc", new {TypeID = 55, CloseDate = DateTime.Now.Date});

It's a pain working with dataparameter[]

feeleen:
Fixes #2089

### DIFF
--- a/Source/LinqToDB/Data/DataConnectionExtensions.cs
+++ b/Source/LinqToDB/Data/DataConnectionExtensions.cs
@@ -106,6 +106,29 @@ namespace LinqToDB.Data
 		}
 
 		/// <summary>
+		/// Executes command using <see cref="CommandType.StoredProcedure"/> command type and returns results as collection of values, mapped using provided mapping function.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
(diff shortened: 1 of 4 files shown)


sdanyliv:
Thanks for that!
Even such simple tasks help us to improve usability. We appreciate any help, there are a lot of urgent tasks to do and we can not switch from process.

sdanyliv:
Please try that overloads do not fall in collision with other methods, that's all.

feeleen:
@sdanyliv do you want me to add some additional tests to the Test project for this? or just check than calling this overrides will compile? I added some calls in a test project like

`var set32 = db.QueryProc<VariableResult>("[VariableResults]",
					new { ReturnFullRow = 1 }).First();`

it does compile

sdanyliv:
I would prefer to make test call using each overload.
We need that for regression detection. It really helps when we refactor legacy parts.
There are already QueryProc calls in our tests, we just need to duplicate them using this new overload.

Thanks!

feeleen:
fixed my copy/paste typo in a test project.. )

PR #2096, opened 2020-02-19
DataContextTransaction: pass cancellation tokens to BeginTransactionAsync

evilbloodydemon:
and fixed typos


MaceWindu:
Thanks! We already addressed token issue in v3, but looks like didn't copied fix to v2

PR #2109, opened 2020-02-29
Restore enums in expressions

### LINKED ISSUE #2041: MapValue and greater/smaller comparison
Hi,

I discovered an issue with enum mappings. Consider the following unit test:
```cs
[Test]
public void MappedEnumGreaterAndLessComparison([IncludeDataSources(ProviderName.Access)] string context)
{
	{
		var result = db.GetTable<Person>()
			.Where(_ => _.Gender > Gender.Female)
			.Where(_ => _.Gender <= Gender.Male)
			.ToList();

		Assert.IsTrue(result.All(_ => _.Gender == Gender.Male));
	}
}
```
I would expect the linq2db to produce the following query (similar to the != operator):
```sql
SELECT
	[_].[FirstName],
	[_].[PersonID],
	[_].[LastName],
	[_].[MiddleName],
	[_].[Gender]
FROM
	[Person] [_]
WHERE
	[_].[Gender] > 'F' AND [_].[Gender] <= 'M'
```
But it produces the following query, which of course breaks down with exception 'Data type mismatch in criteria expression.' (Access provider):
```sql
SELECT
	[_].[FirstName],
	[_].[PersonID],
	[_].[LastName],
	[_].[MiddleName],
	[_].[Gender]
FROM
	[Person] [_]
WHERE
	[_].[Gender] > 1 AND [_].[Gender] <= 0
```
Would it be possible to improve linq2db, so that the mapping is applied also for other comparison types (<, >, <=, >=)? Thank you in advance for any kind of help/feedback and best regards,

Boris

MaceWindu:
Fix #2041

PS:
Perfect example of premature optimization:
- first your optimize expression in wrong place instead of doing it in expression compiler
- then you still need same optimization in expression compiler, because of manually created expressions

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -1708,8 +1708,8 @@ namespace LinqToDB.Linq.Builder
 
 			switch (nodeType)
 			{
-				case ExpressionType.Equal:
-				case ExpressionType.NotEqual:
+				case ExpressionType.Equal    :
+				case ExpressionType.NotEqual :
 
 					var p = ConvertObjectComparison(nodeType, context, left, context, right);
 					if (p != null)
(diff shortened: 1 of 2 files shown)


MaceWindu:
Merged from #2110

PR #2112, opened 2020-02-29
SQLite DateTime roundtrip improvements

### LINKED ISSUE #2107: BulkInsert doesn't properly set UTC in SqlLite
Using a DateTime object, doing a single insert properly sets the date to: 2020-02-21 16:00:00Z
However, when doing a BulkInsert, the date is stored as 2020-02-21 16:00:00  with no Z and no way to tell that it's Z time.

I modified \DataProvider\SQLite\SQLiteMappingSchema.cs  to the below:
__________________________________________________________________________________________
static void ConvertDateTimeToSql(StringBuilder stringBuilder, DateTime value)
{
	string Z = (value.Kind == DateTimeKind.Utc) ? "Z" : "";
	if (value.Millisecond == 0)
	{
		var format = value.Hour == 0 && value.Minute == 0 && value.Second == 0 ?
			"'{0:yyyy-MM-dd}'" :
			"'{0:yyyy-MM-dd HH:mm:ss" + Z + "}'";
		stringBuilder.AppendFormat(format, value);
	}
	else
	{
		stringBuilder
			.Append(string.Format("'{0:yyyy-MM-dd HH:mm:ss.fff" + Z + "}", value).TrimEnd('0'))  
			.Append('\'');
	}
}
__________________________________________________________________________________________

### Environment details
linq2db version: 2.9.7
Database Provider: SqlLite
Operating system: Win10
.NET Framework: 4.7

MaceWindu:
Fix #2107 

Added tests which show that:
- both providers return Kind=Unspecified
- system.data.sqlite cannot handle some testcases (microsoft provider works fine)

Probably we should do following:
- generate proper ISO8601 literal taking into account Kind
- fix system.data.sqlite issues if possible


MaceWindu:
I'm postponing work on this till v3, as I observe a lot of inconsistencies in both providers behavior:
- S.D.S doesn't support unittime/julian dates (probably could be hacked-around with custom reader+writer(!))
- S.D.M works as expected with Z-suffixed literals for any column type but ignores Kind=UTC if parameters used (doesn't make any sense to me)

Need to retest it in v3 with latest versions of both providers and decide what to do with it

to11mtm:
Just wanted to add some thoughts...

As you've mentioned the inconsistencies in providers helps nothing here, nor does the differences in types returned.

System.Data.SQLite allows you to bind a .NET function as a UDF. Unfortunately Microsoft.Data.SQLite does not let you do so unless you are on 2.2.0 or newer (which is only .NetStandard 2.0). I mention this because I did have the idea of using a UDF to 'normalize' the dates in queries. Semi-psuedocode example below:

```
[SQLiteFunction(Name = "EPOCHDATE", Arguments = 1, FuncType = FunctionType.Scalar)]
class EpochDate: SQLiteFunction
{
   public override object Invoke(object[] args)
   {
      var str = Convert.ToString(args[0]);
      if (int.TryParse(str, out int epochTime))
      {
           return epochTime;
      }
      else if (str.Contains("z"))
      {
           //Convert to epoch Time
      }
   }
} 

//on connection:
connection.RegisterFunction(typeof(EpochDate).GetCustomAttribute<SQLiteFunction>(), new EpochDate());
```

PR #2128, opened 2020-03-12
DbDataType refactoring and more NRT annotations

MaceWindu:
- migrate SqlParameter, SqlField, SqlDataType classes to use DbDataType
- extend DbDataType with precision and scale attributes
- add more NRT annotations around affected code
- fixed issues, discovered during code annotation with NRT

~still need to finish regression testing and maybe trim unused constructors from refactored AST types~ done


MaceWindu:
image

Source/LinqToDB/ServiceModel/LinqServiceSerializer.cs
```diff
@@ -597,9 +632,8 @@ void Visit(IQueryElement e)
 						{
 							var fld = (SqlField)e;
 
-							// All could have type for scalar subqueries
```
MaceWindu: This is right now broken, need to look into it (* scalar query over linq)

PR #2129, opened 2020-03-12
Fix for #2118. Added Take/Skip parametrization.

### LINKED ISSUE #2118: Pagination parametrization
Hi,

As I can see,  IQuerable<T>.Skip and IQuerable<T>.Take are not parameterized.
E.g. query on database side is 
        ...
        OFFSET 60 ROWS FETCH NEXT 60 ROWS ONLY 

and should be 
       ...
       OFFSET @__p_1 ROWS FETCH NEXT @__p_2 ROWS ONLY

This is the case in EF Core.
For large queries where only difference is in pagination, there is noticeable delay because Sql Server can't reuse existing execution plan. 

### Environment details
linq2db version:  2.9.6
Database Server: Sql Server 2016
Database Provider: Microsoft
Operating system:  Windows 10
.NET Framework: .Net Core 3.1

sdanyliv:
Fixes #2118

Tasks to do:

- [x] Check caching performance - it should work anyway, before fix, Take/Skip queries almost not cached.
- [x] Cleanup Playground project


igor-tkachev:
I think Take/Skip parametrization was implemented 10 years ago, wasn't it?

sdanyliv:
Nope. Even now we have to improve Equals for Expressions to compare Take/Skip methods specially.

PR #2178, opened 2020-04-12
Fix another case of compiler being 'helpful'

### LINKED ISSUE #2166: Unnecessary conversion of numeric values prevents index usage
Query:

```C#
var list = await Db.AddressInstances
	.With("ReadPast, Index(IX_AddressInstance_Id_NotParsed)")
	.Where(x => x.RawDataStatus == (byte)RawDataStatus.NotParsed)
	.Take(100)
	.Select(x => new
	{
		Id = x.Id,
		RawData = x.RawData,
		RawDataHash = x.RawDataHash,
		RawDataEventId = x.RawDataEventId,
		RawDataEventTimeUtc = x.RawDataEventTimeUtc,
	})
	.ToListAsync(cancellationToken);
```

Generated SQL:

```SQL
SELECT TOP (100)
	[x].[Id],
	[x].[RawData],
	[x].[RawData_Hash],
	[x].[RawData_EventId],
	[x].[RawData_EventTimeUtc]
FROM
	[parsed].[AddressInstance] [x] WITH (ReadPast, Index(IX_AddressInstance_Id_NotParsed))
WHERE
	Convert(Int, [x].[RawData_Status]) = 0
```

`Convert(Int, [x].[RawData_Status])` is generated even though `RawDataStatus.NotParsed` is explicitly cast to `byte`.

The code above results in an SqlException:

```
System.Data.SqlClient.SqlException: 'Query processor could not produce a query plan because of the hints defined in this query. Resubmit the query without specifying any hints and without using SET FORCEPLAN.'
```

This means that the query will not use the index even if there is no table hint.

If `Convert(Int, [x].[RawData_Status]) = 0` is replaced with `[x].[RawData_Status] = 0` the query works as expected.

### Environment details
linq2db version: 2.9.7
Database Server: SQL Server 2019
Database Provider: SqlServer2017
Operating system: Windows Server 2019
.NET Framework: .NET Standard 2.1

MaceWindu:
Fix #2166


MaceWindu:
Actually, I want to add a bit more tests

MaceWindu:
Actually I have richer imagination than roslyn developers, new tests didn't show any issues

sdanyliv:
Better to move such tests to standard linq tests (better filename and class names). It is much complex than original issue.

PR #2180, opened 2020-04-13
Release 2.9.8

MaceWindu:
- run tests
    - [x] netfx
    - [x] netcore1
    - [x] netcore2

We plan this release as last one in v2 branch with 3.0.0 as next non-preview release.
After this release we will replace `master` branch content with v3 code and archive v2 to separate branch.

@linq2db/testers we plan to release on thursday (16/04)

[release notes](https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#release-298)

### DIFF
--- a/Build/linq2db.Default.props
+++ b/Build/linq2db.Default.props
@@ -1,6 +1,6 @@
 ﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<PropertyGroup>
-		<Version>2.9.7.0</Version>
+		<Version>2.9.8.0</Version>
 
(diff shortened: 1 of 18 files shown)


detoxhby:
Ran test on .NET Framework 4.8 and .NET Core 3.1, seems fine ✅ 

**Note**: we've have found a premature optimization problem where a variable based equality check (`entity.Col == someVarFromOutside`) combined with multiple boolean entity column checks gets casted to constant expression and thus throws exception. The solution was to nest the equality check part into a base query. We don't have the time nor the human resource to give enough data to reproduce this now (the data model is huge and the query is complex); will try to report it later and the fix would be included in 3.0-rc.

PR #1814, opened 2019-07-17
[WIP] Asp.Net integration

sdanyliv:
Added new package for native support linq2db in Asp.Net projects.

### Taks to do
- [ ] Make options immutable
- [ ] Caching provider information
- [ ] Tests

PR #2182, opened 2020-04-16
Asp.Net integration

### LINKED ISSUE #1514: Is there easy way to set ConnectionString and ProviderName by json file in Asp.Net Core 2.2?

MaceWindu:
Resubmit #1814 as github decided to close it when target branch were removed...

Added new package for native support linq2db in Asp.Net projects.

### Taks to do
- [x] Make options immutable
- [ ] Caching provider information
- [ ] Tests

Fix #1514


hahn-kev:
Since I introduced the options builder the options class is now immutable. 
I'm not sure I understand the Caching provider information task, is this something we want as a configuration option? or something else?

sdanyliv:
Sorry I have no time to look at this deeper. But I really want to reuse options in linq2db project itself without any additional dependencies.

This PR should stay as an Asp.NET Core integration, but we also need options mechanism for linq2db.
Do you have a time to help us with this?

I see this something like in EF Core, but without dependency on Microsoft libraries.
```
optionsBuilder
   .UseSqlServer("connection string", builder => 
      {
         builder.ServerVersion = 2017,
         builder.MappingSechema = ms,
        ...
      }
);

optionsBuilder
   .UsePostreSql("connection string", builder => 
      {
         builder.ServerVersion = v12,
         builder.MappingSechema = ms,
         builder.UseNodaTime() // it is a sample how we can extend current provider by services
         builder.UseGeospatial()
      }
);
```

We are trying to avoid global configuration settings and make linq2db more configurable.
You can see this mess in https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Common/Configuration.cs

hahn-kev:
Ahh ok, yeah that makes sense. I'm not sure I have time for that now however, but I would like to help get the asp.net integration finished. 

On that subject I have merged master into this branch now that it includes all the v3 features. I had done that previously by rebasing this branch onto the v3 branch but I just squashed and cherry picked that onto this branch and then merged master into it. 

I think there's no reason we can't use all the options and option builders without asp.net. Really the only thing connecting them is the code to allow us to configure the IoC container. Everything else related to the configuration is seperate. So I think we should pretty easily be able to add more configuration options using the builder without taking more dependencies on asp.net.

sdanyliv:
Ok, let's polish current PR without additional stuff.
Could you add logging to options? It should create DataConnection with assigned logging events and redirect them to logger (LoggerFactory?)

hahn-kev:
Alright, I've added some initial work for logging using the existing tracing code. It's not ideal and I've added some todo comments where we should remove some of the getters to prevent using them, I'm not really sure of a good solution for that at the moment.

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -36,16 +36,15 @@ public partial class DataConnection : ICloneable
 		/// <summary>
 		/// Creates database connection object that uses default connection configuration from <see cref="DefaultConfiguration"/> property.
 		/// </summary>
-		public DataConnection() : this((string?)null)
+		public DataConnection() : this(new LinqToDbConnectionOptionsBuilder())
```
sdanyliv, line 39: Maybe I'm missing something, but Builder is builder for options. Options should be separated class.

sdanyliv:
For AspNet integration we have to use not just TraceSwitch but LoggerFactory
builder.UseLoggerFactory
Like this https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontextoptionsbuilder.useloggerfactory?view=efcore-3.1

People become used to DI.

Source/LinqToDB/Data/DataConnection.cs
hahn-kev: Yes it is, the constructor that's being called here is on [line 220](https://github.com/linq2db/linq2db/pull/2182/files#diff-72b988f98cfdeadd8d3e668bbfe51a51R220-R223) which is a private constructor that just calls `builder.Build()` and passes the result to the constructor that takes the actual options class. We could add `Build()` at the end of this call and remove the private constructor, but then we'd have to do that on every other public constructor like this, so this seemed simpler.

hahn-kev:
Alright, so what are you thinking? like an adapter that uses the factory to make a logger which makes a delegate for the WriteTrace function in the data connection. Similar to this
```C#
var logger = loggerFactory.CreateLogger<DataConnection>();
connection.WriteTraceLine = (message, displayName, level) => logger.Log(level, message);
```
it would have to be set via the options builder but you get the point, and I'd have to convert the level into the correct enum.

sdanyliv:
I have implemented similar function in EF Core integration.
https://github.com/linq2db/linq2db.EntityFrameworkCore/blob/master/Source/LinqToDB.EntityFrameworkCore/LinqToDBForEFToolsImplDefault.cs#L1233

Also you can find how I have integrated linq2db in such case.

hahn-kev:
ok, but in order to do something like that then linq2db would have to depend on the Logger abstractions library. I think we should rework how logging works to depend on a standard interface, but I think that's a bit out of scope for this PR.

sdanyliv:
Currently we have to avoid additional dependencies. Let’s do that only for AspNet integration.

hahn-kev:
Alright, I've added a method for using logger factory, and also a method for using the default logging that's in the IoC container.

They're used like this
```C#
builder.UseLoggerFactory(factory);
//or
builder.UseDefaultLogging(serviceProvider)
```

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -273,21 +212,145 @@ public DataConnection(string? configurationString)
...
-				? asyncDbConection
-				: AsyncFactory.Create(transaction.Connection);
-			TransactionAsync   = AsyncFactory.Create(transaction);
-			_closeTransaction  = false;
-			_closeConnection   = false;
-			_disposeConnection = false;
+			IDbConnection?  localConnection  = null;
+			IDbTransaction? localTransaction = null;
+			
+			switch (options.SetupType)
```
sdanyliv, line 243: Do we really need this switch?

hahn-kev: I think so, if we want the DataConnection to work the same as it did before the options class was introduced. There's basically a type for each of the different types of constructors there used to be, then the code for each was moved into the switch.

We might be able to determine the same thing by checking parameters and falling back if something isn't provided. But then handling errors would get more complicated. For example: Connection string type requires a data provider and throws an error if there isn't one, however with a Configuration string it's ok if there's not a DataProvider. We might not want to allow a null DataProvder, but when I made these changes I tried to make sure to maintain backwards compatibility, if you're fine with changing that then I'd be happy to make changes, but I'm not sure what the requirements are for the data connection, so I just tried to make sure the code all worked the same as it did before.

NuGet/linq2db.AspNet.nuspec
```diff
@@ -0,0 +1,41 @@
...
+			</group>
+			<group targetFramework=".NETStandard2.0">
+				<dependency id="linq2db" version="2.6.4"/>
+			</group>
+			<group targetFramework=".NETCoreApp1.0">
+				<dependency id="linq2db" version="2.6.4"/>
+			</group>
+			<group targetFramework=".NETCoreApp2.0">
+				<dependency id="linq2db" version="2.6.4"/>
+			</group>
```
sdanyliv, line 29: Update dependencies.

Source/LinqToDB/Common/Logging/LoggingExtensions.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Diagnostics;
+using LinqToDB.Data;
```
sdanyliv, line 3: Move linq2db usings in namespace part.

Source/LinqToDB/Common/Logging/LoggingExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+			string message,
+			string displayName,
+			TraceLevel level)
+		{
+			switch (context)
+			{
+				case DataConnection connection:
+					connection.WriteTraceLineConnection(message, displayName, level);
+					break;
+					default:
```
sdanyliv, line 19: Something wrong with idents.

Source/LinqToDB/Configuration/LinqToDbConnectionOptions.cs
```diff
@@ -0,0 +1,84 @@
+using System;
+using System.Data;
+using System.Diagnostics;
+using LinqToDB.Data;
+using LinqToDB.DataProvider;
+using LinqToDB.Mapping;
+
```
sdanyliv, line 7: Move linq2db usings in namespace part.

Source/LinqToDB/Configuration/LinqToDbConnectionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+					DbTransaction = builder.DbTransaction;
+					break;
+			}
+
+			MappingSchema = builder.MappingSchema;
+			DataProvider = builder.DataProvider;
+			OnTrace = builder.OnTrace;
+			TraceLevel = builder.TraceLevel;
+			WriteTrace = builder.WriteTrace;
+		}
```
sdanyliv, line 54: Correct formatting.

Source/LinqToDB/Configuration/LinqToDbConnectionOptionsBuilder.cs
```diff
@@ -0,0 +1,213 @@
+using System;
+using System.Data;
+using System.Diagnostics;
+using LinqToDB.Data;
+using LinqToDB.DataProvider;
+using LinqToDB.DataProvider.SqlServer;
+using LinqToDB.Mapping;
+
```
sdanyliv, line 8: Move linq2db usings in namespace part.

hahn-kev:
I've fixed the feedback you gave. There's not many unit tests, though any tests using the public constructors are using the new connection options. I can write up some tests that will also serve as an example for how to use the new asp.net package.

sdanyliv:
I like that. @MaceWindu should review NuGet package correctness.
@David-Mawer, could you review that also, in future you can change your templates to use this new library.

MaceWindu:
@hahn-kev,
That's really cool. We definitely should continue to move in this direction with other settings in next releases.

Could you:
- add xml-docs for types in linq2db.aspnet project
- create documentation article in docs repository and reference it from readme/release notes

@linq2db
Also, are we fine with linq2db.AspNet name or better use linq2db.AspNetCore?

My changes:
- fixed new nuget publishing and build targets
- bumped dependencies. They are backward-compatible by API surface, so it should be ok to enforce higher versions. But we can go with lowest if needed
- fixed nullability

hahn-kev:
Alright, I've added and updated some xml docs. 

I also realized that there's already some tests that serve an example of how to use the Asp.Net package, they can be found [here](https://github.com/linq2db/linq2db/pull/2182/files#diff-6384b373416617b485a47144696e504eR296-R322).

 I'd be happy to write some docs though I'm not really sure where to start. Is there a PR that did a similar change I could look at? I'm thinking an article should probably be added under [Getting Started](https://linq2db.github.io/articles/get-started/) with some examples of how to configure Asp.Net.

@MaceWindu thanks for also cleaning up some style issues I left behind.

As for the project name, doesn't matter to me, though at the moment I think it's usable in .Net Framework and Core, so calling it Asp.Net Core might dissuade any Framework people, not a big deal either way.

MaceWindu:
@hahn-kev , for article example check this PR https://github.com/linq2db/docs/pull/23, last 3 files:
- associations.md 
- toc.yml 

Basically, you write article as markup file and add it to corresponding TOC.
I would recommend to use VS Code for it as it has good markup preview support. We already have workspace file for VSCode in docs repository `docs.code-workspace`.

hahn-kev:
Thanks for the tips, I've pushed my PR and a simple review of how to use the ASP.NET Integration for the release notes.

PR #2186, opened 2020-04-19
Feature/partial update

### LINKED ISSUE #2185: Partial Update v2
related to #1044 

Maybe linq2db "Update" method could have additional parameter with list of Properties (non-pk) to update?
If I use my own change tracking code to monitor properties that changed - then I want to pass these props to Update method. It's unnessesary to update every property in this case

feeleen:
Partial Update. Fix #2185


sdanyliv:
@feeleen,  I think it should be new overload method, to do not break existing users code.

Source/LinqToDB/Linq/QueryRunner.Update.cs
```diff
@@ -88,20 +92,20 @@ public static class Update<T>
 				var type = GetType<T>(obj!, dataContext);
 				var entityDescriptor = dataContext.MappingSchema.GetEntityDescriptor(type);
 				var ei               = Configuration.Linq.DisableQueryCache || entityDescriptor.SkipModificationFlags.HasFlag(SkipModification.Update)
```
sdanyliv, line 94: If columnFilter is present, cache should not be used.

Tests/Linq/Linq/ComplexTests2.cs
```diff
@@ -364,6 +364,44 @@ public void TestUpdateWithTypeAndBaseType([DataSources] string context)
 			}
 		}
 
+
+		[Test]
+		public void TestUpdateWithColumnFilter([DataSources] string context)
```
sdanyliv, line 369: Better to do not create test in this file. There are UpdateTests for that.

Issue #1740, opened 2019-05-29
Hana SQL: ExecuteProcAsync issue


MichielDG:
I've using linq2db for SapHana and want to use the method ExecuteProcAsync from the DataConnectionExtensions class. But when I execute the flowing code 
```c#
await dataConnection.ExecuteProcAsync("TriAPI_AddPlanningElement",
                new DataParameter("@parentKey", parameters.ParentKey, DataType.VarChar),
                new DataParameter("@description", parameters.Description, DataType.VarChar),
                new DataParameter("@calendarKey", parameters.CalendarKey, DataType.VarChar),
                new DataParameter("@scheduleStart", parameters.ScheduleStart, DataType.DateTime),
                new DataParameter("@scheduleFinish", parameters.ScheduleFinish, DataType.DateTime),
                new DataParameter("@returnStatus", returnStatus, DataType.Int32) { Direction = ParameterDirection.InputOutput },
                new DataParameter("@returnKey", planningKey, DataType.VarChar) { Direction = ParameterDirection.InputOutput, Size = 20 });
```

I get an exception like this:
```
Sap.Data.Hana.HanaException (0x80004005): invalid name of function or procedure: TRIAPI_ADDPLANNINGELEMENT: line 1 col 6 (at pos 5)
   at Sap.Data.Hana.HanaCommand._ExecuteNonQuery()
   at Sap.Data.Hana.HanaCommand.ExecuteNonQuery()
   at System.Data.Common.DbCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LinqToDB.Data.DataConnection.<ExecuteNonQueryAsync>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LinqToDB.Data.CommandInfo.<ExecuteAsync>d__35.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #2188, opened 2020-04-21
ACCESS ODBC provider

### LINKED ISSUE #333: Access DB ODBC support?
My web host requires an ODBC connection for MS Access databases.  How do I create a connection string for an ODBC DSN?

MaceWindu:
- adds ODCB Access provider support. Fix #333
- "unbans" 1899/12/30 date for Access provider
- fix types for Access OleDb provider schema
- fixed select procedures result schema load for OleDb

Notes and limitations of provider:
- `DateTime` type stripped of milliseconds on both read and write  (ODBC provider limitation)
- `Int64` parameters not supported (worked for OleDb). Should be provider limitation, but I didn't tested with v15 database, which supports `long long` type
- `ExtendedAnsiSQL=1` connection string parameter is highly recommended
- [SCHEMA] schema information missing some procedures. Known cases: parameter-less insert procedures. See [#35444](https://github.com/dotnet/runtime/issues/35444)
- [SCHEMA] primary/foreign keys information not returned, as it is not possible to distinguish them from indexes with ODBC schema API. See [#35442](https://github.com/dotnet/runtime/issues/35442)
- [SCHEMA] OleDb provider reports `BINARY` as `VARBINARY` (OLEDB provider bug)
- [SCHEMA] OleDb provider reports some not-null INT columns as of COUNTER type  (OLEDB provider bug)
- [SCHEMA] ODBC provider incorrectly reports some not-nullable columns as nullable (ODBC provider bug)
- T4 test template generates incorrect procedure calls, but we don't plan to ship T4 template for ODBC for now. Should be fixed later when we implement more provider-aware procedure API (see #1740)


MaceWindu:
Another confirmation that ADO.NET GetSchema API should be avoided when possible. Both OLE DB and ODBC providers return incorrect data, and both in different places...

MaceWindu:
This is ready. I still need will try to fix schema for SELECT procedures, as I discovered we don't generate result tables for them in T4 templates

PR #2203, opened 2020-04-29
Feature/partial update

MaceWindu:
Resubmit #2186


feeleen:
just noticed that something went wrong.. 
how it could affect bulkcopy tests?

MaceWindu:
Ignore it, that oracle test is unstable in azure for yet unknown reason. and sqlserver failures on macos also unrelated - something broken on macos recently

PR #2215, opened 2020-05-10
Implement provider-specific bulk copy for MySqlConnector

### LINKED ISSUE #2113: Add support for provider-specific BulkCopy for MySqlConnector 0.62.0+
MySqlConnector added bulk copy api recently, should by quick task to support it if we will not hit any issues
https://mysqlconnector.net/overview/version-history/

MaceWindu:
Fix #2113

Limitations:
 - provider-specific bulk copy available only for MySqlConnector 0.67.0+ (previous versions have different API or no API)
- tables with YEAR columns not supported (provider limitation)

Also fix database object quotation logic to properly quote identifiers with ` and dots. Note that this removes following escaping quirks we had:
- if identifier staring from `, it was considered escaped and no escaping applied
- if identifier (for table, schema or database name) contained dots, it was split into components and escaped separately


MaceWindu:
@sdanyliv  review

MaceWindu:
feature postponed till next release, as we found issues with escaping of database objects in bulk copy, which requires research.

MaceWindu:
- bumped min required version to 0.67.0 for bulk copy to work
- fixed mysql identifiers escaping logic (see first message for details)

MaceWindu:
if linux/osx tests will not fail again - this is ready

PR #2223, opened 2020-05-19
FromSql parameter accessors support

MaceWindu:
Still need to fix FTS tests asserts, as now we generate parameters there


MaceWindu:
ne prigodillos

PR #2248, opened 2020-06-01
Added old style LoadWithAsTable API for simplifying backward compatibility.

sdanyliv:
@detoxhby, please review,

### DIFF
--- a/Source/LinqToDB/LinqExtensions.LoadWith.cs
+++ b/Source/LinqToDB/LinqExtensions.LoadWith.cs
@@ -57,16 +57,6 @@ namespace LinqToDB
 			return table;
 		}
 
-		/// <summary>
-		/// Provides support for queryable LoadWith/ThenLoad chaining operators.
-		/// </summary>
-		/// <typeparam name="TEntity">The entity type.</typeparam>
-		/// <typeparam name="TProperty">The property type.</typeparam>
-		// ReSharper disable once UnusedTypeParameter
-		public interface ILoadWithQueryable<out TEntity, out TProperty> : IQueryable<TEntity>, IAsyncEnumerable<TEntity>
-		{
-		}
-
 		class LoadWithQueryable<TEntity, TProperty> : ILoadWithQueryable<TEntity, TProperty>
 		{
 			private readonly IQueryable<TEntity> _query;
--- a/Source/LinqToDB/LinqExtensions.LoadWith.cs
+++ b/Source/LinqToDB/LinqExtensions.LoadWith.cs
@@ -33,7 +33,7 @@ namespace LinqToDB
 		/// // loads records from Reference2 collection association for each loaded Reference1 record
 		/// // loads records from Reference3 association for each loaded Reference2 record
 		/// // note that a way you access collection association record (by index, using First() method) doesn't affect
-		/// // query results and allways select all records
+		/// // query results and always select all records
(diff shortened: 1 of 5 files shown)


Source/LinqToDB/LinqExtensions.LoadWith.cs
```diff
@@ -13,6 +13,50 @@ namespace LinqToDB
...
+		/// db.Table1.LoadWithAsTable(r => r.Reference1.Reference2);
+		///
+		/// // loads records from Table1 with References collection association loaded for each Table1 record
+		/// db.Table1.LoadWithAsTable(r => r.References);
+		///
+		/// // loads records from Table1 with Reference1 collection association loaded for each Table1 record
+		/// // loads records from Reference2 collection association for each loaded Reference1 record
+		/// // loads records from Reference3 association for each loaded Reference2 record
+		/// // note that a way you access collection association record (by index, using First() method) doesn't affect
+		/// // query results and allways select all records
```
detoxhby, line 36: typ0

detoxhby:
Another note: shouldn't the `ILoadWithQueryable` interface be directly in `LinqToDB` namespace?

Feels a bit weird using it as a member and having `LinqToDB.LinqExtensions.ILoadWithQueryable` as type (or even without `LinqToDb` root).

sdanyliv:
Good catch, moving.

detoxhby:
Document-side: I think there should be a warning in upgrade notes that the previous behaviour of mutating state for `ITable` is different than the new `LoadWith` as it now returns a new instance.

I know it's non-compliant way of re-using the same instance, but working code that forgot to reassign the result of `LoadWith` call will break unintentionnally after the upgrade.

detoxhby:
Did some testing in older projects and with mimiced LoadWithAsTable works really strange.
(essentially using `LinqToDB.Linq.MethodHelper.GetMethodInfo(LinqToDB.LinqExtensions.LoadWith, table, selector)` to trick CanBuild recognition...)

There are cases where the LoadWith calls gone missing and the relations were empty even if the final Table.Expression contains the calls. It does not happen if small amount (1-2) calls were used.

I will wait for the merged version of this and retest to make sure it's not the result of my helper, but I think there should be some deeper reason why this happens (probably related to the new interface).

sdanyliv:
What does it mean mimiced version? `LoadWithAsTable` code is fully copied from previous version.
But under hood it can be side effects for example the following code 
```cs
table.LoadWith(e => s.SomeProps.First())
```
Will return table record with only one detail record, But in 2.0 version (not checked) it will return all detail records.

Anyway `LoadWith` has three overloads:
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Reflection/Methods.cs#L77-L87

detoxhby:
No, these are simple simple member access (no .First() or nested access).

By "mimicking", I did this to force the query builder do it's job even without `"LoadWithAsTable"` in CanBuild...:
```csharp
public static ITable<T> LoadWithAsTable<T>(
    this ITable<T> table,
    Expression<Func<T, object>> selector)
            where T : class
{
    if (table == null) throw new ArgumentNullException(nameof(table));

    table.Expression = Expression.Call(
        null,
        LinqToDB.Linq.MethodHelper.GetMethodInfo(LinqToDB.LinqExtensions.LoadWith, table, selector),
        new[] { table.Expression, Expression.Quote(selector) });

        return table;
    }
```
```

Issue #1764, opened 2019-06-16
Release 3.0 roadmap


MaceWindu:
Approximate list of planned changes for 3.0 release

Removals:
- [x] netstandard 1.6 / netcoreapp1.0 support #1533
- [x] netcoreapp2.0 support #1917
- [x] npgsql 2 support (#1260, #1942)

Changes/New Features
- [x] netcoreapp2.1 support #1917
- [x] [MERGE] move Merge API to query AST  (#1512)
- [x] [MERGE][RemoteContext] support Merge API in remote context  (#1512)
- [x] [MERGE] support dynamic properties in Merge API  (#1512)
- [x] [RemoteContext] support provider-specific data readers in remote context (#730, #1512 )
- [x] [SQL] OUTPUT clause support (#1703)
- [x] [SQL] linked servers support (#1357)
- [x] [PROVIDERS] Microsoft.Data.SqlClient provider (#1895)
- [x] [PROVIDERS] SAP.HANA Core provider + odbc provider available for netcore #1917
- [x] nullable reference types annotations support (C#8)
- [x] Migrate to Azure Pipelines (#1748)
- [x] [SQL][PostgreSQL] UPDATE FROM improvement (#1278)
- [x] New async API support for providers that already implemented it (npgsql, mysqlconnector) (#1942)
- [x] Eager loading (#1756)

Also check [this page](https://github.com/linq2db/linq2db/wiki/Release-Notes-3.0.0) for list of changes

genusP:
Please add #832 to roadmap

MaceWindu:
It is actually work in progress already #1756. I didn't added it here, because there are chances to release it before 3.0

yuanxiongwei:
When will it be released

sdanyliv:
Approximately in month or two. But you can use it right now https://dev.azure.com/linq2db/linq2db/_packaging?_a=feed&feed=linq2db%40Local

MaceWindu:
We plan to publish previerw1 today with already implemented functionality https://github.com/linq2db/linq2db/pull/1936

detoxhby:
**Note**: for the last few releases, there were no notifications for the `@testers` of linq2db, which makes the whole initiative pointless...

I volunteer to test out the RC version when arrives, to make sure the changes would not cause awkwardly big refactors in production systems after the release of 3.0.

MaceWindu:
Finally we are ready to release preview 2 with a lot of new features tomorrow.
[...]

PR #2257, opened 2020-06-03
Added possibility to Disable Grouping Guard in query.

sdanyliv:
@detoxhby, please review also. 

Original conversation: https://github.com/linq2db/linq2db/issues/1764#issuecomment-638169815

### DIFF
--- a/Source/LinqToDB/Linq/Builder/AsSubQueryBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/AsSubQueryBuilder.cs
@@ -1,10 +1,11 @@
 ﻿using System.Linq;
 using System.Linq.Expressions;
-using LinqToDB.Expressions;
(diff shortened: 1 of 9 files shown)


detoxhby:
Ah, seems like the direct native `IGrouping<TKey, TElement>` overload is missing and cannot be used in LINQ expression:
```csharp
for user in ctx.Users
group user by user.Type into g
select new UserTypeDescriptor()
{
    UserType = g.Key,
    Users = g.DisableGuard().ToList() // g is IGrouping<,> type here here as range variable
}
```

sdanyliv:
I’ll check this variant. But in this case, better to put AsEnumerable(), always.

sdanyliv:
If i can detect this variant, i can make AsEnumerable() call automatically ))
Here i need to think.

PR #2262, opened 2020-06-06
[CTE Troubleshooting] Temporary branch for bug localization.

sdanyliv:
@filippobottega, let's play here with bug localization.

I have added duplication check into low level functions.

Checkout this branch and add `linq2db.csproj` to your solution. Reference your projects to `linq2db.csproj`.
After these steps you can debug linq2db.


MaceWindu:
Close?

filippobottega:
Thank you!

PR #2270, opened 2020-06-14
Update MySqlConnector to 1.0

bgrainger:
MySqlConnector 1.0 changes the primary namespace to `MySqlConnector`: mysql-net/MySqlConnector#824; this updates all relevant code.

This starts the discussion around updating MySqlConnector to 1.0, but doesn't need to be merged until it's out of beta.


MaceWindu:
Thanks!
I think we will drop support for MySqlConnector < 1.0.0 if nobody will request support. We can support older versions, but don't see a reason to do it for now.

bgrainger:
Updated MySqlConnector 1.0 and rebased PR.

PR #2273, opened 2020-06-15
[Feature] Value Converters

sdanyliv:
I think it is another most requested feature.
In mapping builder you can define `prop.HasConversion(..)` functions.

In this PR:
- Fixed found problems with parameters.
- Corrected `FluentMappingBuilder`.
- Made breaking change to `ColumnDescriptor.GetValue` - from now it can not accept `MappingSchema` as parameter.

TODO:
- [x] Pass tests
- [x] Self code review


Source/LinqToDB/Expressions/ConvertFromDataReaderExpression.cs
```diff
@@ -81,50 +105,55 @@ public Expression Reduce(IDataContext dataContext, IDataReader dataReader, Expre
...
 			}
 
-			if (toType.IsEnum)
+			if (converter != null)
+			{
+				// we have to prepare read expression to conversion
+				//
+				var expectedType = converter.FromProviderExpression.Parameters[0].Type;
+				
+				if (converter.HandlesNulls)
```
MaceWindu, line 114: shouldn't it be `if (!converter.HandlesNulls)` ? probably we should add tests for this flag

Source/LinqToDB/Expressions/ConvertFromDataReaderExpression.cs
```diff
@@ -81,50 +105,55 @@ public Expression Reduce(IDataContext dataContext, IDataReader dataReader, Expre
...
-			{
-				var variable = Variable(ex.Type);
-				var assign   = Assign(variable, ex);
-
-				ex = Block(new[] { variable }, new[] { assign, conv.GetBody(variable) });
-			}
-			else
-			{
-				ex = conv.GetBody(ex);
-			}
```
MaceWindu, line 122: just remove? no ApplyLambdaToExpression call?

Source/LinqToDB/Expressions/ConvertFromDataReaderExpression.cs
MaceWindu: actually here we read initial value to pass to converter, so condition is correct

PR #2276, opened 2020-06-16
[Feature] Weak Join

sdanyliv:
Added possibility to define `WeakInnerJoin` and `WeakLeftJoin`.
Example of usage:
```cs
var query =
	from t in table
	from j1 in forJoin.WeakInnerJoin(j1 => t.Id == j1.ParentId)
	from j2 in forJoin.WeakInnerJoin(j2 => t.Id == j1.Id)
	select new
	{
		T = t,
		J1 = j1,
		J2 = j2
	};
```
It adds benefits if you simplify projection. For example if you select from this query field which do not belong to Weak join, it will be removed automatically.
The following query will contain NO JOINS and will run a lot faster.
```cs
query.Select(e => e.t);
```

TODO:
- [x] More tests

### DIFF
--- a/Source/LinqToDB/Linq/Builder/AllJoinsBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/AllJoinsBuilder.cs
@@ -17,7 +17,7 @@ namespace LinqToDB.Linq.Builder
 		{
 			return
 				methodCall.IsQueryable("Join") && methodCall.Arguments.Count == 3
-				|| !rightNullableOnly && methodCall.IsQueryable("InnerJoin", "LeftJoin", "RightJoin", "FullJoin") && methodCall.Arguments.Count == 2
+				|| !rightNullableOnly && methodCall.IsQueryable("InnerJoin", "WeakInnerJoin", "LeftJoin", "WeakLeftJoin", "RightJoin", "FullJoin") && methodCall.Arguments.Count == 2
 				|| rightNullableOnly && methodCall.IsQueryable("RightJoin", "FullJoin") && methodCall.Arguments.Count == 2;
 		}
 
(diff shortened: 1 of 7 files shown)


sdanyliv:
Looks like it needs deep analysis. Postpone to next release.

igor-tkachev:
What is a usage scenario?

MaceWindu:
rebased

jods4:
I am not sure this is the best design.
The conditions in which a join is actually "weak" are not well defined for developer and some welcome optimizations such as the one discussed in #2815 will not automatically apply.

From the example in the original post above, the inner joins would be weak and could be removed without changing the result only if their cardinality was known to be 1..1 (so they don't filter out the result, nor multiply the result lines).

I am musing the idea that the key part here is knowing the cardinality of joins, as a min..max number of matching rows.

Some joins/applies can have interesting default, for example all outer joins and apply have a min of 1 as they never filter row. Which immediately imply that they could be dropped in the context of an [NOT] EXISTS, which is #2815.

Sometimes the user knows more about cardinality than linq2db, for ex. because a foreign key is not nullable or points to a unique key. And if the user can provide this info to linq2db it could lead to various optimizations in various contexts in a more controlled way than a `WeakLeftJoin`.

So an idea could be to have an extra parameter to the joins function and the original example would look like:
```cs
var query =
	from t in table
	from j1 in forJoin.InnerJoin(j1 => t.Id == j1.ParentId, Cardinality.ExactlyOne)
	from j2 in forJoin.InnerJoin(j2 => t.Id == j1.Id, Cardinality.ExactlyOne)
	select new
	{
		T = t,
		J1 = j1,
		J2 = j2
	};
```
I think the benefits are:
- The meaning and correctness is more clear to the end-user;
- It's more discoverable and intuitive than an additional collection of `WeakXXX` overloads.
- The optimization can be assessed by linq2db based on context (more about this below).

For the 3rd point, the thing is, one join is not always "weak" in the same sense in every context.
In the original example a join is truly weak if it results in _exactly one line_.
Inside an EXISTS, a join is weak if it results in _at least one line_.

Finally, that cardinality could sometimes have better default values if the model uses for example `HasUniqueKey`, so linq2db would automatically be able to apply some optimizations without any change in user code.

What do you think?

jods4:
Another benefit I just thought of: maybe you can merge other existing optimizations into this.
I'm thinking of #2490.

Basically, some APPLY or CROSS JOIN could deduce their cardinality automatically based on what they're joining to.
In #2490 I noted that aggregates without GROUP BY always result into exactly 1 row.

So if the optimizer flows this information into the join cardinality, it can then apply the same logic from this PR and there's no need for a special rule in #2490.

Some similar optimizations could be automatically made if a dynamic query results into an always false condition (e.g. X in empty list). Then the cardinality is 0 and this PR could remove the joins in some conditions:
- for INNER, CROSS the whole query will never have results (FALSE is inside EXISTS)
- for OUTER, it can be removed if unused because it won't multiply rows.

sdanyliv:
Outdated

jods4:
@sdanyliv out of curiosity, is there some metadata in v6 that users can use to indicate the cardinality of joins, e.g. so that an unused INNER JOIN could be removed because it's 1-to-1 anyway?

sdanyliv:
Yes, join has `Cardinality` and `IsSubqueryExpression` - which means one record.

Issue #817, opened 2017-08-23
Incorrect query generated when Take/Skip used for empty select


MaceWindu:
linq2db generates invalid SQL for queries like that for some providers (DB2, SQL Server, SQL CE):
```
// also with other Take/Skip combinations
db.GetTable<Person>().Take(1).Select(_ => new { }).ToList();
```
I will create PR with tests.

### Environment details
linq2db version: *1.9.0-prerelease*

PR #2287, opened 2020-06-19
Fix SQL Server optimizations for TOP/OFFSET/FETCH

### LINKED ISSUE #2282: Enable TOP in subqueries for MS-SQL
When using the MS SQL Server 2017 data provider, `FirstOrDefault()` in a subquery does not translate to `SELECT TOP 1` in the produced SQL, generating a slightly more complex execution plan.

### Steps to reproduce

I am executing code as follows:
```csharp
var provider = new SqlServerDataProvider(ProviderName.SqlServer, SqlServerVersion.v2017);
var db = new DataContext(provider, Configuration.GetConnectionString("testdb"));
var categories = db.GetTable<Category>();
var productCategories = db.GetTable<Product_Category_Mapping>();
var products = db.GetTable<Product>();
var test4 =
    from product in products
    select new {
        Product = product,
        Category = (
            from cat in categories
            join pcat in productCategories on cat.Id equals pcat.CategoryId
            where pcat.ProductId == product.Id
            select new { cat.Id, cat.Name }).FirstOrDefault()
    };
var test4ret = await test4.Take(10).ToListAsync();
```

With these classes:
```csharp
class Product
{
    public int Id { get; set; }

}
class Product_Category_Mapping
{
    public int Id { get; set; }
    public int ProductId { get; set; }
    public int CategoryId { get; set; }
    public bool IsFeaturedProduct { get; set; }
    public bool DisplayOrder { get; set; }
}
class Category
{
    public int Id { get; set; }
    public int ParentCategoryId { get; set; }
}
```

Produced SQL:
```sql
DECLARE @take Int -- Int32
SET     @take = 10
DECLARE @take_1 Int -- Int32
SET     @take_1 = 1

SELECT TOP (@take)
        [product_1].[Id],
        [product_1].[Name],
        [t2].[Id],
        [t2].[Name],
        [t2].[is_empty]
FROM
        [Product] [product_1]
                OUTER APPLY (
                        SELECT
                                [t1].[Id],
                                [t1].[Name],
                                [t1].[is_empty]
                        FROM
                                (
                                        SELECT
                                                [cat].[Id],
                                                [cat].[Name],
                                                1 as [is_empty],
                                                ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) as [RN]
                                        FROM
                                                [Category] [cat]
                                                        INNER JOIN [Product_Category_Mapping] [pcat] ON [cat].[Id] = [pcat].[CategoryId]
                                        WHERE
                                                [pcat].[ProductId] = [product_1].[Id]
                                ) [t1]
                        WHERE
                                [t1].[RN] <= @take_1
                ) [t2]
```

Expected SQL:
```sql
DECLARE @take Int -- Int32
SET     @take = 10
DECLARE @take_1 Int -- Int32
SET     @take_1 = 1

SELECT TOP (@take)
        [product_1].[Id],
        [product_1].[Name],
        [t2].[Id],
        [t2].[Name],
        [t2].[is_empty]
FROM
        [Product] [product_1]
                OUTER APPLY (
                        SELECT TOP (@take_1)
                                [t1].[Id],
                                [t1].[Name],
                                [t1].[is_empty]
                        FROM
                                (
                                        SELECT
                                                [cat].[Id],
                                                [cat].[Name],
                                                1 as [is_empty]
                                        FROM
                                                [Category] [cat]
                                                        INNER JOIN [Product_Category_Mapping] [pcat] ON [cat].[Id] = [pcat].[CategoryId]
                                        WHERE
                                                [pcat].[ProductId] = [product_1].[Id]
                                ) [t1]
                ) [t2]
```
Or possibly this: (which produces an identical execution plan)
```sql
DECLARE @take Int -- Int32
SET     @take = 10
DECLARE @take_1 Int -- Int32
SET     @take_1 = 1

SELECT TOP (@take)
        [product_1].[Id],
        [product_1].[Name],
        [t1].[Id],
        [t1].[Name],
        [t1].[is_empty]
FROM
        [Product] [product_1]
                OUTER APPLY (
                        SELECT TOP (@take_1)
                                [cat].[Id],
                                [cat].[Name],
                                1 as [is_empty]
                        FROM
                                [Category] [cat]
                                        INNER JOIN [Product_Category_Mapping] [pcat] ON [cat].[Id] = [pcat].[CategoryId]
                        WHERE
                                [pcat].[ProductId] = [product_1].[Id]
                ) [t1]
```

Here is the differences in the actual execution plan with the produced SQL vs the expected SQL: (acutal on top, expected on bottom)
image

### Environment details
linq2db version: 3.0.0-rc.1
Database Server: Microsoft Azure SQL Database
Database Provider: SqlServer v2017
Operating system: Windows 10 Pro 1909
.NET Framework: .Net Core 3.1

Shane32:
Fixes #2282 

This fix performs different checks for each version of SQL Server to match the supported features.

SQL Server 2000:
* Skip/Take not supported for updates
* Skip not supported at all (either via ROW_NUMBER or via OFFSET) - handled by SqlProviderFlags
* Only constants are allowed in TOP parameters - handled by SqlProviderFlags

SQL Server 2005/2008:
* Skip supported via ROW_NUMBER

SQL Server 2012/2017:
* Skip supported via OFFSET, but requires ORDER BY

This fix adds a parameter to `BasicSqlOptimizer.ReplaceTakeSkipWithRowNumber` to specify if the replacement should operate only on queries with Skip, allowing operations using TOP to pass through unaltered.  This is exposed as part of the public API, however, so it is a breaking change.  I can duplicate this function within `SqlServerSqlOptimizer` if it is desired to not change the public API.  Note that currently this function is only used by MS SQL Server, and the new flag is always going to be `true`, so another option is to change the functionality of the function without changing the signature.

I also consolidated the `ConvertExpression` and `ConvertConvertFunction` code into `SqlServerSqlOptimizer`, since the code was identical in all versions of the Sql optimizer.  This does not change public APIs and does not affect operation in any way.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2287) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2287) before we can accept your contribution.<br/><br/>:x: Shane32 [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2287)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=2287) it.</sub>

Source/LinqToDB/DataProvider/SqlServer/SqlServerSqlOptimizer.cs
```diff
@@ -15,15 +15,6 @@ protected SqlServerSqlOptimizer(SqlProviderFlags sqlProviderFlags, SqlServerVers
...
 		}
 
-		public override SqlStatement TransformStatement(SqlStatement statement)
-		{
-			statement = SeparateDistinctFromPagination(statement);
-			statement = ReplaceDistinctOrderByWithRowNumber(statement);
-			statement = ReplaceTakeSkipWithRowNumber(statement, false);
-			statement = QueryHelper.OptimizeSubqueries(statement);
-			return statement;
-		}
```
Shane32, line 25: This code was only previously used by `SqlServer2005SqlOptimizer`.  The code has been moved there.

Source/LinqToDB/DataProvider/SqlServer/SqlServer2005SqlOptimizer.cs
```diff
@@ -9,14 +9,16 @@ public SqlServer2005SqlOptimizer(SqlProviderFlags sqlProviderFlags) : base(sqlPr
...
-			expr = base.ConvertExpression(expr);
+			//SQL Server 2005 supports ROW_NUMBER but not OFFSET/FETCH
 
-			if (expr is SqlFunction)
-				return ConvertConvertFunction((SqlFunction)expr);
-
-			return expr;
+			statement = SeparateDistinctFromPagination(statement);
+			statement = ReplaceDistinctOrderByWithRowNumber(statement);
+			statement = ReplaceTakeSkipWithRowNumber(statement, false, true);
```
Shane32, line 18: With the `true` parameter, now only queries with Skip will be converted here.

Source/LinqToDB/DataProvider/SqlServer/SqlServerSqlOptimizer.cs
sdanyliv: It's ok, let's check regressions ))

Source/LinqToDB/DataProvider/SqlServer/SqlServer2008SqlOptimizer.cs
```diff
@@ -5,57 +5,18 @@ namespace LinqToDB.DataProvider.SqlServer
...
 		{
 		}
 
 		public override SqlStatement TransformStatement(SqlStatement statement)
 		{
-			statement = SeparateDistinctFromPagination(statement);
-			statement = ReplaceTakeSkipWithRowNumber(statement, false);
+			//SQL Server 2008 supports ROW_NUMBER but not OFFSET/FETCH
 
-			CorrectRootSkip(statement.SelectQuery!);
```
Shane32, line 21: `CorrectRootSkip` was useless here since `ReplaceTakeSkipWithRowNumber` executed first, removing all skips/takes.  It has been removed

Source/LinqToDB/DataProvider/SqlServer/SqlServer2008SqlOptimizer.cs
```diff
@@ -5,57 +5,18 @@ namespace LinqToDB.DataProvider.SqlServer
...
+			statement = SeparateDistinctFromPagination(statement);
+			statement = ReplaceTakeSkipWithRowNumber(statement, false, true);
```
Shane32, line 17: With the `true` parameter, now only queries with Skip will be converted here.

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
```diff
@@ -16,14 +16,23 @@ protected SqlServer2012SqlOptimizer(SqlProviderFlags sqlProviderFlags, SqlServer
 
 		public override SqlStatement TransformStatement(SqlStatement statement)
 		{
-			if (statement.IsUpdate())
-				statement = ReplaceTakeSkipWithRowNumber(statement, false);
```
Shane32, line 20: Is special code needed for update queries with skip/take to convert them to use ROW_NUMBER ?  If so, it should probably be added for 2005 and later versions, not just 2012/2017.  Is there a test (sample LINQ/SQL query) that I should review?

sdanyliv: Search for Take, Skip
https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Update/UpdateTests.cs

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
```diff
@@ -16,14 +16,23 @@ protected SqlServer2012SqlOptimizer(SqlProviderFlags sqlProviderFlags, SqlServer
...
-			else
-			{
-				statement = ReplaceTakeSkipWithRowNumber(statement, true);
-				CorrectRootSkip(statement.SelectQuery!);
```
Shane32, line 24: Previously, the code would convert all subqueries with skip/take to use ROW_NUMBER, and then the root would be patched to add an ORDER BY if necessary with OFFSET.  I tested OFFSET in subqueries and it works fine.  So these were replaced with a recursive function to update the query and all subqueries to include a simple ORDER BY if necessary.  Note that an ORDER clause is also necessary for ROW_NUMBER so performance should be no worse if this conversion takes place.

Source/LinqToDB/DataProvider/SqlServer/SqlServerSqlOptimizer.cs
```diff
@@ -85,48 +109,5 @@ public override ISqlExpression ConvertExpression(ISqlExpression expr)
 			return expr;
 		}
 
-		public ISqlExpression ConvertConvertFunction(SqlFunction func)
```
Shane32, line 88: This function was consolidated into `ConvertExpression` since it was used by all descendant classes in the same manner.

sdanyliv:
@Shane32, as you can see `UpdateTestJoinSkip` test failed.

Shane32:
Yup thanks!  Looks like also `TestUnorderedSkip` and a couple variations failed also.  I'll check that out.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1921,15 +1921,29 @@ protected SqlStatement SeparateDistinctFromPagination(SqlStatement statement)
 		/// <param name="onlySubqueries">Indicates when transformation needed only for subqueries.</param>
 		/// <returns>The same <paramref name="statement"/> or modified statement when transformation has been performed.</returns>
 		protected SqlStatement ReplaceTakeSkipWithRowNumber(SqlStatement statement, bool onlySubqueries)
```
Shane32, line 1923: This function has been reverted to its original signature with original functionality, to eliminate breaking changes with third party APIs.  This method may be removed, or marked as depreciated, as it is not used anymore.  An overload has been added for the updated functionality.

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
```diff
@@ -16,14 +16,25 @@ protected SqlServer2012SqlOptimizer(SqlProviderFlags sqlProviderFlags, SqlServer
...
+		}
+
+		/// <summary>
+		/// Adds an ORDER BY clause to queries using OFFSET/FETCH, if none exists
+		/// </summary>
+		protected SqlStatement AddOrderByForSkip(SqlStatement statement)
+		{
+			ConvertVisitor.Convert(statement, (visitor, element) => {
+				if (element is SelectQuery query && query.Select.SkipValue != null && query.OrderBy.IsEmpty)
+					query.OrderBy.ExprAsc(new SqlValue(typeof(int), 1));
```
Shane32, line 35: I have no idea why `TestUnorderedSkip("SqlServer.2017.LinqService")` fails when `TestUnorderedSkip("SqlServer.2017")` runs fine.  I'm not sure how to test it on my dev machine to determine what will fix the problem.  For this commit, I tried changing `query.OrderBy.ExprAsc(new SqlExpression("1"));` in the line above to instead use `SqlValue`.  

I would guess that it has to do with the fact that the test selects no columns -- it is written as:
```csharp
var result = db.GetTable<Person>().Skip(1).Select(_ => new { }).ToList();
```

The prior code within `CorrectRootSkip` would pull keys from the table specified in the FROM section into the SELECT.  I can only guess that this is the reason that the test fails now whereas it did not before.

I can use the code from the removed `CorrectRootSkip` function in here instead, which would probably fix the problem, but it looks to be pretty stupid code, unless I'm totally missing something.

I would suggest changing the test to return a column from the table.

Shane32: Side note: I tested MS SQL 2017 and found that `ORDER BY 1` is optimized out of the execution plan, whereas `ORDER BY (SELECT NULL)` does not.  So that is why I have `ORDER BY 1` versus the old `ORDER BY (SELECT NULL)`.

Shane32:
Please review the above comments and confirm:
- Should I remove the old `ReplaceTakeSkipWithRowNumber` method, or mark as deprecated?
- Should I change the `TestUnorderedSkip` tests to return a column?
- Any other comments on why this test would fail with the `SqlServer.2017.LinqService` provider but not the `SqlServer.2017` provider?  Any way for me to test it locally?

Or you can run another set of tests.  Thanks!!!

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
Shane32: Also: I searched all the tests throughout the code base for `new { }`.  I cannot find any other tests that do not return any columns from the underlying table besides the ones here in `Issue817Tests.cs`.  The only other match for `new { }` is in the `GroupByExtensionsTests.cs`.  If it is desired to be able to execute such a query, I suggest that the tests for that functionality be moved to a separate test.

Shane32:
I did find a relevant PR - #817 - and I still can't understand why there should be code in the skip/take processing to handle selecting an empty query, nor why that would even be desirable.  In any case, I can write code to re-implement the patch that was done at that time.  So long as nobody tries to select _nothing_, it should not matter anyway.

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
sdanyliv: LinqService tests let for us. We will fix them.

Shane32:
I'm installing SQL 2000, 2005, 2008, 2012, 2014, and 2017 in my dev environment to properly test this before posting further code revisions.  So far, I'm unable to get the SQL Server 2000 create script (and various tests) to execute successfully on SQL 2000 Standard SP4 (on Windows 2008 x86), so I might just leave that edition alone, but it looks like I will be able to properly test the other editions.  The error executing the create script within SQL Server 2000 is `Could not find stored procedure 'sys.sp_addextendedproperty'`.  Let me know if you're testing with MSDE or if there's anything else I need to know for SQL Server 2000 testing.

Shane32:
Still failing a couple tests for SQL Server 2005.  I'll work on it tomorrow.

Shane32:
All set now; please confirm with another set of azure regression tests.

This PR also fixes another bug where deleting entries with Skip and Take did not work.  Tests have been added, which failed before this update.  I didn't run the new tests against any non-SQL providers, or SQL 2000, so the excluded provider list may need to be updated (or bugs fixed in their optimizers).  Please let me know if you'd like me to add an issue for tracking purposes.

MaceWindu:
I'm currently trying to add more sql server versions to CI. As for SQL2000 we actually don't have real SQL2000 version for testing right now, so I'm not surptised we added some incompatible sql to scripts. Will fix this specific issue you mentioned (you can just remove it from script), but it could be not the only one issue.

Shane32:
Perhaps I can look at SQL Server 2000 testing in another PR, since I already have SQL Server 2000 running in my dev environment.

Shane32:
It appears that the only failing tests are the same tests that are failing in other PRs.

Shane32:
By the way, the `CorrectEmptyRoot` function converts `SELECT *` queries into `SELECT {key}` queries.  This only exists to prevent `Issue817` tests with the `SqlServer.{version}.LinqService` provider from failing.  Those tests with the SQL Server 2000-2008 provider have already been marked with `ActiveIssue`.  If you like, I can remove the `CorrectEmptyRoot` function and then adjust the `ActiveIssue` to include all versions of SQL Server rather than just SQL Server 2000-2008.

Or, I can remove the `ActiveIssue` flags from applying to the SQL Server 2000-2008 providers, as the `CorrectEmptyRoot` patch fixes the `Issue817` tests for those providers.

Please let me know which is preferred, and I will update the PR.

One other question: the SQL Server 2005 optimizer contained a call to `OptimizeSubqueries`.  This call was not utilized within the other SQL Server optimizers.  This call seems to only work properly with SELECT queries.  As it is now, I have left the `OptimizeSubqueries` call in the SQL Server 2005 provider, and have not added it to any of the other optimizers.  Shall I add it to the other SQL Server optimizers also, or remove it from the SQL Server 2005 optimizer, or leave it as-is?  Let me know if a change should be made here.

MaceWindu:
I think you can just mark it with active issue. If I'm not mistaken, `select *` query caused by current implementation of `AlternativeBuildSql` method, which should be fixed to explicitly specify selected columns instead of * (also will remove rownumber column from selection)

Source/LinqToDB/DataProvider/SqlServer/SqlServer2012SqlOptimizer.cs
MaceWindu: This is expected to fail. Just mark it with active issue, as we shouldn't generate `select *` queries

MaceWindu: Or if you can - you can try to fix it to return something like `select 1` (we anyway don't consume selected data here)

MaceWindu: > Any other comments on why this test would fail with the SqlServer.2017.LinqService provider but not the SqlServer.2017 provider?

Because for linqservice we don't support * in select list so all such queries fail for it. While we can fix support for such queries, I don't want to do it, as we shouldn't generate * in first place, so such failing tests are good indicator that we need to fix something.

MaceWindu:
@Shane32, could you post remaining questions in single message as it quite hard to track remaining questions

Tests/Linq/Update/DeleteTests.cs
```diff
@@ -328,6 +328,103 @@ where ids.Contains(p.ParentID)
 			}
 		}
 
+		[Test]
+		public void DeleteTakeOrdered(
+			[DataSources(
```
MaceWindu, line 333: You can use IncludeDataSources attribute to explicitly specify tested providers instead of specifying lots of excludes

Shane32:
@MaceWindu I've updated the code as you indicated, removing the patch for the SELECT * issue from the SQL Server optimizers and flagging the appropriate tests as an active issue.

This is the only remaining question:
> One other question: the SQL Server 2005 optimizer contained a call to OptimizeSubqueries. This call was not utilized within the other SQL Server optimizers. This call seems to only work properly with SELECT queries. As it is now, I have left the OptimizeSubqueries call in the SQL Server 2005 provider, and have not added it to any of the other optimizers. Shall I add it to the other SQL Server optimizers also, or remove it from the SQL Server 2005 optimizer, or leave it as-is? Let me know if a change should be made here.

If you're satisfied with the PR as-is, I suggest running one final set of azure tests and then merging.  Thanks!

MaceWindu:
@sdanyliv could you check `OptimizeSubqueries` question? It's your area. Otherwise I think we can merge it tomorrow so it will be available in 3.0.0

sdanyliv:
@Shane32, `OptimizeSubqueries` were introduced for revert work of `SeparateDistinctFromPagination` if it is doable and maybe other functions which increase nesting via `QueryHelper.WrapQuery`.

Shane32:
@sdanyliv Then I suggest adding it to the 2008 optimizer also, as it also requires `SeparateDistinctFromPagination`.  Would you agree?  The 2012/2017 optimizer should not create any redundant nesting, as it only wraps the root for specific DELETE/UPDATE queries when necessary.

sdanyliv:
@Shane32, sounds reasonable.

Shane32:
Thanks!  The code is updated as you suggested.  I am running local tests now for SQL Server 2005, 2008, 2012, 2014 and 2017; they will be done in about 15 minutes.  I do not anticipate any issues.

@MaceWindu If you can cancel the prior azure pipelines run and start it again with the latest push, that would be great.

I will confirm when all tests pass on my local machine.

Shane32:
All the tests pass for SQL Server 2005-2017 here in my local environment (other than the preexisting MiniProfiler test failures with SQL Server 2005, fixed in PR #2293).

PR #2303, opened 2020-06-29
Implement .Net Standard 2.1 IAsyncEnumerable + add new build targets

Shane32:
This PR comprises the following changes:
* Changes existing IAsyncEnumerable definitions within .Net 4.5 or .Net 4.6 to match .Net Standard 2.1 signatures
* Imports MS backwards compatibility layer (bcl) for .Net Standard 2.0 projects
* Adds .Net Standard 2.1 and .Net Core 3.0 targets for linq2db project
* Adds .Net Core 3.0 target for test projects

Notes:
* A set of tests against SQL Server 2017 pass on my machine
* As of now, for .Net 4.5 / 4.6 projects, DisposeAsync() returns `Task` rather than `ValueTask`.
* The old `IAsyncEnumerable` and `IAsyncEnumerator` interfaces are still within the `LinqToDB.Async` namespace (for .Net 4.5 / 4.6 projects).  I cannot easily make them `internal` since some other public interfaces inherit from them.  I can move them to the `Compatibility` folder and move them into the `System.Collections.Generic` namespace, if you like.

Unless otherwise directed, I plan to add the following changes to this PR:
* Import `System.Threading.Tasks.Extensions` for .Net 4.5 / 4.6 projects, so that `ValueTask` can be utilized for `DisposeAsync` signatures, unifying the DisposeAsync signature across all targets
* Change `IAsyncDbConnection` and `IAsyncDbTransaction` to implement `IAsyncDisposable` (they have explicitly defined `Task DisposeAsync()` methods now)
* Add `ToAsyncEnumerable` method for .Net Standard 2.0+ projects to `LinqToDB.LinqExtensions` (map to `AsyncExtensions.AsAsyncEnumerable`)

Please review the code so far, and also let me know if you have any comments about the remaining proposed changes.  If you like, I can make the proposed additional changes in a separate PR.

PR #2304, opened 2020-06-30
Draft: Use ValueTask for .Net 4.5 / .Net 4.6

Shane32:
This builds on #2303 and imports the `System.Threading.Tasks.Extensions` nuget package for .Net 4.5 and .Net 4.6.  However, it doesn't work.  So far I get this error:
```
System.MethodAccessException : Attempt by method 'LinqToDB.Linq.QueryRunner+<ExecuteQueryAsync>d__14`1<T>.MoveNext()' to access method 'LinqToDB.Async.IAsyncDisposable.DisposeAsync()' failed.
```
It seems that the problem occurs when the JiT compiler attempts to compile the `QueryRunner.ExecuteQueryAsync` code, which contains a reference to the `DisposeAsync()` method -- so in a sense, it's not a runtime error, although it's occurring at runtime.

I've tried multiple techniques to get past this problem, but all have failed.  If anyone has any ideas, please take a look and/or try making additional changes to the code.

I've also changed the `await using` blocks to use more basic `try..finally` blocks, but that didn't help either.  Once the underlying problem is solved, it would probably be fairly trivial to write a `ConfigureAwait` handler for `IAsyncDisposable` so that `await using` can be used.

### DIFF
--- a/NuGet/linq2db.nuspec
+++ b/NuGet/linq2db.nuspec
@@ -9,8 +9,12 @@
 			linq linq2db LinqToDB ORM database DB SQL SqlServer Access SqlCe SqlServerCe MySql Firebird SQLite Sybase Oracle ODP PostgreSQL DB2 Informix SapHana
 		</tags>
 		<dependencies>
-			<group targetFramework=".NETFramework4.5" />
-			<group targetFramework=".NETFramework4.6" />
+			<group targetFramework=".NETFramework4.5">
+				<dependency id="System.Threading.Tasks.Extensions"    version="4.5.4" />
+			</group>
+			<group targetFramework=".NETFramework4.6">
+				<dependency id="System.Threading.Tasks.Extensions"    version="4.5.4" />
+			</group>
 			<group targetFramework=".NETStandard2.0">
 				<dependency id="Microsoft.CSharp"                     version="4.7.0" />
 				<dependency id="System.ComponentModel.Annotations"    version="4.7.0" />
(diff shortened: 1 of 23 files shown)


Shane32:
All the new code for this PR is in the very last commit:
https://github.com/linq2db/linq2db/pull/2304/commits/af38d5706901782ca9d5048c2c5ea9bf9d91fc30

MaceWindu:
it fails because we declare return type as Task but return ValueTask actually. I've changed one place as proof-of-concept, but I really don't like this situation:
- there is no compiler error for this situation, so we need actually inspect all callers and fix them (and probably their callers)
- it will result in same error for users of async methods in net45/net46 code
- CAS generate error for async tests if we have AllowPartiallyTrustedCallers attribute and removing it also is breaking change

For me it looks like major breaking change which is not worth it. Probably we better maintain several ifdefs in our code to support netfx targets

MaceWindu:
quite sad situation, but implementing this will be major disaster judging from netfx compiler/runtime behavior

Shane32:
I agree.  Close this PR?

MaceWindu:
WAit a second, I'm trying something

MaceWindu:
ok, close. I tried to use AsTask, but AsyncTaskMethodBuilder triggered anyway instead of AsyncValueTaskMethodBuilder. Looks like netfx compiler cannot handle ValueTask in Task method properly...

PR #2312, opened 2020-07-03
Update AsyncDbConnection & AsyncDbTransaction to utilize ValueTask

Shane32:
This updates the following for .Net Standard 2.0+ targets:
* The following methods now return a `ValueTask`, matching .Net Standard 2.1 methods:
  * `IAsyncDbConnection.BeginTransactionAsync`
  * `IAsyncDbConnection.DisposeAsync`
  * `IAsyncDbTransaction.DisposeAsync`
* The same methods listed above are optimized for `ValueTask` when using reflection to create an `IAsyncDbConnection` or `IAsyncDbTransaction`

There is an additional change for .Net Standard 2.1:
* When using the `AsyncDbConnection` and `AsyncDbTransaction` classes to wrap `DbConnection` and `DbTransaction` classes (historically used to wrap synchronous `DbConnection` and `DbTransaction` classes with async methods), async methods will map to .Net Standard 2.1 async methods when available.  If the provider does inherit from `DbConnection` or `DbTransaction`, but did not implement the async methods, .Net will call the underlying synchronous method.

I've also added code to look for `DbConnection` and `DbTransaction` classes on .Net Standard 2.1, and skip reflection within `AsyncFactory.Create` and map async methods directly to the .Net Standard 2.1 async methods.  However, this is commented out, as if the database provider was compiled against .Net Standard 2.0 or prior, it will not implement the async methods, and instead the default .Net implementation of the async methods will call the synchronous version -- in this case, using reflection to obtain a handle to the async methods would be preferred.

It is likely possible to utilize the updated `AsyncDbConnection` and `AsyncDbTransaction` classes when wrapping providers that are known to be compiled against .Net Standard 2.1.  I did not look into this.

I only have SQL Server running, and as this code is highly dependent on the provider in use, it will need to be tested against all providers.


Source/LinqToDB/Async/AsyncFactory.cs
```diff
@@ -70,6 +74,15 @@ public static IAsyncDbConnection Create(IDbConnection connection)
...
 				return asyncConnection;
 
+/* this does not work when connecting to a .Net Standard 2.0 library with async methods,
+ *   since the async methods will hide (not override) the .Net Standard 2.1 async methods
+#if NETSTANDARD2_1 || NETCOREAPP3_1
+			// wrap the asynchronous methods already available
+			if (connection is DbConnection dbConnection)
+				return new AsyncDbConnection(dbConnection);
+#endif
+*/
```
Shane32, line 84: Please review.  I suggest leaving the comment block in place.

Source/LinqToDB/Async/AsyncFactory.cs
```diff
@@ -87,6 +100,15 @@ public static IAsyncDbTransaction Create(IDbTransaction transaction)
...
 				return asyncTransaction;
 
+/* this does not work when connecting to a .Net Standard 2.0 library with async methods,
+ *   since the async methods will hide (not override) the .Net Standard 2.1 async methods
+#if NETSTANDARD2_1 || NETCOREAPP3_1
+			// wrap the asynchronous methods already available 
+			if (transaction is DbTransaction dbTransaction)
+				return new AsyncDbTransaction(dbTransaction);
+#endif
+*/
```
Shane32, line 110: Please review.  I suggest leaving the comment block in place.

Source/LinqToDB/Async/AsyncFactory.cs
```diff
@@ -139,50 +177,74 @@ private static async Task<IAsyncDbTransaction> Wrap<TTransaction>(Task<TTransact
...
 			// Availability:
 			// - (stub) DbConnection
-			var openAsync               = CreateDelegate<Func<IDbConnection, CancellationToken, Task>, IDbConnection>(type, "OpenAsync", _tokenParams, _tokenParams, _tokenParams, false);
+			var openAsync               = CreateDelegate<Func<IDbConnection, CancellationToken, Task>, IDbConnection>(type, "OpenAsync", _tokenParams, _tokenParams, _tokenParams, false, false);
 
-			// ValueTask CloseAsync()
+			// Task CloseAsync(CancellationToken)
 			// Availability:
-			var closeAsync              = CreateDelegate<Func<IDbConnection, Task>, IDbConnection>(type, "CloseAsync", Array<Type>.Empty,   _tokenParams     , _noTokenParams,    false)
+			var closeAsync              = CreateDelegate<Func<IDbConnection, Task>, IDbConnection>(type, "CloseAsync", Array<Type>.Empty,   _tokenParams     , _noTokenParams,    false, false)
```
Shane32, line 222: I changed the comment here, as it appears the code was not searching for a `ValueTask CloseAsync()` as previously indicated.  I did not change the code to match the comment.

Source/LinqToDB/Data/RetryPolicy/RetryingDbConnection.cs
```diff
@@ -113,15 +113,29 @@ public object Clone()
...
+			=> _connection.BeginTransactionAsync(cancellationToken);
+
+		public ValueTask<IAsyncDbTransaction> BeginTransactionAsync(IsolationLevel isolationLevel, CancellationToken cancellationToken = default)
+			=> _connection.BeginTransactionAsync(isolationLevel, cancellationToken);
+#else
+		public new ValueTask<IAsyncDbTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default)
+			=> _connection.BeginTransactionAsync(cancellationToken);
+
+		public new ValueTask<IAsyncDbTransaction> BeginTransactionAsync(IsolationLevel isolationLevel, CancellationToken cancellationToken = default)
+			=> _connection.BeginTransactionAsync(isolationLevel, cancellationToken);
```
MaceWindu, line 133: I don't think we need those methods, as underlying implementation already calls BeginDbTransactionAsync https://github.com/dotnet/runtime/blob/master/src/libraries/System.Data.Common/src/System/Data/Common/DbConnection.cs#L87

we do same for BeginTransaction

Source/LinqToDB/Async/AsyncFactory.cs
MaceWindu: We shouldn't have this code, as it breaks RegisterConnectionFactory functionality. If user wants to resolve connection in this way - he will register it using RegisterConnectionFactory

Source/LinqToDB/Async/AsyncFactory.cs
MaceWindu: Same here. This is a wrong place for such shortcuts

Source/LinqToDB/Async/AsyncFactory.cs
MaceWindu: Hmm, I wonder why I even added it in first place if there is no availability details. Let it be like that for now

MaceWindu:
Hmm, github managed to rebase PR, but failed to preserve diff :)

Source/LinqToDB/Data/RetryPolicy/RetryingDbConnection.cs
Shane32: These methods return `IAsyncDbTransaction`s, and are necessary to implement `IAsyncDbConnection`.  `BeginDbTransactionAsync` is necessary to implement `DbConnection`, which this class inherits from.  The `BeginTransactionAsync` methods of `DbConnection` are sealed, as they point to `BeginDbTransactionAsync`.

Source/LinqToDB/Async/AsyncFactory.cs
Shane32: Shall I delete this code and the comment (in both places)?

MaceWindu: Yep

Shane32:
'Squash and merge' commits are seen as different commits than the commits they comprise of.  Merging the "squash and merged" commit back into the history of the other branch (this branch), as you just did, fixes it.

Source/LinqToDB/Async/AsyncFactory.cs
Shane32, line 84:
```suggestion
```

Source/LinqToDB/Async/AsyncFactory.cs
Shane32, line 110:
```suggestion
```

Source/LinqToDB/Async/AsyncFactory.cs
Shane32: 👍

Source/LinqToDB/Data/RetryPolicy/RetryingDbConnection.cs
MaceWindu: Whoops, sorry. Completely overlooked return type

PR #2314, opened 2020-07-03
Bulk copy async support

Shane32:
Here's a draft of adding `BulkCopyAsync` to `IDataProvider`.  Here are the three signatures (one old + two new):
```csharp
BulkCopyRowsCopied       BulkCopy<T>     (ITable<T> table, BulkCopyOptions options, IEnumerable<T> source);
Task<BulkCopyRowsCopied> BulkCopyAsync<T>(ITable<T> table, BulkCopyOptions options, IEnumerable<T> source);
Task<BulkCopyRowsCopied> BulkCopyAsync<T>(ITable<T> table, BulkCopyOptions options, IAsyncEnumerable<T> source);
```
For any .Net Standard 2.1 data provider that supports `DbDataReader` as an input and has an asynchronous bulk copy method, such as `Microsoft.Data.SqlClient`, the 3rd method will asynchronously read data from the input source.  See here:

https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlBulkCopy.cs#L1038

This draft favors a few extra heap allocations per BulkCopy call over duplicating code.  This does not occur within a loop.  It is designed to still call the synchronous underlying methods when `BulkCopy` is called.

Note that the `BulkCopyReader` class is now a generic class - this was necessary to support `IAsyncEnumerable<T>`.  If we add a constraint of `where T : class` to these bulk copy methods, then `IAsyncEnumerable<T>` could be directly cast to `IAsyncEnumerable<object>` and BulkCopyReader would not need to be a generic class.

So far, the code won't work because the `SQLServerProviderAdapter.SqlBulkCopy` class does not yet have the `WriteToServerAsync` call added.  I have not yet looked at any other provider.


Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -76,6 +179,11 @@ protected internal static string GetTableName<T>(ISqlBuilder sqlBuilder, BulkCop
 		#region ProviderSpecific Support
 
 		protected void TraceAction(DataConnection dataConnection, Func<string> commandText, Func<int> action)
+		{
+			TraceActionAsync(dataConnection, commandText, () => Task.FromResult(action())).Wait();
+		}
+
+		protected async Task TraceActionAsync(DataConnection dataConnection, Func<string> commandText, Func<Task<int>> action)
 		{
```
Shane32, line 187: This deduplicates this trace code at the cost of a single `Task<T>` allocation.  Since `action` (and the rest of the code in `TraceActionAsync`) will always run synchronously, any errors will bubble up, so the call to `Wait()` is not actually necessary.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -216,38 +476,107 @@ protected BulkCopyRowsCopied MultipleRowsCopy1(MultipleRowsHelper helper, IEnume
...
+			helper.RowsCopied.RowsCopied++;
+			helper.CurrentCount++;
+		}
+
+		private void MultipleRowsCopy2Finish(MultipleRowsHelper helper)
+		{
+			helper.StringBuilder.Length -= " UNION ALL".Length;
+		}
+
+		protected BulkCopyRowsCopied MultipleRowsCopy3(
```
Shane32, line 499: MultipleRowsCopy3 is not used anywhere.  Should it be removed?  I already added async functions for it, but maybe it should just be deleted.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -128,13 +236,93 @@ protected void TraceAction(DataConnection dataConnection, Func<string> commandTe
 
 		#region MultipleRows Support
 
-		protected BulkCopyRowsCopied MultipleRowsCopy1<T>(
-			ITable<T> table, BulkCopyOptions options, IEnumerable<T> source)
+		protected BulkCopyRowsCopied MultipleRowsCopy1<T>(ITable<T> table, BulkCopyOptions options, IEnumerable<T> source)
+			=> MultipleRowsCopy1(new MultipleRowsHelper<T>(table, options), source);
+
+		protected BulkCopyRowsCopied MultipleRowsCopy1(MultipleRowsHelper helper, IEnumerable source)
```
Shane32, line 242: To deduplicate code between `MultipleRowsCopy1`, `MultipleRowsCopy1Async(...IEnumerable)` and `MultipleRowsCopy1Async(...IAsyncEnumerable)`, I've moved the bulk of its code into three private functions:
1. `MultipleRowsCopy1Prep` - runs first
2. `MultipleRowsCopy1Add` - runs for each item in `source`
3. `MultipleRowsCopy1Finish` - runs before calling `helper.Execute`

Ditto for `MultipleRowsCopy2` and `MultipleRowsCopy3`.

Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs
```diff
@@ -19,9 +20,71 @@ public SqlServerBulkCopy(SqlServerDataProvider provider)
...
+					options,
+					(columns) => new BulkCopyReader<T>(connections.Item1, columns, source),
+					true);
+			}
+
+			return MultipleRowsCopyAsync(table, options, source);
+		}
+#endif
+
+		private Tuple<DataConnection, IDbConnection, IDbTransaction?>? GetProviderConnection<T>(ITable<T> table)
```
Shane32, line 87: `ValueTuple` isn't supported by .Net 4.5 / 4.6 without adding a nuget package.  I could create a struct to hold these variables, but in this case I just used a `Tuple`.  Unfortunately you can't name the members with a `Tuple`.

Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs
```diff
@@ -33,10 +96,25 @@ public SqlServerBulkCopy(SqlServerDataProvider provider)
...
+
+		private async Task<BulkCopyRowsCopied> ProviderSpecificCopyInternal<T>(
+			DataConnection                                          dataConnection,
+			IDbConnection                                           connection,
+			IDbTransaction?                                         transaction,
+			ITable<T>                                               table,
+			BulkCopyOptions	                                        options,
+			Func<List<Mapping.ColumnDescriptor>, BulkCopyReader<T>> createDataReader,
+			bool                                                    runAsync)
+		{
```
Shane32, line 113: Changes were made here to allow the same function to process sync/async requests, again to deduplicate code and at the cost of a `Task<T>` allocation for the return value.  `source` was replaced by `createDataReader` since the `BulkCopyReader<T>` can service both `IEnumerable<T>` and `IAsyncEnumerable<T>` sources.

Shane32:
SQL Server is working, and most of the providers that just map to sql insert statements.  Need to finish support for a few other providers.  Need to add extension methods.  Need to review everything again.  Need to reformat 2 code blocks.

Shane32:
Need to add tests also.

Shane32:
Honestly had no idea how much code this would be to add!!!  But nearly done now.

Shane32:
@MaceWindu The `SqlServerTools` and similar classes often contain a bulk copy method which always maps directly to the extension method of the data connection (regardless of if it's an actual SqlServer data connection).  The names and parameters of the bulk copy method vary from provider to provider.  I would like to suggest we delete these methods, requiring the consumer to either use the extension method provided by the `DataConnection`, or use the bulk copy method on the `IDataProvider` interface of the provider.  Or if the intent is to provide a bulk copy method with parameters specific to the provider, then I would suggest we provide this method for only providers with provider-specific bulk copy methods, and name the methods all alike.  Please let me know what you think.

MaceWindu:
I also don't like those methods as they just duplicate existing functionality. @sdanyliv  what do you think about removing them?

sdanyliv:
I’m not using them. Usually i call DataConnection method extensions.
@igor-tkachev, what do you think? Do you use them directly?

MaceWindu:
As an option we can just mark them obsolete and don't touch them (don't add async versions)

viceroypenguin:
`[Obsolete]` also protects binary compatibility. Could probably delete them finally in 4.0 eventually.

MaceWindu:
Update on npgsql bulkcopy:
- WriteAsync method should be optional, as it was added in npgsql 4.1.0+
- also starting from 4.1.0+ there is `public ValueTask DisposeAsync()` method (for all targets)

Shane32:
Thank you @MaceWindu - good to know.  That’s the only thing left to implement - ngpsql async support - but it might be helpful to run a full set of tests to assess if everything else is working well so far. 

I do have a plan for ngpsql implementation, but I will probably have to write that particular bulk loading code from scratch. I installed postgre so I can properly test it locally.

Running tests now against postgre and SQL Server on my machine, there’s a single failing test due to an apparent failure to properly account for daylight saving time. I just want to be sure that’s not a problem that I introduced.

Thanks!

Shane32:
Todo: fix MySQL WriteToServerAsync2 (name won't match)

Shane32:
@MaceWindu @sdanyliv Many of the database providers have provider-specific bulk upload methods, but MultipleRows method is set as default.  Should this be changed for any of the providers?
```
Provider          Has provider specific       Async?      Current default
-----------------------------------------------------------------------------
Access                    No                              MultipleRows
DB2                       Yes                  No         MultipleRows
Firebird                  No                              MultipleRows
Informix                  Yes                  No         ProviderSpecific
MySQL                     Yes                  Yes        MultipleRows
Oracle                    Yes                  No         MultipleRows
PostgreSQL                Yes                  Yes        MultipleRows
SapHana                   Yes                  Yes        MultipleRows
SqlCe                     No                              MultipleRows
SQLite                    No                              MultipleRows
SqlServer                 Yes                  Yes        ProviderSpecific
Sybase                    Yes                  No         MultipleRows
```

Source/LinqToDB/DataProvider/MySql/MySqlProviderAdapter.cs
```diff
@@ -428,6 +430,14 @@ internal class MySqlBulkCopy : TypeWrapper
 					PropertySetter((MySqlBulkCopy this_) => this_.BulkCopyTimeout),
 					// [7]: set DestinationTableName
 					PropertySetter((MySqlBulkCopy this_) => this_.DestinationTableName),
+					// [8]: WriteToServerAsync
+					new Tuple<LambdaExpression, bool>
+					((Expression<Func<MySqlBulkCopy, IDataReader, CancellationToken, Task>>     )((MySqlBulkCopy this_, IDataReader dataReader, CancellationToken cancellationToken) => this_.WriteToServerAsync (dataReader, cancellationToken)), true),
+#if !NET45 && !NET46
+					// [9]: WriteToServerAsync
+					new Tuple<LambdaExpression, bool>
+					((Expression<Func<MySqlBulkCopy, IDataReader, CancellationToken, ValueTask>>)((MySqlBulkCopy this_, IDataReader dataReader, CancellationToken cancellationToken) => this_.WriteToServerAsync2(dataReader, cancellationToken)), true),
```
Shane32, line 439: This doesn't work because the name `WriteToServerAsync2` doesn't match `WriteToServerAsync`.

Shane32:
@MaceWindu This code is ready for you to take a look at the best way to handle the two different signatures of `WriteToServerAsync` for MySQL.  I can probably patch it in the same way I handled npgsql's generic `WriteAsync<T>` method, if you like.  Otherwise, you can add code to this branch and/or create a PR referencing this branch.  Without further changes, the MySql providers that have a `WriteToServerAsync` function that returns a `ValueTask` will fallback to synchronous execution.  On my machine, this did not occur, as the `MySqlConnector` provider's `WriteToServerAsync` function returned a `Task` and successfully executed asynchronously in both .Net 4.6 and .Net Core.  The `MySql` provider did not seem to provide a BulkCopy class at all, so all operations with that provider downgraded to `MultipleRowsCopy` method.

The tests currently fail now because they fail on the master branch, due to #2329 breaking `WhereBooleanTest2`.  Once the master branch is fixed, we can re-run the tests, and everything should pass.  I've installed MySql 8, PostgreSQL 12, and various flavors of SQL Server locally, which all work fine with the new code.  For those providers, I've verified that bulk uploading properly uses the new asynchronous calls to the provider's bulk loader.

In order to perform bulk upload tests, I've changed the two tests within `Tests.xUpdate.BulkCopyTests` to test both synchronous and asynchronous operation for each combination of providers and options via a new `asyncMode` parameter.  The parameter is as follows:
* 0 - execute synchronously
* 1 - execute asynchronously with IEnumerable
* 2 - execute asynchronously with IAsyncEnumerable (.Net Core only)

Because one of the other parameters is `BulkCopyType`, the `ProviderSpecific` bulk copy mode is tested synchronously and asynchronously for every provider.  Although there were some bulk copy tests scattered elsewhere, often alongside other provider-specific tests, I only added one other test (for SQL Server).  I can duplicate the other tests for async testing if you like.

Note that many of the providers do not have `ProviderSpecific` set as the default bulk upload method -- see chart posted here: https://github.com/linq2db/linq2db/pull/2314#issuecomment-655626678 .  Let me know if I should change any of the defaults for the 3.0.0 release.

There are a few blocks of code which need to be realigned.  I will do so after you've reviewed the code, or else github will flag the code as having changed.

Please let me know if there are any other changes to be made.  Thanks!!

MaceWindu:
@Shane32 regarding default method type - it existed before I joined project, so I tried to not change values for it.

Actually I don't like this setting at all as it doesn't make much sense to me and I don't see logic behind it. My opinion - let it be as it is now.

Shane32:
Thanks! It's ready to be merged then.

MaceWindu:
Not yet, I will push my changes to fix support mysql and also fix npgsql, as you didn't exposed async support for netfx target, which is a bit complicated too due to use of ValueTypes

Shane32:
I’ll be interested to see what you come up with!

MaceWindu:
My changes:
- added more async tests
- added async BC support for NETFX targets and fixed mysqlconnector async BC support (by adding more functionality to TypeMapper)
- for async provider-specific BC we will log it as `INSERT ASYNC BULK` (so we can assert it in tests and users could see if async version used)
- removed record writer cache for sync postgresql bulk copy
- [TypeMapper] moved type mapper types to separate folder
- [TypeMapper]: verify return type of method when mapping methods (needed for MySql to not pick wrong method)
- [TypeMapper]: resolve mapped method name using `TypeWrapperNameAttribute` (allows to map method to method with other name, needed for MySql)
- [TypeMapper]: add `CustomMapperAttribute` and `ICustomMapper` interface to define custom converters (for method return type for now), added `ValueTaskToTaskMapper` custom mapper to map npgsql methods with `ValueTask`\`ValueTask<T>` results

MaceWindu:
I broke sqlserver/oracle provider mappings. Will fix asap

Shane32:
Nice!  Everything looks great to me.  Just as a note, there's two functions which need to be "tabbed over":
* `SqlServerBulkCopy.ProviderSpecificCopyInternal`
* `SapHanaBulkCopy.ProviderSpecificCopyInternal`

Source/LinqToDB/Common/EnumerableHelper.cs
```diff
@@ -1,10 +1,32 @@
...
+using System.Threading.Tasks;
 
 namespace LinqToDB.Common
 {
 	internal class EnumerableHelper
 	{
+#if !NET45 && !NET46
+		public static IEnumerable<T> AsyncToSyncEnumerable<T>(IAsyncEnumerator<T> enumerator)
+		{
+			while (enumerator.MoveNextAsync().Result)
```
viceroypenguin, line 13: Are we guaranteed that any `IAsyncEnumerator<>` from L2DB that would be provided to this method won't block (via what we provide, etc.)? 

Also, I believe the proper way to get the result is `.GetAwaiter().GetResult()`, which will throw any exceptions properly instead of an `AggregateException`.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -21,16 +24,70 @@ public virtual BulkCopyRowsCopied BulkCopy<T>(BulkCopyType bulkCopyType, ITable<
 			}
 		}
 
-		protected virtual BulkCopyRowsCopied ProviderSpecificCopy<T>(ITable<T> table, BulkCopyOptions options, IEnumerable<T> source)
+		public virtual Task<BulkCopyRowsCopied> BulkCopyAsync<T>(
+			BulkCopyType bulkCopyType, ITable<T> table, BulkCopyOptions options, IEnumerable<T> source, CancellationToken cancellationToken)
+		{
+			switch (bulkCopyType)
```
viceroypenguin, line 30: Recommend new switch notation:
```
return bulkCopyType switch
{
  BulkCopyType.MultipleRows => MultipleRowsCopyAsync(...),
  BulkCopyType.RowByRow     => RowByRowCopyAsync(...),
  _                         => ProviderSpecificCoypAsync(...),
};
```

@MaceWindu opinion?

Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs
```diff
@@ -22,6 +24,67 @@ public SqlServerBulkCopy(SqlServerDataProvider provider)
...
+			var connections = TryGetProviderConnections(table);
+			if (connections.HasValue)
+			{
+				return ProviderSpecificCopyInternal(
+					connections.Value,
+					table,
+					options,
+					(columns) => new BulkCopyReader<T>(connections.Value.DataConnection, columns, source),
+					false,
+					default).Result;
```
viceroypenguin, line 37: `.GetAwaiter().GetResult()`?

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
MaceWindu: I don't mind, but I think we should start adding analyzers to enforce stuff like that instead of having mix of various styles

viceroypenguin: I know of a couple of Roslyn analyzers that will support doing so. I'll do some research next week if you want.

viceroypenguin: Actually, some of this stuff can be enforced just by updating the `.editorconfig` file. I'll do some touchups on that file for sure first.

MaceWindu: yep, editorconfig also needs improvement from current basic state :)

Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs
Shane32: Good catch.  Didn't realize `Result` threw an `AggregateException`

Source/LinqToDB/Common/EnumerableHelper.cs
Shane32: Done, and also in `BulkCopyReader.MoveNext`.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
Shane32: I updated the switch to be a switch expression here in `BasicBulkCopy`

Source/LinqToDB/DataProvider/SqlServer/SqlServerBulkCopy.cs
Shane32: Done (although since the `runAsync` parameter is `false`, async calls are never made and exceptions should bubble up properly anyway).  Also changed `MySqlBulkCopy` and `SapHanaBulkCopy` in the same manner.

Shane32:
Let me know if there's anything else you'd like me to do on this PR, @MaceWindu .

Source/LinqToDB/Common/EnumerableHelper.cs
```diff
@@ -1,10 +1,32 @@
...
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
+		{
+			foreach (var item in enumerable)
+			{
+				yield return item;
+			}
+		}
+#endif
+
 		/// <summary>
```
igor-tkachev, line 30: Should we make these methods extension methods? Also I would rename it to AsSyncEnumerable and AsAsyncEnumerable.

Shane32: Makes sense to me.  Note that they are in an `internal` class, and the existing `Batch` methods in this class are not extension methods.

viceroypenguin: Extension methods would require moving them to a new top-level static class as well

MaceWindu:
Probably it's time to merge it.

Issue #2313, opened 2020-07-03
T4 model converts unsigned tinyint(3) in a sbyte


azaroth75:
T4 model converts unsigned tinyint(3) in a sbyte (using .NET5.0)

### Steps to reproduce

database table
```sql
CREATE TABLE `inverterLogs` (
	`tStamp` DATETIME NOT NULL DEFAULT current_timestamp(),
	`idInverter` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0' COMMENT '',
	`data` TINYBLOB NOT NULL DEFAULT '0',
	`operatingMode` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0',
	`faultCodeH` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0',
	`faultCodeL` TINYINT(3) UNSIGNED NOT NULL DEFAULT '0'
)
COMMENT=''
COLLATE='utf8_general_ci'
ENGINE=InnoDB
;
```

conversion result
```c#
[Table("inverterLogs")]
public partial class InverterLog
{
	[Column("tStamp"),        NotNull] public DateTime TStamp        { get; set; } // datetime
	/// <summary>
	/// id inverter di riferimento
	/// </summary>
	[Column("idInverter"),    NotNull] public sbyte    IdInverter    { get; set; } // tinyint(3) unsigned
	[Column("data"),          NotNull] public byte[]   Data          { get; set; } // tinyblob
	[Column("operatingMode"), NotNull] public sbyte    OperatingMode { get; set; } // tinyint(3) unsigned
	[Column("faultCodeH"),    NotNull] public sbyte    FaultCodeH    { get; set; } // tinyint(3) unsigned
	[Column("faultCodeL"),    NotNull] public sbyte    FaultCodeL    { get; set; } // tinyint(3) unsigned
}
```

### Environment details
linq2db version: v3.0.0-rc.1
Database Server: 10.3.21-MariaDB
Database Provider: MySqlConnector  v3.0.0-rc.1
Operating system: w10 1903
.NET Framework: .NET5.0

MaceWindu:
What's wrong with sbyte?

azaroth75:
it's signed

MaceWindu:
Hmm, probably it shouldn't

azaroth75:
thank you!

azaroth75:
I'm sorry to bother you again about this. 
[...]

PR #2316, opened 2020-07-03
Add unsigned tinyint support for MySQL

viceroypenguin:
Fix for #2313.


MaceWindu:
Thanks, @viceroypenguin, but this is a bit more complicated. I will submit bigger fix today

PR #2318, opened 2020-07-04
Introduce query flags into query cache

### LINKED ISSUE #2306: Globally disabled group guard is not respected for specific case
(related to #2257)

Globally disabled group guard (`Configuration.Linq.GuardGrouping = false;`) is not respected (_throws generic exception_) in cases like the following:

```csharp
query.GroupBy(x => x.UserId).ToDictionary(x => x.Key, x => x.Select(y => y.SubstitutedUserId).ToList());
```

```
Exception message:
LinqToDB.LinqToDBException : You should explicitly specify selected fields for server-side GroupBy() call or add AsEnumerable() call before GroupBy() to perform client-side grouping.
    Set Configuration.Linq.GuardGrouping = false to disable this check.
    Additionally this guard exception can be disabled by extension GroupBy(...).DisableGuard().
    NOTE! By disabling this guard you accept additional Database Connection(s) to the server for processing such requests.
```

### Environment details
linq2db version: v3.0.0-rc.1
Database Server: Sql Server 2017
Database Provider: SqlClient
Operating system: Windows 10 1909
.NET Framework: 4.8

MaceWindu:
Fix #2306

### DIFF
--- a/Source/LinqToDB/Linq/Query.cs
+++ b/Source/LinqToDB/Linq/Query.cs
@@ -261,8 +261,41 @@ namespace LinqToDB.Linq
 		#endregion
 
 		#region Query cache
+		[Flags]
+		enum QueryFlags
+		{
+			None             = 0,
+			/// <summary>
+			/// Bit set, when group by guard set for connection.
+			/// </summary>
+			GroupByGuard     = 0x1,
+			/// <summary>
+			/// Bit set, when inline parameters enabled for connection.
+			/// </summary>
+			InlineParameters = 0x2,
+		}
+
 		class QueryCache
 		{
+			class QueryCacheEntry
+			{
+				public QueryCacheEntry(Query<T> query, QueryFlags flags)
+				{
+					// query doesn't have GetHashCode now, so we cannot precalculate hashcode to speed-up search
+					Query = query;
+					Flags = flags;
+				}
+
+				public Query<T>   Query { get; }
+				public QueryFlags Flags { get; }
+
+				// accepts components to avoid QueryCacheEntry allocation for cached query
+				public bool Compare(IDataContext context, Expression queryExpression, QueryFlags flags)
+				{
+					return Flags == flags && Query.Compare(context, queryExpression);
+				}
+			}
+
 			// lock for cache instance modification
 			readonly object _syncCache    = new object();
 			// lock for query priority modification
@@ -270,7 +303,7 @@ namespace LinqToDB.Linq
 
 			// stores all cached queries
 			// when query added or removed from cache, query and priority arrays recreated
-			Query<T>[] _cache   = Array<Query<T>>.Empty;
+			QueryCacheEntry[] _cache   = Array<QueryCacheEntry>.Empty;
 
 			// stores ordered list of query indexes for Find operation
 			int[]      _indexes = Array<int     >.Empty;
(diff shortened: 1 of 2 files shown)


MaceWindu:
Looks like we need to add guard grouping global flag to query cache key, otherwise changing flag doesn't affect cached queries

MaceWindu:
or invalidate cache on flag change

viceroypenguin:
I'd vote for adding the flag to the query cache key. Then if the flag is left as false globally, but enabled on a per query basis, everything still operates as expected?

MaceWindu:
Yep, discussed it with @sdanyliv  and decided to add flags to cache, as it will also later will allow us to move flags to data connection level instead of having them as global settings.
per-query flag change (disabling) using `DisableGuard` is not affected by global setting change, as it is a part of query and already stored in cached query.

MaceWindu:
added initial support for query flags to query cache. Unfortunatelly this is not full fix (still there are changes to have race conditions in multithreaded environment for GroupBy guard setting).
In future we should move global flags to data connection level and invalidate cache on global settings (defaults) change.

@detoxhby, while this fix probably will help with your situation,  I would recommend to invalidate caches in guard setting change (or don't change it after application setup) while we don't have connection-level setting

Source/LinqToDB/Linq/Query.cs
```diff
@@ -335,17 +368,17 @@ public void TryAdd(IDataContext dataContext, Query<T> query)
...
-							if (cache[i].Compare(dataContext, query.Expression!))
+							if (cache[i].Compare(dataContext, query.Expression!, flags))
 								// already added by another thread
 								return;
 					}
 
 					// create new cache instance and reorder items according to priorities to inprove Find without
 					// reorder lock
-					var newCache      = new Query<T>[cache.Length == CacheSize ? CacheSize : cache.Length + 1];
+					var newCache      = new QueryCacheEntry[cache.Length == CacheSize ? CacheSize : cache.Length + 1];
```
viceroypenguin, line 378: Can we save an allocation here by not initializing a new array if `cache.Length == CacheSize`?

Source/LinqToDB/Linq/Query.cs
```diff
@@ -438,14 +471,24 @@ public static Query<T> GetQuery(IDataContext dataContext, ref Expression expr)
...
 
-			var query = _queryCache.Find(dataContext, expr);
+			// calculate query flags
+			var flags = QueryFlags.None;
+			if (dataContext.InlineParameters)
+				flags |= QueryFlags.InlineParameters;
+			// TODO: here we have race condition due to flag being global setting
+			// to fix it we must move flags to context level and remove global flags or invalidate caches on
+			// global flag change
+			if (Configuration.Linq.GuardGrouping)
```
viceroypenguin, line 481: Technically, even invalidating caches on a change will not fix the race condition. It only removes cache entries that have potentially been created under the race condition. The only proper fix is going to be adding it to the context level.

viceroypenguin:
PS: Approved because neither comment needs to be addressed before 3.0

Source/LinqToDB/Linq/Query.cs
MaceWindu: Agree

Source/LinqToDB/Linq/Query.cs
MaceWindu: This allocation happens not because we want to increase cache, but because we want to create new cache version (instance).

viceroypenguin: Sorry, didn't pay attn to the race condition if you don't allocate. All good.

PR #2323, opened 2020-07-06
Update SAP HANA tests to use latest provider

MaceWindu:
Nothing interesting:
- removed hack needed for provider to load (SAP fixed it)
- disabled failing test as they break binary parameters

Looks like SAP guys still stick to their strategy as both issues are too trivial:
1. Unit-testing is not for "enterprise" development
2. If we don't have bugs, how can we sell support?


MaceWindu:
Also moving access oledb testing to netcore 3.1 from 2.1 due to a lot of AV

PR #2329, opened 2020-07-08
Fix RC2 regression with EXISTS function being wrapped in CASE/IIF

MaceWindu:
With new IsTrue predicate we were skipping optimizations for ExprExpr simple cases.


Shane32:
This PR breaks the WhereBooleanTest2 test for most providers

MaceWindu:
@Shane32 it was an error in azure pipelines config

MaceWindu:
Will provide another way to fix it

MaceWindu:
new fix #2332

PR #2337, opened 2020-07-09
override ToString for ILoadWithQueryable to return base query sql

### LINKED ISSUE #2336: SQL-Text of ILoadWithQueryable
Hi,

is it anyhow possible to get the generated SQL string from the new ILoadWithQueryable like it is done with ITable?

MaceWindu:
Fix #2336


MaceWindu:
Note that it will return sql for main query only. Additional queries will not be returned

PR #2347, opened 2020-07-15
Various fixes for parameters.

### LINKED ISSUE #2346: LinqToDb produces a query that wraps Int64 DateTime.Ticks value in SQLite DateTime() function?
Hello, I've changed my code recently to store any DateTime values as Int64 values by setting up my DataConnection like this:

```
            var connection = new DataConnection(ProviderName.SQLiteMS, connectionString)
            {
                InlineParameters = true
            };

            connection.MappingSchema.SetDataType(typeof(DateTime), DataType.Int64);
            connection.MappingSchema.SetConvertExpression(
                (long ticks) => new DateTime(ticks, DateTimeKind.Unspecified)
            );
            connection.MappingSchema.SetConvertExpression(
                (DateTime dt) => dt.Ticks
            );
            connection.MappingSchema.SetValueToSqlConverter(
                typeof(DateTime),
                (sb, dType, v) => sb.Append(((DateTime) v).Ticks.ToString(CultureInfo.InvariantCulture))
            );
```

and that works fine for me so long as I don't apply a WHERE to any queries I do. When I do apply a WHERE, this is an example of a query produced:

```
SELECT
        [x].[DateTime],
        [x].[Value]
FROM
        [fae5697f79034df99e0ad32292446f5a_testOut] [x]
WHERE
        DateTime([x].[DateTime]) > DateTime(637304378093620000) AND
        DateTime([x].[DateTime]) < DateTime(637304382196380000)
```

The WHERE condition is created using Linq.Expressions (small but hopefully sufficient cut of code):

```
var left = Expression.Property(x, p);
var right = Expression.Constant(dt, typeof(DateTime));
var e = Expression.GreaterThan(left, right);
var whereCondition = Expression.Lambda<Func<T, bool>>(e, x);

...
queryable = queryable.Where(whereCondition);
...
```

Sorry for not providing more code or any mapping classes. All the mapping classes I'm using are generated dynamically at run time with IL.Emit so I don't have any more code to give that wouldn't make things more confusing. If necessary I can try making a solution that reproduces the problem, but I'm hoping that I'm just missing something like an extra setting on the DataConnection that I wasn't able to find and you can see the problem easily.

linq2db v2.9.8
linq2db.SQLite.MS
Ubuntu 18.04
.NET Core 3.1

### LINKED ISSUE #2403: Spaces result in invalid SQL
Spaces in properties/fields result in invalid SQL:

```f#

open System    
open LinqToDB
open LinqToDB.Mapping
open LinqToDB.Data

[<Table(Name = "records")>]
type Record =
    {
      [<Column(Name = "id"); PrimaryKey; Identity>]
      Id : int 
    }
let ``some int`` = 2
let q = 
    use db = new DataConnection(DataProvider.MySql.MySqlDataProvider(""),connectionString)
    let table : Record ITable = db.GetTable()
    query {
        for i in table do 
        where(i.Id = ``some int``)
        select(i)
    } |> string

q
```

Resulting SQL (value of q):
```sql
--   MySql
DECLARE @some int Int32
SET     @some int = 2

SELECT
	`i`.`id`
FROM
	`records` `i`
WHERE
	`i`.`id` = @some int
```

sdanyliv:
Fixes #2346
Fixes #2403

Tasks TODO:

- [x] Fixed parameter inlining.
- [x] Removed unnecessary DateTime conversion.
- [x] Introduce `InlineParameters()` queryable extension.
- [x] Introduce global setting for Take/Skip parametrisation.


Tests/Base/TestBase.cs
```diff
@@ -1192,6 +1192,21 @@ public void Dispose()
 		}
 	}
 
+	public class ParametrizeTakeSkip : IDisposable
```
viceroypenguin, line 1195: spelling: `ParameterizeTakeSkip`

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -191,6 +191,11 @@ public static class Linq
 			/// Into GROUP BY equivalent if syntax is not supported
 			/// </summary>
 			public static bool KeepDistinctOrdered = true;
+
+			/// <summary>
+			/// Enables Take/Skip parametrization.
+			/// </summary>
+			public static bool ParametrizeTakeSkip = true;
```
viceroypenguin, line 198: spelling: `ParameterizeTakeSkip`

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -2272,7 +2272,7 @@ ISqlExpression GetParameter(Expression ex, MemberInfo member)
 			if (member is MethodInfo mi)
 				member = mi.GetPropertyInfo();
 
-			var vte  = ReplaceParameter(_expressionAccessors, ex, _ => { });
+			var vte  = ReplaceParameter(_expressionAccessors, ex, false, _ => { });
```
viceroypenguin, line 2275: clarity: can we do `forceConstant: false` for readability?

Source/LinqToDB/Common/Configuration.cs
MaceWindu: found another 3 places where we use `parametrize` "word" (all in comments). makes sense to fix them too.

PR #2352, opened 2020-07-17
CommandBehavior.SchemaOnly causes problems while generating SQL Server SP

### LINKED ISSUE #2348: CommandBehavior.SchemaOnly causes problems while generating SQL Server SP
linq4db missed some of my SQL Server procedures during t4 generation. Debugging LoadServerMetadata(dataConnection) shown that these procedures have ResultException. Exceptions were caused by using [CommandBehavior.SchemaOnly](https://github.com/linq2db/linq2db/issues/792#issuecomment-321282351), which internally uses [SET FMTONLY](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-fmtonly-transact-sql?view=sql-server-ver15) which is deprecated since SQL SERVER 2012.
Strange thing is that MS's System.Data still uses FMTONLY and there are no plans to for using sp_describe_first_result_set instead.

I actually do not care about evil FMTONLY as I do not use any procedures that return result sets as this an evil as well (IMHO).
Still it causes problems. Sample SQL code:

```
drop procedure if exists test 
go

create procedure test (@src NVARCHAR(MAX)) as begin
declare @value NVARCHAR(MAX)
set @value = JSON_QUERY(@src,'$.value')
select @value as value --even if comment it out FMTONLY will throw
end
go

--this throws Msg 13609, Level 16, State 2, Procedure test, Line 4 [Batch Start Line 8] JSON text is not properly formatted. Unexpected character '.' is found at position 0.
set FMTONLY OFF; SET FMTONLY ON;
exec test ''
SET FMTONLY OFF; 
go


--this produces correct data
--is_hidden column_ordinal name                                                                                                                             is_nullable system_type_id system_type_name                                                                                                                 max_length precision scale collation_name                                                                                                                   user_type_id user_type_database                                                                                                               user_type_schema                                                                                                                 user_type_name                                                                                                                   assembly_qualified_type_name                                                                                                                                                                                                                                                                                 xml_collection_id xml_collection_database                                                                                                          xml_collection_schema                                                                                                            xml_collection_name                                                                                                              is_xml_document is_case_sensitive is_fixed_length_clr_type source_server                                                                                                                    source_database                                                                                                                  source_schema                                                                                                                    source_table                                                                                                                     source_column                                                                                                                    is_identity_column is_part_of_unique_key is_updateable is_computed_column is_sparse_column_set ordinal_in_order_by_list order_by_is_descending order_by_list_length tds_type_id tds_length  tds_collation_id tds_collation_sort_id
----------- -------------- -------------------------------------------------------------------------------------------------------------------------------- ----------- -------------- -------------------------------------------------------------------------------------------------------------------------------- ---------- --------- ----- -------------------------------------------------------------------------------------------------------------------------------- ------------ -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ ----------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- --------------- ----------------- ------------------------ -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- ------------------ --------------------- ------------- ------------------ -------------------- ------------------------ ---------------------- -------------------- ----------- ----------- ---------------- ---------------------
--0         1              value                                                                                                                            1           231            nvarchar(max)                                                                                                                    -1         0         0     SQL_Latin1_General_CP1_CI_AS                                                                                                     NULL         NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                                                                                                                                                                                                         NULL              NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                             0               0                 0                        NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                             NULL                                                                                                                             0                  NULL                  0             1                  0                    NULL                     NULL                   NULL                 231         65535       13632521         52
exec sp_describe_first_result_set  @tsql =  N' exec test '''' '
```

Is there a possibility to change CommandBehavior.SchemaOnly to sp_describe_first_result_set in linq2db?

Is there an option to skip calling CommandBehavior.SchemaOnly for SPs?

Is there an option to ignore errors and generate all SPs as having no result sets?

igor-tkachev:
Fixes #2348

### DIFF
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -194,6 +194,19 @@ namespace LinqToDB.Common
 		}
 
 		/// <summary>
+		/// SqlServer specific global settings.
+		/// </summary>
+		[PublicAPI]
+		public static class SqlServer
+		{
+			/// <summary>
+			/// if set to true SchemaProvider uses <see cref="System.Data.CommandBehavior.SchemaOnly"/> to get metadata.
+			/// Otherwise the sp_describe_first_result_set sproc is used instead.
+			/// </summary>
+			public static bool UseSchemaOnlyToGetSchema = false;
+		}
+
+		/// <summary>
 		/// Linq over WCF global settings.
 		/// </summary>
 		[PublicAPI]
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -200,8 +200,8 @@ namespace LinqToDB.Common
 		public static class SqlServer
 		{
 			/// <summary>
-			/// if set to true SchemaProvider uses <see cref="System.Data.CommandBehavior.SchemaOnly"/> to get metadata.
-			/// Otherwise the sp_describe_first_result_set sproc is used instead.
+			/// if set to true, SchemaProvider uses <see cref="System.Data.CommandBehavior.SchemaOnly"/> to get metadata.
+			/// Otherwise the sp_describe_first_result_set sproc is used.
 			/// </summary>
 			public static bool UseSchemaOnlyToGetSchema = false;
 		}
--- a/Source/LinqToDB/Data/DataConnectionExtensions.cs
+++ b/Source/LinqToDB/Data/DataConnectionExtensions.cs
@@ -375,7 +375,6 @@ namespace LinqToDB.Data
 
 		#endregion
 
-		
 		#region Query
 
 		/// <summary>
@@ -1163,6 +1162,7 @@ namespace LinqToDB.Data
 		#endregion
 
 		#region Query with template
+
 		/// <summary>
 		/// Executes command and returns results as collection of values of specified type.
 		/// </summary>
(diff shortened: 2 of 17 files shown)


Source/LinqToDB/Common/Configuration.cs
```diff
@@ -193,6 +193,19 @@ public static class Linq
...
+		/// SqlServer specific global settings.
+		/// </summary>
+		[PublicAPI]
+		public static class SqlServer
+		{
+			/// <summary>
+			/// if set to true, SchemaProvider uses <see cref="System.Data.CommandBehavior.SchemaOnly"/> to get metadata.
+			/// Otherwise the sp_describe_first_result_set sproc is used.
+			/// </summary>
+			public static bool UseSchemaOnlyToGetSchema = false;
```
sdanyliv, line 206: I think it should be part of `GetSchemaOptions` class

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -1200,6 +1200,43 @@ public static IEnumerable<T> Query<T>(this DataConnection connection, T template
...
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="template">This value used only for <typeparamref name="T"/> parameter type inference, which makes this method usable with anonymous types.</param>
+		/// <param name="sql">Command text.</param>
+		/// <param name="parameters">Command parameters.</param>
+		/// <returns>Returns collection of query result records.</returns>
+		public static IEnumerable<T> QueryProc<T>(this DataConnection connection, T template, string sql, params DataParameter[] parameters)
+		{
+			return new CommandInfo(connection, sql, parameters).QueryProc(template);
+		}
```
sdanyliv, line 1215: I think procedure needs overload for `IAsyncEnumerable`.

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -1200,6 +1200,43 @@ public static IEnumerable<T> Query<T>(this DataConnection connection, T template
...
+		/// <para>Last case will convert all mapped columns to <see cref="DataParameter"/> instances using following logic:</para>
+		/// <para> - if column is of <see cref="DataParameter"/> type, column value will be used. If parameter name (<see cref="DataParameter.Name"/>) is not set, column name will be used;</para>
+		/// <para> - if converter from column type to <see cref="DataParameter"/> is defined in mapping schema, it will be used to create parameter with colum name passed to converter;</para>
+		/// <para> - otherwise column value will be converted to <see cref="DataParameter"/> using column name as parameter name and column value will be converted to parameter value using conversion, defined by mapping schema.</para>
+		/// </param>
+		/// <returns>Returns collection of query result records.</returns>
+		public static IEnumerable<T> QueryProc<T>(this DataConnection connection, T template, string sql, object? parameters)
+		{
+			return new CommandInfo(connection, sql, parameters).QueryProc(template);
+		}
```
sdanyliv, line 1238: I think procedure needs overload for `IAsyncEnumerable`.

Source/LinqToDB/Common/Configuration.cs
igor-tkachev: What about SqlServer specific features?

sdanyliv: Really don't know, but currently is see that it is like BulkCopy options, some of them are provider specific.

Source/LinqToDB/Data/DataConnectionExtensions.cs
igor-tkachev: I agree. We do not have async methods for many cases. It should be separate task.

Source/LinqToDB/Common/Configuration.cs
igor-tkachev: BulkCopy options can vary from query to query. UseSchemaOnlyToGetSchema is a global setting.

sdanyliv: It is vary in LINQPad, also it will be more convinient in T4 template to correct only GetSchemaOption instead of setting global option.

Source/LinqToDB/Data/DataConnectionExtensions.cs
viceroypenguin: Created issue #2386 to track

PR #2357, opened 2020-07-21
Fix for issue #2356

### LINKED ISSUE #2356: Setting of DateTimeKind does not work
We have this code in our backend:

     mappingSchema.SetConvertExpression<DateTime, DateTime>(dt => DateTime.SpecifyKind(dt, DateTimeKind.Utc));

but it does not seem to work.

Don't know if it happend with update to v3 or before. Using Postgres Database

jogibear9988:
Fixes #2356


sdanyliv:
Test i not relevant to the issue. It is not Postgres.

jogibear9988:
Issue is also present in Sqlserver& Sqlite

sdanyliv:
Which column type do you have in Postgres? It is important!

jogibear9988:
I know that there are some types wich have no timezoneinfo.
Because of that we introduced that converter.

sdanyliv:
Let's check for regressions.

jogibear9988:
is it possible to create a hotfix release with this fix?

jogibear9988:
@sdanyliv & @MaceWindu thanks for finding the issue

Issue #1166, opened 2018-05-30
Implement Merge into updatable query or CTE


Corey-M:
The current merge API accepts only `ITable<TTarget>` as a merge target.  In Microsoft SQL Server at least it is valid to have any updatable query or CTE as the target of a merge statement.

Specifically this is useful when updating certain types of history tables or slowly-changing dimensions (Type 2 specifically) in data warehouse environments.  In the case of Type2 SCDs the merge target is the subset of rows that are currently active.  All of the terms of the merge act on the selected subset of the data rows.

For example, a partial update of a Type2 SCD:

	MERGE INTO (SELECT * FROM myTargetTable WHERE IsCurrent = 1) t
	USING mySourceTable s
	ON s.ID = t.ID

	-- Close changed records
	WHEN MATCHED AND s.SomeIntField <> t.SomeIntField THEN
		UPDATE SET EndTime = GETDATE(), IsCurrent = 0

	-- Close deleted records
	WHEN NOT MATCHED BY SOURCE THEN
		UPDATE SET EndTime = GETDATE(), IsCurrent = 0

	-- Append new records
	WHEN NOT MATCHED BY TARGET THEN
		INSERT (ID, StartTime, EndTime, IsCurrent, SomeIntField)
		VALUES (s.ID, GETDATE(), '31-Dec-2999', 1, s.SomeIntField)
	;

If the merge is run against all of the rows in the table it will have unintended consequences.

I acknowledge that this may be out of scope, and Data Warehouse updates should almost always be done with stored procedures, but it seems like a desirable feature to implement.

MaceWindu:
Interesting, didn't know it is possible. I would say it is not out-of-scope, but it will take some time to implement as we have other things on our radar ATM.

Corey-M:
@MaceWindu I've been doing some data warehouse stuff recently - all stored procedures of course - and was toying with the idea of doing the same sort of thing in C#.  Looking over the new Merge API it struck me as odd that the target was `ITable<T>` instead of `IQueryable<T>` when the SQL Server MERGE statement supports queries as targets.

The downside is that you can't easily validate whether the target is writable when you're building the SQL statement.  You have to trust that the coder knows what they are doing.

This isn't a high priority, just a matter of completeness.  I think in general once you get to this level of complexity it's safer and cleaner to use stored procedures.

Corey-M:
[...]

PR #2363, opened 2020-07-23
Merge supports IQueryable<T> targets

viceroypenguin:
- [x] General CTE support with Merge statements
- [x] `.AsCte()` expression supported as Merge target
- [x] Inform users to call `.AsCte()` if they attempt to use a complex query as the merge target. 
- [x] Fix LinqSerializer for `With` field
- [x] `.AsCte()` expression supported as Merge source


MaceWindu:
Which databases support CTE/Query as target?
I'm checking docs and see:
- sql sever: table/view
- db2 iSeries/zOs: table/view
- sap hana: table
- ifx: table/view
- sap ase: table/view
- oracle: table/view
- firebird: table/view

Looks like only db2 LUW declares CTE support in v11.5 and something called `fullselect`

MaceWindu:
ok, there is stuff like that for sql server https://aboutsqlserver.com/2011/10/09/sunday-t-sql-tip-merge-into-cte-as-the-target/

What I don't understand is why it is not documented here https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15

viceroypenguin:
Yeah, I've been using CTE as the target for years. Huge advantage when working w/ a detail table that needs to be MERGEd for a single parent pkid. Use the CTE to restrict to the subset for the pkid, and then deletes in the MERGE only delete for that pkid and not the entire rest of the table.

Source/LinqToDB/Linq/Builder/MergeBuilder.Merge.cs
```diff
@@ -11,20 +12,25 @@ internal class Merge : MethodCallBuilder
...
 
 			protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
 			{
 				// Merge(ITable<TTarget> target, string hint)
 				var target = builder.BuildSequence(new BuildInfo(buildInfo, methodCall.Arguments[0], new SelectQuery()) { AssociationsAsSubQueries = true });
 
-				var targetTable = ((TableBuilder.TableContext)target).SqlTable;
-
-				var merge = new SqlMergeStatement(targetTable)
+				if (!(target is TableBuilder.TableContext tableContext))
```
sdanyliv, line 24: I think you need helper function to throw this error. Also you have to check that tableContext.SelectQuery.IsSimple.

viceroypenguin:
NB: Fixes #1166

viceroypenguin:
Unit Test `MergeUsingCteWhere()` failed, it couldn't bind the predicate lambda properly. I haven't run it in a while though. I'll rebase and try again today - see if it's been fixed through the other things we've done.

viceroypenguin:
@MaceWindu can you take a look at why `MergeUsingCteWhere()` isn't working? It is essentially the same as `MergeUsingCteJoin()`, but the join is handled differently. In `MergeUsingCteJoin()`, it's using the two separate key functions, where `MergeUsingCteWhere()` uses a predicate lambda. There's something in how it's looking for the CTE `.Id` column that's throwing it off.

Source/LinqToDB/Linq/Builder/TableBuilder.CteTableContext.cs
```diff
@@ -133,6 +133,8 @@ public override SqlInfo[] ConvertToSql(Expression? expression, int level, Conver
 					ConvertToSql(null, 0, ConvertFlags.All);
 				}
 
+				expression = SequenceHelper.CorrectExpression(expression, this, queryContext);
```
MaceWindu, line 136: @sdanyliv : fix location

Source/LinqToDB/Linq/Builder/MergeBuilder.MergeInto.cs
```diff
@@ -11,20 +12,27 @@ internal class MergeInto : MethodCallBuilder
 			protected override bool CanBuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
 			{
 				return methodCall.Method.IsGenericMethod
-					&& LinqExtensions.MergeIntoMethodInfo == methodCall.Method.GetGenericMethodDefinition();
+					&& (   LinqExtensions.MergeIntoMethodInfo1 == methodCall.Method.GetGenericMethodDefinition()
+						|| LinqExtensions.MergeIntoMethodInfo2 == methodCall.Method.GetGenericMethodDefinition());
```
sdanyliv, line 16: Better to do that as in other Builders.

Source/LinqToDB/LinqExtensions.Merge.cs
```diff
@@ -17,8 +17,10 @@ public static partial class LinqExtensions
...
+		internal static readonly MethodInfo MergeIntoMethodInfo1                      = MemberHelper.MethodOf(() => MergeInto<int, int>(null!, null!, null!))                            .GetGenericMethodDefinition();
+		internal static readonly MethodInfo MergeIntoMethodInfo2                      = MemberHelper.MethodOf(() => MergeInto<int, int>(null!, (IQueryable<int>)null!))                  .GetGenericMethodDefinition();
 		internal static readonly MethodInfo UsingMethodInfo1                          = MemberHelper.MethodOf(() => Using<int, int>(null!, (IQueryable<int>)null!))                      .GetGenericMethodDefinition();
 		internal static readonly MethodInfo UsingMethodInfo2                          = MemberHelper.MethodOf(() => Using<int, int>(null!, (IEnumerable<int>)null!))                     .GetGenericMethodDefinition();
 		internal static readonly MethodInfo UsingTargetMethodInfo                     = MemberHelper.MethodOf(() => UsingTarget<int>(null!))                                             .GetGenericMethodDefinition();
```
sdanyliv, line 26: Move that to Methods.cs

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -179,15 +179,39 @@ void FinalizeCte(SqlStatement statement)
 			{
 				var foundCte  = new Dictionary<CteClause, HashSet<CteClause>>();
 
-				select.SelectQuery.Visit(foundCte, static (foundCte, e) =>
-					{
-						if (e.ElementType == QueryElementType.SqlCteTable)
+				if (select is SqlMergeStatement merge)
```
sdanyliv, line 182: Why do we need here special processing?

MaceWindu:
@viceroypenguin also would be nice to enable all merge providers for tests to check if any of them support CTE too.

MaceWindu:
Made some tests with failing providers:
- SAP HANA: supports CTE in source query (CTE should be moved inside source clause as part of source query) but linq2db doesn't support CTE for SAP HANA right now
- Oracle: also supports cte if it is a part of source query. Also oracle doesn't like `_` as table alias
- DB2: version 11.5 we use on CI declares CTE support in target so it makes sense to test it (note that syntax is differ from sql server: https://www.ibm.com/docs/en/db2/11.5?topic=statements-merge). Also it makes sense to test if source CTE as part of select will work
- informix: CTE support added in v14 we use on CI, makes sense to test is as part of source query
- firebird: CTE works in source if moved to source query

so basically, for most of providers we should move source CTE to source query when we generate SQL and test CTE in target for DB2

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
viceroypenguin: `SqlMergeStatement` inherits from `SqlStatementWithQueryBase` so that the `With` (CTE support) behavior will work properly. However, it does not use the `SelectQuery` property. It uses the `Target` and `Source` properties, which each have their own context and possible CTEs. This means it requires special support here to handle identifying CTEs in either the `Target` or `Source` branch. Possibly we could simplify to moving the target to the `SelectQuery` property, but that doesn't feel right.

Source/LinqToDB/Linq/Builder/MergeBuilder.MergeInto.cs
viceroypenguin: This code matches the other Merge builders (see Merge.InsertWhenNotMatched.cs, Merge.UpdateWhenMatched.cs, Merge.MergInto.cs, etc.).  I can change to doing it by name, but to match, I would need to do all of the merge builders similarly?

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
sdanyliv: Why you have changed inheritance for `SqlMergeStatement` ?

Source/LinqToDB/Linq/Builder/MergeBuilder.MergeInto.cs
viceroypenguin: This currently matches the other Merge builders (Merge.InsertWhenNotMatched.cs, Merge.DeleteWhenMatched.cs, etc.). Would recommend that this is the preferred form personally. 

If you want, though, I can move all of the merge builders to the string versions? Or on a separate PR I can update the other builders in general to do by method comparison rather than string.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
viceroypenguin: so that the `With` (CTE support) behavior will work properly.

Source/LinqToDB/LinqExtensions.Merge.cs
viceroypenguin: Confirming: Do you want me to move this whole region to Methods.cs?

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
sdanyliv: Yeah, under `Merge` section, something like that.

Tests/Linq/Update/MergeTests.IQueryableSource.cs
```diff
@@ -0,0 +1,194 @@
...
+				AssertRow(InitialTargetData[0], result[0], null, null);
+				AssertRow(InitialTargetData[1], result[1], null, null);
+				AssertRow(InitialTargetData[2], result[2], null, 203);
+				AssertRow(InitialTargetData[3], result[3], null, null);
+				AssertRow(InitialSourceData[2], result[4], null, null);
+				AssertRow(InitialSourceData[3], result[5], null, 216);
+			}
+		}
+
+		[Test]
```
MaceWindu, line 43: disable tests for other providers with ActiveIssue(3015...)

Tests/Linq/Update/MergeTests.IQueryableSource.cs
```diff
@@ -0,0 +1,194 @@
...
+		public void MergeFromCte([MergeDataContextSource] string context)
```
MaceWindu, line 105: disable tests for other providers with ActiveIssue(3015...)

Tests/Linq/Update/MergeTests.IQueryableSource.cs
```diff
@@ -0,0 +1,194 @@
...
+				AssertRow(InitialTargetData[1], result[1], null, null);
+				AssertRow(InitialTargetData[2], result[2], null, 203);
+				AssertRow(InitialTargetData[3], result[3], null, null);
+				AssertRow(InitialSourceData[2], result[4], null, null);
+				AssertRow(InitialSourceData[3], result[5], null, 216);
+			}
+		}
+
+		[Test]
+		public void MergeUsingCteJoin([MergeDataContextSource] string context)
```
MaceWindu, line 135: disable tests for other providers with ActiveIssue(3015...)

Tests/Linq/Update/MergeTests.IQueryableSource.cs
```diff
@@ -0,0 +1,194 @@
...
+				AssertRow(InitialTargetData[1], result[1], null, null);
+				AssertRow(InitialTargetData[2], result[2], null, 203);
+				AssertRow(InitialTargetData[3], result[3], null, null);
+				AssertRow(InitialSourceData[2], result[4], null, null);
+				AssertRow(InitialSourceData[3], result[5], null, 216);
+			}
+		}
+
+		[Test]
+		public void MergeUsingCteWhere([MergeDataContextSource] string context)
```
MaceWindu, line 165: disable tests for other providers with ActiveIssue(3015...)

MaceWindu (approved):

sdanyliv (approved):

PR #2371, opened 2020-07-26
Update README.md

imba-tjd:
It seems that .net core 3.0 has this class.

https://docs.microsoft.com/en-us/dotnet/api/system.configuration.configurationmanager?view=netcore-3.1#moniker-applies-to

Change the title if you want.


MaceWindu:
Thanks!

PR #2390, opened 2020-08-04
Fix for #2378. Corrected OrderBy merging for optimized subqueries.

### LINKED ISSUE #2378: Linq2db v3.x - Combined OrderBy has different order
We had following linq query:

                var qryUnsorted =
                    from tt in s.Query<TextTranslationDTO>()
                    select tt;

                
               var qrySorted1 = from tt in qryUnsorted
                    orderby (from l in s.Query<LanguageDTO>()
                        where l.AlternativeLanguageID == tt.LanguageId // 1. Order by
                        select l.LanguageID).Count()
                    select tt;

                
                var qrySorted2 = from tt in qrySorted1
                                 join t in s.Query<TextDTO>() on tt.TextId equals t.Id
                                 orderby t.ServerOnlyText // 2. Order by
                                 select tt;

                var translation = qrySorted2.Select(tt => tt.Text).FirstOrDefault();
                var qryString = s.LastQuery();

This Generates following query in linq2db 3.x:

	SELECT TOP (@take)
		[tt].[Text]
	FROM
		[Common_Translation_TextTrans] [tt]
			INNER JOIN [Common_Translation_Texts] [t] ON [tt].[TextId] = [t].[Id]
	ORDER BY
		(
			SELECT
				Count(*)
			FROM
				[Common_Language] [l]
			WHERE
				[l].[AlternativeLanguageID] = [tt].[LanguageId]
		),
		[t].[ServerOnlyText]

but in linq2db 2.x the orderbys where switched

sdanyliv:
Fixes #2378.

### DIFF
--- a/Source/LinqToDB/Common/Utils.cs
+++ b/Source/LinqToDB/Common/Utils.cs
@@ -119,5 +119,41 @@ namespace LinqToDB.Common
 				currentCounters.Add(name, startDigit);
 			}
 		}
+
+		public static void RemoveDuplicates<T>(this IList<T> list, IEqualityComparer<T>? comparer = null)
+		{
+			if (list.Count <= 1)
+				return;
+
+			var hashSet = new HashSet<T>(comparer ?? EqualityComparer<T>.Default);
+			var i = 0;
+			while (i < list.Count)
+			{
+				if (hashSet.Add(list[i]))
+					++i;
+				else
+				{
+					list.RemoveAt(i);
+				}
+			}
+		}
+
+		public static void RemoveDuplicates<T, TKey>(this IList<T> list, Func<T, TKey> keySelector, IEqualityComparer<TKey>? comparer = null)
+		{
+			if (list.Count <= 1)
+				return;
+
+			var hashSet = new HashSet<TKey>(comparer ?? EqualityComparer<TKey>.Default);
+			var i = 0;
+			while (i < list.Count)
+			{
+				if (hashSet.Add(keySelector(list[i])))
+					++i;
+				else
+				{
+					list.RemoveAt(i);
+				}
+			}
+		}
 	}
 }
--- a/Source/LinqToDB/Common/Utils.cs
+++ b/Source/LinqToDB/Common/Utils.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 
 namespace LinqToDB.Common
 {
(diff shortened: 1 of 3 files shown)


Source/LinqToDB/Common/Utils.cs
```diff
@@ -119,5 +120,60 @@ public static class Utils
 				currentCounters.Add(name, startDigit);
 			}
 		}
+
+		public static void RemoveDuplicates<T>(this IList<T> list, IEqualityComparer<T>? comparer = null)
```
viceroypenguin, line 124: Could this be simplified as a call:
```
RemoveDuplicates<T>(...)
{
    RemoveDuplicates(list, x => x, comparer);
}
```

If you make the lambda a static local method, it should not even add an allocation.

sdanyliv: I had this variant, but was not so lazy and created another one with less allocations/invokations )

viceroypenguin:
Regarding the duplicate orderby fix, what should be the expected behavior here?
```
				var firstOrder =
					from p in db.Parent
					orderby p.ParentID, p.OtherColumn
					select p;

				var secondOrder =
					from p in firstOrder
					join pp in db.Parent on p.ParentID equals pp.ParentID
					orderby p.ParentID descending 
					select p;
```

Since `p.ParentID` is now not directly next to each other, I don't think _removing_ the first one is the proper answer. Maybe only remove if they're consecutive? I don't know that I have a good answer, just covering corner case here.

Source/LinqToDB/Common/Utils.cs
viceroypenguin: Sure. Just hate to see unnecessarily duplicated code. You're right though, no way to save the allocation on generating the delegate for the function I don't think.

sdanyliv:
I thought about this. I remove only last elements
```cs
order by field1, other1, field1 desc, other2, field1, other2 desc
```
Will be simplified to
```cs
order by field1, other1, other2
```
Ordering by the same field later do nothing because it is already sorted by this field. And if second field is used for sorting it means that previous one is equal. So even `descending/ascending` will not count.

viceroypenguin:
Ok, re: the `RemoveDuplicates()` code, I agree with you. The Unit Test doesn't match that, I don't think, because it confirms that `p.ParentID` is _descending_, not _ascending_, and the first instance is _ascending_.

sdanyliv:
last oderby is superior to previous. It is original issue ;) so descending wins here.

Issue #2395, opened 2020-08-06
DST issue in npgsql bulk upload with TSTZRange data type for Postrgre


Shane32:
When bulk inserting data into postgre, with the `tstzrange` date type, npgsql native bulk upload ignores the exact time of day when converting from UTC.

This causes conversion errors for any date that occurs on a DST changeover day after the changeover time.

To test this issue on a computer, bulk insert data of the `tstzrange` data type where one of the dates consist of a DST changeover date with a time of midnight.  It will convert successfully.  Then try with 4am; it will fail to convert successfully.

Note: don't test fall dates using times between 1am and 2am, since they occur twice due to DST (so those times represent two possible UTC times).  The .Net `TimeZoneInfo` class assumes 1am is after the change, not before it.

Shane32:
Note that during the bulk upload of the tstzrange values, dates of a DateTimeKind property set to Unspecified are interpreted as Utc, whereas data returned from a select query is in the local time zone.

Shane32:
Ideally `tstzrange` should be of the type `NpgsqlRange<DateTimeOffset>`, which is not yet supported by npgsql.  See here:
* https://github.com/npgsql/npgsql/issues/2436 - Support range with more than one CLR type
* https://github.com/npgsql/npgsql/pull/2438 - Range cleanup and support range with two CLR types

This still doesn't fix the conversion bug within npgsql for DateTime types, but at least people would have a bug-free option to set and return values.

Shane32:
Thanks to @MaceWindu I've realized that I misinterpreted the failing test.  Bulk uploading data interprets dates with DateTimeKind set to Unspecified as Utc; so it is inserting Utc dates and returning Local times.  Hence, the changeover times I tested were off by 5 hours (eastern time).  Once the test is reconfigured to insert Utc times at the changeover time, and compares the results with those times converted to local time, all tests pass successfully.

Seems like unintuitive behavior, but that's what it does.

See here, where it by default returns times in local time:
https://github.com/npgsql/npgsql/blob/a43058a4ace95e4a74fe22a7fa4d77e6787d6cf3/src/Npgsql/TypeHandlers/DateTimeHandlers/TimestampTzHandler.cs#L55

And see here, where it interprets DateTimeKind.Unspecified as Utc:
[...]

PR #2345, opened 2020-07-14
Fixed a few unit tests.

igor-tkachev:
Some unit tests do not work on my local. For example DB2 tests fail with the following errors:

```
  Expected Was
  1.1 --- 1.1000000000000000000000000
  2.5 --- 2.5000000000000000000000000
  4.0 --- 4.0000000000000000000000000
  4.5 --- 4.5000000000000000000000000
  5.5 --- 5.5000000000000000000000000
  6.6 --- 6.6000000000000000000000000
  9.0 --- 9.0000000000000000000000000
  9.6 --- 9.6000000000000000000000000
  10.8 --- 10.8000000000000000000000000
  11.4 --- 11.4000000000000000000000000
  11.4 --- 11.4000000000000000000000000
```

PR #2396, opened 2020-08-06
Separate broken npgsql tstzrange tests for postgre

Shane32:
See #2395 

This separates the tests failing due to DST from `TestRangeBulkCopy` and `TestRangeBulkCopyAsync` and into a separate test called `TestRangeBulkCopyDST`.  The new test is marked with `[ActiveIssue]`

### DIFF
--- a/Tests/Linq/DataProvider/PostgreSQLTests.cs
+++ b/Tests/Linq/DataProvider/PostgreSQLTests.cs
@@ -1514,11 +1514,15 @@ namespace Tests.DataProvider
 			public int Id { get; set; }
 
 			[Column(DbType = "daterange")]
-			public NpgsqlRange<DateTime> DateRange { get; set; }
+			public NpgsqlRange<DateTime> DateRangeInclusive { get; set; }
+
+			[Column(DbType = "daterange")]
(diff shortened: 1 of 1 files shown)


Tests/Linq/DataProvider/PostgreSQLTests.cs
```diff
@@ -1580,14 +1580,14 @@ public void TestRangeBulkCopy([IncludeDataSources(TestProvName.AllPostgreSQL)] s
 					items.Select(t => new
 					{
 						t.Id,
-						t.DateRange,
+						DateRange = new NpgsqlRange<DateTime>(new DateTime(2000 + t.Id, 2, 3), true, new DateTime(2000 + t.Id, 3, 4), false),
```
Shane32, line 1583: This line fixes the test failure introduced in #2345

Shane32:
Based on the fact that `TestRangeBulkCopy` passed in Azure before #2345 broke it, I would guess that the Azure server running the tests is configured to use a US time zone subject to DST.  If that assumption is true, `TestRangeBulkCopyDST` should also pass when run in Azure.

MaceWindu:
Looked into what's going on and added comments to document what and why is going on.

PR #2407, opened 2020-08-13
Add generic constraint of class type to CreateTempTable

Shane32:
See #1934 

This PR prevents a call of `CreateTempTable<int>(...)`.  Note that a generic constraint on `TempTable<T>` was not added due to conflicts elsewhere in the codebase.

PR #2408, opened 2020-08-13
Add asynchronous overloads for CreateTempTable

Shane32:
This PR adds `CreateTempTableAsync<T>` overloads to `LinqToDB.DataExtensions`.

Notes:
* `TempTable<T>` now implements `IAsyncDisposable` for compatible frameworks
* Since constructors cannot be asynchronous, I have added `CreateAsync` static methods to `TempTable` to facilitate this.  Those static methods call a protected constructor to create the class after `CreateTableAsync` has been called.
* Some of the overloads include an _optional_ `action` that will run before the table is populated via a bulk copy operation.  This optional `action` parameter is now expected to be an asynchronous delegate.  If I were to add another overload for a synchronous action, the compiler would not be able to select which overload was intended to be used.  So I only included the asynchronous `action` parameter for async calls.
* The new extension methods include a `where T : class` clause, similar to #2407 .
* I have added a few more tests to verify basic asynchronous operation.
* Xml comments are almost identical to the synchronous methods.  Lmk if I should adjust them to say "Asynchronously creates a new temporary..." and "Returns an asynchronous task representing...." and so on.
* I have not added support for `IAsyncEnumerable`; shall I?  If someone has that need, they can of course create the temp table first and bulk upload with a separate call.

### DIFF
--- a/Source/LinqToDB/TempTable.cs
+++ b/Source/LinqToDB/TempTable.cs
@@ -164,7 +164,7 @@ namespace LinqToDB
 		/// <param name="table">Table instance.</param>
 		protected TempTable(ITable<T> table)
 		{
-			_table = table;
+			_table = table ?? throw new ArgumentNullException(nameof(table));
 		}
 
 		/// <summary>
--- a/Source/LinqToDB/TempTable.cs
+++ b/Source/LinqToDB/TempTable.cs
@@ -336,33 +336,7 @@ namespace LinqToDB
 		/// <returns>Number of records, inserted into table.</returns>
 		public long Insert(IQueryable<T> items)
 		{
-			var type = typeof(T);
-			var ed   = _table.DataContext.MappingSchema.GetEntityDescriptor(type);
-			var p    = Expression.Parameter(type, "t");
-
-			var l = _setterDic.GetOrAdd(type, t =>
-			{
-				if (t.IsAnonymous())
-				{
-					var nctor   = (NewExpression)items.Expression.Find(e => e.NodeType == ExpressionType.New && e.Type == t)!;
-					var members = nctor.Members
-						.Select(m => m is MethodInfo info ? info.GetPropertyInfo()! : m)
-						.ToList();
-
-					return Expression.Lambda<Func<T,T>>(
-						Expression.New(
-							nctor.Constructor,
-							members.Select(m => ExpressionHelper.PropertyOrField(p, m.Name)),
-							members),
-						p);
-				}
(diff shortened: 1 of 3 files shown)


Source/LinqToDB/TempTable.cs
```diff
@@ -155,6 +158,141 @@ public class TempTable<T> : ITable<T>, ITableMutable<T>, IDisposable
...
+			BulkCopyOptions? options = null,
+			string? databaseName     = null,
+			string? schemaName       = null,
+			string? serverName       = null)
+		{
+			if (db == null) throw new ArgumentNullException(nameof(db));
+			if (items == null) throw new ArgumentNullException(nameof(items));
+
+			var table = await CreateAsync(db, tableName, databaseName, schemaName, serverName)
+				.ConfigureAwait(false);
```
sdanyliv, line 233: Here and later we have to sue global option for ConfigureAwait, please check other places for sample

Shane32: Whoops I forgot!  👍

Shane32:
@MaceWindu Want to look at this commit: https://github.com/linq2db/linq2db/pull/2408/commits/dac579771289d643cfd53a1e6db7102e2c49f11e which fixes using the row-by-row bulk copy operations on an `ITable<T>` where the table name is not inferred from the model?  This should fix the failing tests with the Access database provider.

MaceWindu:
Looks valid

MaceWindu:
Could you check bulk copy code to see if we have other places like that? or you already checked?

Shane32:
It seems that both the `MultipleRowsCopy` function and the native bulk upload providers use the `BasicBulkCopy.GetTableName` protected function to create a table name.  Part of the code in that method is:
```csharp
var serverName   = options.ServerName   ?? table.ServerName;
var databaseName = options.DatabaseName ?? table.DatabaseName;
var schemaName   = options.SchemaName   ?? table.SchemaName;
var tableName    = options.TableName    ?? table.TableName;
```
Since the `RowByRowCopy` copy calls `Insert` repeatedly, and since `Insert` accepts separate `serverName`, `databaseName`, `schemaName`, and `tableName` parameters, I added `?? table.TableName` and similar:
```csharp
table.DataContext.Insert(item, options.TableName ?? table.TableName, options.DatabaseName ?? table.DatabaseName, options.SchemaName ?? table.SchemaName, options.ServerName ?? table.ServerName);
```

PR #2437, opened 2020-08-22
Fix for #2434. String interpolation handling.

### LINKED ISSUE #2434: Syntax error (missing operator) in query expression
Recently I refactored some code and started to receive exceptions like `Syntax error (missing operator) in query expression '[t1].[Nam] [x].[Nam]'`. See `ORDER BY` with space between fields.

```SQL
SELECT
	[x].[Id],
	[t1].[Nam],
	[x].[Nam],
	[x].[Del]
FROM
...
ORDER BY
	[t1].[Nam] [x].[Nam]
```

This bug has been introduced because I changed details of expression methods
```c#
[ExpressionMethod(nameof(FullnameExpr), IsColumn = true)]
public string Fullname { get; set; }

private static Expression<Func<FtqWorkstationGroupDTO, string>> FullnameExpr() =>
```
from string concatenation
```c#
f => f.SectorPart.Name + " " + f.Name;
```
to string interpolation
```c#
f => $"{f.SectorPart.Name} {f.Name}";
```
With concatenation query is ok
```SQL
SELECT
	[x].[Id],
	[t1].[Nam],
	[x].[Nam],
	[x].[Del]
FROM
...
ORDER BY
	[t1].[Nam] + ' ' + [x].[Nam]
```
In my query I have ordering by `Fullname` like so:
```c#
return db.FtqWorkstationGroups
    .Where(x => x.SectorId == sectorId)
    .OrderBy(x => x.Fullname);
```
Is it some kind of limitation or a bug?

### Environment details
linq2db version: 3.0.1
Database Server: MS Access via OleDb

MaceWindu:
Fixes #2434

### DIFF
--- a/Tests/Linq/Linq/WhereTests.cs
+++ b/Tests/Linq/Linq/WhereTests.cs
@@ -1575,11 +1575,13 @@ namespace Tests.Linq
 		{
 			using (var db = GetDataContext(context))
 			{
-				var cnt = db.Person.Count(p => p.LastName + ", " + p.FirstName == $"{p.LastName}, {p.FirstName}"
-				                               && "<" + p.LastName + ", " + p.FirstName + ">" ==
-				                               $"<{p.LastName}, {p.FirstName}>"
-				                               && "<" + p.LastName + p.FirstName + ">" == $"<{p.LastName}{p.FirstName}>"
-				);
+				var cnt = db.Person
+					.Count(p => p.LastName + ", " + p.FirstName == $"{p.LastName}, {p.FirstName}"
+					            && "<" + p.LastName + ", " + p.FirstName + ">" == $"<{p.LastName}, {p.FirstName}>"
+					            && "<" + p.LastName + p.FirstName + ">" == $"<{p.LastName}{p.FirstName}>"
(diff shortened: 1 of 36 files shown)


Tests/Linq/Linq/WhereTests.cs
```diff
@@ -1570,6 +1570,22 @@ public void OptionalObjectInCondition([DataSources(false)] string context)
...
 
+		[Test]
+		public void SrtingInterpolationTests([DataSources(false)] string context)
+		{
+			using (var db = GetDataContext(context))
+			{
+				var cnt = db.Person.Count(p => p.LastName + ", " + p.FirstName == $"{p.LastName}, {p.FirstName}"
+				                               && "<" + p.LastName + ", " + p.FirstName + ">" ==
+				                               $"<{p.LastName}, {p.FirstName}>"
+				                               && "<" + p.LastName + p.FirstName + ">" == $"<{p.LastName}{p.FirstName}>"
```
MaceWindu, line 1581: maybe add case with escaped {}? like `{{test}}`

sdanyliv: Good point, will add tests.

PR #2453, opened 2020-08-29
Add a ConcurrentDictionary cache to back ReflectionExtensions.GetItemType

to11mtm:
While working on another Library I noticed that calls `ReflectionExtensions.GetItemType` added up in certain cases (Mostly around `Insert<T>` and bulk copies.) With the benchmarks I'm running off a local branch I'm seeing some performance boosts for large series of individual writes.

### DIFF
--- a/Source/LinqToDB/Extensions/ReflectionExtensions.cs
+++ b/Source/LinqToDB/Extensions/ReflectionExtensions.cs
@@ -660,7 +660,7 @@ namespace LinqToDB.Extensions
 			return false;
 		}
 		
-		static readonly ConcurrentDictionary<Type?,Type?> getItemTypeCache = new ConcurrentDictionary<Type?, Type?>();
+		static readonly ConcurrentDictionary<Type,Type?> getItemTypeCache = new ConcurrentDictionary<Type, Type?>();
 		
 		[return: NotNullIfNotNull("type")]
 		public static Type? GetItemType(this Type? type)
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -659,39 +659,44 @@ public static bool IsGenericEnumerableType(this Type type)
...
 		[return: NotNullIfNotNull("type")]
 		public static Type? GetItemType(this Type? type)
 		{
 			if (type == null)
 				return null;
+			return getItemTypeCache.GetOrAdd(type, (t) =>
+			{
+				if (type == typeof(object))
+					// if it possible to have null here or we should remove check?
+					return type.HasElementType ? type.GetElementType() : null;
```
sdanyliv, line 674: I think you have to use `t` parameter instead of creating closure,

sdanyliv (changes requested):

Source/LinqToDB/Extensions/ReflectionExtensions.cs
to11mtm: Good Catch. Thanks!

to11mtm:
I'm also seeing some warnings from build due to nullable references, I'll clean those up.

MaceWindu (approved):

PR #2455, opened 2020-08-30
Test baselines

MaceWindu:
Implement test baselines.

Done:
- baselines generation
- use fixed values in tests where possible (Guid.NewGuid, DateTime.Now)
- disabled baselines for tests that cannot produce static baselines: multithreaded tests, tests with server-side guid/current date/id generation
- implemented linqservice logging for baselines
- removed duplicate testcases to avoid baselines conflicts
- fixed names for some tests to avoid file path size limitations
- refactored Azure Pipelines yaml to use templates to reduce copy-paste
- implement baselines changes PR for azure test runs to https://github.com/linq2db/linq2db.baselines/pulls repository
- refactored SqlTable.Fields property to expose ordered collection instead of dictionary to avoid hash-dependent field order
- fixed issues with code, dependent on non-documented Type.GetMembers() results ordering
- fix F# record constructor mapping to use compiler metadata properly
- added new test providers for sql server 2016 and 2019, removed unversioned `SqlServer` test provider
- added procedures sorting by schema+name in schema provider for DB2 to avoid non-stable resutls order
- fixed discovered concurrency issues in fluent metadata reader
- use [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) format for `DateTime` and `DateTimeOffset` parameter values in SQL log for providers without datetime/datetimeoffset literals. This allows us to generate same debug SQL for any cultute/locale
- replace all uses of `Console` in tests with `TestContext`

---
##### What baseline is?
Baseline is a text file for single test case, which contains SQL, generated during test. More precisely it contains text from BeforeExecute trace event. Changes to this file allows us to detect regressions (or improvements) in sql generation.
Baseline file name structure:
`<baselines_root_directory>\<framework_target>\<context_aka_provider_name>\<test\name\space>\<test_class_name>\<testcase_name>.sql`

Conditions for baseline file generation:
- `<User>DataProviders.json` contains path to baseline root directory (`BaselinesPath` property)
- baseline root directory exists
- testcase generated sql

##### Where baselines stored
Because total number of baselines for all test targets is approx. 300+k files and hundreds of megabytes, we store them in separate [repository](https://github.com/linq2db/linq2db.baselines).

##### Azure integration
All azure test pipelines will use baselines generation and will create pr to baselines repo for detected baselines changes. Baselines repository configured to use only squash-merge prs with automatic pr branch removal.
For each `linq2db` source PR with baselines changes we create only one baselines PR, which will be reused by all testruns from `linq2db` PR. After `linq2db` PR merged, user should also merge baselines PR.

### Known Issues
- ~parallel baselines pushes could conflict with other test jobs. We should add rebase retries to mitigate it (non-blocking issue and could be done later)~ resolved. we will try to rebase up to 10 times before giveup
- baselines for several tests disabled as they generate non-repeatable sql for some reason. needs debugging


MaceWindu:
baselines will be uploaded by separate commits to make code reviewable as there are thouthands of them

Source/LinqToDB/Metadata/FluentMetadataReader.cs
```diff
@@ -11,8 +11,11 @@ namespace LinqToDB.Metadata
 
 	public class FluentMetadataReader : IMetadataReader
 	{
-		readonly ConcurrentDictionary<Type,List<Attribute>>                       _types          = new ConcurrentDictionary<Type,List<Attribute>>();
-		readonly ConcurrentDictionary<Type,ConcurrentDictionary<MemberInfo,byte>> _dynamicColumns = new ConcurrentDictionary<Type,ConcurrentDictionary<MemberInfo,byte>>();
+		// don't forget to put lock on List<Attribute> when access it
+		readonly ConcurrentDictionary<Type,List<Attribute>>                            _types          = new ConcurrentDictionary<Type,List<Attribute>>();
+		// set used to guaranty uniqueness
+		// list used to guaranty same order for columns in select queries
```
viceroypenguin, line 17:
```suggestion
		// set used to guarantee uniqueness
		// list used to guarantee same order for columns in select queries
```

Source/LinqToDB/Metadata/FluentMetadataReader.cs
```diff
@@ -78,14 +85,26 @@ public T[] GetAttributes<T>(Type type, MemberInfo memberInfo, bool inherit = tru
 		public void AddAttribute(MemberInfo memberInfo, Attribute attribute)
 		{
 			if (memberInfo.IsDynamicColumnPropertyEx())
-				_dynamicColumns.GetOrAdd(memberInfo.DeclaringType!, new ConcurrentDictionary<MemberInfo, byte>()).TryAdd(memberInfo, 0);
+			{
+				var dynamicColumns = _dynamicColumns.GetOrAdd(memberInfo.DeclaringType!, new Tuple<ISet<MemberInfo>, IList<MemberInfo>>(new HashSet<MemberInfo>(), new List<MemberInfo>()));
+
+				lock (dynamicColumns)
+					if (dynamicColumns.Item1.Add(memberInfo))
```
viceroypenguin, line 92: Why do we need both the `ISet<>` and the `IList<>`? Since we're calling `.ToArray()` every time anyway, can't we just call that on the `ISet<>` instead of keeping the collection twice in two different data structures?

Tests/Base/BaselinesWriter.cs
```diff
@@ -0,0 +1,73 @@
...
+
+			var context = GetTestContextName(test);
+
+#if NET46
+			var target = "net46";
+#elif NETCOREAPP2_1
+			var target = "core21";
+#elif NETCOREAPP3_1
+			var target = "core31";
+#endif
```
viceroypenguin, line 27:
```suggestion
#else
#error "Build Target must be specified here."
#endif
```

viceroypenguin: I know technically it will be build error anyway, but better to mark the error early w/ clear message.

Tests/Base/TestBase.cs
```diff
@@ -543,38 +575,20 @@ protected IEnumerable<ParentInheritanceBase> ParentInheritance
...
-		}
 
 		private   List<ParentInheritance1>? _parentInheritance1;
-		protected List<ParentInheritance1>  ParentInheritance1
-		{
-			get
-			{
-				return _parentInheritance1 ??=
+		protected List<ParentInheritance1> ParentInheritance1 => _parentInheritance1 ??=
 					ParentInheritance.Where(p => p is ParentInheritance1).Cast<ParentInheritance1>().ToList();
```
viceroypenguin, line 583:
```suggestion
		protected List<ParentInheritance1> ParentInheritance1 => 
			_parentInheritance1 ??=
					ParentInheritance.Where(p => p is ParentInheritance1).Cast<ParentInheritance1>().ToList();
```

Tests/Base/TestBase.cs
```diff
@@ -543,38 +575,20 @@ protected IEnumerable<ParentInheritanceBase> ParentInheritance
...
-			}
-		}
 
 		private   List<ParentInheritanceBase4>? _parentInheritance4;
-		protected List<ParentInheritanceBase4>   ParentInheritance4
-		{
-			get
-			{
-				return _parentInheritance4 ??= Parent
+		protected List<ParentInheritanceBase4> ParentInheritance4 => _parentInheritance4 ??= Parent
```
viceroypenguin, line 586:
```suggestion
		protected List<ParentInheritanceBase4> ParentInheritance4 =>
			_parentInheritance4 ??= Parent
```

Tests/Base/TestBase.cs
```diff
@@ -23,23 +18,40 @@
...
-	using System.Diagnostics.CodeAnalysis;
 	using LinqToDB.DataProvider.Informix;
 	using Model;
 	using NUnit.Framework.Internal;
 	using Tools;
 
-	//	[Order(1000)]
 	public class TestBase
 	{
+		protected class TestData
```
viceroypenguin, line 37:
```suggestion
		protected static class TestData
```

viceroypenguin, line 37: Should we add a ticket to research a Roslyn analyzer for the tests to error if someone writes `Guid.NewGuid()` or similar instead of using these? Will probably be obvious in the output baselines, but might make it easier on future dev work.

Tests/Linq/DataProvider/DB2Tests.cs
```diff
@@ -526,7 +526,7 @@ public void BulkCopyLinqTypes([IncludeDataSources(CurrentProvider)] string conte
 									MoneyValue    = 1000m + n,
 									DateTimeValue = new DateTime(2001, 1, 11, 1, 11, 21, 100),
```
viceroypenguin, line 527: Need to switch to `TestData`?

Tests/Linq/DataProvider/DB2Tests.cs
```diff
@@ -557,7 +557,7 @@ public async Task BulkCopyLinqTypesAsync([IncludeDataSources(CurrentProvider)] s
...
```
viceroypenguin, line 558: Need to switch to `TestData`?

Source/LinqToDB/Metadata/FluentMetadataReader.cs
MaceWindu: IList added to make sure that GetDynamicColumns method returns columns in same order as they were added to mapping. We don't have support for column ordering in queries right now (only for create table), so it is currently used only to make sure we have same sql for dynamic property queries. With this change we can generate stable baselines for dynamic property tests.

viceroypenguin: Makes sense. Can we add a comment in the file somewhere so future readers of the code understand?

viceroypenguin: Derp. There is a comment. I just didnt' connect the two.

Tests/Base/TestBase.cs
MaceWindu: We can do it, but there are several exceptions where this method used (mostly in tests that test Guid.NewGuid, DateTime.Now, etc members mapping to sql)

Tests/Linq/DataProvider/DB2Tests.cs
MaceWindu: It is static, but probably yes, will do

Tests/Base/TestBase.cs
viceroypenguin: We should be able to codify exceptions in some way. I'll create an issue after this PR closes

Tests/Linq/DataProvider/DB2Tests.cs
MaceWindu: decided not to touch it for now.

viceroypenguin (approved):

Shane32:
@MaceWindu sorry I don't have time to look at this real deep but I like where this PR is headed; looks good!  I had a few thoughts in my mind, which you probably already thought of, but I thought I'd just jot them down here anyway:
- Does the entire "hundreds of megabytes" repo get downloaded for each test run?  Any issues with that?  Seems like a lot of wasted bandwidth.  I'm guessing you already figured this out.
- Automating this will be extremely useful, as any new test (which there is typically one or more new tests for any PR) will require changes in this other repo.
- Regarding the fact that it creates so many PRs for a single source PR:
  - What if the code created a branch in the other repo with the same name as the branch it was testing?  So it would only replace the files in the proper folder that is applicable to its test run, adding another commit to the branch if it already existed.  Then if no PR exists for that branch, it would be created.  This has a benefit that with future commits/tests, no additional PRs are created; merely the PR is updated in sync with the branch.  Synchronization might be an issue, of course.

Just some ideas in case you've not already thought of them.  Again, I like this concept.

MaceWindu:
@Shane32 

>  Does the entire "hundreds of megabytes" repo get downloaded for each test run? Any issues with that? Seems like a lot of wasted bandwidth. I'm guessing you already figured this out.

Azure tests show that it adds ~1 minute for repo download to each test run. Seems ok to me. Also after I will finish polishing this PR, I will reset repository content to remove current commits, generated during testing.

>  What if the code created a branch in the other repo with the same name as the branch it was testing? So it would only replace the files in the proper folder that is applicable to its test run, adding another commit to the branch if it already existed. Then if no PR exists for that branch, it would be created. This has a benefit that with future commits/tests, no additional PRs are created; merely the PR is updated in sync with the branch. Synchronization might be an issue, of course.

That's what I was planning to implement after I finish fixing unstable tests, looks like I'm on right track :smile:

Tests/Base/TestBase.cs
```diff
@@ -23,23 +18,49 @@
...
 	using Model;
 	using NUnit.Framework.Internal;
 	using Tools;
 
-	//	[Order(1000)]
 	public class TestBase
 	{
+		protected static class TestData
+		{
+			public static readonly DateTimeOffset DateTimeOffset    = new DateTimeOffset(2020, 2, 29, 17, 54, 55, 123, TimeSpan.FromMinutes(45)).AddTicks(1234);
```
MaceWindu, line 39: that's fun. using non-current TZ offet uncovered potential issue in oracle datetimeoffset mapping

MaceWindu: Created #2499 to fix it after baselines.

MaceWindu:
Test baselines changed by this PR. Don't forget to merge/close baselines PR after this pr merged/closed.

sdanyliv (approved):
Awesome! Just resolve one comment.

PR #2477, opened 2020-09-09
Fix for #2471. Added TransactionScope check when loading preambles.

### LINKED ISSUE #2471: LoadWith (also via ProjectTo) fails when run within TransactionScope
LoadWith (also via ProjectTo) fails when run within TransactionScope.  

Since [this one](https://github.com/linq2db/linq2db/issues/962) was closed, but the issue persist, I'm opening a new one.

It'd be great if you could investigate and resolve the problem.

### Example

```c#
// Db model

[Table(Schema = "dbo", Name = "Parent")]
public partial class Parent
{
    [Column, PrimaryKey, NotNull]
    public int Id { get; set; }

    [Column, NotNull]

    #region Associations

    /// <summary>
    /// FK_Parent_Child_BackReference
    /// </summary>
    [Association(ThisKey = "Id", OtherKey = "ParentId", CanBeNull = true, Relationship = Relationship.OneToMany, IsBackReference = true)]
    public IEnumerable<Child> Children { get; set; }

    #endregion
}

[Table(Schema = "dbo", Name = "Child")]
public partial class Child
{
    [Column, PrimaryKey, NotNull]
    public int Id { get; set; }

    [Column, Nullable]

    [Column, NotNull]
    public int ParentId { get; set; }

    #region Associations

    /// <summary>
    /// FK_Parent_Child
    /// </summary>
    [Association(ThisKey = "ParentId", OtherKey = "Id", CanBeNull = false, Relationship = Relationship.ManyToOne, KeyName = "FK_Parent_Child", BackReferenceName = "Children")]
    public Parent Parent { get; set; }

    #endregion
}


// Entities
public class ParentModel
{
    public int Id { get; set; }
    public ICollection<ChildModel> Children { get; set }
}

public class ChildModel
{
    public int Id { get; set; }
}


// Mappings
Mapper.CreateMap<Parent, ParentModel>();
Mapper.CreateMap<Child, ChildModel>();

// asp.net core config
LinqToDB.Common.Configuration.Linq.AllowMultipleQuery = true;
services.AddLinqToDbContext<IDb, Db>((provider, options) =>
{
    options.UseSqlServer("connectionString");
});


// run it
public class SomeRepository : ISomeRepository
{
    private readonly IDb _db;
    private readonly IMapper _mapper;

    public SomeRepository(IDb db, IMapper mapper)
    {
        _db = db;
        _mapper = mapper;
    }

    public async Task<Parent> GetParentById(int id)
    {
        using (var transaction = new TransactionScope(TransactionScopeOption.Required, new TransactionOptions { IsolationLevel = IsolationLevel.Serializable }, TransactionScopeAsyncFlowOption.Enabled))
        {
            var result = await _db.GetTable<Parent>()
                                  .Where(x => x.Id == id)
                                  .ProjectTo<ParentModel>(_mapper.ConfigurationProvider) // which implicitly joins AnotherTable
                                  .FirstAsync();

            transaction.Complete();

            return result;
        }
    }
}
```

### Error
The COMMIT TRANSACTION request has no corresponding BEGIN TRANSACTION

### Environment details
linq2db version: *3.1.2*
Database Server: *Azure Sql Server*
Database Provider: *SQL Server*
Operating system: *Windows, Mac OS*
.NET Framework: *.net core 3.1.300*

sdanyliv:
Fixes #2471

### DIFF
--- a/Source/LinqToDB/Data/TransactionScopeHelper.cs
+++ b/Source/LinqToDB/Data/TransactionScopeHelper.cs
@@ -25,8 +25,8 @@ namespace LinqToDB.Data
 
 			if (assembly != null)
 			{
-				var t = assembly.GetType("System.Transactions.ContextData");
-				var currentDataProperty = t.GetProperty("TLSCurrentData", BindingFlags.NonPublic | BindingFlags.Static);
+				var t = assembly.GetType("System.Transactions.Transaction");
+				var currentDataProperty = t.GetProperty("Current", BindingFlags.Public | BindingFlags.Static);
 				if (currentDataProperty != null)
 				{
 					var body   = Expression.MakeMemberAccess(null, currentDataProperty);
--- a/Source/LinqToDB/Data/TransactionScopeHelper.cs
+++ b/Source/LinqToDB/Data/TransactionScopeHelper.cs
@@ -7,7 +7,7 @@ namespace LinqToDB.Data
 {
 	internal static class TransactionScopeHelper
 	{
-		static Func<object?> _getCurrentScopeDelegate = GetTransactionScopeFunc();
+		static readonly Func<object?> _getCurrentScopeDelegate = GetTransactionScopeFunc();
 
 		public static bool IsInsideTransactionScope
 		{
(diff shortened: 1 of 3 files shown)


MaceWindu (approved):

Source/LinqToDB/Data/TransactionScopeHelper.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+
+namespace LinqToDB.Data
+{
+	internal static class TransactionScopeHelper
+	{
+		static readonly Func<object?> _getCurrentScopeDelegate = GetTransactionScopeFunc();
```
viceroypenguin, line 10: What is the point of doing this as a `Func<object?>` instead of calling `GetTransactionScopeFunc()` directly inside the property? I don't see where this func is getting changed so something different, which would be the primary benefit.

viceroypenguin (approved):

Source/LinqToDB/Data/TransactionScopeHelper.cs
sdanyliv: You mean to compile `Func<bool>` instead?

viceroypenguin: I mean the property could be:
```
public static bool IsInsideTransactionScope =>
  GetTransactionScopeFunc() != null;
```

This field appears unnecessary.

sdanyliv: And it is wrong. We have to compile delegate into static field and call it when it is needed.
Dynamic delegate just make `Transaction.Current` value retrieving.

viceroypenguin: Ah, yes. Sorry. I wasn't reading that correctly. That's what I get for reading at 5am... :)

PR #2481, opened 2020-09-14
Fix for #2479. Correct Parameter usage

### LINKED ISSUE #2479: Incorrect order of serverName, databaseName, schemaName parameters in DataExtensions.cs
When I call 
`db.InsertWithIdentityAsync(entity, schemaName: schema.Name)`

I get exception about my PostgreSQL table does not exist.

I discovered that my `schemaName` value actually are set to `serverName`.

It's because of incorrect order of `serverName, databaseName, schemaName` parameters in [DataExtensions.cs](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataExtensions.cs)

Using named parameters prevents such kind of bugs.

viceroypenguin:
Fixes #2479

### DIFF
--- a/Source/LinqToDB/DataExtensions.cs
+++ b/Source/LinqToDB/DataExtensions.cs
@@ -203,7 +203,7 @@ namespace LinqToDB
 		public static int Insert<T>(this IDataContext dataContext, T obj,
 			string? tableName = null, string? databaseName = null, string? schemaName = null, string? serverName = null)
 		{
-			return Insert<T>(dataContext, obj, null, tableName, serverName, databaseName, schemaName);
+			return Insert<T>(dataContext, obj, null, tableName: tableName, databaseName: databaseName, schemaName: schemaName, serverName: serverName);
 		}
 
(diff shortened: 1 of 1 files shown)


MaceWindu (approved):

MaceWindu:
Too bad C# doesn't have tagged types...

viceroypenguin:
Agreed! Though it is under discussion for C# 10, in the form of enums w/ attached values.

viceroypenguin:
https://github.com/dotnet/csharplang/blob/master/proposals/discriminated-unions.md

MaceWindu:
It looks like more [ADT](https://en.wikipedia.org/wiki/Algebraic_data_type) (e.g. variant in Nemerle). Here it would be enought to have compile time types like one from here https://github.com/dotnet/csharplang/issues/3768 or tagged types from TypeScript.
Anyway,  I'm glad it's finally on radar.

sdanyliv (approved):

PR #2489, opened 2020-09-15
Fix for #2478. Added explicit aggregation column completion.

### LINKED ISSUE #2478: Bug: invalid removal of unused aggregation
We have found a case where linq2db simplifies a query too aggressively and produces an incorrect result.

Our high-level use case is that we produce a paginated list.
To do this we construct a complex query, which is then executed twice: 
- `.Count()` to have the total number of rows
- `.Skip().Take().ToList()` to fetch one page of data

During the `.Count()` execution, no field is selected, which leads to massive simplifications made by linq2db.
From a perf perspective this is very nice, unfortunately we found out one bad simplification that produces incorrect results.

The main query performs CROSS APPLY with an aggregation as a sub-query.

The C# code goes something like this (much simplified, this example could be a simple inner join/group by but in reality we several cross apply):
```csharp
from p in db.GetTable<Parent>()
from c in (from t in db.GetTable<Child>()
           where c.ParentId == p.Id
           group t by 1 into g
           select new { Count = g.Count() })
select new { p.Id, c.Count }
```
This generates the following SQL (correct):
```sql
select p.id, c.count
from Parent p
cross apply (
  select count(*) 
  from Child c
  where c.parentId = p.id
) c
```

But when this request goes through `.Count()`, `c.Count` is not selected anymore and linq2db simplifies everything down to:
```sql
select count(*)
from Parent p
inner join (
  select c.parentId 
  from Child c 
  group by c.parentId
) c on c.parentId = p.id
```

But removing `count(*)`, even though it was unused, is incorrect.

In SQL, a query that has no group by and returns aggregations _always_ returns a single row. If no row matches, then a single result with `count(*)` being 0 is returned.

Linq2db removed the unused `count(*)` and put a `group by` in its place, but the behaviour is different. 
If there is no matching row, then the sub-select returns nothing and the `inner join` fails and discards the row.

-> Linq2db needs to recognize sub-queries that return aggregates without group by.
It needs knowledge that such queries always return _exactly_ one row.
This actually means that a correct and more efficient optimization is to completely remove such sub-query if its results are not used.

### Environment details
linq2db version: 3.1.2
Database Server: Oracle 12c
Database Provider: Oracle Managed Provider
Operating system: Windows
.NET Framework: .net 4.7.2

sdanyliv:
Fixes #2478.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
@@ -119,6 +119,7 @@ namespace LinqToDB.Linq.Builder
 				var expr   = BuildExpression(FieldIndex, Sql);
 				var mapper = Builder.BuildMapper<object>(expr);
 
+				CompleteColumns();
 				QueryRunner.SetRunQuery(query, mapper);
 			}
 
--- a/Source/LinqToDB/Linq/Builder/AllAnyBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/AllAnyBuilder.cs
@@ -97,6 +97,7 @@ namespace LinqToDB.Linq.Builder
 				var expr   = Builder.BuildSql(typeof(bool), 0, sql);
 				var mapper = Builder.BuildMapper<object>(expr);
 
+				CompleteColumns();
 				QueryRunner.SetRunQuery(query, mapper);
 			}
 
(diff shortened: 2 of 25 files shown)


MaceWindu (approved):

viceroypenguin (approved):
Honestly, most of this is still above my head. As long as tests pass, I'm good

Issue #2466, opened 2020-09-02
Allocations when using custom Mapping schemas with DataConnection


to11mtm:
I've noticed that if a MappingSchema is passed into DataConnection, we new up another MappingSchema (to integrate the DataProvider's Mapping Schema).

In one of my codebases, using `Clone` on an existing DataConnection showed a marked improvement in benchmark results (as, among other things, it avoided re-creating the merged Schema).

`Clone` works for the moment but I'm not sure if it's robust for all scenarios one may encouter.

 - Is there a way to avoid this? I think if you passed in a shared `DataProvider` that would work but I don't know if those are considered thread-safe.

 - Would it make sense to add an overload for DataConnection to accept a 'pre-built' mapping schema:

```
public class BuiltMappingSchema
	{
              public BuiltMappingSchema(string providerName, string connectionString,
			MappingSchema schemaToBuild):this(DataConnection.GetDataProvider(providerName, connectionString), schemaToBuild)
              public BuiltMappingSchema(IDataProvider provider, MappingSchema schemaToBuild)
            //make the mapping schema and put it somewhere
         }
```

  - I don't think? caching is an option with current setup of MappingSchema...

sdanyliv:
I have plans to look at this functionality closer, but I do not think it is good idea to create such overload.
Will check what we can do here.

to11mtm:
I had an idea about this that would be less hacky.

If User Provides either Provider or Provider Name with Mapping Schema, we could create specific property in `LinqToDbConnectionOptionsBuilder` and create `BuiltMappingSchema` property appropriately; constructor for `DataConnection` can then utilize to avoid current pitfalls of custom mapping schemas.

If this is OK option I can do.

sdanyliv:
I have Idea:
Add to `DataConnection` static MemoryCache _combinedSchemas;
Schematically:
```cs
		public DataConnection AddMappingSchema(MappingSchema mappingSchema)
		{ 
                       var key = new { MappingSchema.ConfigurationID, mappingSchema.ConfigurationID };
                       MappingSchema  = _combinedSchemas.GetOrAdd(key, 
                  	     o => 
				{
					o.SlidingExpiration = Common.Configuration.Linq.CacheSlidingExpiration;
					return new MappingSchema(mappingSchema, MappingSchema);
				});
			_id  = null;

			return this;
		}
```

to11mtm:
I can do this part for the moment. I'll get a PR in.

PR #2496, opened 2020-09-19
Add MappingSchema MemoryCache to DataConnection

to11mtm:
Use a MemoryCache for Mapping Schemas added to a DataConnection for reduced GC Pressure. Should help with #2466


sdanyliv:
Have you tested difference in allocations?

to11mtm:
Haven't looked at the raw profiler data yet but based on raw numbers for a test doing 1000 Open->Write1Row->Close:

  - Using Cloned DataConnection for each insert:  3000-3400 msg/Sec
  - Creating DataConnection with Built `Linq2DbOptions` containing mapping schema
      - Before Patch: 2700-3000 msg/Sec
      - After Patch: 2900-3200 msg/Sec

MaceWindu (changes requested):
I like this change, but it breaks scenario where user wants to modify mapping schema only for specific connection instance. At least we should alter comment on this method to reflect new behavior.

to11mtm:
> 
> 
> I like this change, but it breaks scenario where user wants to modify mapping schema only for specific connection instance. At least we should alter comment on this method to reflect new behavior.

Could you explain? I thought that the logic would handle most (sane) use cases since the key is based on the MappingSchema configurationId (unless library user gives different mapping schema same names, auto-increment will handle uniqueness on the rest)

MaceWindu:
I'm talking about code like that:
```cs
// method1
using(var db = GetConnection())
{
    // do some stuff
}
// method2
using(var db = GetConnection())
{
    // modify mapping schema. as connection instance has own copy of schema, it will not affect other connections
    db.MappingSchema.AddSomething
    // do some stuff
}

```

but such approach is not a "sane" one, that's why I'm fine if we break it

MaceWindu (changes requested):
don't merge, miniprofiler-related tests fail

MaceWindu (approved):
tests fixed. We were using different mapping schemas with same name in test and this pr broke this scenario, so I changed mapping schema name in test to be unique.

MaceWindu:
@sdanyliv , @Shane32 , @viceroypenguin are we good with this change? Need to decide ASAP, as I wan't to finalize next release

viceroypenguin (approved):

viceroypenguin:
I'm good with pulling it in.

sdanyliv (approved):

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1594,15 +1595,24 @@ public virtual void RollbackTransaction()
...
 		/// <summary>
 		/// Adds additional mapping schema to current connection.
 		/// </summary>
+		/// <remarks><see cref="DataConnection"/> will share <see cref="LinqToDB.Mapping.MappingSchema"/> instances that were created by combining same mapping schemas.</remarks>
 		/// <param name="mappingSchema">Mapping schema.</param>
 		/// <returns>Current connection object.</returns>
 		public DataConnection AddMappingSchema(MappingSchema mappingSchema)
 		{
-			MappingSchema = new MappingSchema(mappingSchema, MappingSchema);
+			var key = new { BaseSchema = MappingSchema.ConfigurationID, AddedSchema = mappingSchema.ConfigurationID };
```
Shane32, line 1609: This could be a value tuple... but C# does define `.Equals` for anonymous types as needed for this use

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1594,15 +1595,24 @@ public virtual void RollbackTransaction()
 		/// Gets list of query hints (writable collection), that will be used only for next query, executed through current connection.
 		/// </summary>
 		public  List<string>  NextQueryHints => _nextQueryHints ??= new List<string>();
+		
+		private static readonly MemoryCache _combinedSchemas = new MemoryCache(new MemoryCacheOptions());
```
Shane32, line 1599: I recommend that we always set the `SizeLimit` option when creating `MemoryCache` instances.  This will require the `Size` property to be set within `GetOrCreate` below.

Tests/Linq/Data/MiniProfilerTests.cs
```diff
@@ -1389,6 +1389,9 @@ public async Task TestPostgreSQL([IncludeDataSources(TestProvName.AllPostgreSQL)
 			var provider = new PostgreSQLDataProvider(version);
 			using (var db = CreateDataConnection(provider, context, type, "Npgsql.NpgsqlConnection, Npgsql"))
 			{
+				// needed for proper AllTypes columns mapping
+				db.AddMappingSchema(new MappingSchema(context));
```
Shane32, line 1393: I'm guessing that a new `MappingSchema` always creates a new `ConfigurationID`.  Assuming so, having a size limit prevents poorly written client code of the above nature (meaning, where they create a new `MappingSchema` for each created `DataConnection`) from adding an unbounded number of entries to the cache (limited only by the sliding expiration).

Source/LinqToDB/Data/DataConnection.cs
MaceWindu: we don't use value tuple in our code to avoid dll hell issue for  users of older .net framework versions

Source/LinqToDB/Data/DataConnection.cs
MaceWindu: If test baselines generation will fail again, I'll look into it, otherwise - we can do it in next release :smile:

PR #2514, opened 2020-09-25
DateTimeOffset fixes

### LINKED ISSUE #2508: Postgres Exception 'Column 'minute' does not exist' when querying DateTimeOffset field
The following statement raised an exception: Npgsql.PostgresException: '42703: column "minute" does not exist. There is no 'minute' column in the table.
Statement:
```cs
using (Postgres.Logging.DB objDB = new Postgres.Logging.DB("connectionstring"))
{
    var lstData = objDB.t_online.Where(y => y.update_date <= DateTimeOffset.Now.AddMinutes(-10)).ToList();
}
```
linq2db Version 3.1.2 got no problems with this.

Exception message:
Stack trace:
 Npgsql.PostgresException: '42703: column "minute" does not exist.

### Steps to reproduce

**PostgreSQL Table Definition:**
```sql
CREATE TABLE public.t_online
(
  online_id integer NOT NULL DEFAULT nextval('t_online_online_id_seq'::regclass),
  application_id integer NOT NULL,
  user_id integer NOT NULL,
  workstation_id integer NOT NULL,
  loglevel_id integer NOT NULL,
  update_date timestamp with time zone NOT NULL DEFAULT now(),
  CONSTRAINT t_online_online_id_pk PRIMARY KEY (online_id),
  )
WITH (
  OIDS=FALSE
);
```

**T4 generated code:**
```cs
        public ITable<t_online>           t_online           { get { return this.GetTable<t_online>(); } }

        [Table(Schema="public", Name="t_online")]
	public partial class t_online
	{
		[PrimaryKey, Identity] public int            online_id      { get; set; } // integer
		[Column,     NotNull ] public int            application_id { get; set; } // integer
		[Column,     NotNull ] public int            user_id        { get; set; } // integer
		[Column,     NotNull ] public int            workstation_id { get; set; } // integer
		[Column,     NotNull ] public int            loglevel_id    { get; set; } // integer
		[Column,     NotNull ] public DateTimeOffset update_date    { get; set; } // timestamp (6) with time zone
	}
```
**This code generates the exception:**
```cs
using (Postgres.Logging.DB objDB = new Postgres.Logging.DB("connectionstring"))
{
    var lstData = objDB.t_online.Where(y => y.update_date <= DateTimeOffset.Now.AddMinutes(-10)).ToList();
}
```

### Environment details
linq2db version: 3.1.3
Database Server: Postgres 9.6
Database Provider: Npgsql.dll v4.14
Operating system: Win 10 2004
.NET Framework: 4.7

### LINKED ISSUE #2512: DateTimeOffset.Add regression in 3.1.3
@viceroypenguin , looks like #2382 broke some cases that worked before.
I have following code that worked before:
```cs
db.Table.Insert(() => new Record()
{
    Field = DateTimeOffset.Now.Add(timespanValue)
});
```
now I get following error
```
'Sql.CurrentTzTimestamp.Add(...)' cannot be converted to SQL.
```

Looks like before we used client-side parameter evaluation, now it tries to convert it to server-side code and fails for some reason.

viceroypenguin:
Fixes #2508 
Fixes #2512 (temporary hack for now)

- Adds DateDiff/DateAdd functions back to PostgreSQL
- Limit test cases in non-SqlServer contexts due to lack of TZ support


Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -607,7 +607,9 @@ public void SetInfo(MappingSchema mappingSchema)
 
 			#region DateTimeOffset
 
-			{ M(() => DateTimeOffset.Now                   ), N(() => L<DateTimeOffset>                      (()                              => Sql.CurrentTzTimestamp                                 )) },
+			// Disabled for now. See #2512 (https://github.com/linq2db/linq2db/issues/2512)
+			// { M(() => DateTimeOffset.Now                   ), N(() => L<DateTimeOffset>                      (()                              => Sql.CurrentTzTimestamp                                 )) },
```
sdanyliv, line 611: Strange, why it is not working?

MaceWindu: https://github.com/linq2db/linq2db/issues/2512#issuecomment-700336176

MaceWindu (approved):

MaceWindu:
Issue2508Test test fails for pgsql and sql server produce conflicting test baselines for new test (non-constant parameters?) https://dev.azure.com/linq2db/linq2db/_build/results?buildId=3529&view=results

viceroypenguin:
Yes. Hence why I need to do merge stuff. :)  I'll fix it up today.

sdanyliv:
@MaceWindu, i’ve checked, to support `DateTimeOffset.New.Add(timespan)` we have to support timespan addition. It is another big task across different data providers. Better to create new issue.

MaceWindu:
This is not my issue - I don't really care about Add(timespan) translation. My issue is that code, which worked before using client-side evaluation, now broken, because it fails to convert to SQL and doesn't try client evalualuation as fallback option.

sdanyliv:
it is also an issue with translator. We have to evaluate expression in such case.

viceroypenguin:
I think the root of #2512 is that I would expect that a naked `DateTimeOffset.Now` to be translated to the server side. However, that is not (currently) compatible with ignoring `DateTimeOffset.Add(TimeSpan)` being rendered client side. So for now, I've removed the server side translation, `Sql.DateTimeOffset` is available for forced evaluation of the server side current time.

I don't think we need to support `DateTimeOffset.Add(TimeSpan)` directly, as there is no good way to evaluate it server side in any version of SQL. What would be nice, however, is a way to mark `DateTimeOffset.Add(TimeSpan)` as a method that is un-translatable, so that if it is used in a context that can be converted to client side evaluation.

If this is possible, then `DateTimeOffset.Now.Add(TimeSpan)` will get converted client-side, but `DateTimeOffset.Now` will get executed server side. 

NB: this separation of behavior is also necessary on the `DateTime.Add(TimeSpan)` side. It's just that no one is relying on `DateTime.Now.Add(TimeSpan)` to work since it never has.

MaceWindu:
@viceroypenguin, now when we have baselines, please don't merge PRs before tests are done and baselines generated, if it is not urgent. (I will trigger generation on master)

PR #2522, opened 2020-10-01
Add missing field cloning for merge source queries

MaceWindu:
Fixes `Table 'LinqToDB.SqlQuery.SqlMergeSourceTable' not found` exceptions from merge queries with queryable source.


sdanyliv (approved):

MaceWindu:
@viceroypenguin I doubt it, as in your case it fails in linq parse phase, and my fix works with AST. But who knows. I spent 3 days to catch this bug :)

PR #2535, opened 2020-10-09
Moved SQL comment generation to SqlBuilder

nsarris:
Query runner built commands add SQL comments on the top, including provider details and any parameters. 

This breaks providers like the DB2i OleDb provider that don't support comments, or perhaps a provider that has different notation than double dash.

I moved the actual printing of the comments to the SQL Builder, so that each provider can override its behaviour.

This is part one of DB2i OleDb provider support.

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.QueryRunner.cs
+++ b/Source/LinqToDB/Data/DataConnection.QueryRunner.cs
@@ -70,18 +70,19 @@ namespace LinqToDB.Data
 
 				var sb = new StringBuilder();
 
-				sb.Append("-- ").Append(_dataConnection.ConfigurationString);
+				var comments = new StringBuilder();
+
+				comments.Append(_dataConnection.ConfigurationString);
 
 				if (_dataConnection.ConfigurationString != _dataConnection.DataProvider.Name)
-					sb.Append(' ').Append(_dataConnection.DataProvider.Name);
+					comments.Append(' ').Append(_dataConnection.DataProvider.Name);
 
 				if (_dataConnection.DataProvider.Name != sqlProvider.Name)
-					sb.Append(' ').Append(sqlProvider.Name);
+					comments.Append(' ').Append(sqlProvider.Name);
 
-				sb.AppendLine();
-
+				sqlProvider.PrintComments(sb, comments.ToString());
 				sqlProvider.PrintParameters(sb, _preparedQuery.Parameters);
-
+				
 				var isFirst = true;
 
 				foreach (var command in _preparedQuery.Commands)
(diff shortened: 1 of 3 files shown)


MaceWindu:
Why do you need it? `GetSqlText` used for logging and header/parameters section could contain invalid sql already. E.g. for parameter types without literal support

MaceWindu:
I don't mind to have comment generation support on sql builder (with `bool inline` parameter) as it could be useful later at least for hints generation on some databases, but I'm against removing provider information from trace data.
If you have a good reason to do it - this information should be moved to trace object properties.

nsarris:
Sorry if wasn't clear in my initial comment, I am trying to use an ADO provider that doesn't support comments in sql command statements. The query runner injects those comments at the beggining of the actual SQL command text, not in the trace data or trace infrastructure. 

If this wasn't intened then perhaps change the query runner code to redirect comments and parameter data to traces. Otherwise please provide a way for a provider to ommit inline comments in SQL commands.

MaceWindu:
I think you are doing something wrong, `GetSqlText` method used to produce sql for diagnostics (logging, query ToString calls), that's why it includedes comments with additional information and this output is not intended for execution by linq2db.
Actual sql that is passed to database is build by `DataConnection.QueryRunner.GetCommand` (we don't generate comments here).

So wether you:
- use `GetSqlText` in wrong place (but I don't think you even need to touch linq2db core classes for provider implementation)
- issue is not with comments

nsarris:
I wasn't aware GetSqlText was a trace/diagnostics method. I located the problem through some of the tests breasking, I guess there are tests that invoke GetSqlText and then try to execute the produced sql. 

I'll take a look at the tests that break and let you know.

MaceWindu:
FYI, ms access also doesn't support comments, so we should have notice invalid use of this method for executed sql generation already if it was in linq2db code.

nsarris:
I copied all the linq2db tests over, so the issue is with certain tests where you handle Access specifically:

```
sql = string.Join(Environment.NewLine, sql.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
					.Where(line => !line.StartsWith("-- Access")));
```

I guess I can simply handle the issue in the tests in the same manner. I'll take a look at it during the weekend and let you know.

nsarris:
Ok, so it was just tests that executed the sql command generated by GetSqlText. I stripped the comments off in the test like you do with access and all tests are passing now, so I don't need this merged.

However this PR itslef simply passes control of the comment generation to the SqlBuilder and doesn't disable trace data information (unless overriden in the builder). In fact if you take a look I simply followed the pattern of PrintParameters, which as also called in GetSqlText. So I leave it up to you to decide if it's usefull or not.

MaceWindu:
I prefer to do it later when we introduce comments generation support

PR #2536, opened 2020-10-09
Added spaces to specific places to support iseries oledb driver

nsarris:
The DB2i OleDb provider needs spaces in specific places, like after commas in fields and values, after SELECT in subqueries etc. I added those spaces in a way that they don't alter the printout of queryies.

This is part two of DB2i OleDb provider support.


MaceWindu:
:no_mouth: 
Could you add comments in those places so we will not remove spaces accidentally later

nsarris:
Sure thing, comments added

viceroypenguin (approved):

MaceWindu:
Don't merge yet. need to fix sql asserts for several tests

MaceWindu:
@nsarris are there any public documentation on this behavior?

nsarris:
Yes, it's here: https://www.ibm.com/support/pages/apar/SE57049

It refers to the .net provider generally, but it only happens when using the OleDb provider, it seems to work with the native provider and the ODBC provider.

nsarris:
No, I found it simply through trial and error. The provider obviously parses the text. It breaks if there is no space after commas in columns or values or datatypes (e.g. Decimal(10,10) fails). End of line characters don't seem to matter in that aspect.

It also needs spaces after subquery parenthesis and the wierdest of all, after the SELECT keyword in subqueries.

I could drop support for this provider all together and only support the ODBC provider, but they both have their quirks so I would prefer to give consumers both options.

MaceWindu:
It always amuse me how "strange" is code from big bois :)

nsarris:
> It always amuse me how "strange" is code from big bois :)

Don't get me started, IBM's integration tools are some of the worse pieces of code I ever had to work with. The worse part of it is error messages that make no sense at all...

sdanyliv:
I think we have to do that only for DB2 provider. I don’t like that we increase size of the SQL.
it can be done by virtual property in SQLBuilder.

MaceWindu:
We decided to implement more generic formatting options for brackets and commas: #2544

PR #2543, opened 2020-10-13
Remove bogus exception

### LINKED ISSUE #2534: Another LINQPad crash when connecting to Access 2003 mdb database
image
I'm trying to use the 'LINQ to DB' driver for the first time. My connection string is:
`Provider=Microsoft.ACE.OLEDB.16.0;Data Source=<<Path to mdb>>;User Id=Admin;Password=;`
I'm using this same connection string with 'OleDbConnection()' from a c# console app without any problem. 
The 'Test' button says the connection is successful.
The database has over 30 tables. 
Maybe 'LINQ to DB' is objecting to just one, or more, of the queries in this database, but as it has over 200, the question is 'which one?'.
I've Googled the error "Operation is not valid due to the current state of the object" but I don't understand whats going wrong. Maybe something to do with 'System.Data.Linq.ForeignKeyReferenceAlreadyHasValueException()' ?

MaceWindu:
Fix #2534

### DIFF
--- a/Source/LinqToDB/DataProvider/Access/AccessSchemaProviderBase.cs
+++ b/Source/LinqToDB/DataProvider/Access/AccessSchemaProviderBase.cs
@@ -32,7 +32,6 @@ namespace LinqToDB.DataProvider.Access
 			{
 				if (dataType.ToLower() == "text")
 					return length == 1 && !options.GenerateChar1AsString ? typeof(char) : typeof(string);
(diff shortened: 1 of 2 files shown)


MaceWindu:
TLDR: we don't know what type was used in place of exception, but we shouldn't throw exception in this place and let provider to use object as fallback type.

viceroypenguin (approved):

PR #2545, opened 2020-10-13
#2526 Changed default escaping options for Oracle

### LINKED ISSUE #2526: Change default escaping options for Oracle provider
Could you try and set
```cs
OracleTools.DontEscapeLowercaseIdentifiers = false;
```
option?
I wonder why we have it set to true by default.

_Originally posted by @MaceWindu in https://github.com/linq2db/linq2db.EntityFrameworkCore/issues/67#issuecomment-704788338_

edsdck:
Fix #2526 

Simply changed the default value to false and updated the summary.

Let me know if there's any problems with this as it's my first issue 😬


MaceWindu:
I suspect we will have some test failures due to incorrect mappings. let's see

edsdck:
> I suspect we will have some test failures due to incorrect mappings. let's see

All right, I'll have a look.

edsdck:
@MaceWindu 
I feel like I could just set `DontEscapeLowercaseIdentifiers` to be true for the tests so mapping would work again. If this solution would be good enough, could you let me know if there's any configuration for tests, so I could drop it there?

I feel like wasting a lot of time looking at the wrong places.

viceroypenguin:
@edsdck  @MaceWindu I think with the number of failed tests here, there's too high risk of being a breaking change for existing users. Might should wait until a future 4.0 version?

sdanyliv:
Don't worry, there is just problem in escaping trigger name. We will handle that. It is not critical change and should work in any cases.

MaceWindu:
I will handle it tomorrow. Already have recent experience with similar change for another db

edsdck:
> I will handle it tomorrow. Already have recent experience with similar change for another db

All right, I'll have a look afterwards

MaceWindu:
just change property value thay said..

- updated oracle.sql test script to use escaping for most of db objects we use
- added missing identity trigger and sequence name escaping in sql builder
- added fallback to SQL-based bulk copy from native bulk copy, if target table columns require escaping. Such tables cannot be used with ODP.NET bulk copy due to provider bug ODP.NET
- fixed `OracleXmlTable` to esape column names when needed

MaceWindu (approved):

viceroypenguin (approved):

PR #2548, opened 2020-10-14
Issue/2547

jogibear9988:


sdanyliv:
I will stop pipeline. It is not a fix but avoids others to run tests.

jogibear9988:
Icould not test against a sql server.
Hopfully my tests works and fails

jogibear9988:
@sdanyliv seems the test did work.
do we already have fixed this, or does my test not reproduce the error?

sdanyliv:
@jogibear9988, test is incomplete. I do not see what to fix.

jogibear9988:
closed at the moment, cause it does not reproduce the issue

PR #2565, opened 2020-10-20
Implement DateDiff for Oracle 

### LINKED ISSUE #2564: Oracle - Exception running Query:    ORA-00904: "DATEDIFF": ungültiger Bezeichner
I've the following query:

                var from = DateTime.UtcNow.AddDays(-1);
                var to = DateTime.UtcNow;
                var qry = (from m in s.GetTable<MessageEventDTO>()
                            where m.TimestampGone.HasValue &&
                             m.TimestampGenerated >= @from &&
                             m.TimestampGenerated <= to &&
                             m.MessageClassName == "Error"
                            group m by new { m.ExternID1, m.TranslatedMessageGroup, m.TimestampGenerated.Hour } into tgGroup
                            select new 
                            {
                                MessageText = tgGroup.Min(x => x.TranslatedMessage1).Trim(),
                                GroupName = tgGroup.Key.TranslatedMessageGroup,
                                Hour = tgGroup.Key.Hour,
                                Count = tgGroup.Count(),
                                DurationInSeconds = (long)tgGroup.Sum(x => (x.TimestampGone.Value - x.TimestampGenerated).TotalMilliseconds),
                            });
                var data = qry.ToList();

I get the following Exception:

	Message: 
		Oracle.ManagedDataAccess.Client.OracleException : ORA-00904: "DATEDIFF": ungültiger Bezeichner
	  Stack Trace: 
		OracleConnectionImpl.VerifyExecution(Int32& cursorId, Boolean bThrowArrayBindRelatedErrors, SqlStatementType sqlStatementType, Int32 arrayBindCount, OracleException& exceptionForArrayBindDML, Boolean& hasMoreRowsInDB, Boolean bFirstIterationDone)
		OracleCommandImpl.VerifyExecution(OracleConnectionImpl connectionImpl, Int32& cursorId, Boolean bThrowArrayBindRelatedErrors, OracleException& exceptionForArrayBindDML, Boolean& hasMoreRowsInDB, Boolean bFirstIterationDone)
		OracleCommandImpl.ExecuteReader(String commandText, OracleParameterCollection paramColl, CommandType commandType, OracleConnectionImpl connectionImpl, OracleDataReaderImpl& rdrImpl, Int32 longFetchSize, Int64 clientInitialLOBFS, OracleDependencyImpl orclDependencyImpl, Int64[] scnForExecution, Int64[]& scnFromExecution, OracleParameterCollection& bindByPositionParamColl, Boolean& bBindParamPresent, Int64& internalInitialLOBFS, OracleException& exceptionForArrayBindDML, OracleConnection connection, OracleLogicalTransaction& oracleLogicalTransaction, IEnumerable`1 adrianParsedStmt, Boolean isDescribeOnly, Boolean isFromEF)
		OracleCommand.ExecuteReader(Boolean requery, Boolean fillRequest, CommandBehavior behavior)
		OracleCommand.ExecuteDbDataReader(CommandBehavior behavior)
		DbCommand.ExecuteReader(CommandBehavior behavior)
		DbCommandProcessorExtensions.ExecuteReaderExt(IDbCommand cmd, CommandBehavior commandBehavior)
		<>c__DisplayClass29_0.<ExecuteDbDataReader>b__0()
		RetryPolicyBase.ExecuteImplementation[TResult](Func`1 operation)
		RetryPolicyBase.Execute[TResult](Func`1 operation)
		RetryingDbCommand.ExecuteDbDataReader(CommandBehavior behavior)
		DbCommand.ExecuteReader(CommandBehavior behavior)
		DbCommandProcessorExtensions.ExecuteReaderExt(IDbCommand cmd, CommandBehavior commandBehavior)
		DataConnection.ExecuteReader(IDbCommand command, CommandBehavior commandBehavior)
		DataConnection.ExecuteReader(CommandBehavior commandBehavior)
		DataConnection.ExecuteReader()
		QueryRunner.ExecuteReader()
		<ExecuteQuery>d__12`1.MoveNext()
		List`1.ctor(IEnumerable`1 collection)
		Enumerable.ToList[TSource](IEnumerable`1 source)
		DatabaseReadTests.ReadComplexOracleTest() line 151

jogibear9988:
Fixes #2564


viceroypenguin (approved):
LGTM. Check w/ @MaceWindu or @sdanyliv for final approval though

viceroypenguin:
@jogibear9988  Looks like Oracle treats the difference between two datetime values as a timespan of some sort. You'll need to look at how to convert it on the Oracle side into the individual parts rather than doing the multiply/divide stuff you have right now.

jogibear9988:
@viceroypenguin do I need to do something that the builds do not fail? it sais something about rebase, but i did that

viceroypenguin:
@jogibear9988 looks like the failures in the last two tests are things @MaceWindu needs to fix since they're part of how the baselines project gets merged. I'll let him review and fix.

viceroypenguin (approved):

Source/LinqToDB/Sql/Sql.DateTime.cs
```diff
@@ -892,12 +892,36 @@ public void Build(ISqExtensionBuilder builder)
...
+				var expStr = part switch
+				{
+					// DateParts.Year        => "({1} - {0}) / 365",
+					// DateParts.Month       => "({1} - {0}) / 30",
+					DateParts.Week        => "(CAST ({1} as DATE) - CAST ({0} as DATE)) / 7",
+					DateParts.Day         => "(CAST ({1} as DATE) - CAST ({0} as DATE))",
+					DateParts.Hour        => "(CAST ({1} as DATE) - CAST ({0} as DATE)) * 24",
+					DateParts.Minute      => "(CAST ({1} as DATE) - CAST ({0} as DATE)) * 1440",
+					DateParts.Second      => "(CAST ({1} as DATE) - CAST ({0} as DATE)) * 86400",
+					DateParts.Millisecond => "(CAST ({1} as DATE) - CAST ({0} as DATE)) * 86400000",
```
MaceWindu, line 911: this doesn't look right. ranges less that day rounded to 1 day?

MaceWindu (changes requested):
I don't see detailed tests for new diff implementation. We already have datediff tests so it makes sense to enable them for oracle.

MaceWindu:
Ah, didn't noticed you enabled them. let me check once more

Source/LinqToDB/Sql/Sql.DateTime.cs
MaceWindu, line 911: Why we use date here? It doesn't work with less that second parts. I suggest to change it to `TIMESPAN` and add corresponding datepart tests as I don't see it in our `DateTimeFunctionsTests`

MaceWindu (changes requested):

MaceWindu:
Ok, this works, but we need to fix milliseconds support

Source/LinqToDB/Sql/Sql.DateTime.cs
jogibear9988: Don't know what I should do now.
If I cast to timestamp it does not work.
I'm not an oracle expert.

jogibear9988:
found this:
http://www.dba-oracle.com/t_timestamp_math_elapsed_times.htm

uses this code

     (extract(hour from tim_col2)-extract(hour from tim_col1))*3600+
       (extract(minute from tim_col2)-extract(minute from tim_col1))*60+
      * extract(second from tim_col2)-extract(second from tim_col1))*1000 ms from check_time

should we switch to this? or make it optional?
it make's not really sense to use this code if DataType is date

jogibear9988:
it's also missing day.
so if we use this, the sql get's much complexer (and maybe slower)

Source/LinqToDB/Sql/Sql.DateTime.cs
MaceWindu: it works (in sql), but produce interval instead of number of days (https://docs.oracle.com/database/121/SQLRF/sql_elements001.htm#g196492), so it timestamp used, additional sql required to convert interval value to number

jogibear9988: yes, I think then we need code like this https://github.com/linq2db/linq2db/pull/2565#issuecomment-716196525
as dates are not with milliseconds, maybe it's possible to leave the code like this.
if someone needs ms support, maybe then he could work on this

jogibear9988: i've no oracle system atm, so i could only test on customer system, and so it takes forever

jogibear9988:
@MaceWindu what about this now? could we merge this?
we could create a extra ticket for oracle millisecond support with datetime types

MaceWindu:
I would prefer to have milliseconds support (we could use other sql only for them). Don't like incomplete features

jogibear9988:
me too, but I‘ve no oracle to work at the moment (and no time). at earliest i could work on this in february next year. but i use a fork of linq2db at the moment cause i need that support, what I don‘t like.
ant it works better with the patch then before, also before there was no error that it would not work, there is only incorrect sql.

jogibear9988:
maybe we create an extra issue oracle millisecond support, and assign to me. and i look to fix at start of next year

MaceWindu:
I will check it a bit later today. Shouldn't be a lot of work

MaceWindu:
ok, oracle support added, also fixed overflow for db2 on large intervals

viceroypenguin (approved):

PR #2577, opened 2020-10-25
Issue 2168 - Modify T4 name conflict detection for data types to include table names instead of only column names

### LINKED ISSUE #2168: Wrong T4 generation for ColumnAttribute when table name is "DataType"
linq2db version: 2.9.7 (and earlier, I guess)

When there's a **table** named "DataType" the T4 generation creates wrong code for the ColumnAttribute attribute because _DataType_ is not fully qualified as _LinqToDB.DataType_.

Generated wrong code:
`[Column(DbType="uniqueidentifier", DataType=DataType.Guid)    , PrimaryKey, NotNull]`
Compiler error: 'DataType' does not contain a definition for 'Guid'

Right code should be:
`[Column(DbType="uniqueidentifier", DataType=LinqToDB.DataType.Guid)    , PrimaryKey, NotNull]`

There's already a fix in _LinqToDB.ttinclude_, line 283 where there's a check if one of the columns is named "DataType".
`var dataTypePrefix = t.Columns.Values.Any  (c => c.MemberName == "DataType") ? "LinqToDB." : ""`

I could fix the problem by changing that line to 
`var dataTypePrefix = t.Columns.Values.Any  (c => c.MemberName == "DataType") || Tables.Values.Any(tbl => tbl.Name == "DataType") ? "LinqToDB." : "";`

Proposal: Please integrate this (or a similar fix).

dipique:
Fixes #2168 

The issue is that `DataType` references in attributes would sometimes create ambiguous reference errors if there is a table named `DataType`. This ambiguous reference is already detected and prevented for field names; this fix expands that detection to include table names as well. When the condition is detected, `DataType` is fully qualified to `LinqToDB.DataType`.

### DIFF
--- a/Source/LinqToDB.Templates/LinqToDB.ttinclude
+++ b/Source/LinqToDB.Templates/LinqToDB.ttinclude
@@ -274,7 +274,7 @@ void GenerateTypesFromMetadata()
 		var dbTypeMaxLen   = t.Columns.Values.Max  (c => (int?)(c.ColumnType.Length)) ?? 0;
 		var dataTypeMaxLen = t.Columns.Values.Where(c => c.DataType != null).Max  (c => (int?)(c.DataType.Length)) ?? 0;
 		var nameConflict   = t.Columns.Values.Any  (c => c.MemberName == "DataType") ||
-								Tables.Values.Any  (tbl => tbl.Name   == "DataType");
+		                        Tables.Values.Any  (tbl => tbl.Name   == "DataType");
 		var dataTypePrefix = nameConflict ? "LinqToDB." : "";
 
 		foreach (var c in t.Columns.Values)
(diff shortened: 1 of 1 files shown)


Source/LinqToDB.Templates/LinqToDB.ttinclude
```diff
@@ -273,7 +273,9 @@ void GenerateTypesFromMetadata()
 			: ToStringLiteral(c.ColumnName).Length)) ?? 0;
 		var dbTypeMaxLen   = t.Columns.Values.Max  (c => (int?)(c.ColumnType.Length)) ?? 0;
 		var dataTypeMaxLen = t.Columns.Values.Where(c => c.DataType != null).Max  (c => (int?)(c.DataType.Length)) ?? 0;
-		var dataTypePrefix = t.Columns.Values.Any  (c => c.MemberName == "DataType") ? "LinqToDB." : "";
+		var nameConflict   = t.Columns.Values.Any  (c => c.MemberName == "DataType") ||
+		                        Tables.Values.Any  (tbl => tbl.Name   == "DataType");
```
MaceWindu, line 277: shouldn't it be `tbl.TypeName` here?

dipique: It definitely should. Updated.

MaceWindu (approved):

PR #2589, opened 2020-11-03
Add QueryAsync extension methods.

perfectsquircle:
I may be missing something, but DataConnectionExtensions seems to missing the Async counterparts to `Query<T>(...)` methods. I was a little confused when they didn't show up in intellisense.

image

Let me know if there is a better way to access QueryAsync, or if they're omitted intentionally.

### DIFF
--- a/Source/LinqToDB/Data/DataConnectionExtensions.cs
+++ b/Source/LinqToDB/Data/DataConnectionExtensions.cs
@@ -1017,6 +1017,69 @@ namespace LinqToDB.Data
 		#region Query async
 
 		/// <summary>
+		/// Executes command asynchronously and returns results as collection of values of specified type.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="sql">Command text.</param>
+		/// <param name="cancellationToken">Asynchronous operation cancellation token.</param>
+		/// <returns>Returns a task with a collection of query result records.</returns>
+		public static Task<IEnumerable<T>> QueryAsync<T>(this DataConnection connection, string sql, CancellationToken cancellationToken = default)
+		{
+			return new CommandInfo(connection, sql).QueryAsync<T>(cancellationToken);
+		}
+
+		/// <summary>
+		/// Executes command asynchronously and returns results as collection of values of specified type.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="sql">Command text.</param>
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2589) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2589) before we can accept your contribution.<br/><br/>:x: perfectsquircle [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2589)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=2589) it.</sub>

MaceWindu (changes requested):
Thanks. Please add CancellationToken parameter to new methods.

perfectsquircle:
Ok, updated the PR. Let me know if there's anywhere I should be adding test coverage or documentation.

MaceWindu:
Looked at those extensions closely and see that it's a bit messy:
-  we already have async versions of those methods in form of `QueryToListAsync` and `QueryToArrayAsync`
- those methods have two versions - with token and without token

@sdanyliv any good reason we implemented it like that?

@perfectsquircle 
if you don't mind - could you add async version for remaining 5 Query methods (9 in total, you added QueryAsync for 4 of them)

Regarding tests - it should be easy to add async version of test using test for sync version but I don't see even sync version of test for most of them, only 3 of 9 methods are tested in DataExtensionTests.cs using following tests TestScalar1, TestScalar2, TestObject2. We would appreciate if you can add both sync and async tests for remaining `Query` methods too.

sdanyliv:
@MaceWindu, maybe it is better to return `IAsyncEnumerable` for these methods?

perfectsquircle:
@MaceWindu if you feel like QueryAsync would be redundant with QueryToListAsync and QueryToArrayAsync, that's fine. I'd rather use existing methods than spread the public interface even wider, and making DataConnectionExtensions even bigger.

I can also just do

```csharp
dataConnection.SetCommand("select 123").QueryAsync<int>()
```

MaceWindu:
It is not like I think it will be redundant, but I agree that we should postpone it for now. You unintentionally rised issue with our async API surface - it is not systematic. Earliest versions of async API provide ToList/ToArray methods, later we started to use Task<IEnumerable> but probably IAsyncEnumerable is correct answer here. Also cancellation token use differ between methods.

We should take timeout to think and decide on async API guidelines. After that we will review/refactor existing API.

PR #2591, opened 2020-11-04
Test runner support

igor-tkachev:
This is a technical issue to support ParallelTestRunner and should affect test projects only.

Main changes:

- Default database configuration is SQLite.Default. This database has to be excluded from any tests that updates data.
- Regular LinqService is replaced with TestLinqServiceClient that does not use TCP/IP for testing.


Tests/Base/TestBase.cs
```diff
@@ -250,6 +253,61 @@ static TestBase()
...
+
+								break;
+							}
+						}
+					}
+				}
+				catch (Exception e)
+				{
+					TestExternals.Log(e.ToString());
+					Console.WriteLine(e);
```
MaceWindu, line 291: Don't use `Console` in tests

Tests/Base/TestBase.cs
```diff
@@ -250,6 +253,61 @@ static TestBase()
...
+			{
+				if (Directory.Exists(dataPath))
+					Directory.Delete(dataPath, true);
+
+				Directory.CreateDirectory(dataPath);
+
+				foreach (var file in Directory.GetFiles(databasePath, "*.*"))
+				{
+					var destination = Path.Combine(dataPath, Path.GetFileName(file));
+					Console.WriteLine("{0} => {1}", file, destination);
```
MaceWindu, line 305: `Console` call

Tests/Base/TestBase.cs
```diff
@@ -273,10 +331,10 @@ static TestBase()
 		}
 
 #if NET472
-		const  int          IP        = 22654;
-		static bool         _isHostOpen;
-		static LinqService? _service;
-		static object       _syncRoot = new object();
+		const           int          IP = 22654;
```
MaceWindu, line 334: pr summary says we don't use network connections for tests. is it still needed?

Tests/Base/TestBase.cs
```diff
@@ -374,22 +435,25 @@ static void OpenHost(MappingSchema? ms)
...
 		{
 			if (configuration.EndsWith(".LinqService"))
 			{
 #if NET472
 				OpenHost(ms);
 
 				var str = configuration.Substring(0, configuration.Length - ".LinqService".Length);
-				var dx  = new TestServiceModelDataContext(IP) { Configuration = str };
+
+				var dx  = testLinqService
```
MaceWindu, line 447: Any value in using old tcp/ip context?

Tests/Base/TestBase.cs
```diff
@@ -1505,6 +1567,28 @@ public void Dispose()
 		}
 	}
 
+	public static class TestExternals
```
MaceWindu, line 1570: Maybe it makse sense to add single logging method to TestBase, which will write logs to TestContex/FileLog based in configuration?

Tests/Base/Tests.Base.csproj
```diff
@@ -11,6 +11,7 @@
 
 		<PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
 		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.3" />
+		<PackageReference Include="FirebirdSql.Data.FirebirdClient" Version="7.5.0" />
```
MaceWindu, line 14: Why do we need to call ClearPool if we call to ClearAllPools already?

Tests/Linq/Linq/DateTimeFunctionsTests.cs
```diff
@@ -843,11 +843,11 @@ public static DateTime Truncate(DateTime date, long resolution)
 
 		[Test]
 		public void AddDynamicFromColumn(
-			[DataSources(TestProvName.AllInformix)] string context, 
+			[DataSources(TestProvName.AllInformix)] string context,
 			[Values(
-				Sql.DateParts.Day,
```
MaceWindu, line 848: endless war between @sdanyliv 's spaces and you :smile: Probably we should add some analyzer for it

Tests/Linq/RunnerSupport.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+	public static class RunnerSupport
+	{
+		public static string? GetConfiguration(ITest test)
+		{
+			if (!test.HasChildren && test.Method is {} method)
+			{
+				var parameters = method.GetParameters();
+
+				if (test.Arguments.Length == parameters.Length)
```
MaceWindu, line 15: I recently added helper to get context from test. Check NUnitUtils.GetContext

Tests/linq2db.testproj
```diff
@@ -0,0 +1,14 @@
+{
+  "Assemblies": [
```
MaceWindu, line 2: what is this file?

MaceWindu (approved):
some minor issues. Where is test-runnner itself?

PR #2599, opened 2020-11-06
remove UseBinaryAggregateExpression setting

### LINKED ISSUE #2593: Set default value of LinqToDB.Common.Configuration.Linq.UseBinaryAggregateExpression to true
This option is very usefull for applications with small thread stack size like web-application, where query could easily fail with big but not huge condition without it.

MaceWindu:
Fix #2593

### DIFF
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -159,13 +159,6 @@ namespace LinqToDB.Common
 			public static bool GuardGrouping = true;
 
 			/// <summary>
-			/// Used to optimize huge logical operations with large number of operands like expr1.and.expr2...and.exprN into balanced tree.
-			/// Without this option, such conditions could lead to <seealso cref="StackOverflowException"/>.
-			/// Default value: <c>false</c>.
-			/// </summary>
-			public static bool UseBinaryAggregateExpression;
-
-			/// <summary>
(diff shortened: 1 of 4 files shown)


viceroypenguin:
The Issue suggest changing the default, this removes the option entirely.
1. Is this a breaking change (due to the removal of the property)? Could we leave the property but mark it deprecated with a note that it doesn't do anything, so that it's not a breaking change?
2. Based on the change here, I'm assuming we don't have any desire to allow the possibility to set to false anymore? Is the concern that originated the option in the first place no longer present?

Code LGTM otherwise.

MaceWindu:
History behind this setting is folllowing:
- initially it was implemented using custom `Expression` element with option to enable it
- later it was rewritten to another approach where we flatten expression tree to decrease level of nesting without use of custom `Exresssion` element

So now we have:
- option name doesn't match implementation for a long time
- it is disabled for no good reason by default

I decided to remove option completely as:
- it doesn't make sense anymore to have option for one of many steps in expression parsing pipeline
- it doesn't make sense to disable it (only if you want less performant parsing of big conditions and occasional stack overflow errors)
- as it was disabled by default, not many people used it and removing option will be breaking change for small amount of people, who just need to remove 1 line from their code

viceroypenguin (approved):

PR #2601, opened 2020-11-06
Non-table column excluded from sub-query group by (3.1.3 regression)

MaceWindu:
Added test to reproduce issue

```
Exception: Microsoft.Data.SqlClient.SqlException
    Message  : Column 't1.c1' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.
```

```sql
SELECT
    	Count(*)
    FROM
    	(
    		SELECT
    			[t1].[c1] as [Key_1],
    			Count(*) as [Count_1]
    		FROM
    			(
    				SELECT
    					1 as [c1]
    				FROM
    					[Person] [_]
    				UNION ALL
    				SELECT
    					2 as [c1]
    				FROM
    					[Person] [_1]
    			) [t1]
-- GROUP BY WAS HERE
    	) [_2]
    WHERE
    	[_2].[Count_1] > 1 AND [_2].[Key_1] = 1
    ```


MaceWindu:
:-/ looks group by like optimization should be disabled for informix

PR #2604, opened 2020-11-06
Add DateDiff function for Access

### LINKED ISSUE #2598: Sql.DateDiff returns exception 'No value given for one or more required parameters.'
Hi,

Below raises `System.Data.OleDb.OleDbException: 'No value given for one or more required parameters.'`.

```c#
private class TestClass
{
    public DateTime Date { get; set; }
}

using (var temptable = new TempTable<TestClass>(DbContext))
{
    var query = from d in temptable
                select Sql.DateDiff(Sql.DateParts.Day, d.Date, Sql.CurrentTimestamp);

    var result = query.ToArray();
}
```
```SQL
SELECT
    DateDiff(day, [d].[Date], Now)
FROM
    [TestClass] [d]
```

The problem is that `Sql.DateParts.Day` gives `day` (no quotes) and should give `'d'` (in single quote). See here: [link](https://www.w3schools.com/sql/func_msaccess_datediff.asp)

### Environment details
linq2db version: 3.1.6
Database Server: MsAccess
Database Provider: OleDb
Operating system: Win10
.NET Framework: 4.7.2

yarecky1:
Fixes #2598

### DIFF
--- a/Source/LinqToDB/Sql/Sql.DateTime.cs
+++ b/Source/LinqToDB/Sql/Sql.DateTime.cs
@@ -900,25 +900,25 @@ namespace LinqToDB
 				var startDate = builder.GetExpression(1);
 				var endDate = builder.GetExpression(2);
 
-				var expStr = "DATEDIFF(";
+				var expStr = "DATEDIFF('";
 
 				expStr += part switch
 				{
-					DateParts.Year        => "'yyyy'",
-					DateParts.Quarter     => "'q'",
-					DateParts.Month       => "'m'",
-					DateParts.DayOfYear   => "'y'",
-					DateParts.Day         => "'d'",
-					DateParts.WeekDay     => "'w'",
-					DateParts.Week        => "'ww'",
-					DateParts.Hour        => "'h'",
-					DateParts.Minute      => "'n'",
-					DateParts.Second      => "'s'",
-					DateParts.Millisecond => "'s'",
+					DateParts.Year        => "yyyy",
+					DateParts.Quarter     => "q",
+					DateParts.Month       => "m",
+					DateParts.DayOfYear   => "y",
+					DateParts.Day         => "d",
+					DateParts.WeekDay     => "w",
+					DateParts.Week        => "ww",
+					DateParts.Hour        => "h",
+					DateParts.Minute      => "n",
+					DateParts.Second      => "s",
+					DateParts.Millisecond => "s",
 					_                     => throw new ArgumentOutOfRangeException(),
 				};
 
-				expStr += ", {0}, {1})";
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2604) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2604) before we can accept your contribution.<br/><br/>:x: yarecky1 [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2604)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=2604) it.</sub>

yarecky1:
@sdanyliv OK, Finished.

sdanyliv:
Enable tests for Access. Just remove exclusion.
https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Linq/DateTimeFunctionsTests.cs#L1207

Source/LinqToDB/Sql/Sql.DateTime.cs
```diff
@@ -892,13 +892,49 @@ public void Build(ISqExtensionBuilder builder)
...
+					DateParts.Quarter     => "q",
+					DateParts.Month       => "m",
+					DateParts.DayOfYear   => "y",
+					DateParts.Day         => "d",
+					DateParts.WeekDay     => "w",
+					DateParts.Week        => "ww",
+					DateParts.Hour        => "h",
+					DateParts.Minute      => "n",
+					DateParts.Second      => "s",
+					DateParts.Millisecond => "s",
```
MaceWindu, line 917: as DateDiff for Access doesn't support milliseconds (and access in general), I think we should throw exception for this qualifier (user always could use seconds directly)

MaceWindu (changes requested):

jogibear9988:
I don‘t like that we throw on ms. Cause then you need to maintain different code for different databases, if you like ms precission where it‘s available

viceroypenguin:
@jogibear9988 Honestly, I would argue that the exception is the proper response. If I am using l2db with one database engine, and then I switch to another to another engine, I would expect that l2db would hide variances where possible and make it visible that the difference is insurmountable. If Access does not support ms, and I am unaware of the nuance of Access, I would want to know up front that Access doesn't support ms, instead of getting a result that is possibly incorrect based on how my code is written.

MaceWindu (approved):

PR #2606, opened 2020-11-07
Fix some of the Sql functions + add tests for non-db part

### LINKED ISSUE #2605: Bug in Sql.Function Left?
Hi,

Just reading the code and found this:
```c#
[Sql.Function(                      PreferServerSide = true)]
[Sql.Function(PN.SQLite, "LeftStr", PreferServerSide = true)]
public static string? Left(string? str, int? length)
{
    return length == null || str == null || str.Length < length? null: str.Substring(1, length.Value);
}
```
Why startIndex = 1 in `str.Substring(1, length.Value)`?
Shouldn't be there startIndex = 0?
(startIndex = The zero-based starting character position of a substring in this instance.)

yarecky1:
Fixes #2605

### DIFF
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -510,8 +510,7 @@ namespace LinqToDB
 		[Sql.Function]
 		public static string? Reverse(string? str)
 		{
-			if (string.IsNullOrEmpty(str))
-				return str;
+			if (string.IsNullOrEmpty(str)) return str;
 
 			var chars = str!.ToCharArray();
 			Array.Reverse(chars);
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -589,18 +589,21 @@ namespace LinqToDB
 		[Sql.Function(PN.Sybase, "Str_Replace")]
 		public static string? Replace(string? str, string? oldValue, string? newValue)
 		{
-			return str == null || oldValue == null || newValue == null ?
-				null :
-				str.Replace(oldValue, newValue);
+			if (str == null || oldValue == null || newValue == null) return null;
+			if (str.Length == 0)                                     return str;
+			if (oldValue.Length == 0)                                return str; // Replace raises exception here.
+
+			return str.Replace(oldValue, newValue);
 		}
 
 		[Sql.Function]
 		[Sql.Function(PN.Sybase, "Str_Replace")]
 		public static string? Replace(string? str, char? oldValue, char? newValue)
 		{
-			return str == null || oldValue == null || newValue == null ?
-				null :
-				str.Replace(oldValue.Value, newValue.Value);
+			if (str == null || oldValue == null || newValue == null) return null;
+			if (str.Length == 0)                                     return str;
+
+			return str.Replace(oldValue.Value, newValue.Value);
 		}
 
 		[Sql.Function]
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -525,7 +525,7 @@ namespace LinqToDB
 			if (length.Value < 0)              return null;
 			if (length.Value > str.Length)     return str;
 
-			return str.Substring(0, length.Value);
+			return str[..length.Value];
 		}
 
 		[Sql.Function(                       PreferServerSide = true)]
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -566,19 +566,23 @@ namespace LinqToDB
 		}
 
 		[Sql.Function(Name = "LPad")]
-		public static string? PadLeft(string? str, int? totalWidth, char? paddingChar)
+		public static string? PadLeft(string? str, int? length, char? paddingChar)
 		{
-			return str == null || totalWidth == null || paddingChar == null ?
-				null :
-				str.PadLeft(totalWidth.Value, paddingChar.Value);
+			if (str == null || length == null || paddingChar == null) return null;
+			if (length.Value < 0)                                     return null;
+			if (length.Value <= str.Length)                           return str.Substring(0, length.Value);
+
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Sql/Sql.cs
```diff
@@ -508,7 +508,7 @@ public static bool Like(string? matchExpression, string? pattern, char? escapeCh
 		[Sql.Function(PN.SQLite, "LeftStr", PreferServerSide = true)]
 		public static string? Left(string? str, int? length)
 		{
-			return length == null || str == null || str.Length < length? null: str.Substring(1, length.Value);
+			return length == null || str == null || str.Length < length? null: str.Substring(0, length.Value);
```
viceroypenguin, line 511: I'm not sure that we should leave it such that `str.Length < length` means the return value is `null`. I think the result for that condition should be `str`. @MaceWindu please confirm?

MaceWindu: Yes, if we mimic db behavior, we shold return str here

yarecky1: I'll correct this, similar functions and tests as you propose.

yarecky1:
@MaceWindu @viceroypenguin 
I will change
```c#
public static string? Substring(string? str, int? startIndex, int? length)
{
    return str == null || startIndex == null || length == null ?
           null :
	   str.Substring(startIndex.Value, length.Value);
}
```
to
```c#
public static string? Substring(string? str, int? startIndex, int? length)
{
    return str == null || startIndex == null || startIndex.Value < 1 || length == null ?
           null :
	   str.Substring(startIndex.Value - 1, length.Value);
}
```
Added: `startIndex.Value < 1` condition and `-1` to parameter `startIndex.Value`.

This is because database uses 1-based position parameter, not 0-based index.

Which also means, we should rename parameter from `startIndex` to just `start` to not mislead users. See [here](https://www.w3schools.com/sql/func_sqlserver_substring.asp). This will change 6 files in 24 places.

Do you agree?

MaceWindu:
Agree

viceroypenguin:
Agree

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -508,24 +521,35 @@ public static bool Like(string? matchExpression, string? pattern, char? escapeCh
...
 		public static string? Right(string? str, int? length)
 		{
-			return length == null || str == null || str.Length < length?
-				null :
-				str.Substring(str.Length - length.Value);
+			if (length == null || str == null) return null;
+			if (length.Value < 0)              return null;
+			if (length.Value > str.Length)     return str;
+
+			return str.Substring(str.Length - length.Value);
```
viceroypenguin, line 539:
```suggestion
			return str[^length.Value..];
```

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -508,24 +521,35 @@ public static bool Like(string? matchExpression, string? pattern, char? escapeCh
 		[Sql.Function(PN.SQLite, "LeftStr", PreferServerSide = true)]
 		public static string? Left(string? str, int? length)
 		{
-			return length == null || str == null || str.Length < length? null: str.Substring(1, length.Value);
+			if (length == null || str == null) return null;
+			if (length.Value < 0)              return null;
+			if (length.Value > str.Length)     return str;
+
+			return str.Substring(0, length.Value);
```
viceroypenguin, line 528:
```suggestion
			return str[..length.Value];
```

Source/LinqToDB/Sql/Sql.cs
yarecky1: I'm not convinced of introducing new syntax but will not discuss :)

Source/LinqToDB/Sql/Sql.cs
yarecky1: I'm not convinced of introducing new syntax but will not discuss :)

Source/LinqToDB/Sql/Sql.cs
viceroypenguin: Ah. Forgot that it requires extra types in earlier FX versions. :)

yarecky1:
@MaceWindu OK. This is it. No more string functions to fix or worth to create tests.

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -415,9 +415,16 @@ public static TTo From<TFrom>(TFrom obj)
 		[Sql.Function  (PN.SQLite,   "Substr",                          PreferServerSide = true)]
 		[Sql.Expression(PN.Firebird, "Substring({0} from {1} for {2})", PreferServerSide = true)]
 		[Sql.Function  (PN.SapHana,  "Substring",                       PreferServerSide = true)]
-		public static string? Substring(string? str, int? startIndex, int? length)
+		public static string? Substring(string? str, int? start, int? length)
 		{
-			return str == null || startIndex == null || length == null ? null : str.Substring(startIndex.Value - 1, length.Value);
+			if (str == null || start == null || length == null) return null;
+			if (start.Value < 1 || start.Value > str.Length) return null;
```
MaceWindu, line 421: that's is interesting case. e.g. various dbs behave differently. E.g. sqlserver doesn't return null for those cases it returns:
- empty string
- or part of string if zero or negative start used and start + length > 1

Let's accept this behavior for out-of-range parameters as it is not possible to emulate all dbs for it. Basically idea is: for valid parameters we implement expected behavior, for invalid parameters - behavior could differ from db behavior...

MaceWindu (approved):

Source/LinqToDB/Sql/Sql.cs
yarecky1: Ok, so let's keep it that way. One change here for parameter out of range will force me to change behaviour in other places. Some function doesn't exists in sqlserver so I had to check what happens then. I preferred to return null for all abnormal parameters unless I could confirm that function would return empty string.

yarecky1:
Could somebody re-run azp?

MaceWindu (approved):

PR #2613, opened 2020-11-10
Fix for #2588. Removed unnecessary null result check.

### LINKED ISSUE #2588: Queryable.DefaultIfEmpty not working?
Hi,
I can't get Max value when collection is null.

```c#
public class TestClass
{
    public int Id { get; set; }
}

public async Task TestMethodAsync()
{
    using (var db = new DbContext())
    using (var table = db.CreateTempTable<TestClass>())
    {
        var value = await db.GetTable<TestClass>()
            .Where(x => x.Id == 0)
            .Select(x => x.Value)
            .MaxAsync();
    }
}
```
Throws:
```
System.InvalidOperationException:
Function Max returns non-nullable value, but result is NULL. Use nullable version of the function instead.
in LinqToDB.Linq.Builder.AggregationBuilder.AggregationContext.CheckNullValue(IDataRecord reader, Object context)
```

So I try:
* `DefaultIfEmpty` after `Select`
* `DefaultIfEmpty` after `Where`
* `Sql.AsNullable`

```c#
var value = await db.GetTable<TestClass>()
    .Where(x => x.Id == 0)
    .Select(x => x.Value)
    .DefaultIfEmpty()
    .MaxAsync();

var value = await db.GetTable<TestClass>()
    .Where(x => x.Id == 0)
    .DefaultIfEmpty(new TestClass())
    .Select(x => x.Value)
    .MaxAsync();

var value = await db.GetTable<TestClass>()
    .Where(x => x.Id == 0)
    .Select(x => Sql.AsNullable(x.Value))
    .MaxAsync();
```
All above three raise the same exception.

What works is to cast property to nullable and then use null-coalescing operator:
```c#
var value = await db.GetTable<TestClass>()
    .Where(x => x.Id == 0)
    .Select(x => (int?)x.Value)
    .MaxAsync() ?? 0;
```

I don't want to use OrderByDescending and then FirstOrDefault.
```c#
var value = await db.GetTable<TestClass>()
    .Where(x => x.Id == 0)
    .Select(x => x.Value)
    .OrderByDescending(x => x)
    .FirstOrDefaultAsync();
```

Is there any other better and more intuitive way? Why doesn't `DefaultIfEmpty` work?

### Environment details
linq2db version: 3.1.5
Database Server: MsAccess
Database Provider: OleDb
Operating system: Win10
.NET Framework: 4.7.2

sdanyliv:
Fixes #2588

### DIFF
--- a/Tests/Linq/UserTests/Issue2588Tests.cs
+++ b/Tests/Linq/UserTests/Issue2588Tests.cs
@@ -1,0 +1,43 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using LinqToDB;
+using LinqToDB.Mapping;
+using NUnit.Framework;
+
+namespace Tests.UserTests
+{
+	[TestFixture]
+	public class Issue2588Tests : TestBase
+	{
+		[Table]
+		class TestClass
+		{
+			[Column] public int Id    { get; set; }
+			[Column] public int Value { get; set; }
+		}
+
+		[Test]
+		public async Task SampleSelectTest([DataSources] string context)
+		{
+			using (var db = GetDataContext(context))
+			using (var table = db.CreateLocalTable<TestClass>())
+			{
+				var value1 = await db.GetTable<TestClass>()
+					.Where(x => x.Id == 0)
+					.Select(x => x.Value)
+					.MaxAsync();
+
+				var value2 = await db.GetTable<TestClass>()
+					.Where(x => x.Id == 0)
+					.Select(x => Sql.ToNullable(x.Value))
+					.MaxAsync();
+
+				var value3 = await db.GetTable<TestClass>()
+					.Where(x => x.Id == 0)
+					.Select(x => x.Value)
+					.DefaultIfEmpty()
+					.MaxAsync();
+			}
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


viceroypenguin:
Question: current behavior for `IEnumerable.Max()` is to throw an exception if the collection is empty and the type is not-nullable. I believe that's where the expectation came from for the existing code. It will, however, work if `.DefaultIfEmpty()` is used, or if the type is nullable. Do we want to try to match that behavior? (In the Unit Test you wrote: Case 1 should throw an exception, Case 2 and 3 should work as written). 

```csharp
Array.Empty<int>().Max();                // throws exception
Array.Empty<int?>().Max();               // returns 0
Array.Empty<int>().DefaultIfEmpty.Max(); // returns null
```

sdanyliv:
@viceroypenguin, thanks for that. Will correct.

viceroypenguin (approved):

Tests/Linq/UserTests/Issue2588Tests.cs
```diff
@@ -0,0 +1,56 @@
...
+					.Select(x => x.Value)
+					.DefaultIfEmpty()
+					.MaxAsync();
+
+				Assert.AreEqual(0, value2);
+
+				//var value3 = await db.GetTable<TestClass>()
+				//	.Where(x => x.Id == 0)
+				//	.Select(x => x.Value)
+				//	.DefaultIfEmpty(5)
```
MaceWindu, line 49: Right now default value is ignored. Wether it should work or throw exception, that custom default value not supported in this context.

MaceWindu (approved):

viceroypenguin (approved):

PR #2625, opened 2020-11-15
[pgsql] fix unsigned integers mapping

### LINKED ISSUE #2624: UInt64 and Npgsql
When using BulkInsertAsync, I'm seeing this error:

System.InvalidCastException: 'Can't write CLR type System.UInt64 with handler type NumericHandler'

When using BulkCopyAsync I get an index of of range exception. I'm guessing these errors have the same source. I am using Npgsql. EF Core 5.0 is creating rows for my ulong of type Numeric with precision 20, scale 0, and dimensions 0.

MaceWindu:
Fix #2624

adds support for `ushort`, `uint` and `ulong` data in parameters in bulk copy

### DIFF
--- a/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLMappingSchema.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLMappingSchema.cs
@@ -5,6 +5,7 @@ using System.Text;
 namespace LinqToDB.DataProvider.PostgreSQL
 {
 	using LinqToDB.Common;
+	using LinqToDB.Data;
 	using LinqToDB.SqlQuery;
 	using Mapping;
 	using System.Data.Linq;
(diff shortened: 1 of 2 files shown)


sdanyliv (approved):

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLMappingSchema.cs
```diff
@@ -36,6 +37,11 @@ protected PostgreSQLMappingSchema(string configuration, params MappingSchema[] s
 			AddScalarType(typeof(string),          DataType.Text);
 			AddScalarType(typeof(TimeSpan),        DataType.Interval);
 			AddScalarType(typeof(TimeSpan?),       DataType.Interval);
+
+			// npgsql doesn't support unsigned types except byte (and sbyte)
+			SetConvertExpression<ushort?, DataParameter>(value => new DataParameter(null, value == null ? (int?)null     : (int)value    , DataType.Int32)  , false);
```
viceroypenguin, line 42: This covers saving to DB. Does this cover queries that return a `ushort?`/`uint?`/`ulong` coming from the db? I know that's not a concern for bulk copy, but if we're adding the mapping...

MaceWindu: Those conversions already exist and covered by test

viceroypenguin: Then I'm good.

viceroypenguin (approved):

PR #2627, opened 2020-11-15
[MiniProvider] Fix datareader expression resolve for wrapped provider

### LINKED ISSUE #2499: MiniProfiler.GetProviderSpecificFieldType could hide real column data type
Discovered during refactoring in https://github.com/linq2db/linq2db/pull/2455.
For oracle timestamptz it returns `DateTime` type instead of oracle-specific type for `DateTimeOffset` mapping, which leads to read of value with incorrect offset.
We should try (if possible) to fix DateTime mapping and properly resolve `GetProviderSpecificFieldType` call

MaceWindu:
Fix #2499


Source/LinqToDB/Linq/DataReaderWrapCache.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Data;
+using System.Linq.Expressions;
+using LinqToDB.Expressions;
+using LinqToDB.Mapping;
+
+namespace LinqToDB.Linq
+{
+	internal static class DataReaderWrapCache
+	{
+		private static readonly ConcurrentDictionary<Tuple<Type, MappingSchema>, Func<IDataReader, IDataReader>?> _readerMappings = new ConcurrentDictionary<Tuple<Type, MappingSchema>, Func<IDataReader, IDataReader>?>();
```
viceroypenguin, line 12: For the number of times this will change, I was recently introduced to a faster way than `ConcurrentDictionary<,>`. See [here](https://ayende.com/blog/192324-A/always-profile-the-hidden-cost-of-serializing-large-object-graphs-to-json), or proof [here](https://github.com/viceroypenguin/DictionaryBenchmark/blob/master/DictionaryBenchmark/BenchmarkDotNet.Artifacts/results/DictionaryBenchmark.DictionaryTest-report-github.md). Cost is probably negligible, but FYI. No comments from me for using the easier path of `ConcurrentDictionary<,>`, just wanted to offer options.

viceroypenguin (approved):

Source/LinqToDB/Linq/DataReaderWrapCache.cs
MaceWindu: Interesting idea, but I don't see (or missed) statement that dictionary copy constructor is thread-safe

MaceWindu:
Will be merged later as I want to estimate performance impact of this change

Source/LinqToDB/Linq/DataReaderWrapCache.cs
sdanyliv: I don't like static cache. Better to create similar using MemoryCache. We do not know which solution user can use and how many mapping schema can be created. As a key use not MappingSchema itself but its Id.

sdanyliv (changes requested):
Use `MemoryCache` instead of `ConcurrentDictionary`.

Source/LinqToDB/Linq/DataReaderWrapCache.cs
viceroypenguin: @sdanyliv has a good point on MemoryCache, so I'll defer to that for actual advice.

To answer the question, the thread safe is because you create a new dictionary and then assign the dictionary after the fact. It relies on the copy being made before the assignment, so the dictionary exposed by the class field is only ever read from, not written to. It also relies on the being able to generate duplicate values and throw away the lost one, because the race condition means one thread may not get to keep the value it created in the dictionary and that key's value must be regenerated next time through.

MaceWindu: Ah, true.

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1520,12 +1520,14 @@ public StringComparer ColumnNameComparer
 		public EntityDescriptor GetEntityDescriptor(Type type)
 		{
 			var key = new { Type = type, ConfigurationID };
```
MaceWindu, line 1522: this method was allocating a lot due to being hotpath, now it should be better, as don't allocate closue object here anymore, only key. Wonder if it makes sense to use struct for key here performance-wise.

When we have time, we should mark closure-less lambdas static and try refactor remaining to use context/avoid closures.

Source/LinqToDB/Common/Internal/Cache/MemoryCacheExtensions.cs
```diff
@@ -106,6 +106,23 @@ public static TItem GetOrCreate<TItem>(this IMemoryCache cache, object key, Func
...
+			where TKey: notnull
+		{
+			if (!cache.TryGetValue(key, out var result))
+			{
+				var entry = cache.CreateEntry(key);
+				result = factory(entry, key, context);
+				entry.SetValue(result);
+				// need to manually call dispose instead of having a using
+				// in case the factory passed in throws, in which case we
+				// do not want to add the entry to the cache
```
viceroypenguin, line 119: I'm not sure I understand your comment here. A `using` would always call `.Dispose()`, but even w/ a using, if the factory throws, it won't call `.SetValue()`. Do you not want to call `.Dispose()` if we throw?

Source/LinqToDB/Mapping/MappingSchema.cs
viceroypenguin: I wonder if we should also consider either re-implementing `ValueTuple<>` or revisiting pulling in System.ValueTuple. Neither would be for query use, but would be for key use to reduce allocations in situations like this.

Source/LinqToDB/Common/Internal/Cache/MemoryCacheExtensions.cs
viceroypenguin: Ah. `entry.Dispose()` is the call that actually plugs the value into the `cache`.

viceroypenguin: Would the following make more sense?
```csharp
result = factory(entry, key, context);
using (var entry = cache.CreateEntry(key))
  entry.SetValue(result);
```

If we create the entry after we execute the factory, we don't have to worry about whether the entry gets saved into the cache or not.

viceroypenguin (approved):
Approve in general principle.

sdanyliv (approved):

Source/LinqToDB/Common/Internal/Cache/MemoryCacheExtensions.cs
MaceWindu: that's code from MS (https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs) and actually I see they changed it 2 month ago due to memory leak issue. Probably we need to sync this code

viceroypenguin: I still don't see why creating the entry before calling factory makes sense. I wonder if msft is assuming factory won't throw?

PR #2635, opened 2020-11-18
[Tests] Support connection string pointers.

nsarris:
Allows for a connecionstring enclosed in brackets to be translated as a reference to another connectionstring name. Due to the architecture of providers, a schema or non database related option has to spawn a different provider, so you might be able to reuse the same connection string to test different providers.

This is PR No.1 of a series related to help porting tests to 3rd party providers (like IBM i).


MaceWindu:
not sure why you need it, but I'm fine with it. I will merge all changes into one branch to run our tests there

PR #2636, opened 2020-11-18
[Tests] multischema fix

nsarris:
Some providers (like the IBM i ) support loading multiple schemas (libraries in IBM i) so the schema providers can load tables with the same name. This patch handles schema testing so that it looks for tables in the default schema, and fallsback to schema ignorant search in case the providers doesn't support default schema information.

This is PR No.2 of a series related to help porting tests to 3rd party providers (like IBM i).


MaceWindu:
Not sure why you have those test tables it multiple schemas (probably you test some schema-related stuff).

PR #2637, opened 2020-11-18
[Tests] Stripped all SQL trace comments from tests before execution

nsarris:
A couple of tests stripped comments Access related comments before executing the SQL produced by the tracing mechanism when calling ToString on a query expression. To support more providers that can't handle inline comments in SQL, all comments were removed in said tests before the raw SQL is dispatched for execution.

This is PR No.3 of a series related to help porting tests to 3rd party providers (like IBM i).


MaceWindu:
Agree

PR #2639, opened 2020-11-18
[Tests] Added sttribute to control Upsert tests on unsupported providers.

nsarris:
The default behaviour from the base DataProvider for InsertOrUpdate is to throw an unsupported exception. All of the embedded linq2db providers support it so there is no ability to skip those tests. This patch adds an attribute to control this for tests tha invoke the InsertOrUpdate / InsertOrReplace functions, similarly to Merge. The list of unsupported providers is empty by default.

This is PR No.5 of a series related to help porting tests to 3rd party providers (like IBM i).

### DIFF
--- a/Tests/Base/InsertOrUpdateDataSourcesAttribute.cs
+++ b/Tests/Base/InsertOrUpdateDataSourcesAttribute.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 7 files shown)


MaceWindu:
`InsertOrUpdate` API implementation actually produce multiple queries if database doesn't support single-command upsert, so nothing prevent you from implementing it in iSeries provider I think. Not like I'm fan of it, but it provides single call API to users.

PR #2640, opened 2020-11-18
[Tests] Customization interceptor

nsarris:
This patch helps control which tests will run without editing the source code. It is critical for porting and maintaining the test code base for 3rd party providers.

It basicly providers a class with 3 overridable methods that by default do nothing, and a static class that provides a singleton instance of that class. If someone wants to port the tests to their own project they only have to:

1. Create a new subclass of the interceptor class
2. Switch the singleton instance of the default implementation (this is the only change needed in the original source code)

In this way, the developer can copy the tests over to their project, inject their own list of providers globally, and control which tests should not run, or should run with different filters on a test base. 

This is PR No.6 (and last) of a series related to help porting tests to 3rd party providers (like IBM i).

### DIFF
--- a/Tests/Base/CustomizationSupport/CustomizationSupport.cs
+++ b/Tests/Base/CustomizationSupport/CustomizationSupport.cs
@@ -2,7 +2,7 @@
 {
 	public static class CustomizationSupport
 	{
-		//Replace this instance with a custom implementation to ovveride default behaviour
+		//Replace this instance with a custom implementation to override default behaviour
 		public static readonly CustomizationSupportInterceptor Interceptor = new CustomizationSupportInterceptor();
 	}
 }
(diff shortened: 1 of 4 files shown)


Tests/Base/CustomizationSupport/CustomizationSupport.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Tests
+{
+	public static class CustomizationSupport
+	{
+		//Replace this instance with a custom implementation to ovveride default behaviour
```
MaceWindu, line 5:
```suggestion
		//Replace this instance with a custom implementation to override default behaviour
```

MaceWindu:
I like it. I was thinking about it couple of times - how to support use of our tests with 3rd party providers

PR #2661, opened 2020-11-26
Optimize mapper expression to be compatible with SequentialAccess command behavior

### LINKED ISSUE #1185: ExecuteReaderAsync should use CommandBehavior
The question is well explained here:

https://stackoverflow.com/questions/28543293/entity-framework-async-operation-takes-ten-times-as-long-to-complete/

Linq2db is affected as well. So it has terrible performance with byte[] columns (varbinary(Max)).

CommandBehavior.SequentialAccess increase speed maybe in ten times.

### LINKED ISSUE #2116: Sequential access does not work with calculations in new clause
Exception when querying using option CommandBehavior.SequentialAccess.  

```
Exception message:

System.InvalidOperationException: Invalid attempt to read from column ordinal '0'.  With CommandBehavior.SequentialAccess, you may only read from column ordinal '2' or greater.

Stack trace:

   at System.Data.SqlClient.SqlDataReader.CheckDataIsReady(Int32 columnIndex, Boolean allowPartiallyReadColumn, Boolean permitAsync, String methodName)
   at System.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn)
   at System.Data.SqlClient.SqlDataReader.GetString(Int32 i)
   at lambda_method(Closure , IQueryRunner , IDataReader )
   at LinqToDB.Linq.QueryRunner.Mapper`1.Map(IQueryRunner queryRunner, IDataReader dataReader)
   at LinqToDB.Linq.QueryRunner.<ExecuteQueryAsync>d__11`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at LinqToDB.AsyncExtensions.<ToListAsync>d__6`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Maintenance.Queries.Job.<Execute>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at Maintenance.Queries.Program.<Main>d__0.MoveNext()
```

```c#
using (Storage storage = new Storage())
            {
                var accounts = from account in storage.Accounts
                               select new
                               {
                                   Name = account.Name,
                                   Surname = account.Surname,
                                   Fullname = $"{account.Name} {account.Surname}"
                               };


                var result = await accounts.ToListAsync();

                var data = result.First();

                Console.WriteLine($"{data.Name} {data.Surname}");
            }
```

### Environment details
linq2db version: compiled from master on date of this issue
Database Server: SQL Server 2014

--

## Reason and solution

I think it is limitation of SQL Server client or even ADO.Net at all (but other DB are not tested).  
The reason is related to multiple attempts to access the same value (but it cannot be done in sequential mode).  

If you profile SQL query it will be like this one:

```sql
SELECT
	[account].[Name],
	[account].[Surname]
FROM
	[dbo].[Account] [account]
```

So in reality it gets only one value from database and it makes sense.  
But it is mistake to access value multiple times with reader.GetValue(column).  
Instead it should be cached, so value once read with GetValue should be available in memory while it is needed.  
It will solve a lot of troubles and will provide correct work in all modes.

MaceWindu:
Fix #2116
Fix #1185

Following providers support SequentialAccess behavior:
- Access (OleDb)
- MySql.Data
- npgsql
- System.Data.SqlClient
- Microsoft.Data.SqlClient
- SqlCe

---

To switch linq2db mapper to `SequentialAccess`-compatible mode user should set `LinqToDB.Common.Configuration.OptimizeForSequentialAccess` option to `true`.
It is disabled by default as it is possible to configure mappings that we cannot optimize, so we don't want to introduce breaking change.

For tests I've enabled this optimization and `SequentialAccess` behavior, to make sure it works for existing tests. I think we can leave it ON to ensure it works with future changes.

#### How optimization works
We find all calls to
- IsDBNull method (IsDBNullAsync method not detected for now as our mapper is not async)
- Other data reader instance methods with single integer parameter (column index)

calls moved to begin of mapper expression in form:
```cs
var db_null_X = rd.IsDBNull(x);
// type of variable is Nullable<T> if METHOD returns struct
var value_X = db_null_X ? null : rd.METHOD(x);
```

in places where those methods were called we place following code:
- IsDBNull call: db_null_X variable
- Other call: `value_X` or `value_X.Value` (for struct)

To make this optimization possible, we shouldn't put calls to reader methods to other methods, so I've refactored places where we do it:
- `ChainContext.CheckNullValue`
- `AggregationContext.CheckNullValue`
- `GroupByHelper.GetGrouping`
- `TableContext.DefaultInheritanceMappingException`
- for slow mode mapper added `ColumnReader.GetValueSequential` method for sequential access mode support

This is actually main reason why this optimization is not enabled by default - it is possible to configure reader expression that hides reader method calls and we cannot do anything about it - such mapping should be refactored by author

#### Changes Summary
- bumped npgsql version, used in NETFX tests to 4.1.6 as 4.0.x contains blocking bug in `SequentialAccess` implementation
- updated `DataReader`-based queries in schema providers to read data in `SequentialAccess`-compatible order
- moved `IsDBNull` call from `ChainContext` and `AggregationContext` to parameter, so we can optimize it
- added `LinqToDB.Common.Configuration.OptimizeForSequentialAccess` option
- enabled `SequentialAccess` for all tests
- replaced found existing Func mappers with Expression-based mappers to be able to reorder reader calls for them
- removed unnecessary nested lambda invoke in mapper expression
- added `ColumnReaderAttribute` to mark methods that call data reader column accessor and used in column mapping (we need it only for `Informix`, as it use using, so it cannot be moved around freely)

#### Known Issues
When `SequentialAccess` behavior used and mapper failed, slow-mode wrapper will fail too, as reader already advanced to next columns. Nothing we can do here, user will need to restart query and it will use slow-mode mapper. This basically means slow mapping mode is less useful with `SequentialAccess`.
We have several tests like that where I disabled `SequentialAccess` behavior. Note that those tests could indicate bugs in mapping, that worth of investigation.

TODO:
- [x] check existing reader mappings to be compatible with `SequentialAccess` (Oracle and Informix providers left)
- [x] test all in SlowMode + SequentialAcces
- [x] test all in SequentialAcces
- [x] test all in SlowMode
- [x] test all default


viceroypenguin (approved):

MaceWindu:
@sdanyliv that's not fair :)))

MaceWindu:
Ok, itermediate summary:
Basic implementation is quite simple and could handle:
- multiple column reads
- out-of-order reads

But there are couple of cases that are not so easy:
- inheritance mapping (at least) generate arbitrary-placed IsDBNull calls (solvable)
- same column could be read multiple times using different read expressions and I don't see good solution here yet

MaceWindu:
Uploaded new implementation and updated initial message.

Source/LinqToDB/Linq/QueryRunner.cs
```diff
@@ -146,6 +148,11 @@ public T Map(IDataContext context, IQueryRunner queryRunner, IDataReader dataRea
 						});
 				}
 
+				if (Configuration.OptimizeForSequentialAccess)
+					expression = SequentialAccessHelper.OptimizeMapingExpressionForSequentialAccess(expression, dataReader.FieldCount, false);
```
viceroypenguin, line 152:
```suggestion
					expression = SequentialAccessHelper.OptimizeMappingExpressionForSequentialAccess(expression, dataReader.FieldCount, reduce: false);
```

Source/LinqToDB/Linq/QueryRunner.cs
```diff
@@ -621,21 +627,33 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
...
+			var queryRunnerParam = expression.Parameters[0];
+			var dataReaderParam  = expression.Parameters[2];
+
+			var dataContextVar = expression.Parameters[1];
+			var expressionVar  = expression.Parameters[3];
+			var parametersVar  = expression.Parameters[4];
+			var preamblesVar   = expression.Parameters[5];
 
+			// we can safely assume it is block expression
+			var block = (BlockExpression)expression.Body;
```
viceroypenguin, line 639:
```suggestion
			if (expression.Body is not BlockExpression block)
				throw new LinqException("???");
```

Source/LinqToDB/Linq/SequentialAccessHelper.cs
```diff
@@ -0,0 +1,301 @@
...
+		public static Expression OptimizeMapingExpressionForSequentialAccess(Expression expression, int fieldCount, bool reduce)
+		{
+			if (reduce)
+				expression = expression.Transform(e =>
+				{
+					if (e is ConvertFromDataReaderExpression conv)
+						return conv.Reduce();
+
+					return e;
+				});
```
viceroypenguin, line 25: optional. I think it reads better, but that's just me
```suggestion
				expression = expression.Transform(e =>
					e is ConvertFromDataReaderExpression conv
						? conv.Reduce()
						: e;
```

Source/LinqToDB/Linq/SequentialAccessHelper.cs
```diff
@@ -0,0 +1,301 @@
...
+				}
+
+				return e;
+			};
+
+			expression = expression.Transform(tranformFunc);
+
+			// expression cannot be optimized
+			if (failMessage != null)
+				throw new LinqToDBException($"{nameof(OptimizeMapingExpressionForSequentialAccess)} optimization failed: {failMessage}");
```
viceroypenguin, line 198:
```suggestion
				throw new LinqToDBException($"{nameof(OptimizeMappingExpressionForSequentialAccess)} optimization failed: {failMessage}");
```

Source/LinqToDB/Linq/SequentialAccessHelper.cs
```diff
@@ -0,0 +1,301 @@
...
+					// - method, marked with ColumnReaderAttribute
+					// - ColumnReader.GetValueSequential
+
+					// ColumnReaderAttribute method
+					var attr = call.Method.GetCustomAttribute<ColumnReaderAttribute>();
+					if (attr != null && call.Arguments[attr.IndexParameterIndex] is ConstantExpression c1 && c1.Type == typeof(int))
+						columnIndex = (int)c1.Value;
+
+					// instance method of data reader
+					// check that method accept singe integer constant as parameter
```
viceroypenguin, line 57:
```suggestion
					// check that method accept single integer constant as parameter
```

Source/LinqToDB/Linq/SequentialAccessHelper.cs
```diff
@@ -0,0 +1,301 @@
...
+						// test IsDBNull method by-name to support overrides
+						if (call.Object != null && typeof(IDataReader).IsAssignableFrom(call.Object.Type) && call.Method.Name == nameof(IDataReader.IsDBNull))
+						{
+							var index = columnIndex.Value * 2;
+							if (newVariables[index] == null)
+							{
+								var variable               = Expression.Variable(typeof(bool), $"is_null_{columnIndex}");
+								newVariables[index] = variable;
+								replacements[index] = variable;
+								insertedExpressions[index] = Expression.Assign(variable, call);
```
viceroypenguin, line 74: Alignment

Source/LinqToDB/Linq/SequentialAccessHelper.cs
```diff
@@ -0,0 +1,301 @@
...
+			{
+				if (failed)
+					return e;
+
+				if (e is MethodCallExpression call)
+				{
+					// we work only with instance method of data reader
+					if (!call.Method.IsStatic && typeof(IDataReader).IsAssignableFrom(call.Object.Type))
+					{
+						// check that method accept singe integer constant as parameter
```
viceroypenguin, line 252:
```suggestion
						// check that method accept single integer constant as parameter
```

viceroypenguin (commented):
Most of this is too opaque for me to judge clearly, but if tests pass, I'm good. Spelling and other misc minor improvements

Source/LinqToDB/DataProvider/Informix/InformixDataProvider.cs
```diff
@@ -55,18 +56,21 @@ public InformixDataProvider(string providerName)
 			if (Adapter.TimeSpanType != null) SetProviderField(Adapter.TimeSpanType, typeof(TimeSpan), Adapter.GetTimeSpanReaderMethod, dataReaderType: Adapter.DataReaderType);
 		}
 
+		[ColumnReader(1)]
```
MaceWindu, line 59: WHY: this reader cannot be optimized (if we will move GetXXX method, using will do nothing), so we introduce special attribute to tell linq2db that this method should be treated as GetXXX method of data reader. Number parameter specifies index of column index argument in method signature.

Source/LinqToDB/DataProvider/Oracle/OracleProviderAdapter.cs
```diff
@@ -364,12 +365,13 @@ private static OracleProviderAdapter CreateAdapter(string assemblyName, string c
...
 			body = generator.Build();
 
-			var readOracleDecimalToDecimalAdv = (Expression<Func<IDataReader, int, decimal>>)Expression.Lambda(body, rdParam, indexParam);
 			// workaround for mapper issue with complex reader expressions handling
 			// https://github.com/linq2db/linq2db/issues/2032
-			var compiledReader                = readOracleDecimalToDecimalAdv.Compile();
-			readOracleDecimalToDecimalAdv     = (Expression<Func<IDataReader, int, decimal>>)Expression.Lambda(
-				Expression.Invoke(Expression.Constant(compiledReader), rdParam, indexParam),
+			var compiledReader                = Expression.Lambda(body, oracleDecimalParam).Compile();
+			var readOracleDecimalToDecimalAdv = (Expression<Func<IDataReader, int, decimal>>)Expression.Lambda(
```
MaceWindu, line 371: WHY: this reader expression were refactored to not hide data reader method call from mapping, column mapper logic preserved

Source/LinqToDB/DataProvider/Oracle/OracleTools.cs
```diff
@@ -303,6 +303,7 @@ public static DataConnection CreateDataConnection(IDbTransaction transaction, st
 
 		public static AlternativeBulkCopy UseAlternativeBulkCopy = AlternativeBulkCopy.InsertAll;
 
+		[Obsolete("This field is not used by linq2db. Configure reader expressions on DataProvider directly")]
 		public static Func<IDataReader,int,decimal> DataReaderGetDecimal = (dr, i) => dr.GetDecimal(i);
```
MaceWindu, line 307: WHY: This method wasn't used already by linq2db (probably to refactorings at some point) and this is not the best way to configure reader expressions anyway

Source/LinqToDB/DataProvider/SqlServer/SqlServerDataProvider.cs
```diff
@@ -66,10 +66,6 @@ public SqlServerDataProvider(string name, SqlServerVersion version, SqlServerPro
 				SqlServerVersion.v2017 => new SqlServer2017SqlOptimizer(SqlProviderFlags),
 				_                      => new SqlServer2008SqlOptimizer(SqlProviderFlags),
 			};
-			SetField<IDataReader, decimal>((r, i) => r.GetDecimal(i));
```
MaceWindu, line 69: WHY: 
- duplicates base class logic
- removed data reader configuration from tools

Source/LinqToDB/DataProvider/SqlServer/SqlServerTools.cs
```diff
@@ -368,7 +369,9 @@ public static class Sql
 			public const string OptionRecompile = "OPTION(RECOMPILE)";
 		}
 
+		[Obsolete("This field is not used by linq2db. Configure reader expressions on DataProvider directly")]
 		public static Func<IDataReader,int,decimal> DataReaderGetMoney   = (dr, i) => dr.GetDecimal(i);
+		[Obsolete("This field is not used by linq2db. Configure reader expressions on DataProvider directly")]
```
MaceWindu, line 374: WHY: deliberate breaking change, data reader expressions should be configured only on data provider level

Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
```diff
@@ -106,9 +107,9 @@ public AggregationContext(IBuildContext? parent, IBuildContext sequence, MethodC
 			public int             FieldIndex;
 			public ISqlExpression? Sql;
 
-			static int CheckNullValue(IDataRecord reader, object context)
+			static int CheckNullValue(bool isNull, object context)
 			{
-				if (reader.IsDBNull(0))
+				if (isNull)
```
MaceWindu, line 112: WHY: `IsDBNull` call not visible to mapping expression and cannot be moved, when called inside of function (here and in another similar place). Move it to expression level (to caller)

Source/LinqToDB/Linq/Builder/GroupByBuilder.cs
```diff
@@ -376,44 +376,24 @@ public Expression GetGrouping(GroupByContext context)
 					var itemReader      = CompiledQuery.Compile(lambda);
 					var keyExpr         = context._key.BuildExpression(null, 0, false);
 
-					var keyReader  = Expression.Lambda<Func<IQueryRunner,IDataContext,IDataReader,Expression,object?[]?,TKey>>(
```
MaceWindu, line 379: WHY: Removed nested invoke from mapping expression:
- no need for extra expression invoke if we can embed expression body into mapper (less memory/cpu used by mapper)
- having invoke makes mapper expression rewrite big PITA (you need to add extra parameters to invoke when you move data reader calls from it to outer context)

Source/LinqToDB/Linq/QueryRunner.cs
```diff
@@ -621,21 +626,34 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
...
+			var queryRunnerParam = expression.Parameters[0];
+			var dataReaderParam  = expression.Parameters[2];
+
+			var dataContextVar = expression.Parameters[1];
+			var expressionVar  = expression.Parameters[3];
+			var parametersVar  = expression.Parameters[4];
+			var preamblesVar   = expression.Parameters[5];
 
+			// we can safely assume it is block expression
+			if (expression.Body is not BlockExpression block)
```
MaceWindu, line 638: WHY: another place where we remove nested invoke from mapping expression (here and below)

Tests/Linq/TestsInitialization.cs
```diff
@@ -17,6 +17,10 @@ public class TestsInitialization
 	[OneTimeSetUp]
 	public void TestAssemblySetup()
 	{
+		// uncomment it to run tests with SeqentialAccess command behavior
```
MaceWindu, line 20: For now I decided to not enable `SequentialAccess` behavior for all tests. We have it already tested and because it will not be used a lot, it is better to test `Default` behavior and add regression tests later if something will broke for `SequentialAccess`

MaceWindu (commented):
Just added some implementation notes for review

MaceWindu:
Added separate test provider  (SqlServer.2019.SA) to test linq2db in SequentialAccess mode

sdanyliv (approved):

Issue #2649, opened 2020-11-20
Firebird "arithmetic exception, numeric overflow, or string truncation" on huge sql string


kevinvenclovas:
Hello, im try to call a stored procedure from my c# code. I get every time a "arithmetic exception, numeric overflow, or string truncation string right truncation" Exception. When i call my stored procedure with the same Parameters from the Database directly i get no error like this. 
In my procedure i copy dynamic some entrys with littele changes.   
I generate dynamically sql code from tables with over 200 fields so its becomes a very long sql-string (over 10k chars). After this i execute this statement with:  "Execute Statement GENERATED_SQL;" 

And there is the issue. When i remove the "Execute Statement" line all works fine. With this line i get the error. 
But when i execute my stored procedure Select * from PROCEDURE(...) directly on the database i get no error? 
Why? Can someone help me?

GENERATE SQL STRING
```sql
for 
            select rdb$field_name 
            from rdb$relation_fields 
            where rdb$relation_name = 'BELEG' and rdb$field_name not in ('BELEGTYP','BELEGART','BELEGNR','BELEGDAT') 
            into fieldname 
do begin
            belegselsql = belegselsql||','||fieldname;
            beleginssql = beleginssql||','||fieldname;
end

GENERATED_SQL= beleginssql||') '||belegselsql||
            ' from beleg where belegtyp = '''||IBELEGTYP||
            ''' and belegart = '''||IBELEGART||''' and belegnr = '|| IBelegnr;

Execute Statement GENERATED_SQL;
```

EXCEPTION
```
FirebirdSql.Data.FirebirdClient.FbException : arithmetic exception, numeric overflow, or string truncation
    string right truncation
    At procedure 'ALFA_BELEG_VERSENDER_AENDERN' line: 78, col: 9
      ----> FirebirdSql.Data.Common.IscException : arithmetic exception, numeric overflow, or string truncation
    string right truncation
    At procedure 'ALFA_BELEG_VERSENDER_AENDERN' line: 78, col: 9
  Stack Trace: 
    FbCommand.Fetch()
    FbDataReader.Read()
    CommandInfo.ReadEnumerator[T](IDataReader rd, Boolean disposeReader)+MoveNext()
    List`1.ctor(IEnumerable`1 collection)
    Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

Issue #1189, opened 2018-06-14
Error for ExpressionMethod with default DateTime


Franki1986:
DB-Probider: Firebird
Linq2dB: Version 2.0.0

If I use the following ExpressionMethod I get a FbException for 'Data type unknown'

(The following code is a test sample with no realistic sense)

```
[Table("CUSTOMER", IsColumnAttributeRequired = true)]
public partial class Customer
{

    [Column("ID"), PrimaryKey, NotNull] public int Id { get; set; } // integer

    [Column("NAME"), NotNull] public string Name { get; set; } // varchar(20)

    [ExpressionMethod(nameof(DefaultDateTime), IsColumn = true)]
    //[Column("TO_DELETE")]
    public DateTime? ToDelete{ get; set; }

    static Expression<Func<Customer, DateTime>> DefaultDateTime()
    {
       return p => Sql.AsSql(DateTime.Now); 
    }
}
```

```
List<Customer> customers;
using (var context = new DbContext())
{
   customers= context.GetTable<Customer>().Where(k => k.ToDelete <= DateTime.Now).ToList();
 }
```

Exception:

FirebirdSql.Data.FirebirdClient Information: 0 : Command:
SELECT
	t1.ID,
	t1.NAME,
	'2001-01-01 01:01:01' as c1
FROM
	KUNDE t1
WHERE
	@p1 <= @p2

Parameters:
Name:p1	Type:TimeStamp	Used Value:14.06.2018 09:15:59
Name:p2	Type:TimeStamp	Used Value:14.06.2018 09:15:59

FirebirdSql.Data.FirebirdClient.FbException
  HResult=0x80004005
  Message=Dynamic SQL Error
SQL error code = -804
Data type unknown
  Source=FirebirdSql.Data.FirebirdClient
  StackTrace:
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteReader(CommandBehavior behavior) in C:\Users\Jiri\Documents\devel\NETProvider\working\Provider\src\FirebirdSql.Data.FirebirdClient\FirebirdClient\FbCommand.cs:line 517
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteDbDataReader(CommandBehavior behavior) in C:\Users\Jiri\Documents\devel\NETProvider\working\Provider\src\FirebirdSql.Data.FirebirdClient\FirebirdClient\FbCommand.cs:line 623
[...]

Issue #755, opened 2017-06-27
Firebird queries do not work with CHAR(x) CHARACTER SET OCTETS type


cantak:
I am trying to store some hex data in a database. There are numerous fields, but for my simple test case let's look at a MAC_ADDRESS field. 

```sql
CREATE TABLE FOO (
    MAC_ADDRESS  CHAR(6) CHARACTER SET OCTETS NOT NULL
);
```

A human readable MAC is hex formatted - something like "FF:FF:FF:FF:FF:FF". This can be stored as 6 bytes, each a value of 255. I am storing this data.

No combination of field definitions work for **both** _selecting_ and _insert/update_ on the table. I've tried **char[]**, **byte[]**, **string**, everything.

In the attached example, I am using a string, however as stated I have tried byte[] and other options. Either selection fails or update/insert fails, or both. Never do they all work.
[Code.zip](https://github.com/linq2db/linq2db/files/1106637/Code.zip)

### Environment details
linq2db version: *1.8.2*  
Database Server: *Firebird 2.5.2*  
Database Provider: *Firebird .Net Provider 5.9.1.0*  
Operating system: * Windows 10*  
Framework version: *.NET Framework 4.5.2*

jack128:
You can try to declare own type ```struct MacAddress { public byte[] Value;  }``` and define MappingScheme
```
var schema = new MappingSchema();
schema.SetConverter<MacAddress, DataParameter>(...)
schema.SetConverter<DataParameter,MacAddress>(...)
```

cantak:
Beautiful! Thank you!

ili:
In general your sample should work. If FB supports storing non ASCII (above `chr(127)`) in `CHAR` this should not be a problem, we'll investigate this!

ili:
@cantak have @jack128's solution solved your issue?

I'v took a closer look in to this case and faced with some problems: firebird client reads such field as string, I tried to insert  `ascii_char(50) || ascii_char(100) || ascii_char(150) || ascii_char(200)` and then read it, and there are problems with byte converting. 

How should I read such data directly (without using linq2db only with Firebird .Net Provider?

cantak:
Yes. I have a table with the following fields:

`
    MAC_ADDRESS       CHAR(6) CHARACTER SET OCTETS NOT NULL,
    IP_V4_ADDRESS     CHAR(4) CHARACTER SET OCTETS,
[...]

PR #2693, opened 2020-12-13
[Firebird] Provider versioning

### LINKED ISSUE #1024: Firebird: should use true/false literals for boolean fields instead of 1/0

MaceWindu:
##### Goal
Add Firebird versioning support on version and dialect level to better support features of recent firebird versions and better dialect 1 support.
Fix #1024

Related and planned for fix/review issues:
- #1024: `boolean` type support
- #755 : blob "type" improvements
- #1189: improvements for column type hints
- #1318: map `bigint` to `decimal(18)`/`decimal(19)`(in v4) for dialect1
- #2649: probably provider issue, fixed in 7.10.x. research required

##### Changes
- added separate providers (provider names) for Firebird v2.5, v3, v4 (both dialect1 and dialect3)  = six provider names in total
- [tests] added testing support for all 6 providers including CI
- [tests] old test providers should be renamed (in test config): `Firebird` -> `Firebird.25`, `Firebird3` -> `Firebird.3`
- [tests] renamed several test tables/columns in shared tests to make them work with `Dialect 1` (we cannot escape identifiers in `Dialect 1` and they used reserved words)

For provider changes see notes below.

##### TODO
- make all tests pass
- add tests for changes
- support new v4 data types
- support new sql features (not much, e.g. new syntax for TOP queries)
- enable `default` in tests: #2954

Upstream issues:

http://tracker.firebirdsql.org/browse/DNET-990

there are more, as tests currently crash provider, but they still need identification

---

##### Migration

Old provider implementation (including `Firebird` provider name) is now mapped to `Firebird25` provider (v2.5.x Dialect 3), which tries to to preserve compatibility for existing users.
When provider name is not specified explicitly in configuration, it is possible that linq2db will pick other provider version. Logic here is simple:
- if connection name or provider name contains tokens `3`, `4`, `Dialect1` or `Dialect.1` - corresponding provider version will be selected
- otherwise if `FirebirdTools.AutoDetectProvider` set to `true` (`true` is default value), linq2db will use version that Firebird returns and dialect from connection string

###### `boolean` type support

`boolean` type is a new Firebird type added in version 3.0. Previously we used `char(1)`: `'1'`/`'0'` for `bool` mapping. This mapping is still default one for v2.5 and v3/v4 is switched to `boolean` type by default. For backward compatibility old `char` mapping also supported in v3/v4 but now requires `DataType.Char` applied to column mapping.


MaceWindu:
dialect 1 support not planned (lack of interess)

PR #2695, opened 2020-12-15
Add ExpressionMethod Generator

### LINKED ISSUE #2687: [Idea] Source Generators
@linq2db/developers, maybe someone can try this? 

Idea is to create [Source Generator](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/) which can transform simple methods to ExpressionTree.
For example such class with methods marked by `ExpressionMethodGeneratorAttribute`
```cs
public static class MyBusinessExtensions
{
   [ExpressionMethodGenerator]
   public static PersonDto ToDto(this Person person)
   {
      return new PersonDto
      {
         Id = personId,
         Name = person.Name
      };
   }
}
```
Should be translated to 
```cs
public static class MyBusinessExtensions
{
   [ExpressionMethod("ToDto_SomeUniqueName")]
   public static PersonDto ToDto(this Person person)
   {
      return new PersonDto
      {
         Id = personId,
         Name = person.Name
      };
   }

   static Expression<Func<Person, PersonDto>> ToDto_SomeUniqueName()
   {
      return (person) => new PersonDto
      {
         Id = personId,
         Name = person.Name
      };
   }
}
```
Also as sample we can use https://github.com/robersonliou/MapperGenerator

viceroypenguin:
Fixes #2687

- [x] Implement Generators framework
- [x] Implement `ExpressionMethodGenerator`
- [x] Implement Diagnostic messages for error messages to users.
- [x] Implement SyntaxReceiver


Source/LinqToDB.Generators/LinqToDB.Generators.csproj
```diff
@@ -0,0 +1,31 @@
...
+		<TargetFrameworks>netstandard2.0</TargetFrameworks>
+		<Configurations>Debug;Release</Configurations>
+	</PropertyGroup>
+
+	<PropertyGroup Condition="'$(Configuration)'=='Release'">
+		<DebugType>pdbonly</DebugType>
+		<DebugSymbols>true</DebugSymbols>
+	</PropertyGroup>
+
+	<ItemGroup>
```
MaceWindu, line 19: could you extract sourcelink stuff from linq2db.source.props into separate props file and import it here and in `linq2db.source.props` ?

MaceWindu (approved):
Looks good.
Does it make sense to handle some C# expressions, not supported by `S.L.E`, like `??`, `?.`?

Source/LinqToDB.Generators/LinqToDB.Generators.csproj
viceroypenguin: You can't debug generators as a consumer. In fact, it's a PITA to even debug as the writer. The sourcelink stuff won't add any value for this project. I can still do it if you want though.

MaceWindu: Then why you added it here?

NuGet/linq2db.nuspec
```diff
@@ -33,8 +33,13 @@
 	</metadata>
 
 	<files>
+		<!-- Primary reference files -->
 		<file src="..\Source\LinqToDB\bin\Release\**\linq2db.pdb" target="lib\" />
 		<file src="..\Source\LinqToDB\bin\Release\**\linq2db.xml" target="lib\" />
 		<file src="..\Source\LinqToDB\bin\Release\**\linq2db.dll" target="lib\" />
+
+		<!-- Generators -->
+		<file src="..\Source\LinqToDB.Generators\bin\Release\netstandard2.0\LinqToDB.Generators.dll"
```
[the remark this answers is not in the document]
viceroypenguin: That makes sense. I'll switch it.

viceroypenguin:
> Does it make sense to handle some C# expressions, not supported by `S.L.E`, like `??`, `?.`?

I think the compiler will catch that when actually compiling the generated code. Probably won't be descriptive to the user though. That's a whole 'nother level of analysis though. I'd say let's do that in v2 of this thing if it looks like an issue.

Source/LinqToDB.Generators/AnalyzerReleases.Unshipped.md
```diff
@@ -0,0 +1,10 @@
+﻿; Unshipped analyzer release
+; https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+
+### New Rules
+Rule ID | Category | Severity | Notes
+--------|----------|----------|-------
+LDBGEN001 | LinqToDB.ExpressionMethodGenerator | Error | Diagnostics
```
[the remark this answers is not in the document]
viceroypenguin: "Officially", they should only get moved to shipped right before we do the actual release... I'm good with moving it to shipped now though. /shrug

https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md

Source/LinqToDB.Generators/LinqToDB.Generators.csproj
viceroypenguin: Copied and Pasted. :)

Source/LinqToDB.Generators/AnalyzerReleases.Unshipped.md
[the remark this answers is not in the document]
MaceWindu: Ah, I see. `for the upcoming or unshipped`

sdanyliv:
I'ld like to see more complex tests with several parameters.

Source/LinqToDB.Generators/LinqToDB.Generators.csproj
viceroypenguin: Removed

NuGet/linq2db.nuspec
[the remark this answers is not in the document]
viceroypenguin: Switched

viceroypenguin:
> I'd like to see more complex tests with several parameters.

Added one more complicated one. Anything specific you'd like to see in another one?

NuGet/linq2db.nuspec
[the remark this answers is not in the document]
MaceWindu: `LinqToDB.Generators.dll` reference not renamed

Source/LinqToDB.Analyzers/LinqToDB.Analyzers.csproj
```diff
@@ -0,0 +1,24 @@
...
+
+	<PropertyGroup>
+		<TargetFrameworks>netstandard2.0</TargetFrameworks>
+		<Configurations>Debug;Release</Configurations>
+	</PropertyGroup>
+
+	<PropertyGroup Condition="'$(Configuration)'=='Release'">
+		<DebugType>pdbonly</DebugType>
+		<DebugSymbols>true</DebugSymbols>
+	</PropertyGroup>
```
MaceWindu, line 17: probably also not needed, as we cannot debug it

viceroypenguin: Removed

NuGet/linq2db.nuspec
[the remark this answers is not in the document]
viceroypenguin: Whoops. Fixed.

NuGet/PackLocal.cmd
```diff
@@ -1,4 +1,4 @@
-ECHO OFF
+@ECHO OFF
```
viceroypenguin, line 1: @MaceWindu  so that `ECHO OFF` itself doesn't show up in the command line output

MaceWindu: TIL :point_up:

MaceWindu (approved):

Source/LinqToDB.Analyzers/ExpressionMethodGenerator.cs
```diff
@@ -0,0 +1,225 @@
...
+			var compilation = context.Compilation;
+			if (compilation.GetTypeByMetadataName("LinqToDB.Mapping.GenerateExpressionMethodAttribute") is not { } attributeSymbol) return;
+
+			foreach (var c in GetCandidates(context, receiver).GroupBy(g => g.containingType))
+			{
+				var containingType = c.Key;
+				if (containingType.DeclaringSyntaxReferences.First().GetSyntax() is not ClassDeclarationSyntax containingTypeSyntax)
+					continue;
+
+				if (!containingTypeSyntax.Modifiers.Any(m => m.Text == "partial"))
```
igor-tkachev, line 61: Can we add partial ourselves if it's not specified? And report a warning?

Source/LinqToDB.Analyzers/ExpressionMethodGenerator.cs
```diff
@@ -0,0 +1,225 @@
...
+				.WithSemicolonToken(SyntaxFactory.Token(SyntaxKind.SemicolonToken))
+				.WithAttributeLists(
+					SyntaxFactory.SingletonList(
+						SyntaxFactory.AttributeList(
+							SyntaxFactory.SingletonSeparatedList(
+								SyntaxFactory.Attribute(
+									SyntaxFactory.IdentifierName("LinqToDB.ExpressionMethodAttribute"),
+									SyntaxFactory.AttributeArgumentList(
+										SyntaxFactory.SingletonSeparatedList(SyntaxFactory.AttributeArgument(
+											SyntaxFactory.LiteralExpression(SyntaxKind.StringLiteralExpression, SyntaxFactory.Literal(newMethodName))))))))));
```
igor-tkachev, line 210: Any chance to use **nameof(__xxxxxx)** expression?

igor-tkachev (approved):

Source/LinqToDB.Analyzers/ExpressionMethodGenerator.cs
jogibear9988: see here: https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/

Can I modify/rewrite existing code with a Source Generator?
No. As mentioned earlier, Source Generators do not allow you to rewrite user source code. We do not intend on allowing them to this. They can only augment a compilation by adding C# source files to it.

sdanyliv (commented):
Well, it looks complicated for end user.

- Class should be dined as partial
- Method should be partial
- No way do define over property.

i suggest to change ExpressionMethodAttribute a little and maybe create base class. It should have at least one virtual method GetLambdaExpression. Then we can derive `ExpressionMethodGenerator` from base class return lambda in any case.

For generation we can always create static class MySuperClass_generated and generate methods in that class.

viceroypenguin:
* `static` classes cannot have inheritance, so using a base class would eliminate automatic generation for extension methods.
* Also, inheritance reduces options on non-static classes (static method on a DTO that may be tiered for example).
* Without inheritance, if we want to connect auto-generation to original method, we will have to do convention based reflection for the related class/method. This is not very difficult, since the `[GenerateExpressionMethod]` attribute will indicate we need to go look for it. However, it would require some minor work to tell l2db how to look.

Source/LinqToDB.Analyzers/ExpressionMethodGenerator.cs
viceroypenguin: Is there a purpose to doing so? In the compiled code, `nameof(...)` translates into a string literal, so there's no difference in the compiled code. The only difference is in the generated code, so only useful if we want to pretty-print for users to look at.

viceroypenguin:
Merry Christmas @sdanyliv - Now doesn't require `partial` on methods (still requires `partial` on classes though, no way around that) *and* supports properties. Still want to write some more unit tests, so not ready for merge yet.

MaceWindu:
There are 4 oracle tests that changed baselines due to sequence value change.

This means they use `InsertWithIdentity` call to generate sequence value and miss call to `ResetPersonIdentity` or `ResetAllTypesIdentity` method.

MaceWindu (approved):

MaceWindu:
> There are 4 oracle tests that changed baselines due to sequence value change.

addressed here #2758

viceroypenguin:
@sdanyliv @MaceWindu any reason for or against this in 3.3.0? I don't particularly care, it was built on an idea from @sdanyliv

MaceWindu:
@viceroypenguin I would say let's merge it after tests passed.

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
```diff
@@ -0,0 +1,114 @@
...
+			public int Id { get; set; }
+			public string Name { get; set; } = null!;
+		}
+
+		public partial class ExprPerson
+		{
+			public int Id { get; set; }
+			public string Name { get; set; } = null!;
+
+			[GenerateExpressionMethod(MethodName = "Test1")]
```
sdanyliv, line 29: Why do we need `MethodName`?

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
```diff
@@ -0,0 +1,114 @@
...
+			};
+		}
+
+		[GenerateExpressionMethod]
+		public static PersonDto ToDtoArrow(ExprPerson person)
+			=> new PersonDto
+			{
+				Id = person.Id,
+				Name = person.Name
+			};
```
sdanyliv, line 70: There are also needed tests for overloads.

sdanyliv (commented):
Please check my comments.

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
viceroypenguin: Good point - overloads create clashing expression method functions unless you use `MethodName`. What would you prefer as proper behavior here? a) let the compile fail, or b) generate an L2DB error, or c) something else? (I vote for b) myself.)

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
viceroypenguin: I added `MethodName` because there is a non-zero chance that someone will want to be able to use the expression for some reason besides just in l2db. The method name provides a way to expose that expression if desired.

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
sdanyliv: Generate method name using signature. MethodName + Full Parameter type names is the best approach I think.

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
sdanyliv: Good idea, but in this case we have to support that for whole library life ;) I. don't have ideas how it can be useful for other consumers.

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
viceroypenguin: Done!

Tests/Linq/Generators/ExpressionMethodGeneratorTests.cs
viceroypenguin: 🙄 fine.... whatever... 😂  Done!

MaceWindu:
I don't understand. Nothing big changed with last commits except merge with master, but now it doesn't build with a lot of errors...

MaceWindu:
nevermind, just some glitch, after rebuild it complains about code referencing removed `MethodName` property

MaceWindu:
@viceroypenguin, reminder about broken branch

viceroypenguin:
Thanks @MaceWindu. Will revisit when I have time to do so.

PR #2699, opened 2020-12-15
Array.Length mapping support, byte[].Length translation to SQL

### LINKED ISSUE #2691: Don't client-side large objects [Possible 3.x regression]
> **This might be a 3.x regression.** 
The code below was developed in 2.x and I'm 90% confident that it did what I want.
We recently noticed a serious perf regression and observed the (new?) behavior described here.

Take a model that maps a DB blob into a .net `byte[]`.
```c#
class File 
{
  public int Id { get; set; }
  public byte[] Data { get; set; }  // blob in DB
}
```

This query attempts to fetch the size of the first 20 records:
```c#
var q = from f in db.Files select new { f.Name, Length = f.Data.Length };
q.Take(20).ToList();
```

The SQL looks like this:
```sql
select Name, Data
from Files
where rownum <= 20
```

My gripe here is that linq2db fetches all of `Data` and chooses to evaluate `Data.Length` on the client-side.

Generally speaking, I love how linq2db evaluates as much as possible client-side. 
It gives access to all .net features that dbs don't have and the results are consistent with other c# code (there's always minor discrepancies in the implementation of "similar" functions). 

In this specific instance, I don't love it as much because `Data` is a heavy, costly object. Just imagine each blob is on average 5Mo. That query moves 100Mo over the network!!
It's especially nasty because it still "works". We noticed the problem after significant perf. degradation.

I would suggest that `byte[].Length` is special-cased to always be server-side evaluated.
I.e. above query should generate:
```sql
select Name, LENGTH(Data)
from Files
where rownum <= 20
```

Side note: I couldn't find a built-in `Sql.Length()` that takes `byte[]`, so working around this requires a bit of knowledge (I created my own length function).

### Environment details
linq2db version: 3.1.4
Database Server: Oracle 12 
Database Provider: Managed
Operating system: Windows 
.NET Framework: Core 2.1

jogibear9988:
Fix #2691

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
@@ -506,7 +506,7 @@ namespace LinqToDB.Linq.Builder
 							var ll = Expressions.ConvertMember(MappingSchema, ue.Operand?.Type, ue.Operand!.Type.GetProperty(nameof(Array.Length))!);
 							if (ll != null)
 							{
-								var ex = convertMemberExpression(expr, ue.Operand!, ll);
+								var ex = СonvertMemberExpression(expr, ue.Operand!, ll);
 
 								return new TransformInfo(ex, false, true);
 							}
--- a/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
@@ -616,7 +616,7 @@ namespace LinqToDB.Linq.Builder
 
 			return result;
 
-			static Expression convertMemberExpression(Expression expr, Expression root, LambdaExpression l)
+			static Expression СonvertMemberExpression(Expression expr, Expression root, LambdaExpression l)
 			{
 				var body  = l.Body.Unwrap();
 				var parms = l.Parameters.ToDictionary(p => p);
--- a/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
@@ -501,41 +501,17 @@ namespace LinqToDB.Linq.Builder
 				switch (expr.NodeType)
 				{
 					case ExpressionType.ArrayLength:
-						var ue = (UnaryExpression)expr;
-						var ll = Expressions.ConvertMember(MappingSchema, ue.Operand?.Type, ue.Operand.Type.GetProperty(nameof(Array.Length)));
-						if (ll != null)
 						{
-							var body  = ll.Body.Unwrap();
-							var parms = ll.Parameters.ToDictionary(p => p);
-							var ex    = body.Transform(wpi =>
+							var ue = (UnaryExpression)expr;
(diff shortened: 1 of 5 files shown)


jogibear9988:
I don't think the ArrayLength Expression Handling in Optimizer is correct... Maybe you have other suggestions how to solve.

MaceWindu (approved):

Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -496,10 +496,46 @@ public Expression ExposeExpression(Expression expression)
...
+
+									throw new LinqToDBException($"Can't convert {wpi} to expression.");
+								}
+
+								return wpi;
+							});
+
+							if (ex.Type != expr.Type)
+								ex = new ChangeTypeExpression(ex, expr.Type);
+
```
sdanyliv, line 535: I don't like this copy/paste. Better extract common function from MemberAccess handler.

Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -629,6 +615,37 @@ public Expression ExposeExpression(Expression expression)
 			_exposedCache[expression] = result;
 
 			return result;
+
+			static Expression convertMemberExpression(Expression expr, Expression root, LambdaExpression l)
```
sdanyliv, line 619:
```suggestion
			static Expression СonvertMemberExpression(Expression expr, Expression root, LambdaExpression l)
```

Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -516,32 +528,7 @@ public Expression ExposeExpression(Expression expression)
...
-
-										throw new LinqToDBException($"Can't convert {wpi} to expression.");
-									}
-
-									return wpi;
-								});
-
-								if (ex.Type != expr.Type)
-									ex = new ChangeTypeExpression(ex, expr.Type);
+								var ex = convertMemberExpression(expr, me.Expression!, l);
```
sdanyliv, line 531:
```suggestion
								var ex = СonvertMemberExpression(expr, me.Expression!, l);
```

Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -496,10 +496,22 @@ public Expression ExposeExpression(Expression expression)
...
 
 				switch (expr.NodeType)
 				{
+					case ExpressionType.ArrayLength:
+						{
+							var ue = (UnaryExpression)expr;
+							var ll = Expressions.ConvertMember(MappingSchema, ue.Operand?.Type, ue.Operand!.Type.GetProperty(nameof(Array.Length))!);
+							if (ll != null)
+							{
+								var ex = convertMemberExpression(expr, ue.Operand!, ll);
```
sdanyliv, line 509:
```suggestion
								var ex = СonvertMemberExpression(expr, ue.Operand!, ll);
```

sdanyliv (approved):

viceroypenguin (approved):

PR #2705, opened 2020-12-17
Null check in expression method transaltion to SQL

MaceWindu:
(not 3.2.0 regression)

generated sql miss null check:
```sql
 SELECT
    	IIF(EXISTS(
    		SELECT
    			*
    		FROM
    			[Person] [p]
    		WHERE
    			[p].[PersonID] = CAST(N'SHOULD NOT BE CALLED' AS INT)
    	), 1, 0)
```


MaceWindu:
Workaround - wrap null case into `Sql.AsSql`:
```cs
return value => value == null ? Sql.AsSql<int?>(null) : Fail(value.Value);
```

sdanyliv (approved):

PR #2713, opened 2020-12-19
Bad `IS NULL` predicate generation

MaceWindu:
Fails at least for SQL Server (will add list of dbs a bit later)


MaceWindu:
sql server
```sql
SELECT
    	IIF(EXISTS(
    		SELECT
    			*
    		FROM
    			[Person] [_]
    		WHERE
    			(IIF([_].[MiddleName] = N'123', 1, 0) = IIF(IIF([_].[MiddleName] = N'1', N'test', [_].[MiddleName]) = N'test', 1, 0) OR IIF([_].[MiddleName] = N'123', 1, 0) IS NULL AND (IIF([_].[MiddleName] = N'1', N'test', [_].[MiddleName]) = N'test') IS NULL)
    	), 1, 0)
```

MaceWindu:
Actually only sql server fail test

PR #2714, opened 2020-12-19
Refactor nugets generation

### LINKED ISSUE #1595: Pack.bat does not work
It is not possible to compile and create nuget packages from command line.

`
Attempting to build package from 'linq2db.MySql.nuspec'.
The element 'metadata' in namespace 'http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd' has incomplete content. List of possible elements expected: 'authors, version' in namespace 'http://schemas.microsoft.com/packaging/2010/07/nusp
ec.xsd'. This validation error occurred in a 'metadata' element.
`

#### Steps to reproduce
Just run _Pack.bat_ file.

MaceWindu:
- use latest nuget.exe for publishing (5.8)
- build nuspecs/nugets in separate folder (.\BuiltNuGet) to avoid nuspecs rewrite/folder pollution. Fix #1595
- deploy symbols in nupkg for Azure Artifacts and in snupkg for nuget
- cleanup .gitignore from most of irrelevant stuff


Build/Azure/pipelines/default.yml
```diff
@@ -83,25 +87,33 @@ stages:
       inputs:
         filePath: '$(Build.SourcesDirectory)/Build/BuildNuspecs.ps1'
         workingDirectory: '$(Build.SourcesDirectory)'
-        arguments: -path $(Build.SourcesDirectory)/NuGet/*.nuspec -version $(packageVersion) -branch $(Build.SourceBranchName)
+        arguments: -path $(Build.SourcesDirectory)/NuGet/*.nuspec -buildPath $(Build.SourcesDirectory)/BuiltNuGet -version $(packageVersion) -branch $(Build.SourceBranchName)
```
viceroypenguin, line 90: Is there no $(Build.Artifacts) directory or $(Build.Output) or similar that is outside of the Sources subfolder? Then you won't pollute the source directory at all.

viceroypenguin (commented):
Can we download nuget.exe from nuget.org instead of hard-coding a binary in the source tree? Maybe use https://dist.nuget.org/win-x86-commandline/latest/nuget.exe so we don't have to be version specific?

MaceWindu:
nuget.exe here for local builds. On azure it is already installed.

Build/Azure/pipelines/default.yml
MaceWindu: That's for local build (see #1595). For azure we don't care about files location as it is disposable VM anyway

viceroypenguin:
> nuget.exe here for local builds. On azure it is already installed.

I suppose I'm good with that then.

MaceWindu:
I would be happy to remove nuget.exe from repo, but it is not installed by default even with visual studio (it use nuget libaries instead), so it means people need to install it and add to path somehow.

Build/Azure/pipelines/default.yml
viceroypenguin: Ah. Ok. I'm used to build where they re-use the source location between builds and deliver output to other folders to make the git cleanup easier.

viceroypenguin (approved):

viceroypenguin:
Suggestion for not including the nuget file. Look for the file in that location. If it doesn't exist, prompt the user "We need the latest nuget.exe at location X in order to generate nuget packages. Would you like us to automatically download this file?", and then download from the url to that location. If the user wants to manually download it and put it there, they can do that too, since the prompt will give them the location, etc. Add nuget.exe to the .gitignore so people don't accidentally try to check the file in too.

sdanyliv (approved):
I hope it will work )

MaceWindu:
@viceroypenguin  you won :smile:

jogibear9988:
How about adding nuget: https://www.nuget.org/packages/NuGet.CommandLine/

viceroypenguin:
@MaceWindu haha! Later today, I'll look and see if there's a way we can add auto download if user says yes. What's there works for me though.

PR #2716, opened 2020-12-20
Incorrect GROUP BY optimization when applied to subquery

MaceWindu:
count moved inside subquery, which produce incorrect results
```sql
   SELECT
    	1,
    	Count(*)
    FROM
    	[Person] [_]
    WHERE
    	1 = 0
```


MaceWindu:
Shouldn't we exclude constants from group by if there are non-constant group-by expressions defined? I see from baselines that it re-introduced constants into group by for such queries.

MaceWindu:
Anyway, I'm fine with it for now, as we need this fix released.

sdanyliv:
Actually it is correct
```cs
select count(*) from table
```
always return one record, but with group by - it can be 0 or N records.

Issue #2676, opened 2020-12-07
(Question) Sometimes ambient transaction is ignored


Powerz:
Hi guys,

I have a small question regarding the way Linq2Db treats ambient transactions (TransactionScope).

In the example (see below), if the first query runs before an instance of transaction scope is created, then the ambient transaction will be ignored. However, if the first query runs after the transaction scope is created, then the transaction will be respected.

This behavior looks more like a bug (otherwise, in my view, it deserves to be documented), but maybe it was a design decision. So is this a bug? (and if it is, would you guys fix it, please?)

### Steps to reproduce
Comment and uncomment the first line in the example posted below.

```c#
// if uncomment this line, the ambient transactions will be ignored => a new record will be committed to the database
// without this line, the ambient transaction will work => no new data will be committed to the database

// var user = await _db.Table<User>().FirstAsync();

using var transaction = new TransactionScope(TransactionScopeOption.Required, TransactionScopeAsyncFlowOption.Enabled);

await _db.InsertAsync(new User
                      {
                          Name = Guid.NewGuid().ToString(),
                          Email = $"{Guid.NewGuid()}@test.com"
                      });

throw new ApplicationException();
transaction.Complete();
```

### Environment details
linq2db version: 3.1.2, 3.1.6
Database Server: Azure Sql Server
Database Provider: SQL Server
Operating system: Windows, Mac OS
.NET Framework: .net core 3.1.300, 5.0

MaceWindu:
In second case you create `TransactionScope` after connection already created. In such cases you need to enlist transaction in connection explicitly:
```cs
((DbConnection)_db.Connection).EnlistTransaction(Transaction.Current);
```

MaceWindu:
linq2db doesn't call `EnlistTransaction` for user.

And I would say we shouldn't  do it. IMHO `TransactionScope` was bad design example from Microsoft.

@linq2db what do you think?

viceroypenguin:
I would say that l2db should either completely ignore `TransactionScope` or respect the expected behavior (that creating a `TransactionScope` after the initial query still gets enlisted when the first query after the TS is created). Personally, I just use `DataConnection.CreateTransactionAsync()` and `.CommitTransactionAsync()`, since it's more convenient for async behavior than `TransactionScope`.

Maybe this could be the next analyzer - identifying uses of `TransactionScope` and issuing a warning that behavior may not be as expected.

MaceWindu:
[...]

Issue #86, opened 2014-06-03
Recognise bit(1) as Bool


HeatFlex:
MySQL Schema Provider must recognise column type bit(1) as Boolean. Currently it recognises bit(1) as ulong.

MaceWindu:
I'm reverting this change, as `x` in mysql `tinyint(x)` doesn't affect allowed range and should be mapped to byte/sbyte

MaceWindu:
Note to myself:
add option to map tinyint(1) to bool for backward compatibility

IceReaper:
The current approach is:
```csharp
case "tinyint":
{
	var size = precision > 0 ? precision : length;
	if (columnType == "tinyint(1)" || size == 1)
		return typeof(bool);
	return columnType?.Contains("unsigned") == true ? typeof(byte) : typeof(sbyte);
}
//case "tinyint"           : return columnType?.Contains("unsigned") == true ? typeof(byte)   : typeof(sbyte);
```

In our software, we (need to, please dont suggest a workaround) parse a database to generate the table classes on runtime.
Our customer has a column `tinyint(1)` where he stores values `-9 to 9`. As it is `not unsigned`, linq2db keeps generating `bool` as its datatype, which is clearly wrong. Basicaly the correct answer would be the commented out line.

If the case of interpreting `tinyint(1)` as `bool` is required, a compatibility flag must be set.

Currently it interprets `tinyint(1)` as `bool` without any option to override, which will `always break` and `cause data loss` when using linq2db, as the result is interpreted as a `bool`, overwriting any other value than 0 with 1.

MaceWindu:
Scaffold utility allows you to owerride default type mappings https://linq2db.github.io/articles/CLI.html#type-mapping-interceptor

IceReaper:
We cannot use any CLI stuff in our application workflow. Can we somehow specify that in this particular usecase?
```csharp
await using var connection = new DataConnection(model.GetProviderName(), model.ToConnectionString());
await connection.EnsureConnectionAsync(cancellationToken);
return connection.DataProvider.GetSchemaProvider().GetSchema(connection, new() { GetProcedures = false });
```

MaceWindu:
Nope, you need to postprocess GetSchema outputs

MaceWindu:
Decided to not change it for now

PR #2728, opened 2020-12-22
Fix multiple active resultsets per-connection support for supporting providers

### LINKED ISSUE #2643: Exception: There is already an open DataReader associated with this Command which must be closed first
.Net Version 4.6.1
Microsoft.Data.SqlClient 2.01
linq2db 3.1.6

**With Oracle:**
no Problem
**With SQL-Server:**
Exception: There is already an open DataReader associated with this Command which must be closed first.

**Code:**
```cs
using (DbContext dbContext = new DbContext(dbProvider))
{
  IEnumerable<string> list = dbContext.MENUS_LOG.Select(obj => obj.DATATYPE).Distinct();
   foreach(string strItem in list)
   {
      dbContext.MENUS_LOG.Where(objWhere => objWhere.DATATYPE == strItem)
                .Update(dbContext.MENUS_LOG, objWhere => new MENUS_LOG{ DATATYPE = strItem + "X"});
   }
}
```

### DataProvider:
**Oracle:**
```cs
m_IDataProvider = new LinqToDB.DataProvider.Oracle.OracleDataProvider(ProviderName.OracleManaged,
LinqToDB.DataProvider.Oracle.OracleVersion.v12);
```

**SQL-Server:**
```cs
m_IDataProvider = new LinqToDB.DataProvider.SqlServer.SqlServerDataProvider(
ProviderName.SqlServer,
LinqToDB.DataProvider.SqlServer.SqlServerVersion.v2017,
LinqToDB.DataProvider.SqlServer.SqlServerProvider.MicrosoftDataSqlClient);
```

### ConnectionString:
**Oracle:**
```cs
@"USER ID=UserDummy;CONNECTION TIMEOUT=60;PASSWORD=XXXXXX;DATA SOURCE=PKSABNAHME12;STATEMENT CACHE SIZE=0;CONNECTION LIFETIME=0;INCR POOL SIZE=1;STATEMENT CACHE PURGE=True";
```

**SQL-Server:**
```cs
@"Data Source=MSTEST17\MSTEST17;Initial Catalog=DUMMYDB;Integrated Security=True;Connect Timeout=60;Multiple Active Result Sets=True;Multi Subnet Failover=True;";
```

**Exception:**
```
System.InvalidOperationException
HResult=0x80131509
Message=There is already an open DataReader associated with this Command which must be closed first.
Source=Microsoft.Data.SqlClient
StackTrace:
at Microsoft.Data.SqlClient.SqlInternalConnectionTds.ValidateConnectionForExecute(SqlCommand command)
at Microsoft.Data.SqlClient.SqlConnection.ValidateConnectionForExecute(String method, SqlCommand command)
at Microsoft.Data.SqlClient.SqlCommand.ValidateCommand(String method, Boolean async)
at Microsoft.Data.SqlClient.SqlCommand.InternalExecuteNonQuery(TaskCompletionSource1 completion, String methodName, Boolean sendToPipe, Int32 timeout, Boolean& usedCache, Boolean asyncWrite, Boolean inRetry) at Microsoft.Data.SqlClient.SqlCommand.ExecuteNonQuery() at LinqToDB.Data.DataConnection.ExecuteNonQuery(IDbCommand command) at LinqToDB.Data.DataConnection.ExecuteNonQuery() at LinqToDB.Data.DataConnection.QueryRunner.ExecuteNonQueryImpl(DataConnection dataConnection, PreparedQuery preparedQuery) at LinqToDB.Data.DataConnection.QueryRunner.ExecuteNonQuery() at LinqToDB.DataContext.QueryRunner.ExecuteNonQuery() at LinqToDB.Linq.QueryRunner.NonQueryQuery(Query query, IDataContext dataContext, Expression expr, Object[] parameters, Object[] preambles) at LinqToDB.Linq.QueryRunner.<>c__DisplayClass32_0.<SetNonQueryQuery>b__0(IDataContext db, Expression expr, Object[] ps, Object[] preambles) at LinqToDB.Linq.ExpressionQuery1.System.Linq.IQueryProvider.Execute[TResult](Expression expression)
at LinqToDB.LinqExtensions.Update[TSource,TTarget](IQueryable1 source, ITable1 target, Expression`1 setter)
```

MaceWindu:
Fix #2643

Providers tests shown that some providers:
- some providers support mutliple active data readers created from single command
- some providers support multiple active data readers if each data reader use own command
- some providers doesn't suppoort multiple data readers in any form on single connection (mysql.data, mysqlconnector, npgsql, sql server client in non-mars mode)
- some providers dont like if command associated with data reader disposed before we finished reading data from data reader
- some providers clear parameters collection if command disposed

To not introduce new configuration settings this PR contain following changes that work for all providers:
- we will not reuse command, that used to create data reader query, but instead of it we will dispose this command when data reader disposed
- ~we will copy command parameters to `DataConnection.LastParameters` propery instead of setting it to parameters collection from command. This will preserve last parameters even if command was disposed~ `LastParameters` is gone to Oblivion
- `DataConnection.Command` property becames even less useful with those changes and even misleading, so we obsolete it (or remove, it this PR will be postponed till v4 release)
- to provide better replacement for `DataConnection.Command`, introduced new event `DataConnection.OnCommandInitialized`, triggered after command initialized with sql and parameters. Event allows command replacement.

Other changes:
- call `DisposeAsync` on data reader instead of `Dispose` in more places
- disable native bulk copy for Sybase against temp table (not supported by provider)

Remaining tasks:
- [x] fix `OnCommandInitialized` event to fire after parameters set
- [x] get rid of `DataConnection.LastParameters` property


Source/LinqToDB/DataProvider/Access/AccessODBCDataProvider.cs
```diff
@@ -68,6 +68,11 @@ public override ISchemaProvider GetSchemaProvider()
 			return new AccessODBCSchemaProvider();
 		}
 
+		public override IDisposable? ExecuteScope(DataConnection dataConnection)
+		{
+			return new CallOnDisposeRegion(() => dataConnection.DisposeCommand());
```
sdanyliv, line 73: Avoid not needed closure allocation. Pass dataConnection into constructor. Same places below.

Source/LinqToDB/DataProvider/Sybase/SybaseDataProvider.cs
```diff
@@ -108,6 +108,17 @@ public override ISchemaProvider GetSchemaProvider()
 			return new SybaseSchemaProvider(this);
 		}
 
+		public override IDisposable? ExecuteScope(DataConnection dataConnection)
+		{
+			// https://github.com/linq2db/linq2db/issues/2643
+			// also don't do it for managed provider:
+			// https://github.com/DataAction/AdoNetCore.AseClient/issues/203
+			if (Name == ProviderName.Sybase)
+				return new CallOnDisposeRegion(() => dataConnection.DisposeCommand());
```
sdanyliv, line 117: Add provider flag for this.

sdanyliv (changes requested):

Tests/Linq/Data/DataConnectionTests.cs
```diff
@@ -1213,7 +1213,136 @@ public class Product
...
+				TestProvName.AllSqlServer,
+				TestProvName.AllSybase,
+				ProviderName.SQLiteMS,
+				TestProvName.AllSQLiteClassic,
+				ProviderName.SqlCe,
+				TestProvName.AllOracle,
+				TestProvName.AllAccess,
+				ProviderName.SapHanaNative,
+				ProviderName.DB2)]
+		string context)
```
viceroypenguin, line 1230:
```suggestion
				string context)
```

viceroypenguin, line 1230:
```suggestion
				ProviderName.DB2)] string context)
```

Tests/Linq/Data/DataConnectionTests.cs
```diff
@@ -1213,7 +1213,136 @@ public class Product
...
```
viceroypenguin, line 1295:
```suggestion
				ProviderName.DB2)] string context)
```

viceroypenguin (approved):

MaceWindu:
- moved command dispose to provider flags
- added recently added provider flags to flags equals/gethashcode
- fixed MARS support for sap hana odbc and informix providers and moved them to supported list

MaceWindu:
on hold, implementation uncovered unexpected expectations in our code to have command object in some places :)

MaceWindu:
moving it to 3.3 as command disposal makes command unavailable on data connection after query and we need to review depending usecases carefully. And maybe introduce some extra API to expose command data in such cases.

bgrainger:
Following on from https://github.com/npgsql/npgsql/issues/462#issuecomment-756787766:

> PostgreSQL simply doesn't have a mechanism that would allow an efficient MARS implementation.

This is also true for the MySQL database protocol. After a query is sent to the MySQL server by the client, the server starts streaming the entire result set back to the client. Until the "EOF" packet is received (from the server), no other client/server communication can take place. At that point, the first result set is no longer "active", so it's never possible to have _multiple_ active result sets.

The best you can do is buffer all rows in-memory on the client. Due to performance/memory usage concerns, this should be an explicit decision by the application (which can be as simple as `.Query<T>(...).ToList()` in Dapper).

MaceWindu:
Changes to DataConnection.Command.

Because with command disposal and recreation with MARS enabled, DataConnection.Command property becomes less useful and predictable, I've deprecated it alongsize with two other public API for command management:
- `Command {get;set;}`
- `DisposeCommand()`: internal API with public access modifier, replaced with internal `DisposeCurrentCommand()`
- `CreateCommand()`: internal API with public access modifier, replaced with internal `CreateNewCommand()`

Alternatives for disposed APIs:
- access to command parameters: alteady existing `DataConnection.LastParameters` (also fixed T4 generation for stored procedures to use this property)
- access to command SQL: already existing property `DataConnectionLastQuery`
- new `DataConnection.OnCommandInitialized` event added. Event triggered after command initialization and allow command inspection and replacement
- existing `DbCommandProcessorExtensions.Instance` extension point

Other changes:
- following `Command` property deprecation, most of code that used it switched to non-creating `CurrentCommand` property
- fixed `DataConnection.ExecuteNonQuery` to not ignore command parameter
- MARS support disabled by default and hidden behind `Configuration.EnableMultipleActiveResultSetsSupport` option.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -41,6 +41,23 @@ public static class Configuration
...
+		/// <list type="bullet">
+		/// <item>MySql.Data</item>
+		/// <item>MySqlConnector</item>
+		/// <item>Npgsql</item>
+		/// <item>System.Data.SqlClient (without MARS enabled in connection string)</item>
+		/// <item>Microsoft.Data.SqlClient (without MARS enabled in connection string)</item>
+		/// <item>FirebirdSql.Data.FirebirdClient</item>
+		/// </list>
+		/// Note that enabling this feature will also dispose <see cref="DataConnection.Command"/> property after command
+		/// execution.
```
viceroypenguin, line 56:
```suggestion
		/// Note that enabling this feature means the <see cref="DataConnection.Command"/> will be
		/// disposed after every command execution.
```

Tests/Linq/Data/DataConnectionTests.cs
```diff
@@ -1214,8 +1214,130 @@ public class Product
...
+			}
+
+			try
+			{
+				using (var db = new TestDataConnection(context))
+				{
+					db.GetTable<TransactionScopeTable>().Insert(() => new TransactionScopeTable() { Id = 1 });
+					using (var transaction = new TransactionScope(TransactionScopeOption.Required, TransactionScopeAsyncFlowOption.Enabled))
+					{
+						// this query will be executed outside of TransactionScope transaction as it wasn't enlisted into connection
```
viceroypenguin: Maybe add another comment that this may be updated per #2676 ?

Source/LinqToDB/SqlProvider/SqlProviderFlags.cs
```diff
@@ -30,6 +30,17 @@ public class SqlProviderFlags
 		public bool        IsUpdateSetTableAliasSupported { get; set; }
 		public TakeHints?  TakeHintsSupported             { get; set; }
 
+		/// <summary>
+		/// Provider allows multiple active recordsets (data readers) within one connection, if associated with
+		/// different <see cref="IDbCommand"/> instances.
+		/// </summary>
+		public bool        SupportsMARSWithNewCommand     { get; set; }
```
viceroypenguin, line 37: Should be `SupportsMarsWithNewCommand`, per naming convention: "The PascalCasing convention, used for all identifiers except parameter names, capitalizes the first character of each word (including acronyms over two letters in length)" [here](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions)

Source/LinqToDB/Common/Configuration.cs
viceroypenguin: Also, maybe this sentence goes into the `<remarks>` tag instead of the `<summary>` tag?

MaceWindu:
interesting side-effect - when both command and data reader disposed, System.Data.SQLite and native sybase provider clear parameters collection, so LastParameters is empty when mars option enabled.
Which means this option in not compatible with output parameters for stored procedures. Actual only for native sybase provider, so it is not a big deal, but maybe we should copy parameters for this provider with mars enabled?

MaceWindu:
ADO.NET is fun (not really).

sdanyliv:
@MaceWindu, probably it is better to change `LastParameters` to `DataParameter[]?` type?

MaceWindu:
@sdanyliv I don't want to allocate array for each query. Taking into account sqlite/sybase providers support I will move command disposal to init phase of next command instead.

MaceWindu:
@linq2db, I really want to see this PR in next release if we agree on breaking changes:
- users requre to rerun their T4 templates if they have stored procedures in model
- `DataConnection.Command` property is obsoleted and will contain no data after data reader (select) queries (as it will contain fresh instance of command)

viceroypenguin:
Depending on how many users have stored procedures in their model (I don't, but I know there are plenty of environments where sprocs are required by DBAs, etc.), #1 is probably enough of a breaking change to require 4.0. I make no opinion on whether _this_ release should be 4.0 vs this waits until we're ready for 4.0; only that this should not be in a point release for the library. Or at least for the .ttinclude packages.

Secondary concern: what about users that use sprocs but don't use the model to generate? Will they be forced to update their hand-generated models?

MaceWindu:
> what about users that use sprocs but don't use the model to generate? Will they be forced to update their hand-generated models?

if they used `Command` property to access output parameters then yes for two reasons:
- they will get obsolete warnings
- if they used `QueryProc` API, that returns data sets, `Command` will be useless as it will be erased with new instance

viceroypenguin:
Then I'd say mark `[Obsolete]` for v4 and remove in v5. This requires users to make changes for correct operation, so we should use v4 to indicate to users that they should be aware and make changes if necessary. IMHO. Open to other opinions though.

MaceWindu:
Ok, then I think we will do following:
- cherrypick various small fixes from this pr (missing ConfigureAwait, DisposeAsync, sybase bulk copy fixes, etc) into 3.3
- create v4 branch, where we will accumulate changes for v4 and configure publishing to azure artifacts

MaceWindu:
Also I'm reverting #86 here, as 1 in `tinyint(1)` for mysql doesn't mean it could have only 1 bit of data.

Source/LinqToDB/Data/DataConnection.QueryRunner.cs
```diff
@@ -16,6 +16,7 @@ namespace LinqToDB.Data
 	using Common;
 	using SqlQuery;
 	using SqlProvider;
+	using LinqToDB.DataProvider;
```
sdanyliv, line 19: Move inside namespace.

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1218,47 +1232,55 @@ public  int   CommandTimeout
...
-		private IDbCommand? _command;
-
 		/// <summary>
-		/// Provides acess to current <see cref="_command"/> instance. Used for logs to avoid command instance creation
-		/// in <see cref="Command"/> getter if <see cref="_command"/> is not initialized.
+		/// Event, triggered after command initialized, but not yet executed.
+		/// Event could be used for command inspection, modification or replacement.
 		/// </summary>
-		internal IDbCommand? GetCurrentCommand() => _command;
+		public event OnCommandInitializedEventHandler? OnCommandInitialized;
```
sdanyliv, line 1239: Think about interceptors, looks like DataConnection has too much events.

jods4: Note that `Command` was useful as an easy way to setup some provider-specific properties in a simple one-shot way for the next query.  For example, set `FetchSize` on an `OracleCommand`.

This would still be possible with `OnCommandInitialized` and is somewhat more intuitive than knowing you could manipulate the next command, but it's also more convoluted.

For a one-shot parameter, you need to subscribe the event, write the actual code, then unsubscribe (ideally in a `finally` block in case something goes wrong but the `DataConnection` can be used further).

Interceptors are a powerful concept but it would be nice to also have a simple API to modify the next query.
For example, maybe:
```csharp
db.NextCommand(c => (c as OracleCommand).FetchSize = 100));
// Do something
db.Table.ToList();
```

MaceWindu:
Notes on implementation:
- interceptors supported on both `DataConnection` and `DataContext` instances
- interceptors supported by `DataConnection` fluent configuration builder
- to leverage `DataConnection.Command` removal migration issues, mentioned by @jods4 (and to support single query-scoped cases), I've added extensions that provide access to command using one-time interceptors
- while implementation is inspired by `EF.Core` interceptors, it is different, especially in aggregation part to support one-time interceptors
- aggregation, removal and one-time events infrastructure is internal implementation and not exposed to users

MaceWindu:
As requested I will add one more interceptor to this PR to better demonstrate this functionality

MaceWindu:
- replaced `OnBeforeConnectionOpen`, `OnBeforeConnectionOpenAsync`, `OnConnectionOpened`, `OnConnectionOpenedAsync` events on `DataConnection` with interceptor.
- I didn't added one-time helpers for those events as I don't see good use-case for them right now
- [functionality degradation]: it was possible to unsubscribe from events but we don't allow interceptor removal. I don't see it as a big isssue without seeing real use-case for it (same as previous entry). Also user could emulate required behavior by setting disable flag it in custom interceptor
- found out that `DataContext.CloseAfterUse` actually wasn't working properly and were disposing connection always. Fixed. We should cherrypick it to master

viceroypenguin:
> * ~we will copy command parameters to `DataConnection.LastParameters` propery instead of setting it to parameters collection from command. This will preserve last parameters even if command was disposed~ `LastParameters` is gone to Oblivion
> * `DataConnection.Command` property becames even less useful with those changes and even misleading, so we obsolete it (or remove, it this PR will be postponed till v4 release)

Suggestions: Per the above, we should add another PR to `Obsolete` both of these things in 3.4. I'm not opposed to removing in 4.0, given major version num and breaking changes are allowed, but would like to have a little lead time for the (probably rare) users that are using those things.

Source/LinqToDB/Configuration/LinqToDbConnectionOptions.cs
```diff
@@ -111,6 +114,10 @@ internal LinqToDbConnectionOptions()
 		/// Gets custom trace writer to use with <see cref="DataConnection"/> instance.
 		/// </summary>
 		public Action<string?, string?, TraceLevel>? WriteTrace          { get; }
+		/// <summary>
+		/// Gets list of interceptors to use with <see cref="DataConnection"/> instance.
+		/// </summary>
+		public List<IInterceptor>?                   Interceptors        { get; }
```
viceroypenguin, line 120: Question: do we want a user to be able to edit this list after it's been assigned here? Perhaps it should be `IReadOnlyList<>` to indicate inflexibility?

Source/LinqToDB/Configuration/LinqToDbConnectionOptionsBuilder.cs
```diff
@@ -27,6 +29,7 @@ public class LinqToDbConnectionOptionsBuilder
 		public Action<TraceInfo>?                    OnTrace             { get; private set; }
 		public TraceLevel?                           TraceLevel          { get; private set; }
 		public Action<string?, string?, TraceLevel>? WriteTrace          { get; private set; }
+		public List<IInterceptor>?                   Interceptors        { get; private set; }
```
viceroypenguin, line 32: Question: do we want a user to be able to edit this list after it's been assigned here? Perhaps we should use a `private List<>` so we can add to it, but public api is `public IReadOnlyList<>? Interceptors => _interceptors;`.

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -355,32 +331,32 @@ IEnumerable<T> ReadEnumerator<T>(IDataReader rd, IDisposable? scope, bool dispos
...
 							}
 
 							yield return result;
 
-						} while (rd.Read());
+						} while (rd.DataReader!.Read());
 					}
+
+					if (rd.Command?.Parameters.Count > 0 && Parameters?.Length > 0)
+						rd.OnBeforeCommandDispose = RebindParameters;
```
viceroypenguin, line 353: These two lines are repeated throughout the file. Should abstract to `SetRebindParameters(rd);`

MaceWindu:
> Suggestions: Per the above, we should add another PR to Obsolete both of these things in 3.4.

We cannot obsolete them as we cannot provide alternative in v3 for it

sdanyliv (approved):

PR #2756, opened 2021-01-03
Release 3.2.3

MaceWindu:


MaceWindu:
windows sql 2019 tests fail for windows as we use express version here, which doesn't support `CATALOG_COLLATION` option. will fix after release

PR #2774, opened 2021-01-16
Lower allocations to improve ConvertBinaryToSql performance

to11mtm:
When profiling some benchmarks I noticed that there was a lot of traffic around `.ConvertBinaryToSql` (especially in bulk insert scenarios). 

I did some research and adapted [this](https://stackoverflow.com/a/24343727/2937845) into something that worked with StringBuilder.

Doing 'before/after' tests with this change show a pretty big improvement on paths where `.ConvertBinaryToSql` is called (with payloads in the 512b-1kb range, as much as 30% in a library benchmark against fast local IO)

### DIFF
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -45,18 +45,26 @@ namespace LinqToDB.Common
 		
 		/// <summary>
 		/// Determines the length after which logging of binary data in SQL will be truncated.
-		/// Use a value of -1 to disable. Set to 0 to truncate all binary data.
 		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
 		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
 		/// </summary>
+		/// <remarks>
+		/// This value defaults to 100.
+		/// Use a value of -1 to disable and always log full binary.
+		/// Set to 0 to truncate all binary data.
+		/// </remarks>
 		public static int MaxBinaryParameterLengthLogging { get; set; } = 100;
 
 		/// <summary>
 		/// Determines the length after which logging of string data in SQL will be truncated.
-		/// Use a value of -1 to disable. Set to 0 to truncate all binary data.
 		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
 		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
 		/// </summary>
+		/// <remarks>
+		/// This value defaults to 200.
+		/// Use a value of -1 to disable and always log full string.
+		/// Set to 0 to truncate all string data.
+		/// </remarks>
 		public static int MaxStringParameterLengthLogging { get; set; } = 200;
 
 		public static class Data
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 
 using JetBrains.Annotations;
+using LinqToDB.Linq;
 
 namespace LinqToDB.Common
 {
@@ -40,6 +41,13 @@ namespace LinqToDB.Common
 		/// queries, so this optimization could be used for <see cref="CommandBehavior.Default"/> too.
 		/// </summary>
 		public static bool OptimizeForSequentialAccess = false;
+		
+		/// <summary>
+		/// Determines the length after which logging of binary data in SQL will be truncated.
+		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
+		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
+		/// </summary>
+		public static int MaxByteLengthLogging { get; set; } = 1024;
 
 		public static class Data
 		{
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -47,7 +47,14 @@ namespace LinqToDB.Common
 		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
 		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
 		/// </summary>
-		public static int MaxByteLengthLogging { get; set; } = 1024;
+		public static int MaxBinaryParameterLengthLogging { get; set; } = 100;
+
+		/// <summary>
+		/// Determines the length after which logging of string data in SQL will be truncated.
+		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
+		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
+		/// </summary>
+		public static int MaxStringParameterLengthLogging { get; set; } = 200;
 
 		public static class Data
 		{
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -1,10 +1,11 @@
 ﻿using System;
 
 using JetBrains.Annotations;
-using LinqToDB.Linq;
+
 
 namespace LinqToDB.Common
 {
+	using LinqToDB.Linq;
 	using Data;
 	using Data.RetryPolicy;
 	using System.Data;
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -44,6 +44,7 @@ namespace LinqToDB.Common
 		
 		/// <summary>
 		/// Determines the length after which logging of binary data in SQL will be truncated.
+		/// Use a value of -1 to disable. Set to 0 to truncate all binary data.
 		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
 		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
 		/// </summary>
@@ -51,6 +52,7 @@ namespace LinqToDB.Common
 
 		/// <summary>
 		/// Determines the length after which logging of string data in SQL will be truncated.
+		/// Use a value of -1 to disable. Set to 0 to truncate all binary data.
 		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
 		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
 		/// </summary>
--- a/Source/LinqToDB/Common/StringBuilderExtensions.cs
+++ b/Source/LinqToDB/Common/StringBuilderExtensions.cs
@@ -12,7 +12,7 @@ namespace LinqToDB.Common
 			var result = new uint[256];
 			for (int i = 0; i < 256; i++)
 			{
-				string s =i.ToString("X2");
+				string s = i.ToString("X2");
 				result[i] = ((uint)s[0]) + ((uint)s[1] << 16);
 			}
 			return result;
--- a/Source/LinqToDB/Common/StringBuilderExtensions.cs
+++ b/Source/LinqToDB/Common/StringBuilderExtensions.cs
@@ -35,9 +35,9 @@ namespace LinqToDB.Common
 		public static void AppendByteArrayAsHexViaLookup32(this StringBuilder sb, byte[] bytes)
 		{
 			var lookup32 = _lookup32;
-			for (int i = 0; i < bytes.Length; i++)
+			foreach (var b in bytes)
 			{
-				var val = lookup32[bytes[i]];
+				var val = lookup32[b];
 				sb.Append((char)val);
 				sb.Append((char) (val >> 16));
 			}
--- a/Source/LinqToDB/Common/StringBuilderExtensions.cs
+++ b/Source/LinqToDB/Common/StringBuilderExtensions.cs
@@ -1,0 +1,46 @@
+﻿using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace LinqToDB.Common
+{
+	public static class StringBuilderExtensions
+	{
+		private static readonly uint[] _lookup32 = CreateLookup32();
+        
+		private static uint[] CreateLookup32()
+		{
+			var result = new uint[256];
+			for (int i = 0; i < 256; i++)
+			{
+				string s =i.ToString("X2");
+				result[i] = ((uint)s[0]) + ((uint)s[1] << 16);
+			}
+			return result;
+		}
+        
+		/// <summary>
+		/// Appends an array of bytes to a <see cref="StringBuilder"/> in hex (i.e. 255->FF)
+		/// format utilizing a static lookup table to minimize allocations.
+		/// </summary>
+		/// <param name="sb">The <see cref="StringBuilder"/> to append to</param>
+		/// <param name="bytes">The byte array to append in hex</param>
+		/// <remarks>
+		/// The implementation here was chosen based on:
+		/// https://stackoverflow.com/a/624379/2937845
+		/// Which indicated that  https://stackoverflow.com/a/24343727/2937845's
+		/// implementation of ByteArrayToHexViaLookup32 was the fastest method
+		/// not involving unsafe 
+		/// </remarks>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static void AppendByteArrayAsHexViaLookup32(this StringBuilder sb, byte[] bytes)
+		{
+			var lookup32 = _lookup32;
+			for (int i = 0; i < bytes.Length; i++)
+			{
+				var val = lookup32[bytes[i]];
+				sb.Append((char)val);
+				sb.Append((char) (val >> 16));
+			}
+		}
+	}
+}
--- a/Source/LinqToDB/DataProvider/Access/AccessMappingSchema.cs
+++ b/Source/LinqToDB/DataProvider/Access/AccessMappingSchema.cs
@@ -1,9 +1,10 @@
 ﻿using System;
 using System.Text;
-using LinqToDB.Common;
+
 
 namespace LinqToDB.DataProvider.Access
 {
+	using Common;
 	using Mapping;
 	using SqlQuery;
 	using System.Data.Linq;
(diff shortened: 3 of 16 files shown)


Source/LinqToDB/Common/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static void AppendByteArrayAsHexViaLookup32(this StringBuilder sb, byte[] bytes)
+		{
+			var lookup32 = _lookup32;
+			for (int i = 0; i < bytes.Length; i++)
+			{
+				var val = lookup32[bytes[i]];
+				sb.Append((char)val);
+				sb.Append((char) (val >> 16));
+			}
```
viceroypenguin, line 43: Can you please review and compare if this might be better? I believe that there are optimizations the JIT does on a `foreach` on an array vs using an index.
```suggestion
			foreach (var b in bytes
			{
				var val = lookup32[b];
				sb.Append((char)val);
				sb.Append((char) (val >> 16));
			}
```

to11mtm: It looks like it's either a wash or a little bit worse. 

On one hand, the generated ASM is essentially identical or only one byte better for the `foreach` version.

On the other hand, `foreach` comes in with ~6 more bytes of IL. I tossed both versions into sharplab:

https://sharplab.io/#gist:360c05b61e74e8c6ecdec7a3ee34797c

MaceWindu:
BTW, maybe you want to address another issue in this area? When we generate sql statement log, we write parameter values and for binary parameters it generates full literal, which leads to OOM-exception in string builder on big binaries.
There is no big value in binary parameter value logged and I propose to log only first N bytes for it (maybe with some option like `long MaxBinaryParameterLogSize` with -1 for unlimited loging)

Source/LinqToDB/Common/StringBuilderExtensions.cs
viceroypenguin: Regarding benchmarks: You should always use Benchmark.NET to run benchmarks. The library is written to correctly test micro-benchmarks and has tools to manage various things that can affect benchmark performance. Here's an updated copy of your code w/ Benchmark.NET: http://share.linqpad.net/x3j7fk.linq On my machine, the `foreach` is .1us faster, but within the margin of error. I would say it is worth the difference (given how negligible the difference is) in either direction to have more idiomatic code (i.e. use the `foreach`), but that's my personal opinion.

to11mtm: >I would say it is worth the difference (given how negligible the difference is) in either direction to have more idiomatic code (i.e. use the foreach), but that's my personal opinion.

I'm fine with switching it to `foreach` given the lack of difference, but if we are adding something like @MaceWindu 's suggested `MaxBinaryParameterLogSize` for SqlText logging purposes we may need to stick with the for loop

to11mtm:
> 
> 
> BTW, maybe you want to address another issue in this area? When we generate sql statement log, we write parameter values and for binary parameters it generates full literal, which leads to OOM-exception in string builder on big binaries.
> There is no big value in binary parameter value logged and I propose to log only first N bytes for it (maybe with some option like `long MaxBinaryParameterLogSize` with -1 for unlimited logging)

I wouldn't mind giving it an attempt, but peeking at how we are building those out I would have to take a deep dive at things to make sure change only impacts fetching for logging versus when we require the built SQL. May take more time.

Are `.GetSqlText` and/or `PrintParameters` ever used for actual SQL sent To DB and not just debugging/etc? if answer is that they are never used for actually sending for DB, I had idea; in `PrintParameters`, if item is byte array, we could send in either an `ArraySegment` (Net45/etc) or `Span<char>` (NetStandard2.0+) to an overload with the span/segment sliced to appropriate length based on setting.

 - in `PrintParameters`, if item is byte array, use

MaceWindu:
> Are .GetSqlText and/or PrintParameters ever used for actual SQL sent To DB and not just debugging/etc? 

nope, this is intended for debug/logging only.

MaceWindu:
It would be nice to have some triming indication in log, like ellipsis after parameter value

to11mtm:
>  some triming indication in log, like ellipsis after parameter value

I think easiest thing to do at least for `PrintParameters` would be something like this; it avoids having to poke at `ValueToSqlBuilder` too much:

```
foreach (var p in parameters)
				{
					sb.Append("DECLARE ");
					PrintParameterName(sb, p);
					sb.Append(' ');
					PrintParameterType(sb, p);
					sb.AppendLine();

					sb.Append("SET     ");
					PrintParameterName(sb, p);
					sb.Append(" = ");
					if (p.Value is byte[] bytes                           &&
					    bytes.Length > Configuration.MaxByteLengthLogging &&
					    ValueToSqlConverter.CanConvert(typeof(byte[])))
					{
						var trimmed =
							new byte[Configuration.MaxByteLengthLogging];
						Array.Copy(bytes, 0, trimmed, 0,
							Configuration.MaxByteLengthLogging);
						ValueToSqlConverter.TryConvert(sb, trimmed);
						sb.AppendFormat(
							$"-- Truncated for logging, actual length is {bytes.Length}");
					}
					else if (!ValueToSqlConverter.TryConvert(sb, p.Value))
						FormatParameterValue(sb, p.Value);
					sb.AppendLine();
				}
```

Source/LinqToDB/Common/StringBuilderExtensions.cs
to11mtm: > 
> 
> Regarding benchmarks: You should always use Benchmark.NET to run benchmarks. The library is written to correctly test micro-benchmarks and has tools to manage various things that can affect benchmark performance. Here's an updated copy of your code w/ Benchmark.NET: http://share.linqpad.net/x3j7fk.linq 

BTW thank you for showing me how to use Benchmark.NET in Linqpad... IDK why I never tried it that way.

to11mtm:
@MaceWindu I think the above would cover the major case? The only things I think it would not cover are:

 - `InlineParameters` I'm assuming would still have problems with this; however I would assume in that case they would still get an OOM executing the SQL anyway, so I think it's not concern
 - If someone inlines a hard coded byte array directly in a query expression, I'm assuming this will not catch it. But I feel like that case is a bit too absurd of a use case to worry about.

MaceWindu:
If user use binary literals, it's not much we can do about it.
Also could you add `Binary` type support in print parameters?

to11mtm:
@MaceWindu Can you see if my latest commit works for our needs?

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -40,6 +41,13 @@ public static class Configuration
 		/// queries, so this optimization could be used for <see cref="CommandBehavior.Default"/> too.
 		/// </summary>
 		public static bool OptimizeForSequentialAccess = false;
+		
+		/// <summary>
+		/// Determines the length after which logging of binary data in SQL will be truncated.
+		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
+		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
+		/// </summary>
+		public static int MaxByteLengthLogging { get; set; } = 1024;
```
MaceWindu, line 50: - add notes that -1 means no-trimming
- not sure it makes sense to have such big limit by default, something around 100 should be enough?
- I would prefer `MaxBinaryLengthLogging` name

Also I think we should handle another unlimited type here: string (with own setting limited to e.g. 200 chars)

Source/LinqToDB/Common/StringBuilderExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+	public static class StringBuilderExtensions
+	{
+		private static readonly uint[] _lookup32 = CreateLookup32();
+        
+		private static uint[] CreateLookup32()
+		{
+			var result = new uint[256];
+			for (int i = 0; i < 256; i++)
+			{
+				string s =i.ToString("X2");
```
MaceWindu, line 15: spaces formatting around `=`

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3081,7 +3083,33 @@ public virtual StringBuilder PrintParameters(StringBuilder sb, IEnumerable<IDbDa
...
+					    bytes.Length > Configuration.MaxByteLengthLogging &&
+					    ValueToSqlConverter.CanConvert(typeof(byte[])))
+					{
+						var trimmed =
+							new byte[Configuration.MaxByteLengthLogging];
+						Array.Copy(bytes, 0, trimmed, 0,
+							Configuration.MaxByteLengthLogging);
+						ValueToSqlConverter.TryConvert(sb, trimmed);
+						sb.Append(
+							$"-- Truncated for logging, actual length is {bytes.Length}");
```
MaceWindu, line 3096: as we print comment below variable, change comment to: `value above truncated for logging, actual length is {bytes.Length}` ?

MaceWindu (changes requested):

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
to11mtm: got it

Source/LinqToDB/Common/StringBuilderExtensions.cs
to11mtm: Done.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 
 using JetBrains.Annotations;
+using LinqToDB.Linq;
```
sdanyliv, line 4: Move `using LinqToDB.Linq;` inside namespace.

Source/LinqToDB/DataProvider/Access/AccessMappingSchema.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Text;
+using LinqToDB.Common;
```
sdanyliv, line 3: Move `using LinqToDB.Common;` inside namespace.

Source/LinqToDB/DataProvider/Firebird/FirebirdMappingSchema.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Text;
+using LinqToDB.Common;
```
sdanyliv, line 3: Move `using LinqToDB.Common;` inside namespace.

Source/LinqToDB/DataProvider/SQLite/SQLiteMappingSchema.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Globalization;
 using System.Text;
+using LinqToDB.Common;
```
sdanyliv, line 4: Move `using LinqToDB.Common;` inside namespace.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -40,6 +41,22 @@ public static class Configuration
...
+		/// </summary>
+		public static int MaxBinaryParameterLengthLogging { get; set; } = 100;
+
+		/// <summary>
+		/// Determines the length after which logging of string data in SQL will be truncated.
+		/// Use a value of -1 to disable. Set to 0 to truncate all binary data.
+		/// This is to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
+		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
+		/// </summary>
+		public static int MaxStringParameterLengthLogging { get; set; } = 200;
```
viceroypenguin, line 59: I believe that commentary about specific values belongs in the `<remarks />` tag instead of in the `<summary />` tag. @MaceWindu can correct me if he doesn't like this?
```suggestion
		/// <summary>
		/// Determines the length after which logging of string data in SQL will be truncated.
		/// Truncation is recomended to avoid Out-Of-Memory exceptions when getting SqlText from <see cref="TraceInfo"/>
		/// or <see cref="IExpressionQuery"/> for logging or other purposes.
		/// </summary>
		/// <remarks>
		///  This value defaults to 200. 
		/// Use a value of -1 to disable truncation.
		/// Use a value of 0 to truncate all data.
		/// </remarks>
		public static int MaxStringParameterLengthLogging { get; set; } = 200;
```

Source/LinqToDB/DataProvider/SapHana/SapHanaMappingSchema.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace LinqToDB.DataProvider.SapHana
+﻿using LinqToDB.Common;
```
sdanyliv, line 1: Move `using LinqToDB.Common;` inside namespace.

Source/LinqToDB/DataProvider/Sybase/SybaseMappingSchema.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Text;
+using LinqToDB.Common;
```
sdanyliv, line 3: Move `using LinqToDB.Common;` inside namespace.

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -2,11 +2,13 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.Linq;
 using System.Data.SqlTypes;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using LinqToDB.Linq;
```
sdanyliv, line 11: Move `using LinqToDB.Linq;` inside namespace.

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3081,7 +3083,47 @@ public virtual StringBuilder PrintParameters(StringBuilder sb, IEnumerable<IDbDa
 					sb.Append("SET     ");
 					PrintParameterName(sb, p);
 					sb.Append(" = ");
-					if (!ValueToSqlConverter.TryConvert(sb, p.Value))
+					if (p.Value is byte[] bytes                           &&
+					    Configuration.MaxBinaryParameterLengthLogging >=0 &&
```
sdanyliv, line 3087:
```suggestion
					    Configuration.MaxBinaryParameterLengthLogging >= 0 &&
```

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3081,7 +3083,47 @@ public virtual StringBuilder PrintParameters(StringBuilder sb, IEnumerable<IDbDa
...
+						var trimmed =
+							new byte[Configuration.MaxBinaryParameterLengthLogging];
+						Array.Copy(bytes, 0, trimmed, 0,
+							Configuration.MaxBinaryParameterLengthLogging);
+						ValueToSqlConverter.TryConvert(sb, trimmed);
+						sb.Append(
+							$"-- value above truncated for logging, actual length is {bytes.Length}");
+					}
+					else if (p.Value is Binary binaryData &&
+					         Configuration.MaxBinaryParameterLengthLogging >=0 &&
```
sdanyliv, line 3100:
```suggestion
					         Configuration.MaxBinaryParameterLengthLogging >= 0 &&
```

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3081,7 +3083,47 @@ public virtual StringBuilder PrintParameters(StringBuilder sb, IEnumerable<IDbDa
...
+						var trimmed =
+							new byte[Configuration.MaxBinaryParameterLengthLogging];
+						Array.Copy(binaryData.ToArray(), 0, trimmed, 0,
+							Configuration.MaxBinaryParameterLengthLogging);
+						ValueToSqlConverter.TryConvert(sb, trimmed);
+						sb.Append(
+							$"-- value above truncated for logging, actual length is {binaryData.Length}");
+					}
+					else if (p.Value is string s && 
+					         Configuration.MaxStringParameterLengthLogging >=0 &&
```
sdanyliv, line 3115:
```suggestion
					         Configuration.MaxStringParameterLengthLogging >= 0 &&
```

Source/LinqToDB/Common/Configuration.cs
to11mtm: > add notes that -1 means no-trimming

Done, along with notes that 0 effectively truncates all data which may be useful in some scenarios.

> not sure it makes sense to have such big limit by default, something around 100 should be enough?

I went with the limit of 100; we may want even less but can tweak in future I'm sure.

> I would prefer MaxBinaryLengthLogging name

I changed up the name a bit, might be a little -too- wordy but wanted clarity, because when I tried naming the other one `MaxStringLengthLogging` felt very confusing, like user might not get a `SqlText` longer than that value. So I settled with `MaxBinaryParameterLengthLogging` and `MaxStringParameterLengthLogging`. We may want to tweak further.

> Also I think we should handle another unlimited type here: string (with own setting limited to e.g. 200 chars)

Done.

Source/LinqToDB/Common/Configuration.cs
to11mtm: I'll defer to you and @MaceWindu on that one.

to11mtm:
I'm noticing that the .sql files generated by baselines are having bits truncated by this, do we need to make a change in CI somewhere to account for this change?:


https://github.com/linq2db/linq2db.baselines/pull/517/files#diff-65291d4185c3e9839c0e707c97a3debf442d7f9410756d44748bca31f91c74f1R14

MaceWindu:
I'm fine with baselines values truncation, but I noticed that comment placed after parameter value instead of new line

to11mtm:
> 
> 
> I'm fine with baselines values truncation, but I noticed that comment placed after parameter value instead of new line

I've corrected so that comment will appear on a new line in these cases.

to11mtm:
@MaceWindu anything else on this before I re-dev and merge? I don't want to miss 3.3.0 if possible

MaceWindu (approved):

sdanyliv (approved):

viceroypenguin (approved):

PR #2350, opened 2020-07-17
Update .editorconfig file

viceroypenguin:
* Update the .editorconfig file to enforce style
* Update some style nits project wide

PR #2778, opened 2021-01-18
chore: Set indent_size to 4 on GitHub

worldbeater:
This PR sets `indent_size` to 4 for the linq2db repository by updating the `.editorconfig` file. With recent changes made to the GitHub online code viewer, the `indent_size` for C# (the tab size) now defaults to 8, and this might hurt the readability of the code in certain cases. E.g. without `indent_size=4` the code looks as such:

With `indent_size=4` it would look like this by default:

Thanks for making and maintaining the best ORM for .NET btw.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,7 @@ root = true
 
 [*]
 indent_style = tab
+indent_size = 4
 insert_final_newline = true
 end_of_line = crlf
 


MaceWindu:
@linq2db, any objections?
I already use 4 for tab size, so I'm fine with this change, but if somebody has other preferences, it will enforce 4 for him.

viceroypenguin:
I seem to remember some discussion about this when I updated the .editorconfig. I don't know what the real conclusion was, but I remember taking it out because... /vague hand-waving. I'm certainly open for putting it in though!

MaceWindu:
it was about setting it to 2 :) https://github.com/linq2db/linq2db/pull/2350/files#r456585138

MaceWindu:
I would prefer github recognize github-specific option, but it looks like there is no such option

viceroypenguin:
Then as long as @igor-tkachev  is good with it, I'm good with it. :)

sdanyliv (approved):

igor-tkachev (approved):

viceroypenguin (approved):

Issue #2787, opened 2021-01-20
SQL CASE WHEN THEN


HypeillingerChillinger:
Cannot create a CASE WHEN THEN similar to https://stackoverflow.com/questions/936028/linq-case-statement

Here is my Linq2Db example

```C#
public class MyProjection
{
    public Car Car {get;set;}
    public Tyre Tyre {get;set;}
}

    public class Linq2DbPlayground
    {
        public void Start()
        {
            var context = new DataConnection(new SQLiteDataProvider(ProviderName.SQLiteClassic), "Data Source=:memory:");
            context.CreateTable<Car>();
            context.CreateTable<Tyre>();
            context.CreateTable<Group>();
            context.CreateTable<CarTyreMapping>();
            context.CreateTable<Brand>();

            var fluentMappingBuilder = context.MappingSchema.GetFluentMappingBuilder();

            var brandBuilder = fluentMappingBuilder.Entity<Brand>();
            brandBuilder.Property(x => x.Id).IsPrimaryKey();
            brandBuilder.Association(x => x.Groups, x => x.Id, x => x.BrandId);

            var groupBuilder = fluentMappingBuilder.Entity<Group>();
            groupBuilder.Property(x => x.Id).IsPrimaryKey();
            groupBuilder.Association(x => x.Cars, x => x.Id, x => x.GroupId);
            groupBuilder.Association(x => x.Brand, x => x.BrandId ,x => x.Id);

            var carBuilder = fluentMappingBuilder.Entity<Car>();
            carBuilder.Property(x => x.Id).IsPrimaryKey();
            carBuilder.Association(x => x.Group, x => x.GroupId, x => x.Id);

            var tyreHistoryMappingBuilder = fluentMappingBuilder.Entity<CarTyreMapping>();

            var tyreBuilder = fluentMappingBuilder.Entity<Tyre>();
            tyreBuilder.Property(x => x.Id).IsPrimaryKey();

            context.Insert(new Car { Id = 1, Name = "MyCar", IsSet = true });
            context.Insert(new Car { Id = 2, Name = "Bla" });
            context.Insert(new Group { Id = 1, Name = "group 1", IsSet = false });
            context.Insert(new Tyre { Id = 1, Name = "MyTyre", CarId = 1 });
            context.Insert(new CarTyreMapping { CarId = 1, TyreId = 1, Time = DateTime.UtcNow.AddDays(-2) });

            var carTable = context.GetTable<Car>();
[...]

PR #2788, opened 2021-01-20
Avoid selection of columns missing from final proojection

jogibear9988:
Test for #2787


sdanyliv:
Don't use new tables, enough is defined by default like db.Parent.

jogibear9988:
@sdanyliv simpilfied

MaceWindu:
moved to https://github.com/linq2db/linq2db/tree/issue/regression-tests

PR #2810, opened 2021-02-01
Fix for #2797. Added missed IsPredicate and IsWindowFunction for SqlExpression.

### LINKED ISSUE #2797: Incorrect Query generation for Ternary
SQL Server; l2db 3.2.3. Introduced after 3.1.6

Table and column are not important, only that the column is numeric (int, tinyint, etc.)
```csharp
context.Equipment.Select(e => Sql.AsSql(Sql.Between(e.EquipmentTypeId, 2, 5) ? 0 : 1)).ToList();
```
Renders to the following SQL:
```sql
SELECT
	IIF([e].[EquipmentTypeId] BETWEEN 2 AND 5 = 1, 0, 1)
FROM
	[dbo].[Equipment] [e]
```

There is a false comparison (`=1`) when the value is already a bool expression, so the syntax doesn't work right.

sdanyliv:
Fixes #2797

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
@@ -209,7 +209,7 @@ namespace LinqToDB.DataProvider.Oracle
 			return expr;
 		}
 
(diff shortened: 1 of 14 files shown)


MaceWindu (approved):
except UpdateTestJoinTake test failing for oracle for some reason...

viceroypenguin (approved):
LGTM! Thanks for the quick response on this! Failing tests aren't related to this PR, so I'm good with them as is.

PR #2820, opened 2021-02-07
Stub Type.GetInterfaceMap on CoreRT

### LINKED ISSUE #2680: SQLite.Ms wants to use System.Data.Sqlite in Release mode
I use linq2db and linq2db.SQLite.MS. Compiling to Debug mode does not produce any Error.
But on doing this in Release mode, the first time it tries to create a DataConnection it shuts down with the Error below.

```
Exception message: 
System.InvalidOperationException
  HResult=0x80131509
  Nachricht = Cannot load assembly System.Data.SQLite
  Quelle = <Die Ausnahmequelle kann nicht ausgewertet werden.>
  Stapelüberwachung:
<Die Ausnahmestapelüberwachung kann nicht ausgewertet werden.>

Stack trace:
   at LinqToDB.DataProvider.SQLite.SQLiteProviderAdapter.CreateAdapter(String assemblyName, String clientNamespace, String prefix) in D:\a\1\s\Source\LinqToDB\DataProvider\SQLite\SQLiteProviderAdapter.cs:line 52
   at LinqToDB.DataProvider.SQLite.SQLiteProviderAdapter.GetInstance(String name) in D:\a\1\s\Source\LinqToDB\DataProvider\SQLite\SQLiteProviderAdapter.cs:line 78
   at LinqToDB.DataProvider.SQLite.SQLiteDataProvider..ctor(String name, MappingSchema mappingSchema) in D:\a\1\s\Source\LinqToDB\DataProvider\SQLite\SQLiteDataProvider.cs:line 32
   at LinqToDB.DataProvider.SQLite.SQLiteDataProvider..ctor(String name) in D:\a\1\s\Source\LinqToDB\DataProvider\SQLite\SQLiteDataProvider.cs:line 27
   at LinqToDB.DataProvider.SQLite.SQLiteTools.<>c.<.cctor>b__21_0() in D:\a\1\s\Source\LinqToDB\DataProvider\SQLite\SQLiteTools.cs:line 17
   at System.Func`1.Invoke()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode) in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\shared\System\Lazy.cs:line 324
```

### Steps to reproduce
I use the code below to build my LinqToDbConnectionOptions. Nothing else has been used.

```c#
LinqToDB.Common.Configuration.Linq.AllowMultipleQuery = true;
            var connectionString = @"Data Source=internal.db";

            // create options builder
            var builder = new LinqToDbConnectionOptionsBuilder();

            // configure connection string
            builder.UseSQLite(connectionString);

            return builder.Build();
```

### Environment details
linq2db version: *3.1.6*
Database Server: *?*
Database Provider: *SQlite MS*
Operating system: *Windows*
.NET Framework: *.Net Standard 2.0*

MaceWindu:
Fix #2680 

Note that this issue applies only to CoreRT runtime, enabled by `Compile with .NET Native tool chain` build option.
Implementation for this method were added to corert by [this](https://github.com/dotnet/corert/pull/8144) PR but it looks like it was never released and whole CoreRT stuff [looks like dead horse](https://github.com/dotnet/corert/issues/7200).

@Daveismus, I'm replacing GetInterfaceMap implementation with noop method, which means you cannot use interfaces in your data model (linq2db will not see them). Another option to make it work is to disable AOT buid for your project...

### DIFF
--- a/Source/LinqToDB/Extensions/ReflectionExtensions.cs
+++ b/Source/LinqToDB/Extensions/ReflectionExtensions.cs
@@ -254,17 +254,23 @@ namespace LinqToDB.Extensions
 				_getInterfaceMap = (t, i) => t.GetInterfaceMap(i);
 				try
 				{
-					return _getInterfaceMap(interfaceType):
+					return _getInterfaceMap(type, interfaceType);
 				}
 				catch (PlatformNotSupportedException)
 				{
-					// porting of https://github.com/dotnet/corert/pull/8144 requires a lot of fragile platform-specific reflection
-					// so we'll just "unsupport" interfaces in mappings for UWP without native implementation
-					_getInterfaceMap = (t, i) => default;
+					// porting of https://github.com/dotnet/corert/pull/8144 is not possible as it requires access
+					// to non-public runtime data and reflection doesn't work in corert
+					_getInterfaceMap = (t, i) => new InterfaceMapping()
+					{
+						TargetType       = t,
+						InterfaceType    = i,
+						TargetMethods    = Array.Empty<MethodInfo>(),
+						InterfaceMethods = Array.Empty<MethodInfo>()
+					};
 				}
 			}
 
-			return _getInterfaceMap(interfaceType):
+			return _getInterfaceMap(type, interfaceType);
 #else
 			return type.GetInterfaceMap(interfaceType);
 #endif
--- a/Source/LinqToDB/Extensions/ReflectionExtensions.cs
+++ b/Source/LinqToDB/Extensions/ReflectionExtensions.cs
@@ -241,12 +241,41 @@ namespace LinqToDB.Extensions
 			return type.GetMember(name);
 		}
 
+#if NETSTANDARD2_0
+		private static Func<Type, Type, InterfaceMapping>? _getInterfaceMap;
+#endif
+		public static InterfaceMapping GetInterfaceMapEx(this Type type, Type interfaceType)
+		{
+			// native UWP builds (corert) had no GetInterfaceMap() implementation
+			// (added here https://github.com/dotnet/corert/pull/8144)
+#if NETSTANDARD2_0
+			if (_getInterfaceMap == null)
+			{
+				_getInterfaceMap = (t, i) => t.GetInterfaceMap(i);
+				try
+				{
+					return _getInterfaceMap(interfaceType):
+				}
+				catch (PlatformNotSupportedException)
+				{
+					// porting of https://github.com/dotnet/corert/pull/8144 requires a lot of fragile platform-specific reflection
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -288,7 +317,7 @@ static void GetAttributesTreeInternal(List<object> list, Type type)
 				{
 					var getAttr = false;
 
-					foreach (var mi in type.GetInterfaceMap(intf).TargetMethods)
+					foreach (var mi in type.GetInterfaceMapEx(intf).TargetMethods)
```
viceroypenguin, line 320: If we return `default` above, does that mean we're going to get an NRE on `.TargetMethods`?

Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -241,12 +241,41 @@ public static MemberInfo[] GetPublicMemberEx(this Type type, string name)
...
+			// (added here https://github.com/dotnet/corert/pull/8144)
+#if NETSTANDARD2_0
+			if (_getInterfaceMap == null)
+			{
+				_getInterfaceMap = (t, i) => t.GetInterfaceMap(i);
+				try
+				{
+					return _getInterfaceMap(interfaceType):
+				}
+				catch (PlatformNotSupportedException)
```
viceroypenguin, line 259: If we catch `PlatformNotSupportedException` even once, doesn't that mean we'll get it for all functions? So maybe would be better to use a simple `bool? flag` of some kind to track unsupported (`null` means untested, `true`/`false` mean known, etc.

MaceWindu: we will enter this `if` only first  time when  `_getInterfaceMap` is not yet initialized

Source/LinqToDB/Extensions/ReflectionExtensions.cs
MaceWindu: yep, will fix it tomorrow

Source/LinqToDB/Extensions/ReflectionExtensions.cs
viceroypenguin: Ah. Yup. Misread that as a dictionary to be initialized for each (type, type) rather than a func. I'm good w/ this.

MaceWindu:
Tested stub with sqlite - as expected only 4 tests that use interface mappings failed. Also managed to run some queries in UWP using MS sqlite provider.

viceroypenguin (approved):

PR #2824, opened 2021-02-08
SQL translation for string.IsNullOrWhitespace

### LINKED ISSUE #2816: string.IsNullOrWhitespace cannot be converted to SQL
In Linq2Db 3.2.3 
string.IsNullOrWhitespace is not supported

```C#
var carTable = context.GetTable<Car>();
carTable.Where(x => !string.IsNullOrWhitespace(x.Name));
```

jogibear9988:
Fix #2816

Databases with limited support:
- Access: only whitepace characters recognized by Access `*trim` functions will be considered as whitespace (0x20 (SPACE), 0x1680 (OGHAM SPACE MARK), 0x205F (MEDIUM MATHEMATICAL SPACE), 0x3000 (IDEOGRAPHIC SPACE)). Proper implementation requires 24 nested `Replace` calls and doesn't work anyway without [sandbox mode](https://support.microsoft.com/en-us/office/turn-sandbox-mode-on-or-off-to-disable-macros-8cc7bad8-38c2-4a7a-a604-43e9a7bbc4fb)
- Informix: while tests pass, I'm not very confident it will work for all encoding for non-ascii whitespaces (probably works now due to some unicode normalization done by informix)

### DIFF
--- a/Tests/Linq/UserTests/Issue2816Tests.cs
+++ b/Tests/Linq/UserTests/Issue2816Tests.cs
@@ -159,6 +159,15 @@ namespace Tests.UserTests
 					// full support
 					return true;
 
+				case ProviderName.Access:
+				case ProviderName.AccessOdbc:
+					// only 4 characters including space supported
+					return character == 0x20
+						|| character == 0x1680
+						|| character == 0x205F
+						|| character == 0x3000;
+
+
 					//	case ProviderName.Informix:
 					//	case ProviderName.InformixDB2:
 					//		// teh winner!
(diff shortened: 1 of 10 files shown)


Tests/Linq/UserTests/Issue2816Tests.cs
```diff
@@ -0,0 +1,47 @@
...
+					var query = from p in table
+								where string.IsNullOrWhiteSpace(p.Text)
+								select p;
+
+					var res = query.ToList();
+
+					var sql = ((DataConnection) db).LastQuery;
+
+					Assert.AreEqual(res.Count, 3);
+				}
```
sdanyliv, line 43: Use `AssertQuery` here

jogibear9988: ?? AsserQuery? I don't use the sql, the variable was only there to llok at the sql

sdanyliv: Yes `AssertQuery(query)` instead of these three lines of code.

Tests/Linq/UserTests/Issue2816Tests.cs
```diff
@@ -0,0 +1,43 @@
...
+			using (var db = GetDataContext(context))
+			{
+				using (var table = db.CreateLocalTable<TestClass>())
+				{
+					db.Insert(new TestClass() { Id = 0, Text = "a" });
+					db.Insert(new TestClass() { Id = 1, Text = null });
+					db.Insert(new TestClass() { Id = 2, Text = " " });
+					db.Insert(new TestClass() { Id = 3, Text = "  " });
+					db.Insert(new TestClass() { Id = 4, Text = " m" });
+
```
sdanyliv, line 33: For speedup test, just put array into `CreateLocalTable` class.

MaceWindu:
Surprisingly, best compatibility with .net implementation shown by Informix

viceroypenguin (approved):

MaceWindu:
I will extract `NOT` into AST so it could be optimized for cases like !string.IsNullOrWhitespace() to not have two negations

viceroypenguin (approved):

MaceWindu:
that's funny (not really). Something changed in azure ubuntu image between 3 and 20 february and now sybase image doesn't work (but work with macos, so we will use it for now for testing). Will look into it later.

PR #2826, opened 2021-02-08
Fix issue 2823. Guid data type suppport for Firebird data provider.

### LINKED ISSUE #2823: FirebirdSqlBuilder.BuildDataTypeFromDataType doesnt  handle Guid data type
By default DataType.Guid is converted to "Guid" string. But there is not Guid data type in Firebird, usually they use char(16) charset octets
Test:
```cs
		[Table]
		class TableWithGuid
		{
			[Column] public Guid Guid   { get; set; }			
			[Column] public string Data { get; set; }
		}

		[Test]
		public void CreateTableWithGuid([DataSources(false)] string context)
		{			
			using (var table = db.CreateLocalTable<TableWithGuid>())
			{
				var guid = Guid.NewGuid();
				table.Insert(() => new TableWithGuid { Guid = guid, Data = "My data" });
				Assert.AreEqual("My data", table.Where(x => x.Guid == guid).Select(x => x.Data).First());
			}
		}
```
Failed with exception
```
    FirebirdSql.Data.FirebirdClient.FbException : unsuccessful metadata update
    CREATE TABLE TABLEWITHGUID failed
    SQL error code = -607
    Invalid command
    **Specified domain or source column GUID does not exist**
```

Fix:
add to FirebirdSqlBuilder.BuildDataTypeFromDataType(SqlDataType type, bool forCreateTable)
```cs
case DataType.Guid: StringBuilder.Append("CHAR(16) CHARACTER SET OCTETS"); break;
```

### LINKED ISSUE #2833: [Firebird] Guid literal generation (ValueToSqlConverter)
Current behavior:
```cs
db.InlineParameters = true;
var guid = Guid.NewGuid(); 
db.MyTable.Insert(() => new MyTable { GuidValue = guid });
``` 
will generate
```sql
INSERT INTO MyTable
(
	GuidValue
)
VALUES
(
	'b999f87c-fcc0-4858-b155-264205d8c36a'
)
```
It would be nice to generate native format, something like 

```sql
INSERT INTO MyTable
(
	GuidValue
)
VALUES
(
	X'48DE4C2A3FFF422DABB48357FF84E27F'
)
```
but it's breaking change!
Is it OK to add flag to FirebirdConfiguration ??
```
/// <summary>
/// Specifies how linq2db stored Guids. If true then GUIDs enforced stored as strings
/// </summary>
public static bool IsGuidEncodedAsString { get; set; } = true;
```
IsGuidEncodedAsString = true means old behavior, 

| IsGuidEncodedAsString 	| DataType 	| Result SQL value                       	|
|:---------------------:	|----------	|----------------------------------------	|
| true                  	| any      	| 'b999f87c-fcc0-4858-b155-264205d8c36a' 	|
| false                 	| Guid     	| X'48DE4C2A3FFF422DABB48357FF84E27F'    	|
| false                 	| != Guid  	| 'b999f87c-fcc0-4858-b155-264205d8c36a' 	|

jack128:
Fix #2823
Fix #2833


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2826) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2826) before we can accept your contribution.<br/><br/>:x: jack128 [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=2826)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=2826) it.</sub>

Source/LinqToDB/DataProvider/Firebird/FirebirdSqlBuilder.cs
```diff
@@ -122,6 +122,8 @@ protected override void BuildDataTypeFromDataType(SqlDataType type, bool forCrea
 
 					StringBuilder.Append(" CHARACTER SET UNICODE_FSS");
 					                                                                                      break;
+
+				case DataType.Guid: StringBuilder.Append("CHAR(16) CHARACTER SET OCTETS");                break;
```
MaceWindu, line 126: note to other reviewers: this is correct type for GUID in firebird, e.g. see https://firebirdsql.org/refdocs/langrefupd25-intfunc-gen_uuid.html

Tests/Linq/Linq/IssueTests.cs
```diff
@@ -691,6 +691,25 @@ void Query(bool isNull)
 			return isNull ? (short?)null : 1234;
 		}
 
+		[Test]
+		public void Issue2823Guid([DataSources(false, ProviderName.AccessOdbc)] string context)
```
MaceWindu, line 695: please fix data sources attribute to [IncludeDataSources(true, TestProvName.AllFirebird)]

MaceWindu (changes requested):
Thanks! Just fix providers used for test

Tests/Linq/Linq/IssueTests.cs
```diff
@@ -691,6 +691,25 @@ void Query(bool isNull)
 			return isNull ? (short?)null : 1234;
 		}
 
+		[Test]
+		public void Issue2823Guid([IncludeDataSources(true, TestProvName.AllFirebird)] string context)
```
MaceWindu, line 695:
```suggestion
		public void Issue2823Guid([IncludeDataSources(true, TestProvName.AllFirebird)] string context, [Values] bool useParameters)
```

Tests/Linq/Linq/IssueTests.cs
```diff
@@ -691,6 +691,25 @@ void Query(bool isNull)
...
+		{
+			using(var db = GetDataContext(context))
+			using(var table = db.CreateLocalTable<TableWithGuid>())
+			{
+				var guid = Guid.NewGuid();
```
MaceWindu, line 700:
```suggestion
				db.InlineParameters = !useParameters;
				var guid = Guid.NewGuid();
```

MaceWindu (approved):
Not directly related, but would be nice to add literal generation for guid. I've added suggested changes to test, so if you want to implement it - check `SetValueToSqlConverter` calls in `FirebirdMappingSchema`

Tests/Linq/Linq/IssueTests.cs
jack128: Well, there is a mess with GUIDs in FB NET Provider due that issue http://tracker.firebirdsql.org/browse/DNET-509, so that will be not so easy. I will try in days.

MaceWindu: ... I see it's quite popular issue when there is no database type for guid and binary used. microsoft sqlite provider also was hit by it :smile:

jack128: No, I will create new issue for discussion and separated PR, it's too complex theme.

Source/LinqToDB/DataProvider/Firebird/FirebirdMappingSchema.cs
```diff
@@ -27,6 +28,13 @@ protected FirebirdMappingSchema(string configuration) : base(configuration)
 			SetValueToSqlConverter(typeof(byte[])  , (sb, dt, v) => ConvertBinaryToSql(sb, (byte[])v));
 			SetValueToSqlConverter(typeof(Binary)  , (sb, dt, v) => ConvertBinaryToSql(sb, ((Binary)v).ToArray()));
 			SetValueToSqlConverter(typeof(DateTime), (sb, dt, v) => BuildDateTime(sb, dt, (DateTime)v));
+			SetValueToSqlConverter(typeof(Guid)    , (sb, dt, v) => ConvertGuidToSql(sb, dt, (Guid)v));
+		}
+
+		static FirebirdMappingSchema()
+		{
+			var fbConnectionType = FirebirdProviderAdapter.GetInstance().ConnectionType;
+			_useLegacyGuidEncoding = (fbConnectionType.Assembly is Assembly assembly) && (assembly.GetName().Version < new Version(6, 0, 0, 0));
```
MaceWindu, line 37: This is not a correct way to initialize schema with version-specific behavior.
1. version load should be done in provider adapter (`FirebirdProviderAdapter`)
2. provider instance should configure schema on creation based on version. Check how some other providers do it. E.g. `InformixDataProvider` - they create mapping schema as concatenation of static schema instance and schema with version-specific configuration created by adapter

MaceWindu (changes requested):

Source/LinqToDB/DataProvider/Firebird/FirebirdMappingSchema.cs
jack128: Fixed

MaceWindu:
Plan to review it later this week and include into 4.0 preview 1

MaceWindu:
@jack128 
I've decided to remove workaround for client < 6 as:
- it was applied to literal generation, but from issue description I see that problem is not here but in incorrect read logic. This means we will write incorrect data into database (not talking about situation when user update client)
- even v6 client is quite old and I don't think users should expect seamless work of pre-v6 client

If we really want to fix it on our side, we should register custom reader expression, which will reverse bits after we get data from provider. But don't think it worth doing it now without request from user.

MaceWindu (approved):

Source/LinqToDB/DataProvider/Firebird/FirebirdSqlBuilder.cs
viceroypenguin: Note is accurate. Can we add comment here in code for future reviewers who don't look at this PR specifically?

viceroypenguin (approved):

sdanyliv (approved):

PR #2829, opened 2021-02-09
 Fix for #2800. Added additional check for null comparison.

### LINKED ISSUE #845: Possibly wrong SQL with ExpressionPredicate
First I must say that the final result is working as expected. But when I tried to see generated SQL through _SQL Server Profiler_ I think that it have unnecessary _NOT Boolean_ expression.

This is my ExpressionPredicate
`public static Expression<Func<Employee, Department, bool>> DepartmentPredicate => ( e, d ) => e.DepartmentId == d.DepartmentId && !d.Deleted;`

And this is generated SQL
```
SELECT
*
FROM
    [Common].[Employee] [t4]
    LEFT JOIN [Common].[Department] [t1] ON [t4].[DepartmentId] = [t1].[DepartmentId] AND NOT (1 = [t1].[Deleted])
```
Notice `NOT (1 = [t1].[Deleted])`.
I think this should be `0 = [t1].[Deleted]`.

Any thoughts on this?

### Environment details
linq2db version: *1.9.0*  
Database Server: *SQL Server*  
Framework version: *.NET Framework 4.6.1*

### LINKED ISSUE #2800: ExpressionMethod Exception on SQLite - No change on Oracle
The given ``` ExpressionMethod``` produces an exception on SQLite with null
```  carTable.Where(x => x.FilterBySpecialString(null)).ToList();```
but it works for false and true

On Oracle there is no exception but it produces the same select for null, false and true as if it was not used at all.

```C#
    public class Car
    {
        public long Id {get;set;}
        public string Name {get;set;}
    } 

    public static class CarExtensions
    {
        [ExpressionMethod(nameof(FilterBySpecialStringExpression))]
        public static bool FilterBySpecialString(this Car car, bool? isSpecial = false)
        {
            return !isSpecial.HasValue || isSpecial.Value && car.Name == "Special" || !isSpecial.Value && car.Name != "Special";
        }

        public static Expression<Func<Car, bool?, bool>> FilterBySpecialStringExpression()
        {
            return (x, isSpecial) => !isSpecial.HasValue || isSpecial.Value && x.Name == "Special" || !isSpecial.Value && x.Name != "Special";
        }
    }


    public class Linq2DbPlayground
    {
        public void Start()
        {
            var context = new DataConnection(new SQLiteDataProvider(ProviderName.SQLiteClassic), "Data Source=:memory:");
            context.CreateTable<Car>();

            var fluentMappingBuilder = context.MappingSchema.GetFluentMappingBuilder();

            var carBuilder = fluentMappingBuilder.Entity<Car>();
            carBuilder.Property(x => x.Id).IsPrimaryKey();
            carBuilder.Association(x => x.Group, x => x.GroupId, x => x.Id);

            context.Insert(new Car { Id = 1, Name = "MyCar", IsSet = true });

            var carTable = context.GetTable<Car>();

            var result = carTable.Where(x => x.FilterBySpecialString(null)).ToList();

            var lastQuery = context.LastQuery;
        }
    }
```

sdanyliv:
Fixes #845
Fixes #2800

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
@@ -1998,7 +1998,9 @@ namespace LinqToDB.Linq.Builder
 					expression = l;
 				}
 
-				if (value != null && expression != null)
+				if (value != null 
+				    && expression != null 
+				    && !(expression.ElementType == QueryElementType.SqlValue && ((SqlValue) expression).Value == null))
 				{
 					var isNot = !value.Value;
 					var withNull = false;
@@ -3007,9 +3009,12 @@ namespace LinqToDB.Linq.Builder
 
 		static bool NeedNullCheck(ISqlExpression expr)
 		{
+			if (!expr.CanBeNull)
+				return false;
+
 			if (null != new QueryVisitor().Find(expr, e => e.ElementType == QueryElementType.SelectClause))
 				return false;
-			return expr.CanBeNull;
+			return true;
 		}
 
 		#endregion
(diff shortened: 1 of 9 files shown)


MaceWindu (approved):

MaceWindu:
Looks like #845 also fixed by this PR

Issue #2854, opened 2021-02-27
Index was outside the bounds of the array


tiperes:
I am a Tech Lead in large software factory and our team has been using linq2db in the last 2 years.
We use System.Linq.Dynamic.Core library to build dynamically LINQ queries at runtime and linq2db as the LINQ provider for querying an Oracle database.

In the last core components revision, we have tested the upgrade of linqtodb from version 3.2.1 to version 3.2.3.
All applications depending on this library stopped working.
We have reverted back to linqtodb version 3.2.1.

This is not critical for us, at least for now, but we would like to continue to receive updates from this library.
Feel free to request any additional detail that you need from me.

Thanks!

**Exception message:**
   Index was outside the bounds of the array.

**Stack trace:**
```
   at LinqToDB.DataProvider.Oracle.Oracle11SqlBuilder.IsValidIdentifier(String name)
   at LinqToDB.DataProvider.Oracle.Oracle11SqlBuilder.Convert(StringBuilder sb, String value, ConvertType convertType)
   at LinqToDB.SqlProvider.BasicSqlBuilder.ConvertInline(String value, ConvertType convertType)
   at LinqToDB.SqlProvider.BasicSqlBuilder.GetPhysicalTableName(ISqlTableSource table, String alias, Boolean ignoreTableExpression, String defaultDatabaseName)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildPhysicalTable(ISqlTableSource table, String alias, String defaultDatabaseName)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildTableName(SqlTableSource ts, Boolean buildName, Boolean buildAlias)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildFromClause(SqlStatement statement, SelectQuery selectQuery)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildSelectQuery(SqlSelectStatement selectStatement)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildSql(Int32 commandNumber, SqlStatement statement, StringBuilder sb, OptimizationContext optimizationContext, Int32 indent, Boolean skipAlias)
   at LinqToDB.SqlProvider.BasicSqlBuilder.BuildSql(Int32 commandNumber, SqlStatement statement, StringBuilder sb, OptimizationContext optimizationContext, Int32 startIndent)
[...]

PR #2855, opened 2021-02-28
test for issue #2854

jogibear9988:


sdanyliv:
@jogibear9988, how about creating test using DynamicLINQ?

jogibear9988:
I don't think the error has anything to do with dynamic linq.
Dynamic Linq only creates Expressions so why should I need it?
Did my tests not fail? Cause they do on my system

sdanyliv:
It is just good for regression. I assume that validation function has a bug, but why empty string come to this function?

jogibear9988:
but why empty string come to this function? 
??? empty string to wich function? I don't understand....

sdanyliv:
Original issue has call stack: `LinqToDB.DataProvider.Oracle.Oracle11SqlBuilder.IsValidIdentifier`

jogibear9988:
Maybe original issue is an other problem.
I only thougt, after linq2db Update I've also an IndexOutOfRange exception. Maybe could be the same root cause.

sdanyliv:
So, this Is not test for the original issue. And why I'm asking to add DynamicLINQ variant.

jogibear9988:
But I still think dynamic linq is not relevant.
He need to tell what query he'd like to do, so we can create a test for it.

Expressions from c# queries and the ones created from dynamic linq look the same.

jogibear9988:
We use dynamic linq all over our system, and those querys all work also we newest linq2db.
Only this one I posted here fails

jogibear9988:
@sdanyliv should I create an extra issue for my problem?

sdanyliv:
If your issue has another call stack yes. But better create additional DynamicLINQ test here.

jogibear9988:
I've asked the author for his query. Yes mine has an other callstack. I'll create a new issue and move the test

PR #2860, opened 2021-03-01
Virtualize DataContext.Dispose

### LINKED ISSUE #2853: DataContext's Dispose method should be accessible in child classes
I have a class that inherits `DataContext` and I need some additional code on dispose.

The problem is that `DataContext`'s Dispose method is private and `IDisposable` is implemented explicitly, so I can't override it, nor can I implement `IDisposable` and call it.

A simple solution will be to make it `virtual` so inherited classes can override it if needed.

### Environment details
linq2db version: 3.1.3

MaceWindu:
Fix #2853

### DIFF
--- a/Source/LinqToDB/DataContext.cs
+++ b/Source/LinqToDB/DataContext.cs
@@ -386,7 +386,10 @@ namespace LinqToDB
 		/// <inheritdoc />
 		public Action<EntityCreatedEventArgs>? OnEntityCreated { get; set; }
 
-		void IDisposable.Dispose()
+		/// <summary>
+		/// Closes underlying connection and fires <see cref="OnClosing"/> event (only if connection existed).
+		/// </summary>
+		public virtual void Dispose()
 		{
 			_disposed = true;
 			Close();
--- a/Source/LinqToDB/DataContext.cs
+++ b/Source/LinqToDB/DataContext.cs
@@ -386,10 +386,16 @@ namespace LinqToDB
 		/// <inheritdoc />
 		public Action<EntityCreatedEventArgs>? OnEntityCreated { get; set; }
 
+		void IDisposable.Dispose()
+		{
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/DataContext.cs
```diff
@@ -386,7 +386,10 @@ IDataContext IDataContext.Clone(bool forNestedQuery)
 		/// <inheritdoc />
 		public Action<EntityCreatedEventArgs>? OnEntityCreated { get; set; }
 
-		void IDisposable.Dispose()
+		/// <summary>
+		/// Closes underlying connection and fires <see cref="OnClosing"/> event (only if connection existed).
+		/// </summary>
+		public virtual void Dispose()
```
viceroypenguin, line 392: Per [this document](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose). Also, update the `void Dispose(bool)` to only process if `disposing` is `true` 

```suggestion
		void IDisposable.Dispose()
		{
			Dispose(disposing: true);
			GC.SuppressFinalize(this);
		}
		
		/// <summary>
		/// Closes underlying connection and fires <see cref="OnClosing"/> event (only if connection existed).
		/// </summary>
		protected virtual void Dispose(bool disposing)
```

viceroypenguin (changes requested):

Source/LinqToDB/DataContext.cs
MaceWindu: Why? We don't have finalizer in DataContext.
I was always puzzled by this overcomplicated implementation...

MaceWindu: ah, whatever... let's imagine finalizer in derived class :smile:

viceroypenguin (approved):

sdanyliv (approved):

PR #2871, opened 2021-03-04
Added ANY functions for arrays.

### LINKED ISSUE #1660: postgres Array queries do not seem to be supported!
[Table(Name = "dd")]
public partial class dd
{
        [Column(Name = "dd", DataType = DataType.VarChar, Length = 255)]
        public System.String[] source { get; set; }
}

using (var con = DB.GetConnection())
 {
      var sql = con.Entity.dd.Where(c => c.source.Contains("2"));
      var data2 = sql.ToList();
}

throw:
LinqToDB.Linq.LinqException:“'c.source.Contains("2")' cannot be converted to SQL.”

sdanyliv:
Partially fixe_s #1660

### DIFF
--- a/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLExtensions.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLExtensions.cs
@@ -310,6 +310,42 @@ namespace LinqToDB.DataProvider.PostgreSQL
 			throw new LinqException($"'{nameof(ArrayToString)}' is server-side method.");
 		}
 
+		[Sql.Extension("{value} = ANY({array})", ServerSideOnly = true, CanBeNull = true, Precedence = Precedence.Primary, IsPredicate = true)]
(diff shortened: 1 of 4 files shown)


MaceWindu (approved):

Evengard:
Does this support standart linq methods such as Contains and/or Any? Or it is a special method only?

MaceWindu:
for now special methods only.
Not sure we should map generic `Contains`/`Any` methods by default. @sdanyliv ?

sdanyliv:
We can, but it needs additional effort.

Evengard:
Well, people first impulse would to attempt to use the generic Any/Contains methods first, only to end up with an exception. While, for example, window functions do need some explicit syntax, the array type is basically an enumeration, so seems logical to work with it like with any other enumerations... And the "ANY()" method from pgsql seems to be logical to be mapped to Linqs Any, even the names match.

That's just my two cents though. I have no ideas how hard is it, as it is a really complex library.

PR #2878, opened 2021-03-06
[PostgreSQL] ANY methods rename

MaceWindu:
Proposed rename of methods, introduced by #2871


sdanyliv (approved):

MaceWindu:
@viceroypenguin  what do you think? or should it use `Equals` instead of `Equal`?

viceroypenguin:
I'm good w/ "equal to" over "equals", but I wonder if we should add the word `If` in there, if that would read better?`Sql.Ext.PostgreSQL().ValueIsEqualToAny(value, array)`, `Sql.Ext.PostgreSQL().ValueIsLessThanAny(value, array)`, etc.  That would make the "equal to" much more natural as well.

viceroypenguin (approved):

PR #2879, opened 2021-03-07
Partial fix for transaction isolation level for multi-statement eager load sql server queries

### LINKED ISSUE #2790: The transaction isolation level is not reset when the connection is closed
Hello.

I have been migrating a large project from Entity Framework to linq2db and decided to split migration of one especially large database into 2 parts - after the first deployment one part of it will work with EF and the other with linq2db. After this deployment, I noticed that rare deadlocks are starting to occur in EF queries (that code was not changed). After digging around, I realized that some of EF queries use _RepeatableRead_ isolation level, but we use our default (_ReadCommited_ _Snapshot_) in all queries to this database. The single place where connection with _RepeatableRead_ may occur is that when you use _DefaultMultiQueryIsolationLevel_ setting (after wrapping all non transactional EF queries in _ReadCommited_ transaction deadlocks were stopped).

The problem is: **transaction isolation level is not reset to default when the connection is closed and after reuse it continues to use the last isolation level unless a new one is specified**. I read that until Sql Server 2014 _cp_reset_connection_ does not reset isolation level, but we use Sql Server 2017.

So I have a few questions:
1) When is _DefaultMultiQueryIsolationLevel_ setting used? (I assume that when called _LoadWith_ to load some collection)
2) I almost sure that is not linq2db bug, but anything is possible, so what do you know about that problem?

Thank you in advance.

### Environment details
linq2db version: 3.2.3
Database Server: Sql Server 2017
.NET Framework: net472

MaceWindu:
Fix #2790

### Problem:

When user execute eager load query which requires multiple queries to load data and there is no transaction available, linq2db will automatically wrap all queries into transaction with predefined isolation level (see below).
Due to `SqlClient` behavior where it doesn't reset current transaction isolation level even after connection was returned to connection pool, it changes transaction isolation level for pooled connection.
It is not a problem, when user always use explicit transactions and specify transaction isolation level, but when auto-commit transactions used they start to use isolation level introduced by eager load transaction.

### Solutions:
1. Reset transaction isolation level after eager load query to old value. Will require two additional roundtrips to server to query and reset isolation level.
2. Manually start transaction using `BEGIN TRAN` sql query to avoid isolation level change by `BeginTransaction` API.

This PR implements 2nd option as most lightweigth and more expected when auto-commit transactions used by application. Unfortunatelly it it not possible to tell `SqlClient` to not touch transaction level for `BeginTransaction` API (`IsolationLevel.Unspecified` will be replaced with `IsolationLevel.ReadCommitted`), so we need to manage transaction manually using `BEGIN TRAN` queries.

Due to MARS limitations, this fix will not work for MARS-enabled connections https://docs.microsoft.com/en-us/archive/blogs/cbiyikoglu/mars-transactions-and-sql-error-3997-3988-or-3983. It is still possible to fix it for MARS, but it will require eager load refactoring to batch all queries into single batch.

#### Isolation levels used by providers for eager load transactions:
- Oracle: ReadCommitted
- SQLite: Serializable
- Access: Unspecified
- Others: RepeatableRead
- Sql Server: was: RepeatableRead -> new : Unspecified/ReadCommitted (MARS)

### DIFF
--- a/Source/LinqToDB/DataProvider/DataProviderBase.cs
+++ b/Source/LinqToDB/DataProvider/DataProviderBase.cs
@@ -433,5 +433,8 @@ namespace LinqToDB.DataProvider
 #endif
 
 		#endregion
+
+		public virtual RawTransaction CreateRawTransaction(DataConnection dataConnection)
+			=> throw new InvalidOperationException($"{nameof(CreateRawTransaction)} API not suppored by {Name} provider");
 	}
 }
--- a/Source/LinqToDB/DataProvider/IDataProvider.cs
+++ b/Source/LinqToDB/DataProvider/IDataProvider.cs
@@ -60,5 +60,7 @@ namespace LinqToDB.DataProvider
 #if !NETFRAMEWORK
 		Task<BulkCopyRowsCopied> BulkCopyAsync<T>(ITable<T> table, BulkCopyOptions options, IAsyncEnumerable<T> source, CancellationToken cancellationToken);
 #endif
+
+		RawTransaction     CreateRawTransaction(DataConnection dataConnection);
 	}
 }
--- a/Source/LinqToDB/DataProvider/RawTransaction.cs
+++ b/Source/LinqToDB/DataProvider/RawTransaction.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using LinqToDB.Data;
+
+namespace LinqToDB.DataProvider
+{
+	/// <summary>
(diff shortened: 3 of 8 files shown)


MaceWindu:
I didn't checked if similar issue exists for other databases yet. I think first we should decide if we are going to accept this PR

MaceWindu:
I don't like it... Also several tests now fail as explicit transaction not detected by current logic

MaceWindu:
Started checking other databases : some of them affected, some not affected.

I think we should take another way here: leave current implementation intact and introduce a way to customize isolation level for eager load transactions

PR #2875, opened 2021-03-05
Corrected UPDATE FROM generation for Sybase.

sdanyliv:

Issue #798, opened 2017-08-12
Drop operation incorrectly handles throwExceptionIfNotExists parameter false value


MaceWindu:
Discovered while writing documentation. Right now this method will catch any exceptions during drop table operation when `throwExceptionIfNotExists` parameter value set to `false`.
This is not correct, because it should mask only missing table exceptions.

### Environment details
linq2db version: *1.8.3*

PR #2883, opened 2021-03-09
Prevent errors within TempTable.Dispose

### LINKED ISSUE #2882: TempTable.Dispose can throw an exception
### Problem

`TempTable.Dispose` can mask exceptions by the nature of the fact that a failed SQL statement can rollback a transaction, and the `Dispose` method will try to drop a table that does not exist anymore.

Consider the following sample:

```cs
await using var dc = GetDataConnection();
await using var transaction = await dc.BeginTransactionAsync(cancellationToken);
await using var table = await dc.CreateTempTableAsync("#Temp_Products", productsToUpdate, cancellationToken: cancellationToken);
await dc.GetTable<Product>()
    .Merge()
    .Using(table)
    .On(db => db.Id, table => table.Id)
    .UpdateWhenMatched((db, table) => new Product { StockQuantity = table.StockQuantity })
    .MergeAsync(cancellationToken);
await transaction.CommitAsync();
```

Then assume that there is a duplicate `Id` entry in `productsToUpdate`, causing MS SQL Server to give the following error:

```
The MERGE statement attempted to UPDATE or DELETE the same row more than once. This happens when
a target row matches more than one source row. A MERGE statement cannot UPDATE/DELETE the same row
of the target table multiple times. Refine the ON clause to ensure a target row matches at most one source row,
or use the GROUP BY clause to group the source rows.
```

The above error is thrown as an exception by linq2db.  MS SQL Server 2017 will also rollback the SQL transaction automatically, regardless of whether SET XACT_ABORT is set to ON or OFF.

The .NET runtime then attempts to dispose the table while passing the exception up the stack.  However, the table does not exist anymore.  So SQL reports this error, which is thrown by linq2db as an exception.

```
Cannot drop the table '#TempTarget', because it does not exist or you do not have permission.
```

This masks the actual cause of the exception, which is the merge error.

It is also noteworthy that the transaction, which is also now invalid, is also disposed of by .NET.  SQL should be returning this error:

```
The ROLLBACK TRANSACTION request has no corresponding BEGIN TRANSACTION.
```

But I assume that `DbTransaction.RollbackAsync()` is ignoring the error.

### Suggestion

I suggest that we change the `DropTable.Dispose()` method as follows: (and the DisposeAsync method similarly)

```cs
public virtual void Dispose()
{
	//_table.DropTable();
	_table.DropTable(throwExceptionIfNotExists: false);
}
```

It seems to me that as a general rule of thumb, `Dispose()` should not be throwing an error.

### Workarounds

If the temporary table is created with the `TableOptions.CheckExistence` flag then this error does not occur.

### Other notes

I have not tested this with databases other than SQL Server.  It is likely that not all of them behave the same way that SQL does, automatically reverting an active transaction upon a failed MERGE statement.  However, I am not sure that it matters much, considering the problem and suggested fix.

### Environment details
linq2db version: current `master` branch
Database Server: SQL Server 2017
Operating system: Win10
.NET Framework: .NET 5

Shane32:
Fix for issue #2882 including test.

I also wrote a test that demonstrates how a merge statement can cause SQL Server to rollback a transaction, causing a temporary table to be deleted before `TempTable.Dispose()` is called.  However, I did not include it because it seemed unnecessarily complex, and the test would only work for database engines that have the same semantics.

Fix #2882


Shane32:
@MaceWindu I assume we should be waiting for sybase to run successfully, due to the number of baseline changes?

Any idea if that's a temporary problem?

MaceWindu:
currently sybase works only for macos. to run macos tests you need to use test-sybase pipeline and merge your branch with recent master changes.

MaceWindu:
anyway you need to wait for https://github.com/linq2db/linq2db/pull/2875 first for sybase as it fix couple of failing tests for it (and will update outdated baselines for sybase)

Shane32:
@sdanyliv @MaceWindu Should I merge or wait for #2875 ?  (Either way I should pull changes from master and rerun tests due to #2884 / #2885 affecting baselines.)

MaceWindu:
I think proper fix will be to not try to drop table twice
FYI somewhat related issue: #798

Shane32:
> I think proper fix will be to not try to drop table twice

Is that possible?

Given the example scenario listed in #2882, we would need to:
1. determine if the table was created while inside a transaction
2. determine if the transaction was rolled back automatically by SQL (or rolled back manually by the user)
3. skip deletion of the table if both 1 & 2 are true

Even if we write the above logic, it would still seem incorrect for an unexpected exception to throw during a call to Dispose.

See: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1065?view=vs-2019#dispose-methods

> A System.IDisposable.Dispose method should not throw an exception. Dispose is often called as part of the cleanup logic in a 
finally clause. Therefore, explicitly throwing an exception from Dispose forces the user to add exception handling inside the finally clause.
> 
> The Dispose(false) code path should never throw exceptions, because Dispose is almost always called from a finalizer.

Also see: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

> To help ensure that resources are always cleaned up appropriately, a Dispose method should be idempotent, such that it is callable multiple times without throwing an exception. Furthermore, subsequent invocations of Dispose should do nothing.

The current codebase also violates this second rule as (I think, didn't verify) it will try to drop the table again upon a second call to Dispose.

I'm not saying we should, but one could argue that the proper way to write `TempTable.Dispose()` would be something like this:

```cs
bool _dropped = false;
void Dispose()
{
    if (!_dropped)
    {
        try
        {
             DropTable(throwIfNotExists: false);
        }
        catch
        {
        }
        _dropped = true;
    }
}
```

Consider this sample:

```cs
try
{
    using (var table = CreateTempTable(...))
    {
        //do some logic
        //uh-oh, we lost internet connectivity, and a call to the database throws an error
    }
}
catch (Exception ex)
{
    // exception handling and log error here
}
```

What is the proper exception and stack trace to be logged?  With the current codebase, or even with the PR as it is now (if we assume bug #798 is fixed), it would log a failure within the `TempTable.Dispose()` method.  However, ideally it would log an error where it first occurred.

Having all exceptions skipped of course has its drawback.  If in the above sample, the inner logic gracefully finished, now we would not have any exception even though the loss of internet caused the temp table to remain.

We should keep in mind also that local temporary tables (e.g. `#MyTable` in SQL Server) will automatically be dropped upon the database connection being closed.  So as long as the `#` character was used in the table name, cleanup will automatically occur regardless of whether `Dispose()` is executed or not.  Of course `CreateTempTable` is not limited to local temp tables.

MaceWindu:
I mean we shouldn't perform drop in `Dispose` if user explicitly called `Drop` method already.

Shane32:
> Failed to rebase baselines PR on origin/master. Delete branch and re-run tests. Error code 1

Shane32:
@MaceWindu Do we have any shortcuts for deleting the baselines PR & branch so tests can be re-run?  Or can you do so manually?

sdanyliv:
@Shane32, if there conflicts in baselines. Drop baseline PR manually and delete baseline branch.

Shane32:
> @Shane32, if there conflicts in baselines. Drop baseline PR manually and delete baseline branch.

I don't have permission in that repo

Shane32:
Perhaps the "developers" group should be given write access to that repo.... not sure what the thought process was there, but I'm okay with whatever. 🤷‍♂️

Shane32:
(I also can't merge a baseline PR on that repo....)

MaceWindu:
@viceroypenguin , @Shane32 added you to baselines repo

Shane32:
@MaceWindu so now run `/azp run test-sybase` for the sybase tests?

sdanyliv:
@Shane32, it is wasting time. Something wrong with docker image.

Shane32:
Ok.  Well, then this should be ready to merge, unless anyone else has a comment.

sdanyliv:
@Shane32, let @MaceWindu to do that. If you do not hurry for sure.

Shane32:
No problem.  Just wanted to be sure nobody was waiting for me on it.

MaceWindu:
I'm still opposed to double Drop behavior. I don't see why TempTable.Dispose should call Drop second time if it was already called.

Shane32:
> I'm still opposed to double Drop behavior. I don't see why TempTable.Dispose should call Drop second time if it was already called.

I have no problem with that -- but this PR is not designed to fix that issue, nor will fixing that issue fix the issue that this PR addresses.

I can change the test in this PR so that if, in the future, someone decides to prevent `Dispose` from dropping a table that was dropped via `Drop`, the test will still work properly.  I'm thinking I could probably write code like this (for purposes of testing this PR):

1. tempTable = CreateTempTable(...)
2. table2 = GetTable(...) or similar to open a separate instance of the same table via name
3. table2.Drop
4. tempTable.Dispose <-- verify that no exception occurs

Is this what you're thinking, @MaceWindu  ?

Shane32:
@MaceWindu I can also provide a test for the exact issue outlined in #2882 -- a failed merge command causing SQL Server to implicitly cause a rollback of a transaction, which deletes a temporary table within the transaction.  However, the test will be based on SQL Server-specific semantics, so I was attempting to write a generic test replicating the underlying cause of the bug.  A Drop followed by a Dispose was an easy way to do that, but if there is a plan to prevent Dispose from calling Drop after Drop is called manually, then the test should be changed.

MaceWindu:
@Shane32, agree that calling Drop allows us to test it easier. Let's try your proposal with dropping table using new instance of ITable in test.

MaceWindu (approved):

PR #2902, opened 2021-03-18
[Feature] TagQuery (TagWith)

### LINKED ISSUE #2903: Feature - query tag (TagWith)
There is one helpful feature in ef core: TagWith. It adds custom tag as a comment before SELECT query. This feature can be very helpful to identify the heaviest queries in a database, because such tag will be stored together with query text in the database query plan cache.

I haven't found anything similar in linq2db and I think it would be great to implement this here as well.

jlukawska:
This PR implements TagWith feature #2903 

Fix #2903


Source/LinqToDB/SqlQuery/Tag.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace LinqToDB.SqlQuery
+{
+	public class Tag : IQueryElement, ICloneableElement
```
MaceWindu, line 7: all ast classes are prefixed with `Sql`, also this is actually just comment, so we should use generic `SqlComment` name

Source/LinqToDB/SqlQuery/Tag.cs
```diff
@@ -0,0 +1,43 @@
...
+	{
+		public QueryElementType ElementType => QueryElementType.Tag;
```
MaceWindu, line 9: QueryElementType.Comment;

Source/LinqToDB/SqlQuery/Tag.cs
```diff
@@ -0,0 +1,43 @@
...
+
+			return clone;
+		}
+
+		public StringBuilder ToString(StringBuilder sb, Dictionary<IQueryElement, IQueryElement> dic)
+		{
+			return BuildSql(sb, Value);
+		}
+
+		internal static StringBuilder BuildSql(StringBuilder sb, string? tagValue)
```
MaceWindu, line 30: sql generation should be moved to SqlBuilder as it is db-dependent:
- access doesn't support comments
- instead of newline removal we should generate multiple single-line comments or multiline comment if supported

MaceWindu (changes requested):

Source/LinqToDB/SqlQuery/Tag.cs
jlukawska: Ok, I've moved sql generation to BasicSqlBuilder, but it doesn't seem very good to duplicate the code in SqlComment.ToString(). I guess it is called for debugging purposes only?

Source/LinqToDB/DataProvider/Access/AccessSqlBuilderBase.cs
```diff
@@ -301,5 +301,9 @@ protected override void BuildMergeStatement(SqlMergeStatement merge)
 		{
 			throw new LinqToDBException($"{Name} provider doesn't support SQL MERGE statement");
 		}
+
+		protected override void BuildTag(SelectQuery selectQuery)
```
sdanyliv, line 305: I think we have to add tags to `SqlStatement` base class.

Source/LinqToDB/Linq/Builder/TagQueryBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+	{
+		protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
+		{
+			var sequence = builder.BuildSequence(new BuildInfo(buildInfo, methodCall.Arguments[0]));
+
+			var tag = (string?)methodCall.Arguments[1].EvaluateExpression();
+
+			if (!string.IsNullOrWhiteSpace(tag))
+			{
+				sequence.SelectQuery.Tag.Parts.Add(tag!);
```
sdanyliv, line 16: Move tags to `ExpressionBuilder` class. We will reuse them when building `SqlStatement`.

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -295,6 +295,7 @@ protected virtual void BuildUpdateQuery(SqlStatement statement, SelectQuery sele
 
 		protected virtual void BuildSelectQuery(SqlSelectStatement selectStatement)
 		{
+			BuildStep = Step.Tag;           BuildTag(selectStatement.SelectQuery);
```
sdanyliv, line 298: Every `SqlStatement` builders should have this BuildTag call.

Source/LinqToDB/SqlQuery/SelectQuery.cs
```diff
@@ -75,6 +78,8 @@ internal SelectQuery(int id)
 		public SqlWhereClause   Having  { get; private set; } = null!;
 		public SqlOrderByClause OrderBy { get; private set; } = null!;
 
+		public SqlComment       Tag     { get; private set; } = null!;
```
sdanyliv, line 81: Move to `SqlStatement`

sdanyliv (commented):
When you move tags from `SelectQuery` to `SqlStatement`. Corect this method:
https://github.com/linq2db/linq2db/blob/8246efebdfbd3a7e5ab838b5425dce80a9147da0/Source/LinqToDB/Linq/Query.cs#L241
Like this
```cs
internal override void Init(IBuildContext parseContext, List<ParameterAccessor> sqlParameters)
{
	var statement = parseContext.GetResultStatement();
	statement.Tags.AddRange(parseContext.Builder.Tags);
	Queries.Add(new QueryInfo
	{
		Statement          = statement,
		ParameterAccessors = sqlParameters,
	});
}
```

Source/LinqToDB/LinqExtensions.cs
```diff
@@ -3422,5 +3422,61 @@ internal static IQueryable<TSource> SelectDistinct<TSource>(this IQueryable<TSou
...
+		/// db.Table.TagWith("my tag");
+		/// </code>
+		/// </summary>
+		/// <typeparam name="TSource">Table record mapping class.</typeparam>
+		/// <param name="source">Source data query.</param>
+		/// <param name="tagValue">Tag text to be added as comment before generated query.</param>
+		/// <returns>Query with tag.</returns>
+		[LinqTunnel]
+		[Pure]
+		public static IQueryable<TSource> TagQuery<TSource>(this IQueryable<TSource> source, string tagValue)
```
sdanyliv, line 3441:
```suggestion
		public static IQueryable<TSource> TagQuery<TSource>(this IQueryable<TSource> source, [SqlQueryDependent] string tagValue)
```

MaceWindu:
I will add my commit today or tomorrow with extra tests and support for new ast element in visitors

MaceWindu:
@jlukawska 
I've made some changes:
- switched to block comment generation instead of multiple line comments with sanitation of comment test to avoid sql injection
- added tests for all statements (except SqlCreateTableStatement)
- fixed visitors so comment not lost on ast conversion. also requred for remote context serialization to work

Remaining issues (should be easy to fix as they all are similar):
- DropTable has no tag
- InsertOrUpdate has no tag for most of providers
- Insert has no tag for MySql
- TruncateTable has no tag for DB2 and postgresql
- Merge has no tag

Tests/Base/TestBase.cs
```diff
@@ -1350,6 +1349,11 @@ public virtual void OnAfterTest()
 			CustomTestContext.Release();
 		}
 
+		protected string GetCurrentBaselines()
+		{
+			return CustomTestContext.Get().Get<StringBuilder>(CustomTestContext.BASELINE)?.ToString() ?? string.Empty;
```
MaceWindu, line 1354: completely forgot that baselines not collected when they are not configured. Would be nice if you will refactor it to collect always and just don't save them if they are not configured

jlukawska:
> Remaining issues (should be easy to fix as they all are similar):
> 
>     * DropTable has no tag
> 
>     * InsertOrUpdate has no tag for most of providers
> 
>     * Insert has no tag for MySql
> 
>     * TruncateTable has no tag for DB2 and postgresql
> 
>     * Merge has no tag

Done almost all things: merge is not ready yet (I want to run merge tests on my machine first).

sdanyliv:
@jlukawska, do not forge to correct this one https://github.com/linq2db/linq2db/pull/2902/files#r602726103

jlukawska:
I'll fix Oracle failures, but what about Sybase? The error is: "There is no server listening at localhost:5000." Seems that the database server is not able to start.

MaceWindu:
sybase currently doesn't work (works only for mac, but it is not configured for external PRs)

Tests/Base/TestBase.cs
jlukawska: @MaceWindu if you say "they are not configured" you mean BaselinePath setting (then it seems to work as you expect) or maybe BASELINE_DISABLED checking?

MaceWindu: hmm, you are right, looks like it is already collecting baselines and [BaselinesPath](https://github.com/linq2db/linq2db/blob/master/CONTRIBUTING.md#userdataprovidersjson-example-with-description) only enables writing to disk

MaceWindu (approved):

sdanyliv (approved):

PR #2906, opened 2021-03-20
Add operation type to trace events, more async disposable support

### LINKED ISSUE #2898: TraceInfo: ExecuteNonQuery vs ExecuteReader difference
For trace events, there is no way to tell ExecuteNonQuery from ExecuteReader or ExecuteScalar.
ExecuteNonQuery method produces BeforeExecute and AfterExecute trace steps, but ExecuteReader
has an additional Completed step which can be used to measure fetch time.

Consider an application that is able to log its SQL statements. Each statement should appear only once.
For ExecuteNonQuery commands the application displays the start time and execution time.
For ExecuteReader commands the application displays the start time, the execution time and the fetch time.

To track execution time, the app can just handle AfterExecute trace steps.
But for data reader, the app should handle Completed trace steps instead.
Currently, there is no way to tell whether Completed step is going to follow or not.

As a quick hack, the app can assume that commands starting from "SELECT " are ExecuteReader.
But that's not accurate at all. Some databases can use SELECT to execute stored procedures,
so there is nothing wrong with executing SELECT statements using ExecuteNonQuery.

Can we add something like an OperationType to the TraceInfo?

MaceWindu:
Fix #2898

Also:
- set `IsAsync` flag for all trace events
- add `CloseAsync` to `IDataContext`
- add `IAsyncDisposable` to `IDataContext` and `IQueryRunner`
- obsolete `DataConnection`'s `DisposeAsync` and `CloseAsync` methods with `CancellationToken` parameter

### DIFF
--- a/Source/LinqToDB/Async/AsyncDbConnection.cs
+++ b/Source/LinqToDB/Async/AsyncDbConnection.cs
@@ -42,7 +42,7 @@ namespace LinqToDB.Async
 			=> Task.FromResult(AsyncFactory.Create(BeginTransaction()));
 #elif !NETSTANDARD2_1PLUS
 		public virtual ValueTask<IAsyncDbTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default)
-			=> new ValueTask<IAsyncDbTransaction>(AsyncFactory.Create(BeginTransaction()));
+			=> new (AsyncFactory.Create(BeginTransaction()));
 #else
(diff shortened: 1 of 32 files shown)


MaceWindu:
no review yet. I will add IAsyncDisposable to query runner and fix event operations here

Shane32 (approved):

sdanyliv (approved):

PR #2911, opened 2021-03-21
Dependecies update

MaceWindu:
- Update dependencies:
  - ~[Microsoft.NET.Test.Sdk](https://www.nuget.org/packages/Microsoft.NET.Test.Sdk): 16.8.3 -> 16.9.1~ borken, waiting for 16.9.2
  - [NUnit3TestAdapter](https://www.nuget.org/packages/NUnit3TestAdapter): 3.17.0 -> 4.0.0-beta.1
  - [FastExpressionCompiler](https://www.nuget.org/packages/FastExpressionCompiler): 3.0.0-preview-07 -> 3.0.0
  - [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection): 3.1.12 -> 3.1.13
  - [Microsoft.Extensions.Logging.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions): 3.1.12 -> 3.1.13
  - [Microsoft.Data.Sqlite](https://www.nuget.org/packages/Microsoft.Data.Sqlite): 5.0.3 -> 5.0.4
  - [MySqlConnector](https://www.nuget.org/packages/MySqlConnector): 1.3.0 -> 1.3.2
  - [Oracle.ManagedDataAccess](https://www.nuget.org/packages/Oracle.ManagedDataAccess): 19.10.1 -> 19.11.0
  - [Oracle.ManagedDataAccess.Core](https://www.nuget.org/packages/Oracle.ManagedDataAccess.Core): 2.19.101 -> 2.19.110
  - [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json): 12.0.3 -> 13.0.1
- fix net472 testing from VS
- fix potential deadlocks on task await

### DIFF
--- a/Source/LinqToDB/Async/SafeAwaiter.cs
+++ b/Source/LinqToDB/Async/SafeAwaiter.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 
 namespace LinqToDB.Async
 {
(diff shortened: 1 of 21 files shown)


viceroypenguin (approved):
I assume APTCA had something to do w/ testing in VS?

MaceWindu:
Yep, after adding ValueTask for IAsyncDisposable it became unusable

sdanyliv (approved):

Source/LinqToDB/Async/SafeAwaiter.cs
```diff
@@ -0,0 +1,27 @@
...
+	/// <summary>
+	/// Provides deadlock-free task await helpers.
+	/// </summary>
+	internal static class SafeAwaiter
+	{
+#if NATIVE_ASYNC
+		public static T Run<T>(Func<ValueTask<T>> task)
+		{
+			// awaited ValueTask retrieved in Task.Run context as doing it in main thread could cause deadlock too
+			var awaitable = Task.Run(async () => await task().ConfigureAwait(Common.Configuration.ContinueOnCapturedContext));
```
viceroypenguin, line 15: Do we still need `.ConfigureAwait()` considering that we've cleared the context by running it inside of `Task.Run()`? In all my contentious task work (i.e. doing the same thing), just running in `Task.Run()` was enough, and didn't need an add'l `.ConfigureAwait()`

viceroypenguin: Actually, can probably be simplified to `Task.Run(() => task())`, since without the `.ConfigureAwait()`, there's no need for the add'l `async`/`await` compiler abstraction and `Task.Run()` will handle a naked `ValueTask`/`Task` for us.

Source/LinqToDB/Async/SafeAwaiter.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace LinqToDB.Async
+{
+	/// <summary>
+	/// Provides deadlock-free task await helpers.
+	/// </summary>
+	internal static class SafeAwaiter
+	{
+#if NATIVE_ASYNC
+		public static T Run<T>(Func<ValueTask<T>> task)
```
viceroypenguin, line 12: Should there be `Run<T>(Func<Task<T>>)` version as well? Or are we only using `ValueTask<>` in the places that matter?

MaceWindu: we don't have use-case for Task

Source/LinqToDB/Async/SafeAwaiter.cs
MaceWindu: Task.Run have no Func<ValueTask> overload

MaceWindu: I don't really care about `ConfigureAwait` here, but at least in this form there is no build warning :)

viceroypenguin: Ha! :D Then I'm good leaving as is.

PR #2914, opened 2021-03-24
Multi-table insert

jods4:
Fix #2831

This is the same branch as #2899 but this time created from linq2db repo so that I can run tests.

Every change requested in that other PR was done.
I also added some basic tests, please review if they're ok for you.

I did not add tests for `IEnumerable`, [see this comment](https://github.com/linq2db/linq2db/pull/2899#issuecomment-804445109)

PR #2915, opened 2021-03-25
Release 3.3.0

MaceWindu:


jods4:
@MaceWindu if you're doing a new release soon, it would make me very happy if we could include multi-table insert into it. 
#2914 is done, I have addressed all feedback and included tests. I'm just having trouble running the tests, Oracle MERGE tests seem to fail.

MaceWindu:
We don't plan to postpone or include additional features to current release, but nothing prevents us from doing another release later, even next week

PR #2923, opened 2021-04-03
Drop SQL Server 2000 support

### LINKED ISSUE #2812: Announcement: Drop SQL Server 2000 support in v4
We plan to drop SQL Server 2000 support from linq2db in next major version (v4).

1. There is no support for it already in .net core (it uses different protocol which wasn't ported to .net core)
2. We don't perform regular testing for it

If anyone interested in SQL Server 2000 long-term support, we could extract it to separate external project.

MaceWindu:
Fix #2812

Press "F"


viceroypenguin (approved):
F

sdanyliv (approved):

PR #2924, opened 2021-04-04
Don't release data context connection on persistent context

MaceWindu:
Cherry-pick fix from #2728


MaceWindu:
blocked. looks like it introduced connection leak in tests, needs investigation

MaceWindu:
This could be `by-design` behavior, so we shouldn't probably "fix" it in v3

PR #2941, opened 2021-04-10
Convert more code to interceptors

### LINKED ISSUE #2927: Interceptors migration
For version 4 we add interceptors support. To get familirar with e.g. check [EF.Core](https://docs.microsoft.com/en-us/ef/core/logging-events-diagnostics/interceptors) implementation.

We plan to remove bunch of existing static options and events and replace them with interceptors. #2728 already introduced initial support with couple of interceptors (also check [release notes](https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#interceptors))

Below is the list of additional interceptors we should implement:
- [x] static `DbCommandProcessorExtensions` and `IDbCommandProcessor` interface [here](https://github.com/linq2db/linq2db/tree/master/Source/LinqToDB/Data/DbCommandProcessor) should be replaced with corresponding events on `ICommandInterceptor`
- static `RetryPolicy` [factory](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Common/Configuration.cs#L321) replaced with event on `IConnectionInterceptor`
- [x] `IDataContext.OnClosing` [event](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/IDataContext.cs) + `DataConnection.OnClosing/OnClosed` [events](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Data/DataConnection.cs#L1109-L1113) + `DataContext.OnCLosing` [event](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataContext.cs#L384) replaced with events on `IConnectionInterceptor`. Note that removal of event from `IDataContext` could require interceptors support added to remote context
- [x] `IEntityServices` [interface](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/IEntityServices.cs) +  `DataConnection.OnEntityCreated` [delegate](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Data/DataConnection.cs#L1116) and `DataContext.OnEntityCreated` [delegate](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataContext.cs#L387) replaced by event on `IConnectionInterceptor` (or new interceptor like IEntityInterceptor?)

Consider migration:
- [Compilation.SetExpressionCompiler()](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Common/Configuration.cs) replaced with interceptor. Probably no, as it doesn't fit good for interceptor.

@linq2db please provide your comments on proposed changes if any or suggest additional interceptor events

### LINKED ISSUE #2934: Add LinqToDbConnectionOptions support to DataContext
Allow fluent configuration support of DataContext same way we do it for DataConnection

MaceWindu:
Fix #2934
Fix #2927

Interceptors:
- `IDbCommandProcessor` interface (and `DbCommandProcessorExtensions` static setting) replaced with same events on `ICommandInterceptor` interceptor interface
- `IEntityServices` interface in `IDataContext` (`OnEntityCreated` event) replaced with `IDataContextInterceptor.EntityCreated` interceptor event
- `DataConnection.OnClosed` + `IDataContext.OnClosing` events replaced with events on `IDataContextInterceptor` (separate events for `IDataContext.Close` and `IDataContext.CloseAsync`). So instead on 1+2 events we support all of them on all contexts
- `IDataContext` interface updated with `AddInterceptor`/`GetInterceptors` methods
- `IDataContextInterceptor` support added to remote context
- `IDataContextInterceptor` currently passed down do `DataConnection` from `DataContext` (expected?)

Other fixes/improvements/refactorings:
- mark `Dispose` methods virtual for `RemoteDataContextBase`
- replace internal `Option<T>` class with public struct `Option<T>` struct and use it in new interceptors
- add `DataContext(LinqToDbConnectionOptions options)` constructor to `DataContext` to support options builder in `DataContext` too
- `DataContext.CloneDataConnection` and `DataContext.CreateDataConnection` signatures changed to use `LinqToDbConnectionOptions`
- updated SQLite SQL builder to explicitly specify `temp` schema for temporary tables to avoid name conflicts (could be backported to v3 if anyone interested)
- `DataContext.QueryRunner.DataContext` now returns data context instance instead of underlying DataConnection
- add `DataContext.ReleaseQueryAsync()` to avoid sync connection close from async code

Cleanup:
- removed never unused `autoCommitOnDispose` parameter from `DataContext.BeginTransactionAsync` and `DataContext.BeginTransaction` methods
- removed `Tools.IsNullOrEmpty(string)`, `Tools.IsNullOrWhiteSpace(string)` methods that just call corresponding `string` methods and used non-systematically
- remove setter from `IQueryRunner.DataContext` property

Tests:
- update tests infrastructure a bit to support `SequentialAccessCommandInterceptor` and interceptors in general. Call context creation from `TestBase` for tests that don't do it without good reason


MaceWindu:
Added LinqToDbConnectionOptions support to datacontext as initial commit. Will add interceptors with next commits

Source/LinqToDB/DataContext.cs
```diff
@@ -282,7 +337,7 @@ internal DataConnection GetDataConnection()
 		private void AssertDisposed()
 		{
 			if (_disposed)
-				throw new ObjectDisposedException(GetType().FullName);
+				throw new ObjectDisposedException(nameof(DataContext));
```
sdanyliv, line 340: Better to leave previous code. It helps if there is descendant from DataContext.

Source/LinqToDB/DataContext.cs
```diff
@@ -318,74 +373,41 @@ Expression IDataContext.GetReaderExpression(DbDataReader reader, int idx, Expres
...
-				ConfigurationString     = ConfigurationString,
-				ConnectionString        = ConnectionString,
 				KeepConnectionAlive     = KeepConnectionAlive,
-				DataProvider            = DataProvider,
 				ContextID               = ContextID,
-				MappingSchema           = MappingSchema,
-				InlineParameters        = InlineParameters,
-				_commandInterceptors    = _commandInterceptors?.Clone(),
-				_connectionInterceptors = _connectionInterceptors?.Clone(),
+				InlineParameters        = InlineParameters
```
sdanyliv, line 396: Why `InlineParameters` not in options?

MaceWindu: Cause it never was here. We can include it probably

MaceWindu:
Second commit is about `IDbCommandProcessor` replacement with interceptors:

- `IDbCommandProcessor` interface (and `DbCommandProcessorExtensions` static setting) replaced with same events on `ICommandInterceptor` interceptor interface
- remove `Tools.IsNullOrEmpty(string)`, `Tools.IsNullOrWhiteSpace(string)` methods that just call corresponding `string` methods and used non-systematically
- replace internal `Option<T>` class with public struct `Option<T>` struct and use it in new interceptors
- mark `Dispose` methods virtual for `RemoteDataContextBase`
- update tests infrastructure a bit to support `SequentialAccessCommandInterceptor` and interceptors in general. Call context creation from `TestBase` for tests that don't do it without good reason
- updated SQLite SQL builder to explicitly specify `temp` schema for temporary tables to avoid name conflicts (could be backported to v3 if anyone interested)

MaceWindu:
Next one: `OnEntityCreated` event replaced with IDataContext-level interceptor

- `IEntityServices` interface in `IDataContext` (`OnEntityCreated` event) replaced with `IDataContextInterceptor.EntityCreated` interceptor event
- `IDataContext` interface updated with `AddInterceptor`/`GetInterceptors` methods
- `IDataContextInterceptor` support added to remote context
- remove setter from `IQueryRunner.DataContext` property (more setters removal planned)
- `DataContext.QueryRunner.DataContext` now returns data context instance instead of underlying DataConnection
- `IDataContextInterceptor` currently passed down do `DataConnection` from `DataContext` (expected?)

MaceWindu:
Another one: Close context events

- `DataConnection.OnClosed` + `IDataContext.OnClosing` events replaced with events on `IDataContextInterceptor` (separate events for `IDataContext.Close` and `IDataContext.CloseAsync`). So instead on 1+2 events we support all of them on all contexts
- add `DataContext.ReleaseQueryAsync()` to avoid sync connection close from async code

MaceWindu:
@linq2db I would like to merge it into v4 branch if there is no objections to reduce merge strain. While I have a lot of ideas for future refactorings, it could be done using new PRs later. It would be easier to review and I actually want to concentrate to model generators now, so I don't left behind work that is technically done...

MaceWindu:
I've updated first message with full  list of changes from separate commits

MaceWindu:
Currently I'm going through baselines and most of them are expected changes like:
- new `temp.` prefix for temp tables in sqlite tests
- replacement of `ToList` with `Count()` in data context tests

If there will be some unexpected changes I will point it here late

MaceWindu:
release and migration notes updated

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -213,21 +216,29 @@ public async Task DisposeAsync()
...
+			await CloseAsync().ConfigureAwait(Configuration.ContinueOnCapturedContext);
 		}
 
 		#region ExecuteNonQueryAsync
 
-		protected virtual Task<int> ExecuteNonQueryAsync(DbCommand command, CancellationToken cancellationToken)
+		protected virtual async Task<int> ExecuteNonQueryAsync(DbCommand command, CancellationToken cancellationToken)
 		{
-			return CurrentCommand!.ExecuteNonQueryExtAsync(cancellationToken);
+			var result = Option<int>.None;
```
viceroypenguin, line 226: Why `Option<int>` instead of `int?`? `int?` will have better support from the JIT, and on the last line, you can do `result ?? await ...` or similar. Or leave the last line alone, since `int?` has the same interface.

viceroypenguin: Nvm - consistency w/ others.

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -301,37 +301,25 @@ Expression BuildTableExpression(bool buildBlock, Type objectType, Tuple<int, Sql
...
-					var args = new EntityCreatedEventArgs
-					{
-						Entity      = entity,
-						DataContext = context
-					};
-
-					onEntityCreated(args);
-
-					return args.Entity;
+					args   = args ?? new DataContextEventData(context);
```
viceroypenguin, line 307: `args ??= new DCED(...);`

Source/LinqToDB/ServiceModel/RemoteDataContextBase.cs
```diff
@@ -331,36 +332,40 @@ protected void ThrowOnDisposed()
...
-			return TaskEx.CompletedTask;
+			if (_contextInterceptors != null)
+				_contextInterceptors.Apply((interceptor, arg) => interceptor.OnClosed(arg), new DataContextEventData(this));
 		}
 
-		void Close()
+		async Task IDataContext.CloseAsync()
 		{
-			OnClosing?.Invoke(this, EventArgs.Empty);
+			if (_contextInterceptors != null)
```
viceroypenguin, line 344: There appears to be two checks for `if (_contextInterceptors != null)`. Either something should be in between them (to do closing), or they should be in the same `if(){ ... }` block.

viceroypenguin (approved):
Nothing stands out to me. I'm good w/ doing a v4 preview with this.

sdanyliv (approved):

PR #2945, opened 2021-04-10
Improve thread-safety

### LINKED ISSUE #2940: Changing TraceSwitch at runtime
I have a configuration in my apps configuration to turn on Linq2Db trace.
Basically, I do this at startup:
```csharp
if (Config.LogSql)
{
  DataConnection.TurnTraceSwitchOn();
  DataConnection.WriteTraceLine = (message, _, _) => Log.Debug(message);
}
```

My config is live-reloaded when it changes. The trace is usually off in production (of course), but it would be useful for me to be able to turn it on shortly to diagnose some issue and then turn it back off, without stopping my app.

Basically my idea is to always define `WriteTraceLine` and then call `TurnTraceSwitchOn(level)` with `TraceLevel.Info` or `TraceLevel.Off` whenever configuration changes.

I would like to ensure a few things:

1. `WriteTraceLine` is not called when trace level is `Off`, right? There's no harm in setting it always?
2. Calling `TurnTraceSwitchOn(TraceLevel)` at runtime would actually work an be applied to all new connections opened afterwards?
3. Is this thread-safe, even if threads are concurrently working with opened `DataConnection`?

From a glance at the code, I have the impression that the answers are:
1. Yes
2. Yes, it would even apply to all connections alive if they don't override the default trace switch.
3. Almost. 

For that last point, my analysis is that changing the trace level is assigning a new immutable `TraceSwitch` reference to a global static, so code that is running will see either the previous or the new one and both option will run fine.

The only catch is that there is no synchronisation when writing the `TraceSwitch` variable:
https://github.com/linq2db/linq2db/blob/59d927f26d2effa5a460f467bec129bb867b6d41/Source/LinqToDB/Data/DataConnection.cs#L584-L599

Publishing a reference is an atomic operation that is fine if you do a _release_ write (memory fence).
The problem here is that memory writes re-ordering could mean another thread could see the new `TraceSwitch` reference published before the `TraceSwitch` instance is fully constructed (i.e. a partially initialised instance, which can of course lead to very elusive bugs).

You can easily fix that by doing `Volatile.Write(ref _traceSwitch, new TraceSwitch())` when publishing the new trace switch, or mark `_traceSwitch` as `volatile`.

jods4:
The "Publish a ref atomically" pattern is safe only with a memory half-fence that prevents writes to be reordered after the publishing write (aka _Release Write_).

Otherwise compiler and hardware are free to reorder writes as long as it isn't noticeable on a single thread, which can lead to other threads seeing the published reference pointing to a partially initialized object.

Fixes #2940

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.cs
+++ b/Source/LinqToDB/Data/DataConnection.cs
@@ -581,11 +581,11 @@ namespace LinqToDB.Data
 		/// defaults to off unless library was built in debug mode.
 		/// <remarks>Should only be used when <see cref="TraceSwitchConnection"/> can not be used!</remarks>
 		/// </summary>
-		public  static TraceSwitch  TraceSwitch
+		public static TraceSwitch TraceSwitch
 		{
 			// used by LoggingExtensions
 			get => _traceSwitch;
-			set => _traceSwitch = value;
+			set => Volatile.Write(ref _traceSwitch, value);
 		}
 
 		/// <summary>
--- a/Source/LinqToDB/Metadata/MetadataReader.cs
+++ b/Source/LinqToDB/Metadata/MetadataReader.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 
 namespace LinqToDB.Metadata
 {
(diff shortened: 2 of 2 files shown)


viceroypenguin (approved):

MaceWindu (approved):
not like I'm convinced that there is a problem, but have no objections

MaceWindu:
@jods4 , please rebase and you can merge it

PR #2946, opened 2021-04-11
Feat: IsDistinctFrom

jods4:
This PR adds `Sql.IsDistinctFrom` and `Sql.IsNotDistinctFrom`, which are intended to map to the standard `IS [NOT] DISTINCT FROM`, which is basically `=` but supports `NULL` and always returns `TRUE` or `FALSE`, never `UNKNOWN`.

This is a useful function to have at hand, esp. since the equivalent "old-school" SQL is a bit long.
Notice this SQL standard is poorly supported (by only 3 LinqToDB providers) but many have proprietary variations such as `IS [NOT]` in SQLite or `<=>` in MySql. Tests will show if I got all of them right on the first try (very unlikely!).

This is also the intended behavior of equality comparisons when `CompareNullsAsValues = true`.
You may want to consider whether `==` should be translated into `IS NOT DISTINCT FROM` in this case. Currently it becomes `a = b or a is null and b is null` but as you can see from this PR, many DB have better proprietary syntax than that.
Also, I noticed that the current translation isn't 100% correct as it can still return `UNKNOWN`, for e.g. when `a = 1 and b is null`. This usually isn't a problem as `UNKNOWN` eventually coerces to `FALSE`, but in complex expressions this could lead to bugs.

This PR implements `IS DISTINCT FROM` as a custom expression, I'm not sure if you'd want to merge it into the expression engine as a basic block. As is it works fine but it lacks the smarts for example to convert a `!a.IsDistinctFrom(b)` into `a.IsNotDistinctFrom(b)`, to notice that if both expression are not-null it could be simplified to a simple equality, or to notice that if a value is null and the other non-nullable, this would always be false (resp. true). I'm not sure any of this is worth the effort.

Notice I didn't put a `PreferServerSide`, which means that if both arguments are constants (unlikely but hey), this predicate can be evaluated client-side and dropped.


Source/LinqToDB/Sql/Sql.cs
```diff
@@ -132,6 +132,42 @@ public static bool NotBetween<T>(this T? value, T? low, T? high)
...
+
+		[Sql.Expression(               "CASE WHEN {0} = {1} OR {0} IS NULL AND {1} IS NULL THEN 1 ELSE 0 END = 1", IsPredicate = true)]
+		[Sql.Expression(PN.DB2,        "{0} IS NOT DISTINCT FROM {1}",                                             IsPredicate = true)]
+		[Sql.Expression(PN.Firebird,   "{0} IS NOT DISTINCT FROM {1}",                                             IsPredicate = true)]		
+		[Sql.Expression(PN.MySql,      "{0} <=> {1}",                                                              IsPredicate = true)]
+		[Sql.Expression(PN.Oracle,     "DECODE({0}, {1}, 1, 0) = 1",                                               IsPredicate = true)]
+		[Sql.Expression(PN.PostgreSQL, "{0} IS NOT DISTINCT FROM {1}",                                             IsPredicate = true)]
+		[Sql.Expression(PN.SQLite,     "{0} IS {1}",                                                               IsPredicate = true)]
+		public static bool IsNotDistinctFrom<T>(this T value, T? other) where T: struct => EqualityComparer<T?>.Default.Equals(value, other);
+
```
MaceWindu, line 170: Something doesn't look right regarding nullability here.

What about:
- `Func<T>(T?, T?) where T: struct` - both nullable
- `Func<T>(T, T?) where T: struct` - ommit left value null check for fallback expression
- `Func<T>(T?, T) where T: struct` - ommit right value null check for fallback expression
- `Func<T>(T?, T?) where T: class` - same as first one as we cannot overload on NRT annotation

jods4: @MaceWindu I know it doesn't look right... but it is!
Those 2 overloads are all that is required to make any combination compile.
Check this full test matrix:
https://sharplab.io/#v2:CYLg1APgxAdgrgGwQQwEYIKYAIMzZgWAChiABARgDYtSAmLAUQA8AXLAb2K25oGYaqWVAHthCLAEEAPABUAfAAoWACwCWAZywysTADRasATwCUWALxysLAE5wMAbi49S/CtRFjJsxSo0G9WgD8RqYA7soY1tgyIFjqNnAAxmwWVrYOxAC+xGSCdFoY8ZqcRNxuNAAsWgrGTiU8zuQADHHk5lgARB366vRmnR2OpQ0CTcEw6m398Ej6E31YMwhDI1iqMGyqU1hN+qoLTSsj6yzjW+1Lc/sXiMtODfc8APRPWABKGABm6vrISFiJYQAW1Q62QLFUwgmj24zBYADoJApJj1aMYjg04YjkeQ5r10TDGKxsRNcXE0fZVoSsUjSXiKYTCZNsfiMTxmbTWYTSSyGcMGjzOXyqfznq8AGp/Ow/LB/cSAkFgiFQ9TU4lIrZ7YUjGkKTWLfYE0Ww9UKGD6w1sk0I2n683ah7GtbkbGWwlbElup3ml0ah08H2e7XZIiZIA=

This code works and is a good first step for me to check if I have to correct codegen for all providers.

But I believe this call should be compiled into a `SqlPredicate` instead of just using a fixed expression.
That way we could optimize the generated code depending on the runtime nullability of its parameters, invert it, etc.
It can also be used for `==` when `CompareNullsAsValues` is true, leading to simplification in `SqlBuilder` and improvements in codegen.

jods4:
Instead of using `[Expression("...")]` I have turned `IsDistinctFrom` into its own `SqlPredicate`.

This enables a lot more finesse when handling various cases that can be simplified.
Here's a table with all supported cases and the SQL generated. 
`PN` is a possibly null value, `NN` is never null.
The operator is symetric, permutations of `A` and `B` are not explicitely in the table.

| A | B | A.IsDistinctFrom(B)
|---|---|---
| 1 | 2 | true (client-side evaluation)
| NN | null | true
| PN | null | A IS NOT NULL
| NN | NN | A <> B
| PN | PN | A IS DISTINCT FROM B
| PN | NN | A IS DISTINCT FROM B

For cases that can't be simplified and need `A IS DISTINCT FROM B`, here's what each provider uses:
| Provider | SQL
|---|---
| BasicSqlProvider | A IS DISTINCT FROM B
| Access | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1
| DB2 | A IS DISTINCT FROM B
| Firebird | A IS DISTINCT FROM B
| Informix | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1
| MySql | A <=> B
| Oracle | DECODE(A, B, 0, 1) = 1
| Postgre | A IS DISTINCT FROM B
| SAP Hana | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1
| SQL CE | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1
| SQLite | A IS NOT B
| SQL Server | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1
| Sybase | CASE WHEN A = B OR A IS NULL AND B IS NULL THEN 0 ELSE 1 END = 1

I will add more unit tests later.

jods4:
I did NOT change any existing code, in particular how `==` is handled when `CompareNullsAsValues` is true.

In theory that operator could be directly converted into `SqlPredicate.IsDistinct` and `ExprExpr` could be for SQL semantics only.
Sadly, contrary to what I wrote previously, it wouldn't simplify the optimizer/builder much because `ExprExpr` needs to keep `WithNull` for its handling of other operators: `<`, `<=`, `>=`, `>`.
I would find even more confusing that `WithNull` applies to some operators (greater/less) and not others (equals).

Completely removing  `WithNull` from `ExprExp` is not a small task and I'm not even sure what the semantics of `>` are with `CompareNullsAsValues`... so I'm not gonna tackle that now.

The work on `SqlPredicate.IsDistinct` is a first step in that direction, if anyone wants to pursue that.

jods4:
@sdanyliv @MaceWindu I was reviewing my failing tests and I found a LinqToDb bug in Informix.

This query:
```csharp
db
  .SelectQuery(() => new { ID = 1 })
  .Count(x => 5.IsNotDistinctFrom(5)); // Client-evaluated to true, any condition will do...
```

Generates (Informix):
```sql
SELECT FIRST 1
Count(*)
FROM SYSTABLES
```

And that returns 63, instead of expected 1.

I see what happens here. `SYSTABLES` is used a bit like `DUAL` in Oracle except it contains multiple rows.
`SELECT FIRST 1` is supposed to return only the first row, but that doesn't work here because of `COUNT(*)` aggregation that sees all rows before `FIRST 1` is applied.

For reference, test source:
https://github.com/linq2db/linq2db/blob/is-distinct/Tests/Linq/Linq/IsDistinctFromTests.cs#L80
Test run failure: 
https://dev.azure.com/linq2db/linq2db/_build/results?buildId=5634&view=ms.vss-test-web.build-test-results-tab&runId=118328&resultId=101359&paneView=debug

jods4:
@sdanyliv @MaceWindu I have one weird failure for MySQL on Linux... I don't think it specific to IS DISTINCT FROM.
Some tests generate this query:
```sql
SELECT
Count(*) as cnt
WHERE
1 = 0
```
And it fails with:
```
MySql.Data.MySqlClient.MySqlException : You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'WHERE
1 = 0' at line 3
```
I'm not familiar with this provider, maybe you _need_  a FROM in this query?
Did I just found another LinqToDb bug?

MaceWindu:
@jods4 regarding mysql failure - it fails only on old mysql (e.g. 5.5). I'll check what we can do here

MaceWindu:
yep, as you suggested, `WHERE` required `FROM` for older mysql: https://docs.oracle.com/cd/E19957-01/mysql-refman-5.5/sql-syntax.html#select

MaceWindu (approved):

viceroypenguin (approved):

MaceWindu:
@sdanyliv check my last commit - with your changes to predicates, bool `Sql.Expr` is not recognized as predicate anymore, so I've added predicate flag to `ExprBuilder`.

sdanyliv (approved):

PR #2949, opened 2021-04-12
Better nulls handling in Contains

jods4:
1. LinqToDb attempts to sniff `null` constants in `Sql.In` arguments and include them in a comparison (with `x IN (1,2) or x IS NULL` and `x NOT IN (1, 2) and x IS NOT NULL`). 
This is a non-standard SQL behavior that really should have been behind `CompareNullsAsValues`, but it is not. To not break backwards compatiblity, I created a new configuration to disable it: `CheckNullsInContains`.
The case `null not in (null)` can become just too confusing, so implicitly if `CompareNullsAsValues` is true, then `CheckNullsInContains` is implicitly true as well.

I did include client-side support to keep things coherent. This applies to `Sql.In`, but not to `Array.Contains` of course, whose client-side behavior is defined by C#.

I noticed regular LinqToDb code makes use of `Sql.In` internally as a helper method, which isn't great esp. if its exact behaviour depends on user configuration. I did a review of all usages and none passes `null` as a list value, so that's ok.

 2. I fixed a bug in SQL builder where negated IN translated `a not in (1, null)` into `a not in (1) or a is not null` which is incorrect because it's true for `1`. The correct version is `a not in (1) and a is not null`.

3. Removed a case of double parens `(())`

4. Added a bunch of tests for those cases.

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -2135,7 +2135,7 @@ namespace LinqToDB.SqlProvider
 		{
 			var firstValue      = true;
 			var len             = StringBuilder.Length;
-			var checkNull       = Configuration.Linq.CheckNullInContains || Configuration.Linq.CompareNullsAsValues;
+			var checkNull       = Configuration.Linq.CheckNullsInContains || Configuration.Linq.CompareNullsAsValues;
 			var hasNull         = false;
 			var count           = 0;
 			var multipleClauses = false;
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -98,6 +98,12 @@ namespace LinqToDB.SqlProvider
 			return SqlOptimizer.ConvertElement(MappingSchema, element, OptimizationContext) as T;
 		}
 
+		private StringBuilder RemoveInlineComma()
+		{
+			StringBuilder.Length -= InlineComma.Length;
+			return StringBuilder;
+		}
+
 		#endregion
 
 		#region BuildSql
@@ -2127,11 +2133,12 @@ namespace LinqToDB.SqlProvider
 
 		void BuildInListValues(SqlPredicate.InList predicate, IEnumerable values)
 		{
-			var firstValue = true;
-			var len        = StringBuilder.Length;
-			var hasNull    = false;
-			var count      = 0;
-			var longList   = false;
+			var firstValue      = true;
+			var len             = StringBuilder.Length;
+			var checkNull       = Configuration.Linq.CheckNullInContains;
+			var hasNull         = false;
+			var count           = 0;
+			var multipleClauses = false;
 
 			SqlDataType? sqlDataType = null;
 
@@ -2139,29 +2146,30 @@ namespace LinqToDB.SqlProvider
 			{
 				if (count++ >= SqlProviderFlags.MaxInListValuesCount)
 				{
-					count    = 1;
-					longList = true;
+					count           = 1;
+					multipleClauses = true;
 
 					// start building next bucked
 					firstValue = true;
-					StringBuilder.Remove(StringBuilder.Length - 2, 2).Append(')');
-					if (predicate.IsNot)
-						StringBuilder.Append(" AND ");
-					else
-						StringBuilder.Append(" OR ");
+					RemoveInlineComma()
+						.Append(')')
+						.Append(predicate.IsNot ? " AND " : " OR ");
 				}
 
-				object? val = value;
-
-				if (val is ISqlExpression sqlExpr && sqlExpr.TryEvaluateExpression(OptimizationContext.Context, out var evaluated))
+				if (checkNull)
 				{
-					val = evaluated;
-				}
+					object? val = value;
 
-				if (val == null)
-				{
-					hasNull = true;
-					continue;
+					if (val is ISqlExpression sqlExpr && sqlExpr.TryEvaluateExpression(OptimizationContext.Context, out var evaluated))
+					{
+						val = evaluated;
+					}
+
+					if (val == null)
+					{
+						hasNull = true;
+						continue;
+					}
 				}
 
 				if (firstValue)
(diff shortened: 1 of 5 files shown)


jods4:
One more possible bug that I caught after the PR:
`BuildInListPredicate` starts with:
```csharp
if (p.Values == null || p.Values.Count == 0)
{
	BuildPredicate(new SqlPredicate.Expr(new SqlValue(false)));
}
```
Shouldn't that be `new SqlValue(!IsNot)`?
`x in ()` is false, but shouldn't `x not in ()` be true?

jods4:
Here's a summary of the SQL generated by the flags in each possible situation:

### CompareNullsAsValues = true
Same behavior as master, default behavior.
`CheckNullsInContains` is ignored when `CompareNullsAsValues` is true.

| C# | SQL 
|---|---
| `x.In(1, 2)` | `x IN (1, 2)`
| `x.In(1, null)` | `x IN (1) OR x IS NULL`
| `x.NotIn(1, 2)` | `x NOT IN (1, 2) OR x IS NULL`
| `x.NotIn(1, null)` | `x NOT IN (1) AND x IS NOT NULL`

### CompareNullsAsValues = false[, CheckNullsInContains = true]
Same behavior as master, alternative config.
This combination isn't very intuitive, it exists because of back compat.

| C# | SQL 
|---|---
| `x.In(1, 2)` | `x IN (1, 2)`
| `x.In(1, null)` | `x IN (1) OR x IS NULL`
| `x.NotIn(1, 2)` | `x NOT IN (1, 2)`
| `x.NotIn(1, null)` | `x NOT IN (1) AND x IS NOT NULL`

### CompareNullsAsValues = false, CheckNullsInContains = false
New in this PR.
Pure SQL semantics without changes.

| C# | SQL 
|---|---
| `x.In(1, 2)` | `x IN (1, 2)`
| `x.In(1, null)` | `x IN (1, null)`
| `x.NotIn(1, 2)` | `x NOT IN (1, 2)`
| `x.NotIn(1, null)` | `x NOT IN (1, null)`

MaceWindu:
Do we really need new option? Isn't it better to generate sql based only on `CompareNullsAsValues` and database support for NULL in IN?

jods4:
@MaceWindu I agree with you it would be better to base SQL generation on `CompareNullsAsValues` only.

The problem with this is backward compatibility.

You can see all cases in the tables I put above.
In v3.3, you must choose between one of the first 2. 

What I would like is option 3, which is straight SQL without additional clauses.
If we were designing from scratch, I think it could simply replace option 2.

Changing this today would change the SQL codegen of existing users and can break their application if they depend on this.

MaceWindu:
@jods4 something to change in v4 as part of global options removal work?

jods4:
@MaceWindu yes, that sounds like the right time to revamp this. 
Just be careful how much you're willing to break in v4. 
A major version is semantically the right time to do breaking changes, but users will still need to upgrade and the more was changed the harder it'll be for them (although I'm ok with that move).

I still believe this is worthy of being merged in v3.4, even if you remove the new option in v4:
- Users who want SQL semantics can have it right now;
- This PR adds a bunch of tests that I had (have...) a hard time making pass. The semantics of those variations are surprisingly tricky so it's good to have them covered by tests, if only to have a reference for the expected behavior;
- In fact, this PR contains a bug fix in addition to the new option;
- This provides an easier migration path to v4. Users that turn `CheckNullsInContains` off in addition to `CompareNullsAsValues` will be forward compatible with v4.

Source/LinqToDB/Tools/SqlExtensions.cs
```diff
@@ -59,7 +80,9 @@ public static bool In<T>(this T value, T cmp1, T cmp2)
 		[ExpressionMethod(nameof(InImpl5))]
 		public static bool In<T>(this T value, T cmp1, T cmp2, T cmp3)
 		{
-			return object.Equals(value, cmp1) || object.Equals(value, cmp2) || object.Equals(value, cmp3);
+			var comparer = EqualityComparer<T>.Default;
```
MaceWindu, line 83: BTW, I was thinking to do the same in memory PR, but isn't it breaking change if T implements custom `Equals` logic?

MaceWindu: Not like I'm against it still

viceroypenguin: I believe any objects that this would be a breaking change are breaking the equals spec regardless, so technically out of bounds for us to _have_ to worry about. Whether we choose to or not... This, however, it a much more performant method, because it doesn't involve 6 separate boxes for value types, so my vote is firmly for this.

MaceWindu:
@jods4 I'm struggling to understand rationale behind it.

If I understand what's going on here this option allow user to disable `null` support for `IN` in `CompareNullsAsValues = false` mode (default mode)? Or in other words enables strict `IN` semantics without extra null checks support. Right?

MaceWindu (changes requested):
one case doesn't work:
`withNullContains: true, withNullCompares: false` generates 
`([s].[Int] NOT IN (-1, 2) OR [s].[Int] IS NULL)`
but should
`[s].[Int] NOT IN (-1, 2)`

Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -2127,41 +2133,43 @@ void BuildInListPredicate(ISqlPredicate predicate)
...
 		void BuildInListValues(SqlPredicate.InList predicate, IEnumerable values)
 		{
-			var firstValue = true;
-			var len        = StringBuilder.Length;
-			var hasNull    = false;
-			var count      = 0;
-			var longList   = false;
+			var firstValue      = true;
+			var len             = StringBuilder.Length;
+			var checkNull       = Configuration.Linq.CheckNullsInContains || Configuration.Linq.CompareNullsAsValues;
```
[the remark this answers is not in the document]
MaceWindu: @sdanyliv

jods4:
> @jods4 I'm struggling to understand rationale behind it.
> 
> If I understand what's going on here this option allow user to disable `null` support for `IN` in `CompareNullsAsValues = false` mode (default mode)? Or in other words enables strict `IN` semantics without extra null checks support. Right?

Yes, main driver here is to have a mode where the generated SQL is predictable and matches exactly the C# code.

Sometimes this can also be handy because putting `NULL` inside a `IN` expression can be a way to disable a condition without actually removing it (as `in (null)` would never be true).

Source/LinqToDB/Tools/SqlExtensions.cs
MaceWindu: Hmm, you are right. I thought `EqualityComparer<T>.Default` will not call `Equals`

jods4:
> one case doesn't work:
> `withNullContains: true, withNullCompares: false` generates
> `([s].[Int] NOT IN (-1, 2) OR [s].[Int] IS NULL)`
> but should
> `[s].[Int] NOT IN (-1, 2)`

I assume you did `s.Int.NotIn(-1, 2)` ? 
That doesn't match my comment above, so it's probably a bug. I'll look into it and check why it wasn't caught by a test.
https://github.com/linq2db/linq2db/pull/2949#issuecomment-819874784

jods4:
> > one case doesn't work:
> > `withNullContains: true, withNullCompares: false` generates
> > `([s].[Int] NOT IN (-1, 2) OR [s].[Int] IS NULL)`
> > but should
> > `[s].[Int] NOT IN (-1, 2)`
> 
> I assume you did `s.Int.NotIn(-1, 2)` ?
> That doesn't match my comment above, so it's probably a bug. I'll look into it and check why it wasn't caught by a test.
> [#2949 (comment)](https://github.com/linq2db/linq2db/pull/2949#issuecomment-819874784)

@MaceWindu I believe this is because I never managed to make my tests work properly.
See this comment I made in Slack:
https://linq2db.slack.com/archives/C9JGUKWD8/p1618436002020600

Basically, everything seems right when ran individually, but I'm seeing codegen from different options when all tests run in CI. 
In this case, code behavior is the one from `predicate.WithNull == true` although `CompareNullsAsValues` is supposed to be false!?!

MaceWindu:
still needed or we can close and drop branch?

jods4:
@MaceWindu This is superseded by #3514, I'm gonna close it.
Can you please not drop the branch before #3514 is all good and merged? Just in case I want to look at what I did in this version.

PR #2952, opened 2021-04-13
Fix for #2947. Added support for case (in)sensitive search.

### LINKED ISSUE #2947: string.StartsWith/EndsWith do not respect StringComparison arg.
This bug hits atleast Firebird and Sqlite providers.

```c#
		[Test, Combinatorial]
		public void SearchStringTest(
			[IncludeDataSources(TestProvName.AllSQLite, TestProvName.AllFirebird)] string context, 
			[Values(nameof(string.StartsWith), nameof(string.EndsWith))] string methodName,
			[Values]bool ignoreCase)
		{
			var stringComparison = ignoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
			Expression<Func<Model.Person, bool>> GetPredicate(string param) => methodName switch
			{
				nameof(string.StartsWith) => (Model.Person p) => p.LastName.StartsWith(param, stringComparison),
				nameof(string.EndsWith) => (Model.Person p) => p.LastName.EndsWith(param, stringComparison),				
				_ => throw new ArgumentException("Unknown method name " + methodName)
			};

			{
				Assert.IsTrue(db.Person.Any(GetPredicate("Pupkin")));
				Assert.AreEqual(ignoreCase, db.Person.Any(GetPredicate("pupkin")));
			}			
		}
```

Also there is FB specific issue in FirebirdSqlOptimizer.ConvertSearchStringPredicate. "starting with" is case-sensitive, but linq2db use it in "ignorecase" branch

### Environment details
linq2db version: 3.3
Database Server: Firebird

sdanyliv:
Fixes #2947


Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1662,6 +1660,19 @@ ValueTypeExpression ReplaceParameter(IDictionary<Expression,Expression> expressi
 
 		ISqlPredicate ConvertPredicate(IBuildContext? context, Expression expression)
 		{
+			static bool IsIgnoreCase(MethodCallExpression mc)
```
jack128, line 1663: I tried to fix the issue this way, but run into some cache problem. For example 
```
			StringComparison stringComparison;
			Expression<Func<Model.Person, bool>> GetPredicate(string param) => (Model.Person p) => p.LastName.StartsWith(param, stringComparison);
			{
				stringComparison = StringComparison.CurrentCultureIgnoreCase;
				db.Person.Any(GetPredicate("pupkin"));
				stringComparison = StringComparison.CurrentCulture;
				db.Person.Any(GetPredicate("pupkin")); // 2
			}
```
 linq2db do not call CreateStringPredicate in line 2 (get it from cache???) and search is case-insensitive

sdanyliv: Good point, yes probably cache may leverage this. attempt. Will take care in this. part.

MaceWindu (changes requested):

MaceWindu:
For SQL Server 2019 we use case-sensitive database so tests fail. Which rise interesting issue.
I think we shouldn't depend on default match behavior if user specified comparison mode explicitly.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
sdanyliv: Resolved.

MaceWindu:
Clarification on what's going on here. This PR updates SQL, generated for Contains/StartsWith/EndsWith methods with `StringComparison` parameter:

- no changes: for calls without parameter we compare as is: using server comparison logic, db or column collation (which one is set)
- when `StringComparison` parameter set we try to generate explicit case-sensitive or case-insensitive comparison (depends on parameter value)

viceroypenguin (approved):

PR #2954, opened 2021-04-15
Add Sql.Default<T>()

jods4:
This PR adds an extension to use the SQL `DEFAULT` keyword.
This keyword can be used as an insert value, for example:
```csharp
db.T.Insert(() => new T 
{
  Id = Sql.Default<int>(),
  Name = "Joe",
});
```
Generates:
```sql
insert into T (id, name)
values (default, "Joe")
```
Default is useful when you need to explicitly insert a non-null value but the DB decides what the value actually is, e.g.:
- Identities, sequences;
- Default constraints;
- Triggers.

### DIFF
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -37,7 +37,7 @@ namespace LinqToDB
 		/// Generates 'DEFAULT' keyword, usable in inserts.
 		/// </summary>
 		[Expression("DEFAULT", ServerSideOnly = true)]
-		public static T Default<T>() => throw new NotImplementedException();
+		public static T Default<T>() => throw new LinqException($"{nameof(Default<int>)} is only server-side method.");
 
 		/// <summary>
 		/// Enforces generating SQL even if an expression can be calculated locally.
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -37,7 +37,7 @@ namespace LinqToDB
 		/// Generates 'DEFAULT' keyword, usable in inserts.
 		/// </summary>
 		[Expression("DEFAULT", ServerSideOnly = true)]
-		public static T Default<T>() => throw new LinqException($"{nameof(Default<int>)} is only server-side method.");
+		public static T Default<T>() => throw new LinqException($"Default is only server-side method.");
 
 		/// <summary>
 		/// Enforces generating SQL even if an expression can be calculated locally.
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -34,6 +34,12 @@ namespace LinqToDB
 		}
 
 		/// <summary>
+		/// Generates 'DEFAULT' keyword, usable in inserts.
+		/// </summary>
+		[Expression("DEFAULT", ServerSideOnly = true)]
+		public static T Default<T>() => throw new NotImplementedException();
+
+		/// <summary>
 		/// Enforces generating SQL even if an expression can be calculated locally.
 		/// </summary>
 		/// <typeparam name="T"></typeparam>
--- a/Tests/Linq/Update/InsertTests.cs
+++ b/Tests/Linq/Update/InsertTests.cs
@@ -1609,11 +1609,11 @@ namespace Tests.xUpdate
 		// see https://github.com/linq2db/linq2db/pull/2954#issuecomment-821798021
 		[Test]
 		public void InsertDefault([DataSources(
-			TestProvName.AllMySql, 
-			TestProvName.AllPostgreSQL, 
-			TestProvName.AllOracle, 
-			TestProvName.Sybase, 
-			TestProvName.AllSqlServer)] string context)
+			TestProvName.AllAccess, 
+			TestProvName.AllFirebird, 
+			TestProvName.AllInformix, 
+			TestProvName.AllSapHana, 
+			TestProvName.AllSQLite)] string context)
 		{
 			using var db = GetDataContext(context);
 			try
--- a/Tests/Linq/Update/InsertTests.cs
+++ b/Tests/Linq/Update/InsertTests.cs
@@ -1606,6 +1606,26 @@ namespace Tests.xUpdate
 		}
 
 		[Test]
+		public void InsertDefault([DataSources] string context)
+		{
+			using var db = GetDataContext(context);
+			try
+			{
+				db.Person.Insert(() => new Person
+				{
+					FirstName = "InsertDefault",
+					MiddleName = Sql.Default<string>(),
+					LastName = "InsertDefault",
+					Gender = Gender.Male,
+				});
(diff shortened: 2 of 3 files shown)


sdanyliv:
Maybe run tests and see which providers will fail?

jods4:
I love your optimism 😄

jods4:
@sdanyliv looks like you were right, not all providers support this old, standard feature.
Tests aren't complete yet but it seems at least Access and SQLite don't support it, possibly others we'll see when tests are complete.

What's LinqToDb approach when a standard SQL feature isn't supported everywhere?
Do we keep it in and just specify which providers to run the tests for?

sdanyliv:
Not ready to answer this question. Let's see all failing results.

jods4:
Results are in, most RDMB support `default` except:
- SQLite
- Access
- Firebird
- Informix
- SAP Hana

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -33,6 +33,12 @@ public static partial class Sql
 			throw new LinqException("'AllColumns' is only server-side method.");
 		}
 
+		/// <summary>
+		/// Generates 'DEFAULT' keyword, usable in inserts.
+		/// </summary>
+		[Expression("DEFAULT", ServerSideOnly = true)]
+		public static T Default<T>() => throw new NotImplementedException();
```
MaceWindu, line 40:
```suggestion
		public static T Default<T>() => throw new LinqException($"{nameof(Default<int>)} is only server-side method.");
```

MaceWindu: Like in other db-only extensions

Tests/Linq/Update/InsertTests.cs
```diff
@@ -1605,6 +1605,26 @@ public void Insert16([DataSources] string context)
...
+		{
+			using var db = GetDataContext(context);
+			try
+			{
+				db.Person.Insert(() => new Person
+				{
+					FirstName = "InsertDefault",
+					MiddleName = Sql.Default<string>(),
+					LastName = "InsertDefault",
+					Gender = Gender.Male,
```
MaceWindu, line 1619: alignment missing

MaceWindu:
Can confirm test results:
- Access doesn't have `DEFAULT` placeholder: https://docs.microsoft.com/en-us/office/client-developer/access/desktop-database-reference/insert-into-statement-microsoft-access-sql
- SQLite has only `DEFAULT VALUES` clause which replaces `VALUES` : https://sqlite.org/lang_insert.html
- Firebird added support for `DEFAULT` in v4: https://tracker.firebirdsql.org/browse/CORE-5449, which is not yet added to CI (#2693)
- Informix doesn't have `DEFAULT`, but have another interesting placeholder `USER` for current user id https://www.ibm.com/docs/en/informix-servers/14.10/14.10?topic=statement-values-clause#ids_sqs_0132
- SAP HANA has no `DEFAULT` too https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/2.0.03/en-US/20f7f70975191014a76da70c9181720e.html

Tests/Linq/Update/InsertTests.cs
```diff
@@ -1605,6 +1605,26 @@ public void Insert16([DataSources] string context)
 			}
 		}
 
+		[Test]
+		public void InsertDefault([DataSources] string context)
```
MaceWindu, line 1609: add
```suggestion
		public void InsertDefault([DataSources(TestProvName.AllAccess, TestProvName.AllSQLite, TestProvName.AllFirebird, TestProvName.AllInformix, TestProvName.AllSapHana)] string context)
```

MaceWindu (changes requested):

MaceWindu:
@jods4, as I plan to add firebird4 test provider in #2959 it makes sense to wait for it completed and merged

Tests/Linq/Update/InsertTests.cs
jods4: I think you got it backwards, these are all the providers that do _not_ support this feature.

MaceWindu: `DataSources` excludes providers, passed as parameter

Tests/Linq/Update/InsertTests.cs
jods4: Done in 9cdb39f

Tests/Linq/Update/InsertTests.cs
jods4: I put the providers that supports DEFAULT in 743c29e

jods4:
@MaceWIndu I've done the requested changes, this PR is all ready.
I leave it to you to do the Firebird 4 changes and merge.

Tests/Linq/Update/InsertTests.cs
jods4: > DataSources excludes providers, passed as parameter

Ok from the naming I expected the opposite.
It's actually better that way (e.g. in case you add a new provider).

I reversed the tested providers in 01d27a1.

Tests/Linq/Update/InsertTests.cs
```diff
@@ -1605,6 +1605,33 @@ public void Insert16([DataSources] string context)
 			}
 		}
 
+		// Access, SQLite, Firebird before v4, Informix and SAP Hana do not support DEFAULT in inserted values, 
+		// see https://github.com/linq2db/linq2db/pull/2954#issuecomment-821798021
+		[Test]
+		public void InsertDefault([DataSources(
+			TestProvName.AllAccess, 
+			TestProvName.AllFirebird,
```
MaceWindu, line 1613:
```suggestion
			ProviderName.Firebird, TestProvName.Firebird3, 
```

MaceWindu: but you need to merge master with FB4 support for it to work

jods4: @MaceWindu this PR should be all done with respect to `master`. 
I leave it to you to merge and test it into your FB4 branch. (or feel free to merge the other way round).

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -33,6 +33,12 @@ public static partial class Sql
 			throw new LinqException("'AllColumns' is only server-side method.");
 		}
 
+		/// <summary>
+		/// Generates 'DEFAULT' keyword, usable in inserts.
+		/// </summary>
+		[Expression("DEFAULT", ServerSideOnly = true)]
+		public static T Default<T>() => throw new LinqException($"Default is only server-side method.");
```
MaceWindu, line 40: not happy with it, but was forced to remove `nameof` here due to strange `CS8084` error on build ¯\_(ツ)_/¯

MaceWindu:
@jods4 , please don't forget to add `feature`/`issue` prefixes to branches

MaceWindu (approved):

viceroypenguin (approved):

PR #2956, opened 2021-04-15
Add missing Blob datatype mapping

jods4:
Just a forgotten mapping that prevents `OracleBlob` from working smoothly.

### DIFF
--- a/Tests/Linq/DataProvider/OracleTests.cs
+++ b/Tests/Linq/DataProvider/OracleTests.cs
@@ -3824,7 +3824,7 @@ CREATE TABLE ""TABLE_A""(
 			// Implicit OracleBlob support, no attribute
 			public OracleBlob? BinaryValue { get; set; }
 			// Explicit attribute with DataType = Blob
-			[Column("BINARYVALUE", DataType = DataType.Blob)]
+			[Column("BinaryValue", DataType = DataType.Blob)]
 			public OracleBlob? Blob { get; set; }
 		}
 		
--- a/Tests/Linq/DataProvider/OracleTests.cs
+++ b/Tests/Linq/DataProvider/OracleTests.cs
@@ -3845,7 +3845,7 @@ CREATE TABLE ""TABLE_A""(
 			var inserted = db
 				.GetTable<LinqDataTypesBlobs>()
 				.Where(x => x.ID.In(-10, -20))
-				.Select(x => Sql.Expr<int>("LENGTH(BINARYVALUE)"))
+				.Select(x => Sql.Expr<int>("LENGTH(\"BinaryValue\")"))
 				.ToList();
 
 			tx.Rollback();
--- a/Tests/Linq/DataProvider/OracleTests.cs
+++ b/Tests/Linq/DataProvider/OracleTests.cs
@@ -3817,7 +3817,7 @@ CREATE TABLE ""TABLE_A""(
 			}
 		}
 	
-		[Table("LINQDATATYPES", IsColumnAttributeRequired = false)]
+		[Table("LinqDataTypes", IsColumnAttributeRequired = false)]
 		class LinqDataTypesBlobs
 		{
 			public int ID { get; set; }
--- a/Tests/Linq/DataProvider/OracleTests.cs
+++ b/Tests/Linq/DataProvider/OracleTests.cs
@@ -3850,7 +3850,7 @@ CREATE TABLE ""TABLE_A""(
 
 			tx.Rollback();
 
-			inserted.Should().BeSameAs(new[] { 1, 1 });
+			inserted.Should().Equal(1, 1);
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


viceroypenguin (approved):

sdanyliv:
Bad that here we have no regression tests.

jods4:
I don't know what test to write for this change...

sdanyliv:
Use existing table `LinqDataTypes`, map new class to this table, at least blob column. And mark this column as `DataType.Blob`.

sdanyliv (approved):

jods4:
I attempted to write a test, let's see if it works.

For context, neither of the 2 statements in the test worked before.
`OracleBlob` was converted to a parameter of type `RAW` and that failed.
Explicit `DataType.Blob` was still converted to a parameter of type `RAW` and failed.

We worked around this in 3.3 by using `DataType.VarBinary`, which feels weird in Oracle but works.

/azp run test-all

jods4:
@sdanyliv Tests are OK except 1 Mac and 2 Linux tests where it says the `LinqDataTypes` table doesn't exist??

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -3815,5 +3816,41 @@ public void TestDateTimeNAddTimeSpan([IncludeDataSources(true, TestProvName.AllO
 					).Select(x => x.ID).ToArray();
 			}
 		}
+	
+		[Table("LINQDATATYPES", IsColumnAttributeRequired = false)]
```
MaceWindu, line 3820:
```suggestion
		[Table("LinqDataTypes", IsColumnAttributeRequired = false)]
```

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -3815,5 +3816,41 @@ public void TestDateTimeNAddTimeSpan([IncludeDataSources(true, TestProvName.AllO
...
+		class LinqDataTypesBlobs
+		{
+			public int ID { get; set; }
+			// Implicit OracleBlob support, no attribute
+			public OracleBlob? BinaryValue { get; set; }
+			// Explicit attribute with DataType = Blob
+			[Column("BINARYVALUE", DataType = DataType.Blob)]
```
MaceWindu, line 3827:
```suggestion
			[Column("BinaryValue", DataType = DataType.Blob)]
```

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -3815,5 +3816,41 @@ public void TestDateTimeNAddTimeSpan([IncludeDataSources(true, TestProvName.AllO
...
+			using var blob = new OracleBlob((OracleConnection)dc.Connection);
+			blob.WriteByte(1);
+
+			db.GetTable<LinqDataTypesBlobs>().Insert(() => new LinqDataTypesBlobs { ID = -10, BinaryValue = blob });
+			db.GetTable<LinqDataTypesBlobs>().Insert(() => new LinqDataTypesBlobs { ID = -20, Blob = blob });
+
+			var inserted = db
+				.GetTable<LinqDataTypesBlobs>()
+				.Where(x => x.ID.In(-10, -20))
+				.Select(x => Sql.Expr<int>("LENGTH(BINARYVALUE)"))
```
MaceWindu, line 3848:
```suggestion
				.Select(x => Sql.Expr<int>("LENGTH(\"BinaryValue\")"))
```

MaceWindu (commented):
Fixed table/column names. We use following definition (`oracle.sql`):
```sql
CREATE TABLE "LinqDataTypes"
(
	ID               int,
	"MoneyValue"     decimal(10,4),
	"DateTimeValue"  timestamp,
	"DateTimeValue2" timestamp,
	"BoolValue"      smallint,
	"GuidValue"      raw(16),
	"BinaryValue"    blob         NULL,
	"SmallIntValue"  smallint,
	"IntValue"       int          NULL,
	"BigIntValue"    number(20,0) NULL,
	"StringValue"    VARCHAR2(50) NULL
)
```

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -3815,5 +3816,41 @@ public void TestDateTimeNAddTimeSpan([IncludeDataSources(true, TestProvName.AllO
...
+
+			var inserted = db
+				.GetTable<LinqDataTypesBlobs>()
+				.Where(x => x.ID.In(-10, -20))
+				.Select(x => Sql.Expr<int>("LENGTH(\"BinaryValue\")"))
+				.ToList();
+
+			tx.Rollback();
+
+			inserted.Should().BeSameAs(new[] { 1, 1 });
```
MaceWindu, line 3853: Assert fails, looks like arrays equality not supported by `BeSameAs`

MaceWindu (changes requested):

Tests/Linq/DataProvider/OracleTests.cs
jods4: Ok, I tried `.Equal(1, 1)` according to FluentAssertions docs it should work.

jods4:
@MaceWindu all good now, except Mac tests that timed-out and cancelled... as usual.

MaceWindu (approved):

PR #2957, opened 2021-04-15
Use Oracle date/timestamp literals instead of functions

jods4:
Oracle supports ANSI SQL dates literal, as well as timestamp literals, which are shorter and more readable then calling parsing functions:
- `DATE '2020-01-01'`
- `TIMESTAMP '2020-01-01 10:30:00.1234567 +00:00'`

Sadly, there's no such literal for dates with times, so this has to stay as `TO_DATE('2020-01-01 10:00:00', 'YYYY-MON-DD HH24:MIN-SS')`

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/OracleMappingSchema.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/OracleMappingSchema.cs
@@ -110,22 +110,22 @@ namespace LinqToDB.DataProvider.Oracle
 			switch (dataType.Type.DataType)
 			{
 				case DataType.Date:
-					format = "TO_DATE('{0:yyyy-MM-dd}', 'YYYY-MM-DD')";
+					format = "DATE '{0:yyyy-MM-dd}'";
 					break;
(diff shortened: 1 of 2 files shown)


jods4:
Do you know if this is covered by regression tests?

sdanyliv:
It should be. But I think you may broke lower Oracle versions.

jods4:
How low do we go? AFAIK it's at least supported in Oracle 10+

jods4:
There are regression tests for sure... they failed! 😄 

They actually failed because they assert the presence of a specific literal expansion, which I've changed, so I need to update the tests accordingly. 
Will do later.

jods4:
Interestingly I found out that the literal syntax was already used in some tests, see `OracleTests.TestDateTimeOffset`.

jods4:
Everything is green, only failure is Mac tests that were cancelled

viceroypenguin (approved):

MaceWindu (approved):
:+1:

PR #2967, opened 2021-04-22
Correct DeleteWithOutputAsync() function name

viceroypenguin:


sdanyliv:
Well, looks like we do not have regression test for this function ;)

viceroypenguin:
pish posh. :D

sdanyliv:
And missed again. Tested on Sync version.

viceroypenguin:
Yes, all of the tests were only done w/ the sync version. Only added sync to match the other sync tests. I can double all of the tests for the async versions if you want? Seems somewhat extraneous, but...

sdanyliv:
Believe me it worth efforts. We can accidentally change something and async versions may fail.

viceroypenguin:
PS: Blame @MaceWindu  - he's the one who wrote the InsertWithOutput tests that I copied from originally... 😉

sdanyliv:
It's me ;) Really messed up with count of new methods.

viceroypenguin:
There. Punk... :)

sdanyliv (approved):

MaceWindu (changes requested):
tests fail, probably something wrong with asserts

MaceWindu (approved):

Issue #2696, opened 2020-12-15
Memory leak in query cache and LoadWith


andrewvk:
When you call LoadWith with caching enabled, the memory consumption is constantly increasing.
![image](https://user-images.githubusercontent.com/803486/102187802-883b5000-3ec5-11eb-96e1-e68f56318900.png)
![image](https://user-images.githubusercontent.com/803486/102187831-92f5e500-3ec5-11eb-96ce-2649a33f8893.png)
It seems that the same query is pushed into the cache on every call
![image](https://user-images.githubusercontent.com/803486/102190539-38f71e80-3ec9-11eb-9110-753749f626fe.png)

### Environment details
linq2db version: *2.9.8*
Database Server: *PostgreSQL*
Operating system: *Debian 10*
.NET Framework: *Core 3.1*

sdanyliv:
Maybe some query sample?

MaceWindu:
@sdanyliv #2670 removed dictionary from CompiledTable, so it could be fixed already.

andrewvk:
Yet another screenshot
![image](https://user-images.githubusercontent.com/803486/102191063-f41fb780-3ec9-11eb-96ae-23f488a6f274.png)

sdanyliv:
I do not need that nice screenshots, but query which fails.

MaceWindu:
@andrewvk could you inspect CompiledTable dictionary cache keys - could you see why there are so many entries (why cache entries not reused?)

andrewvk:
Probably a problematic request.
![image](https://user-images.githubusercontent.com/803486/102191974-3269a680-3ecb-11eb-8d77-dfe7764c43b0.png)

sdanyliv:
@andrewvk, you have not prepared reproducible steps. 
Your stack trace shows that you have used `CompiledQuery` but sample screenshot not related to that.
I have prepared small tests on my side with both variants. All works as expected.

If you want fix to be ready for upcoming release planned this `Thursday`, be more informative.

Anyway, probably you have created new `MappingSchema` for every `DataConnection` and missed query cache.

sdanyliv:
@andrewvk, any news with that? If guys don't know English, they can create post in RSDN.
Anyway, I think that they have problem with `MappingSchema` which should be cached somehow.

MaceWindu:
We addressed some memory issues here #2971

[...]

Issue #2556, opened 2020-10-17
Update performance regression from 2.9.8


pebezo:
Hi,

We've noticed a significant performance regression between 2.9.8 and 3.1.4 when AWS started shutting down our docker instance due to OOM errors. To troubleshoot we downgraded to 2.9.8 which fixed the issue.

To better help track down the source I made a small app that inserts and then updates a bunch of rows in a test table (using MySQL). The inserts are "smooth" and fast, they look like this in dotMemory:

![image](https://user-images.githubusercontent.com/1287097/96350620-b4348380-1084-11eb-9e55-2212a37a12a7.png)

However, updates look like this:

![image](https://user-images.githubusercontent.com/1287097/96350664-d4fcd900-1084-11eb-810f-bc9b1cbe3ae2.png)

On the left you can see the transition from inserts to updates as the memory allocation increases. 

Interestingly, version 2.9.8 also has a similar jagged line, but for whatever reason it is not causing OOM errors.

Also, between version 2.9.8 and 3.1.4 (same with 3.1.5) the inserts take pretty much the same time, but the update has gone from 1min 33s to 2min (almost 30s slower).

Before I dig in any more, is this a known problem that's being worked on?

Thanks,

### Environment details
linq2db version: *2.9.8* and *3.1.4* / *3.1.5*
Database Server: *MySQL* in Docker
Database Provider: *MySQL*
Operating system: *Windows 10*
.NET Framework: *Core 3.1*

MaceWindu:
That's unexpected, as we did a lot of optimizations for v3. Wether we regressed in 3.1 or we didn't noticed something.
To look into it it would be nice to have information about memory allocations and/or small PoC program that does operations, similar to one you do in real app and which demonstrate same memory behavior as real app in v2 and v3

pebezo:
OK... I'll work on coming up with a repro/PoC. 

With the small console app I noticed the slowdown, but the memory usage/patter was similar between 2.9.8 and 3.1.5. In the production code the RAM usage was going up and up (from 30-40MB to several 100s of MB) eventually getting killed by AWS.

MaceWindu:
Any updates?

pblachut:
Hi, I noticed similar issue. Was using version 2.9.8 and after upgrade to 3.1.5 performance was very badly.

![linq2bissue](https://user-images.githubusercontent.com/12517339/98080018-1b1fae00-1e75-11eb-8319-be4181cec472.png)

[...]

PR #2971, opened 2021-04-26
Reduce memory footprint

MaceWindu:
Related issues:
- #2556
- #2696
- https://rsdn.org/forum/prj.rfd/7993035

Already managed to reduce memory consumption greatly, but still need at least to refactor visitors to final form, so disregard current look of them as visitors refactoring still in progress.

Main changes:
- Expression/AST visitors refactoring to use context for actions instead of closures, reduce allocations in visitors
- `MappingSchema`, `MetadataReaders`: removed a lot of allocations around linq/closures use and arrays instantiations
- `NotIn`/`In` use for enums internally: were causing masive memory consumption due to boxing, use of them removed for enums
- use of static arrays instead of `params` parameters in linq method builders
- other small fixes here and there
- fixed couple of issues in visitors/builders, discovered during refactoring

Remaining tasks:
- [ ] try to lower stack pressure to allow our tests pass on macos with small stack size
- [ ] finish visitors/walkers refactoring
  - [x] improve API/allocations around context-less visitors and use static instances of visitor
  - [x] for visitors with context review uses to see if we can make them context-less or have smaller context
  - visitors and visitor-like API TODO:
    - [x] Expressions: Find
    - [x] Expressions: Path
    - [x] Expressions: Transform
    - [x] Expressions: TransformInfo
    - [x] Expressions: VisitAction
    - [x] Expressions: VisitFunction
    - [x] Expressions: EqualsTo
    - [x] QueryVisitor
    - [x] AST: ConvertVisitor
    - [x] AST: Clone
    - [ ] WalkQueries ?
- [x] add benchmark for update

### Preview of current results

before
image
latest
image


jods4:
That's great! 👍 
I noticed when working on various PR that the codebase makes use of tons of short-lived objects (small arrays, strings, string builders, boxed values, delegates with captures).
Sadly that's the kind of thing that is a thousands paper-cuts. Each one individually doesn't change much, but together they add up.

If you're refactoring the visitors, I noticed there are several places that perform linear searches that might be more efficient if they were binary searches or hash tables. It's not memory-related but could improve perf.
In particular, the sequence builder list is currently a 62 items linear search:
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs#L26

The current design is based on calling `CanBuild` predicates, which cannot be optimized, but most cases could be sorted out in a first pass based on a type hashtable.
Consider for example the new `MultiInsertBuilder`. Because all multi-insert methods were put in their own `MultiInsertExtensions` class, it can quickly take responsability of all code calling methods of that type.

Another design, that is not a breaking change as the previous one, could be to decorate extension methods with a `ExpressionBuilderAttribute` indicating the right type. So only stuff that can't be decorated would be searched linearly.

sdanyliv:
@viceroypenguin, well I remember 6 years ago code was exactly the same as @MaceWindu did. But @igor-tkachev replaced it with this ugly cast variant, exactly for reducing stack size. Probably compiler become better but .NET Framework JIT is still old.

viceroypenguin:
Actually, even the JIT is improved, because .net (both fx and core) is 100% on the RyuJIT instead of the original JIT. Lots of stuff has been improved since then too. Tail-call-optimization is one I specifically remember was added to the JIT along the way. Plenty of others as well.

viceroypenguin:
That said, I may be wrong - this one might be complicated enough to get not get properly optimized. But there's enough to suggest research rather than a blanket reject the cast variable.

MaceWindu:
image

jogibear9988:
do you know @ayende. He had written many interesting blog posts about c# perf. 

see https://ayende.com/blog/tags/performance
for example:
https://ayende.com/blog/177409/goto-based-optimizations
https://ayende.com/blog/177377/fast-dictionary-and-struct-generic-arguments

MaceWindu:
Comparison with v2/v3 releases (using release build). I would say having it better/on-par with 2.9.8 is nice taking into account all new features we added since 2.9.8 release.

Cuurent PR:
image
3.3.0:
image
2.9.8:
image

MaceWindu:
image

MaceWindu:
Guys, thanks for all suggestions, but this PR is only targeting easy memory consumption issues, especially introduced by v3 features. They are more than enough to solve regression in this area since v2. Any other perf. issues could be rised separately

Source/LinqToDB/Expressions/Extensions.cs
```diff
@@ -956,42 +182,13 @@ public static Expression GetBody(this LambdaExpression lambda, Expression exprTo
...
-			{
-				var e = Transform(item, func);
-				list.Add((T)e);
-				modified = modified || e != item;
-			}
-
-			return modified ? list : source;
+			return Transform(
+				lambda.Body,
+				new { lambda, exprToReplaceParameter1, exprToReplaceParameter2, exprToReplaceParameter3 },
```
sdanyliv, line 187: Maybe better to pass Parameters instead. of lambda?

Source/LinqToDB/Expressions/Extensions.cs
```diff
@@ -956,42 +182,13 @@ public static Expression GetBody(this LambdaExpression lambda, Expression exprTo
...
+				static (context, e) =>
+					e == context.lambda.Parameters[0] ? context.exprToReplaceParameter1 :
```
sdanyliv, line 189: Quick shortcut here is to check that `e.NodeType == EpressionType.Parameter`

Source/LinqToDB/Common/Converter.cs
```diff
@@ -140,12 +152,14 @@ static bool ToBoolean(char ch)
...
 				var ex = Expression.Lambda<Func<object,object>>(
 					Expression.Convert(
-						b.Transform(e =>
-							e == ps[0] ?
-								Expression.Convert(p, e.Type) :
-							IsDefaultValuePlaceHolder(e) ?
-								new DefaultValueExpression(mappingSchema, e.Type) :
-								e),
+						b.Transform(
+							new { mappingSchema, ps, p },
```
viceroypenguin, line 156: I know we've discussed `System.ValueTuple` before, but is it possibly worth attempting to resolve the issues with it so we can reduce these allocations? Alternatively, writing our own simplified `internal struct ValueTuple` for use at each of these `.Transform()` sites. Each of these `new { }` is another allocation that could be done as a struct carried on the stack instead.

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -51,7 +50,10 @@ public override bool IsParameterDependedElement(IQueryElement element)
 
 					// Oracle saves empty string as null to database, so we need predicate modification before sending query
 					//
-					if (expr.Operator.In(SqlPredicate.Operator.Equal, SqlPredicate.Operator.NotEqual, SqlPredicate.Operator.GreaterOrEqual, SqlPredicate.Operator.LessOrEqual) && expr.WithNull == true)
+					if ((expr.Operator == SqlPredicate.Operator.Equal          ||
+						 expr.Operator == SqlPredicate.Operator.NotEqual       ||
+						 expr.Operator == SqlPredicate.Operator.GreaterOrEqual ||
+						 expr.Operator == SqlPredicate.Operator.LessOrEqual) && expr.WithNull == true)
```
viceroypenguin, line 56: Make the additional condition outside the `()` more obvious by putting it on the next line.
```suggestion
					if ((expr.Operator == SqlPredicate.Operator.Equal          ||
						 expr.Operator == SqlPredicate.Operator.NotEqual       ||
						 expr.Operator == SqlPredicate.Operator.GreaterOrEqual ||
						 expr.Operator == SqlPredicate.Operator.LessOrEqual) && 
					    expr.WithNull == true)
```

Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
```diff
@@ -14,7 +14,8 @@ namespace LinqToDB.Linq.Builder
 
 	class AggregationBuilder : MethodCallBuilder
 	{
-		public static string[] MethodNames = { "Average", "Min", "Max", "Sum" };
+		public  static readonly string[] MethodNames      = { "Average"     , "Min"     , "Max"     , "Sum"      };
```
viceroypenguin, line 17: Should be `IReadOnlyList<string>` for protection. `string[]` is still technically mutable outside of class control, even if `AggregationBuilder` is internal, so `IReadOnlyList<string>` protects sanity

to11mtm: This sounds like a good idea as long as the enumeration codegen doesn't somehow get tripped up (It shouldn't tho)

Source/LinqToDB/Common/Converter.cs
MaceWindu: I will keep it in mind (probably it makes sense to add our own set of ValueTupe-like types for this), but right now I don't see at as something that will help a lot. More effective will be reduce/optimize use of visitors itself.

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
MaceWindu: I'll exchange conditions instead as it looks more performant

Source/LinqToDB/DataProvider/SqlServer/SqlServerMappingSchema.cs
```diff
@@ -142,16 +143,22 @@ static void ConvertCharToSql(StringBuilder stringBuilder, SqlDataType sqlDataTyp
...
-			DataTools.ConvertCharToSql(stringBuilder, start, AppendConversion, value);
+			DataTools.ConvertCharToSql(stringBuilder, start, AppendConversionAction, value);
 		}
 
 		internal static void ConvertDateTimeToSql(StringBuilder stringBuilder, SqlDataType? dt, DateTime value)
 		{
 			var format =
 				dt?.Type.DataType == DataType.DateTime2
-					? dt.Type.Precision == 0
+					? (dt.Type.Precision == 0
```
viceroypenguin, line 153: would a switch expression be cleaner here? i.e.

```csharp
dt.Type.Precision switch
{
  0 => ...,
}
```

viceroypenguin (approved):

Source/LinqToDB/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -125,131 +139,135 @@ static List<Expression> CompactTree(List<Expression> items, ExpressionType nodeT
...
 		}
 
 		public static Expression ExpandExpression(Expression expression)
 		{
 			expression = AggregateExpression(expression);
 
-			var result = expression.Transform(expr =>
+			var result = _expandExpressionTransformer.Transform(expression);
+
+			return result;
```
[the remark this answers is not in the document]
MaceWindu: I'll get yelled at by @sdanyliv if I do it :) he likes to have it for debugging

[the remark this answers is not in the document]
viceroypenguin: 😂

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -272,7 +271,7 @@ static T NormalizeExpressions<T>(T expression, bool allowMutation)
 
 		SelectQuery MoveCountSubQuery(SelectQuery selectQuery, EvaluationContext context)
 		{
-			new QueryVisitor().Visit(selectQuery, e => MoveCountSubQuery(e, context));
+			selectQuery.Visit(new { context, optimizer = this }, static (context, e) => context.optimizer.MoveCountSubQuery(e, context.context));
```
sdanyliv, line 274: Replace with ValueTuple.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2339,10 +2395,10 @@ protected SqlUpdateStatement CorrectUpdateTable(SqlUpdateStatement statement)
 				else if (firstTable.Source is SqlTable newUpdateTable && newUpdateTable != updateTable && QueryHelper.IsEqualTables(newUpdateTable, updateTable))
 				{
 					statement.Update.Table = newUpdateTable;
-					statement.Update = ConvertVisitor.Convert(statement.Update, (v, e) =>
+					statement.Update = statement.Update.Convert(new { updateTable, newUpdateTable }, static (v, e) =>
```
sdanyliv, line 2398: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -165,33 +204,33 @@ public static DbDataType GetDbDataType(ISqlExpression? expr)
...
 		}
-		
+
 		public static void CollectDependencies(IQueryElement root, IEnumerable<ISqlTableSource> sources, HashSet<ISqlExpression> found, IEnumerable<IQueryElement>? ignore = null)
 		{
 			var hash       = new HashSet<ISqlTableSource>(sources);
 			var hashIgnore = new HashSet<IQueryElement>(ignore ?? Enumerable.Empty<IQueryElement>());
 
-			new QueryVisitor().VisitParentFirst(root, e =>
+			root.VisitParentFirst(new { hash, hashIgnore, found }, static (context, e) =>
```
sdanyliv, line 213: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -201,27 +240,27 @@ public static void CollectUsedSources(IQueryElement root, HashSet<ISqlTableSourc
 		{
 			var hashIgnore = new HashSet<IQueryElement>(ignore ?? Enumerable.Empty<IQueryElement>());
 
-			new QueryVisitor().VisitParentFirst(root, e =>
+			root.VisitParentFirst(new { hashIgnore, found }, static (context, e) =>
```
sdanyliv, line 243: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -1104,7 +1157,7 @@ public static TStatement WrapQuery<TStatement>(TStatement statement, SelectQuery
 		/// <returns>Same or new statement with removed joins.</returns>
 		public static SqlStatement JoinRemoval(SqlStatement statement, Func<SqlStatement, SqlJoinedTable, bool> joinFunc)
 		{
-			var newStatement = ConvertVisitor.ConvertAll(statement, (visitor, e) =>
+			var newStatement = statement.ConvertAll(new { joinFunc, statement }, static (visitor, e) =>
```
sdanyliv, line 1160: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -1378,12 +1431,12 @@ public static string ToDebugString(this IQueryElement expr)
 
 		public static SqlCondition CorrectSearchConditionNesting(SelectQuery sql, SqlCondition condition, HashSet<ISqlTableSource> forTableSources)
 		{
-			var newCondition = ConvertVisitor.Convert(condition, (v, e) =>
+			var newCondition = condition.Convert(new { sql, forTableSources }, static (v, e) =>
```
sdanyliv, line 1434: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/SelectQuery.cs
```diff
@@ -174,10 +124,10 @@ public void ForEachTable(Action<SqlTableSource> action, HashSet<SelectQuery> vis
 			foreach (var table in From.Tables)
 				table.ForEach(action, visitedQueries);
 
-			new QueryVisitor().Visit(this, e =>
+			this.Visit(new { query = this, action, visitedQueries}, static (context, e) =>
```
sdanyliv, line 127: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
```diff
@@ -73,7 +72,7 @@ static QueryData GetQueryData(IQueryElement? root, SelectQuery selectQuery, Hash
 		{
 			var data = new QueryData { Query = selectQuery };
 
-			new QueryVisitor().VisitParentFirst(root ?? selectQuery, e =>
+			(root ?? selectQuery).VisitParentFirst(new { selectQuery, visitedHash, data }, static (context, e) =>
```
sdanyliv, line 75: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
```diff
@@ -205,21 +204,21 @@ static void ResolveFields(QueryData data)
 			}
 
 			if (dic.Count > 0)
-				new QueryVisitor().VisitParentFirst(data.Query, e =>
+				data.Query.VisitParentFirst(new { dic, data }, static (context, e) =>
```
sdanyliv, line 207: Replace with ValueTuple.

Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
```diff
@@ -422,16 +419,16 @@ void OptimizeUnions()
 
 		void FinalizeAndValidateInternal(bool isApplySupported, bool optimizeColumns)
 		{
-			new QueryVisitor().Visit(_selectQuery, e =>
+			_selectQuery.Visit(new { optimizer = this, isApplySupported, optimizeColumns }, static (context, e) =>
```
sdanyliv, line 422: Replace with ValueTuple.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -117,307 +113,312 @@ private Expression UpdateNullabilityFromExtension(Expression resultExpr)
 			return resultExpr;
 		}
 
-		TransformInfo TransformExpression(IBuildContext context, Expression expr, bool enforceServerSide, string? alias)
+		public Expression BuildExpression(IBuildContext context, Expression expression, bool enforceServerSide, string? alias = null)
 		{
-			if (_skippedExpressions.Contains(expr))
-				return new TransformInfo(expr, true);
+			return expression.Transform(
+				new { builder = this, context, enforceServerSide, alias },
```
sdanyliv, line 119: Replace with ValueTuple.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -849,19 +871,19 @@ static Expression GetMultipleQueryExpressionLazy(IBuildContext context, MappingS
 
 			// Convert associations.
 			//
-			return expression.Transform(e =>
+			return expression.Transform(new { context, expression, parameters }, static (context, e) =>
```
sdanyliv, line 874: Replace with ValueTuple.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -921,35 +943,35 @@ public Expression BuildMultipleQuery(IBuildContext context, Expression expressio
 
 			// Convert parameters.
 			//
-			expression = expression.Transform(e =>
+			expression = expression.Transform(new { parameters, buildContext = context, builder = this, parms, paramex, enforceServerSide }, static (context, e) =>
```
sdanyliv, line 946: Replace with ValueTuple.

sdanyliv (commented):
It would be great to fix my comments.

Source/LinqToDB/Common/Converter.cs
```diff
@@ -140,12 +140,14 @@ static bool ToBoolean(char ch)
...
```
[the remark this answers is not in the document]
jods4: Just so I know when I write new code: 
I thought ValueTuple was not ok because of "old" .net framework support?
Are you changing the targets for release 4.0?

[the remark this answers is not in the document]
sdanyliv: We have included ValueTuple into library as internal class for older frameworks. So compiler is happy.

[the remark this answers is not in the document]
jods4: I see. Good that we can use it now at least internally.
Shouldn't you add a reference to https://www.nuget.org/packages/System.ValueTuple/ instead? This would re-use the same class if users have included ValueTuple in their own project.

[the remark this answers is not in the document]
sdanyliv: We don't want to force users to fix their projects by `bindingRedirect`  https://stackoverflow.com/questions/42867434/could-not-load-file-or-assembly-system-valuetuple

sdanyliv (approved):
Merge ASAP.

Issue #2960, opened 2021-04-17
BulkCopy(Async) Returning Inserted IDs


to11mtm:
The thought I had is that with most data providers, it would be possible to have `RowByRow` or `MultipleRows` methods of Bulk Copy return the inserted IDs for Auto-incremented primary keys.

 - Postgres: Via `Returning` statement
 - SqlServer: Via Table variable and `OUTPUT` into table variable, selecting table variable on return
 - Oracle: Via some magic involving `BULK COLLECT INTO` (I'm really rusty on my Oracle here...)
 - MySql: It looks like based on this: https://stackoverflow.com/a/53352324/2937845 we could use a simple calculation of `lasttrowid` alongside the number of records inserted to produce the IDs.
 - SqLite: Similar to MySQL, but using `last_insert_rowid()`

This would make certain operations more performant for users, specifically inserting lots of parent-child type records.

From an API Standpoint, I assume falling back to `RowByRow` would be the most simple option to maintain compatibility with providers that may not support returning multiple inserted IDs.

viceroypenguin:
You can use [InsertWithOutput()](https://linq2db.github.io/api/LinqToDB.LinqExtensions.html#LinqToDB_LinqExtensions_InsertWithOutput__1_LinqToDB_ITable___0____0_) (or any of it's variants) to do this. Since `RowByRow` does separate statements for each, calling this method on each row one-by-one has the same net behavior for the insert. However, it uses the `OUTPUT` or `returning` or equivalent statement on each server to return data from the inserted row.

to11mtm:
@viceroypenguin True but that would still require DB Round trips for each row inserted. The idea here is is that we could do something like:

```
var insertedIds = dataConnection.GetTable<SomeTable>().BulkCopyWithIdentity<long>(toInsert);
```

In the optimal case (Provider lets us get or safely infer inserted IDs from MultiRow statements) this could be done with fewer round trips than many `InsertWithOutput` calls in a loop.

I think I confused things by referencing `RowByRow`. In that case we could utilize the behavior of `InsertWithOutput` fairly easily I would think, but to me that's the compatibility/fallback case. I do see `InsertWithOutput()` now has a method that can take an `IQueryable<TSource>` in, but I'm not sure offhand if that can be used cleanly.

viceroypenguin:
[...]

PR #2975, opened 2021-04-29
Allow Parameter Usage on MultipleRowsCopy/MultipleRowsCopyAsync

### LINKED ISSUE #2144: Running BulkCopy without parameters
Why BulkCopy run without parameters?

Simple insert generates SQL with parameters:

> db.Insert(new SubmissionFile() { Status = "Uploaded2"});

`INSERT INTO submission_files
(
	status,
	filename,
	created,
	created_by
)
VALUES
(
	:Status,
	:Filename,
	:Created,
	:CreatedBy
)`

But BulkCopy generates SQL without parameters:
>db.BulkCopy<SubmissionFile>(new List<SubmissionFile>()
{
	new SubmissionFile() { Status = "Uploaded" },
	new SubmissionFile() { Status = "Uploaded2"}
});

`INSERT INTO submission_files
(
	status,
	filename,
	created,
	created_by
)
VALUES
('Uploaded','\\',NULL,'0001-01-01'::date,NULL,'0001-01-01'::date),
('Uploaded2','\\',NULL,'0001-01-01'::date,NULL,'0001-01-01'::date)
`

So, does this implementation has a relation to productivity or is it some else?
Could I use bulk operation with parameters?

to11mtm:
Allow Parameter usage on MultipleRowsCopy/MultipleRowsCopyAsync via `BulkCopyOptions`. Fix #2144


Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -344,20 +344,32 @@ protected async Task TraceActionAsync(DataConnection dataConnection, Func<string
...
+						helper.Parameters.RemoveRange(helper.LastRowParameterIndex, helper.ParameterIndex-helper.LastRowParameterIndex);
+						helper.StringBuilder.Length = helper.LastRowStringIndex;	
+					}
 					finishFunction(helper);
 					if (!helper.Execute())
 						return helper.RowsCopied;
+					if (needRemove)
+					{
+						addFunction(helper, item!, from);	
+					}
```
to11mtm, line 372: So, I think this could -possibly- be simplified a little bit. I put `LastRowParameterIndex` and `LastRowStringIndex` into `MultipleRowsHelper` since `.Execute()` is what is clearing out the other bits we are comparing to. I'm not certain whether that is considered better or worse than just tracking those values locally in these functions.

other option to clean up would be `if(CurrentCount > rowcount )`->`else if (needRemoveCondition)`.

to11mtm:
@viceroypenguin If you get a chance can you beat me up on this one please?

viceroypenguin:
First thing I would suggest is move the `maxParameters` and `maxSqlText` from `MultipleRowsCopyHelperAsync()` method to `MultipleRowsHelper` constructor. Then you can use `maxParameters` to reduce `MaxBatchSize` as necessary (`BatchSize = Math.Min(Math.Max(10, Options.MaxBatchSize ?? 1000), maxParameters / Columns.Count)`; notice the `-1`). This will remove the need to check for max parameters, since the batch size is implicitly less than max parameters. 

The final concern will be the text length. For that, your current solution will work, though I think you can track `LastRowXxx` as local variables rather than properties on the `Helper` object? There's no way to move that boiler-plate that's copied three times into a common method or pattern is there? Seems like just enough code to screw up editing later and get the three out of sync w/ each other.

to11mtm:
> 
> 
> First thing I would suggest is move the `maxParameters` and `maxSqlText` from `MultipleRowsCopyHelperAsync()` method to `MultipleRowsHelper` constructor. Then you can use `maxParameters` to reduce `MaxBatchSize` as necessary (`BatchSize = Math.Min(Math.Max(10, Options.MaxBatchSize ?? 1000), maxParameters / Columns.Count)`; notice the `-1`). This will remove the need to check for max parameters, since the batch size is implicitly less than max parameters.
> 

Thank you for the feedback! I'll give that a try. I was trying not to be too aggressive in this PR since next round (i.e. another PR) I'd like to optimize the build stage so that we don't re-build the string/parameter lists. I think to do it well we'll want to re-do the loop.

> The final concern will be the text length. For that, your current solution will work, though I think you can track `LastRowXxx` as local variables rather than properties on the `Helper` object? There's no way to move that boiler-plate that's copied three times into a common method or pattern is there? Seems like just enough code to screw up editing later and get the three out of sync w/ each other.

We might be able to pull out at least some of the logic, challenge is each one is just different enough where more indirection feels like as much cognitive load as the duped code.

But, I think overall the loop structure can be improved at least; Rather than Adding the item, and then checking length to decide whether to back off, we can check length at the start of the iteration and decide what to do. I think it will make the code a little cleaner, and will probably be cleaner as it will have less jumps.

viceroypenguin:
Definitely agree with re-working to check the length before adding. That would simplify all three back to reasonable levels again and don't need to abstract the three versions of the loop into one. 

Personally, I'd say go ahead and do it, rather than trying to piece-meal in two separate PRs, but that's a personal opinion. On the other hand, if you're going to redo it in a second PR, then what you've got works well enough for now I think.

to11mtm:
> 
> 
> Definitely agree with re-working to check the length before adding. That would simplify all three back to reasonable levels again and don't need to abstract the three versions of the loop into one.
> 
> Personally, I'd say go ahead and do it, rather than trying to piece-meal in two separate PRs, but that's a personal opinion. On the other hand, if you're going to redo it in a second PR, then what you've got works well enough for now I think.

If I wasn't about to deal with a bunch of release stuff at work that's gonna have me busy for a while I'd agree. :)

Question though, based on your other feedback; Should I put the `MaxParameters`/`MaxSqlText` into `SqlProviderFlags`? I think that may be cleaner overall, but it still raises the question of whether we want to have 'Real' `MaxParameters` or ones that make sense for working in .NET

to11mtm:
I looked into setting MaxBatchSize based on param usage, it unfortunately gets a little hacky to work-around the behavior of Oracle's Array-parameter copy. Might be better to get what is here into 3.4.0 for now, so I am marking ready to see if there is other feedback or if we must make more changes first.

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -88,5 +88,10 @@ public BulkCopyOptions(BulkCopyOptions options)
 		/// This callback will not be used if <see cref="NotifyAfter"/> set to 0.
 		/// </summary>
 		public Action<BulkCopyRowsCopied>? RowsCopiedCallback { get; set; }
+
+		/// <summary>
+		/// Gets or sets whether to Always use Parameters for MultipleRowsCopy.
```
MaceWindu, line 93: makes sense to add default value information and behavior in this mode like splitting operation into batches if parameters limit reached for operation.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -15,6 +15,9 @@ namespace LinqToDB.DataProvider
 
 	public class BasicBulkCopy
 	{
+		protected virtual int MaxParameters => 999;
```
MaceWindu, line 18: I think it makes sense to use those options as not max but as default limits and allow user to override them in bulk copy options. This will allow user to workaround issues with incorrect limits

Tests/Linq/Update/BulkCopyTests.cs
```diff
@@ -276,5 +276,22 @@ public void ReuseOptionTest([DataSources(false, ProviderName.DB2)] string contex
 				db.Child. BulkCopy(options, new[] { new Child  { ParentID = 111001 } });
 			}
 		}
+		
+		[Test]
+		public void UseParametersTest([DataSources(false, ProviderName.DB2)] string context)
```
MaceWindu, line 281: What's wrong with DB2?

MaceWindu (changes requested):
Could you provide sources for limits for various providers in comments? I see some of them specified but some not

MaceWindu:
@to11mtm, we plan to release 3.4 next thursday, it would be nice if we can include this one.

to11mtm:
> 
> 
> Could you provide sources for limits for various providers in comments? I see some of them specified but some not

I'll try to find my notes again. Most came from https://www.jooq.org/doc/3.12/manual/sql-building/dsl-context/custom-settings/settings-inline-threshold/ but for the strings I had to look in some other places.

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
to11mtm: Are you suggesting a name to something like `DefaultMaxParameters` then? or is this name OK?

MaceWindu: Yep, `DefaultMaxParameters ` sounds good

to11mtm:
Looks like Firebird may be mad about coercion of nulls. Looking into that and the DB2 failures

MaceWindu:
Firebird doesn't use type information from parameter, so you need to add it using CAST to first select of UNION

to11mtm:
> 
> 
> Firebird doesn't use type information from parameter, so you need to add it using CAST to first select of UNION

Yeah I found FirebirdSqlOptimizer.WrapParameters and cried for a few minutes.

I'm giving it a try with just casts on all rows. Ironically, after our other optimizations to SQL building (i.e. #2774) It appears that at least for SQL Server and maybe postgres, Parameters are slower than our string literals at this point.  my guess/assumption is that the cost of marshaling the parameters is eating away any benefits, and because we aren't treating this as a prepared statement there's no gains to be had there either. My hope is that when myself or someone else does a more thorough refactoring of `BulkCopy` we can address some of this and some of the other cruft in BulkCopy/MultipleRowsHelper.

Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
```diff
@@ -77,7 +79,27 @@ public virtual void BuildColumns(object item, Func<ColumnDescriptor, bool>? skip
 				{
 					var name = ParameterName == "?" ? ParameterName : ParameterName + ++ParameterIndex;
 
-					StringBuilder.Append(name);
+					if (castParameters)
```
MaceWindu, line 82: CAST needed only for first select in union. For other rows db will use type information from first type

Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
```diff
@@ -77,7 +79,27 @@ public virtual void BuildColumns(object item, Func<ColumnDescriptor, bool>? skip
...
+					{
+						StringBuilder.Append("CAST(");
+						StringBuilder.Append(name);
+						StringBuilder.Append(" AS ");
+						StringBuilder.Append(column.DataType);
+						if (!string.IsNullOrEmpty(column.DbType))
```
MaceWindu, line 88: can we use sql builder here as it already have db-specific type builder method that takes into account db-specific stuff

to11mtm: If you could give some pointers on which method to try to use in SQLBuilder I can give it a shot after work.

Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
to11mtm: Yeah in retrospect I don't think it would hurt to make this `if (castParameters && CurrentCount==0)`. I was over-thinking the jump cost lol.

Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
MaceWindu: BuildDataType https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs#L2716
but you need to make it `protected internal`

to11mtm: I ran into some challenges here (we are using `ISqlBuilder`, can't cleanly internal on an interface,) so tried to solve for this by adding to ISqlBuilder Interface,  `StringBuilder    BuildDataType        (StringBuilder sb, SqlDataType dataType);`. I went with returning the stringbuilder because that was the convention for BuildTableName, Convert, and others.

MaceWindu:
@to11mtm, if you are interested in working on this area more I have several suggestions for improvements:
- issue, similar to one you fought with Firebird, still exists for parameter-less mode when first data row contains nulls (no type information issue for multiple databases)
- instead of query generation for each batch, we can reuse generated SQL for batches of same size
- it makes sense to investigate new batching feature from .net 6 : https://github.com/dotnet/runtime/issues/28633 . Here I would add that some providers already support it. IIRC SAP HANA provider repeat command if command use less parameters than you passed to command, but it is provider with positional parameters, so it is not a common case

to11mtm:
> 
> 
> @to11mtm, if you are interested in working on this area more I have several suggestions for improvements:
> 
>     * issue, similar to one you fought with Firebird, still exists for parameter-less mode when first data row contains nulls (no type information issue for multiple databases)
> 
>     * instead of query generation for each batch, we can reuse generated SQL for batches of same size
> 
>     * it makes sense to investigate new batching feature from .net 6 : [dotnet/runtime#28633](https://github.com/dotnet/runtime/issues/28633) . Here I would add that some providers already support it. IIRC SAP HANA provider repeat command if command use less parameters than you passed to command, but it is provider with positional parameters, so it is not a common case

Yeah, I want to do more work on this in future; My big goal is to have version of BulkCopy that will return inserted Identity columns (issue #2960 ), but it is a bigger refactor because every DB that supports returning on MultipleRows has different rules on how to do this right. 

I started looking into re-using the generated SQL, I think for it to be of actual benefit we will need to do more refactoring. Our stringbuild is pretty efficent as-is, to get a worthwhile benefit I think we would have to also `.Prepare()` the statement, and go a little lower level on re-assigning parameter values (at least some data providers, if you add/remove/reassign `DbParameter` on `DbCommand.Parameters` the `.Prepare()` gets thrown out, you have to ONLY touch the actual Value of the existing `DbParameter` object, if that makes sense.)

Source/LinqToDB/DataProvider/BasicBulkCopy.cs
```diff
@@ -15,6 +15,9 @@ namespace LinqToDB.DataProvider
...
+		protected virtual int MaxSqlLength => 100000;
```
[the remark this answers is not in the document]
MaceWindu: COuld be usful. In baselines I see increase in number of roundtrips e.g. for DB2

PR #2984, opened 2021-05-09
Add sql server support to RetrieveIdentity

### LINKED ISSUE #2983: DataExtensions.RetrieveIdentity() returns wrong Identity values
[`DataExtensions.RetrieveIdentity()` ](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Tools/DataExtensions.cs#L23)  build internally query to retrieve the last identity column value calling [`GetMaxValueSql()`](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Tools/DataExtensions.cs#L60).

This returns wrong value because the query that is used is ["SELECT Max({columnName}) FROM {tableName}"](https://github.com/linq2db/linq2db/blob/v3.3.0/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs#L3203) which is not the last identity value. SQL would return the last identity value calling `IDENT_CURRENT('Table_Name')`

The steps to reproduce the issue are:

1. Insert records
2. Go to SQL and delete the records
3. Insert again 

The returned IDs will not match the IDs in  SQL

Note: Turning off the SQL Identity cache does not solve the issue

MaceWindu:
Fix #2983

feature hidden behind `useIdentity` parameter to preserve backward compatibility.

### DIFF
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -1324,5 +1324,27 @@ namespace LinqToDB
 		}
 
 		#endregion
+
+		#region Identity Functions
+		// identity APIs are nternal as:
+		// - there is plans to make them public for now
+		// - support for more providers required
+
+		/// <summary>
+		/// Returns last identity value (current value) for specific table.
+		/// </summary>
+		[Function  (PN.SqlServer    , "IDENT_CURRENT", ServerSideOnly = true)]
+		[Expression(PN.SqlServer2000, "NULL"         , ServerSideOnly = true)]
+		[Expression(                  "NULL"         , ServerSideOnly = true)]
+		internal static object? CurrentIdentity(string tableName) => throw new LinqException($"'{nameof(CurrentIdentity)}' is server side only property.");
+
+		/// <summary>
+		/// Returns identity step for specific table.
+		/// </summary>
+		[Function  (PN.SqlServer    , "IDENT_INCR", ServerSideOnly = true)]
+		[Expression(PN.SqlServer2000, "NULL"      , ServerSideOnly = true)]
+		[Expression(                  "NULL"      , ServerSideOnly = true)]
+		internal static object? IdentityStep(string tableName) => throw new LinqException($"'{nameof(IdentityStep)}' is server side only property.");
+		#endregion
 	}
 }
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -1326,8 +1326,8 @@ namespace LinqToDB
 		#endregion
 
 		#region Identity Functions
-		// identity APIs are nternal as:
-		// - there is plans to make them public for now
+		// identity APIs are internal as:
+		// - there is no plans to make them public for now
 		// - support for more providers required
 
 		/// <summary>
--- a/Source/LinqToDB/Tools/DataExtensions.cs
+++ b/Source/LinqToDB/Tools/DataExtensions.cs
@@ -4,26 +4,34 @@ using System.Linq;
 
 namespace LinqToDB.Tools
 {
+	using System.Text;
 	using Common;
 	using Data;
 	using LinqToDB.Mapping;
+	using LinqToDB.SqlProvider;
 
 	public static class DataExtensions
 	{
 		/// <summary>
-		/// Initializes source columns, marked with <see cref="ColumnAttribute.IsIdentity"/> with values.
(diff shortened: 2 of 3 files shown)


Tests/Linq/DataProvider/SqlServerTests.cs
```diff
@@ -1913,5 +1914,30 @@ public void TestDescriptions([IncludeDataSources(false, TestProvName.AllSqlServe
...
+					db.Insert(new Person() { FirstName = "", LastName = "" });
+					db.Insert(new Person() { FirstName = "", LastName = "" });
+				}
+
+				var lastIdentity = db.Execute<int>("SELECT IDENT_CURRENT('Person')");
+				var step         = db.Execute<int>("SELECT IDENT_INCR('Person')");
+
+				var persons  = Enumerable.Range(1, 10).Select(_ => new Person()).ToArray();
+
+				persons.RetrieveIdentity(db);
```
viceroypenguin, line 1936: Doesn't test `useIdentity = true`?

MaceWindu: whoops, added parameter as last step and didn't updated test

viceroypenguin: For completeness, I'd do what he did, insert some, delete some, insert some more. Would expect to work fine, based on what I read, but safer for sure.

viceroypenguin (approved):

Source/LinqToDB/Tools/DataExtensions.cs
```diff
@@ -41,54 +49,107 @@ public static class DataExtensions
...
-						if (maxValue == null || maxValue == DBNull.Value)
-							maxValue = 0;
+						if (useIdentity)
+						{
+							var sb = new StringBuilder();
+							sqlBuilder.BuildTableName(sb, entityDescriptor.ServerName, entityDescriptor.DatabaseName, entityDescriptor.SchemaName, entityDescriptor.TableName, entityDescriptor.TableOptions);
+							var tableName = sb.ToString();
 
-						var type = Type.GetTypeCode(maxValue.GetType());
+							var identity = context.Select(() => new { last = Sql.CurrentIdentity(tableName), step = Sql.IdentityStep(tableName) });
```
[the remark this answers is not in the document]
MaceWindu: `Sql.CurrentIdentity` maps to database function and will throw exception if you try to execute it outside of query context (client-side execution).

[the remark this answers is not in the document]
KostaVlev: I see it now...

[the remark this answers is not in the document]
KostaVlev: @MaceWindu  using `var identity = context.Select(() => new { last = Sql.CurrentIdentity(tableName), step = Sql.IdentityStep(tableName) });`  sintax is throwing exeption `DataReader associated with this Command which must be closed first.'`

``
var last = Sql.CurrentIdentity(tableName);
var step = Sql.IdentityStep(tableName);
``

works fine ... is this make sence? @MaceWindu

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -1324,5 +1324,27 @@ public static DateTime GetDate()
...
+		// identity APIs are internal as:
+		// - there is no plans to make them public for now
+		// - support for more providers required
+
+		/// <summary>
+		/// Returns last identity value (current value) for specific table.
+		/// </summary>
+		[Function  (PN.SqlServer    , "IDENT_CURRENT", ServerSideOnly = true)]
+		[Expression(PN.SqlServer2000, "NULL"         , ServerSideOnly = true)]
+		[Expression(                  "NULL"         , ServerSideOnly = true)]
```
sdanyliv, line 1338: Why NULL for other providers?

MaceWindu: So `RetrieveIdentity` will not fail if called for non-sql server provider with `useIdentity: true`. I don't plan to implement it for other databases without request as I think whole `RetrieveIdentity` API shouldn't even exist in linq2db :)

sdanyliv (approved):

Source/LinqToDB/Sql/Sql.cs
viceroypenguin: Agreed - there are better ways to manage retrieving the identity. :)

viceroypenguin (approved):

PR #3006, opened 2021-05-24
Recognize more values as empty update in InsertOrUpdate

### LINKED ISSUE #3005: Allow empty update expression in InsertOrUpdate API
see https://rsdn.org/forum/prj.rfd/8014195

We should allow `New` and propably `null` expressions here to improve usability and discovery of this API

MaceWindu:
Fix #3005

alongsize with `r => new Table() { }` empty init statement following values also now recognized:
- `null`
- `record => null`
- `record => new Table()`

### DIFF
--- a/Source/LinqToDB/Linq/Builder/ExpressionContext.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionContext.cs
@@ -4,6 +4,7 @@ using System.Linq.Expressions;
 
 namespace LinqToDB.Linq.Builder
 {
+	using LinqToDB.Common;
 	using LinqToDB.Expressions;
 	using SqlQuery;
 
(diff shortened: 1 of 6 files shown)


sdanyliv (approved):

Source/LinqToDB/Linq/Builder/ExpressionContext.cs
```diff
@@ -64,6 +65,16 @@ public override SqlInfo[] ConvertToSql(Expression? expression, int level, Conver
...
 								}
 							}
+							else if (root.NodeType == ExpressionType.Constant)
+							{
+								if (((ConstantExpression)root).Value == null)
+									return Array<SqlInfo>.Empty;
+							}
+							else if (root.NodeType == ExpressionType.New)
+							{
+								if (((NewExpression)root).Arguments.Count == 0)
```
viceroypenguin, line 75: Correct me if I'm wrong, but `.Arguments` is about the arguments into the constructor, and not the member bindings in the `new { ... }`? So this would incorrectly handle `new Table { ColumnA = 1, }` (since no arguments to the constructor, but member bindings. 

Should this check arguments _and_ `.Members` as well?

MaceWindu: this is `ExpressionType.New`, so there is only constructor call

viceroypenguin: Ah, got caught on the `NewExpression` type, and forgot the `NodeType` as well. I'm good then. :)

viceroypenguin (approved):

PR #3022, opened 2021-05-30
Distinct+OrderBy + Take/Skip fixes

### LINKED ISSUE #2943: Likely bug in BasicSqlOptimizer
I saw this while reading source code:
https://github.com/linq2db/linq2db/tree/42afc3ae9cbdf311ef6dfb4af5152c3914b33263/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs#L2882-L2886

That condition cannot be true, can it?
`projectionItems` being the union of `columnItems` + something else, it cannot have less items than `columnItems` alone.

MaceWindu:
Fix #2943 (part 1)

First reported issue is about unreachable code in `ReplaceDistinctOrderByWithRowNumber` method. This method used only for SQL Server 2005 but when I tried to create test for this code, I found out that tests doesn't work for other databases (at least SQL Server 2008+), e.g. order by clause is not correct.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
@@ -42,8 +42,7 @@ namespace LinqToDB.Linq.Builder
 
 			var wrapped = false;
 
-			if (sequence.SelectQuery.Select.TakeValue != null ||
-				sequence.SelectQuery.Select.SkipValue != null)
+			if (sequence.SelectQuery.Select.HasModifier)
 			{
 				sequence = new SubQueryContext(sequence);
 				wrapped = true;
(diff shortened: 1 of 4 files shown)


MaceWindu:
queries broken for:
- sql server (all)
- access
- sqlce
- sap hana
- db2
- sybase

works for:
- postgresql
- sqlite
- oracle
- firebird
- mysql
- informix

MaceWindu:
tests work now, still need to decide what to do with unused code in ReplaceDistinctOrderByWithRowNumber

MaceWindu:
FYI, right now ReplaceDistinctOrderByWithRowNumber doesn't affect generated SQL at all for existing tests (both DB2 and SQL Server 2005)

PR #3023, opened 2021-05-30
Investigate potentially unused code

MaceWindu:
Address second concern from #2943


MaceWindu:
Great, actually we have test coverage for some providers and could investigate it further

PR #3024, opened 2021-05-30
Update SQLite dependency

MaceWindu:
relese 1.0.114 was broken for non-windows targets, so we need to update once more


MaceWindu:
Postponed, filled issue for sqlite provider https://system.data.sqlite.org/index.html/tktview/e58a0b5baeeea1fad1069550840d2044db50969f

PR #3027, opened 2021-05-30
WIP Row constructors

jods4:
This PR adds support for SQL Row Constructors, which basically are SQL tuples, usually with syntax `ROW (1, 2)` or just `(1, 2)`.

### Design 
Row constructors are only server-side, they're always represented by a call to `Sql.Row(1, 2)` and they can't be passed as parameters.

The returned type is `SqlRow<T1, T2>`. 
Tuples with more elements simply nest `SqlRow` in the second argument: e.g. `SqlRow<int, SqlRow<int, int>>`  for a triple.
This may look inefficient but remember this will never be instantiated. It's only a compilation-time type to enforce static types.
The benefit when compared to `SqlRow<T1, T2, T3>` is that I only declared a single type and declared all its operators (such as `>`, `<`) once.
I created `Sql.Row` overloads up to 8 values. With this design, it's trivial to add more if so desired.

Sadly types are stictly checked by C# and I see no way to relax the constraints a bit when it comes to nullable types.
If one side has a `int?` and the other has not, a cast is required:
```csharp
// Error, SqlRow<int, int> incompatible with SqlRow<int, int?>
// Needs a cast (int?)2
Row(1, 2) == Row(3, null) 
```

Row constructors are always evaluated without `CompareNullsAsValues`. The problem here is that generated row ctor code must behave the same as equivalent code in DB without row ctor, and nothing makes sense (to me) for row ctor expressions.

There is no attempt to completely or partially evaluate a `SqlRow` expression, at least for now. 
For example in `(1, 2, 3) = (4, 5, 6)` one could check if any pair of values are constant that are different, or if one is null, and the whole expression is statically false. If any pair of values are constants that are the same, in theory the expression could be shortened.
Notice that in DB that don't support row ctors natively, this expression is converted by SqlOptimizer into an equivalent predicate and that runs through the optimizer and would be simplified.

### IS [NOT] NULL
The following translations are used when not supported:
```
(1, 2) IS NULL     -> 1 IS NULL AND 2 IS NULL
(1, 2) IS NOT NULL -> 1 IS NOT NULL AND 2 IS NOT NULL
```

### Comparisons
The following translations are used when not supported:
```
(1, 2) = (3, 4)        -> 1 = 3 and 2 = 4
(1, 2) <> (3, 4)       -> 1 <> 3 or 2 <> 4
(1, 2, 3) > (4, 5, 6)  -> 1 > 4 or 1 = 4 and 2 > 5 or 1 = 4 and 2 = 5 and 3 > 6
(1, 2, 3) >= (4, 5, 6) -> 1 > 4 or 1 = 4 and 2 > 5 or 1 = 4 and 2 = 5 and 3 >= 6
Likewise for < and <=
```

### [NOT] IN 
When not supported, expands to a big OR: 
```
(1, 2) IN ((3, 4), (5, 6)) -> (1, 2) = (3, 4) OR (1, 2) = (5, 6)
```
This gets translated further in case the provider doesn't support comparisons either, see above.

### OVERLAPS
As described in a comment below, emulating the precise behavior of OVERLAPS in SQL is practically impossible. 
Client-side C# support would in theory be possible but is much less useful.
Approximate, but still useful, behavior is possible but is left for future improvements, if so desired.

In this PR, OVERLAPS is only available natively, which means in DB that have support and only for specific types (i.e. dates and intervals).

Because of the combinatoric explosion of valid type pairs, the static C# typing of `.Overlaps()` accepts any type at all. At runtime the types are validated and a specific error message is thrown if they are invalid (e.g. numbers).

```
(date'2020-01-01', date'2020-12-31') OVERLAPS (date'2020-07-01', null) -- true
```

### [NOT] BETWEEN
When not supported, converts to two comparisons:
```
1     BETWEEN 2 AND 3 -> 1 >= 2 AND 1 <= 3
1 NOT BETWEEN 2 AND 3 -> 1 < 2 OR 1 > 3
```
This gets translated further in case the provider doesn't support comparisons either, see above.

### Comparisons to (SELECT)
Providers that natively support Row Constructors can compare them to a sub-select query returning a single row of the same cardinality.

This might probably be emulated with a [NOT] EXISTS in other providers but I leave this to future improvements, if someone wants it.

The public API is like any scalar comparison to a sub-select: write a query returning a `Sql.Row()` and ends it with `.Single()`. Or if it might return more than one row, use `.First()` as appropriate.
Since any `object` can be compared for equality, the `.Single()` call is not actually needed for operators `==` and `!=`.

```csharp
from x in db.Source
where Sql.Row("name", x.B) >= 
  (from y in db.OtherSource
   where y.Id == 42
   select Sql.Row(y.Name, 0))
  .Single()
select x
```

Results in :
```sql
where ('name', x.B) >= (select y.Name, 0 from OtherSource y where y.id = 42)
```

Note that it's a runtime error to try to select `SqlRow` in the main SELECT clause, and `SqlRow` must be the only selected value.

> I've tried to define an implicit conversion from `IQueryable<SqlRow>` to `SqlRow` to avoid the `.Single()` and make it easier, but C# forbids the definition of implicit conversions from an interface.
>
> I don't think (?) linq2db has support for `>= [ANY|ALL] (SELECT multiple rows)`, please tell me if it does and I'll check it works properly.

### TBD: more features coming
Complete support would include UPDATE SET.

### Provider support
| Provider | Comparisons | IS NULL | IN | OVERLAPS | BETWEEN | Comp. to (SELECT)
| --- | --- | --- | --- | --- | --- | ---
| Access | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ 
| DB2 | ✅ | ⬇ | ⬇ | ✅ | ✅| ✅ 
| Firebird | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔  
| Informix | ✅ (1) | ⬇ | ✅ (1) | ⛔ | ⬇ | ✅
| MySQL | ✅ | ⬇ | ✅ | ⛔ | ⬇ | ✅
| Oracle | ✅ (2) | ⬇ | ✅ | ✅ | ⬇ | ✅
| Postgre | ✅ | ✅ | ✅ | ✅ | ✅ | ✅
| SQLite | ✅ | ⬇ | ⬇ | ⛔ | ✅ | ✅
| SapHana | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ 
| SqlCe | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ 
| SqlServer | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ 
| Sybase | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ 

Notes:
1. Informix uses `ROW (1, 2)` syntax
2. Oracle needs extra brackets on the right-hand side to disambiguate syntax: `(1, 2) = ( (3, 4) )`


Source/LinqToDB/Sql/Sql.Row.cs
```diff
@@ -0,0 +1,67 @@
...
+			public static bool operator >= (SqlRow<T1, T2> x, SqlRow<T1, T2> y)
+				=> throw new NotImplementedException();
+
+			public static bool operator <= (SqlRow<T1, T2> x, SqlRow<T1, T2> y)
+				=> throw new NotImplementedException();
+		}
+
+		[Sql.Expression(PN.Informix, "ROW ({0}, {1})", ServerSideOnly = true)]
+		[Sql.Expression("({0}, {1})", ServerSideOnly = true)]
+		public static SqlRow<T1, T2> Row<T1, T2>(T1 value1, T2 value2)
```
sdanyliv, line 31: It should be not `Sql.Expression` but additional SQL AST element.

jods4: Does it really matter? I get most things for free that way.

jods4:
☝️  Updated description with IN infos

jods4:
@sdanyliv  @MaceWindu How do you guys feel about  `OVERLAPS`?

SQL:2011 defines an `OVERLAPS` predicate over 2 intervals of time: row constructor of arity 2 with either `(date, date)` or `(date, interval)`.
For example you could do:
```sql
WHERE ('2020-01-01', '2020-01-31') OVERLAPS ('2020-01-15', '2020-01-22')
```
If we have a translation for DB not supporting row constructors, there's no reason not to support _any_ interval, e.g. numbers, and translate it for _all_ providers: `(1, 4) overlaps (7, 8)`.

To me, the problem is that translation. A basic SQL translation would be:
```sql
(a1, a2) overlaps (b1, b2)
b1 < a2 and a1 < b2
```
(note that intervals are inclusive at start and exclusive at end)

But a _faithful_ translation of `OVERLAPS` is made difficult by 3 facts:
1. Intervals might be out-of-order, i.e. it's ok if `a2 < a1`, and/or `b2 < b1`.
2. If intervals have a single `null` value, it is considered the end of interval, and non-null value its start.
3. If two intervals start on the same date, they overlaps even if one of them also ends on that same date (i.e. it's duration is 0).
4. Solution for 1,2,3 should work correctly wrt to null values. E.g. `(1, 10) overlaps (5, null)` should be true because the second interval falls in the middle of the first, regardless of what `null` could be. On the other hand, `(1, 10) overlaps (0, null)` should be `unknown`.

I couldn't find even complex expressions that are 100% correct (how complex? see below).

What do you think I should do?
- Not include `OVERLAPS` at all; 
- Include `OVERLAPS` for providers that support it natively (db2, Oracle, Postgre);
- Include `OVERLAPS` and an imprecise fallback.

If we assume that pairs are in-order, following expression should be correct:
```sql
(a1 > b1 and (a1 < b2 or b1 < b2))
or
(b1 > a1 and (b1 < a2 or b2 < b1))
or
(a1 = b1 and a2 is not null and b2 is not null)
```

As I mentionned above, a simple alternative could be `a1 < b2 and b1 < a2` but this:
- Only works if pairs are in-order;
- Never evaluates to true when some values are `null`;
- Doesn't handle properly the case of `(1, 2) overlaps (1, 1)` (should be true).

Source/LinqToDB/Sql/Sql.Row.cs
sdanyliv: Yes, in this case we can easily transform AST for provider which do not support such feature. Also we can reuse this class for defining more effective Oracle update.
```sql
update some
set (field1, field2) = (1, 2)
```

jods4: @sdanyliv So far I managed to do translations for other providers without too much problem. 
Let's see how it goes for future features.

`UPDATE` is my ultimate goal, it's what I really, really need in my project. 

The rest is just because I want to build a comprehensive linq2db solution, so that the design of `UPDATE` isn't inconsistent with row ctor if they were added in the future...

jods4:
Rebased on master.
☝️ Updated description with OVERLAPS infos

jods4:
☝️ Updated description with BETWEEN infos

jods4:
☝️ Updated description with comparisons to (SELECT)

Issue #2850, opened 2021-02-24
DTO Objects


AndreasWaning:
For me Linq2DB is the best object mapper there is. Thanks for that. I would have one suggestion. When the database classes are created, couldn't you create a DTO (Data Transfer Object) class directly ?
Now I always do this manually. I don't know how others feel about this, I would appreciate it.

regards,
Andreas

MaceWindu:
Could you clarify what you mean?

If you are talking about database scaffolding, we have T4 templates to generate mappings from database (see `linq2db.<DB_NAME>` nuget packages).

AndreasWaning:
This is generate from my Database table:
	[Table(Schema="dbo", Name="JobList")]
	public partial class JobList
	{
		[Column(),                  PrimaryKey,  Identity] public int    ID             { get; set; } // int
		[Column("Job_ID"),          NotNull              ] public int    JobId          { get; set; } // int
		[Column(),                     Nullable          ] public short? Priority       { get; set; } // smallint
       }

now, I need a  (d)ata (t)ransfer (o)bject for transfer the database content from the server to the client and vise versa.

   public class JobListDto
    {
		public int ID { get; set; }
		
		public int JobId { get; set; }
		
		public short Priority { get; set; }
   }

But this class a always create by hand.  Can I create this DTO automatically ?

I am from Germany,m it is hard to explain, what I mean. Hope you understand what I mean.

https://stackoverflow.com/questions/1051182/what-is-a-data-transfer-object-dto

MaceWindu:
Ok, so basically you want to generate another class without attributes.
I can see following options for generation:
- DTO namespace 
- DTO class name template
- enable generation of conversion methods between dto and model

BTW, as workaround you can customize your T4 template to do it already.
Check https://linq2db.github.io/articles/T4.html#useful-members-and-data-structures

you basically need to generate new `Class` instances for each data model table and add them to `Model.Types` collection.

Issue #2803, opened 2021-01-28
Association Property names


myabdolmaleki:
hi

im using linq2db in my projects and its wonderful, but i have a question.

why it create property names like DeleteRecCarta22951Afds? but my Association Property Type is a List of DeleteRecordCatableFinancing class.

linq2db can generate association property name  like "DeleteRecordCatableFinancings"?

thanks a lot

viceroypenguin:
Without seeing your schema or knowing which DB you're using, I'm not sure off the top of my head why it's generating that name for the FK link, but you can change it if you want. Open the `.tt` file you're using for generation and between the `Load<dbtype>Metadata()` call and the 'GenerateModel()' call, add this code:

```csharp
		GetFK("<tablename>", "<db_fk_name>")
			.MemberName = "DeleteRecordCatableFinancings";
```

This will tell the generator to use the new name instead of the one it defaulted to based on the FK name. You can do this for each FK that you want to change names for.

viceroypenguin:
See [here](https://linq2db.github.io/articles/T4.html) for more information on how to modify the generated code.

sdanyliv:
@myabdolmaleki, for such case we have introduced callback methods:
https://linq2db.github.io/articles/T4.html#configuring-generation-process
```cs
// Defines method to generate name for "one" side of association
Func<ForeignKey, string> GetAssociationExtensionSingularName
    = GetAssociationExtensionSingularNameDefault;
// Defines method to generate name for "many" side of association
Func<ForeignKey, string> GetAssociationExtensionPluralName
    = GetAssociationExtensionPluralNameDefault;
```

myabdolmaleki:
Hi

Im looking for automatic way, these approaches you said are handy, my database is sql server.

Thanks

sdanyliv:
I don't get that. Many users have different naming conventions. So we have implemented callback mechanism to define exactly what is needed.

Issue #2766, opened 2021-01-09
Add SequenceNameAttribute generation to T4


MaceWindu:
Currently we don't load sequence information in schema provider and don't generate sequence attributes on model

Issue #2331, opened 2020-07-08
T4 Template DataModel back references are generated as Guids


drewlsvern:
I'm generating C# classes from the T4 templates. The classes are generated without errors but some back references are assigned random guids as opposed to the related table name. Is this by design? If so, why and can it be changed to reflect the correct table name property?

```c#
var result = Db.TblAccounts.LoadWith(x => x._30F767840CF546D08AC8552E413F4488)
                                       .ThenLoad(x => x.BE9C6CA6203A4145AAF505280E679F6B)
                                       .LoadWith(x => x.Contact);
```

### Environment details
linq2db version: 3.0.0-rc.2
Database Provider: Microsoft.Jet.OLEDB.4.0 (Access)
Operating system: Windows 10 Version 2004
.NET Version: .NET Core 3.1 (Console Application)

MaceWindu:
That's not what we observe in our tests ([example](https://github.com/linq2db/linq2db/blob/master/Tests/Tests.T4/Databases/Access.generated.cs#L137)) so I would suspect something in your db causing it. We don't even generate guids in related code anywhere.

Could you provide your mdb file (without data and trim tables to leave minimal required set of tables to reproduce issue)?

drewlsvern:
Attached is trimmed sample with two tables that produce the issue.
Let me know if you need more info.

[Db.Sample.mdb](https://drive.google.com/file/d/1ul6Fret70aqj2fWuLk9s1qhtQRffw3d9/view?usp=sharing)

MaceWindu:
![image](https://user-images.githubusercontent.com/1099238/87056306-b83f4280-c205-11ea-94f4-89f452f2b35a.png)

This is what we get from `GetOleDbSchemaTable(Foreign_Keys)` (collapsed empty columns).
Which means that wether FK relation really has such guid-like name. Or as speculation (cannot find any info on this) maybe FK name is not set and JET provider return some object id in such cases.

MaceWindu:
I don't have access installed, but looking at screens here https://docs.microsoft.com/en-us/office/troubleshoot/access/define-table-relationships I see that threre is no names rendered for relations. Probably when you create relation from Access, it generates guid and use it as a name.

MaceWindu:
We don't do it for other databases, but probably we can detect such names and generate something more user-friendly.

MaceWindu:
As workaround you can:
[...]

PR #3098, opened 2021-07-04
New scaffolding framework

### LINKED ISSUE #1383: Create shared model generation framework to use with T4 and Linqpad

### LINKED ISSUE #1468: MS SQL Temproray tables in procedures. How to build
Good day.
1) For mssql, can i use temporary tables in procedures and with what valid names?
I ran tt template and saw that procedure loaded with error. Error shows some problem with temprory table: _"Invalid object name '#Data' "_

2) Is it possible to locally build the library and add dll to my project?
I want to try debug the app when it loads metadata.

Thank you

### Environment details
linq2db version: *2.6.0*
Database Server: *MS SQL Server 16*
Operating system: *E.g. Windows 10*
Framework version: *.NET Core 2.1*

### LINKED ISSUE #1586: Table named "Relationships"
Hello,

I created my database (MS Sql)  with a table named "Relationship". After running T4 templates, the generated code failed to compile:

image
In this case, the name of the enum `LinqToDB.Mapping.Relationship` conflicts with the generated class `Relationship`. I found these lines in the `linqtodb.ttinclude` file and changed them to use full name of the enum:

```c#
switch (key.AssociationType)
{
	case AssociationType.OneToOne  : aa.Parameters.Add("Relationship=LinqToDB.Mapping.Relationship.OneToOne");  break;
	case AssociationType.OneToMany : aa.Parameters.Add("Relationship=LinqToDB.Mapping.Relationship.OneToMany"); break;
	case AssociationType.ManyToOne : aa.Parameters.Add("Relationship=LinqToDB.Mapping.Relationship.ManyToOne"); break

}

```

However, it did not affect the generated code, it was still the same, with errors. 
Can anyone please suggest a workaround (**besides renaming my table**), or maybe a patch is needed? As a generic rule, generated code should always use fully qualified names.

Thanks.

**P.S.:** Ok I got it, I just referenced *.tt files from my project instead of from `/packages` directory, and the generation went alright. But still, can the generation be fixed to use full names?

### Environment details
linq2db version: *2.6.4*

### LINKED ISSUE #1825: How T4 generates multiple files

### LINKED ISSUE #1897: [Request, T4 template] Support for ParameterDirection.ReturnValue

### LINKED ISSUE #1971: osx, wrong path?
I have this issue on Osx with Visual studio.

`UserDb.tt(5,5): Error: Could not resolve include file '$(LinqToDBT4SqlServerTemplatesDirectory)LinqToDB.SqlServer.Tools.ttinclude'.`

Where does `LinqToDBT4SqlServerTemplatesDirectory` is coming from? I tried in feeding in the absolute path where nuget package is located but then these ttinclude files complain about missing this dll or that dll, that is all because of the path problems.

Is this a bug or?

### LINKED ISSUE #2168: Wrong T4 generation for ColumnAttribute when table name is "DataType"

### LINKED ISSUE #2515: Templates do not work with JetBrains Rider C# IDE (non-windows)
The postgresql package installs a bunch of TT templates that are not compatible with Rider as they take a dependency on Visual Studio assemblies.

linq2db version: 3.1.3
Database Server: postgresql 13
Database Provider: npgsql 5.0.0-alpha1
Operating system: ubuntu 20.04
.NET Framework: netcoreapp3.1

### LINKED ISSUE #2793: question T4 how to custom add FindQuery() without custom LinqToDB.ttinclude
**how can i custom it in _PBNorthwind.tt_ without custom _LinqToDB.ttinclude_?**

this my generated.cs
```c#
public partial class T_Table1
{
	[PrimaryKey(1), NotNull    ] public int       RowId   { get; set; } // integer
	[PrimaryKey(2), NotNull    ] public string    Code         { get; set; } // character varying(50)
	[Column,        NotNull    ] public string    VariableName { get; set; } // character varying(150)
	[Column,        NotNull    ] public short     DataType     { get; set; } // smallint
	[Column,           Nullable] public int?      IntValue     { get; set; } // integer
	[Column,           Nullable] public double?   DoubleValue  { get; set; } // double precision
	[Column,           Nullable] public byte[]    MoneyValue   { get; set; } // bytea
	[Column,           Nullable] public string    StringValue  { get; set; } // character varying
	[Column,           Nullable] public bool?     BoolValue    { get; set; } // boolean
	[Column,           Nullable] public DateTime? DateValue    { get; set; } // date
	[Column,           Nullable] public short?    CurrencyId   { get; set; } // smallint
}

public static T_Table1 Find(this ITable<T_Table1> table, int RowId, string Code)
{
	return table.FirstOrDefault(t =>
		t.RowId == RowId &&
		t.Code == Code);
}
```

i want to add this for every table
```c#
public static IQueryable<T_Table1> FindQuery(this ITable<T_Table1> table, T_Table1 row)
{
	return table.Where(t =>
		t.RowId == row.RowId &&
		t.Code == row.Code);
}
```
i managed to do this by add custom _LinqToDB.ttinclude_ and adding _MakeTypeMembersNamesUnique_
```c#
tableExtensions.Members.Add(
				new Method(
					() => string.Format("IQueryable<{0}>", new ModelType(t.TypeName, true, true).ToTypeName()),
					"FindQuery",
					new Func<string>[] { () => string.Format("this ITable<{0}> table", t.TypeName) }
						.Union(PKs.Select(c => (Func<string>)(() => t.TypeName + " row"))),
					() => new[] { "return table.Where(t =>" }
						.Union(PKs.SelectMany((c,i) =>
						{
							var ss = new List<string>();

							if (c.Conditional != null)
								ss.Add("#if " + c.Conditional);

							ss.Add(BuildColumnComparison2(c, LenDiff(maxNameLen1, c.MemberName), LenDiff(maxNameLen2, c.MemberName), i == nPKs - 1));

							if (c.Conditional != null)
							{
								if (ss[1].EndsWith(");"))
								{
									ss[1] = ss[1].Substring(0, ss[1].Length - 2);
									ss.Add("#endif");
									ss.Add("\t\t);");
								}
								else
								{
									ss.Add("#endif");
								}
							}

							return ss;
						})))
				{
					IsStatic = true
				});
```

purpose is because it didnt save track changes which column is updated

so if i do this, it will update all column
```c#
var get = db.T_Table1.FirstOrDefault(x => x.RowId == 1 && x.Code == "code1");
if (get != null)
{
	/* process data */
	get.IntValue = 5;
	
	/* Update but All Column */
	db.Update(get);
}
```

instead passing key again
```c#
	db.T_Table1.Where(x => x.RowId == 1 && x.Code == "code1").Set(x => x.IntValue, get.IntValue).Update();
```
i prefer use
```c#
	db.T_Table1.FindQuery(get).Set(x => x.IntValue, get.IntValue).Update();
```

so i can easy update the table i just retrieved without passing/knowing the primary key again

### Environment details
linq2db version: 3.2.3
Database: PostgreSQL 12.1
.NET Framework: 4.7.2

MaceWindu:
Fix #1383
Fix #1586 (proper fix instead of old hack)
Fix #1825
Fix #1897 (wasn't actually fixed before)
Fix #1971
Fix #2168 (proper fix instead of old hack)
Fix #2515
Fix #2793
Fix #3345

### Goal
Provide new code-generation library and command-line tool for database model scaffolding to be used by:
- LinqPAD integration
- users

Currently we use:
1. T4 templates
2. Copy of T4 code in LinqPAD integration

Issues with current approach:
- code duplication between T4 and LinqPAD, issues with changes sync between them
- T4 has very poor support outside of Visual Studio and unusable on most of 3rd party implementations due to lack of features (like MSBuild directives support)
- Issues with different target frameworks used to run T4: VS use .net framework, Rider use .net core
- poor support for T4 development, basically you work with text file
- current implementation contains of two parts with mixed responsibilities: schema API in linq2db with unrelated functionality like C# identifiers and types generation and T4 code that mix schema translation and code generation without responsibilities separation
- target codegeneration supports only one language (C#)
- a lot of known and unknown (e.g. discovered during development of current feature) issues with code generation, especially with code correctness of generated code (like invalid identifiers, naming conflicts)

### Architecture

Instead of semi-monolitic approach, here we try to separate code generation into several distinctive steps:
- database schema load in form of new schema API, provided by linq2db that will not include unrelated functionality and bring more flexibility to this API. **NOTE**: initial releases will not have this API and we will have wrapper over current schema API instead
- in-memory data model build, that includes generation of descriptors of data context, entities, stored procedure/function helpers with names, that will be used in code, data model metadata (that will be converted to code model on next step using different metadata generators, e.g. attribute-based or fluent mapping mappings generators)
- conversion of data model to code model (AST), normalization of AST (e.g. fixing naming according to target language naming rules and resolving duplicate name conflicts)
- conversion of code AST to target language. Initial releases will have C# code generator, but later we plan to add at least VB.NET and F# generators as languages that were requested by users before

### Tasks
- [x] packaging: decide on release names for codegen library and commandline tool
- [x] packaging: nuget packaging for codegeneration library
- [x] packaging: nuget packaging for commandline tool
- [x] packaging: deprecate T4 packages and stop automatic publishing on release
- [x] cli: parametrization using command line and configuration file (json)
- [ ] cli: T4 migration preset to use options, similar to T4 defaults to simplify migration to new tool for existing T4 users
- [x] api: finalize schema API
- [x] api: finalize data model build API
- [x] api: finalize code generation API
- [x] api: finalize C# code generator
- [ ] extensibility interfaces to extend data model generation on all steps
- [ ] cli: decide and implement which approach to use for scaffolding customization: no customization for cli, loadable C# file, loadable assembly?
- [ ] [**v.next**] VB.NET code generator
- [ ] [**v.next**] F# code generator. See #1553
- [ ] [**v.next**] TypeScript code generator. See #1233
- [ ] [**v.next**] Fluent mapping generator. See #945
- [ ] linqpad: migration to new library
- [ ] documentation
- [x] xml-docs in code
- [ ] tests
- [ ] [**v.next**] new schema api and deprecation of old API
- [ ] [**v.next**] new schema api: avoid table function/procedure execution to retrieve schema when supported by database (e.g. postgresql table functions)
- [ ] [**v.next**] new schema api: sequence name support. See #2766
- [x] [**v.next**] Generation of data model in multiple classes/file-per-class mode. See #1825
- [ ] option to configure which .net type to use for specific database type instead of current implementation where type is hardcoded
- [ ] more pre-defined customization options. E.g. see #1257
- [ ] migrate some existing non-db T4 templates: `Equatable.ttinclude`
- [ ] decide what to do with WPF-related T4 templates. Discard and unrelared to linq2db or release as separate T4 package?
- [ ] improve association properties/extensions naming. See #2331 (probably should be fixed on schema api side to return empty FK name for such Access keys)
- [x] new FindQuery extension generation. See #2793
- [ ] more built-in methods for association property/extension naming. See #2803
- [ ] [**v.next**] option to generate DTO classes. See #2850
- [ ] [**v.next**] option to generate mappings using C# records (requires fluent mapping support)
- [ ] verify that all applicable T4 logic replicated in new code

There are also a bunch of schema/T4 issues that could be addressed during current feature implmentation or later, which are not mentioned as they probably will be done separately.

---
# Implemented changes

#### Main feature
- implemented scaffolding framework in `linq2db.Tools` project:
   - `Source\LinqToDB.Tools\CodeModel`: code model and code generation (currently C#-only)
   - `Source\LinqToDB.Tools\DataModel`: cotext data model to code AST conversion
   - `Source\LinqToDB.Tools\Metadata`: context metadata model and generation (currently attribute-based only)
   - `Source\LinqToDB.Tools\Naming`: model pluralization and identifier normalization
   - `Source\LinqToDB.Tools\Scaffold`: database schema to context data model conversion
   - `Source\LinqToDB.Tools\Schema`: database schema API (currently wrapper over existing schema API)
- added scaffolding CLI tool 'Source\LinqToDB.CLI' that use new framework
- added cli tool packaging as [`dotnet tool`](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools). Planned nuget name: `linq2db.cli`, dotnet tool name: `dotnet-linq2db`. Target framework: `netcoreapp3.1` (with plan to switch to `net6` after 3.1 EOL)

#### Various scaffolding fixes/improvements
- [Scaffold] Partial fix #1468 - added new scaffold option to ignore schema load errors and generate procedure mappings without return table. Could be enough if procedure doesn't return table or user is not interested in it or plan to add table mapping manually.
- [CodeGeneration] Implement conflict detection for generated identifiers. Properly fix #2168, #1586
- [CodeGeneration] Implement generated C# identifier validation/normalization to produce valid identifier name
- [Scaffold] enable `@return` parameter support for SQL Server stored procedures. Fix #1897 (previous fix wasn't included into T4 by mistake)
- Fix #3345. Implemented `--architecture` cli tool switch to start it as x64 or x86 process. This is needed for unmanaged providers, especially Access. VS2019- always used x86 T4 host and VS2022 use x64 test host, which made it hard to use with such databases.
- [Access] New tool supports schema load for Access using two connections (OLE DB + ODBC). This allows us to load more precise database schema. Both providers return incomplete schema information and with this feature we combine information from both providers.
- [Sybase] set `ColumnType` for procedure schema in schema provider
- [Access] Update provider detection logic to not identify connections with `DataAccess` in name as Access connection (as `DataAccess` is a part of oracle managed provider name)
- [Scaffold] support generation of async stored procedure mappings
- [Scaffold] update `Find` extensions generation (Fix #2793)
  - `FindAsync` method generation
  - `FindQuery` method generation
  - generation of extensions on of context class (previously only on `ITable<T>`)
  - generation of extensions with entity object as parameter instead of separate parameter for each primary key column
  - (results in 12 variations of `Find` generation)
- [Scaffold] interceptors mechanism to cusomize scaffolding using T4 template or pre-built interceptor assembly

#### Other fixes
- add new connection interceptor method `AfterExecuteReader` to handle user-requested case with oracle reader configuration. Proposed workaround was unusable due to use of internal API
- obsoleted nonfunctional `AssociationAttribute` properties (`KeyName`, `BackReferenceName`, `IsBackReference`, `Relationship`)  and `Relationship` enum as they were never used
- obsoleted `DataConnection.OnTrace` static property
- replace use of `StringComparison.Invariant*` to `StringComparison.Ordinal*` enum values

#### Infrastructure changes
- Code sharing: shared code to enable recent C# features for older frameworks moved to shared folder `Source\Shared` from `LinqToDB` project and included to all projects (see `Build/linq2db.Default.props`)
- disabled nuget publishing for T4 templates. Current plan is to obsolete them and publish them only explicitly when we fix something.

###### `InternalsVisibleTo` removal
This attribute used by `linq2db` to open some internal stuff to tests which had several negative consequences:
- we had some API intended for use by users as marked internal (see below)
- [netfx-only] attribute requires strong name so we signed our tests assembly, which prevented it from references to unsigned assemblies. E.g. as `dotMorten.Microsoft.SqlServer.Types` is not signed - we used custom signed build for it instead of nuget

To solve it I've:
- removed attribute usages
- removed tests signing
- opened "internal" API, that became unavailable to tests

Opened API:
- `LinqToDB.Common.Compilation.CompileExpression` methods exposed for `linq2db.Tools` and tests, but still marked as internal API. It's needed mostly for `Tools` and we can remove use of it from tests
- `LinqToDB.Common.EnumerableHelper.Batch` methods exposed (were covered by tests)
- `InvariantCultureRegion` made public (for tests)
- `MappingSchema.CombineSchemas` method made public (for tests)
- `LinqExtensions.Updatable` made public (for tests)
- `Query.Queries` property, `Query<T>` class (for method `ClarCache`, property `CacheMissCount` and method `GetQuery`) and `QueryInfo` class made public for tests
- `InternalExtensions` class, methods `Unwrap`, `IsSameGenericMethod`, `EvaluateExpression`  (for tests)
- `TypeWrapperNameAttribute`, `ValueTaskToTaskMapper`, `CustomMapperAttribute` and `ICustomMapper` made public (should've been public from start)
- `ReflectionExtensions.IsSubClassOf` method (for `linq2db.Tools`)
- `SqlServerTransientExceptionDetector.IsHandled` method (for `linq2db.Tools`)
- `SqlServerTools.QuoteIdentifier` method (for tests)
- `PostgreSQLDataProvider.HasMacAddr8` property (for tests)
- `DataContext` methods `GetDataConnection`, `ReleaseQuery`, `ReleaseQueryAsync` (for tests)

I plan to review it later and probably hide some API back (requires tests refactoring). Also some tests are commented due to internal APIs - will address it too


MaceWindu:
If anyone interested in testing and giving feedback for current dev. build, he can try it using attached nuget (unzip first) [linq2db.cli.0.0.52.zip](https://github.com/linq2db/linq2db/files/7941906/linq2db.cli.0.0.52.zip)
Or for manual build:
- checkout PR branch
- build in release mode
- run `NuGet\TestToolLocal.cmd X` to build and install nuget, where x is a version number (increase it for each new build)

What is missing from this version:
- ability to customize generation using custom code (working on it)
- CLI options defaults not finalized yet
- T4 compatibility mode not configured yet (this is mode when tool use options defaults that generate code matching old T4 generation for more easy migration)

Check nuget readme for starters: https://github.com/linq2db/linq2db/blob/685a75093832e430e41411d72da64f5b40f62051/NuGet/CLI/README.md

MaceWindu:
[linq2db.cli.0.0.83.zip](https://github.com/linq2db/linq2db/files/8010496/linq2db.cli.0.0.83.zip)

- async stored procedures
- FindAsync and FindQuery methods
- `@return` parameter support for mssql procedures

MaceWindu:
- initial extensibility points: schema filtering and type mapper
- t4 and built assembly support as interceptors source

[linq2db.cli.0.0.111.zip](https://github.com/linq2db/linq2db/files/8055719/linq2db.cli.0.0.111.zip)

Source/LinqToDB/Async/AsyncEnumerableExtensions.cs
```diff
@@ -6,10 +6,9 @@
 namespace LinqToDB.Async
 {
 	/// <summary>
-	/// This API supports the LinqToDB infrastructure and is not intended to be used  directly from your code.
-	/// This API may change or be removed in future releases.
+	/// Contains extension methods for <see cref="IAsyncEnumerable{T}"/> interface.
 	/// </summary>
-	internal static class AsyncExtensions
+	public static class AsyncEnumerableExtensions
```
viceroypenguin, line 11: `public` will make the the `.ToListAsync()` method conflict with the one provided by msft (https://github.com/dotnet/reactive/blob/e878589c5889ed19b2fd51c8b18f200c5e497011/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToList.cs#L22). This would be problematic for users who include both.

viceroypenguin (commented):
Public comments from private discussion:

1. Use dotnet provided command line parser (https://github.com/dotnet/command-line-api)
2. Revert many of the `internal` to `public` changes and re-use the `[InternalsVisibleTo]` attributes.

NuGet/TestToolLocal.cmd
```diff
@@ -0,0 +1,16 @@
+ECHO OFF
+
+IF [%1] EQU [] (
+ECHO Version required
+GOTO :EOF
+)
+
+SET VERSION=0.0.%1
```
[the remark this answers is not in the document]
MaceWindu: that's version for local testing

NuGet/linq2db.Tools.nuspec
```diff
@@ -11,13 +11,16 @@
 		<readme>README.md</readme>
 		<dependencies>
 			<group targetFramework=".NETFramework4.5" >
-				<dependency id="linq2db" version="3.0.0" />
+				<dependency id="linq2db"        version="3.0.0"  />
```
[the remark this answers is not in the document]
MaceWindu: version value set during nugets generation. but we can bump it to 4

MaceWindu:
I don't plan to reinstate internalsvisibleto as it creates more harm, but will probably remove some strange tests that need internal access and revert visibility changes

Source/LinqToDB/Linq/Query.cs
```diff
@@ -31,7 +31,7 @@ public abstract class Query
 
 		#region Init
 
-		internal readonly List<QueryInfo> Queries = new (1);
+		public readonly List<QueryInfo> Queries = new (1);
```
viceroypenguin, line 34: Because `List<>` is itself mutable, even though this field is readonly, there is the possibility that someone could take this list and change it outside of our expectations (add `null` entries, remove entries, etc.). It should be an `IReadOnlyList<>` property if it is to be public, though we can have an internal `List<>` field the property references.

Source/LinqToDB.CLI/CommandLine/Commands/ScaffoldCommand.Execute.cs
```diff
@@ -0,0 +1,430 @@
...
+				MetadataBuilders.GetAttributeBasedMetadataBuilder(generator.Language, sqlBuilder),
+				SqlBoolEqualityConverter.Create(generator.Language));
+			var sourceCode = generator.GenerateSourceCode(files);
+
+			Directory.CreateDirectory(output);
+
+			for (var i = 0; i < sourceCode.Length; i++)
+			{
+				// TODO: add file name normalization/deduplication?
+				var fileName = $@"{output}\{sourceCode[i].FileName}";
```
jelical, line 149: `var fileName = $@"{output}{Path.DirectorySeparatorChar}{sourceCode[i].FileName}";`
Otherwise nix like oses will complain

sdanyliv: Or better `Path.Combine`

Source/LinqToDB.CLI/CLI/Commands/CliCommand.cs
```diff
@@ -0,0 +1,189 @@
...
+namespace LinqToDB.CLI
+{
+	/// <summary>
+	/// Base class for CLI commands.
+	/// </summary>
+	internal abstract class CliCommand
+	{
+		private readonly Dictionary<OptionCategory, List<CliOption>>           _optionsByCategory      = new ();
+		private readonly List<CliOption>                                       _optionsWithoutCategory = new ();
+		// TODO: replace with HashSet if not used as it also ensures command name/shortname is unique
```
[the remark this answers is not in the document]
MaceWindu: TODOs will be addressed one-by-one in later releases

PR #3106, opened 2021-07-08
add banned symbols support, ban threadstatic attribute

### LINKED ISSUE #3096: ThreadStatic field _suspended is used in async method
https://github.com/linq2db/linq2db/blob/89e0f7e34413120b4bffa278f9c4d12e3c885758/Source/LinqToDB/Data/RetryPolicy/RetryPolicyBase.cs#L186

```c#
[ThreadStatic]
static volatile bool _suspended;

protected static bool Suspended
{
	get => _suspended;
	set => _suspended = value;
}

//...
Suspended = true;
var result = await operation(cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext);
Suspended = false;
```
You cannot assume you will be in the same thread after `await` even if the `ContinueOnCapturedContext` was `true`.
Moreover the thread can be used by some parallel task while the async `operation` is working and that task can also call the `ExecuteAsync` causing unpredictable state of the Suspended property value.

MaceWindu:
Fix #3096

### DIFF
--- a/Build/BannedSymbols.txt
+++ b/Build/BannedSymbols.txt
@@ -1,0 +1,1 @@
+T:System.ThreadStaticAttribute;don't do it
--- a/Build/linq2db.Default.props
+++ b/Build/linq2db.Default.props
@@ -33,7 +33,7 @@
 
 		<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
 	</PropertyGroup>
-	
+
 	<PropertyGroup>
 		<AnalysisLevel>preview</AnalysisLevel>
 		<EnableNETAnalyzers>true</EnableNETAnalyzers>
@@ -43,4 +43,10 @@
 	<PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'netcoreapp3.1' OR '$(TargetFramework)' == 'net5.0'">
 		<DefineConstants>NETSTANDARD2_1PLUS;$(DefineConstants)</DefineConstants>
 	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" />
+		<AdditionalFiles Include="$(MSBuildThisFileDirectory)/BannedSymbols.txt" />
+	</ItemGroup>
+
 </Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -54,6 +54,7 @@
 		<PackageVersion Include="FSharp.Core" Version="5.0.1" />
 		<PackageVersion Include="Microsoft.AspNet.OData" Version="7.5.8" />
 		<PackageVersion Include="NodaTime" Version="3.0.5" />
+		<PackageVersion Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.2" />
 
 		<!--nuget doesn't have strong name, so we use local self-signed copy-->
 		<!--<PackageVersion Include="dotMorten.Microsoft.SqlServer.Types" Version="1.3.0" />-->
--- a/Source/LinqToDB/Data/RetryPolicy/RetryPolicyBase.cs
+++ b/Source/LinqToDB/Data/RetryPolicy/RetryPolicyBase.cs
@@ -52,8 +52,7 @@ namespace LinqToDB.Data.RetryPolicy
 		/// </summary>
 		protected virtual TimeSpan MaxRetryDelay { get; }
 
-		[ThreadStatic]
-		static volatile bool _suspended;
+		static readonly AsyncLocal<bool> _suspended = new ();
 
 		/// <summary>
 		/// Indicates whether the strategy is suspended. The strategy is typically suspending while executing to avoid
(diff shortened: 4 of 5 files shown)


viceroypenguin (approved):

artelk:
Unfortunately the change in the RetryPolicyBase seems wrong and we cannot just use AsyncLocal.
```c#
static readonly AsyncLocal<bool> _suspended = new();

static bool Suspended
{
    get => _suspended.Value;
    set => _suspended.Value = value;
}

static async Task ExecuteAsync(Func<CancellationToken, Task> operation, CancellationToken cancellationToken = new CancellationToken())
{
	if (Suspended)
	{
		await operation(cancellationToken).ConfigureAwait(false);
		return;
	}

	await ExecuteImplementationAsync(async ct => { await operation(ct).ConfigureAwait(false); return 0; }, cancellationToken).ConfigureAwait(false);
}

async static Task<TResult> ExecuteImplementationAsync<TResult>(
	Func<CancellationToken,
	Task<TResult>> operation,
	CancellationToken cancellationToken)
{
	//...
	Suspended = true;
	isStarted = true; // <-- added
	var result = await operation(cancellationToken).ConfigureAwait(false);
	Suspended = false;
	return result;
	//...
}

static bool isStarted = false;

static async Task Main(string[] args)
{
	var task = ExecuteAsync(_ => Task.Delay(5000));
	await Task.Delay(100);
	Console.WriteLine(isStarted); // true
	Console.WriteLine(Suspended); // false <------- !
}
```
I suspect it wasn't expected to get Suspend==false in the beginning of ExecuteAsync while the operation is running in parallel.

viceroypenguin:
@artelk  - I believe the behavior works as expected. Regarding the behavior of `AsyncLocal<>`, every async context will inherit the value from it's parent, until the value changes. So `Main()` and `ExecuteImplementationAsync()` will share a value until the value gets set in `ExecuteImplementationAsync()`. Then the two values diverge - `Main()` keeps it's value while `ExecuteImplementationAsync()` gets the new value. Any children of `ExecuteImplementationAsync()` (function calls from there) will inherit the new value unless they in turn set their own value.

Updating the task as follows:
```csharp
var task = ExecuteAsync(async _ => { Suspended.Dump(); await Task.Delay(5000); Suspended.Dump(); });
```
will produce two `True` values and one `False` value. (the two true values being from inside the task, the false being from `Main()`). 

Given the above behavior, `AsyncLocal<>` will do exactly what we need, which is to inform us whether we are in a distantly recursive call. `Suspended` will be true if and only if `ExecuteImplementationAsync()` is in the async function call chain before a second call to `ExecuteAsync()`.

MaceWindu:
Also don't see anything wrong with implementation. In your example you see expected side effect of how exectution context flows between threads.

But in retry policy we use it in a way where context copying doesn't affect us.

Actually to avoid those side-effects in my projects I use following trick - I store async data not in context directly but in dictionary and put this dictionary to execution context on top-most level so it flows successfully to underlying async calls.

artelk:
Interesting. Do you mean the Suspend is needed if the operation reenters the Execute/ExecuteAsync method? Yes, in that case the nested Execute/ExecuteAsync won't enter the retry loop.
(I've just read the comment on Suspend property, you are right 😄)
I thought the Suspend was needed for things like this:
```
var parallelTask1 = ExecuteAsync(...);
Execute(...); // or parallelTask2 = ExecuteAsync(...);
```
With current implementation the Execute (or the second ExecuteAsync) will also find Suspend==false. So it will Clear the (shared) ExceptionsEncountered List (not thread safe) and also enter the while loop.

Also the Suspend value is switched to true and false _inside_ the while loop of ExecuteImplementationAsync. The TaskEx.Delay is called when Suspend is false. I supposed that was made intentionally so some parallel tasks can be somehow allowed to enter the while loop while we are waiting for the TaskEx.Delay.

Ok, proposal:
AsyncLocal can be slow in case there are many other AsyncLocals in the execution context. I think it would be better to at least change the Suspend value outside the loop. The ExceptionsEncountered should probably also be AsyncLocal for tread safety and to not mix exceptions from multiple parallel tasks.
I mean something like this:
```c#
static readonly AsyncLocal<bool> _suspended = new ();

protected static bool Suspended
{
  get => _suspended.Value;
  set => _suspended.Value = value;
}

// also changed to AsyncLocal. I would even put both things to a single class to have only one AsyncLocal.
static readonly AsyncLocal<List<Exception>> _exceptionsEncountered = new ();

protected List<Exception> ExceptionsEncountered => _exceptionsEncountered.Value;

public virtual Task<TResult> ExecuteAsync<TResult>(Func<CancellationToken, Task<TResult>> operation,
  CancellationToken cancellationToken = default(CancellationToken))
{
  if (Suspended)
    return operation(cancellationToken);

  var exceptionsEncountered = new List<Exception>();
  exceptionsEncountered.Value = exceptionsEncountered; // touch AsyncLocal only once, pass the list as a parameter
  return ExecuteImplementationAsync(operation, exceptionsEncountered, cancellationToken);
}

// TODO: ExecuteAsync(Func<CancellationToken, Task>) should also be synchronous, with no awaits

async Task<TResult> ExecuteImplementationAsync<TResult>(
  Func<CancellationToken, Task<TResult>> operation,
  List<Exception> exceptionsEncountered, // <-- added
  CancellationToken cancellationToken)
{
  Suspended = true;
  try 
  {
    while (true)
    {

      try
      {
        return await operation(cancellationToken);
      }
      catch (Exception ex)
      {
        //...
        exceptionsEncountered.Add(ex); // don't touch AsyncLocal
        //...
      }

      await TaskEx.Delay()
    }
  }
  finally
  {
    Suspended = false;
  }
}
```

PR #3144, opened 2021-07-31
Don't generate options constructors for context based on DataContext

### LINKED ISSUE #3138: Ctor with LinqToDbConnectionOptions parameter is invalid
When generating the classes for the DB access it creates invalid code when running with specific settings.
Error ocoured with a MySql database, but I think the error is the some also for other databases.

### Steps to reproduce
1. Set `BaseDataContextClass = "LinqToDB.DataContext";` in your Generator.MySql.tt file
2. Save file to generate classes
3. You'll find the following code:
```c#
public partial class MyClassName : LinqToDB.DataContext
{
	// ...
	public MyClassName(LinqToDbConnectionOptions options)
		: base(options) // <-- Error here
	{
		InitDataContext();
		InitMappingSchema();
	}
	// ...
}
```

The parameter options in `base(options)` shows the following error:
CS1503 C# Argument 1: cannot convert from 'LinqToDbConnectionOptions' to 'string?'

### Environment details
linq2db version: 3.4.2
Database Provider: MySQL
Operating system: Windows 10
.NET Framework: 4.5.2

MaceWindu:
fix #3138

- check that base class is not DataContext (will be removed after merge to v4 as there we support such constructors)
- add explicit option to disable options constructors (`GenerateLinqToDbConnectionOptionsConstructors = true`)

### DIFF
--- a/Source/LinqToDB.Templates/DataModel.ttinclude
+++ b/Source/LinqToDB.Templates/DataModel.ttinclude
@@ -21,7 +21,7 @@ string   NamespaceName
 public string   ServerName;
 public string   DatabaseName;
 public string   DataContextName;
-public string   BaseDataContextClass;
+public static string BaseDataContextClass;
 public string   BaseEntityClass;
 public string   OneToManyAssociationType      = "IEnumerable<{0}>";
 
--- a/Source/LinqToDB.Templates/DataModel.ttinclude
+++ b/Source/LinqToDB.Templates/DataModel.ttinclude
@@ -18,10 +18,12 @@ string   NamespaceName
 	set { Model.Namespace.Name = value; }
 }
 
+public static string BaseDataContextClass;
+public static bool   EnforceModelNullability = true;
+
 public string   ServerName;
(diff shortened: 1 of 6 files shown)


sdanyliv (approved):

Source/LinqToDB.Templates/DataModel.ttinclude
```diff
@@ -21,7 +21,7 @@ string   NamespaceName
 public string   ServerName;
 public string   DatabaseName;
 public string   DataContextName;
-public string   BaseDataContextClass;
+public static string BaseDataContextClass;
```
viceroypenguin, line 24: Given the penchant for vertical alignment, why is this set of items not vertically aligned to the longer type?

viceroypenguin (approved):

Source/LinqToDB.Templates/DataModel.ttinclude
MaceWindu: :crying_cat_face: mkay, will fix it

PR #3147, opened 2021-08-02
Allow to use private and protected parameterized constructor for mapping

ImoutoChan:
This PR adds an ability for linq2db to map data to classes with only private or protected constructors. Before this PR only public ones were supported. 

This case is useful when you want to map data to encapsulated rich models, which often have closed constructors.

We can also consider smartly selecting which constructor to use (most suitable based on field names). Right now it will only work for classes with single constructor.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
+++ b/Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
@@ -527,19 +527,28 @@ namespace LinqToDB.Linq.Builder
 				}
 			}
 
-
-			Expression BuildFromParametrizedConstructor(Type objectType,
-				IList<(string Name, Expression? Expr)> expressions)
+			ConstructorInfo SelectParametrizedConstructor(Type objectType)
 			{
-				var constructors = objectType.GetConstructors(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
+				var constructors = objectType.GetConstructors();
 
 				if (constructors.Length == 0)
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3147) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3147) before we can accept your contribution.<br/><br/>:x: ImoutoChan [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3147)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=3147) it.</sub>

ImoutoChan:
What do you think? Can it be merged?

Or did I break something? I don't see which test is broken.

viceroypenguin (approved):

MaceWindu:
Thanks!

It's failed because we don't support db-specific tests on external pr, so I need to run all tests.

Regarding implementation, I think it should be changed a bit - first probe public constructors and if there is not match then try non-public. Otherwise it could be a regression for existing types with public and non-public constructors

MaceWindu (changes requested):
see my comment above

ImoutoChan:
Thank you for your feedback!

I've changed implementation as suggested to try pubic constructor first. I've added one more test to check this specific case. The failed test should also be fixed now.

sdanyliv (approved):

MaceWindu (approved):

Issue #3161, opened 2021-08-07
Nested subquery does not execute


Shane32:
Executing a outer apply within another outer apply can fail in recent versions of linq2db:

```cs
var ret = db.GetTable<Table1>()
	.Select(t1 => new
	{
		//retrieve Table1.Name1
		Name1 = t1.Name1,
		Value1 = db.GetTable<Table2>()
			.Where(x => x.ParentId2 == t1.Id1)
			.Select(t2 => new
			{
                                //retrieve Table2.Name2 via outer apply
				Name2 = t2.Name2,
				Value2 = db.GetTable<Table3>()
					.Where(x => x.ParentId3 == t2.Id2)
					.Select(t3 => new
					{
                                                //retrieve Table3.Name3 and Table3.Id3 via outer apply
						Name3 = t3.Name3,
						Value3 = t3.Id3
					})
					.FirstOrDefault()
			})
			.FirstOrDefault()
	})
	.ToList();
```

While a single cross apply between Table1 and Table2 works fine, adding the cross apply for Table3 breaks the sample.  For instance, the below code works fine:

```cs
var ret = db.GetTable<Table1>()
	.Select(t1 => new
	{
                //retrieve Table1.Name1
		Name1 = t1.Name1,
		Value1 = db.GetTable<Table2>()
			.Where(x => x.ParentId2 == t1.Id1)
			.Select(t2 => new
			{
                                //retrieve Table2.Name2 and Table2.Id2 via outer apply
				Name2 = t2.Name2,
				Value2 = t2.Id2
			})
			.FirstOrDefault()
	})
	.ToList();
```

The exception is the following:

```
System.Data.SqlClient.SqlException : Only one expression can be specified in the select list
when the subquery is not introduced with EXISTS.
```

The faulty code is producing invalid SQL as follows:

```sql
DECLARE @take Int -- Int32
SET     @take = 1
DECLARE @take_1 Int -- Int32
SET     @take_1 = 1

SELECT
	[t1_1].[NAME1],
	[t2].[Name2],
	[t2].[Name3],
	[t2].[Value3]
FROM
	[TABLE1] [t1_1]
		OUTER APPLY (
			SELECT TOP (@take)
				[x_1].[NAME2] as [Name2],
				/**** This portion of the SQL statment should not be 
				      present, and the query would work without it *****/
				(
					SELECT TOP (@take_1)
						[x].[NAME3],
						[x].[ID3]
					FROM
						[TABLE3] [x]
[...]

Issue #3160, opened 2021-08-07
Regression: Nested dictionaries fail in 3.4.0+ (after #3021)


Shane32:
This query fails for all builds after PR #3021 :

```cs
var ret = db.GetTable<Table1>()
	.Select(t1 => new Dictionary<string, object?>()
	{
		{
			"t1",
			db.GetTable<Table2>()
				.Where(x => x.ParentId == t1.Id)
				.Select(t2 => new Dictionary<string, object>()
				{
					{
						"t2",
						db.GetTable<Table3>()
							.Where(x => x.ParentId == t2.Id)
							.Select(t3 => t3.Id)
							.FirstOrDefault()
					}
				})
				.FirstOrDefault()
		}
	})
	.ToList();
```

The exception is:

```
The number of generic arguments provided doesn't equal the arity of the generic type definition. (Parameter 'instantiation')
```

The code works correctly in linq2db 3.3.0

### Environment details
linq2db version: 3.4.0 and newer including 4.0.0 beta
Database Server: SQL Server
Operating system: Windows 10
.NET Framework: .NET 5

sdanyliv:
@Shane32, could you please post model used in this test?

Shane32:
See PR #3162

Shane32:
Branch issues/3160

sdanyliv:
@Shane32, it is really worked before?

sdanyliv:
I'll post workaround before I find time to resolve this issue:
```cs
var query = 
    from t1 in db.GetTable<Table1>()
    from t2 in db.GetTable<Table2>().Where(x => x.ParentId == t1.Id).Take(1).DefaultIfEmpty()
    from t3 in db.GetTable<Table3>().Where(x => x.ParentId == t2.Id).Take(1).DefaultIfEmpty()
    select new Dictionary<string, object?>()
    {
        {
            "t1",
            new Dictionary<string, object>()
                {
                    {
                        "t2",
                        t3.Id
                    }
                }
        }
    };

var ret = query.ToList();
```

Shane32:
After #3028 merged, before #3021 was merged, the tests within #3162 run as follows:

![image](https://user-images.githubusercontent.com/6377684/128760954-8caa0c5f-7a1e-4e6d-b708-0050c552bfc7.png)

My specific use case is most similar to `NestedDictionaryTest3CrossApply`.

After #3021 was merged (the next commit in the history of `master`), the tests within #3162 run as follows:

[...]

PR #3162, opened 2021-08-07
Test case for #3160 and #3161 (currently failing)

Shane32:
See issues:
- #3160
- #3161

### DIFF
--- a/Tests/Linq/UserTests/Issue3160Tests.cs
+++ b/Tests/Linq/UserTests/Issue3160Tests.cs
@@ -316,26 +316,38 @@ namespace Tests.UserTests
 				.Select(t1 => new Dictionary<string, object?>()
 				{
 					{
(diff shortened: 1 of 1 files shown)


Shane32:
I added a bunch of scenarios here.  It's a puzzling combination.  Some of the issues do not appear without the third level of cross apply / subqueries.  And some of the issues that failed in 3.3.0 now fail but with a different message with 3.4.0+.

I'm hoping you can take a look @sdanyliv if you have a chance.  My entire e-commerce application is based on dynamically generated queries which use nested dictionaries and cross applies as demonstrated in these tests, so I'm stuck at 3.3.0 until the regression in #3160 is patched.  I am not seeing bug #3161 in my production code, but noticed it while tracking down #3160.

Shane32:
Replaced by #3391

PR #3163, opened 2021-08-07
Extended readme with more details about mapping and transaction scope

Usaga:
Extended readme with more details about mapping and transaction scope

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,14 @@ From **NuGet**:
 
 ## Configuring connection strings
 
+### Passing Into Constructor
+
+You can simply pass provider name and connection string into `DataConnection` constructor:
+
+```cs
+var db = new LinqToDB.Data.DataConnection(LinqToDB.ProviderName.SqlServer2012,
+   "Server=.\;Database=Northwind;Trusted_Connection=True;Enlist=False;");
+```
 
 ### Using Connection Options Builder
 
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3163) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3163) before we can accept your contribution.<br/><br/>:x: Usaga [sign now](https://cla.dotnetfoundation.org/linq2db/linq2db?pullRequest=3163)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/linq2db/linq2db?pullRequest=3163) it.</sub>

MaceWindu (changes requested):
I'm not sure it is correct to use `Code First` here. While we provide table creation API, we don't have database creation logic and it doesn't have anything with fluent mapping.

Usaga:
I am not sure I understood you right. Fluent Mapping is another way configuring mapping. More flexible. Why you think it related to database creation? We used fluent in our project for configuriong. I is working.

MaceWindu:
Yes, but you mention it as `Fluent configuration (also known as Code First)`.

I had an impression that `code first` term is about creating database from model, but after looking around it looks like one of those buzz-words where everybody has own understanding what it means....

MaceWindu (approved):

MaceWindu:
I would preferer to not use vague terms, but if microsoft use it in ef.core docs, we probably could expect people to understand it in MS way :)

Usaga:
So, I need to replace all "Code First" to "Fluent"?

MaceWindu:
@sdanyliv what you think? I would prefer to avoid `code first` term in this context

sdanyliv:
Code First usually meaning that database is created from code. Better do not use this term with linq2db.

jogibear9988:
also here: https://entityframeworkcore.com/approach-code-first

If you have used Entity Framework in the past, you will probably have heard about the Code First approach.

Code First workflow begins with classes that describe the conceptual model and then Entity Framework generate a database from that model automatically.

jogibear9988:
Microsoft also calls it "Fluent API"
https://docs.microsoft.com/en-us/ef/core/modeling/

PR #3166, opened 2021-08-08
Extended readme with more details about mapping and transaction scope

Usaga:
Extended readme with more details about mapping and transaction scope. Changed "Code First" with more appropriate "Fluent".

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,15 +187,15 @@ public class Product
 }
 ```
 
-There is two other ways to configure POCOs: no configuration at all and using Fluent configuration (also known as Code First).
+There is two other ways to configure POCOs: no configuration at all and using Fluent configuration.
 
 The first one involves no attributes. In this case Linq2Db will use POCO's name as table name and property names as column names. This might seem to be convenient, but there some moments: Linq2Db will not infer primary key even if class has property called "ID", and it will not infer nullability of string properties as there is no way to do so. Also, there will be no associations configured. In most cases the approach is pretty useless.
 
 The second way lets you configure your mapping dynamically, at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through class `MappingSchema`. There is static property `LinqToDB.Mapping.MappingSchema.Default` which might be used to one-time, global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string. 
 
-Code First approach lets you to use all configuration abilities available with attributes. These two approaches are interchangeable in its abilities.
+Fluent approach lets you to use all configuration abilities available with attributes. These two approaches are interchangeable in its abilities.
 
-There one thing which differs these approaches. If you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored. With Code First approach you can configure only thing that require it explicitly. All other properties will be inferred by Linq2Db:
+There one thing which differs these approaches. If you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored. With Fluent approach you can configure only thing that require it explicitly. All other properties will be inferred by Linq2Db:
 
 ```c#
 var builder = MappingSchema.Default.GetFluentMappingBuilder();
--- a/README.md
+++ b/README.md
@@ -75,8 +75,9 @@ From **NuGet**:
 You can simply pass provider name and connection string into `DataConnection` constructor:
 
 ```cs
-var db = new LinqToDB.Data.DataConnection(LinqToDB.ProviderName.SqlServer2012,
-   "Server=.\;Database=Northwind;Trusted_Connection=True;Enlist=False;");
+var db = new LinqToDB.Data.DataConnection(
+  LinqToDB.ProviderName.SqlServer2012,
+  "Server=.\;Database=Northwind;Trusted_Connection=True;Enlist=False;");
 ```
 
 ### Using Connection Options Builder
@@ -168,35 +169,9 @@ See [article](https://linq2db.github.io/articles/get-started/asp-dotnet-core/ind
 
 ## Now let's create a **POCO** class
 
-You can generate those classes from your database using [T4 templates](https://linq2db.github.io/articles/T4.html). Demonstration video could be found [here](https://linq2db.github.io/articles/general/Video.html). Or you can write them manually, using `Fluent configuration`, `Attribute configuration`, or inferring.
+You can generate POCO classes from your database using [T4 templates](https://linq2db.github.io/articles/T4.html).  These classes will be generated using the `Attribute configuration`. Demonstration video could be found [here](https://linq2db.github.io/articles/general/Video.html). 
 
-### Fluent Configuration
-
-This method lets you configure your mapping dynamically at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through the class `MappingSchema`. 
-
-With Fluent approach you can configure only thing that require it explicitly. All other properties will be inferred by linq2db:
-
-```c#
-var builder = MappingSchema.Default.GetFluentMappingBuilder();
-
-builder.Entity<Product>()
-    .HasTableName("Products")
-    .HasSchemaName("dbo")
-    .HasIdentity(x => x.ProductID)
-    .HasPrimaryKey(x => x.ProductID)
-    .Ignore(x => x.SomeNonDbProperty)
-    .Property(x => x.TimeStamp)
-        .HasSkipOnInsert()
-        .HasSkipOnUpdate()
-    .Association(x => x.Vendor, x => x.VendorID, x => x.VendorID, canBeNull: false)
-    ;
-
-//... other mapping configurations
-```
-
-In this example we configured only three properties and one association. We let linq2db to infer all other properties which have to match with column names. However, other associations will not get configured automatically.
-
-There is a static property `LinqToDB.Mapping.MappingSchema.Default` which may be used to define a global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string. 
+Alternatively, you can write them manually, using `Attribute configuration`, `Fluent configuration`, or inferring.
 
 ### Attribute configuration
 
@@ -223,7 +198,7 @@ public class Product
 }
 ```
 
-This approach involves attributes on all properties that should be mapped. This way lets you to configure all possible things linq2db ever supports. You will get all configuration abilities available with fluent configuration. These two approaches are interchangeable in its abilities. There one thing to mention: if you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored:
+This approach involves attributes on all properties that should be mapped. This way lets you to configure all possible things linq2db ever supports. There one thing to mention: if you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored:
 
 ```c#
 using System;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -179,6 +187,36 @@ public class Product
 }
 ```
 
+There is two other ways to configure POCOs: no configuration at all and using Fluent configuration.
```
MaceWindu, line 190:
```suggestion
There are two other ways to configure POCOs: no configuration at all and using Fluent configuration.
```

README.md
```diff
@@ -179,6 +187,36 @@ public class Product
...
+
+The first one involves no attributes. In this case Linq2Db will use POCO's name as table name and property names as column names. This might seem to be convenient, but there some moments: Linq2Db will not infer primary key even if class has property called "ID", and it will not infer nullability of string properties as there is no way to do so. Also, there will be no associations configured. In most cases the approach is pretty useless.
```
MaceWindu, line 192:
```suggestion
The first one involves no attributes. In this case linq2db will use POCO's name as table name and property names as column names (with exact same casing, which could be important for case-sensitive databases). This might seem to be convenient, but there are some moments: linq2db will not infer primary key even if class has property called "ID", and it will not infer nullability of string properties as there is no way to do so. Also, there will be no associations configured. In most cases the approach is pretty useless.
```

README.md
```diff
@@ -179,6 +187,36 @@ public class Product
...
+
+The second way lets you configure your mapping dynamically, at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through class `MappingSchema`. There is static property `LinqToDB.Mapping.MappingSchema.Default` which might be used to one-time, global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string.
```
MaceWindu, line 194:
```suggestion
The second way lets you configure your mapping dynamically at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through class `MappingSchema`. There is a static property `LinqToDB.Mapping.MappingSchema.Default` which might be used to one-time, global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string. 
```

README.md
```diff
@@ -179,6 +187,36 @@ public class Product
...
+
+The second way lets you configure your mapping dynamically, at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through class `MappingSchema`. There is static property `LinqToDB.Mapping.MappingSchema.Default` which might be used to one-time, global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string. 
+
+Fluent approach lets you to use all configuration abilities available with attributes. These two approaches are interchangeable in its abilities.
+
+There one thing which differs these approaches. If you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored. With Fluent approach you can configure only thing that require it explicitly. All other properties will be inferred by Linq2Db:
```
MaceWindu, line 198:
```suggestion
There is one thing which differs with these approaches. If you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored. With Fluent approach you can configure only thing that require it explicitly. All other properties will be inferred by linq2db:
```

README.md
```diff
@@ -179,6 +187,36 @@ public class Product
...
+    .Property(x => x.TimeStamp)
+        .HasSkipOnInsert()
+        .HasSkipOnUpdate()
+    .Association(x => x.Vendor, x => x.VendorId, x => x.Id, canBeNull: false)
+    ;
+
+//... other mapping configurations
+```
+
+In this example we configured only three properties and one association. We let Linq2Db to infer all other properties which have to match with column names.
```
MaceWindu, line 218:
```suggestion
In this example we configured only three properties and one association. We let linq2db to infer all other properties which have to match with column names.
```

README.md
```diff
@@ -555,6 +593,42 @@ using (var transaction = new TransactionScope())
 }
 ```
 
+It should be noted, than there are two base classes for your "context" class: `LinqToDB.Data.DataConnection` and `LinqToDB.DataContext`. The key difference between them is in connection retention behaviour. `DataConnection` opens connection with first query and holds it open until dispose happens. `DataContext` behaves the way you might used to with Entity Framework: it opens connection per query and closes it right after query is done.
```
MaceWindu, line 596:
```suggestion
It should be noted, that there are two base classes for your "context" class: `LinqToDB.Data.DataConnection` and `LinqToDB.DataContext`. The key difference between them is in connection retention behaviour. `DataConnection` opens connection with first query and holds it open until dispose happens. `DataContext` behaves the way you might used to with Entity Framework: it opens connection per query and closes it right after query is done.
```

README.md
```diff
@@ -555,6 +593,42 @@ using (var transaction = new TransactionScope())
...
+db.Update(product);
+
+scope.Dispose();
+
+// no transaction rollback happed, "Lollipop" has been saved
+```
+
+Connection is attached with ambient transaction in moment it opened. `TransactionScope` created after this moment have no effect on this connection. Changing `DataConnection` with `DataContext` in code shown earlier will make transaction scope work as expected: no record will be updated.
+
+Althou, `DataContext` seems to be the right class to choose, it strongly recommended to use `DataConnection` instead. It's behaviour might be corrected with setting `CloseAfterUse` property to `true`:
```
MaceWindu, line 620:
```suggestion
Althou, `DataContext` seems to be the right class to choose, it is strongly recommended to use `DataConnection` instead. It's default behaviour might be changed with setting `CloseAfterUse` property to `true`:
```

MaceWindu (approved):
- fixed some spelling errors
- added additional information about default mapping behavior
- lowercased `linq2db` (we already have `linq2db` vs `Linq To DB`, no need to have even more spellings :smile: )

@viceroypenguin could you please proofread it as a native speaker?

README.md
viceroypenguin:
```suggestion
There are two other ways to configure POCOs: no configuration at all, and using Fluent configuration.
```

README.md
```diff
@@ -70,6 +70,14 @@ From **NuGet**:
 
 ## Configuring connection strings
 
+### Passing Into Constructor
+
+You can simply pass provider name and connection string into `DataConnection` constructor:
+
+```cs
+var db = new LinqToDB.Data.DataConnection(LinqToDB.ProviderName.SqlServer2012,
+   "Server=.\;Database=Northwind;Trusted_Connection=True;Enlist=False;");
```
viceroypenguin, line 79:
```suggestion
var db = new LinqToDB.Data.DataConnection(
   LinqToDB.ProviderName.SqlServer2012,
   "Server=.\;Database=Northwind;Trusted_Connection=True;Enlist=False;");
```

README.md
viceroypenguin: Also, I would like to see `###` sub-headings for `Attribute configuration` above, `No configuration`, and `Fluent configuration`. Plus, we should add some commentary above the example above. The current comment describes how to do it via T4, but not that it can be done manually.

README.md
viceroypenguin: This being the second is because the first is `Attribute configuration` above. 
```suggestion
### Inferred configuration

The second method involves no attributes. In this case linq2db will use the class name as the table name and property names as column names (with exact same casing, which could be important for case-sensitive databases). This may be convenient, but there are some restrictions: linq2db will not infer primary key even if class has property called "ID"; it will not infer nullability of string properties as there is no way to do so; and associations will not be automatically configured. This approach is not generally recommended.
```

viceroypenguin: Actually, I think this should be the third method, due to it's lack of recommendation. Focus on fluent first, and then inferred configuration.

README.md
viceroypenguin:
```suggestion
### Fluent configuration

This method lets you configure your mapping dynamically at runtime. Furthermore, it lets you to have several different configurations if you need so. This kind of configuration is done through the class `MappingSchema`. 
```

viceroypenguin: The following text should be noted at the end of this section, not the beginning. 

```
There is a static property `LinqToDB.Mapping.MappingSchema.Default` which may be used to define a global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string. 
```

README.md
viceroypenguin:
```suggestion
There is one thing that differs with these approaches: if you add at least one attribute into POCO, all other properties should also have attributes, otherwise they will be ignored. With Fluent approach you can configure only thing that require it explicitly. All other properties will be inferred by linq2db:
```

README.md
viceroypenguin:
```suggestion
It should be noted that there are two base classes for your "context" class: `LinqToDB.Data.DataConnection` and `LinqToDB.DataContext`. The key difference between them is in connection retention behaviour. `DataConnection` opens connection with first query and holds it open until dispose happens. `DataContext` behaves the way you might used to with Entity Framework: it opens connection per query and closes it right after query is done.
```

README.md
```diff
@@ -555,6 +593,42 @@ using (var transaction = new TransactionScope())
 }
 ```
 
+It should be noted, than there are two base classes for your "context" class: `LinqToDB.Data.DataConnection` and `LinqToDB.DataContext`. The key difference between them is in connection retention behaviour. `DataConnection` opens connection with first query and holds it open until dispose happens. `DataContext` behaves the way you might used to with Entity Framework: it opens connection per query and closes it right after query is done.
+
+This difference in behavior matters when used with TransactionScope:
```
viceroypenguin, line 598:
```suggestion
This difference in behavior matters when used with `TransactionScope`:
```

README.md
```diff
@@ -555,6 +593,42 @@ using (var transaction = new TransactionScope())
...
+
+product.Name = "Lollipop";
+db.Update(product);
+
+scope.Dispose();
+
+// no transaction rollback happed, "Lollipop" has been saved
+```
+
+Connection is attached with ambient transaction in moment it opened. `TransactionScope` created after this moment have no effect on this connection. Changing `DataConnection` with `DataContext` in code shown earlier will make transaction scope work as expected: no record will be updated.
```
viceroypenguin, line 618:
```suggestion
A `DataConnection` is attached with ambient transaction in moment it is opened. Any `TransactionScope`s created after the connection is created will no effect on that connection. Replacing `DataConnection` with `DataContext` in code shown earlier will make transaction scope work as expected: the created record will be discarded with the transaction.
```

README.md
viceroypenguin:
```suggestion
Although, `DataContext` appears to be the right class to choose, it is strongly recommended to use `DataConnection` instead. It's default behaviour might be changed with setting `CloseAfterUse` property to `true`:
```

viceroypenguin (changes requested):
I'd like to see the changes regarding the three methods of definition before doing final review and approving this update.

Usaga:
I fixed all issues. Should I create new PR all this one will do?

viceroypenguin:
@Usaga This PR will do fine. It updates automatically when you push to your branch.

Given that the T4 produces `Attribute configuration`, I'd say that should be the first method, and `Fluent` should be second. Also, please update the introducing statement to mention attribute configuration:

```
You can generate POCO classes from your database using [T4 templates](https://linq2db.github.io/articles/T4.html).  These classes will be generated using the `Attribute configuration`. Demonstration video could be found [here](https://linq2db.github.io/articles/general/Video.html). 

Alternatively, you can write them manually, using `Attribute configuration`, `Fluent configuration`, or inferring.
```

Usaga:
Done.

README.md
```diff
@@ -175,10 +188,90 @@ public class Product
...
+//... other mapping configurations
+```
+
+In this example we configured only three properties and one association. We let linq2db to infer all other properties which have to match with column names. However, other associations will not get configured automatically.
+
+There is a static property `LinqToDB.Mapping.MappingSchema.Default` which may be used to define a global configuration. This mapping is used by default if no mapping schema provided explicitly. The other way is to pass instance of `MappingSchema` into constructor alongside with connection string.
+
+### Inferred Configuration
+
+This approach involves no attributes at all. In this case linq2db will use POCO's name as table name and property names as column names (with exact same casing, which could be important for case-sensitive databases). This might seem to be convenient, but there are some moments: linq2db will not infer primary key even if class has property called "ID"; it will not infer nullability of string properties as there is no way to do so; and associations will not be automatically configured.
```
viceroypenguin, line 249:
```suggestion
This approach involves no attributes at all. In this case linq2db will use POCO's name as table name and property names as column names (with exact same casing, which could be important for case-sensitive databases). This might seem to be convenient, but there are some restrictions: linq2db will not infer primary key even if class has property called "ID"; it will not infer nullability of string properties as there is no way to do so; and associations will not be automatically configured.
```

PR #3195, opened 2021-08-31
Added preprocessing for conditional expressions.

### LINKED ISSUE #3192: Don't select unnecessary columns
```cs
bool b = false;
var q = from t in db.Tbl1
        select new
        {
            t.C1,
            Other = b ? new { t.C2, t.C3, t.C4 } : null
        };
```
The SQL executed is:
```sql
SELECT
        [t].[C1],
        [t].[C2],
        [t].[C3],
        [t].[C4]
FROM
        [Tbl1] [t]
```
The property `Other` in the results is null. But it loads the columns C2, C3 and C4 from the DB.
Desired behavior: don't load columns which aren't needed.

sdanyliv:
Fixes #3192


MaceWindu (approved):
still want to see baselines when tests finish

MaceWindu (approved):
those changes introduced slight regression in Issue2478Tests.CrossApplyParameter test. Now it selects unused column in subquery, but it doesn't affect logic. We decided to not fix it now due to coplexity of fix

PR #3225, opened 2021-09-23
Improved Enumerable handling in queries.

### LINKED ISSUE #3220: Incorrect query cache behavior
Here is a simple example of two different joins but with the same query and result

```
    [Table("Item")]
    public class Item
    {
        [Column("ID")]
        [Identity, PrimaryKey]
        public int ID { get; set; }
        [Column("Value")]
    }


    public class TestDB: DataConnection
    {
        public TestDB(string connectionString):base(new PostgreSQLDataProvider(), connectionString) {}
        public ITable<Item> Items => GetTable<Item>();
    }

    class Program
    {
        {
            using (var db = new TestDB("Host=host;Port=5432;Database=testDb;Username=postgres;Password=pass"))
            {
                /*
                foreach (var i in Enumerable.Range(0, 1000))
                    db.Items.Value(item => item.Value, i).Insert();
                */

                var firstRange = Enumerable.Range(5, 5).ToList().AsQueryable();
                var secondRange = Enumerable.Range(50, 50).ToList().AsQueryable();

                var query1 = db.Items.Join(firstRange, item => item.Value, i => i, (item, i) => item);
                var firstJoined = query1.ToList();


                var query2 = db.Items.Join(secondRange, item => item.Value, i => i, (item, i) => item);
                var secondJoined = query2.ToList();


                Console.WriteLine(firstJoined.Count);
                Console.WriteLine(secondJoined.Count);
            }

            Console.ReadLine();
        }
    }
```

query1 as SQL:
```
--  PostgreSQL.9.2 PostgreSQL
SELECT
	item_1."ID",
	item_1."Value"
FROM
	"Item" item_1
		INNER JOIN (
			SELECT
				5 as c1
			UNION ALL
			SELECT
				6 as c1
			UNION ALL
			SELECT
				7 as c1
			UNION ALL
			SELECT
				8 as c1
			UNION ALL
			SELECT
				9 as c1
		) t1 ON item_1."Value" = t1.c1
```
generated SQL from query2 has the same query

The results of these queries are the same too. It looks like a bug.

If we wrap it into `NoLinqCache.Scope()` or add `Configuration.Linq.DisableQueryCache = true;` it will work correctly.

### Environment details
linq2db version: tested on 3.4.3 and 3.4.4
Database Server: Postgres 13.3
.NET Framework: tested on 4.5.2 and 4.6.1

sdanyliv:
Fixes #3220


viceroypenguin (approved):

Source/LinqToDB/Linq/Builder/EnumerableContext.cs
```diff
@@ -28,16 +33,99 @@ class EnumerableContext : IBuildContext
...
+		public EnumerableContext(ExpressionBuilder builder, BuildInfo buildInfo, SelectQuery query, Type elementType)
 		{
 			Parent            = buildInfo.Parent;
 			Builder           = builder;
 			Expression        = buildInfo.Expression;
 			SelectQuery       = query;
 			_elementType      = elementType;
 			_entityDescriptor = Builder.MappingSchema.GetEntityDescriptor(elementType);
-			Table             = new SqlValuesTable(source);
+			Table             = BuildValuesTable();
```
MaceWindu, line 44: Is it possible to reduce number of allocated collections in this class?

Source/LinqToDB/DataProvider/SQLite/SQLiteSqlBuilder.cs
```diff
@@ -184,5 +186,47 @@ protected override void BuildIsDistinctPredicate(SqlPredicate.IsDistinct expr)
...
+			{
+				if (i > 0)
+					StringBuilder.Append(InlineComma);
+				var field = valuesTable.Fields[i];
+				StringBuilder.Append("NULL AS ");
+				Convert(StringBuilder, field.PhysicalName, ConvertType.NameToQueryField);
+			}
+
+			StringBuilder
+				.Append(" WHERE 0")
```
MaceWindu, line 211: Is this line a reason we have separate builder for sqlite? Then maybe it's not worth of it?

MaceWindu (approved):
Some questions/potential improvements, but approved in general

Source/LinqToDB/DataProvider/SQLite/SQLiteSqlBuilder.cs
sdanyliv: It is only way that i have found how to set column aliases for VALUES

Issue #3158, opened 2021-08-06
Sql.Property validates its argument incorrectly


jods4:
This check here is supposed to validate the name of the property (at least I suppose, from observed behavior and the comments in code), but it actually checks the `Arguments[0]` which is the target object.

Because of this, `Sql.Property(x, "abc")` sometimes throws an `ArgumentException` when it should not, e.g. when `x` is neither a constant nor a parameter.

https://github.com/linq2db/linq2db/blob/2368b8321a9c5ac817731699b58f6bbb8d9a0310/Source/LinqToDB/Expressions/MemberHelper.cs#L50-L56

This bit of code reproduces the problem for me:
```cs
from t2 in db.Transfers
from s in db.OperationTypes.LeftJoin(s => s.Id == Sql.Property<int?>(t2, "TypeId"))
```

jods4:
Awesome thanks! 🎉
This has been a major pain point in our project.

PR #3240, opened 2021-10-04
[WIP] Fix issue #3158

jods4:


jods4:
Arf that query worked. I'll tweak it to fail later.

jods4:
For the record: `Issue3240` fails on the current version (before fixing), with the following error (which is what was reported in the issue):
image

One insight seems to be that the string column must not exist on the target type to reproduce the problem. I haven't looked deeply into the code but I suppose that if the target property does exist, the query is transformed differently and doesn't even go through the dynamic part.

jods4:
@MaceWindu I thought my commit above would fix the test case but apparently it does not :(
I'm not sure what's wrong now, could you take a look at the provided test case? Thanks!

PR #3252, opened 2021-10-11
Fix incorrect mapping schema id generation for combined schemas

### LINKED ISSUE #3251: Problem with Mapping Schema...
@sdanyliv @MaceWindu 

I've now detected a big issue with mapping schema in our software. (maybe related to #2582)

We create a mapping schema where we map all our types via fluent mapping builder.
During Startup, the mappingschema could already been used, but we need to add types. So we create a new mapping schema with this code (_mappingSchema is the already used one):

       var newMappingSchema = new MappingSchema(_mappingSchema);

with this I add other types, and then newMappingSchema replaces the existing one _mappingSchema...

When I create a DataConnection, I hand over the MappingSchema in the constructor.

Now I have the problem, that sometimes the mapping for my later added types is found, and sometimes not. It seems to depend on startup, maybe if something is read from the database before adding the additional types to the newMappingSchema.

MicrosoftTeams-image

In the screenshot you can see part of the issue. The DataConnection & MappingSchema are wrapped in our own DataLayer.
Se my two variables:

      t2 is from the Mappingschema where I add my Types to. It seems to know the Mapping
      t1 is from the DataConnection, the Schema used for t2 is hand over to the DataConnection, but it seems not to the type then in it's internal MappingSchema

do I do something wrong? Or do we maybe still have a problem?

jogibear9988:
Fix #3251

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.cs
+++ b/Source/LinqToDB/Data/DataConnection.cs
@@ -1601,8 +1601,6 @@ namespace LinqToDB.Data
 		/// </summary>
 		public  List<string>  NextQueryHints => _nextQueryHints ??= new List<string>();
 		
-		private static readonly MemoryCache<(string baseSchemaId, string addedSchemaId)> _combinedSchemas = new (new ());
-
 		/// <summary>
 		/// Adds additional mapping schema to current connection.
 		/// </summary>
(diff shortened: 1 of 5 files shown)


MaceWindu:
Quite big issue, surprised it wasn't discovered till now...

Issue #3305, opened 2021-10-30
complex query from #1875 does not work.


jogibear9988:
they query from the issue #1875 will not work.

i'll at a branch with a test.

jogibear9988:
sqlserver exception:

sqlserver sql:

jogibear9988:
sqlite throws following exception:

{"SQLite Error 1: 'near \"(\": syntax error'."}


sqlite generates following sql:

		WITH [CTE_1]
		(
			[IR_ResourceID],
			[RN],
			[IR_MaterialID],
			[Id],
			[BatchNumber],
			[BundleUnit],
			[ProductStatus],
			[CustomField1],
			[CustomField2],
			[CustomField3],
			[Count],
			[CountLocked],
			[IR_InfeedAdviceID]
		)
		AS
		(
			SELECT
				[x].[ResourceID],
				ROW_NUMBER() OVER(PARTITION BY [x].[ResourceID]),
				[x].[MaterialID],
				[x].[Id],
				[x].[BatchNumber],
				[x].[BundleUnit],
				[x].[ProductStatus],
				[x].[CustomField1],
				[x].[CustomField2],
				[x].[CustomField3],
				COUNT(*) OVER(PARTITION BY [x].[ResourceID]),
				COUNT(([x].[ProductStatus] > 0)) OVER(PARTITION BY [x].[ResourceID]),
				[x].[InfeedAdviceID]
			FROM
				[InventoryResourceDTO] [x]
			WHERE
				[x].[Status] < 99
		)
		SELECT
			[c_1].[Id],
			[c_1].[AisleID],
			[c_1].[MaterialID],
			[a].[Id],
			[a].[AisleNumber],
			[a].[PlantID],
			[a].[Name],
			[m_1].[Id],
			[m_1].[MaterialNumber],
			[m_1].[MaterialDescription_1],
			[m_1].[MaterialDescription_2],
			[m_1].[MaterialDescription_3],
			[m_1].[CategoryABC],
			[m_1].[CategoryCustoms],
			[m_1].[CategoryDimensions],
			[m_1].[CategoryQuality],
			[m_1].[CategoryTemperature],
			[a1].[Id],
			[c1].[Id],
			[c1].[Status],
			[c1].[ResourceLabel],
			[c1].[CustomField1],
			[c1].[CustomField2],
			[c1].[CustomField3],
			[c1].[CustomLong1],
			[c1].[CustomLong2],
			[c1].[CustomLong3],
[...]

PR #3306, opened 2021-10-30
test for issue #3305

jogibear9988:


jogibear9988:
@MaceWindu 

This now seems to work for SQLite, for SqlServer it still fails

PR #3319, opened 2021-11-06
t4 template fix for "assembly can't be resolved" on macos

pavlo-dudka:


MaceWindu:
Thanks. Need to check it still works for Visual Studio, but looks ok

MaceWindu:
Actually no, this fix will break use from VS as VS uses shadow copy. I will edit it a bit to use location only for non-vs hosts

pavlo-dudka:
sorry if it brakes template on Windows/VS
please adjust the changes as needed
the problem on MacOs is caused by removing "file:///", while only "file://" needs to be removed, e.g.
"file:///Users/pdu/.nuget/packages/linq2db.mysql/3.4.5/tools/linq2db.dll" should be changed to "/Users/pdu/.nuget/packages/linq2db.mysql/3.4.5/tools/linq2db.dll"

MaceWindu:
Please check it works after my change.

MaceWindu:
According to documentation it's even worse https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-5.0
under .net 5+ it will throw exception, but we still need to use it for .net framework

MaceWindu (approved):

pavlo-dudka:
> Please check it works for you after my change.

works fine, thank you

pavlo-dudka:
CodeBase property is also used in LinqToDB.Common.Tools.GetFileName, please consider to refactor it too

viceroypenguin (approved):

PR #3321, opened 2021-11-06
Optimize UPDATE SET queries

jods4:
I encountered a blocker while working on my SqlRow branch: I noticed that queries inside UPDATE SET expressions are not visited.

I traced that to the fact that `SqlUpdateStatement.WalkQueries` doesn't visit them, which I fixed in this PR.

Notice that I only visit "full query" statements, which is enough to unlock me. In reality an expression could be:
`SET x = 1 + (select 2)`
and I'm not visiting those queries (not sure how it would be done).

⚠ More tests from me showed that there's a problem with this PR, though -- and I'm not sure how to fix it:

When the query refers to the outer table being updated, linq2db can't find that table source.

I debugged a bit and it seems that it searches the SqlUpdateStatement alright, but instead of looking for the plain Table that is found in that statement, it looks for a weird sub-select expression instead... I'm not sure how that's supposed to be fixed.

### DIFF
--- a/Source/LinqToDB/SqlQuery/SqlUpdateStatement.cs
+++ b/Source/LinqToDB/SqlQuery/SqlUpdateStatement.cs
@@ -49,6 +49,17 @@ namespace LinqToDB.SqlQuery
 			return null;
 		}
 
+		public override void WalkQueries<TContext>(TContext context, Func<TContext, SelectQuery, SelectQuery> func)
+		{
+			base.WalkQueries(context, func);
+
+			foreach (var setItem in Update.Items) {
+				if (setItem.Expression is SelectQuery q) {
+					setItem.Expression = func(context, q);
+				}
+			}
+		}
+
 		public override ISqlTableSource? GetTableSource(ISqlTableSource table)
 		{
 			var result = SelectQuery.GetTableSource(table);


jods4:
In the build results below, you can see plenty of failing UPDATE tests with _Table not found_ that illustrate the problem.

If you want to debug something simpler, anything even minimal that refers to the updated table fails, like this:
```csharp
db.GetTable<Dest>()
  .Set(d => d.A, d => from db.GetTable<Src>() select d.A + 1)
  .Update()
```

jods4:
Awesome, thanks @sdanyliv
I should have time to rebase and test SqlRow this week

jods4:
@sdanyliv FYI I have merged this branch into sql-row branch and results look good!
Thanks!

sep2:
Sorry to bother, but this is not supported yet:
```c#
await Events.Where(x => x.Id == key)
            .Set(d => Sql.Row(d.Title, d.SearchVector),
                t => (from x in Events
                    where t.Id == x.Id
                    select Sql.Row(t.Title, EF
                        .Functions
                        .ToTsVector("test")
                    )).Single())
            .UpdateAsync();
```

model:
```c#
public class Event
{
    [LinqToDB.Mapping.PrimaryKey] public Guid Id { get; set; }
    public string Title { get; set; } = "";
    public NpgsqlTsVector? SearchVector { get; set; } = null;
}
```

```c#
System.InvalidOperationException: The 'ToTsVector' method is not supported because the query has switched to client-evaluation. This usually happens when the arguments to the method cannot be translated to server. Rewrite the query to avoid client evaluation of arguments so that method can be translated to server.
         at Microsoft.EntityFrameworkCore.NpgsqlFullTextSearchDbFunctionsExtensions.ToTsVector(DbFunctions _, String document)
         at lambda_method215(Closure , Expression , IDataContext , Object[] )
         at LinqToDB.Linq.QueryRunner.SetParameters(Query query, Expression expression, IDataContext parametersContext, Object[] parameters, Int32 queryNumber, SqlParameterValues parameterValues)
         at LinqToDB.Linq.QueryRunnerBase.SetCommand(Boolean forGetSqlText)
         at LinqToDB.Data.DataConnection.QueryRunner.ExecuteNonQueryAsync(CancellationToken cancellationToken)
         at LinqToDB.Linq.QueryRunner.NonQueryQueryAsync(Query query, IDataContext dataContext, Expression expression, Object[] ps, Object[] preambles, CancellationToken cancellationToken)
         at LinqToDB.Linq.QueryRunner.NonQueryQueryAsync(Query query, IDataContext dataContext, Expression expression, Object[] ps, Object[] preambles, CancellationToken cancellationToken)
         at LinqToDB.Linq.ExpressionQuery`1.LinqToDB.Async.IQueryProviderAsync.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken)
         at LinqToDB.LinqExtensions.UpdateAsync[T](IUpdatable`1 source, CancellationToken token)
```

Issue #2839, opened 2021-02-14
BulkCopyType.MultipleRows could insert incorrect data into varchar columns in Firebird


MaceWindu:
Discovered by #2829

`MultipleRows` use sql like:
```sql
insert from (
select {columns} from rdb$database
union all
select {columns} from rdb$database
...
union all
select {columns} from rdb$database
)
```

If one of target columns is of varchar type it could result in column values padded with spaces due to following reasons:
- we specify column value as string literal which is typed as CHAR(literal_length)
- firebird converts subquery column type to CHAR(max_literal_len) (basically padding smaller values with spaces)
- values with padding inserted into target varchar column

To fix it we need to type string literals that target varchar column with varchar type (for all record or only for first one if it will be enough, needs testing).

jods4:
This seem to affect test data insertion.

I've had to disable Firebird from test `LeadLagOverloads` (see #3423 for context), because when inserting this data:
```cs
var data = new Issue1799Table3[] 
{
	new() { ProcessID = 1, ProcessName = "One" },
	new() { ProcessID = 2, ProcessName = "Two" },
	new() { ProcessID = 3, ProcessName = "Three" },
	new() { ProcessID = 4, ProcessName = "Four" },
};
```
All `ProcessName` have been inserted with a length of 5 (I suppose that's the max length of `"Three"`), padded with spaces.
Tests expectations then fail, of course, because when we assert something is `"Two"`, it actually comes back from DB as `"Two  "`.

PR #3328, opened 2021-11-09
RETURNING statement support by existing Output API

### LINKED ISSUE #869: Output clause in DELETE, INSERT, UPDATE and MERGE
Merge in mssql can return affected rows with "output_clause". Are there any plans to support it in Merge API?

### LINKED ISSUE #979: OUTPUT: PostgreSQL support (RETURNING  clause)
Using postgres I use "returning *" when inserting or updating data. 
This allows you to get not only the primary key but also all the fields. 
Is it possible to do this with linq2db?

 I think in some cases this is more convenient than inserting data and then selecting them. 
If you use "returning *", you get an atomic operation and you do not need to select the primary key type.

Something like
```csharp
T obj = db.Insert<T>(T item);
T obj = db.Update<T>(T item);
```

### LINKED ISSUE #2958: PostrgreSql error when InsertWithOutputAsync
Returning clause is empty in result sql query

```
Exception message: 42601: syntax error at end of input
Stack trace:
at Npgsql.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|194_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
   at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.ExecuteReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.QueryRunner.ExecuteReaderAsync(CancellationToken cancellationToken)
   at LinqToDB.Linq.QueryRunner.ExecuteQueryAsync[T](Query query, IDataContext dataContext, Mapper`1 mapper, Expression expression, Object[] ps, Object[] preambles, Int32 queryNumber, Func`2 func, TakeSkipDelegate skipAction, TakeSkipDelegate takeAction, CancellationToken cancellationToken)
   at LinqToDB.Linq.QueryRunner.ExecuteQueryAsync[T](Query query, IDataContext dataContext, Mapper`1 mapper, Expression expression, Object[] ps, Object[] preambles, Int32 queryNumber, Func`2 func, TakeSkipDelegate skipAction, TakeSkipDelegate takeAction, CancellationToken cancellationToken)
   at LinqToDB.Linq.ExpressionQuery`1.GetForEachAsync(Action`1 action, CancellationToken cancellationToken)
   at LinqToDB.AsyncExtensions.ToArrayAsync[TSource](IQueryable`1 source, CancellationToken token)
```

### Steps to reproduce
It's need postgres on `localhost:5432` with created database `l2db_tests`.

Test framework: xunit

```c#
public class Test
{
    private readonly ITestOutputHelper _console;

    public Test(ITestOutputHelper console)
    {
        _console = console;
    }

    [Fact]
    public async Task InsertWithOutputError()
    {
        DataConnection.TurnTraceSwitchOn();
        DataConnection.WriteTraceLine = (message, displayName, lvl) =>
        {
            _console.WriteLine($"{lvl}|{displayName}|{message}");
        };

        var connectionString = "Host=localhost;Port=5432;Username=postgres;Password=postgres;Database=l2db_tests;Include Error Detail=true";
        var connectionOptions = new LinqToDbConnectionOptionsBuilder()
            .UsePostgreSQL(connectionString)
            .Build<Con>();
        await using var con = new Con(connectionOptions);
        await using var tmpTable = con.CreateTempTable<Entity>("entities");
        await con.InsertAsync(new Entity()
        {
            Id = "id",
            Name = "name"
        }, "entities");

        var inserted = await tmpTable.InsertWithOutputAsync(tmpTable, e => new Entity
        {
            Id = e.Name,
            Name = e.Id
        }, e => e.Id);

        Assert.Equal("name", Assert.Single(inserted));
        Assert.Collection(tmpTable.OrderBy(x => x.Id).ToArray(),
            e => Assert.Equal("id", e.Id),
            e => Assert.Equal("name", e.Id));
    }

    public class Con : DataConnection
    {
        public Con(LinqToDbConnectionOptions<Con> options) : base(options)
        {
        }

        public ITable<Entity> Entities => GetTable<Entity>();
    }

    public class Entity
    {
        [PrimaryKey, Column("id")]
        public string Id { get; set; }

        [Column("name")]
    }
}
```

In logs you can see sql query with empty `returning` clause

```
...
Info|DataConnection|BeforeExecute
--  PostgreSQL.9.5 PostgreSQL (asynchronously)

INSERT INTO entities
(
	id,
	name
)
SELECT
	t1.name,
	t1.id
FROM
	entities t1
RETURNING

Error|DataConnection|Error
Exception: Npgsql.PostgresException
...
```

### Environment details
linq2db version: 3.3.0
Database Server: PostgreSQL 11.7 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.2.0) 9.2.0, 64-bit
Operating system: ubuntu 20.04
.NET Framework: NET 5.0.5

### LINKED ISSUE #3018: DeleteWithOutput always returns an empty collection
`DeleteWithOutput()` and `DeleteWithOutputAsync()` generate and execute the query correctly, but they always return an empty collection, no matter how many entries are removed from the database table. No exception is thrown.

I'm using them exactly like they are used in [Tests.Linq](https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Update/DeleteWithOutputTests.cs) but I invoke the method directly from the DbContext object. It always returns an empty collection. I'm wondering whether I'm using them correctly or if a certain condition must be met in order for them to work. Some advice would be appreciated.

### Steps to reproduce
- Query any table with one of the two methods above. For example:
```cs
// This deletes a bunch of entries from the table, but returns an Enumerable.Empty<UserEntity>()
var deletedEntities = db.Users
    .Where(x => x.Id > 50)
    .DeleteWithOutput();
```
```cs
// This deletes all entries from the table, but returns an int[0]
var ids = await db.Users
    .DeleteWithOutputAsync(x => x.Id);
```

### Environment details
linq2db version: **3.3.0**
Database Server: **PostgreSQL 13.1**
Database Provider: **Npgsql**
Operating system: **Windows 10**
**.NET 5**

### LINKED ISSUE #3044: UpdateWithOutput and Take(n) generates incorrect Sql
I'm trying out the new UpdateWithOutput functionality (thanks for providing this) but when used with Take(n) it is generating incorrect SQL

``` csharp
ctx.Queue
    .Where(q => q.ProcessStarted == null)
    .Take(1)
    .UpdateWithOutput(q => new Queue { ProcessStarted = DateTime.Now })
```

is creating

```sql
--  SqlServer.2017
DECLARE @Now DateTime2
SET     @Now = '2021-06-08T14:48:46.9614876'
DECLARE @take Int -- Int32
SET     @take = 1

UPDATE
	[t1]
SET
	[t1].[ProcessStarted_2] = @Now
OUTPUT
	[t1].[Id],
	[t1].[ProcessStarted],
	[t1].[Id_1],
	[t1].[ProcessStarted_1]
FROM
	(
		SELECT TOP (@take)
			[DELETED].[Id],
			[DELETED].[ProcessStarted],
			[INSERTED].[Id] as [Id_1],
			[INSERTED].[ProcessStarted] as [ProcessStarted_1],
			[q].[ProcessStarted] as [ProcessStarted_2]
		FROM
			[dbo].[Queue] [q]
		WHERE
			[q].[ProcessStarted] IS NULL
	) [t1]
```

This is giving an SQL exception:

```
Exception: System.Data.SqlClient.SqlException
Message  : The multi-part identifier "DELETED.Id" could not be bound.
The multi-part identifier "DELETED.ProcessStarted" could not be bound.
The multi-part identifier "INSERTED.Id" could not be bound.
The multi-part identifier "INSERTED.ProcessStarted" could not be bound.
```

I would expect it to create

```sql
UPDATE
	[t1]
SET
	[t1].[ProcessStarted] = @Now
OUTPUT
	[DELETED].[Id],
	[DELETED].[ProcessStarted],
	[INSERTED].[Id],
	[INSERTED].[ProcessStarted]
FROM
	(
		SELECT TOP (@take)
			[q].[Id],
			[q].[ProcessStarted]
		FROM
			[dbo].[Queue] [q]
		WHERE
			[q].[ProcessStarted] IS NULL
	) [t1]
```

### Environment details
linq2db version: 3.4.0
Database Server: SqlServer 2017
Database Provider: System.Data.SqlClient
Operating system: Windows10
.NET Framework: .Net Core 3.1

MaceWindu:
Fix #869
Fix #979
Fix #2958
Fix #3018
Fix #3044

- add support for MariaDB, SQLite, PostgreSQL and Firebird to current API
- use of APIs limited by features, provided by databases
  - [MariaDB][SQLite][PostgreSQL] doesn't support history data for `DELETE` and `UPDATE` statements
  - [SQLite] requires SQLite 3.35+, so doesn't work with `Microsoft.Data.SQLite` 5.x and lower (6+ not tested). Works fine with `System.Data.SQLite` as it ships with most recent SQLite version
  - [Firebird] doesn't support `action` in merge
  - [Firebird] doesn't support statements with multiple affected records
  - `*OutputInto*` APIs not supported (no support from databases)
- [PostgreSQL] query from output not supported as it required new APIs. This PR only updates support for already existing APIs
- reduced number of test records in update tests to 9 due to #2839

Remaining Tasks:
- [ ] (nice to have) source columns in output should be replaced with subqueries when update rewritten by `GetAlternativeUpdate`
- [x] support non-table sources in output (required for #3044)
- [x] (**BLOCKING**) Delete for MariaDB doesn't work, as we need to generate alias-free target table for it


jods4:
I wanted to ask if there is any support for Oracle RETURNING?
AFAIK the Oracle feature is shitty because it doesn't directly return rows as a select would, it can only insert into variables. :(

MaceWindu:
No Oracle support as I have no idea how we can do it with variables. If you have any idea how it could work - we can try to add it.
Also there is no DB2 support as their API also strange and more similar to postgresql quaryable returning, which we don't support yet.

MaceWindu:
But with variables I suspect it will be similar to Firebird - only one-record queries work?

jods4:
@MaceWindu 
RE Oracle: returning a single line is simple enough because you can bind each output to a parameter.
Returning bulk lines is tricky and I never took the time to figure out the best approach in plain SQL.
I *think* the best approach that doesn't require pre-defining Oracle temp tables, 
is to use a PL/SQL block and declare a variable of table kind, output into this variable, then bind the table to a ref cursor parameter (this is how you can return one, or more, result sets from a non-select query).

So something something like this:
```sql
declare
  type t_output as table(...);
  v_output t_output;
begin 
  update/insert
  ...
  returning a, b, c bulk collect into v_output;
  
  -- bind v_output to a ref cursor parameter
end;
```

Sorry it's all vague.
OUTPUT is a really powerful feature but it's not very usable in Oracle.

In our complex processes, we usually use temp tables in a way that we then don't need OUTPUT.

to11mtm:
> @MaceWindu RE Oracle: returning a single line is simple enough because you can bind each output to a parameter. Returning bulk lines is tricky and I never took the time to figure out the best approach in plain SQL. I _think_ the best approach that doesn't require pre-defining Oracle temp tables, is to use a PL/SQL block and declare a variable of table kind, output into this variable, then bind the table to a ref cursor parameter (this is how you can return one, or more, result sets from a non-select query).
> 
> So something something like this:
> 
> ```sql
> declare
>   type t_output as table(...);
>   v_output t_output;
> begin 
>   update/insert
>   ...
>   returning a, b, c bulk collect into v_output;
>   
>   -- bind v_output to a ref cursor parameter
> end;
> ```
> 
> Sorry it's all vague. OUTPUT is a really powerful feature but it's not very usable in Oracle.
> 
> In our complex processes, we usually use temp tables in a way that we then don't need OUTPUT.

When I used to work on Oracle, `BULK COLLECT INTO` along with a cursor was how I would roll.

I think in 12c, you don't need to pass a refcursor out necessarily (implicit statement results), but I don't know if that is true for `RETURNING` case or if you'd still need to put it in a cursor/etc and `SELECT` back out.

jods4:
@to11mtm I would be very interested in the full details of what's the most lean way to OUTPUT a list of rows from an Oracle DDL statement in your opinion.

Even better if it can then be integrated into linq2db!

Being able to read the results of a complex insert/update/delete is an extremely powerful feature but Oracle makes it so impractical to use -- esp. compared to SQL Server where basically any DDL simply becomes a SELECT. 😒

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
```diff
@@ -311,12 +311,43 @@ protected override void BuildDataTypeFromDataType(SqlDataType type, bool forCrea
...
+
+				table = tables[0];
+
+				if (deleteStatement.Output != null && tables.Count == 1 && tables[0].Joins.Count == 0)
+				{
+					var t = table;
+
+					while (t is SqlTableSource sqlTableSource)
+					{
+						sqlTableSource.Alias = null;
```
sdanyliv, line 332: It is not thread safe to Modify AST in SQLBuilder. The same AST can be used by other ParameterDepended query.

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
```diff
@@ -311,12 +311,43 @@ protected override void BuildDataTypeFromDataType(SqlDataType type, bool forCrea
...
+						t                    = sqlTableSource.Source;
+					}
+
+					if (t is SqlTable sqlTable)
+						sqlTable.Alias = null;
```
sdanyliv, line 337: Same, AST modification.

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
igor-tkachev: All the threads should set this to null anyway.

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
igor-tkachev: fixed

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
igor-tkachev: fixed

viceroypenguin (approved):

MaceWindu:
@jods4 we are not going to support oralce in this PR (we want to finalize 3.7 and there is a lot of stuff left still). If you want this feature implemented, please create feature request or even better - PR

jods4:
Fair enough.
FYI I've tweeted at Oracle about this.
https://twitter.com/jods16/status/1502764925323776005 
The anonymous block, with a custom type inside, is super ugly but apparently the only way. Not sure that's very practical, I never do it in my hand-written code.

jods4:
@MaceWindu after some back-and-forth with our Oracle friends, they suggested this solution:
https://community.oracle.com/tech/developers/discussion/4496022/odp-net-insert-update-returning-into-bulk-collect-result-set

As you can see, it's most likely too complicated and specific to be integrated in a general purpose library such as linq2db. Maybe some support for more limited RETURNING [INTO | BULK COLLECT] might be an option if someone wants to work on it (I won't).

to11mtm:
> @to11mtm I would be very interested in the full details of what's the most lean way to OUTPUT a list of rows from an Oracle DDL statement in your opinion.
> 
> Even better if it can then be integrated into linq2db!
> 
> Being able to read the results of a complex insert/update/delete is an extremely powerful feature but Oracle makes it so impractical to use -- esp. compared to SQL Server where basically any DDL simply becomes a SELECT. 😒

So, I misremembered on this. We were either dynamically using `%rowtype` and/or inlining `TYPE r IS RECORD` +`TYPE rTbl IS TABLE OF r` in an anonymous PL/SQL block and slamming that into a refcursor for return.

> MaceWindu after some back-and-forth with our Oracle friends, they suggested this solution: https://community.oracle.com/tech/developers/discussion/4496022/odp-net-insert-update-returning-into-bulk-collect-result-set
> 
> As you can see, it's most likely too complicated and specific to be integrated in a general purpose library such as linq2db. Maybe some support for more limited RETURNING [INTO | BULK COLLECT] might be an option if someone wants to work on it (I won't).

So, this might be *easier* than what I was thinking of in some ways, since it doesn't involve a lot of specific types, just ROWID. However I am rusty on my oracle; what are the risks if triggers are involved? i.e. If there is a trigger on insert or update, will we get the post-trigger value? (It's been 5 years since I've done serious Oracle so I'm trying to remember the rules here...)

jods4:
@to11mtm it's still fairly involved.

They made it kind of "simpler" by only returning the rowid instead of the actual data (which avoids having to create a specific type for the output shape).

On the other hand, you need to build a query that computes any output expression you wanted using those rowid. Tackling this in a general fashion is not trivial:
- To access data _before_ update/delete they use Oracle Flashback, which is already not trivial. Given how quick this happens there should be no problem but technically you risk the undo buffer not containing the relevant data anymore, in which case the query fails.
- When doing INSERT..SELECT (let's remind ourselves that Oracle doesn't support multiple tables in UPDATE or DELETE), you could reference other tables in your OUTPUT expression. It's basically impossible to find which rows were related to the inserted row after-the-fact, so Oracle support would need to restrict itself to constants / expressions involving the target table _only_. Mind you, this is probably the most common case and is still very useful.
- Reading from other tables is a bad idea anyway, as you risk inconsistent reads depending on your isolation level and concurrent transactions.

With the limitation above, and if we have faith in Undo log / Flashback, it might be doable. The codegen is fairly involved as you need an anonymous PL/SQL block, multiple queries, potentially including "AS OF" temporal query and a ref cursor parameter.

jods4:
> If there is a trigger on insert or update, will we get the post-trigger value? (It's been 5 years since I've done serious Oracle so I'm trying to remember the rules here...)

Yes, because they basically read the table again after the UPDATE/INSERT, so you'll get the values that are stored inside the table, after the trigger.

sdanyliv (approved):

Issue #3327, opened 2021-11-08
Extension `AsValueInsertable` for IQueryable<T> ?


MihailsKuzmins:
Based on https://stackoverflow.com/questions/69882757/iupdatablet-and-ivalueinsertablet-from-optional-values

IQueryable has an extension `AsUpdatable` which allows building dynamic blocks with `Set()`, but there is no extension which would cast `ITable<T>` to `IValueInsertable<T>`. Would it be possible to add it?

P.S. is it ok that I have created both the stackoverflow question and this issue?

### Environment details
linq2db version: 3.5.1

sdanyliv:
For which purpose do you need that?

MihailsKuzmins:
for the same purpose as `AsUpdatable` with optionals, i.e.
```cs
var optional1 = Optional<int>.None();
var optional2 = Optional<int>.Of(123);

var table = Database.MyTable; // ITable<MyRecord>
if (optional1.HasValue) table = table.Value(x => x.Optional1, optional1.Value);
if (optional2.HasValue) table = table.Value(x => x.Optional2, optional2.Value);

await table.InsertAsync();
```

MihailsKuzmins:
Thank you. Not a perfect solution for my case (generally my extension methods only have `ITable<T>` not the whole `IDataContext`), but it's better than nothing

So if you do not plan to include creation of `IValueInsertable<T>` from `ITable<T>` please feel free to close the issue

sdanyliv:
Well, you can try implement it by yourself. PRs are welcome .

MihailsKuzmins:
I will try to do something, but my VS enters the berserk mode as soon as I try doing something in the solution 😞 

![image](https://user-images.githubusercontent.com/47413092/140995495-b07c7e2f-98bd-4476-9bf5-9a8734f1c5ce.png)

sdanyliv:
Use `linq2db.Playground.slnf` and link only needed tests.

MaceWindu:
merged

PR #3331, opened 2021-11-09
Add AsValueInsertable

MihailsKuzmins:
Issue #3327


sdanyliv:
Nicely done. But we need test when no `Value` is called. Probably it will fail, but we need to figure out what to do with that.

MihailsKuzmins:
Good catch. The values block will be empty so it will fail. I would say that it is a mistake of the developer, not Linq2DB.
But if you have some suggesstions (sorry, I am not experienced to offer something), we could try it 😃

to11mtm:
> Good catch. The values block will be empty so it will fail. I would say that it is a mistake of the developer, not Linq2DB. But if you have some suggesstions (sorry, I am not experienced to offer something), we could try it 😃

We probably don't need to get very fancy. Can we just start with a test that tries to call the insert without any `.Value()`s added and see what it does? Example:

```
[Test]
		public void AsValueInsertableEmptyThrowsTest([DataSources(TestProvName.All] string context)
		{
			using (var table = db.CreateLocalTable<TestInsertOrReplaceTable>())
			{
				var vi = table.AsValueInsertable();

				Assert.Throws(/*???*/,()=> vi.Insert());
			}
		}
```

Ideally, the exception should have a message that may help the user understand that no .Value()s were passed in.

Usaga:
If Values block is empty it is not a mistake. It shouldn't throw or fail any way. This insert operation may be conditional. Like `SaveChanges()` in EF which will do nothing if nothing has changed\inserted.

Programmer may check what `Insert()` returns, and if it return zero (nothing was inserted), then he\she\it may react the way he\she\it wants.

Fail may be needed if we meet some kind of uncertainty.

to11mtm:
You could probably get away with returning a different value for the basic
.Insert case, but InsertWithInt32/64Identity would have potentially
ambiguous behavior short of erroring out.

MihailsKuzmins:
When I run a unit test for `IValueInsertable` without any properties the error message is **really bad** since `insert.Into` is null:
*InsertBuilder.cs line 216*
image

I can only suggest:
- throw another more meaningful exception here
- wrap the foreach block in the if block which checks `insert.Into` for null. There are further null reference errors in `BuildPhysicalTable` where the argument table is null since `insertClause.Into` is passed to it

MaceWindu (approved):

MaceWindu:
@MihailsKuzmins thanks!

MihailsKuzmins:
@MaceWindu thank you for merging the change :)

PR #3336, opened 2021-11-11
Fixed found issues with current code.

sdanyliv:
- Prevent connecting to database while calling `query.ToString()`
- Fixed lambda copy logic for  Eager Loading.

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.cs
+++ b/Source/LinqToDB/Data/DataConnection.cs
@@ -1072,7 +1072,7 @@ namespace LinqToDB.Data
 		/// </summary>
 		public IDbConnection Connection => EnsureConnection().Connection;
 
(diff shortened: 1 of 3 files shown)


Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1239,7 +1239,7 @@ public  IDbCommand  Command
 		/// </summary>
 		public IDbCommand CreateCommand()
 		{
-			var command = Connection.CreateCommand();
+			var command = EnsureConnection(false).CreateCommand();
```
MaceWindu, line 1242: `EnsureConnectionAsync` not affected?

MaceWindu (approved):

viceroypenguin (approved):

PR #3339, opened 2021-11-11
Row constructors

### LINKED ISSUE #2204: [Oracle] UPDATE FROM: implement correlated update support
Hi!

I think Oracle has more good candidate for such thing:
```sql
UPDATE
  [UpdatedEntities]
SET
  ( [Value1], [Value2], [Value3] ) = 
  (
    SELECT
      ([UpdatedEntities].[Value1] * [t].[Value1]) * @int1,
      ([UpdatedEntities].[Value2] * [t].[Value2]) * @int2,
      ([UpdatedEntities].[Value3] * [t].[Value3]) * @int3
    FROM
      [NewEntities] [t] 
    WHERE
      [t].[id] <> @someId AND [UpdatedEntities].[id] = [t].[id]
  )
WHERE
  EXISTS(
    SELECT
      *
    FROM
       [NewEntities] [t6] 
    WHERE
      [t6].[id] <> @someId AND [UpdatedEntities].[id] = [t6].[id]
  )
  ```

_Originally posted by @AndrePScope in https://github.com/linq2db/linq2db/pull/1278#issuecomment-620851672_

### LINKED ISSUE #2657: ORACLE Update with sub-query
```c#
var objWhereQueryAktiv = from objRentenDetails in dbContext.W_AUSZAHLUNG_RENTENDETAIL
                            join objVersicherte in dbContext.VERSICHERTE on objRentenDetails.VERSINT_ID equals objVersicherte.VERSINT_ID
                            where objRentenDetails.REPORT_ID == 123412
                            select new
                            {
                                RentenDetails = objRentenDetails,
                                Versicherte = objVersicherte
                            };

objWhereQueryAktiv.Update(dbContext.W_AUSZAHLUNG_RENTENDETAIL, objUpdate =>
        new W_AUSZAHLUNG_RENTENDETAIL()
        {
            VERS_ID = objUpdate.Versicherte.VERS_ID,
            PERSONAL_ID = objUpdate.Versicherte.PERSONAL_ID,
            NACHNAME = objUpdate.Versicherte.NACHNAME,
            VORNAME = objUpdate.Versicherte.VORNAME
        }
);
```
 ### Generate this BIG Oracle SQL:
```
UPDATE
	W_AUSZAHLUNG_RENTENDETAIL
SET
	W_AUSZAHLUNG_RENTENDETAIL.VERS_ID = (
		SELECT
			objVersicherte.VERS_ID
		FROM
			W_AUSZAHLUNG_RENTENDETAIL objRentenDetails
				INNER JOIN VERSICHERTE objVersicherte ON objRentenDetails.VERSINT_ID = objVersicherte.VERSINT_ID
		WHERE
			objRentenDetails.REPORT_ID = 123412 AND
			W_AUSZAHLUNG_RENTENDETAIL.ADRESS_ID = objRentenDetails.ADRESS_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER = objRentenDetails.EMPFAENGER AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER_ID = objRentenDetails.EMPFAENGER_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.GUELTIG_AB = objRentenDetails.GUELTIG_AB AND
			W_AUSZAHLUNG_RENTENDETAIL.REGLEMENT_ID = objRentenDetails.REGLEMENT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.RENTENART_ID = objRentenDetails.RENTENART_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.REPORT_ID = objRentenDetails.REPORT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.SACHBEARB_ID = objRentenDetails.SACHBEARB_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID = objRentenDetails.VERSINT_ID
	),
	W_AUSZAHLUNG_RENTENDETAIL.PERSONAL_ID = (
		SELECT
			objVersicherte_1.PERSONAL_ID
		FROM
			W_AUSZAHLUNG_RENTENDETAIL objRentenDetails_1
				INNER JOIN VERSICHERTE objVersicherte_1 ON objRentenDetails_1.VERSINT_ID = objVersicherte_1.VERSINT_ID
		WHERE
			objRentenDetails_1.REPORT_ID = 123412 AND
			W_AUSZAHLUNG_RENTENDETAIL.ADRESS_ID = objRentenDetails_1.ADRESS_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER = objRentenDetails_1.EMPFAENGER AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER_ID = objRentenDetails_1.EMPFAENGER_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.GUELTIG_AB = objRentenDetails_1.GUELTIG_AB AND
			W_AUSZAHLUNG_RENTENDETAIL.REGLEMENT_ID = objRentenDetails_1.REGLEMENT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.RENTENART_ID = objRentenDetails_1.RENTENART_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.REPORT_ID = objRentenDetails_1.REPORT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.SACHBEARB_ID = objRentenDetails_1.SACHBEARB_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID = objRentenDetails_1.VERSINT_ID
	),
	W_AUSZAHLUNG_RENTENDETAIL.NACHNAME = (
		SELECT
			objVersicherte_2.NACHNAME
		FROM
			W_AUSZAHLUNG_RENTENDETAIL objRentenDetails_2
				INNER JOIN VERSICHERTE objVersicherte_2 ON objRentenDetails_2.VERSINT_ID = objVersicherte_2.VERSINT_ID
		WHERE
			objRentenDetails_2.REPORT_ID = 123412 AND
			W_AUSZAHLUNG_RENTENDETAIL.ADRESS_ID = objRentenDetails_2.ADRESS_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER = objRentenDetails_2.EMPFAENGER AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER_ID = objRentenDetails_2.EMPFAENGER_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.GUELTIG_AB = objRentenDetails_2.GUELTIG_AB AND
			W_AUSZAHLUNG_RENTENDETAIL.REGLEMENT_ID = objRentenDetails_2.REGLEMENT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.RENTENART_ID = objRentenDetails_2.RENTENART_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.REPORT_ID = objRentenDetails_2.REPORT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.SACHBEARB_ID = objRentenDetails_2.SACHBEARB_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID = objRentenDetails_2.VERSINT_ID
	),
	W_AUSZAHLUNG_RENTENDETAIL.VORNAME = (
		SELECT
			objVersicherte_3.VORNAME
		FROM
			W_AUSZAHLUNG_RENTENDETAIL objRentenDetails_3
				INNER JOIN VERSICHERTE objVersicherte_3 ON objRentenDetails_3.VERSINT_ID = objVersicherte_3.VERSINT_ID
		WHERE
			objRentenDetails_3.REPORT_ID = 123412 AND
			W_AUSZAHLUNG_RENTENDETAIL.ADRESS_ID = objRentenDetails_3.ADRESS_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER = objRentenDetails_3.EMPFAENGER AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER_ID = objRentenDetails_3.EMPFAENGER_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.GUELTIG_AB = objRentenDetails_3.GUELTIG_AB AND
			W_AUSZAHLUNG_RENTENDETAIL.REGLEMENT_ID = objRentenDetails_3.REGLEMENT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.RENTENART_ID = objRentenDetails_3.RENTENART_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.REPORT_ID = objRentenDetails_3.REPORT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.SACHBEARB_ID = objRentenDetails_3.SACHBEARB_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID = objRentenDetails_3.VERSINT_ID
	)
WHERE
	EXISTS(
		SELECT
			*
		FROM
			W_AUSZAHLUNG_RENTENDETAIL objRentenDetails_4
				INNER JOIN VERSICHERTE objVersicherte_4 ON objRentenDetails_4.VERSINT_ID = objVersicherte_4.VERSINT_ID
		WHERE
			objRentenDetails_4.REPORT_ID = 123412 AND
			W_AUSZAHLUNG_RENTENDETAIL.ADRESS_ID = objRentenDetails_4.ADRESS_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER = objRentenDetails_4.EMPFAENGER AND
			W_AUSZAHLUNG_RENTENDETAIL.EMPFAENGER_ID = objRentenDetails_4.EMPFAENGER_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.GUELTIG_AB = objRentenDetails_4.GUELTIG_AB AND
			W_AUSZAHLUNG_RENTENDETAIL.REGLEMENT_ID = objRentenDetails_4.REGLEMENT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.RENTENART_ID = objRentenDetails_4.RENTENART_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.REPORT_ID = objRentenDetails_4.REPORT_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.SACHBEARB_ID = objRentenDetails_4.SACHBEARB_ID AND
			W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID = objRentenDetails_4.VERSINT_ID
	)
```
### Why not generate an SQL with a subquery like this:
```
UPDATE W_AUSZAHLUNG_RENTENDETAIL
   SET(VERS_ID, PERSONAL_ID, NACHNAME, VORNAME)
       =
       (SELECT VERSICHERTE.VERS_ID, VERSICHERTE.PERSONAL_ID, VERSICHERTE.NACHNAME, VERSICHERTE.VORNAME
          FROM VERSICHERTE 
          WHERE VERSICHERTE.VERSINT_ID = W_AUSZAHLUNG_RENTENDETAIL.VERSINT_ID
       )
 WHERE REPORT_ID = 32142
```

### For information Sql-Server generates the following SQL:
```
UPDATE
	[W_AUSZAHLUNG_RENTENDETAIL]
SET
	[W_AUSZAHLUNG_RENTENDETAIL].[VERS_ID] = [objVersicherte].[VERS_ID],
	[W_AUSZAHLUNG_RENTENDETAIL].[PERSONAL_ID] = [objVersicherte].[PERSONAL_ID],
	[W_AUSZAHLUNG_RENTENDETAIL].[NACHNAME] = [objVersicherte].[NACHNAME],
	[W_AUSZAHLUNG_RENTENDETAIL].[VORNAME] = [objVersicherte].[VORNAME]
FROM
	[W_AUSZAHLUNG_RENTENDETAIL] [objRentenDetails]
		INNER JOIN [VERSICHERTE] [objVersicherte] ON [objRentenDetails].[VERSINT_ID] = [objVersicherte].[VERSINT_ID]
WHERE
	[objRentenDetails].[REPORT_ID] = 123412
```

### Environment details
linq2db version: 3.1.6
Database Server: Oracle
Database Provider: OracleManaged
Operating system: Windows 10
.NET Framework: 4.6.1

### LINKED ISSUE #2844: Inefficient SQL generated for Bulk Update from in-memory collection
Hi,

I am attempting to design a generic 'bulk update' method which uses a temp table to store the externally updated rows and joins the temp table with the actual table to do the update operation efficiently: 

```c#
public void UpdateRange<T>(IEnumerable<T> source) where T : class, IEntity
{
    using var connection = _context.CreateLinqToDbConnection();

    connection.BeginTransaction();

    using var tempTable = connection.CreateTempTable(source, tableName: $"{typeof(T).Name}UpdateTemp");

    var entityTable = connection.GetTable<T>();

    var updates = from e in entityTable
        join te in tempTable on e.Id equals te.Id into gj
        from te in gj
        select te;

    updates.Update(entityTable, e => e);

    connection.CommitTransaction();
}
```

I use this strategy as opposed to a `MERGE` statement to support RDBMS that don't support `MERGE`.

When working with the following test model and interfaces:

```c#
public interface IEntity
{
    object Id { get; set; }
}

public interface IEntity<T> : IEntity
{
    new T Id { get; set; }
}

public abstract class Entity<T> : IEntity<T>
{
    public T Id { get; set; }

    object IEntity.Id
    {
        get => Id;
        set => Id = value != null ? (T) value : default;
    }
}

public class Person : Entity<int>
{
    public string FirstName { get; set; }
    public string LastName { get; set; }
}
```
The following SQL is generated:

```SQL
UPDATE
	[People]
SET
	[Id] = (
		SELECT
			[gj].[Id]
		FROM
			[People] [e]
				INNER JOIN [PersonUpdateTemp] [gj] ON [e].[Id] = [gj].[Id]
		WHERE
			[People].[Id] = [e].[Id]
	),
	[LastName] = (
		SELECT
			[gj_1].[LastName]
		FROM
			[People] [e_1]
				INNER JOIN [PersonUpdateTemp] [gj_1] ON [e_1].[Id] = [gj_1].[Id]
		WHERE
			[People].[Id] = [e_1].[Id]
	),
	[FirstName] = (
		SELECT
			[gj_2].[FirstName]
		FROM
			[People] [e_2]
				INNER JOIN [PersonUpdateTemp] [gj_2] ON [e_2].[Id] = [gj_2].[Id]
		WHERE
			[People].[Id] = [e_2].[Id]
	)
WHERE
	EXISTS(
		SELECT
			*
		FROM
			[People] [e_3]
				INNER JOIN [PersonUpdateTemp] [gj_3] ON [e_3].[Id] = [gj_3].[Id]
		WHERE
			[People].[Id] = [e_3].[Id]
	)

```

I was hoping to have a single join and update statement updating each column from the join result. How would I accomplish this? 

I have a repo that contains a test that produces the above SQL [here](https://github.com/Nate129/Linq2DbTests/tree/master) (see the `Linq2DbUpdateTests` class). I retrieved the SQL by setting a breakpoint in the `TestService` class on the last line of the method `connection.CommitTransaction();` and inspecting the Command property's SQLText property.

### Environment details
linq2db version: 3.2.3
Database Server: Sqlite
Database Provider: EF Core (via Linq2db.EntityFrameworkCore
Operating system: Windows 10
.NET Framework: .NET Core

jods4:
Fix #2204 
Fix #2657
Fix #2844

This PR adds support for SQL Row Constructors, which basically are SQL tuples, usually with syntax `ROW (1, 2)` or just `(1, 2)`.

### Design 
Row constructors are only server-side, they're always represented by a call to `Sql.Row(1, 2)` and they can't be passed as parameters.

The returned type is `SqlRow<T1, T2>`. 
Tuples with more elements simply nest `SqlRow` in the second argument: e.g. `SqlRow<int, SqlRow<int, int>>`  for a triple.
This may look inefficient but remember this will never be instantiated. It's only a compilation-time type to enforce static types.
The benefit when compared to `SqlRow<T1, T2, T3>` is that I only declared a single type and declared all its operators (such as `>`, `<`) once.
I created `Sql.Row` overloads up to 8 values. With this design, it's trivial to add more if so desired.

Sadly types are stictly checked by C# and I see no way to relax the constraints a bit when it comes to nullable types.
If one side has a `int?` and the other has not, a cast is required:
```csharp
// Error, SqlRow<int, int> incompatible with SqlRow<int, int?>
// Needs a cast (int?)2
Row(1, 2) == Row(3, null) 
```

Row constructors are always evaluated without `CompareNullsAsValues`. The problem here is that generated row ctor code must behave the same as equivalent code in DB without row ctor, and nothing makes sense (to me) for row ctor expressions.

There is no attempt to completely or partially evaluate a `SqlRow` expression, at least for now. 
For example in `(1, 2, 3) = (4, 5, 6)` one could check if any pair of values are constant that are different, or if one is null, and the whole expression is statically false. If any pair of values are constants that are the same, in theory the expression could be shortened.
Notice that in DB that don't support row ctors natively, this expression is converted by SqlOptimizer into an equivalent predicate and that runs through the optimizer and would be simplified.

### IS [NOT] NULL
The following translations are used when not supported:
```
(1, 2) IS NULL     -> 1 IS NULL AND 2 IS NULL
(1, 2) IS NOT NULL -> 1 IS NOT NULL AND 2 IS NOT NULL
```

### Comparisons
The following translations are used when not supported:
```
(1, 2) = (3, 4)        -> 1 = 3 and 2 = 4
(1, 2) <> (3, 4)       -> 1 <> 3 or 2 <> 4
(1, 2, 3) > (4, 5, 6)  -> 1 > 4 or 1 = 4 and 2 > 5 or 1 = 4 and 2 = 5 and 3 > 6
(1, 2, 3) >= (4, 5, 6) -> 1 > 4 or 1 = 4 and 2 > 5 or 1 = 4 and 2 = 5 and 3 >= 6
Likewise for < and <=
```

### [NOT] IN 
When not supported, expands to a big OR: 
```
(1, 2) IN ((3, 4), (5, 6)) -> (1, 2) = (3, 4) OR (1, 2) = (5, 6)
```
This gets translated further in case the provider doesn't support comparisons either, see above.

### OVERLAPS
As described in a comment below, emulating the precise behavior of OVERLAPS in SQL is practically impossible. 
Client-side C# support would in theory be possible but is much less useful.
Approximate, but still useful, behavior is possible but is left for future improvements, if so desired.

In this PR, OVERLAPS is only available natively, which means in DB that have support and only for specific types (i.e. dates and intervals).

If unsupported types are used, the SQL query is generated anyway and the DB provider error will surface.
```
(date'2020-01-01', date'2020-12-31') OVERLAPS (date'2020-07-01', null) -- true
```

### [NOT] BETWEEN
When not supported, converts to two comparisons:
```
1     BETWEEN 2 AND 3 -> 1 >= 2 AND 1 <= 3
1 NOT BETWEEN 2 AND 3 -> 1 < 2 OR 1 > 3
```
This gets translated further in case the provider doesn't support comparisons either, see above.

### Comparisons to (SELECT)
Providers that natively support Row Constructors can compare them to a sub-select query returning a single row of the same cardinality.

This might probably be emulated with a [NOT] EXISTS in other providers but I leave this to future improvements, if someone wants it.

The public API is like any scalar comparison to a sub-select: write a query returning a `Sql.Row()` and ends it with `.Single()`. Or if it might return more than one row, use `.First()` as appropriate.
Since any `object` can be compared for equality, the `.Single()` call is not actually needed for operators `==` and `!=`.

```csharp
from x in db.Source
where Sql.Row("name", x.B) >= 
  (from y in db.OtherSource
   where y.Id == 42
   select Sql.Row(y.Name, 0))
  .Single()
select x
```

Results in :
```sql
where ('name', x.B) >= (select y.Name, 0 from OtherSource y where y.id = 42)
```

Note that it's a runtime error to try to select `SqlRow` in the main SELECT clause, and `SqlRow` must be the only selected value.

> I've tried to define an implicit conversion from `IQueryable<SqlRow>` to `SqlRow` to avoid the `.Single()` and make it easier, but C# forbids the definition of implicit conversions from an interface.
>
> I don't think (?) linq2db has support for `>= [ANY|ALL] (SELECT multiple rows)`, please tell me if it does and I'll check it works properly.

### UPDATE
For provider that supports it, it's possible to SET `SqlRow` values. 
This can be very useful to set multiple values to the result of a single subquery.

Using a Row Constructor on the right-hand side is not supported by Oracle, which only accepts a subquery in this context.
This could be emulated by a `SELECT 1, 2 FROM DUAL` but I did not code this fallback because, quite frankly, if you're using plain values you can use several regular, single column assignment anyway. So why bother?

Example code:
```csharp
int updated = db
  .GetTable<T>("Table")
  .Where(t => t.Id == 4)
  .Set(
    t => SqlRow(t.A, t.B), 
    t => (from s in db.GetTable<S>("OtherTable") where s.Id == t.OtherId select SqlRow(s.A, 42)).Single()))
  .Set(
    t => SqlRow(t.C, t.D)
    t => SqlRow("John", "Doe")) // Not supported in Oracle
  .Set(t => t.SingleColumn, 0)
  .Update()
```
```sql
update Table t set
  (t.A, t.B) = (select s.A, 42 from OtherTable s where s.Id = t.OtherId),
  (t.C, t.D) = ('John', 'Doe'), -- Not supported in Oracle
  t.SingleColumn = 0
where t.Id = 4
```

Also note that `.Set(t => SqlRow(t.A, t.B), SqlRow(1, 2))` (overload with a value as second parameter instead of a lambda) is not supported as `SqlRow` instances can't exist in C#, only in expressions. 
This is in no way a limitation: if you're dealing with constants you can just use two normal calls `.Set(t => t.A, 1).Set(t => t.B, 2)` which would generate simpler SQL anyway.

### Future possible enhancements
This PR provides complete linq2db support for Row Constructors, including UPDATE statement, which provides a solution to #2204 and #2657.

It is a building block upon which a better UPDATE FROM alternative can be created but I do not intend on providing this translation inside this PR. 

Writing the query manually using `SqlRow` may result in better SQL queries because it gives full control on the `WHERE` and `SET` parts, which can sometimes be shorten in ways the optimizer cannot see.
It would still be very nice to have this as a much better codegen for UPDATE..FROM, if anyone wants to pick that up.

### Provider support
| Provider | Comparisons | IS&nbsp;NULL | IN | OVERLAPS | BETWEEN | Comp. to (SELECT) | UPDATE (SELECT) | UPDATE (literal)
| --- | --- | --- | --- | --- | --- | --- | --- | ---
| Access | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 
| DB2 | ✅ | ⬇ | ⬇ | ✅ | ✅| ⛔ | ✅ | ✅
| Firebird | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔   
| Informix | ✅ (1) (3) | ⬇ | ✅&nbsp;(1) | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 
| MySQL | ✅ | ⬇ | ✅ | ⛔ | ⬇ | ✅ | ⛔ | ⛔ 
| Oracle | ✅ (2) (3) | ⬇ | ✅ | ✅ | ⬇ | ✅ | ✅ | ⛔
| Postgre | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅
| SQLite | ✅ | ⬇ | ⬇ | ⛔ | ✅ | ✅ | ⛔ | ⛔ 
| SapHana | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 
| SqlCe | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 
| SqlServer | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 
| Sybase | ⬇ | ⬇ | ⬇ | ⛔ | ⬇ | ⛔ | ⛔ | ⛔ 

Notes:
1. Informix uses `ROW (1, 2)` syntax
2. Oracle needs extra brackets on the right-hand side to disambiguate syntax: `(1, 2) = ( (3, 4) )`
3. Oracle, Informix only supports equality (= and <>), none of >, >=, <, <=.


jods4:
Sorry guys, it looks like I messed up my previous PR 😅. This is the continuation of the work in #3027.

I have added some initial work on UPDATE, rebased on 3.5.2 and merged the #3321.
With that fix from @sdanyliv included, updates now look good, I'm almost feature-complete.

Teaser codegen:
```sql
UPDATE
SET
        ("Dest1".A, "Dest1".B) = (
                        y.N,
                        "Dest1".A + 1
                FROM
                        "Src" y
        )
WHERE
        "Dest1"."Id" = 1
```

sdanyliv:
I'm still don't like this. Better to create `SqlRow: ISqlExpression` AST. In this case even `SqlSetExpression` will be not changed. Later it is easy to generate Row update automatically.

jods4:
@sdanyliv ok, it's an easy enough change.
It was a little bit of work to create the class `SqlRow : ISqlExpression` but past this work I agree the result is actually better and easier to manipulate.

I am hitting one small roadblock, though :(
Here, when analyzing the setters in an UPDATE:
https://github.com/linq2db/linq2db/blob/61b9e5b06cb1defd23c2d6955439fc0ed52ad6f0/Source/LinqToDB/Linq/Builder/UpdateBuilder.cs#L458

The issue is that here the setters are just `Expression`, not `ISqlExpression`. They are converted into `SqlField` by very specific code in `GetField`.
What's the best way to check and convert them to `SqlRow`? 
I have put `Extension(BuilderType)` on all `Sql.Row()` methods and would rather not duplicate the logic of `SqlRow` creation :/

What do you think?

MaceWindu:
BTW, at some point it would be nice to refactor `SqlSetExpression` to support more forms of `set` like here https://docs.microsoft.com/en-us/sql/t-sql/queries/update-transact-sql?view=sql-server-ver15

sdanyliv:
Use `builder.ConvertToSqlExpression(ctx, fieldExpr, columnDescriptor, false)`. See below code example. Or also it can `ConvertToSql` method

jods4:
Sorry it took me so long, I was quite busy and had no time to work on this. 
Patience is rewarded, though: 

I have finished the refactoring requested by @sdanyliv as well as rounded up a few issues/tasks.
This PR is feature-complete and I don't intend on adding/changing the code anymore 🎉🎉🎉
-- barring any findings during tests, of course.

As such it's a great time to start reviewing the PR @sdanyliv @MaceWindu 
I tried to make "logical" commits but in the end there were so many changes over time that it's better for you to just look at the final diff.
There are less changes than you might expect, albeit a few non-trivial ones. 😉

The last piece missing before merge and release is tests, of course. 
I'll start writing tests for every feature of this PR.

jods4:
There are no tests for Row constructors yet, but for my own sanity I want to check that everything else is still working ok, hence the run above 👆

jods4:
@MaceWindu @sdanyliv I'm done!
This is ready for final review and merging.

The remaining tests failure are unrelated to `SqlRow`, it looks like flacky tests whose results are inconsistent from one run to the next.

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+					Row<int?, int>(i.Three, i.Five)))
+				.Should().Be(0);
+		}
+
+#if NETFRAMEWORK
+		[ActiveIssue("Old SQLite version doesn't support Row Values", Configuration = ProviderName.SQLiteMS)]
+#endif
+		[Test]
+		public void Overlaps(
+			[IncludeDataSources(TestProvName.AllOracle /* TestProvName.AllPostgreSQL, ProviderName.DB2 */)] string context)
```
MaceWindu, line 460: pass `true` as first parameter to enable remote context tests

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+				  .Overlaps(Row(DT.Parse("2020-10-05"), (TimeSpan?)null)))
+				.Should().Be(1);
+		}
+
+#if NETFRAMEWORK
+		[ActiveIssue("Old SQLite version doesn't support Row Values", Configuration = ProviderName.SQLiteMS)]
+#endif
+		[Test]
+		public void EqualToSelect(
+			[IncludeDataSources(
```
MaceWindu, line 494: pass `true` as first parameter to enable remote context tests

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+				 select Row(y.One, y.One + 1, 4)))
+				.Should().Be(1);
+		}
+
+#if NETFRAMEWORK
+		[ActiveIssue("Old SQLite version doesn't support Row Values", Configuration = ProviderName.SQLiteMS)]
+#endif
+		[Test]
+		public void CompareToSelect(
+			[IncludeDataSources(
```
MaceWindu, line 538: pass `true` as first parameter to enable remote context tests

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+					table.Any(u => Row(2, u.Date) > Row(u.Int, t.Date)))
+				.Should().Be(1);
+		}
+
+#if NETFRAMEWORK
+		[ActiveIssue("Old SQLite version doesn't support Row Values", Configuration = ProviderName.SQLiteMS)]
+#endif
+		[Test]
+		public void UpdateRowLiteral(
+			[IncludeDataSources(ProviderName.DB2, TestProvName.AllPostgreSQL)] string context)
```
MaceWindu, line 601: pass `true` as first parameter to enable remote context tests

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+					new Ints { One = 1,   Two = 2,   Three = 3,   Four = 4,   Five = 5,  Nil = (int?)null },
+					new Ints { One = 100, Two = 200, Three = 300, Four = 400, Five = 50, Nil = 600 });
+		}
+
+#if NETFRAMEWORK
+		[ActiveIssue("Old SQLite version doesn't support Row Values", Configuration = ProviderName.SQLiteMS)]
+#endif
+		[Test]
+		public void UpdateRowSelect(
+			[IncludeDataSources(
```
MaceWindu, line 630: pass `true` as first parameter to enable remote context tests

Tests/Linq/Linq/SqlRowTests.cs
```diff
@@ -0,0 +1,695 @@
...
+using System.Linq;
+using FluentAssertions;
+
+using static LinqToDB.Sql;
+using DT = System.DateTime;
+using DTO = System.DateTimeOffset;
+
+namespace Tests.Linq
+{
+	[TestFixture]
```
MaceWindu, line 15: we need test that execute same query with different values in row to ensure query/parameter caching works properly for rows

MaceWindu (changes requested):

Source/LinqToDB/Data/DbCommandProcessor/DbCommandProcessorExtensions.cs
```diff
@@ -38,8 +38,8 @@ public static class DbCommandProcessorExtensions
 			Instance == null ? cmd.ExecuteNonQueryAsync(token) : Instance.ExecuteNonQueryAsync(cmd, token);
 
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static DbDataReader ExecuteReaderExt(this IDbCommand cmd, CommandBehavior commandBehavior) =>
-			Instance == null ? ((DbCommand)cmd).ExecuteReader(commandBehavior) : Instance.ExecuteReader((DbCommand)cmd, commandBehavior);
+		public static IDataReader ExecuteReaderExt(this IDbCommand cmd, CommandBehavior commandBehavior) =>
```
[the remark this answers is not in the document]
jods4: Good catch, this is part of my dev setup and shouldn't have been committed.
It's so that the connections can be mocked, which I find far faster and more convenient than working against a real DB during dev.

Source/LinqToDB/Sql/Sql.Row.cs
```diff
@@ -0,0 +1,170 @@
...
+
+			public override ISqlExpression? GetExpression<TContext>(TContext context, IDataContext dataContext, SelectQuery query,
+				Expression expression, Func<TContext, Expression, Mapping.ColumnDescriptor?, ISqlExpression> converter)
+			{				
+				string? exprStr = Expression;
+				PrepareParameterValues(expression, ref exprStr, true, out var knownExpressions, out var genericTypes);
+				
+				// The main purpose of this derived ExpressionAttribute is to validate that types are valid.
+				// SQL standard only defines OVERLAPS on couples of (nullable) dates, or date + interval types.
+				if (!IsValidRow(knownExpressions![0]!) || !IsValidRow(knownExpressions![1]!))
```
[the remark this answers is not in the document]
jods4: Ok, removed. It's much simpler as I don't need a custom attribute, then.

I think the main reason behind this was to provide extensions for non-date types and fallbacks for RDBMS that don't support OVERLAPS.

This was left out because the semantics of SQL-standard OVERLAPS are too complicated to emulate exactly. You can do a good enough approximation but I didn't want to make that kind of decisions in this PR, so I kept OVERLAPS native-only at the moment.

Tests/Linq/Linq/SqlRowTests.cs
jods4: Noted

Tests/Linq/Linq/SqlRowTests.cs
jods4: Not sure how many tests like that you want and what's the best approach.
Could you add those tests please?

jods4:
@MaceWindu I've done all the requested changes, except the one about tests with varying values, see above.

MaceWindu:
Ok, I've also found some issue with remote context.

MaceWindu:
@jods4 
I've made some fixes, but after marking SqlSetExpression.Column as nullable I've found several places where Rows not taken into account.

Could you please review each such place (now it gives compilation error) and add Rows support with tests or if it is not possible to have rows there - add supporession operator?

jods4:
@MaceWindu I fixed it.
All I had to do is made `Column` throw if the setter is a row expression.
That's the right thing to do in all the places where nullness isn't checked, because it means a row expression is not supported in this context (e.g. in an INSERT).

Codepaths that support rows do appropriate checks, so no other change was required.

MaceWindu:
With exception it again crash in query serializer

MaceWindu:
Also please restore nullability on Column property (with exception) and add suppressions on callsites. I want callers be aware of it being optional.

jods4:
@MaceWindu 
> With exception it again crash in query serializer. That's initial reason why I removed exception and made it nullable.

OK I missed the query serializer as a place that needs to support Row expressions.
As such, I think the proper way to go is to modify the query serializer to check for `Row == null` first.

> But there were around 10 more places where Row check were missing and I'm not sure they all cannot have Row set.

The only place `Row` is supported is inside an update and I tried to update all places related to that.
Query serializer is something I didn't consider as impacted and I was wrong. If there are others we should fix them as they come up.

> Also please restore nullability on Column property (with exception) and add suppressions on callsites. I want callers be aware of it being optional.

We can do that but it's a lot of callsites.
I think the current pattern is ok because if you don't support Row then Column is indeed not nullable.
If you support Row and don't check the nullable Row first, you'll get an exception (hopefully in tests).

MaceWindu:
Not really alot. as I said it's about 10 places where ! should be added

MaceWindu:
> As such, I think the proper way to go is to modify the query serializer to check for Row == null first.

yes, that's how it should be done with current code

jods4:
@MaceWindu I fixed the serialization.
Also made Column nullable as you requested. There was indeed only 12 places impacted, the build said 96 errors but that's because it builds 8 times :(
I kept the exception in anyway, because `Column` was not nullable before and that's how it's intended to be used. Incorrect usage should fail fast to catch bugs.

MaceWindu (approved):

Source/LinqToDB/SqlQuery/SqlSetExpression.cs
```diff
@@ -6,36 +6,93 @@ namespace LinqToDB.SqlQuery
...
+
 		public SqlSetExpression(ISqlExpression column, ISqlExpression? expression)
 		{
-			Column     = column;
+			_column = column;
+			Expression  = expression;
+			RefineDbParameter(column, expression);
+		}
+
+		public SqlSetExpression(ISqlExpression[] row, ISqlExpression? expression)
```
sdanyliv, line 20: Why you have introduced Row? Why not just reuse Column property? SqlRow is already `ISqlExpression`. Everything should be be simpler.

jods4: `Column` is expected to be a single column field, as the name implies. There are 40+ places that reference it with this assumption.

`Row` is a clear name that explicitly tells you it's not a single field setter but a row value setter. It's already an array of expression which is the most straightforward to use. 
Most importantly: it's makes it easy to see which code supports rows (just a handful) and which code is oblivious of rows (most of the 40+ others). In fact Column contains a fail-fast when that assertion is wrong (as was the case of the serializer that I forgot).

jods4: @sdanyliv 
And I forgot the _main_ reason.
This would not be a plain `SqlRowExpression` straight from query.
Look at the update builder and at `GetField()` in particular. This is how the array of expressions is built.
That part was tricky for me and I don't feel confident changing it much.

MaceWindu:
@sdanyliv failing test generates `(:p_1, :p_2) OVERLAPS ((:p_3, :p_4))` but should generate `(:p_1, :p_2) OVERLAPS (:p_3, :p_4)`
@jods, could you clarify when double brackets needed for oracle?

jods4:
@MaceWindu from my experiments, when a row value is on the right-hand side of a binary operator, e.g.:
```
(1, 2) = ( (1, 2) )
(1, 2) <= ( (1, 2) )
```
I believe Oracle parser has an ambiguity at this position and thinks brackets are the usual grouping/precedence brackets.

I see @sdanyliv has refactored `OVERLAPS` into a binary operator. I think this is why it's now handled by my double-bracketing code in Oracle provider.

The relevant code is here:
https://github.com/linq2db/linq2db/pull/3339/files#diff-82c262d6464c4218de01a7298f57148dc8c3a3ad973baf9eaccf24529d3c28ebR255-R267

Maybe you should pass the operator and don't double-bracket `OVERLAPS`.

viceroypenguin (approved):

MaceWindu:
Fix #2204

MaceWindu:
and fix #2657

sdanyliv (approved):

PR #3343, opened 2021-11-13
Update dependencies

MaceWindu:
- Update [Npgsql](https://www.nuget.org/packages/Npgsql): 4.1.9 -> 4.1.10, 5.0.10 -> 5.0.11+6.0.0
- Npgsql 6.0.0 support changes:
  - Explicitly set parameter type for `DataType.DateTime` and `DataType.DateTime2` (used by default for `DateTime`) to `timestamp` to prevent `Npgsql` from treating it as `timestamptz`, when it is not, and fail
  - automatically `Offset`/`Kind` for `DateTime`/`DateTimeOffset` values , used to pass `timestamp`/`timestamptz` values, to simplify migration to `npgsql` 6 for users. If user wants to fix his data himself and receive errrors when he pass invalid data, he can disable those fixes using `PostgreSQLTools.NormalizeTimestampData = false;` setting
    - Reset offset to zero for `DateTimeOffset` values, passed to native `BulkCopy` or as query parameter to prevent `Nnpgsql` complains about non-zero offset
    - Update `Kind` for `DateTime` parameters of `timestamp` type from `UTC` to `Unspecified` to prevent `Nnpgsql` complains about it
    - Update `Kind` for `DateTime` parameters of `timestamptz` type from non-`UTC` to `UTC` to prevent `Nnpgsql` complains about it
    - for other .net types, used to pass `timestamp`/`timestamptz` values (e.g. `NpgsqlRange<T>`), user will need to fix their data manually to comply with `npgsql` 6 requrements
- updated `PostgreSQL` range tests to use `DateTime` values with proper `Kind`
- converted "add/get Hour(s)" tests for `DateTime` and `DateTimeOffset` to evaluate completely on server side to avoid inconsistent results due to UTC <-> local conversions

Remaining tasks:
- support new npgsql 6 APIs
- try to update to M.D.Sqlite 6 once more

### DIFF
--- a/Build/Azure/README.md
+++ b/Build/Azure/README.md
@@ -97,13 +97,13 @@ Legend:
 |MySQL 8<br>[MySql.Data](https://www.nuget.org/packages/MySql.Data/) 8.0.27|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
 |MySQL 8<br>[MySqlConnector](https://www.nuget.org/packages/MySqlConnector/) 0.69.10/1.3.14/2.0.0|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
 |MariaDB 10<br>[MySql.Data](https://www.nuget.org/packages/MySql.Data/) 8.0.27|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 9.2<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 9.3<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 9.5<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 10<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 11<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
-|PostgreSQL 12<br>[Npgsql](https://www.nuget.org/packages/Npgsql/) 4.1.9 (netfx) / 5.0.10 (core)|:x:|:x:|:heavy_check_mark:|:heavy_check_mark:|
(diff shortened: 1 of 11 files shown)


MaceWindu:
@linq2db, guys, your opinion on `PostgreSQLTools.NormalizeTimestampData` setting - should be enable it by default so users could migrate to npgsql 6 easier or disable to encourage users to use proper types?

MaceWindu:
@linq2db/developers 
Still need to detect failing sqlite concurrency tests and fix them, but in general it is ready for review.
Also waiting for your opinion on default value for `PostgreSQLTools.NormalizeTimestampData`

sdanyliv (approved):

Source/LinqToDB/DataProvider/SQLite/SQLiteTools.cs
```diff
@@ -159,6 +159,30 @@ public static void DropDatabase(string databaseName)
...
 		}
 
+		/// <summary>
+		/// Invokes ClearAllPools() method for specified provider.
+		/// </summary>
+		/// <param name="provider">For which provider ClearAllPools should be called:
+		/// <list type="bullet">
+		/// <item><see cref="ProviderName.SQLiteClassic"/>: System.Data.SQLite</item>
+		/// <item><see cref="ProviderName.SQLiteMS"/>: Microsoft.Data.Sqlite</item>
+		/// <item><c>null</c>: both (any)</item>
```
viceroypenguin, line 169: with the `else if`, this will only do `SQLiteMS` if the provider is `null`. Either code or comment should change?

viceroypenguin (approved):

CumpsD:
This is quite an old PR I'm digging up, but I have a question.

We have a query which looks like this:

```csharp
var row = await connection.GetTable<SnapshotRow>()
    .Where(r =>
        r.PersistenceId == persistenceId &&
        r.SequenceNumber <= sequenceNr &&
        r.Created <= timestamp)
    .OrderByDescending(t => t.SequenceNumber)
    .FirstOrDefaultAsync();
```

Where Created is:
```csharp
[Column(DataType = DataType.DateTime2)]
public DateTime Created { get; set; }
```

But mapped with:
```csharp
builder
    .Member(r => r.Created)
    .HasDataType(DataType.Int64)
    .HasConversion(
        r => r.Ticks,
        r => new DateTime(r));
```

Prior to this PR, the query which got generated looked like:

```sql
SELECT
	[r].[persistence_id],
	[r].[sequence_number],
	[r].[created],
	[r].[snapshot],
	[r].[manifest],
	[r].[serializer_id]
FROM
	[snapshot] [r]
WHERE
	[r].[persistence_id] = @persistenceId AND [r].[sequence_number] <= @sequenceNr AND
	[r].[created] <= @timestamp_1
ORDER BY
	[r].[sequence_number] DESC
LIMIT @take
```

But after this PR it looks like:
```sql
SELECT
	[r].[persistence_id],
	[r].[sequence_number],
	[r].[created],
	[r].[snapshot],
	[r].[manifest],
	[r].[serializer_id]
FROM
	[snapshot] [r]
WHERE
	[r].[persistence_id] = @persistenceId AND [r].[sequence_number] <= @sequenceNr AND
	DateTime([r].[created]) <= DateTime(@timestamp_1)
ORDER BY
	[r].[sequence_number] DESC
LIMIT @take
```

Something seems to have happened to `DateTime([r].[created]) <= DateTime(@timestamp_1)` while the only thing that has changed is bumping from 3.5.2 to 3.6.0. It is as if the conversion is ignored?

jogibear9988:
@CumpsD maybe you should create an issue instead of commenting on a closed pull req.

CumpsD:
@jogibear9988 I will, I tried to ask as close as possible to the relevant change first

CumpsD:
@jogibear9988 I will, I tried to ask as close as possible to the relevant change first

I created an issue at https://github.com/linq2db/linq2db/issues/4038 for this sqlite conversion

PR #3349, opened 2021-11-14
test for issue 3342 - need npgsql version 6 to fail

jogibear9988:


MaceWindu:
Well, it will not work - I don't see mapping of `DateTime` to `tstz` type

jogibear9988:
???

you wrote:

but I'm not sure why you need to use dbtype. This should be enough for you:

    this.MappingSchema.AddScalarType(typeof(DateTimeOffset), DataType.DateTimeOffset);
    this.MappingSchema.AddScalarType(typeof(DateTimeOffset?), DataType.DateTimeOffset);

jogibear9988:
okay if I use this:

                     ms.AddScalarType(typeof(DateTime), new SqlDataType(DataType.DateTimeOffset, typeof(DateTime), "timestamptz"));
			ms.AddScalarType(typeof(DateTime?), new SqlDataType(DataType.DateTimeOffset, typeof(DateTime), "timestamptz"));

it works.
didn't do it in first place cause you wrote the other one should be enough. i think i missunderstod...

Issue #3314, opened 2021-11-04
[SQL Server] Implement ISNULL optimization


MaceWindu:
`ISNULL` is a built-in SQL Server function, which calculates parameters only once, `COALESCE` (and `IIF`) is a "macro", converted to `CASE` statement.
For complex parameters like subqueries `CASE` leads to double evaludation of parameters.

We already have tests that demonstate this behavior:
- `SelectTests.Coalesce4`
- `SelectTests.Coalesce5`

marcnet80:
Hello, @MaceWindu. Could you please reivew PR?
Thanks.

PR #3353, opened 2021-11-17
Feature isnull sqlserver

marcnet80:
https://github.com/linq2db/linq2db/issues/3314


MaceWindu:
@marcnet80 , thanks!

I will accept it, but note that this is not a fix for #3314 - #3314 is about automatic generation of `ISNULL` isntead of `COALESCE` in sql queries and is more complex.

For #3314 we cannot just detect complex `COALESCE` parameters and replace it with `ISNULL` - it's an easy part. Trick with `ISNULL` is that it assigns result type using type of first parameter, which could result in data loss when second parameter returned (when parameter types differ, e.g. by length for strings).

I actually have fix for #3314 but it requires tests and I currently busy with other tasks.

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
```diff
@@ -774,5 +774,16 @@ public static bool ContainsPropertyWithLanguage(this ISqlServerExtensions? ext,
...
+		/// <summary>
+		/// Generates 'ISNULL( value, replacementValue )' function.
+		/// </summary>
+		/// <typeparam name="T">Generic type.</typeparam>
+		/// <param name="ext">Extension point.</param>
+		/// <param name="value">Value to test whether is NULL.</param>
+		/// <param name="replacementValue">Value to replace.</param>
+		/// <returns>Function returns a replacementValue if the value is NULL.</returns>
+		[Sql.Extension("ISNULL({value}, {replacementValue})", ServerSideOnly = true)]
+		public static T IsNull<T>(this ISqlServerExtensions? ext, [ExprParameter] T? value, [ExprParameter] T replacementValue) => throw new LinqException($"'{nameof(IsNull)}' is server - side method.");
```
MaceWindu, line 787: Why `replacementValue` is not nullable? It should be possible to use nullable value here too.

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
```diff
@@ -774,5 +774,16 @@ public static bool ContainsPropertyWithLanguage(this ISqlServerExtensions? ext,
...
+
+		/// <summary>
+		/// Generates 'ISNULL( value, replacementValue )' function.
+		/// </summary>
+		/// <typeparam name="T">Generic type.</typeparam>
+		/// <param name="ext">Extension point.</param>
+		/// <param name="value">Value to test whether is NULL.</param>
+		/// <param name="replacementValue">Value to replace.</param>
+		/// <returns>Function returns a replacementValue if the value is NULL.</returns>
+		[Sql.Extension("ISNULL({value}, {replacementValue})", ServerSideOnly = true)]
```
MaceWindu, line 786: Would be nice to have `IsNullableType` set here, but I don't see `IfAllParametersNullable` [here](https://github.com/linq2db/linq2db/blob/cf1c77b45b7734f9dcf31970d6efee78ac2ba11e/Source/LinqToDB/Sql/IsNullableType.cs). If you can add it (and use it in nullability detenction logic) - it would be great.

MaceWindu (changes requested):

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
marcnet80: I wanted to make this function more strict, but I ok with you comment.

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
marcnet80: added

MaceWindu (approved):

Source/LinqToDB/Sql/IsNullableType.cs
```diff
@@ -14,6 +14,7 @@ public enum IsNullableType
 			SameAsSecondParameter,
 			SameAsThirdParameter,
 			SameAsLastParameter,
+			IfAllParametersNullable
```
viceroypenguin, line 17: Reduce future code difference when new enum values are added.

```suggestion
			IfAllParametersNullable,
```

Source/LinqToDB/Sql/Sql.ExpressionAttribute.cs
```diff
@@ -193,6 +193,7 @@ protected bool GetCanBeNull(ISqlExpression[] parameters)
 					case IsNullableType.SameAsThirdParameter   : return SameAs(2);
 					case IsNullableType.SameAsLastParameter    : return SameAs(parameters.Length - 1);
 					case IsNullableType.IfAnyParameterNullable : return parameters.Any(static p => p);
+					case IsNullableType.IfAllParametersNullable: return parameters.All(static p => p);
 				}
```
viceroypenguin, line 197: @MaceWindu should we suggest a switch to the new `switch`-expression here? Not sure you're opinion on those.

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
```diff
@@ -774,5 +774,16 @@ public static bool ContainsPropertyWithLanguage(this ISqlServerExtensions? ext,
...
+
+		/// <summary>
+		/// Generates 'ISNULL( value, replacementValue )' function.
+		/// </summary>
+		/// <typeparam name="T">Generic type.</typeparam>
+		/// <param name="ext">Extension point.</param>
+		/// <param name="value">Value to test whether is NULL.</param>
+		/// <param name="replacementValue">Value to replace.</param>
+		/// <returns>Function returns a replacementValue if the value is NULL.</returns>
+		[Sql.Extension("ISNULL({value}, {replacementValue})", ServerSideOnly = true, IsNullable = Sql.IsNullableType.IfAllParametersNullable)]
```
viceroypenguin, line 786: @MaceWindu  I know you suggested `IfAllParametersNullable`, but I think this should actually just be `SameAsLastParameter`. The whole point of this function on the server side is that if the first value is null, we take the second value, so only the last parameter's nullability is relevant for the nullability of the expression, right?

MaceWindu: Don't think this is correct. E.g. for call with such parameters `ISNULL(not-nullable, nullable)` it will return true, when it should be false. I agree that in that case there is no need for function call, but it's how it could be called by user anyway

Source/LinqToDB/Sql/Sql.ExpressionAttribute.cs
MaceWindu: I have no strong opinion on that. If @marcnet80 wants, he can change it

Source/LinqToDB/DataProvider/SqlServer/SqlServerExtensions.cs
viceroypenguin: Oh. I was reading it as `AnyParametersNullable`, not `AllParametersNullable`. I'm good then.

viceroypenguin (approved):

Source/LinqToDB/Sql/IsNullableType.cs
marcnet80: added

Source/LinqToDB/Sql/Sql.ExpressionAttribute.cs
marcnet80: @viceroypenguin, changed to switch-expression.

Tests/Linq/Linq/IsNullTests.SqlServer.cs
```diff
@@ -0,0 +1,53 @@
...
+		public void TestProperty_IsNull_ServerSideOnly(
+		[IncludeDataSources(TestProvName.Northwind)] string context)
+		{
+			string defaultCategory = "test";
+			using (var db = new NorthwindDB(context))
+			{
+				var query = (from c in db.Category
+							 select Sql.Ext.SqlServer().IsNull(c.CategoryName, defaultCategory));
+
+				Assert.That(query.ToString().Contains("ISNULL([c_1].[CategoryName], @defaultCategory)"));
```
sdanyliv, line 29: Execute query, then get `db.LastQuery`, we need to ensure that database accepts that and query logged for baselines.

Tests/Linq/Linq/IsNullTests.SqlServer.cs
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		[Test]
+		public void TestStatementWithOrderBy_IsNull_ServerSideOnly(
+		[IncludeDataSources(TestProvName.Northwind)] string context)
+		{
+			int  categoryId = 1;
+			using (var db = new NorthwindDB(context))
+			{
+				var supplierQuery = GetSupplierIdWithMaxUnitPrice(categoryId, db);
```
sdanyliv, line 40: Same, execute query.

Tests/Linq/Linq/IsNullTests.SqlServer.cs
marcnet80: added

Tests/Linq/Linq/IsNullTests.SqlServer.cs
marcnet80: added

MaceWindu:
@marcnet80  thanks!

PR #3366, opened 2021-11-22
Add net6.0 support

### LINKED ISSUE #3320: Review use of Assembly.CodeBase and friends propertis in linq2db
CodeBase property is also used in LinqToDB.Common.Tools.GetFileName, please consider to refactor it too

_Originally posted by @pavlo-dudka in https://github.com/linq2db/linq2db/issues/3319#issuecomment-962495871_

This property alongside with some others obsoleted in .net core or even can throw exceptions. We need to review our code in places where we use them.

### LINKED ISSUE #3358: Add net6/C# 10 support
- enable C# 10 (we need it at least for `record struct` support testing)
- add net6 target to tests and CI
- add net6 target (build-only, no release) to `linq2db` to detect nullability issues with recent NRT annotations

Regarding last one - currently don't see a reason to release `net6` version as we don't use anything from it. Could be added when we decide to add built-in `DateOnly`/`TimeOnly` mappings.

Related issues:
- #2977
- #3357

MaceWindu:
Fix #3358
Fix #3320

- enable C# 10 for all projects
- add net6.0 test target
- add net6.0 build target to source projects (just for NRT validation on build, no nuget deployment yet)
- switch to warning as error build mode (no more NRT warnings oversight!)
- fixed/suppressed new NRT warnings
- update pipelines to support windows-2022 image
- add net6.0 build and testing support to pipelines
- remove ubuntu 18 from test matrix
- add pgsql 14 to test matrix


Source/LinqToDB/LinqToDB.csproj
```diff
@@ -6,7 +6,7 @@
 		<RootNamespace>LinqToDB</RootNamespace>
 
 		<DocumentationFile>bin\$(Configuration)\$(TargetFramework)\linq2db.xml</DocumentationFile>
-		<TargetFrameworks>net472;net45;net46;netstandard2.0;netcoreapp2.1;netstandard2.1;netcoreapp3.1</TargetFrameworks>
+		<TargetFrameworks>net45;net46;net472;netstandard2.0;netstandard2.1;netcoreapp2.1;netcoreapp3.1;net6.0</TargetFrameworks>
```
viceroypenguin, line 9: There is an `#ifdef` that refers to .net5, but it is not listed here. Inconsistency?

viceroypenguin (approved):

Source/LinqToDB/LinqToDB.csproj
MaceWindu: we don't build linq2db for net5

viceroypenguin: Sure, but that means the check for `#ifdef NET5_0_OR_GREATER` should be `#ifdef NET6_0_OR_GREATER`, since we don't compile for net5. I guess what I'm asking is why the check for 5 when we don't build for 5?

MaceWindu: nullableattributes.cs is a shared file

viceroypenguin: Ah, didn't know that. Good enough for me then. :)

sdanyliv (approved):

PR #3367, opened 2021-11-23
Address multiple issues with SET queries

MaceWindu:


MaceWindu:
I will split this pr into smaller PRs as it starting to get too many changes and baselines became unreviewable

PR #3021, opened 2021-05-29
Fix for #2895. More aggressive eager loading detection.

sdanyliv:
Fixes #2895

Issue #2895, opened 2021-03-16
Linq2db fails with "The method or operation is not implemented."


v-zubritsky:
Hi! I encounter `NotImplementedException` from [here](https://github.com/linq2db/linq2db/blob/d9832b6e8dd60e4946e26c07f5e4ac241d167f11/Source/LinqToDB/Linq/Builder/SelectContext.cs#L876) I believe. 

Model is probably not in the simplest form to reproduce the error, but I tried to remove all the unnecessary details.

`The method or operation is not implemented.`
```
   at LinqToDB.Linq.Builder.SelectContext.GetContext(Expression expression, Int32 level, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.PassThroughContext.GetContext(Expression expression, Int32 level, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.SelectContext.<>c__DisplayClass55_0.<GetContext>b__0(IBuildContext ctx, Expression ex, Int32 l)
   at LinqToDB.Linq.Builder.SelectContext.ProcessScalar[T](Expression expression, Int32 level, Func`4 action, Func`1 defaultAction, Boolean throwOnError)
   at LinqToDB.Linq.Builder.SelectContext.GetContext(Expression expression, Int32 level, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.SelectManyBuilder.SelectManyContext.GetContext(Expression expression, Int32 level, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.TableBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.SelectManyBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.SelectManyBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
[...]

PR #3384, opened 2021-12-09
Revert "Fix for #2895. More aggressive eager loading detection. (#3021)"

Shane32:
This reverts commit 8fa97de6a45225e70de98979c5d69fede0973ae4.

The reason for the reversion is the regressions listed in #3160 .  A separate PR will be issued to add tests to ensure the regression does not reassert itself, possibly along with minor fixes.


MaceWindu:
What about #2895? At least we shouldn't revert tests (just disable them if they still fail without fix)

Shane32:
I can certainly add a disabled copy of the other test to my subsequent PR. I figured for cleanliness it would be best if this PR was exactly a revert of the other PR. I’m open to suggestions.

Shane32:
@sdanyliv What's the next step here?

PR #3293, opened 2021-10-25
Fix for #3259. Improved aggregates handling.

sdanyliv:
Fixes #3259

PR #3386, opened 2021-12-16
Fix regression, introduced by 3.5.2

MaceWindu:
Introduced by #3293


MaceWindu:
While it doesn't break tests, it reintroduced duplicate columns for same subqueries (e.g. Group* tests). Maybe we can fix it too

PR #3411, opened 2022-01-29
SqlServer System schemas

igor-tkachev:
Data Model of 'INFORMATION_SCHEMA' and 'sys' schemas.


MaceWindu (changes requested):
- What about other providers?
- I think some kind of schema versioning needed and new db versions add/modify/remove schema/system objects

igor-tkachev:
If we need other providers we can implemented it if we get doc/info. Also it should be different PR. Otherwise, we will never finish it as the hints PR.

igor-tkachev:
> * What about other providers?

No plans for other providers yet.

MaceWindu:
Let's move it to 4.0 (rebase needed). We will release 3.7 next week and there is no time to finish this PR and review it.

Source/LinqToDB.Templates/LinqToDB.ttinclude
```diff
@@ -245,9 +245,15 @@ void GenerateTypesFromMetadata()
...
 		var tableAttrs = new List<string>();
 
-		if (ServerName   != null) tableAttrs.Add("Server="    + ToStringLiteral(ServerName));
+
+		if (GenerateDatabaseNameFromTable && t.TableSchema.CatalogName != null)
+		{
+			tableAttrs.Add("Database="  + ToStringLiteral(t.TableSchema.CatalogName));
+		}
+		else
 		if (DatabaseName != null) tableAttrs.Add("Database="  + ToStringLiteral(DatabaseName));
```
viceroypenguin, line 254: This looks questionable. you have an `if`/`else`, but without brackets, this first `if` is actually an `else if`, and the other `if`s here are not part of the else. Either a) line 253 should be explicitly `else if`, or there should be brackets around 254-256.

Source/LinqToDB/Sql/Sql.FunctionAttribute.cs
```diff
@@ -90,12 +90,12 @@ public FunctionAttribute(string configuration, string name, params int[] argIndi
...
 			{
 				var expressionStr = Expression;
-				PrepareParameterValues(expression, ref expressionStr, true, out var knownExpressions, out var genericTypes);
+				PrepareParameterValues(expression, ref expressionStr, true, out var knownExpressions, IgnoreGenericParameters, out var genericTypes);
 
 				if (string.IsNullOrEmpty(expressionStr))
 					throw new LinqToDBException($"Cannot retrieve function name for expression '{expression}'.");
 
-				var parameters = PrepareArguments(context, expressionStr!, ArgIndices, addDefault: true, knownExpressions, genericTypes, converter);
+				var parameters = PrepareArguments(context, expressionStr!, ArgIndices, true, knownExpressions, genericTypes, converter);
```
viceroypenguin, line 98: Preferably, `true`/`false` parameters are named, so that it is obvious to future readers what the parameter is specifying. Is there a reason to remove the naming here?

viceroypenguin (changes requested):

Source/LinqToDB/Sql/Sql.FunctionAttribute.cs
igor-tkachev: ok

Source/LinqToDB.Templates/LinqToDB.ttinclude
igor-tkachev: fixed

viceroypenguin (approved):

PR #3414, opened 2022-02-03
Performance improvement

igor-tkachev:


MaceWindu (approved):

viceroypenguin (approved):

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1645,14 +1661,12 @@ public StringComparer ColumnNameComparer
 		/// <returns>Mapping descriptor.</returns>
 		public EntityDescriptor GetEntityDescriptor(Type type)
```
MaceWindu, line 1662: this method called a lot and was creating a lot of closure objects. With this change it's back due to use of `this`

PR #3424, opened 2022-02-14
Fix #3423 LEAD and LAG default parameter type

### LINKED ISSUE #3423: Bad typing of LEAD and LAG default values
The `default` parameter of LEAD and LAG analytical functions is defined as `int`.
This is incorrect, as the `default` parameter should have the same generic type `T` as the main expression.

https://github.com/linq2db/linq2db/blob/8200bce088a9001bf794e920ed27ac4c9835b6fa/Source/LinqToDB/Sql/Sql.Analytic.cs#L717-L745

 This was unnoticed because the unit tests in `AnalyticTests.cs` all use int expressions...

jods4:
Fix #3423 
This PR fixes the type of `default` in `LEAD` and `LAG`.

It comes with a unit test that uses string as `default` in addition to existing tests that used `int`.

I've also added more `LEAD` and `LAG` convenience overloads because it's pretty uncommon to specify the null modifier.
Now all these are possible: `Lag(expr)`, `Lag(expr, 3)`, `Lag(expr, 3, "default")`, which mirrors the native SQL function specs.

### DIFF
--- a/Tests/Linq/Linq/AnalyticTests.cs
+++ b/Tests/Linq/Linq/AnalyticTests.cs
@@ -1710,8 +1710,8 @@ namespace Tests.Linq
 			TestProvName.AllSybase,
 			ProviderName.SqlCe,
 			TestProvName.AllAccess,
-			// Firebird excluded because of #2839, the test data is inserted with padding and then expectations fail
-			// ProviderName.Firebird,
+			// All Firebird excluded because of #2839, test data is inserted with padding and then expectations fail
+			TestProvName.AllFirebird,
 			TestProvName.MySql55)] string context)
 		{
 			var data = new Issue1799Table3[] 
--- a/Tests/Linq/Linq/AnalyticTests.cs
+++ b/Tests/Linq/Linq/AnalyticTests.cs
@@ -1703,5 +1703,50 @@ namespace Tests.Linq
 				CollectionAssert.AreEqual(new[] { "None", "One" }, lags);
 			}
 		}
+		
+		[Test]
+		public void LeadLagOverloads([DataSources(
+			TestProvName.AllSqlServer2008Minus,
+			TestProvName.AllSybase,
+			ProviderName.SqlCe,
+			TestProvName.AllAccess,
+			ProviderName.Firebird,
+			TestProvName.MySql55)] string context)
+		{
+			var data = new Issue1799Table3[] 
+			{
+				new() { ProcessID = 1, ProcessName = "One" },
+				new() { ProcessID = 2, ProcessName = "Two" },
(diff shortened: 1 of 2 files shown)


Tests/Linq/Linq/AnalyticTests.cs
```diff
@@ -1668,5 +1668,40 @@ select new
 					.ToList();
 			}
 		}
+
+		[Test]
+		public void Issue3423LeadLagDefault([DataSources(
```
sdanyliv, line 1673: Please, do not use IssueXXXTest, just normal name for LAG, LEAD test

jods4: Sure, I've changed the name. 
There were lots of IssueXXXTest just before this one so I thought I would follow suit. 🤷‍♂️

jods4:
Tests are green now! 
Except `EagerLoadMultiLevel` in SQLite, but that's totally unrelated!?! Flaky test?

MaceWindu:
yep

sdanyliv:
6 functions are added, but tested only 2 of them.

jods4:
@sdanyliv yeah, I added the new overloads after the fact, for convenience. 
They're all the same but copy-paste mistakes happen, so I quickly added one new tests for the new overloads right now.

MaceWindu (approved):

MaceWindu:
LeadLagOverloads test fails for Firebird
```
Expected and actual are both <System.String[4]>
Values differ at index [1]
Expected string length 4 but was 5. Strings differ at index 4.
Expected: "Four"
But was:  "Four "
```

MaceWindu:
Looks like issue with `char(x)` vs `varchar(x)` types. Just mark test disabled for firebird and create issue to track it

MaceWindu:
Actually, just add note to https://github.com/linq2db/linq2db/issues/2839 - for some reason all strings are padded to max length with space here too

Tests/Linq/Linq/AnalyticTests.cs
```diff
@@ -1668,5 +1668,86 @@ select new
...
+		}
+		
+		[Test]
+		public void LeadLagOverloads([DataSources(
+			TestProvName.AllSqlServer2008Minus,
+			TestProvName.AllSybase,
+			ProviderName.SqlCe,
+			TestProvName.AllAccess,
+			// Firebird excluded because of #2839, the test data is inserted with padding and then expectations fail
+			// ProviderName.Firebird,
```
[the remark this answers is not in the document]
jods4: Incl. / Excl. always confuses me :(
I made the change.

PR #3427, opened 2022-02-17
Added .NET 6.0+ Hot Reload compatibility

jhudler:
Due to the heavy use of Reflection in static (global) variables in linq2db, running queries after a Hot Reload may issue exceptions due to mismatch metadata info. This pull request re-reads and updates reflection data in such variables every time a hot reload is triggered, avoiding such exceptions.

More info: https://github.com/dotnet/sdk/issues/23736

### DIFF
--- a/Source/LinqToDB/Data/RecordReaderBuilder.cs
+++ b/Source/LinqToDB/Data/RecordReaderBuilder.cs
@@ -5,6 +5,10 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 
+#if NET6_0_OR_GREATER
+[assembly: System.Reflection.Metadata.MetadataUpdateHandler(typeof(LinqToDB.Data.RecordReaderBuilder))]
+#endif
+
 namespace LinqToDB.Data
 {
 	using Expressions;
@@ -16,8 +20,8 @@ namespace LinqToDB.Data
 
 	class RecordReaderBuilder
 	{
-		public static readonly ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
-		public        readonly ParameterExpression DataReaderLocal;
+		public static   ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
+		public readonly ParameterExpression DataReaderLocal;
 
 		public readonly List<ParameterExpression>  BlockVariables   = new ();
 		public readonly List<Expression>           BlockExpressions = new ();
@@ -410,5 +414,11 @@ namespace LinqToDB.Data
 			return expression;
 		}
 
+		#region Hot reload compatibility
+		private static void ClearCache(Type[]? updatedTypes) // Hot reload compatibility
+		{
+			DataReaderParam = Expression.Parameter(typeof(IDataReader),  "rd");
+		}
+		#endregion
 	}
 }
--- a/Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
+++ b/Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
@@ -9,6 +9,10 @@ using System.Reflection;
 
 using JetBrains.Annotations;
 
+#if NET6_0_OR_GREATER
+[assembly: System.Reflection.Metadata.MetadataUpdateHandler(typeof(LinqToDB.Linq.Builder.ExpressionBuilder))]
+#endif
+
 namespace LinqToDB.Linq.Builder
 {
 	using Common;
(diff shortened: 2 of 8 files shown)


MaceWindu:
ping @igor-tkachev , @sdanyliv 

We need to discuss it first
1. we have more places with metadata cache
2. it's a pity microsoft push maintenance of non-production scenarios to users

MaceWindu:
My opinion - we can support it, just need to review more places where we cache metadata like `Methods` class. But I fear it will become clown fiesta in long-term run: new code added, nobody checks for hotreload support, hotreload broken again...

to11mtm:
Throwing my thoughts in, We should keep an eye on perf with these changes. It will be interesting to see how removing readonly impacts JIT optimizer

igor-tkachev:
Can we get some test showing this issue?

jogibear9988:
> Throwing my thoughts in, We should keep an eye on perf with these changes. It will be interesting to see how removing readonly impacts JIT optimizer

Maybe we should keep the readonly and assign via Reflection in the ClearCache function

jogibear9988:
> > Throwing my thoughts in, We should keep an eye on perf with these changes. It will be interesting to see how removing readonly impacts JIT optimizer
> 
> Maybe we should keep the readonly and assign via Reflection in the ClearCache function

The question is, will jitted methods be thrown away when this happens?

jogibear9988:
Found this: https://stackoverflow.com/questions/54380905/why-after-changing-a-static-readonly-field-via-reflection-the-output-of-that-rea so my idea would not work...

jods4:
I hate that initialization and refresh are duplicated code.

This is gonna lead to bugs because of copy-paste mistakes, or a new contributor will see the member but not the refresh and introduce new hot-reload incompatible code, as @MaceWindu said.

I'd like it better if there was a static ctor that called `ClearCache` to initialize instead.

jods4:
Can someone provide a bit more details about hot reload inner workings?
Some of the changes in this PR seem unlikely to happen...

The callback receives `Type[] types` that are types that were modified by user and reloaded, so whose metadata was invalidated, correct?

How can a user of linq2db modify the `ConvertBuilder` class or a .net framework class??
Users can only modify the code they own, can't they?

Source/LinqToDB/Data/RecordReaderBuilder.cs
```diff
@@ -16,8 +20,8 @@ namespace LinqToDB.Data
 
 	class RecordReaderBuilder
 	{
-		public static readonly ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
-		public        readonly ParameterExpression DataReaderLocal;
+		public static   ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
+		public readonly ParameterExpression DataReaderLocal;
```
viceroypenguin, line 24: Changing a `public static readonly` field to be `public static` can introduce bugs in the future. Currently, the internal expectation is that no other classes (including in l2db classes) will change these fields; however, dropping the `readonly` allows other classes to change them.

I would recommend changing to the following:
```suggestion
		public  static  ParameterExpression DataReaderParam  => DataReaderLocal;
		private static ParameterExpression DataReaderLocal   =  Expression.Parameter(typeof(IDataReader), "rd");
```

And updating `ClearCache` to update the private field instead.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -148,13 +152,13 @@ public static void AddBuilder(ISequenceBuilder builder)
...
-		public static readonly ParameterExpression DataContextParam = Expression.Parameter(typeof(IDataContext), "dctx");
-		public static readonly ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
-		public        readonly ParameterExpression DataReaderLocal;
-		public static readonly ParameterExpression ParametersParam  = Expression.Parameter(typeof(object[]),     "ps");
-		public static readonly ParameterExpression ExpressionParam  = Expression.Parameter(typeof(Expression),   "expr");
+		public static ParameterExpression QueryRunnerParam = Expression.Parameter(typeof(IQueryRunner), "qr");
+		public static ParameterExpression DataContextParam = Expression.Parameter(typeof(IDataContext), "dctx");
+		public static ParameterExpression DataReaderParam  = Expression.Parameter(typeof(IDataReader),  "rd");
+		public static ParameterExpression ParametersParam  = Expression.Parameter(typeof(object[]),     "ps");
+		public static ParameterExpression ExpressionParam  = Expression.Parameter(typeof(Expression),   "expr");
```
viceroypenguin, line 161: See comment in `RecordReaderBuilder.cs`; make these properties that refer to a `private static` field that can be updated by `ClearCache`.

Source/LinqToDB/Data/RecordReaderBuilder.cs
Shane32: Makes sense but changing a field to a property is a binary breaking change, correct?

viceroypenguin: Technically yes, but only for code directly accessing it. Since this class (and the other one commented) are both `internal`, then we can safely assume that no consumer code accesses these fields. As such, it would not be an externally breaking change.

However, this is why all `public` values should be properties, even if the class is internal and the property is an auto-property that will get JIT-ted away - it means that implementation can change to a regular field if changes are necessary.

Shane32:
> > > Throwing my thoughts in, We should keep an eye on perf with these changes. It will be interesting to see how removing readonly impacts JIT optimizer
> > 
> > 
> > Maybe we should keep the readonly and assign via Reflection in the ClearCache function
> 
> The question is, will jitted methods be thrown away when this happens?

Yeah, logically, either the JIT optimizer recognizes the readonly flag and assumes the values are static, in which case writing to them via reflection will not be seen in compiled code, or it doesn't, in which case there is no performance penalty to removing `readonly`.  So might as well remove `readonly` (if need be)

Shane32:
> How can a user of linq2db modify the `ConvertBuilder` class or a .net framework class?? Users can only modify the code they own, can't they?

My thoughts as well.  It would seem at first glance that making `Expression.Parameter(typeof(IDataContext), "dctx");` auto-reload should never occur for an end user as `typeof(IDataContext)` should never change for them -- and us linq2db devs can do without hot-reload (especially when there's a trade-off, like making a `static readonly` field writable).

viceroypenguin:
> > > > Throwing my thoughts in, We should keep an eye on perf with these changes. It will be interesting to see how removing readonly impacts JIT optimizer
> > > 
> > > 
> > > Maybe we should keep the readonly and assign via Reflection in the ClearCache function
> > 
> > 
> > The question is, will jitted methods be thrown away when this happens?
> 
> Yeah, logically, either the JIT optimizer recognizes the readonly flag and assumes the values are static, in which case writing to them via reflection will not be seen in compiled code, or it doesn't, in which case there is no performance penalty to removing `readonly`. So might as well remove `readonly` (if need be)

The most important thing about `readonly` for the JIT is assuming that no one will change the value in the middle of a method. Due to the nature of these objects, they'll likely be stored in a shared location, so the JIT won't inline the values directly, just the pointer to them. With `readonly`, JIT can lookup and cache the pointer at the start of the method; without `readonly`, it will lookup the pointer to the object on every access. Most likely not a major perf hit, but only benchmarking can tell.

Shane32:
> I hate that initialization and refresh are duplicated code.
> 
> This is gonna lead to bugs because of copy-paste mistakes, or a new contributor will see the member but not the refresh and introduce new hot-reload incompatible code, as @MaceWindu said.
> 
> I'd like it better if there was a static ctor that called `ClearCache` to initialize instead.

Agree

igor-tkachev (approved):

PR #3443, opened 2022-02-23
Fix for #3432. Localized bug with Outer Apply.

### LINKED ISSUE #3432: LINQ Any translated differently from v2
### Discussed in https://github.com/linq2db/linq2db/discussions/3429

<sup>Originally posted by **hkarask** February 17, 2022</sup>
Hi all,

I'm trying to bump the `linq2sql` version in an older codebase, from `1.10.2` to `2.0.0` and noticed that `LINQ` `Any` gets translated a bit differently:

Given a contrived LINQ statement:

```csharp
var query =
    from task in con.GetTable<Task>()
    from party in con.GetTable<Party>()
        .Where(p => task.AdminPartyId == p.Id ||
                    con.GetTable<PartyAccess>().Any(pa => pa.PartyId == p.Id && pa.Role == Roles.Admin)
        ).DefaultIfEmpty()
...
```

I'm seeing different SQL between versions 1 and 2

### v 1.10.2
```sql
SELECT
 	t.Description,
 	p.Name
 FROM
 	Task t
 		LEFT JOIN Party p ON t.AdminPartyId = p.Id OR EXISTS(
 			SELECT
 				*
 			FROM
 				PartyAccess pa
 			WHERE
 				pa.PartyId = p.Id AND pa.Role = 'Admin'
 		)
```
### v 2.0.0
```sql
SELECT
 	j.Description,
 	t2.Name
 FROM
 	Task t
 		LEFT JOIN (
 			SELECT
 				p.Name,
 				p.Id
 			FROM
 				Party p
 			WHERE
 				(EXISTS(
 					SELECT
 						*
 					FROM
 						PartyAccess pa
 					WHERE
 						pa.PartyId = p.Id AND pa.Role = 'Admin'
 				))
 		) t2 ON t.AdminPartyId = t2.Id
```

There's now a full sub-query instead of join a level filter. This is causing a full table scan for some reason (using MySql 5.6). Is there a way to revert back to the old behaviour?

## 

### Environment details
linq2db version: `1.10.2` & `2.0.0`
Database Server: `MySql 5.6`
Database Provider: `MySql.Data 8.0.25`
.NET Framework: `.NET 4.8`

Thanks!
</div>

sdanyliv:
Fixes #3432

Note that OUTER APPLY will be removed only if `Configuration.Linq.PreferApply = false`

### DIFF
--- a/Tests/Linq/UserTests/Issue3432Tests.cs
+++ b/Tests/Linq/UserTests/Issue3432Tests.cs
@@ -1,0 +1,68 @@
+﻿using System.Linq;
+using FluentAssertions;
+using LinqToDB;
+using LinqToDB.Mapping;
+using NUnit.Framework;
+
+namespace Tests.UserTests
+{
+	[TestFixture]
+	public class Issue3432Tests : TestBase
+	{
+		[Table]
+		public class Task
+		{
+			[Column] public int     AdminPartyId { get; set; }
+			[Column] public string? Description  { get; set; }
+		}
+
+		[Table]
+		public class Party
+		{
+			[Column] public int     Id   { get; set; }
+			[Column] public string? Name { get; set; }
+		}
+
+		[Table]
+		public class PartyAccess
+		{
+			[Column] public int     PartyId { get; set; }
+			[Column] public string? Role    { get; set; }
+		}
+
+		[Test]
+		public void OuterApplyOptimization([IncludeDataSources(TestProvName.AllSqlServer)] string context, [Values]bool preferApply)
+		{
+			const string Admin = "Admin";
+
+			using (var db = GetDataContext(context))
+			using (db.CreateLocalTable<Task>())
+			using (db.CreateLocalTable<Party>())
+			using (db.CreateLocalTable<PartyAccess>())
+			{
+				var query =
+					from task in db.GetTable<Task>()
+					from party in db.GetTable<Party>()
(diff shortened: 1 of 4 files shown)


Tests/Linq/UserTests/Issue3432Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+
+		[Table]
+		public class PartyAccess
+		{
+			[Column] public int     PartyId { get; set; }
+			[Column] public string? Role    { get; set; }
+		}
+
+		[Test]
+		public void OuterApplyOptimization([IncludeDataSources(TestProvName.AllSqlServer)] string context, [Values]bool preferApply)
```
MaceWindu, line 34: could we have mysql tests too?

sdanyliv: Well, it needs LATERAL check. Just simplified.

viceroypenguin (commented):
I don't see where `PreferApply` is actually used by the query system at all? I only see the declaration in Query.cs, and a change in SelectQueryOptimizer.cs that does not rely on that value.

sdanyliv:
`PreferApply` is used when we are trying to convert OUTER APPLY to LEFT JOIN and only in specific cases. Bug was that convert algorithm almost never reached point which checks this flag.

MaceWindu (approved):

viceroypenguin (approved):

PR #3460, opened 2022-03-07
Improve transactions logging

MaceWindu:
- add isolation level to logs
- add async suffix for async transaction operations

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.Async.cs
+++ b/Source/LinqToDB/Data/DataConnection.Async.cs
@@ -304,7 +304,7 @@ namespace LinqToDB.Data
 					dataConnection.OnTraceConnection(new TraceInfo(dataConnection, TraceInfoStep.AfterExecute, traceOperation, true)
 					{
 						TraceLevel    = TraceLevel.Info,
-						CommandText   = dataConnection.TraceSwitchConnection.TraceInfo ? sql : commandText?.Invoke(context),
+						CommandText   = sql,
 						StartTime     = now,
 						ExecutionTime = sw!.Elapsed
 					});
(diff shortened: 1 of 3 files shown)


sdanyliv (approved):

MaceWindu:
@viceroypenguin not sure... it will require addtional field on trace event object. I would prefere to leave it as is for now

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -299,7 +304,7 @@ public async Task DisposeAsync()
 					dataConnection.OnTraceConnection(new TraceInfo(dataConnection, TraceInfoStep.AfterExecute, traceOperation, true)
 					{
 						TraceLevel    = TraceLevel.Info,
-						CommandText   = commandText,
+						CommandText   = dataConnection.TraceSwitchConnection.TraceInfo ? sql : commandText?.Invoke(context),
```
viceroypenguin, line 307: * We already know `dataConnection.TraceSwitchConnection.TraceInfo == true` due to the `if()`
* We know that when it is `true`, the value is already set in `sql`.

```suggestion
						CommandText   = sql,
```

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1639,12 +1640,12 @@ public virtual void RollbackTransaction()
...
 
-				if (TraceSwitchConnection.TraceInfo)
+				if (dataConnection.TraceSwitchConnection.TraceInfo)
 				{
-					OnTraceConnection(new TraceInfo(this, TraceInfoStep.AfterExecute, traceOperation, false)
+					dataConnection.OnTraceConnection(new TraceInfo(dataConnection, TraceInfoStep.AfterExecute, traceOperation, false)
 					{
 						TraceLevel    = TraceLevel.Info,
-						CommandText   = commandText,
+						CommandText   = dataConnection.TraceSwitchConnection.TraceInfo ? sql : commandText?.Invoke(context),
```
viceroypenguin, line 1648: * We already know `dataConnection.TraceSwitchConnection.TraceInfo == true` due to the `if()`
* We know that when it is `true`, the value is already set in `sql`.

```suggestion
						CommandText   = sql,
```

viceroypenguin (approved):

Source/LinqToDB/Data/DataConnection.Async.cs
MaceWindu: whoops, it should be in error report... will fix

PR #3461, opened 2022-03-09
Fix for #3453. Added support for Enum types in PostgreSQL.

### LINKED ISSUE #303: MapValue to Postgres Enum Type
I have been playing around with LINQ2DB so far, and I'm very much liking it. However, I am having troubles with mapping values as per the blog post listed [here](http://blog.linq2db.com/2015/05/using-mapvalueattribute-to-control.html). In particular, given the following, I find that I end up raising NpgsqlExceptions depending on how I structure the code:

_Database.tt_: More or less the same as default, but I added the following line in between `LoadPostgreSQLMetadata(...)` and `GenerateModel()`

```
Tables["methods"].Columns["methodtype"].Type = "ModelMethodTypes";
```

_Form1.cs_: This is the minimal amount of code needed to reproduce this error:

``` c#
using System;
using System.Windows.Forms;
using DataModel;
using LinqToDB;
using LinqToDB.Mapping;

public enum ModelMethodTypes
{
    [MapValue("TypeA")] TypeA,
    [MapValue("TypeB")] TypeB,
    [MapValue("TypeC")] TypeC,
    [MapValue("TypeD")] TypeD
}


namespace TestLinq2DB
{
    public partial class Form1 : Form
    {
        public Form1()
        {
            InitializeComponent();

            DoModelThing(ModelMethodTypes.Capture);
        }

        private static int DoModelThing(ModelMethodTypes methodType)
        {
             using (var db = new TestDB())
            {
                var q = db.methods.Insert(() => new method
                {
                    methodtype = methodType,  // LINE 33 - This is where the problem happens
                    creationdate = DateTimeOffset.Now
                });

                return q;
            }           
        }
    }
}
```

Additionally, the following is the code I use to initialize my Postgres database (`initdb.sql`):

``` sql
CREATE TYPE method_type AS ENUM (
    'TypeA',
    'TypeB',
    'TypeC',
    'TypeD'
);

CREATE TABLE methods (
    id bigserial PRIMARY KEY,
    methodtype method_type NOT NULL,
    creationdate timestamptz NOT NULL
);
```

The problem is that when I call the method like this, an `NpgsqlException` is raised, telling me the following: 

``` text
Npgsql.NpgsqlException occurred
  BaseMessage=column "methodtype" is of type method_type but expression is of type text
  Code=42804
  ErrorCode=-2147467259
  File=src\backend\parser\parse_target.c
  HResult=-2147467259
  Hint=You will need to rewrite or cast the expression.
  InternalPosition=0
  Line=510
  Message=42804: column "methodtype" is of type method_type but expression is of type text
  MessageText=column "methodtype" is of type method_type but expression is of type text
  Position=77
  Routine=transformAssignedExpr
  Severity=ERROR
  Source=Npgsql
  StackTrace:
       at Npgsql.NpgsqlConnector.DoReadSingleMessage(DataRowLoadingMode dataRowLoadingMode, Boolean returnNullForAsyncMessage, Boolean isPrependedMessage)
       at Npgsql.NpgsqlConnector.ReadSingleMessage(DataRowLoadingMode dataRowLoadingMode, Boolean returnNullForAsyncMessage)
       at Npgsql.NpgsqlCommand.Execute(CommandBehavior behavior)
       at Npgsql.NpgsqlCommand.ExecuteNonQueryInternal()
       at Npgsql.NpgsqlCommand.ExecuteNonQuery()
       at LinqToDB.Data.DataConnection.ExecuteNonQuery() in i:\linq2db\Source\Data\DataConnection.cs:line 579
  InnerException: 
```

Normally, I would say that this is a problem with Npgsql, but I can't confirm why it would be doing this. I say that in particular because I can swap out my LINE 33 in the `Form1.cs` file with the following:

```
methodtype = ModelMethodTypes.TypeA,
```

and the code works without fail. What could possibly be causing this problem, and what am I doing wrong to map to user-defined types using linq2db? A short-term solution would be to change my database schema to incorporate characters or integer values in place of an enum type, however I would much rather have the ability to map 1:1 my enumeration type in C# to an enum type in PostgreSQL. For that matter, using Npgsql without an ORM, the following can be done: 

```
NpgsqlConnection.RegisterEnumGlobally<ModelMethodTypes>("method_type");
```

which registers the two together globally. In any case,  I figured since this works one way (calling the enum explicitly) but not the other (passing the enum value in as a parameter first), this must be either a bug or I have mis-configured something horribly. I've tried looking at the documentation but beyond the blog post referenced above I cannot find anything that details how `MapValue` works or how I should go about tweaking it.

### LINKED ISSUE #3368: PostGIS extension in bulk copy
Do I understand correctly that PostGIS (Spatial and Geographic objects for PostgreSQL) are not supported for BulkCopy with ProviderSpecific type (BINARY COPY)? In the documentation PostGIS data types are not mentioned ([link](https://linq2db.github.io/articles/sql/Bulk-Copy.html#postgresql-provider-specific-bulk-copy)) and in the source code itself ([link](https://github.com/linq2db/linq2db/blob/881adf49c15cf974a9494e680b9b138873c0d9e3/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLDataProvider.cs#:~:text=private%20void%20ConfigureTypes())) I cannot see handling of NpgsqlDbType.Geometry. When I try to run it in my project I always get following exception for one of PostGIS column with ```geometry(Geometry,4326)``` data type:

```
Exception: (Cannot guess PostgreSQL type for column. Specify type explicitly in column mapping.)
```

### LINKED ISSUE #3453: PostgreSQL ENUM type support
I have been trying for past few days to figure out how to use PostgreSQL ENUM type with linq2db.
I'm using latest Linq2Db v3.6.0 with .NET 6.0.
I read issues #2894 #303 #2576 (and some other issues)
Tried different things but the only thing that worked was using simple MapValueAttribute and enabling inline parameters. But I do not want to enable inline parameters I want a solution for parametrized queries.
For testing I'm using an example from issue #2894 (slightly changed)

```
DROP TABLE IF EXISTS schedule;
DROP TYPE IF EXISTS time_unit;
CREATE TYPE time_unit AS ENUM ('hour', 'day');
CREATE TABLE IF NOT EXISTS schedule
(
  id SERIAL CONSTRAINT schedule_pk PRIMARY KEY,
  unit time_unit NOT NULL,
  amount INT NOT NULL
);
```

```
namespace Models
{
    [Table(Schema = "public", Name = "schedule")]
    public class Schedule
    {
        [Column("id"), PrimaryKey, Identity] public int Id { get; set; } // integer
        [Column("unit")] public TimeUnit Unit { get; set; } // USER-DEFINED
        [Column("amount")] public int Amount { get; set; } // integer
    }

    public enum TimeUnit
    {
        [MapValue("hour")]
        Hour,
        [MapValue("day")]
        Day,
    }
}
```

```
using LinqToDB;
using LinqToDB.Data;
using Models;

DataConnection.TurnTraceSwitchOn(System.Diagnostics.TraceLevel.Verbose);
DataConnection.WriteTraceLine = (s1, s2, level) => Console.WriteLine($"{level}: {s1}, {s2}");
var db = new DataConnection(ProviderName.PostgreSQL, "Server=localhost;Port=5432;Database=test;User Id=postgres;Password=postgres;");
db.InlineParameters = true;

try
{
    await db.GetTable<Schedule>().Where(x => x.Unit == TimeUnit.Day).ToListAsync();
}
catch { }

try
{
    await db.InsertAsync(new Schedule { Unit = TimeUnit.Day, Amount = 1 });
}
catch { }
```

When using inline parameters query and insert works fine.

```
SELECT
        x.id,
        x.unit,
        x.amount
FROM
        "public".schedule x
WHERE
        x.unit = 'day'

INSERT INTO "public".schedule
(
        unit,
        amount
)
VALUES
(
        'day',
        1
)
```
But if I disable inline parameters then query still works but insert fails.

```
--  PostgreSQL.9.5 PostgreSQL (asynchronously)
DECLARE @Unit Text(3) -- String
SET     @Unit = 'day'
DECLARE @Amount Integer -- Int32
SET     @Amount = 1

INSERT INTO "public".schedule
(
        unit,
        amount
)
VALUES
(
        :Unit,
        :Amount
)
, DataConnection
Error: Error
Exception: Npgsql.PostgresException
Message  : 42804: column "unit" is of type time_unit but expression is of type text
```

I tried multiple suggestions but non of them worked.

Suggestion to manually map enums did not help
```
Npgsql.NpgsqlConnection.GlobalTypeMapper.MapEnum<TimeUnit>("public.time_unit");
LinqToDB.Mapping.MappingSchema.Default.SetDefaultFromEnumType(typeof(TimeUnit), typeof(TimeUnit));
```

Tried to use SetValueToSqlConverter:
```
var ms = new MappingSchema();
ms.SetValueToSqlConverter(typeof(TimeUnit), (sb, dt, v) =>
{
    Console.WriteLine($"SetValueToSqlConverter({dt},{v}");
    sb.Append('\'').Append(((TimeUnit)v).ToString().ToLower()).Append('\'');
});
var db = new DataConnection(ProviderName.PostgreSQL, ..., ms);
```
It seemed like a good direction but this method was never triggered,
it did not matter if I used MapValueAttribute or not.

Tried to use converter from enum to string:
```
var ms = new MappingSchema();
ms.SetConverter<TimeUnit, string>(value =>
{
    Console.WriteLine($"SetConverter<TimeUnit, string>({value})");
    return value.ToString().ToLower();
});
ms.SetConverter<string, TimeUnit>(value =>
{
    Console.WriteLine($"SetConverter<string, TimeUnit>({value})");
    return Enum.Parse<TimeUnit>(value, true);
});
var db = new DataConnection(ProviderName.PostgreSQL, ..., ms);
```
Having converter between TimeUnit and string also did not help.
Conversion methods are triggered but insert still fails.

Then tried converter from enum to DataParameter:
```
var ms = new MappingSchema();
ms.SetConverter<TimeUnit, DataParameter>(value =>
{
    Console.WriteLine($"SetConverter<TimeUnit, DataParameter>({value})");
    return new DataParameter("p", value.ToString().ToLower(), "time_unit");
});
var db = new DataConnection(ProviderName.PostgreSQL, "...", ms);
```
This one was triggered both on query and insert and both failed. Last argument dbType seems to have no effect.

Then tried another version:
```
ms.SetConverter<TimeUnit, DataParameter>(value =>
{
    Console.WriteLine($"SetConverter<TimeUnit, DataParameter>({value})");
    return new DataParameter("", value, "time_unit");
}, new DbDataType(typeof(TimeUnit)), new DbDataType(typeof(DataParameter), "time_unit"));
```
This one triggered only on insert and the statements were unchanged and execution failed.

Is there any way to make PostgreSQL ENUM type to work without inline parameters?

I know I can make this work if I use check constraint or a foreign key solution, for example:
```
DROP TABLE IF EXISTS schedule;
CREATE TABLE IF NOT EXISTS schedule
(
    id SERIAL CONSTRAINT schedule_pk PRIMARY KEY,
    unit VARCHAR(10) CHECK (unit IN ('hour', 'day'))  NOT NULL,
    amount INT NOT NULL
);
```
But I really want to use ENUM type.

sdanyliv:
Fix #303
Fix #3453
Fix #3368

### DIFF
--- a/Source/LinqToDB/DataProvider/BasicBulkCopy.cs
+++ b/Source/LinqToDB/DataProvider/BasicBulkCopy.cs
@@ -352,7 +352,11 @@ namespace LinqToDB.DataProvider
 			int                                       maxParameters,
 			int                                       maxSqlLength)
 		{
-			var adjustedBatchSize = helper.Options.UseParameters ?  Math.Min(helper.BatchSize, helper.Options.MaxParametersForBatch.GetValueOrDefault(maxParameters) / helper.Columns.Length): helper.BatchSize;
+			var adjustedBatchSize = helper.Options.UseParameters
+				? Math.Min(helper.BatchSize,
+					helper.Options.MaxParametersForBatch.GetValueOrDefault(maxParameters) / helper.Columns.Length)
+				: helper.BatchSize;
+
 			prepFunction(helper);
 
 			foreach (var item in source)
--- a/Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
+++ b/Source/LinqToDB/DataProvider/MultipleRowsHelper.cs
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace LinqToDB.DataProvider
 {
@@ -9,8 +11,7 @@ namespace LinqToDB.DataProvider
 	using Mapping;
 	using SqlProvider;
 	using SqlQuery;
-	using System.Threading;
-	using System.Threading.Tasks;
+	using Extensions;
 
 	public class MultipleRowsHelper<T> : MultipleRowsHelper
 		where T : notnull
@@ -35,7 +36,6 @@ namespace LinqToDB.DataProvider
 			MappingSchema  = dataConnection.MappingSchema;
 			Options        = options;
 			SqlBuilder     = DataConnection.DataProvider.CreateSqlBuilder(MappingSchema);
-			ValueConverter = MappingSchema.ValueToSqlConverter;
 			Descriptor     = MappingSchema.GetEntityDescriptor(entityType);
 			Columns        = Descriptor.Columns
 				.Where(c => !c.SkipOnInsert || c.IsIdentity && options.KeepIdentity == true)
@@ -48,7 +48,6 @@ namespace LinqToDB.DataProvider
 		public readonly DataConnection      DataConnection;
 		public readonly MappingSchema       MappingSchema;
 		public readonly BulkCopyOptions     Options;
-		public readonly ValueToSqlConverter ValueConverter;
 		public readonly EntityDescriptor    Descriptor;
 		public readonly ColumnDescriptor[]  Columns;
 		public readonly SqlDataType[]       ColumnTypes;
@@ -86,7 +85,7 @@ namespace LinqToDB.DataProvider
 				var column = Columns[i];
 				var value  = column.GetValue(item);
 
-				if (Options.UseParameters || skipConvert(column) || !ValueConverter.TryConvert(StringBuilder, ColumnTypes[i], value))
+				if (Options.UseParameters || skipConvert(column) || !MappingSchema.TryConvertToSql(StringBuilder, ColumnTypes[i], value))
 				{
 					var name = ParameterName == "?" ? ParameterName : ParameterName + ++ParameterIndex;
 
--- a/Source/LinqToDB/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
@@ -572,6 +572,11 @@ namespace LinqToDB.DataProvider.PostgreSQL
 				((Expression<Func<NpgsqlBinaryImporter, object?, NpgsqlDbType, CancellationToken, Task>>)((NpgsqlBinaryImporter this_, object? value, NpgsqlDbType type, CancellationToken token) => this_.WriteAsync(value, type, token)), true),
 				// [9]: Write
 				(Expression<Action<NpgsqlBinaryImporter, object?, NpgsqlDbType                        >>)((NpgsqlBinaryImporter this_, object? value, NpgsqlDbType type) => this_.Write(value, type)),
+				// [10]: Write
+				(Expression<Action<NpgsqlBinaryImporter, object?, string                              >>)((NpgsqlBinaryImporter this_, object? value, string dataTypeName) => this_.Write(value, dataTypeName)),
+				// [11]: WriteAsync
+				new Tuple<LambdaExpression, bool>
+				((Expression<Func<NpgsqlBinaryImporter, object?, string      , CancellationToken, Task>>)((NpgsqlBinaryImporter this_, object? value, string dataTypeName, CancellationToken token) => this_.WriteAsync(value, dataTypeName, token)), true),
 			};
 
 			public NpgsqlBinaryImporter(object instance, Delegate[] wrappers) : base(instance, wrappers)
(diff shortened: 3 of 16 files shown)


MaceWindu:
#303 probably too could be referenced

MaceWindu (approved):

viceroypenguin (approved):

PR #3474, opened 2022-03-19
Improve types support by providers

### LINKED ISSUE #2839: BulkCopyType.MultipleRows could insert incorrect data into varchar columns in Firebird

MaceWindu:
- fix #2839
- improve `Guid` type support:
  - [DB2] fixed table creation for entity with `Guid?` columns. Update literal generation to generate `BX'...'` literal instead of `CAST(x'...' as char(16) for bit data)`
  - [Firebird] add missing `CreateTable` support (`CHAR(16) CHARACTER SET OCTETS`) and add binary literal generation. Added `Sql.Ext.Firebird().UuidToChar(Guid)` extension to map `UUID_TO_CHAR` Firebird function
  - [Informix] add missing `CreateTable` support (`VARCHAR(36)`). Literal value example: `d8948d42-7b56-4dd3-b522-51a07c818e14`
  - [SQL CE] add missing `CreateTable` support (`UNIQUEIDENTIFIER`)
  - [Sybase ASE] add missing `CreateTable` support (`VARCHAR(36)`). Literal value example: `d8948d42-7b56-4dd3-b522-51a07c818e14`
  - [SQLite] remove `CAST(... as blob)` around binary literal for `Guid`
  - [Oracle] replace `CAST(... as raw(16))` with `HEXTORAW('...')` for `Guid` literal
- improve `Byte` type support:
  - [DB2] Byte type mapped to `SMALLINT`
  - [Informix] Byte type mapped to `SMALLINT`, fixed byte values support in parameters
- improve mapped and unmapped enums support:
  - [Sybase] Fix enums support in native provider bulk copy
- improve `bool` support:
  - [Sybase] Create non-nullable `bit` column (type doesn't support nullability in Sybase ASE) by `CreateTable`

### DIFF
--- a/Data/Create Scripts/Firebird.sql
+++ b/Data/Create Scripts/Firebird.sql
@@ -167,7 +167,7 @@ CREATE TABLE "DataTypeTest"
 	"DateTime_"       TIMESTAMP,
 	"Decimal_"        DECIMAL(10, 2),
 	"Double_"         DOUBLE PRECISION,
-	"Guid_"           CHAR(38),
+	"Guid_"           CHAR(16) CHARACTER SET OCTETS,
 	"Int16_"          SMALLINT,
 	"Int32_"          INTEGER,
 	"Int64_"          NUMERIC(11),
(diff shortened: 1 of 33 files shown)


viceroypenguin (approved):
Any likely conflicts between this and the Row Constructors PR?

PR #3484, opened 2022-03-28
Remove .net 5 testing support

MaceWindu:


viceroypenguin:
Even if .net 5 is EOL, there are going to be a lot of people still on VS'19 who will be using .net 5 with no chance to upgrade to .net 6 until they can convince an upgrade to VS'22. I'd say we should keep .net 5 in the base project, even if we no longer test it in the test projects.

MaceWindu:
We never had 5.0 TFM for non-test projects.

PR #3492, opened 2022-04-09
Change defaults for Configuration.ContinueOnCapturedContext

### LINKED ISSUE #3097: Still, why not just ConfigureAwait(false)?
I believe there is nothing in linq2db code that would require executing the continuations in the caller side sync context.
Example
```c#
public static async Task<List<T>> ToListAsync<T>(
	this IAsyncEnumerable<T> source, 
	CancellationToken        cancellationToken = default)
{
	var result = new List<T>();
	var enumerator = source.GetAsyncEnumerator(cancellationToken);
	await using (enumerator.ConfigureAwait(Common.Configuration.ContinueOnCapturedContext))
	{
		while (await enumerator.MoveNextAsync().ConfigureAwait(Common.Configuration.ContinueOnCapturedContext))
		{
			result.Add(enumerator.Current); // <---- THIS LINE
		}
	}

	return result;
}
```
I think there is no reason to make it execute "THIS LINE" in the UI thread (if the caller is an UI app).

MaceWindu:
Fix #3097
Update default value for `Configuration.ContinueOnCapturedContext` to `false`

### DIFF
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -62,7 +62,7 @@ namespace LinqToDB.Common
 		/// Defines value to pass to <see cref="Task.ConfigureAwait(bool)"/> method for all linq2db internal await operations.
 		/// Default value: <c>false</c>.
 		/// </summary>
-		public static bool ContinueOnCapturedContext = false;
+		public static bool ContinueOnCapturedContext;
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


sdanyliv (approved):

viceroypenguin (approved):
`false` is a better default for this. Good to change for v4.

Shane32:
:+1:

PR #3507, opened 2022-04-16
Refactor signatures to use more readonly collections 

MaceWindu:
Another set of cleanups for major version release

Update collection types to use interfaces/readonly interfaces where possible in signatures
- `HashSet<T>` -> `ISet<T>` (`IReadOnlySet<T>` not used due to availability only in net5+)
- IDictionary<,>` -> 'IReadOnlyDictionary<,>`
- Dictionary<,> -> IDictionary<,>` or 'IReadOnlyDictionary<,>`
- use of `ContainsKey` + value accessor pattern replaced with TryGetValue for dictionaries
- add `IReadOnlyDictionary<,>.ContainsKey` translation

Remains:
- review use of `ICollection`, `IEnumerable`, `List`, `IList`, `[]`
- replace use of ugly `Tuple` type with records or `ValueTuple`


jods4:
Why use `ISet` and `IDictionary` inside private apis?
They don't enforce a readonly contract, and they add an indirection in JIT, i.e. it's slower.

PR #3514, opened 2022-04-23
Better Contains/In translation when CompareNullsAsValues is false

jods4:
This is a rework of #2949 for release 4.0.

🚨 **Breaking change for v4.0** 🚨 
It changes the generated SQL of `Contains`, `In`, `NotIn` when `CompareNullsAsValues` is `false`: 
in this case values are not sniffed anymore and a straightforward `[NOT] IN` is generated.

Some examples:
`CompareNullsAsValues == true`, as today, no change(*):
```
a.In(1, null) -> a IN (1) OR a IS NULL
a.NotIn(1, null) -> a NOT IN (1) AND a IS NOT NULL (*)
a.NotIn(1) -> a NOT IN (1) OR a IS NULL
```
(*) This is actually a fix by this PR, it used to be incorrectly translated into `a not in (1) or a is not null`.

`CompareNullsAsValues == false` does not peek at values anymore and performs no null comparisons. SQL semantics apply.
```
a.In(1, null) -> a IN (1, NULL)
a.NotIn(1, null) -> a NOT IN (1, NULL)
a.NotIn(1) -> a NOT IN (1)
```

Of course, this linq2db magic only works with expressions that can be evaluated client-side (in C#). 
When the values are SQL fields/expressions, you get the latter behavior regardless.

I added a bunch of tests to cover all this.

There are two differences with PR #2949:
1. The new behavior is not keyed off of a new option, which was the point of redoing this PR for 4.0;
2. I have decided not to modify client-side implementation of `In` and `NotIn` to take `CompareNullsAsValues` into account. 

I figured it was better this way because it limits this risk of breaking code, especially for users that call `In` and `NotIn` from their own code as convenience functions. There are many differences between client-side and server-side evaluations anyway and generally client-side evaluation works like you'd expect from C#, which is often different from SQL. So it's more coherent if client-side `In` and `NotIn` always work "like C#".


jods4:
All tests are green, the only failure is an unrelated test that was ok in previous runs.
Looks like a random failure to me:
```
TearDown : 
System.InvalidOperationException : 
Baseline already in use:
/home/vsts/work/1/s/baselines/core31/Oracle.11.Managed/Tests/xUpdate/TruncateTableTests/Tests.xUpdate.TruncateTableTests.TruncateIdentityNoResetTest(Oracle.11.Managed).sql
```

MaceWindu:
ignore TruncateIdentityNoResetTest, it's oracle mystery

Tests/Linq/Linq/ContainsTests.cs
```diff
@@ -0,0 +1,103 @@
...
+			count = src.Count(s => !s.Int.In(Array.Empty<int?>()));
+			count.Should().Be(2);
+		}
+
+		[Test]
+		public void AllNulls(
+			// Excluded Access from tests because it seems to have non compliant behavior.
+			// It is the only DB that returns 1 for `WHERE Int NOT IN (null, null)`
+			[DataSources(TestProvName.AllAccess)] string context,
+			[Values(true, false)]                 bool   withNullCompares)
```
MaceWindu, line 82: FYI, `ValuesAttribute` pass all valid values by default for bool and enum types (defined enum fields)

jods4: Good to know, thanks!

Tests/Base/ScopedSettings.cs
```diff
@@ -43,6 +43,7 @@ public class WithoutComparisonNullCheck : IDisposable
 		public WithoutComparisonNullCheck()
 		{
 			Configuration.Linq.CompareNullsAsValues = false;
+			Query.ClearCaches();
```
MaceWindu, line 46: If we touch this area - it is a good time to add this setting to query cache flags instead of cache reset.
See
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/QueryFlagsHelper.cs

MaceWindu (changes requested):

Tests/Base/ScopedSettings.cs
jods4: This is just a small fix for the tests.

If you look at Dispose, you can see that when we go from `false` back to `true` the cache is cleared (so that queries built with the other setting are not reused in future tests).

But it makes no sense that when we go from `true` to `false` we don't do the same. In fact, this is the reason why some tests where failing in my previous PR. When the same function is run multiple times with different data and different `CompareNullsAsValues` settings, you end up in a situation where the `true` cached query is reused in a `false` run and ruins everything.

I tend to agree that adding this as a cache parameter makes sense, but it should be its own PR.

MaceWindu: it is basically couple of lines fix, so I would prefer to fix it now (if you prefer it done in separate PR - I'm fine with it). We will try to release 4.0 RC tomorrow.

jods4: I think it's unrelated to this, but you can do it here if you want. I don't mind.

viceroypenguin (approved):

PR #3502, opened 2022-04-12
Add DateOnly support

viceroypenguin:

PR #3515, opened 2022-04-24
Extract types from Sql into Sql.Types

jods4:
🚨 **Source-breaking change for 4.0** 🚨 
DB types that were in `Sql` have been moved into `Sql.Types`. Those are typically used with the `Convert` function.

```csharp
// 3.x
var q = 
  from t in db.Table 
  select Sql.Convert(Sql.BigInt, t.Value);

// 4.0
var q = 
  from t in db.Table
  select Sql.Convert(Sql.Types.BigInt, t.Value);
```

If you use `using static`:
```csharp
using static LinqToDB.Sql;
// Add this line for 4.0
using static LinqToDB.Sql.Types;

var q = 
  from t in db.Table 
  select Convert(BigInt, t.Value);

// Alternatively
var q = 
  from t in db.Table 
  select Convert(Types.BigInt, t.Value);
```

### Why?
This change is intended to make `using static LinqToDB.Sql` easier. 
When you do that, the problem in v3.x is that names like `DateTime` or `DateTimeOffset` become ambiguous with `using System`.
So if the file contains `DateTime.Today` for example, you have to prefix it.

### What's in a name?
I settled on an inner class `Sql.Types` after considering:
- `SqlTypes` class, but that clashes with `System.Data`.
- `SqlType` class, but we already have this one in our SQL Server provider.
- `Sql.Type` but I was concerned that `using static` would conflict with `System.Type`.

I can still change this PR if you have a better naming idea.

### Tests
As can be seen in the PR changes, all those members are well covered by the `Convert` tests.


MaceWindu (approved):

Shane32 (approved):

Source/LinqToDB/Sql/SqlType.cs
```diff
@@ -0,0 +1,164 @@
+using PN = LinqToDB.ProviderName;
```
viceroypenguin, line 1: Recommend file be renamed to Sql.Types.cs

viceroypenguin (approved):
TODO: Will have conflicts with #3502 - specifically Expressions.cs. Some lines overlap, and this PR does not (yet) cover the `DateOnly` section. Also, SqlType.cs will need to cover `DateOnly`. (or #3502 will need to address these conflicts).

Source/LinqToDB/Sql/SqlType.cs
jods4: Absolutely. I first named the class `SqlType` but found out it clashes with the SqlServer provider. 
Renamed the class but forgot the file, it's done now.

jods4:
@viceroypenguin This is a simple change that shouldn't be hard to merge. 
If it's accepted for v4.0, I'd recommend you back-merge this branch into #3502 and add `DateOnly` and `TimeOnly` to the new file.

jods4:
Previous test run was all green except one instance of Access tests. 
Looks like flacky tests to me:
```
The active test run was aborted. Reason: Test host process crashed : Fatal error. 0xC0000005
```

MaceWindu:
@jods4, please merge with DateOnly changes

jods4:
@MaceWindu do you mean this branch into DateOnly? or the other way round?

MaceWindu:
master into this branch

igor-tkachev (approved):

jods4:
@MaceWindu Thanks for doing the merge!

PR #3519, opened 2022-04-25
Fix issue w/ Converter Usage

viceroypenguin:


sdanyliv:
@viceroypenguin, it is something that we cannot predict. `ValueConverter` is applied to `Entity.Id` not anonymous object. We can resolve converter only in `On` part of Merge query, but it's too late, Values table do not have this information.

Issue #3535, opened 2022-04-30
ExprExpr is a mess under CompareNullsAsValues


jods4:
I wanted to work on `ExprExpr` and it looked massively inconsistent to me, so I ran short experiment with every case.

With `CompareNullsAsValues == false`, everything is as expected, operators match 1-to-1. 
Inverted operators are simply reversed, which is also correct.

Here's the meat of this issue: `CompareNullsAsValues == true` results.
**EDIT:** so `CompareNullsAsValues` is supposed to give the same results of C#. In C# lifted nullable comparisons are always `false` when one of the two values is `null`. I've updated the results below accordingly.
```
-- (1) a == b 🟩
a = b or a is null and b is null

-- (2) a != b 🟩
a <> b or a is null and b is not null or a is not null and b is null

-- (3) a >= b  🟦
a >= b

-- (4) a > b 🟦
a > b

-- (5) a <= b 🟦
a <= b

-- (6) a < b 🟦
a < b

-- (7) !(a == b) 🟩
a <> b or a is null and b is not null or a is not null and b is null

-- (8) !(a != b) ❌
a = b or a is null and b is not null or a is not null and b is null

-- (9) !(a >= b) ❌
a < b or a is null and b is null

-- (10) !(a > b) 🟩
a <= b or a is null or b is null

-- (11) !(a <= b) ❌
a > b or a is null and b is null

-- (12) !(a < b) 🟩
a >= b or a is null or b is null
```
(1) is correct. Relying on `is not distinct from` could be shorter SQL and possibly (?) better optimized (could maybe use index?).
(2) is correct as well, note that being the opposite of (1), generating `not (a = b or a is null and b is null)` would be quite shorter. Probably neither is optimizable by execution engine anyway.
(7) is the same as (2), correct. As noted `NOT (1)` might be more straightforward here.
(8) is wrong, should be the same as (1). Note that `(null, 1)` returns true, which is incorrect if you apply C# semantics.

Inequalities (3), (4), (5), (6) simply apply no special condition. When there's a null value it doesn't force `unknown` into `false`, which is usual for linq2db and very rarely makes a difference, especially since linq2db tries to simplify `not` in expressions (see below).

[...]

PR #3537, opened 2022-05-01
Null comparisons changes, incl. tests and fix for #3535

jods4:
This PR contains:
- New tests for every possible comparison, shows 3 bad inverted nullable comparisons as described in #3535 
- Fix for #3535
- A few minor refactorings and fixes
- ~Fixes (nope) #1767~
- Obsolete option `CompareNullsAsValues`, introducing a new option `CompareNulls` that accepts 3 different enum values. `LikeCSharp` works like `CompareNullsAsValues = true`; `LikeSqlExceptParameters` works like `CompareNullsAsValue = false` for backward compatibility. 
- Introduce a third new option: `LikeSql` that compiles queries in a straightforward way like `CompareNullsAsValues = false`, except it does not sniff parameter values: the only time when `==` and `!=` are not compiled to `=` and `<>` is when they involve a `null` or `default` literal.

```csharp
// Assuming CompareNullsAsValues = false
class Table 
{
  int? Value { get; set; }
}

// With any option, null and default still work as you'd expect:
db.Table.Where(t => t.Value == null);    // WHERE value IS NULL
db.Table.Where(t => t.Value != default); // WHERE value IS NOT NULL

// Comparisons to any other expression, even one that can be client-side evaluated, 
// is a plain equality when `CompareNulls = LikeSql`
int? missing = null; 
db.Table.Where(t => t.Value == missing);
// CompareNulls = LikeCSharp or LikeSqlExceptParameters: WHERE value IS NULL -- TRUE
// CompareNulls = LikeSql: WHERE value = @missing -- UNKNOWN

// If you want to include null-equality when using LikeSql,
// you should use IsNotDistinctFrom:
db.Table.Where(t => t.Value.IsNotDistinctFrom(missing))
// Always: WHERE value IS NULL
```

> **Note**
> For Oracle, that treats empty strings like null, that behavior extends to `""`.
> `x == ""` is compiled into `x IS NULL`.
> `x == str` compiles to `x = str` even when `str == ""` when using `LikeSql`.


jods4:
Initial commit with tests for every case is in, and the three operations I expected to be incorrect in #3535 are failing.

sdanyliv (dismissed):

jods4:
This is done, all remaining tests failure are `DirectoryNotFoundException` related to the baselines of new tests during Teardown.
I'm not familiar with the baselines, anyone has an idea?
> TearDown : System.IO.DirectoryNotFoundException : Could not find a part of the path 'D:\a\1\s\baselines\net472\SQLite.Classic.MPM.LinqService\Tests\Linq\CompareWithNullTests\Tests.Linq.CompareWithNullTests.Functional(SQLite.Classic.MPM.LinqService,False,(x =0x003E (Convert(x.EnumA) 0x003C Convert(x.EnumB)), System.Int32[], System.Int32[])).sql'

MaceWindu:
Path
`D:\a\1\s\baselines\net472\SQLite.Classic.MPM.LinqService\Tests\Linq\CompareWithNullTests\Tests.Linq.CompareWithNullTests.Functional(SQLite.Classic.MPM.LinqService,False,(x =0x003E (Convert(x.EnumA) 0x003C Convert(x.EnumB)), System.Int32[], System.Int32[])).sql`

is 261 characters, which is more than windows could digest. You should try to reduce test name length

jods4:
@MaceWindu I changed the way the parameters are passed to reduce baselines length, but I think the baseline repo was broken by previous runs.

I don't know how baselines work exactly, but it looks like tests try to checkout a baseline PR created by previous run and still fail on Windows because of (previously) too long filenames, see this log:
https://dev.azure.com/linq2db/linq2db/_build/results?buildId=8780&view=logs&j=abf8768c-4802-539b-ea89-b8847b4f4778&t=b87f150f-3651-5ff2-afd3-dc00aeeed2af

Can you delete the baseline PR or something like that?

MaceWindu:
#1767 has `has-tests` label, which means we have tests marked with active issue. Those tests should be enabled

MaceWindu:
https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Linq/WhereTests.cs#L1956-L1990

MaceWindu (changes requested):
enable issue tests. Also will need to review baselines (currently github PR page gives me unicorns instead of files...)

viceroypenguin (dismissed):

MaceWindu:
Checked tests for #1767 and cannot confirm it is fixed.

MaceWindu (dismissed):

MaceWindu:
We postpone it a bit as @sdanyliv has concerns regarding breaking changes. Please discuss it with him later

jods4:
@sdanyliv I'm back, we can discuss this here on in chat if you want.

MaceWindu:
@jods4 could you clarify reasoning behing breaking change as it is not clear to me.

- if it is to preserve setting semantics then why we generate `IS [NOT] NULL` for literals
- if it required for some other reason - could you clarify it?

sdanyliv:
I'm really don't like this breaking change. It will immediately break user user code. Comparing null parameter has no sense - always `false`.

Any technical complications doing that?

jods4:
### Motivation 
I'm doing this because the behaviour of `CompareNullsAsValues = false` is unexpected, incoherent, and actively prevents you from writing some SQL expressions.

(1) With `CompareNullsAsValues = false` my expectation is that C# compiles into the straight SQL equivalent.
Making `T.x == y` a true expression when `y` is null *is C# semantics*, and we've opted into *SQL semantics* so that's unexpected.

(2) It's incoherent: why `T.x == T.y` is false (assuming `x` and `y` nulls), but `T.x == y` is true??
With SQL semantics I can do that with `IsNotDistinct(T.x, T.y)`, why is `T.x == y` different?

(3) It prevents you from writing the SQL `T.x = y` when parameter `y` might be null.

@sdanyliv 
> Comparing null parameter has no sense - always false.

Always false when the parameter is null, but the parameter is not always null.

Turns out this is sometimes desirable, in fact this is how I became aware of this.
I wrote a query like this and I assumed the condition would be false when `y` is null. We hit bugs because linq2db changed the query into a `T.x is null` and suddenly some rows were true.
(Why did I not test more? I should have... I was porting old Dapper code and I didn't foresee any issue.)
I fixed the bug by refactoring into `y != null && T.x = y`, which is weird when you work with SQL semantics.

As a bonus, it has minor side-benefits like always compiling to the same query regardless of parameter value, which enables SQL plan re-use.

### What about T.x == null ?
Obviously comparing to a constant null makes no sense, ever. 
I see `== null` as a convenient syntax for `IS NULL`, which is a straightforward translation of the C# expression.

I think this is intuitive, regardless of the change above.

### Breaking change
Yes, this changes some queries so it's definitely a breaking change if you depend on the old behaviour.

How can we move forward? 
A major release? 
A more fine-grained option? `NullSemantics = Sql | CSharp | Legacy`, where `CompareNullsAsValues` becomes obsolete and maps `true` to `CSharp` and `false` to `Legacy`? 
If we did that, we'd better be sure there are no more translations like that in linq2db! I remember removing some from `Contains` / `IN` before...

jods4:
PS: it's hard to say how many people would be affected, but from our projects:
We always use `CompareNullsAsValues = false`, we have tons of super-complex queries... and we only hit this once.
I'd say it's not a super common pattern.

MaceWindu:
I personally approve this change, let's see @sdanyliv opinion

igor-tkachev:
The problem you are trying to solve is easily fixed by using the following:

`x.A == Sql.AsNotNull(x.B)` 

In this case you explicitly see what I do here (ignoring NULLs). (Maybe we need shorter version of this method?)

Your way to simplify generated SQL using options/flags is quite questionable. However, people like to shoot off their stuff from time to time. Just do not break compatibility. If you still need this feature, please add another option/flag (for example ShootOffMyEggs) and make default behavior compatible with the previous version.

BTW: Your parametrized test generates 2160 unit tests on my database set. Can we reduce this number? Do we need to test it over all the databases?

jods4:
@igor-tkachev that's a cleaner work-around than `y != null && ...`.

I still think the current behaviour is a `CompareNullsAsValues` leak: if linq2db didn't have `CompareNullsAsValues` at all, I doubt the variable comparison would behave this way. 

But given that the case is rare and your work-around not too horrible, I'm gonna give myself some time to sleep over it before I decide if it's worthy of an option or not.

> Your parameterized test generates 2160 unit tests on my database set. Can we reduce this number? Do we need to test it over all the databases?

It tests all possible combinations and highlighted a few bugs, so I think the test is worth having. 
What it tests shouldn't be different _per database_, so we could limit this test to one, or a few, providers.

This is a different discussion but the test suite is huge, requires real DB and takes a long time to run.
I wonder if locally we couldn't test a lot of functionality without a DB at all just by checking the generated SQL against a validated output. That would be much faster.

In CI when before a merge we should actually run every SQL against real DB of course, because there's always surprises.

MaceWindu:
I support @jods4 opinion here - if user enabled C# semantics for nulls using `CompareNullsAsValues` option - we should generate SQL as-is.

igor-tkachev:
I think you can add new option like UseSqlSemantic or something like that and do any changes you need for this. In this case there will be no braking change questions. Also we can use this option for some other SQL/C# semantic distinctions if needed.

For tests. Do we need to test it for all the possible databases? Can we do it for a couple like SQLite?

jods4:
@igor-tkachev  The new tests could run on one or two providers (which?) only. 
The code/bug they actually test is the same for all providers.

sdanyliv:
@igor-tkachev, EF Core has `UseRelationalNulls`, we can reuse this name in new options.

igor-tkachev:
OK. I would use more generic name to cover all possible C#/SQL semantic difference cases.

MaceWindu:
@jods4 looks like we need to introduce 3rd mode for compatibility reasons. If you will be able to refactor PR this week, it will be in next release.

jods4:
Last changes:
1. I made the new `ExprExpr` tests run for SQLite only. It really doesn't matter which provider we use to run them, they should be identical for all providers. I picked SQLite because it's the most convenient provider to run locally.

2. I added a new enum-based option `CompareNulls` that accepts 3 modes of operation: `LikeCSharp` (aka `CompareNullsAsValues = true`, `LikeSql` the changes introduced by this PR, and `LikeSqlExceptParameters` which works like `CompareNullsAsValues = false` for back-compat). `CompareNullsAsValues` is a proxy property to read/writes the enum behind the same. I marked it obsolete to direct old and new users towards the more fine-grained option. (Note that turning a field into a property is a binary-breaking change, users need to recompile to upgrade.)
3. I added a test case to illustrate the difference between the 3 modes (runs with SQLite only -- identical behavior for all providers).
4. I tweaked `OracleOptimizer` handling of `""` (which is equivalent to null -- for Oracle) and fixed at least one bug in doing so. I added one basic Oracle-only test to demonstrate check how `x == ""` works with each option.

5. I tweaked `IsParameterDependent` to be `false` when `CompareNulls = LikeSql`. There is no more parameter sniffing here so let's take advantage of it and stick queries in cache!

6. Back-merged master so we're ready to merge this PR! :)
7. All tests are green. Speaking of tests, please note that we now in theory have 3 options for null-handling. With the exception of the new tests designed to demonstrate of each option, all old tests still run in 2 modes only. `LikeCSharp` and the new `LikeSql`. I didn't have to change anything to make tests pass despite going from `LikeSqlExceptParameter` to a strict `LikeSql`.

MaceWindu:
@jods4 please merge conflicts. Also @sdanyliv wants to postpone it till v6 as it introduce a lot of conflicts with parser refactorings

jods4:
@MaceWindu in that case I'm gonna wait till @sdanyliv merges his refactoring before resolving any conflict. I don't want to do it twice.

sdanyliv:
Right, postpone till v6. A lot of changes here from my side.

MaceWindu:
Ping, @jods4

viceroypenguin:
Rebased PR is merged, closing.

PR #3540, opened 2022-05-02
Add net6.0 to the l2db nuspec

viceroypenguin:


MaceWindu:
fck

MaceWindu (approved):

MaceWindu:
I need to verify other projects too. E.g. I see linq2db.aspnet also builds for net6 but not included into nuget. Don't remember when net6 introduced here, probably we should remove it

viceroypenguin:
linq2db.aspnet added net6.0 in #3366 with the other projects. however, I don't see any specific need for it other than possibly referencing a newer version of base libraries.

MaceWindu:
Ah, right `add net6.0 build target to source projects (just for NRT validation on build, no nuget deployment yet)`, so we are fine

PR #3550, opened 2022-05-07
Add packages support

### LINKED ISSUE #399: Oracle T4 templates missing packages and certain stored procedures
I tried to generate the metadata infor for my oracle database. The t4 template successfully generated all the tables with few procedures. It failed to generate certain stored procedures and all packages. Let me know if calling packages is supported?

### LINKED ISSUE #3536: Add package/module support
Discovered when tried to fix unstable baselines for DB2 https://github.com/linq2db/linq2db.baselines/commit/c32b9fa4e3c2dd41ce64203fcb2456f1e82bce8a

Turned out there are two REMOVE procedures in different modules but schema load procedure doesn't detect it and load incorrect metadata.

Supported at least by oracle and db2 (could contain procedures, functions, types, variables but not tables/views).

As it will requires quite big refactoring of database object naming functionality it probably will be postponed till v5 release, but I will try to fix it before v4 final release.

Related issue #399

MaceWindu:
Fix #399
Fix #3536

Adds support for functions and procedures in packages.

Changes:
- REFACTORING: refactor database object name to use single `SqlObjectName` record instead of 4 string values
- REFACTORING: replace 3 `SqlBuilder` methods to generate table/object name with single method
- add support for `Package` name component to supported databases
- update T4 and scaffold tool codegeneration to generate package name for table functions
- FIX: fix duplicate schema names for oracle 11 and 12+ dialects leading to incorrect query cache lookup results
- ~FIX: implement workaround for `MySql.Data` stored procedure call to support procedures from package~ reverted, breaks `SchemaOnly` behavior
- FIX: add `TABLE()` wrappers to table function calls for `Oracle 11` and `DB2 LUW` dialects
- [Oracle]: support packages, scalar and table functions in schema/scaffold
- [Firebird]: support packages, scalar and table functions (AKA selectable procedures) in schema/scaffold
- [DB2 LUW]: support modules, scalar and table functions in schema/scaffold

Package support per-database:
- [Firebird 3+ (3.0.6+)](https://firebirdsql.org/file/documentation/release_notes/html/en/3_0/rnfb30-psql-packages.html)
- [DB2 LUW (named `module`)](https://www.ibm.com/docs/en/db2/11.5?topic=data-modules)
- [MariaDB 10.3.5+](https://mariadb.com/kb/en/create-package/)
- [Oracle](https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/CREATE-PACKAGE-BODY-statement.html#GUID-68526FF2-96A1-4F14-A10B-4DD3E1CD80BE)
- [SAP HANA 2 (named `library`)](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/62263ce35ac74488a397553fcb25a7d6.html)

---

### Provider-specific issues

Some providers do everything they can to prevent you from calling package members:
- [WILL NOT FIX] `MySql.Data` cannot call package procedure using `CommandType.StoredProcedure`. User should call procedure using `CALL proc_name` sql query (or switch to `MySqlConnector` finally) as fixing it in `linq2db` will break some other queries
- [NOT FIXABLE] SAP HANA native provider cannot call library functions and procedures as SAP HANA use `:` as library member name separator and provider replaces all `:` with `?` (monkey see parameter prefix character `:`, monkey do brain-dead `replace all`)
- [NOT FIXABLE] MariaDB doesn't expose package information in schema tables, making it impossible to scaffold schema procedures/function
- [NOT FIXABLE] SAP HANA doesn't expose package information in schema tables, making it impossible to scaffold schema procedures/function


Source/LinqToDB/Linq/QueryRunner.Update.cs
```diff
@@ -29,10 +29,10 @@ public static class Update<T>
...
 				var sqlTable = new SqlTable<T>(dataContext.MappingSchema);
 
-				if (tableName    != null) sqlTable.PhysicalName = tableName;
-				if (serverName   != null) sqlTable.Server       = serverName;
-				if (databaseName != null) sqlTable.Database     = databaseName;
-				if (schemaName   != null) sqlTable.Schema       = schemaName;
+				if (tableName    != null) sqlTable.TableName = sqlTable.TableName with { Name     = tableName    };
+				if (serverName   != null) sqlTable.TableName = sqlTable.TableName with { Server   = serverName   };
+				if (databaseName != null) sqlTable.TableName = sqlTable.TableName with { Database = databaseName };
+				if (schemaName   != null) sqlTable.TableName = sqlTable.TableName with { Schema   = schemaName   };
```
jods4, line 35: Note that this generates a large amount of assembly code and potentially performs a lot of copying.

`TableName` is a `record struct SqlObjectName` containing 5 pointers.
Each of those 4 assignments generates a full copy of `TableName` into a temp local, performs the modification, and then copies the temp local back over `TableName`. That's 40 `movsd` instructions right there, that are all executed if you provide all the options.

In comparison, the previous code had only 4 `movsd` instructions, 10 times less!

If we want to go with a record here, it seems to me it'd be better to pass those overrides to `SqlTable<T>` ctor and inside it create a new `SqlObjectName` by using `??` between each field of the `EntityDescriptor` and the override.

This pattern is found at least 10 more times in the other `QueryRunner` statements that are modified in this PR, so overall it'd shorten both the source code and the size of JITed code by a fair bit.

viceroypenguin (approved):

PR #3551, opened 2022-05-10
[WIP] Update to C#10

viceroypenguin:


sdanyliv:
Don’t do that right now. It is terrible difficult to merge collisions with upcoming changes. We have at least two very big PR which will be affected.

viceroypenguin:
> Don’t do that right now. It is terrible difficult to merge collisions with upcoming changes. We have at least two very big PR which will be affected.  

Yes, WIP/Draft to be completed after those two. Merge from those into this one will not be difficult.

jods4:
I am working on a very large change, too.

Issue #2421, opened 2020-08-18
Syntax error in the JOIN operation for eager loading LoadWith()


yarecky1:
I just wanted to play a little with LoadWith() we spoke yesterday about and found bug?

5 tables:
```
* User               -> PK int UserId,                  string Name
* Position           -> PK int PositionId,              string Name
* Sector             -> PK int SectorId,                string Name

* UserPosition       -> PK int UserPositionId,       FK int UserId,         FK int PositionId (many-to-many)
* UserPositionSector -> PK int UserPositionSectorId, FK int UserPositionId, FK int SectorId   (many-to-many)

User can have many Positions (Roles) and as a combination UserPosition can be in many Sectors.
Vice versa: Sector can have many Users each playing different roles.
```
I want to get list of all sector names with user names 

### Case 1 (working)

```c#
var y = _context.Value.FtqSectors
    .Select(x => new
    {
        SectorName = x.Name,
        UserName = x.UserPositionSectors.Select(t => t.UserPosition.User.Fullname)
    });

var z = y.ToList();
```
Generated two queries

```SQL
SELECT
	[x].[Id] as [Key_1],
	[User_1].[Lastname],
	[User_1].[Firstname]
FROM
	(([tblSector] [x]
		INNER JOIN [tblUserPositionSector] [detail] ON ([x].[Id] = [detail].[SectorId]))
			INNER JOIN [tblUserPosition] [UserPosition] ON ([detail].[UserPositionId] = [UserPosition].[UserPositionId]))
		INNER JOIN [tblUser] [User_1] ON ([UserPosition].[UserId] = [User_1].[UserId])

SELECT
	[x].[Nam] as [SectorName],
	[x].[Id]
FROM
	[tblSector] [x]
```

### Case 2 (not working)
I want to limit Position to `PositionId == 3` so I add `WHERE` clause.

```c#
var y = _context.Value.FtqSectors
    .Select(x => new
    {
        SectorName = x.Name,
        UserName = x.UserPositionSectors.Where(t => t.UserPosition.PositionId == 3).Select(t => t.UserPosition.User.Fullname)
    });
```
Generated output

```SQL
SELECT
	[x].[Id] as [Key_1],
	[User_1].[Lastname],
	[User_1].[Firstname]
FROM
	(([tblSector] [x]
		INNER JOIN [tblUserPositionSector] [detail] ON ([x].[Id] = [detail].[SectorId] AND [UserPosition].[PositionId] = 3))
[...]

PR #3558, opened 2022-05-17
Fix nullable association join type

### LINKED ISSUE #3557: Association - wrong sql generated
Expected LEFT OUTER JOIN for 'public SubData1 SubData1 { get; }' while got INNER JOIN:
what we got:

```sql
FROM
	[Data] [i]
		INNER JOIN [SubData1] [a_SubData1] ON [i].[Id] = [a_SubData1].[Id]
```

what was expected

```sql
FROM
	[Data] [i]
		LEFT JOIN [SubData1] [a_SubData1] ON [i].[Id] = [a_SubData1].[Id]
```

### Steps to reproduce
```c#
  public static class TestExample
  {
    public class Result
    {
      public int    Id     { get; set; }
      public string Reason { get; set; }
    }

    [Table]
    public class SubData2
    {
      [Column] public int    Id     { get; set; }
      [Column] public string Reason { get; set; }
    }

    [Table]
    public class SubData1
    {
      [Column] public int    Id     { get; set; }
      
      [Association(ThisKey = nameof(Id), OtherKey = nameof(SubData2.Id))]
      public IEnumerable<SubData2> SubData2s { get; }
    }

    [Table]
    public class Data
    {

      [Association(ThisKey = nameof(Id), OtherKey = nameof(TestExample.SubData1.Id), CanBeNull = true)]
      public SubData1 SubData1 { get; }
    }

    public static void Test()
    {
      using var dc = new Dc();

      var result = dc.GetTable<Data>()
        .Select(
          i =>
            new Result
            {
              Id     = i.Id,
              Reason = i.SubData1.SubData2s.Select(s => s.Reason).FirstOrDefault(),
            })
        .ToList();
    }
  }
```

### Environment details
linq2db version: latest git
Database Server: sql server 2008
Database Provider: sql server 2008
Operating system: win10
.NET Framework: 4.7.2

MaceWindu:
Fix #3557


MaceWindu:
1. Main question is: why we enforced `inner join` in those cases - it is expected behavior or bug? I would say it is a bug
2. Access tests start to fail due to limited join syntax (see #2421)

sdanyliv (dismissed):

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -1255,7 +1255,7 @@ public IBuildContext GetContext(Expression? expression, int level, BuildInfo bui
 						}
 						else
 						{
-							var tableLevel  = FindContextExpression(expression, level, true, true)!;
+							var tableLevel  = FindContextExpression(expression, level, false, true)!;
```
viceroypenguin, line 1258: Prefer to have names on the boolean parameters to be able to evaluate the change.

viceroypenguin (approved):

PR #3573, opened 2022-05-21
Refactor BasicSqlBuilder.DataProvider

jods4:
- Make it non-nullable (it is never null)
- Make it generic to get rid of all the casts in derived classes

PostgreSqlBulkCopy could get rid of its casts if `CreateSqlBuilder` used co-variance, but our old .net targets don't allow that.
BasicSqlBuilder could also use the curiously recursive pattern but I feel like it's not worth doing here.

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilderBase.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilderBase.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Data;
 using System.Linq;
 
 namespace LinqToDB.DataProvider.Oracle
@@ -11,16 +10,14 @@ namespace LinqToDB.DataProvider.Oracle
 	using Mapping;
 	using System.Data.Common;
 
-	abstract partial class OracleSqlBuilderBase : BasicSqlBuilder
+	abstract partial class OracleSqlBuilderBase : BasicSqlBuilder<OracleDataProvider>
 	{
-		public OracleSqlBuilderBase(IDataProvider? provider, MappingSchema mappingSchema, ISqlOptimizer sqlOptimizer, SqlProviderFlags sqlProviderFlags)
+		public OracleSqlBuilderBase(OracleDataProvider provider, MappingSchema mappingSchema, ISqlOptimizer sqlOptimizer, SqlProviderFlags sqlProviderFlags)
 			: base(provider, mappingSchema, sqlOptimizer, sqlProviderFlags)
-		{
-		}
+		{ }
 
-		protected OracleSqlBuilderBase(BasicSqlBuilder parentBuilder) : base(parentBuilder)
-		{
-		}
+		protected OracleSqlBuilderBase(OracleSqlBuilderBase parentBuilder) : base(parentBuilder)
+		{ }
 
 		protected override void BuildSelectClause(SelectQuery selectQuery)
 		{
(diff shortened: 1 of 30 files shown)


jods4:
This doesn't work because of remoting.
There's dynamic code generation in remoting that creates a `SqlBuilder` of the desired type, but with a hard-coded null `DataProvider`:
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Remote/RemoteDataContextBase.cs#L241-L257

This is ugly and I don't the energy to untangle it for now.

Why does remoting even need to instantiate a SqlBuilder provider, as a SQL tree is remoted, not the final string?

Source/LinqToDB/DataProvider/Oracle/OracleSqlBuilderBase.cs
```diff
@@ -548,14 +545,10 @@ void AppendSchemaPrefix(StringBuilder sb, string? schema)
...
-				var param = provider.TryGetProviderParameter(dataContext, parameter);
-				if (param != null)
-					return provider.Adapter.GetDbType(param).ToString();
-			}
-
-			return base.GetProviderTypeName(dataContext, parameter);
+			var param = DataProvider.TryGetProviderParameter(dataContext, parameter);
+			return param != null
+				? DataProvider.Adapter.GetDbType(param).ToString()
+				: base.GetProviderTypeName(dataContext, parameter);
```
viceroypenguin, line 551: This code appears duplicated in several places (Oracle, Sql, Sybase, etc.). Was apparently necessary previously due to the `is <db>DataProvider`, but maybe could be made common in the base class now?

jods4: Not all builders do that, but several do, yes.
The cast was required because `TryGetProviderParameter` is not part of the interface. 
Maybe this could all be simplified if it was added to the interface.

viceroypenguin: That was going to be my first recommendation, which would have simplified several parts of this PR? Any reason not to add it to the interface? Is it a public API that others are likely to have implemented?

jods4:
Ok, I'll do what I wanted to do differently and dismiss this PR.
If someone thinks this was a good idea and wants to pursue, please feel free.

The culprit here is remoting:
1. Remote clients shouldn't have DB provider libraries to run queries.
2. As the name says `DataProvider` is tightly coupled to DB provider, so it shouldn't be needed in remote client.
3. `SqlBuilder` contains a `DataProvider` 
4. `SqlBuilder` is needed to build provider-specific queries and is instantiated by remote client.
5. (4) + (3) contradict (1) but it works because the remoting client works with a null `DataProvider` in `SqlBuilder`. The code looks a bit hacky, builders must do many casts / null checks to work.

Ideas to clean that up:
I think the "simpler" redesign would be to cleanly separate provider-dependent and independent code.
If `SqlBuilder` is needed to build remote queries, then its API should not need `DataProvider` those functions should be moved/refactored somewhere else, maybe in `DataProvider` itself.

The "more complex" design IMHO is that even relying on provider-specific `SqlBuilder` in remote client is leaking server abstractions. 
It means the client needs to know what DB the server is running and have the required classes available -> this is easy as every provider is in the main linq2db dll, but in a better world where each DB has its own linq2db.provider.dll, this would be very leaky.
A better design could build a provider-independent query remotely, then this query would be adjusted to be provider-specific by the server.
I don't think that's realistic with the current codebase, more core improvements are IMHO required before something like that could be attempted.

PR #3590, opened 2022-05-30
Fix error reporting for connection open

### LINKED ISSUE #861: OnTrace does not create a TraceInfoStep == Error if no connection was established
LinqToDB-Version: 1.9.0
DB-Provider: Firebird 2.1

When no connection is established and you access the database e.g. to load entities, the OnTrace delegate does not get an information about it. The resulting TraceInfoStep is always on Completed.

```
DataConnection.OnTrace += delegate (TraceInfo info)
                {
                    if (info.TraceInfoStep == TraceInfoStep.Error)
                    {
                        // If no connection is established, this will never arrive!!
                        Log.Error("Error while accessing database!");
                    }
                    else if (info.TraceInfoStep == TraceInfoStep.AfterExecute || info.TraceInfoStep == TraceInfoStep.Completed)
                    {
                        //..
                    }
                };
```

Only workaround till now is to always check the Connection property in the OnTrace delegate:

```
DataConnection.OnTrace += delegate (TraceInfo info)
                {
                    try
                    {
                        // This throws an exception if no connection was established...
                        var conn = info.DataConnection?.Connection;
                    }
                    catch (Exception)
                    {
                        Log.Error("Error while accessing database!");
                        return;
                    }
                    if (info.TraceInfoStep == TraceInfoStep.Error)
                    {
                       Log.Error("Error while accessing database!");
                    }
                    else if (info.TraceInfoStep == TraceInfoStep.AfterExecute || info.TraceInfoStep == TraceInfoStep.Completed)
                    {
                       //...
                    }
                };
```

viceroypenguin:
Fixes #861


sdanyliv:
Probably easies way to test this was to create DataConnection using a tons of available constructor overloads.

viceroypenguin:
> Probably easies way to test this was to create DataConnection using a tons of available constructor overloads.

a) consistency, b) didn't want to have to provide a specific provider to the constructor, c) flexibility in case someone wants to confirm behavior on other db systems in the future (though code is db agnostic, so shouldn't be affected...)

Shane32 (approved):

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -144,20 +144,24 @@ await _connectionInterceptor.ConnectionOpeningAsync(new (this), _connection.Conn
...
+						Exception  = ex,
+					});
+				}
 
+				// OCE should always be re-thrown instead of encapsulated
+				if (ex is OperationCanceledException)
 					throw;
-				}
+
+				throw new LinqToDBException("Unable to open connection to database", ex);
```
MaceWindu, line 164: Why we wrap expection here? I would expect it to break exception handlers or user side.

viceroypenguin: * yes, may break existing exception handlers that don't already trap `LinqToDBException`. 
* this is correct behavior - consumer should be able to trap a single exception for any db instead of the constituent exception from each db library. 
* if we don't want to do this, I can revert the change here and update in the non-async version as well. behavior should be consistent across both versions.

viceroypenguin: Case in point, in merging your devart library changes into my branch, I had a conflict on one of the unit tests, because it had a different exception than the "normal" `ArgumentException`. This tells me that always doing a `LinqToDBException`, at least for the Open, makes more sense long-term.

MaceWindu: @linq2db ping

MaceWindu: My vote is to revert it. Except being breaking change (e.g. for me) it makes connection open a special case where all other exceptions from other provider methods still preserve original exception and it is hard to see for me why open should differ.

MaceWindu: Also I don't agree it simplifies exception handing:
- it adds additional exception type you need to handle if you already analyzed exceptions and performed different actions based on exception
- and for catch-all fans it adds nothing new

viceroypenguin: > My vote is to revert it. Except being breaking change (e.g. for me) it makes connection open a special case where all other exceptions from other provider methods still preserve original exception and it is hard to see for me why open should differ.

That's a fair point. I guess that is a discussion decided 10 years ago: should exceptions be from the original provider or consistent. On one hand, a consistent exception means that a consumer can receive a `DataConnection` and do whatever without concern to the underlying database. On the other, you already still have to know the underlying database in many cases due to corner cases and more complicated magic that you want to do. 

Since it's something already decided in favor of provider exceptions, I'll defer to that decision. Will update, but it'll take a short bit to rewrite the unit tests to be more inclusive to various exceptions.

Shane32: > My vote is to revert it. Except being breaking change (e.g. for me) it makes connection open a special case where all other exceptions from other provider methods still preserve original exception and it is hard to see for me why open should differ.

I would say that exceptions should be treated consistently across the library. If they are generally not wrapped, I wouldn’t do so here either.

MaceWindu (approved):

PR #3606, opened 2022-06-03
[WIP] Support C# nullable references

### LINKED ISSUE #2690: Question: Are the Nullable/NotNull attributes still needed in C#8 with Nullable Reference Types?
I see that the Nullable and NotNull attributes are needed for marking class types (mainly string) where we previously had no way to know if they were meant to be nullable or not. But now in C#8, you can just set the type as `string?` and the compiler will mark it with a Nullable attribute.

So, my question is, is explicitly setting Nullable/NotNull still required for LinqToDb when using it in C# 8 or 9?

I tried looking in the documentation and the issues, but I couldn't find any relevant info.

jods4:
Fixes #2690

👉 This change applies to .net 6.0+ only.

When there is no explicit `CanBeNull` annotation, i.e. `[Column]` has not set `CanBeNull` and there is no `[Nullable]`, before falling back to default rules, we use the C# nullable reference type info if there is one available.

Work to be done:
- [x] `[Column]` configuration
- [ ] `[Association]` configuration
- [ ] Add one test or two

### DIFF
--- a/Source/LinqToDB/Mapping/ColumnDescriptor.cs
+++ b/Source/LinqToDB/Mapping/ColumnDescriptor.cs
@@ -93,25 +93,6 @@ namespace LinqToDB.Mapping
 				StorageInfo = expr.Member;
 			}
 
-			var defaultCanBeNull = false;
-
-			if (columnAttribute.HasCanBeNull())
-				CanBeNull = columnAttribute.CanBeNull;
-			else
-			{
-				var na = mappingSchema.GetAttribute<NullableAttribute>(MemberAccessor.TypeAccessor.Type, MemberInfo, attr => attr.Configuration);
-
-				if (na != null)
-				{
-					CanBeNull = na.CanBeNull;
-				}
-				else
-				{
-					CanBeNull        = mappingSchema.GetCanBeNull(MemberType);
-					defaultCanBeNull = true;
-				}
-			}
-
 			if (columnAttribute.HasIsIdentity())
 			{
 				IsIdentity = columnAttribute.IsIdentity;
@@ -130,9 +111,8 @@ namespace LinqToDB.Mapping
 
 			SkipOnInsert = columnAttribute.HasSkipOnInsert() ? columnAttribute.SkipOnInsert : IsIdentity;
 			SkipOnUpdate = columnAttribute.HasSkipOnUpdate() ? columnAttribute.SkipOnUpdate : IsIdentity;
-
-			if (defaultCanBeNull && IsIdentity)
-				CanBeNull = false;
+			
+			CanBeNull = AnalyzeCanBeNull(columnAttribute);
 
 			if (columnAttribute.HasIsPrimaryKey())
 				IsPrimaryKey = columnAttribute.IsPrimaryKey;
@@ -173,6 +153,39 @@ namespace LinqToDB.Mapping
 				SkipBaseAttributes    = skipValueAttributes;
 				SkipModificationFlags = SkipBaseAttributes.Aggregate(SkipModification.None, (s, c) => s | c.Affects);
 			}
+		}
+
+		private bool AnalyzeCanBeNull(ColumnAttribute columnAttribute)
+		{
+			if (columnAttribute.HasCanBeNull())
+				return columnAttribute.CanBeNull;
+			
+			var na = MappingSchema.GetAttribute<NullableAttribute>(
+				MemberAccessor.TypeAccessor.Type, 
+				MemberInfo, 
+				attr => attr.Configuration);
+			if (na != null)
+				return na.CanBeNull;
+				
+#if NET6_0_OR_GREATER
+			// Extract info from C# Nullable Reference Types if available.
+			// Note that this should also handle Nullable Value Types.
+			var context = new NullabilityInfoContext();
+			var nullability = MemberInfo switch 
+			{
+				PropertyInfo p => context.Create(p).ReadState,
+				FieldInfo    f => context.Create(f).ReadState,
+				             _ => NullabilityState.Unknown,
+			};
+			
+			if (nullability != NullabilityState.Unknown)
+				return nullability == NullabilityState.Nullable;
+#endif
+
+			if (IsIdentity) 
+				return false;
+
+			return MappingSchema.GetCanBeNull(MemberType);
 		}
 
 		/// <summary>


Shane32:
Just FYI, the implementation in .NET 6 has a lot of bugs, although honestly it's unlikely they will affect NRT interpretation for basic models.  See following links:
- https://github.com/dotnet/runtime/pull/63556
- https://github.com/dotnet/runtime/pull/64143

The bug fixes will make it in .NET 7 but (as far as I know) they did not patch .NET 6.

The source code that MS uses will run on any .NET Standard 2.0 library; it is not limited to .NET 6.  We can copy MS's files over to allow reading NRT annotations on pre-.NET 6 versions, or use a source-only NuGet package like [Nullability.Source](https://www.nuget.org/packages/Nullability.Source/) to read NRT annotations on pre-.NET 6 versions.  I have used this in [GraphQL.NET](https://github.com/graphql-dotnet/graphql-dotnet) to read NRT annotations without having a 3rd party dependency.

Note that the Nullability.Source package only takes effect for pre-.NET 6 runtimes, which means that code running on any version other than .NET 6 (prior to or later than) will have the patched code whereas .NET 6 will have the unpatched code.  Since it is source code, it is possible that it may break a build, for example if the build process checks the source files for proper formatting, or if a global using changes the meaning of a type.

Shane32:
As it relates to the feature in general, I'm all for it, although it's probably a breaking change.  I also noticed recently that EF also reads NRT annotations, when enabled.

MaceWindu:
Also check if this is apply here : https://github.com/dotnet/efcore/issues/27474

jods4:
This is a nice feature but it seems to be a bigger can of worms than I'd anticipated.

First I observe a significant number of our tests fail with `NotImplementedException` inside `System.Reflection` methods.
At first glance this is quite surprising to me.
```
NotImplementedException
  at System.Reflection.MemberInfo.GetCustomAttributesData()
  at System.Reflection.NullabilityInfoContext.Create(PropertyInfo propertyInfo)
  at LinqToDB.Mapping.ColumnDescriptor.AnalyzeCanBeNull(ColumnAttribute columnAttribute)
```

@Shane32 **RE bugs:** I've seen the issues when doing initial research for this feature.
For our basic usage, I don't think there should be problems -- from what I've seen so far.
Notice that if an edge-case is incorrect, it's possible for users to set `CanBeNull` explicitly and work around it.

**RE Nullability package:** Why not, what's the opinion of others?
We could use this to bring this feature to .net before 6.0. 
Personally I'm ok with this new feature being .net 6+ only, as long as it's advertised as such. 
For old projects that can't easily update, like .net fx projects, I think it is unlikely they are nullable-enabled anyway, as that would require using a recent compiler + a nuget package to add all the useful nullable attributes + they don't get as much benefits as their BCL is not annotated.

We would need to decide if we take source deps or dll, which has implications when running on .net 6+ as you mentioned.

@MaceWindu What is the best course of action in your opinion?
Trimming can remove nullability attributes, as can another feature that removes them on private/internal members only (would apply if someone activate this feature and use internal classes for data model).
I understand that EF forces this feature off for all consumers?
If we have users that care about size, I'd rather lean towards linq2db working without exceptions in these contexts, users simply have to continue using `[NotNull]` or `[Columns(CanBeNull = false)]` attributes.
And we should document that when trimming, `NullabilityInfoContextSupport` should be set to true if users want to depend on it.
What's bad about this is that trimming is typically only done when publishing and might go unnoticed during dev. 

I wonder how much of all this we can reliably test?

jods4:
About trimming:
It's not as bad as I thought.

There's a feature switch that dictates whether Nullable attributes can be trimmed.
`NullabilityInfoContext.Create()` throws an `InvalidOperationException` when the switch is unset.

This means we can detect that nullable annotations are not available even during dev before publish/trimming.
We could even write a Debug-level message about it.

I like this more than what EF did, because users that are size-conscious like Blazor or MAUI can still trim to the max but have to rely on manual `[NotNull]`.
For most users, who don't trim this changes nothing.

MaceWindu:
Regarding availability/dependency: I would prefer it awailable anywhere and referenced as source file in Source/Shared folder as we do for other version-dependent features

Shane32:
> Regarding availability/dependency: I would prefer it awailable anywhere and referenced as source file in Source/Shared folder as we do for other version-dependent features

I would agree.  There's no reason not to include this feature for all platforms that support it; I would not lock the feature to only .NET 6+ users.

We could also implement a global switch to enable/disable the functionality.  This would allow this feature to be a non-breaking addition to 4.0.0 if we default to disabled.  It might also help to solve issues with trimming and platform support, if those platforms have support defaulted to false.

I don't know enough about trimming, but if trimming is a concern, perhaps there's a way to trim the NRT analysis conditionally or from platforms that expect to use trimming, or something, so that compilation will trim the `NullabilityInfoContext` class (either from our code or .NET 6's code).

jods4:
Ok, so I added `Nullability.Source` into the project, but **note that it needs .net standard 2.0+**.

So, by targets:
- .net 4.5 and 4.6 still have no support for NRT.
- Anything between .net 4.6 and .net 6.0 has the source code of `NullabilityInfoContext` embedded into linq2db and works.
- .net 6.0 and above have no embedded source and use the BCL version.

jods4:
I also added a check to see if `NullabilityInfoContext` is supported, i.e. if nullable attributes are not aggressively trimmed such as in MAUI or Blazor builds. (Note that "regular" trimmed builds do not have this feature on by default, it's opt-in.)

In this case, we just skip the `NullabilityInfoContext` code and run like today (i.e. users have to rely on `[Column]` and `[NotNull]`).

If users are ok with slightly bigger binaries, they can enable NRT analysis by disabling the trimming of those attributes with this property in their csproj:
`<NullabilityInfoContextSupport>true</NullabilityInfoContextSupport>`

Note that this check is based on msbuild configuration, not actual publishing/trimming. 
So behavior is identical during dev/tests and published exe.

jods4:
As anticipated by @Shane32, the embedded source creates problems because lint rules are errors.

Anyway to disable those errors on included assets? If that helps, the files have a `*.pp` extension.

jods4:
This should be ready for merge now:
- I added an option `UseNullableTypesMetadata` so that there is no back-compat issue.

- I added support for `[Association]`, when it is not a collection.

- I added some metadata tests.

- I suppressed warnings from the imported sources so that the build passes.

- Back-merged master.

Directory.Packages.props
```diff
@@ -13,6 +13,8 @@
 		<PackageVersion Include="Microsoft.CSharp"                                      Version="4.7.0"         />
 		<PackageVersion Include="Microsoft.Bcl.AsyncInterfaces"                         Version="7.0.0"         />
 		<PackageVersion Include="System.ComponentModel.Annotations"                     Version="5.0.0"         />
+		<!--Source of NullabilityInfoContext for runtimes before .net 6-->
```
MaceWindu, line 16: update to 2.1.0

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -101,6 +102,52 @@ public static class Configuration
 		/// Set to 0 to truncate all string data.
 		/// </remarks>
 		public static int MaxStringParameterLengthLogging { get; set; } = 200;
+
```
MaceWindu, line 105: Would be nice to support new options with `UseNullability` + `WithNullability` configuration helpers. Only question is where to put this option in `DataOptions`. Maybe to `LinqOptions` (then move it to `Configuration.Linq`)?

Source/LinqToDB/Mapping/DynamicColumnInfo.cs
```diff
@@ -106,6 +107,10 @@ public override object[] GetCustomAttributes(bool inherit)
 		public override object[] GetCustomAttributes(Type attributeType, bool inherit)
 			=> Array<object>.Empty;
 
+		/// <inheritdoc cref="MemberInfo.GetCustomAttributesData()"/>
```
MaceWindu, line 110: Why do we need it?

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -101,6 +102,52 @@ public static class Configuration
...
+			get => _useNullableTypesMetadata;
+			set 
+			{
+				// Can't change the default value of "false" on platforms where nullable metadata is unavailable.				
+				if (value) EnsureNullableReflectionSupported();					
+				_useNullableTypesMetadata = value;
+			}
+		}	
+
+		private static void EnsureNullableReflectionSupported()
```
MaceWindu, line 126: Maybe add separate class for NRT helpers?
- `EnsureNullableReflectionSupported` method
- `AnalyzeCanBeNull` method from association descriptor
- `AnalyzeCanBeNull` method from from column descriptor

MaceWindu (changes requested):

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -101,6 +102,52 @@ public static class Configuration
...
+		private static bool _useNullableTypesMetadata;
+		/// <summary>
+		/// Whether or not Nullable Reference Types annotations from C#
+		/// are read and taken into consideration to determine if a
+		/// column or association can be null.
+		/// Nullable Types can be overriden with explicit CanBeNull
+		/// annotations in [Column] or [Nullable].
+		/// </summary>
+		/// <remarks>Defaults to false.</remarks>
+		public static bool UseNullableTypesMetadata
```
sdanyliv, line 115: Since in it will be targeted to Version 5.0, it should be duplicated in `DataOptions`

MaceWindu:
Build fails after merge with master

jods4:
@MaceWindu there seems to be 2 different errors.

1. **'ThrowHelper' is inaccessible due to its protection level** in `Configuration.cs`. 
I have no checked where the declaration is, it's strange that this is a protection level issue.
What's the final decision regarding ThrowHelpers? Worst case the easy fix is to replace with a `throw`.

2. **The type or namespace name 'NullabilityInfoContext' could not be found**
That's strange, because it should be behind `#if` preprocessor on targets that are not supported (neither native, nor by the Nuget package).
I'll look into it later.

MaceWindu:
1. yep, use `throw`. protection level error probably because it finds internal class in runtime assemblies

Source/LinqToDB/Common/Configuration.cs
jods4: @MaceWindu @sdanyliv can you tell me a bit more about the new Options handling coming up in v5?
"should be duplicated"? Is the new Options not a replacement of the current system? 
Do you expect that some users of v5 will stick to existing code?

Directory.Packages.props
jods4: Done.

Source/LinqToDB/Common/Configuration.cs
MaceWindu: If you check `Configuration.Linq` properties, you will notice that options is not stored there anymore. Now they stored in new `LinqOptions` settings class, used by new Options implementation, where:
-  `Configuration.Linq.Options` : application-wide default set of options
- user could override settings on context level using `DataOptions` class and extension methods to set specific option value

So what I'm asking here:
- add new NRT option to `LinqOptions` class with boolean property in `Configuration.Linq` class to set application-wide behavior
- add extension methods for `DataOptions` and `LinqOptions` classes to modify this new setting (like it's done to other properties in `LinqOptions`)
- instead of access to `Configuration.Linq.UseNullableTypesMetadata` application-wide property in code you should pass current context options to code, that use it

Source/LinqToDB/Mapping/DynamicColumnInfo.cs
jods4: Some tests failed, because base implementation in `MemberInfo` simply throws `NotImplementedException`.

I think it's invoked by this PR because we're now probing Nullability attributes on every mapped member.

Source/LinqToDB/Common/Configuration.cs
jods4: Done: factored out `Nullability` that contains code related to C# NRT.

`AnalyzeCanBeNull` are still in place because they do a little more than that.
E.g. in `AssociationDescriptor` there's a check whether the association is a list, and there's a ton of other heuristics in `ColumnDescriptor`.

Source/LinqToDB/Common/Configuration.cs
jods4: Does it work with mapping data? 
Nullability info is cached in mapping, namely in `ColumnDescriptor` and `AssociationDescriptor`.

`ColumnDescriptor` has a `MappingSchema`, so if there's a distinct `MappingSchema` per configuration that is doable. Although `ColumnDescriptor` implements `IConfigurationId` there doesn't seem to be any access to datacontext-specific configuration in it (yet)?

`AssociationDescriptor` doesn't even have a reference to the `MappingSchema`, although it's constructed in 2 different places, that both could provide a `MappingSchema` if required.

@MaceWindu as you've designed the new options system, maybe it's easier if you do the conversion here? I feel like there's some things to design with respect to accessing options from mappings?
I have addressed all the other feedback on this PR.

viceroypenguin:
> Ok, so I added `Nullability.Source` into the project, but **note that it needs .net standard 2.0+**.

Would recommend at this point to use `PolySharp` to polyfill missing attributes. This may be a more general recommendation for replacing other polyfill code we have manually copied into l2db codebase, and so it should be a separate PR.

viceroypenguin (approved):

.gitignore
```diff
@@ -24,3 +24,5 @@ linq2db.sln.ide/
 /.temp
 .vs/
 packages/
+.vscode/
+Source/Test/
```
MaceWindu, line 28: btw, what is source/test?

jods4: My bad! This is local stuff I did not intend on committing. ☹
Just remove it.

Source/LinqToDB/Common/Configuration.cs
MaceWindu: @jods4 , yep, you are right. Options bould to context and it doesn't work well with mapping schema. Let's leave it as it for now

MaceWindu (dismissed):

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -401,7 +399,7 @@ public virtual bool ShouldSkip(object obj, EntityDescriptor descriptor, SkipModi
 		/// Gets value converter for specific column.
 		/// </summary>
 		public IValueConverter? ValueConverter  { get; }
-
+
```
Shane32, line 402:
```suggestion

```

Source/LinqToDB/Mapping/Nullability.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+using System.Reflection;
+
+namespace LinqToDB.Mapping
+{
+	static class Nullability
+	{
+		private static Lazy<bool> _isSupported = new(() => 
+		{
+#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER || NET5_0_OR_GREATER
```
Shane32, line 10: I believe .NET 5 targets include NETCOREAPP2_0_OR_GREATER.

Ref: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

```suggestion
#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER
```

Source/LinqToDB/Mapping/Nullability.cs
```diff
@@ -0,0 +1,64 @@
...
+				throw new NotSupportedException(
+					"Nullable metadata is not available. " +
+					"Check that you run on .net standard 2.0 or greater, " + 
+					"and that you don't trim nullable metadata during build (see NullabilityInfoContextSupport).");
+			}
+		}
+
+		public static bool TryAnalyzeMember(MemberInfo member, out bool isNullable)
+		{
+#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER || NET5_0_OR_GREATER
```
Shane32, line 44: ditto

```suggestion
#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER			
```

Source/LinqToDB/Mapping/Nullability.cs
```diff
@@ -0,0 +1,64 @@
...
+			return false;
+		});
+
+		public static void EnsureSupport()
+		{
+			if (!_isSupported.Value)
+			{
+				throw new NotSupportedException(
+					"Nullable metadata is not available. " +
+					"Check that you run on .net standard 2.0 or greater, " +
```
Shane32, line 37:
```suggestion
					"Check that you run on .NET Standard 2.0 or greater, " + 
```

Source/LinqToDB/Properties/GlobalSuppressions.cs
```diff
@@ -0,0 +1,11 @@
+using System.Diagnostics.CodeAnalysis;
+
+#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER || NET5_0_OR_GREATER
```
Shane32, line 3:
```suggestion
#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER
```

Tests/Linq/Mapping/CanBeNullTests.cs
```diff
@@ -0,0 +1,168 @@
...
+			[Association(CanBeNull = false, ThisKey = "Fake", OtherKey = "Id")]
+			public IList<Related>? C2 { get; set; } = null!;
+		}
+
+		sealed class Related
+		{
+			public int Id { get; set; }
+		}
+
+#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER || NET5_0_OR_GREATER
```
Shane32, line 80:
```suggestion
#if NET461_OR_GREATER || NETSTANDARD2_0_OR_GREATER || NETCOREAPP2_0_OR_GREATER
```

Shane32 (dismissed):

Shane32:
Looks great!

Source/LinqToDB/Mapping/Nullability.cs
jods4: Yeah, I believe that might actually be the case.
If you're 100% sure you can remove it, otherwise it doesn't hurt anyway.

.gitignore
MaceWindu, line 28:
```suggestion
```

PR #3657, opened 2022-07-14
Add a test for Associations

jods4:
Exercises ExpressionPredicate + CanBeNull = false

I expect this to fail, in my project this configuration results in an OUTER JOIN despite CanBeNull = false.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/AssociationHelper.cs
+++ b/Source/LinqToDB/Linq/Builder/AssociationHelper.cs
@@ -119,8 +119,8 @@ namespace LinqToDB.Linq.Builder
 
 				if (expressionPredicate != null)
 				{
-					shouldAddDefaultIfEmpty = true;
-					shouldAddCacheCheck = true;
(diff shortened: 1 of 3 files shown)


MaceWindu:
@jods4 there are several issues:
- expression-based association ignores CanBeNull - fixed
- count ignores associations in select - "fixed"
- optimizer removes "unused" join - not fixed

jods4:
@MaceWindu Thanks!

Regarding the last 2 "optimization" points:

On one hand, the optimizer should not remove inner joins from select (count or otherwise) because they can remove rows (and more generally besides associations: also multiply rows). Unless, of course, we have some metadata that says the join is 1:1.

On the other hand, we could debate if the `Include` API applies to SELECT that doesn't return the target entity, like a `Count`. There is an argument to be made that maybe `Include` is only meaningful if the parent entity is part of the resultset, so it's ok to ignore in a `Count`. I think this is open to interpretation.

MaceWindu:
replaced by #3921

PR #3627, opened 2022-06-25
ClickHouse provider

MaceWindu:
Fix #1796

**List of changes**

ClickHouse support:
- added `ClickHouse` provider with support of:
  - MySQL interface using [MySqlConnector](https://github.com/mysql-net/MySqlConnector/)
  - HTTP(S) interface using [ClickHouse.Client](https://github.com/DarkWanderer/ClickHouse.Client)
  - TCP (binary) interface using [Octonica.ClickHouseClient](https://github.com/Octonica/ClickHouseClient)
- ClickHouse database scaffold support
  -  only from CLI tool (all 3 interfaces supported)
  - only tables/views scaffolded (ClickHouse doesn't provide metadata for functions)
- added new bulk copy options `BulkCopyOptions.MaxDegreeOfParallelism` and `BulkCopyOptions.WithoutSession` for `ClickHouse.Client` HTTP provider
- added `Data/Setup Scripts/clickhouse.cmd` script to setup local docker instance of ClickHouse (Windows, requires WSL)

**Notes**

- PostgreSQL protocol [not supported](https://github.com/ClickHouse/ClickHouse/issues/18611)
- MySQL interface [doesn't support](https://github.com/ClickHouse/ClickHouse/issues/28935) connection reuse (AKA connection pooling). Must be disabled in connection string
- Parameters not supported due to poor support by providers/ClickHouse
- for types support and type support issues per-provider check type mapping table `Source/LinqToDB/DataProvider/ClickHouse/README.md`
- Many tests disabled for ClickHouse due to lack of support for correlated queries in ClickHouse and limited support of `ON` join conditions
- `InsertOr[Update/Replace]` API not implemented due to `ClickHouse` limitations
- Custom hints/query extensions will be implemented later as separate PR
- LinqPAD support will be added later

***Non-default settings for better test compatibility***

- `join_use_nulls: 1` : enable `NULL` values for missing outer records in joins
- `mutations_sync: 1` : execute delete/update queries synchonously to not fail multi-query tests
- `allow_experimental_object_type: 1`: to test `JSON` type
- `allow_experimental_geo_types: 1`: to test `Geo` types

PR #3686, opened 2022-08-08
Fix datetime strings generation

MaceWindu:
Extracted from #3627

Fixes generation of datetime strings to have fixed length for parts with leading zeroes, e.g. generate `01:02` time part instead of `1:2`

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
@@ -37,7 +37,7 @@ namespace LinqToDB.SqlProvider
 		{
 			FixRootSelect (statement);
 			FixEmptySelect(statement);
-			FinalizeCte(statement);
+			FinalizeCte   (statement);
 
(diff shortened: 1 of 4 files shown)


sdanyliv (approved):

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2557,6 +2557,36 @@ protected static bool IsDateDataType(ISqlExpression expr, string dateName)
 			};
 		}
 
+		protected static bool IsSmallDateTimeType(ISqlExpression expr, string typeName)
+		{
+			return expr.ElementType switch
+			{
+				QueryElementType.SqlDataType   => ((SqlDataType)expr).Type.DataType == DataType.SmallDateTime,
```
sdanyliv, line 2564: Maybe it is better to use `QueryHelper.GetDbDataType(...)`

viceroypenguin (approved):

PR #3690, opened 2022-08-09
Obsolete `SqlDataType.GetDataType`

MaceWindu:
Another part of #3627

- obsolete `SqlDataType.GetDataType` method and update existing calls to use `MappingSchema.GetDataType`. Having and using two different type definitions is not a good idea, mkay?
- remove casts to `object` from expression tree as it introduces unnecessary `CAST AS VARIANT` nodes to AST. caused by object-typed APIs like `FromSql(string, params object[])`
- remove useless casts of `enum`s to `Enum` from expression tree (caused by use of `Enum` typed API, e.g. methods of `Enum` class)
- **[BREAKING]**: Conversion functions that doesn't specify exact type (`System.Convert.ToDecimal`, `LinqToDB.Common.Convert`) change behavior for decimal type. They will generate cast to decimal type definition, specified in mapping schema (usually `decimal` without precision and scale set). Previously they used hardcoded `decimal(29, 10)` type.
  - Workarounds:
    - register suitable decimal type mapping in mapping schema (not recommended): `mappingSchema.AddScalarType(typeof(decimal), new SqlDataType(new DbDataType(typeof(decimal), DataType.Decimal, null, null, <precision>, <scale>)));`
    - use better-typed convert API, e.g. `Sql.Convert(Sql.Types.Decimal(<precision>,<scale>), convertedValue)`
  - List of affected providers:
    - DB2: `decimal(29, 10)` -> `decimal`
    - Firebird: `decimal(18, 10)` -> `decimal`
    - SQL Sever CE: `decimal(29, 10)` -> `decimal`
    - SQL Sever: `decimal(29, 10)` -> `decimal`
    - MySQL/MariaDB: `decimal(29, 10)` -> `decimal`
    - SAP/Sybase ASE: `decimal(29, 10)` -> `decimal`
    - Oracle: `decimal(29, 10)` -> `decimal`

This PR is not finished - I will tune mappings based on baselines to reduce number of baseline regressions. Cannot be avoided completely as we used two incompatible type definitions at the same time in different places.

After all 4 PRs merged I will create one more PR with `ConvertBuilder` changes (last one)


viceroypenguin (dismissed):

sdanyliv (dismissed):

MaceWindu:
Extra `()` appeared in `TestTableValueFunction` in baselines is correct (even if it looks unnecessary)

Source/LinqToDB/DataProvider/Firebird/FirebirdSqlOptimizer.cs
```diff
@@ -3,6 +3,7 @@
 
 namespace LinqToDB.DataProvider.Firebird
 {
+	using System.Collections.Generic;
```
viceroypenguin, line 6: nit: move to other `using System.*;` at top of file.

Source/LinqToDB/DataProvider/Firebird/FirebirdSqlOptimizer.cs
```diff
@@ -343,6 +297,47 @@ private SqlStatement WrapParameters(SqlStatement statement, EvaluationContext co
...
+
+				// part of output clause
+				if (parents[i] is SqlOutputClause)
+					return true;
+
+				// insert or update keys used in merge source select query
+				if (parents[i] is SqlSetExpression set
+					&& i == 2
+					&& (parents[i - 1] is SqlInsertClause || parents[i - 1] is SqlUpdateClause)
+					&& parents[i - 2] is SqlInsertOrUpdateStatement insertOrUpdate
```
viceroypenguin, line 320: Since `i==2` per line 318, simplify constants here:

```suggestion
					&& (parents[1] is SqlInsertClause || parents[1] is SqlUpdateClause)
					&& parents[0] is SqlInsertOrUpdateStatement insertOrUpdate
```

Source/LinqToDB/Sql/Sql.ExpressionAttribute.cs
```diff
@@ -8,6 +8,7 @@
 
 namespace LinqToDB
 {
+	using LinqToDB.Common;
 	using LinqToDB.Expressions;
```
viceroypenguin, line 12:
```suggestion
	using Common;
	using Expressions;
```

viceroypenguin (approved):

PR #3730, opened 2022-08-23
Group by eager load parameters fix

### LINKED ISSUE #3636: SQL generation bug
this bug happens on all provider (I tested on SQL server and PostgreSQL).
This is a weird bug. My coworker wrote a very long and complicated LinQ, and can never got the right result. (actually we got complaint from our client, so this is a painful experience) So I simplified the Linq, and eventually figureout there might be a bug with grouping.
The final Linq May not make sense to you, that's because I simplified it, and the you won't see the business logic in it any more. 

Anyway, here is the code to recreate the bug:
basically, the parameter passed into the final SQL is wrong the second time. 
When I call Test(2), it looks fine, but the finally SQL looks odd, even the result is correct. The finally SQL created two parameter for myid:  :myid, :ID2

This is fine, but when I call Test(85), funny things start to happen:
:myid = 2
:ID2 = 85
were passed in. But they both should be 85.

It's really odd, but I recreated the test case, and always got the same wrong result.
You can simply copy the following code, and run it, and see it yourself.

I print the parameter value using the intercepter.

```cs
using static System.Console;
using LinqToDB;
using LinqToDB.Mapping;
using LinqToDB.Data;
using LinqToDB.Interceptors;
using System.Data.Common;
AppDataContext db = new("PostgreSQL", "conn");
db.CreateTable<T1>();
db.CreateTable<T2>();

db.Insert(new T1 { ID = 1, ID2 = 2 });
db.Insert(new T1 { ID = 1, ID2 = 85 });
db.Insert(new T1 { ID = 2, ID2 = 2 });
db.Insert(new T1 { ID = 2, ID2 = 85 });

db.Insert(new T2 { ID = 1, ID2 = 2 });
db.Insert(new T2 { ID = 1, ID2 = 2 });

Test(2);
Test(85);


void Test(int myid)
{
	var q = from s in db.GetTable<T1>().Where(x => x.ID2 == myid)
			join o in db.GetTable<T2>().Where(x => x.ID2 == myid) on s.ID equals o.ID into temp1
			from order in temp1.DefaultIfEmpty()
			group s by s.ID into g
			select g;

	WriteLine(System.Text.Json.JsonSerializer.Serialize(q.FirstOrDefault()));
}

public partial class AppDataContext : DataConnection
{
	public AppDataContext(string providerName, string connectionString) : base(providerName, connectionString)	
	{
		AddInterceptor(new MyCMD());
	}

	class MyCMD : LinqToDB.Interceptors.CommandInterceptor
	{
		public override DbCommand CommandInitialized(CommandEventData eventData, DbCommand command)
		{
			Console.WriteLine(command.CommandText);
			foreach (DbParameter p in command.Parameters) Console.WriteLine(p.ParameterName + " : " + p.Value);
			return base.CommandInitialized(eventData, command);
		}
	}
}

[Table]
public class T1
{
	/// <summary>
	/// ID//
	/// </summary>
	[Column("id")] public int ID { get; set; } // integer
	[Column("id2")] public int ID2 { get; set; } // integer
	[Column("id3")] public int ID3 { get; set; } // integer
}


[Table]
public class T2
{
	/// <summary>
	/// ID//
	/// </summary>
	[Column("id")] public int ID { get; set; } // integer
	[Column("id2")] public int ID2 { get; set; } // integer
}
```

MaceWindu:
Fix #3636

Updated test to test issue properly and re-introduced reverted recently fix

### DIFF
--- a/Build/Azure/net60/clickhouse.json
+++ b/Build/Azure/net60/clickhouse.json
@@ -2,7 +2,8 @@
 	"NET60.Azure": {
 		"Providers": [
 			"ClickHouse.Octonica",
-			"ClickHouse.Client",
+			// disabled due https://github.com/DarkWanderer/ClickHouse.Client/issues/152
+			//"ClickHouse.Client",
 			"ClickHouse.MySql"
 		]
 	}
--- a/Build/Azure/netcoreapp31/clickhouse.json
+++ b/Build/Azure/netcoreapp31/clickhouse.json
@@ -2,7 +2,8 @@
 	"CORE31.Azure": {
 		"Providers": [
 			"ClickHouse.Octonica",
-			"ClickHouse.Client",
+			// disabled due https://github.com/DarkWanderer/ClickHouse.Client/issues/152
+			//"ClickHouse.Client",
 			"ClickHouse.MySql"
 		]
 	}
--- a/Build/Azure/pipelines/templates/test-matrix.yml
+++ b/Build/Azure/pipelines/templates/test-matrix.yml
@@ -930,7 +930,7 @@ jobs:
           enable_fw_netcore31: true
           enable_fw_net60: true
           is_experimental: false
-          docker_login: true
+          docker_login: false
           ${{ if or(contains(parameters.db_filter, '[all]'), contains(parameters.db_filter, '[saphana.all]')) }}:
             enabled: true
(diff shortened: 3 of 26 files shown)


viceroypenguin (dismissed):

sdanyliv (approved):

MaceWindu:
changes not related to PR, just SAP and MariaDB decided to broke at the same time...

PR #3732, opened 2022-08-23
ArrayBindCount must be reset to 0 after BulkCopy

### LINKED ISSUE #3731: Regression in Oracle array-bound BulkCopy
After updating to latest 4.1.1 (from 3.x), we have a huge regression in BulkCopy when using the proprietary Oracle strategy using array-bound parameters.

All queries on the same `DataConnection` after `BulkCopy` fail with an `InvalidCastException`, that can't convert a random type (int32, string, DateTime) to Array.

It seems linq2db tries to re-use `DbCommand` objects, but it doesn't reset them properly.
Specifically, after the bulk copy, the `ArrayBindCount` property is not set back to 0.
So future queries that re-use the command are still in array binding mode... and fail because their parameters are not set to array values, of course.

jods4:
Fixes #3731


viceroypenguin:
Any chance you could add a unit test for the bug you encountered and this fixes to confirm validity?

jods4:
@viceroypenguin don't have much time right now. I'm running the Oracle tests to ensure no regression, and I can easily test the built package in my projects to confirm the fix.

viceroypenguin (dismissed):

Source/LinqToDB/DataProvider/Oracle/OracleDataProvider.cs
```diff
@@ -149,17 +147,15 @@ public override DbCommand InitCommand(DataConnection dataConnection, DbCommand c
...
-						{
-							Adapter.SetArrayBindCount(rawCommand, value.Length);
-							break;
-						}
-					}
+				// Always reset SetArrayBindCount, in case it was modified by a previous command.
+				// When using array bindings, _all_ parameters must be arrays and have the same length.
+				if (Adapter.SetArrayBindCount != null)
+				{
+					int arrayBindCount = parameters?.Length > 0 && parameters[0] is { IsArray: true, Value: object[] value }
```
sdanyliv, line 154: Why just first parameter? `parameters[0]`

jods4: Because when you use ArrayBind all parameters must be arrays of the same size, as I indicated in the comment on line 151.

With ArrayBind, the SQL command will be executed N times with values [0..N] from each parameter.

The only consumer of this right now is the alternate BulkCopy and you can easily check that the implementation puts an array in each parameter.

MaceWindu (approved):

Issue #3658, opened 2022-07-14
CanBeNull is not respected in association without keys


jods4:
When using an association that is defined by an `ExpressionPredicate`, it looks like `CanBeNull` is ignored and the generated SQL is always an `OUTER JOIN`, even if you set `CanBeNull = false`.

We have no test for `ExpressionPredicate` + `CanBeNull = false`, so I added one in PR #3657.
As expected, it fails.

(If you look at the test case and SQL carefully, you'll notice that because I don't use any association field (I just did a quick `count(*)`), the OUTER JOIN is dropped altogether. That's still a valid test case, as an INNER JOIN would not be dropped.)

PR #3751, opened 2022-08-28
Force nullability on lef-joined columns

MaceWindu:


MaceWindu:
@sdanyliv promised test + test for another one issue with CanBeNull

jods4:
@MaceWindu If you're tweaking Association CanBeNull, could you look at #3658 ?
It's a small fix and I've already provided a test case in PR #3657

sdanyliv (approved):

MaceWindu:
Only `Contains11` test baselines changes are suspicious, but looks like change is valid to me

jods4:
This is a great step forward. 👍
When I was working on refactoring the optimizer I encountered plenty of issues because `CanBeNull` is sometimes imprecise, which can cause subtle bugs when the SQL generation depends on it.

The changes are not straightforward to review, does this fix `CanBeNull` wherever it's read? 
- Accessing `CanBeNull` inside the join condition respects the source expression nullability?
- Accessing `CanBeNull` outside the join is always true for fields derived from `[LEFT|RIGHT|OUTER] [JOIN|APPLY]`?

Looking at the changes it looks like this is achieved by wrapping the expression with one level of pass-through `{0}` expression.
These are quite annoying, is it not possible to do this cleanly, with a flag on the `TableSource` that is taken into account for its field references? (The tricky part is that this flag should not apply when inside the join condition.)

I'm surprised more code isn't removed with this PR, I recall `ExprExpr` conversion calls a helper that, if I remember correctly, checks for extra nullable scenarios like outer joins (which shouldn't be necessary anymore).

MaceWindu:
This is just a workaround for current parser. New parser handles it in better way: there is no association context anymore and instead of {0} expression it introduce new AST element SqlNullabilityExpression

jods4:
@MaceWindu great.

Any news regarding the new parser being merged? 
I'm really scared by the scope of that merge! 😅
It also blocks a lot of other impacting changes (I wanted to work on simplifying our AST).

sdanyliv:
I would like to finish it as soon as possible. But sometimes I'm stuck with found workarounds starts writing new proper code.

PR #3770, opened 2022-09-08
[WIP] Introducing an AST factory

jods4:
This is an attempt at improving the internals of linq2db, there is no change for users.
As linq2db is large and complex, the changes in the PR are just "getting started" and far from complete, although as I will discuss below it should probably be merged incrementally in multiple steps..

## Main idea: add an AST factory
The core idea of this PR is to introduce a factory class to create our SQL abstract syntax tree.
So, instead of new'ing up nodes with `new SqlPredicate.IsNull(...)` and friends, we call into `ast.IsNull(...)`.

## What's the difference anyway?!
`new T` must return an instance of `T`, whereas a factory could return any `ISqlExpression`.
This opens up quite a few possibilities:

### Local optimizations
The AST factory can apply local transformations and return an equivalent, optimized expression. I think I've seen this referred to as _optimization by construction_.

For example, if you call `and(...)` with a series of nodes that incl. a `false` constant, the factory would just return a `false` constant; Nested calls to `coalesce(...)` can be flattened in a single node.

Today, these transformations are generally done by `SqlOptimizer`. Doing them at construction time largely reduces the number of allocations, which is beneficial for performance.

To illustrate, consider the expression `isnull(row(x, y))` in a provider that doesn't support `row`.
Today, compilation could go like this:
```
isnull( row(x, y) )         # Parsed AST -> 2 nodes (in addition to x, y)
and( isnull(x), isnull(y) ) # Optimizer lowers to non-row code -> 3 new nodes
and( isnull(x), false )     # Optimizer notices y cannot be null, simplifies -> 2 new nodes
false                       # Optimizers sees "and" is always false -> 0 new node (hopefully reuses false node)
```
With a builder, the same process happens but immediately at construction without visiting anything, and without constructing intermediate nodes:
```
ast.row(x, y)           # Temporary row(x, y) must still be built by expression parser -> 1 node.
ast.isnull( row(x, y) ) # This call notices the row and immediately calls SQL equivalent -> 0 node.
ast.and( 
  ast.isnull( x ),      # Allocates a temporary isnull(x) -> 1 node
  ast.isnull( y )       # Notices y is non-nullable and returns a cached "false" node -> 0 allocation
)                       # ast.and call notices one parameter is false and returns it -> 0 allocation
```
Summary: current design allocates 7 nodes (in addition to x, y); factory only allocates 2!

> **Note**
> This doesn't remove or replace `SqlOptimizer`. It is still useful for _global_ optimizations, i.e. transforms that need more context than just the current node.

### Interning
Common constants or expressions such as `null`, `0`, `1`, `true`, `false` are cached and re-used, reducing allocations.
If the codebase is 100% going through the AST factory, we can take advantage of the fact that values are interned and make reference equalities rather than pattern matching to check if something is a well-known value like `0`.

### Provider-specific translations
The factory could be a static class, but I've made it an instance, so that we can have provider-specific factories.
This enable provider-specific transformations, such as lowering unsupported expressions like `SqlRow` in the previous example.

### Enforcing invariants
The factory can enforce some constraints on our AST nodes, for example we could guarantee that a some structures like `or`, `coalesce` are always flattened, which makes further processing more efficient.

### Performance
As explained above, a factory can apply transformations at construction without allocating nodes.
This improves performance because it reduces GC pressure and requires less "visiting".
### Separation of concerns
Some optimizations are actually made by the expression parser.
This complicates the parser with things that are not its responsibilities. 
Furthermore, those optimizations are either not applied later when the tree is transformed, or duplicated.
With a dedicated class that consistently does these transforms, the calling code can be simplified.

I think that if you look at the code that is already modified in this PR, the result is IMHO generally much cleaner -- and I'm only getting started.
### Future-proof
This is a first step, that will help future refactorings. 
I think our SQL AST is not ideal and there are changes I would like to make for cleaner code and more efficient execution.
This abstraction layer, once complete, will make it easier to apply those changes project-wide.

### Provider-independent remoting?
This is a long-term idea that is not on my map for now, but anyway:
Instead of serializing a provider specific AST on the client, we could call into a "generic" AST factory that does nothing but creates a neutral structure to serialize these calls.
On the server, we can rebuild the AST by making the same calls into a provider-specific factory.
This could make the client provider-agnostic, which is nice to have, IMHO.

## Rollout
Obviously this is a very large-scale change, as there are calls to AST constructors all over the place.
I've started the work just so that you can get a better grasp of what the code would actually look like.

I think this is gonna be too annoying to merge to fully do in a single PR.
**If we commit to this idea**, I think we should regularly merge progress. Not all the code calls into the factory, but everything is still working and can be released.

There's a lot to do, in waves:
- Replace all ctors usage;
- Make AstFactory provider-specific, move provider transformations into it;
- Move local optimizations in factory (from SqlOptimizer and other places);

### DIFF
--- a/Source/LinqToDB/Sql/Sql.Row.cs
+++ b/Source/LinqToDB/Sql/Sql.Row.cs
@@ -79,7 +79,7 @@
 			{
 				var args = Array.ConvertAll(builder.Arguments, x => builder.ConvertExpressionToSql(x));
 				builder.ResultExpression = new SqlSearchCondition(new SqlCondition(false,
-					new SqlPredicate.ExprExpr(args[0], SqlPredicate.Operator.Overlaps, args[1], false)));
+					new SqlPredicate.ExprExpr(args[0], SqlPredicate.Operator.Overlaps, args[1], null)));
 			}
 		}
 	}
--- a/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
@@ -18,12 +18,15 @@ namespace LinqToDB.SqlProvider
 	{
 		#region Init
 
-		protected BasicSqlOptimizer(SqlProviderFlags sqlProviderFlags)
+		protected BasicSqlOptimizer(SqlProviderFlags sqlProviderFlags, AstFactory ast)
 		{
 			SqlProviderFlags = sqlProviderFlags;
+			this.ast         = ast;
 		}
 
 		public SqlProviderFlags SqlProviderFlags { get; }
+
+		protected readonly AstFactory ast;
 
 		#endregion
 
@@ -1082,9 +1085,7 @@ namespace LinqToDB.SqlProvider
 					    func.Parameters[2] is SqlSearchCondition sc &&
 					    func.Parameters[1].TryEvaluateExpression(context, out var v1) && v1 is null)
 					{
-						if (isTrue.IsNot)
-							return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
-						return sc;
+						return isTrue.IsNot ? ast.Not(sc) : sc;
 					}
 				}
 			}
@@ -1202,7 +1203,7 @@ namespace LinqToDB.SqlProvider
 							return sc;
 
 						if (Equals(value, v2) && !sc.CanBeNull)
-							return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+							return ast.Not(sc);
 					}
 				}
 			}
@@ -1320,37 +1321,23 @@ namespace LinqToDB.SqlProvider
 
 					if (expr.WithNull == null && (expr.Operator == SqlPredicate.Operator.Equal || expr.Operator == SqlPredicate.Operator.NotEqual))
 					{
-						if (expr.Expr2 is ISqlPredicate)
+						if (expr.Expr2 is ISqlPredicate pred2)
 						{
 							var boolValue1 = QueryHelper.GetBoolValue(expr.Expr1, context);
 							if (boolValue1 != null)
 							{
-								ISqlPredicate transformed = new SqlPredicate.Expr(expr.Expr2);
 								var isNot = boolValue1.Value != (expr.Operator == SqlPredicate.Operator.Equal);
-								if (isNot)
-								{
-									transformed =
-										new SqlPredicate.NotExpr(expr.Expr2, true, Precedence.LogicalNegation);
-								}
-
-								return transformed;
+								return isNot ? ast.Not(pred2) : new SqlPredicate.Expr(expr.Expr2);
 							}
 						}
 
-						if (expr.Expr1 is ISqlPredicate)
+						if (expr.Expr1 is ISqlPredicate pred1)
 						{
 							var boolValue2 = QueryHelper.GetBoolValue(expr.Expr2, context);
 							if (boolValue2 != null)
 							{
-								ISqlPredicate transformed = new SqlPredicate.Expr(expr.Expr1);
 								var isNot = boolValue2.Value != (expr.Operator == SqlPredicate.Operator.Equal);
-								if (isNot)
-								{
-									transformed =
-										new SqlPredicate.NotExpr(expr.Expr1, true, Precedence.LogicalNegation);
-								}
-
-								return transformed;
+								return isNot ? ast.Not(pred1) : new SqlPredicate.Expr(expr.Expr1);
 							}
 						}
 					}
@@ -1378,7 +1365,7 @@ namespace LinqToDB.SqlProvider
 				{
 					var expr = (SqlPredicate.NotExpr)predicate;
 
-					if (expr.IsNot && expr.Expr1 is SqlSearchCondition sc)
+					if (expr.Expr1 is SqlSearchCondition sc)
 					{
 						if (sc.Conditions.Count == 1)
 						{
@@ -1583,11 +1570,11 @@ namespace LinqToDB.SqlProvider
 		{
 			var newPredicate = !between.IsNot
 				? new SqlSearchCondition(
-					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.GreaterOrEqual, between.Expr2, withNull: false)),
-					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.LessOrEqual,    between.Expr3, withNull: false)))
+					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.GreaterOrEqual, between.Expr2, withNull: null)),
+					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.LessOrEqual,    between.Expr3, withNull: null)))
 				: new SqlSearchCondition(
-					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.Less,    between.Expr2, withNull: false), isOr: true),
-					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.Greater, between.Expr3, withNull: false)));
+					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.Less,    between.Expr2, withNull: null), isOr: true),
+					new SqlCondition(false, new SqlPredicate.ExprExpr(between.Expr1, SqlPredicate.Operator.Greater, between.Expr3, withNull: null)));
 
 			return newPredicate;
 		}
@@ -1954,7 +1941,7 @@ namespace LinqToDB.SqlProvider
 									be.SystemType,
 									be.Expr1,
 									be.Operation,
-									ConvertExpressionImpl(PseudoFunctions.MakeConvert(new SqlDataType(DataType.VarChar, typeof(string), len.Value), new SqlDataType(be.Expr2.GetExpressionType()), be.Expr2), visitor),
+									ConvertExpressionImpl(ast.Convert(new SqlDataType(DataType.VarChar, typeof(string), len.Value), be.Expr2), visitor),
 									be.Precedence);
 							}
 
@@ -1967,7 +1954,7 @@ namespace LinqToDB.SqlProvider
 
 								return new SqlBinaryExpression(
 									be.SystemType,
-									ConvertExpressionImpl(PseudoFunctions.MakeConvert(new SqlDataType(DataType.VarChar, typeof(string), len.Value), new SqlDataType(be.Expr1.GetExpressionType()), be.Expr1), visitor),
+									ConvertExpressionImpl(ast.Convert(new SqlDataType(DataType.VarChar, typeof(string), len.Value), be.Expr1), visitor),
 									be.Operation,
 									be.Expr2,
 									be.Precedence);
@@ -2003,8 +1990,10 @@ namespace LinqToDB.SqlProvider
 			return expression;
 		}
 
-		protected virtual ISqlExpression ConvertFunction(SqlFunction func)
+		protected virtual ISqlExpression ConvertFunction(ISqlExpression expr)
 		{
+			if (expr is not SqlFunction func) return expr;
+			
 			switch (func.Name)
 			{
 				case "Average": return new SqlFunction(func.SystemType, "Avg", func.Parameters);
@@ -2454,7 +2443,7 @@ namespace LinqToDB.SqlProvider
 								return new SqlPredicate.Expr(new SqlValue(p.IsNot));
 
 							if (p.IsNot)
-								return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+								return ast.Not(sc);
 
 							return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
 						}
@@ -2500,7 +2489,7 @@ namespace LinqToDB.SqlProvider
 							return new SqlPredicate.Expr(new SqlValue(p.IsNot));
 
 						if (p.IsNot)
-							return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+							return ast.Not(sc);
 
 						return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
 					}
--- a/Source/LinqToDB/SqlQuery/AstFactory.cs
+++ b/Source/LinqToDB/SqlQuery/AstFactory.cs
@@ -1,0 +1,262 @@
+using System;
+using Config = LinqToDB.Common.Configuration;
+
+namespace LinqToDB.SqlQuery
+{
+	public class AstFactory
+	{
+		#region Ctor
+
+		public AstFactory() 
+		{
+			// REVIEW(jods): a real True/False predicate node might be nice for DBs that have BOOLEAN support
+			TruePredicate = Equal(One, One);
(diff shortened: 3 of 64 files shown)


Source/LinqToDB/SqlQuery/AstFactory.cs
```diff
@@ -0,0 +1,262 @@
...
+			=> new SqlBinaryExpression(sysType, a, "|", b, Precedence.Bitwise);
+
+		public ISqlExpression BitXor(ISqlExpression a, ISqlExpression b, Type sysType)
+			=> new SqlBinaryExpression(sysType, a, "^", b, Precedence.Bitwise);
+
+		#endregion
+
+		#region Expressions:Logic
+
+		public ISqlExpression Coalesce(ISqlExpression a, ISqlExpression b, Type sysType)
```
viceroypenguin, line 166: Should return `SqlFunction`?

Source/LinqToDB/SqlQuery/AstFactory.cs
```diff
@@ -0,0 +1,262 @@
...
+				_ => new[] { a, b },
+			};
+
+			return new SqlFunction(sysType, PseudoFunctions.COALESCE, isAggregate: false, isPure: true, values)
+			{
+				CanBeNull = values.All(v => v.CanBeNull),
+			};
+		}
+
+		public ISqlExpression If(ISqlExpression condition, ISqlExpression whenTrue, ISqlExpression whenFalse, Type sysType)
```
viceroypenguin, line 186: Should return `SqlFunction`?

Source/LinqToDB/SqlQuery/AstFactory.cs
```diff
@@ -0,0 +1,262 @@
...
+			{ 
+				CanBeNull = whenTrue.CanBeNull | whenFalse.CanBeNull,
+			};
+		}
+
+		#endregion
+
+		#region Expressions:Types
+
+		public ISqlExpression Convert(SqlDataType toType, ISqlExpression value, SqlDataType? fromType = null)
```
viceroypenguin, line 208: Should return `SqlFunction`?

Source/LinqToDB/Sql/Sql.Row.cs
```diff
@@ -79,7 +79,7 @@ public void Build(ISqExtensionBuilder builder)
 			{
 				var args = Array.ConvertAll(builder.Arguments, x => builder.ConvertExpressionToSql(x));
 				builder.ResultExpression = new SqlSearchCondition(new SqlCondition(false,
-					new SqlPredicate.ExprExpr(args[0], SqlPredicate.Operator.Overlaps, args[1], false)));
+					new SqlPredicate.ExprExpr(args[0], SqlPredicate.Operator.Overlaps, args[1], null)));
```
viceroypenguin, line 82:
```suggestion
					new SqlPredicate.ExprExpr(args[0], SqlPredicate.Operator.Overlaps, args[1], withNull: null)));
```

Source/LinqToDB/SqlQuery/AstFactory.cs
```diff
@@ -0,0 +1,262 @@
...
+		{
+			// Flatten nested Coalesce calls
+			var ua = QueryHelper.UnwrapExpression(a);
+			var ub = QueryHelper.UnwrapExpression(b);
+			ISqlExpression[] values = (ua, ub) switch
+			{
+				(SqlFunction { Name: "Coalesce" or PseudoFunctions.COALESCE } ca, 
+				 SqlFunction { Name: "Coalesce" or PseudoFunctions.COALESCE } cb)    => ToArray(ca.Parameters, cb.Parameters),
+				(SqlFunction { Name: "Coalesce" or PseudoFunctions.COALESCE } ca, _) => ToArray(ca.Parameters, b),
+				(_, SqlFunction { Name: "Coalesce" or PseudoFunctions.COALESCE } cb) => ToArray(a, cb.Parameters),
```
viceroypenguin, line 176: Per the commentary of the PR, these cases should be recognizing an existing `SqlFunction` object and updating it rather than preparing an array for a brand new `SqlFunction` object in l180. This probably means that `SqlFunction.Parameters` should use an `IReadOnlyList<>` or similar, where `SqlFunction provides an `internal void AddParameter(ISqlExpression a)` that updates the underlying list of that property.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2003,8 +1990,10 @@ public virtual ISqlExpression ConvertExpressionImpl(ISqlExpression expression, C
 			return expression;
 		}
 
-		protected virtual ISqlExpression ConvertFunction(SqlFunction func)
+		protected virtual ISqlExpression ConvertFunction(ISqlExpression expr)
```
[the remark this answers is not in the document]
jods4: Yeah the `ConvertFunction` business was annoying, I had to hack something to be able to proceed without making too many deep changes for now.

Long-term, I would like to see `ConvertFunction` be gone, seems to me its purpose is to switch function names so that correct SQL is generated. This could also happen at SQL generation time without modifying the AST at all, but I don't want to engage in too many refactorings at the same time.

> Would recommend AstFactory.Coalesce() returns a SqlFunction.

Seems against the principle of AST factory to constraint what is returned too much.
For instance, one local optimization I have on the map for later is that `Coalesce` can drop all its arguments after the 1st one that cannot be null.
If this happens to be the 1st argument, `Coalesce` can just return the first argument. In that case there's no guarantee that `Coalesce` returns a `SqlFunction`, could be any expression.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2454,7 +2443,7 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
 								return new SqlPredicate.Expr(new SqlValue(p.IsNot));
 
 							if (p.IsNot)
-								return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+								return ast.Not(sc);
 
 							return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
```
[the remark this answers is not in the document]
jods4: For sure will do!
I've worked very sparingly on `SqlOptimizer` so far, tried to not touch everything at once.
I will simplify for sure when `new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction)` gets replaced by a factory call, too.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2500,7 +2489,7 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
 							return new SqlPredicate.Expr(new SqlValue(p.IsNot));
 
 						if (p.IsNot)
-							return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+							return ast.Not(sc);
 
 						return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
 					}
```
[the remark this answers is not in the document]
jods4: Same, will do when `new Expr` gets replaced.

Source/LinqToDB/SqlQuery/AstFactory.cs
jods4: As explained above, I don't like this idea because we can implement local transforms that wouldn't be functions, such as returning the first `a` directly if it cannot be null.

Source/LinqToDB/SqlQuery/AstFactory.cs
jods4: Same as `Coalesce`. 
A big part of the game will be to simplify constant conditions (esp. when called after parameters are baked in), so this _will_ return either `whenTrue` or `whenFalse` sometimes -> the return value can't be more specific than `ISqlExpression`.

Source/LinqToDB/SqlQuery/AstFactory.cs
jods4: Same as `Coalesce` and `If`, rather not.

Source/LinqToDB/Sql/Sql.Row.cs
jods4: Just noticed this was a bug and fixed it _en passant_ (almost all `withNull: false` are bugs and should be `null` instead, easy confusion in what `withNull` means).

Eventually `new ExprExpr` will be replaced by a factory call.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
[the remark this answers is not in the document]
viceroypenguin: Sounds good. Wonder if a lot of the remaining functionality in this method could be converted to Function attributes or expression mappings.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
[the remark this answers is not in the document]
viceroypenguin: Sure, I only commented because you made the change on the other bit to be a ternary, but then only changed the `ast.Not()` bit on this one.

Source/LinqToDB/SqlQuery/AstFactory.cs
jods4: Yeah, I'm still trying to come up with better ways to reduce allocations and would love new ideas here.

I think our SQL AST nodes should be **immutable**. Maybe not in the strict sense, I'd maybe allow having flags set later during analysis / optimizations, but as far as what what the node is/represents.
Immutability enables re-using the node safely, for example when expanding an expression `a == b` into `a == b or a is null and b is null` without cloning the whole `a` and `b` subtree.
Another idea is that we could cache and reuse sub-expressions across various compilations.

I'm not set on what's the best way to reduce copying of immutable structures, there are options. I'd like to balance clean code and efficiency.
Some ideas:

1. This specific `coalesce` example might be common and bad when converting a long `??` chain. 
To avoid this we could complement it by some custom code in the expression converter to immediately create the flattened node.
I have done this for `or`/`and` predicates, the Expression converter immediately flattens them before calling the factory, although the factory would also flatten (by copying) later tree transforms that would put an `or` inside an `or`.

2. Maybe nodes could be mutable until they're cloned? Could be indicated by a bit on the node. 
This makes any code that want to mutate the node less clean because it has to check the bit and it needs 2 code paths for the case where it can mutate and the case where it needs a copy :(

3. For nodes that are nicer being mutated when they're first created (`or`/`and` immediately come to mind in current linq2db design), maybe we could have a Builder pattern, where a mutable `struct` is used until we're done and then an immutable node is produced.

4. Maybe we can use more sophisticated structures inside ast nodes, e.g. a collection of nodes that can share and reuse a buffer when stuff is appended?
BTW in this regard, I think we would benefit from using frugal collections, i.e. structs with optimized storage for 2-4 elements, instead of generic arrays/list, as many nodes commonly have very few children. That would cut down allocations some more... Lots of ideas but I don't want to touch the AST nodes until the factory is fully in place.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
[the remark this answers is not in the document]
jods4: Maybe, I haven't got my head too deep into that yet.

Gotta say one challenge when trying to improve the code is that you hit some design quirks and then it's a rabbit hole to work-around or fix them.

Other instances where I had to hack around:
- I've changed `Invert` to return a `ISqlPredicate` instead of `IQueryElement`... turned out every caller had a cast already! 😱 .
- I've changed `NotExpr` to take a `ISqlPredicate` instead of `ISqlExpression`. Makes sense and was already the case *almost* everywhere.

jods4:
### About SqlBuilders
SqlBuilder is the last step of the pipeline, it takes an optimized tree and prints a SQL string.

I've noticed some builders create a bunch of AST nodes, the main reason is to print those nodes as an alternative for "standard" nodes in the tree. These kind of transformations are often done by SqlOptimizer but not always...

I'm not gonna introduce the AST factory inside SqlBuilder classes. This is a last step past every optimizations and the provider-specific builders know exactly what they want to print, so let them be.

It is quite wasteful to create those temporary nodes. It'd be good to remove them with some more direct calls to SqlBuilder helpers that can print expressions without allocating new nodes. 

A good example is the `BuildFunction(string name, ISqlExpression[] parameters)` that is currently private but could be made protected. It's much more efficient to call this directly rather than to print a `new SqlFunction(name, parameters)`.

Source/LinqToDB/SqlQuery/AstFactory.cs
viceroypenguin: You should do some research on red-green trees from Roslyn: https://ericlippert.com/2012/06/08/red-green-trees/. It may help you find a better balance between immutability and memory pressure. I think the gist of it is similar to option 2, where you have internal mutable trees and public immutable trees once cloned.

viceroypenguin:
> ### About SqlBuilders
> SqlBuilder is the last step of the pipeline, it takes an optimized tree and prints a SQL string.
> ...

I'll agree with this comment. I think ultimately there should be three distinct phases: building (when initial LINQ calls are made), optimizing (SqlOptimizer and co.), and rendering (SqlBuilder and co.).  Currently, it sounds like optimizing and rendering are blended, with some optimization happening during the rendering phase and some pre-rendering happening during the optimization phase. It would definitely be a bit cleaner if the optimization was independent of rendering, with maybe some flags available from the data provider that affect certain features during the optimization phase.

Source/LinqToDB/SqlQuery/AstFactory.cs
jods4: Maybe I change my mind every day but seeing some stuff that's done with the AST today, I'm having second-thoughts about immutability. I'm still unsure what the best design will be, but in all cases it'll be easier to tweak AST representation with the factory in place.

RE: red-green trees I don't think they would help with building/mutating the AST, as the green part is immutable and basically matches what our AST is and does today. 
The design might be interesting if we want to maximize the sharing of immutable nodes (the green part), maybe even across queries, while allowing a mutable layer on top, e.g. for flags and other annotations (would be the red part).

My current thoughts: 
- The builder approach is most promising if we want to optimize allocations to the max during `Expression` conversion.
- Another design that would work well during optimizations too: add a flag indicating that the node is shared (hence immutable -- but most nodes would not be) and mutating methods. Those methods would mutate and return the same node when they can, and return a fresh modified copy when they can't.

The second design pushed to the max would seriously reduce allocations during full tree optimizations by `SqlOptimizer`. Currently, when something is modified, every parent node of the tree is replaced by a new one -> in many cases that may  not be required if you just swap child references.

sdanyliv:
@viceroypenguin, I've did that separation, and I've got performance degradation ;) So, SqlBuilder also optimises SQL on the fly. It is critical for queries which depends on parameter values.

viceroypenguin:
Converted to draft until updated against mainline

igor-tkachev:
I don't see any advantage in this approach.

We tried something similar in bltoolkit, in the very first versions. You can still find parts of this approach in the code. Nothing interesting came out of it.

In your case, it's even worse. You're unnecessarily coupling a bunch of code to a new object, which by the way has no state, but you're instantiating it.

Maybe if you make it static, it will look better?

What would be really interesting is a refactoring of SQL AST so that it can be conveniently used with pattern matching.

jods4:
@igor-tkachev 
> Maybe if you make it static, it will look better?

It sure would but there's a big trade-off. If it's static it can't be provider-specific.

igor-tkachev:
@jods4 

> It sure would but there's a big trade-off. If it's static it can't be provider-specific.

We have the `SqlBuilder` class to encapsulate provider-specific functionality.. We don't need to smear this all over the code. Also I do not see any provider-specific code in `AstFactory`, so it can be easily converted into static. In this case it can be useful.

As it is, it's more problems than it's worth.

PR #3785, opened 2022-09-22
Added custom test initialization logic.

nsarris:
Just another custom provider test helper, it's a hook on test setup.


jods4:
How is this meant to be used, given it's a static method?
Is it just to put some code in there when running tests locally -- and never meant to be committed?

nsarris:
> How is this meant to be used, given it's a static method? Is it just to put some code in there when running tests locally -- and never meant to be committed?

Correct, this how the whole customization infrstructure was setup. This is to facilitate third party providers, because you have to copy all tests over.

MaceWindu (approved):

PR #3812, opened 2022-10-13
Binary class equality operators fix

### LINKED ISSUE #3811: Binary class equality operators not working correctly
When comparing System.Data.Linq.Binary with operators, null is equal to any value because of typo in null checks:
```c#
if (binary1 is null && binary1 is null)
```

Can be reproduced simply enough, no DB needed:
```c#
using System;
using System.Data.Linq;

Binary a = null;
Binary b = new Binary(new byte[] { 1, 2, 3 });
if (a == b)
    Console.WriteLine("problems with ==");
if (!(a != b))
    Console.WriteLine("problems with !=");
```

Reproduced on:
LINQ to DB 4.3.0
.NET 6

arthur-suleimanov:
fix #3811

### DIFF
--- a/Source/LinqToDB/Compatibility/System/Data/Linq/Binary.cs
+++ b/Source/LinqToDB/Compatibility/System/Data/Linq/Binary.cs
@@ -43,7 +43,7 @@ namespace System.Data.Linq
 
 		public static bool operator ==(Binary? binary1, Binary? binary2)
 		{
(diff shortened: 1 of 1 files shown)


jods4 (approved):

sdanyliv (approved):

MaceWindu (approved):

MaceWindu:
thanks!

PR #3826, opened 2022-11-11
Improve log when db type can not be determined

timbze:
Helps to figure out which column is the issue when creating table with type that can not be determined.

Log was:
> Unhandled exception. LinqToDB.LinqToDBException: Database type cannot be determined automatically and must be specified explicitly

Now (diff in bold):
> Unhandled exception. LinqToDB.LinqToDBException: Database **column** type cannot be determined automatically and must be specified explicitly **for system type NodaTime.Instant**


MaceWindu (approved):

MaceWindu:
Thanks!

Issue #3840, opened 2022-11-22
Query Produces Exception in SQLServer - Property ist not defined


jogibear9988:
I try the following Code in SQLServer
(not exctly this code, but smth similar)

	public class Test
	{
		public virtual DateTime? StartDateTime { get; set; }

		public virtual TimeSpan? PreNotification { get; set; }
	}

	[TestFixture]
	public class Issue3999Tests : TestBase
	{
		[Test]
		public void Test3999([IncludeDataSources(TestProvName.AllSqlServer)] string configuration)
		{
			var ms = new MappingSchema();
			var mb = ms.GetFluentMappingBuilder();
			mb.Entity<Test>()
			   .HasTableName("Common_Topology_Locations")
			   .Property(e => e.StartDateTime)
			   .Property(e => e.PreNotification);

			using (var db = GetDataContext(configuration, ms))
			{
				using (db.CreateLocalTable<Test>())
				{
					var qry = from t in db.GetTable<Test>()
							  select new
							  {
								  StartDateTime = t.StartDateTime,
								  PreNotification = t.PreNotification,
								  NotificationDateTime = Sql.DateAdd(Sql.DateParts.Millisecond, -1 * t.PreNotification!.Value.Milliseconds, t.StartDateTime)
							  };
					var lst = qry.ToList();
					var sql = ((DataConnection)db).LastQuery;
				}				
			}
		}
	}

jogibear9988:
![MicrosoftTeams-image](https://user-images.githubusercontent.com/364896/203297911-45abf2db-6b4b-4221-bf35-61bcff0267bc.png)

MaceWindu:
Test pass for me

MaceWindu:
Actually reproduced with a bit different setup

MaceWindu:
looks like another case of same .net 7 bug

jogibear9988:
No, we had this issue in our Net6 application. I'll look when my tests work again.

MaceWindu:
7.0.1 released

jogibear9988:
@MaceWindu I updated the test. With Values in database it fails

MaceWindu:
![image](https://user-images.githubusercontent.com/1099238/214533321-b5e12f72-8264-43fb-9cd6-1209b0bed41a.png)

PR #3848, opened 2022-11-28
test for #3840

jogibear9988:
#3840


Source/LinqToDB/Mapping/FluentMappingBuilder.cs
```diff
@@ -22,6 +21,14 @@ public class FluentMappingBuilder
 
 		#region Init
 
+		/// <summary>
+		/// Creates new MappingSchema and fluent mapping builder for it.
+		/// </summary>
+		public FluentMappingBuilder()
```
MaceWindu, line 27: Just note. FluentMappingBuilder has very unintuitive behavior when we pass mapping schema to it - it use attribute from passed schema to derive and override mapping properties from it (implemented explicitly for some attributes only like ColumnAttribute). 
We should add this information to constructors summary to make difference clear to users

igor-tkachev: As we removed `mappingSchema.GetFluentMappingBulder()` method, actually, this is just a short version of 
```
var ms = new MappingSchema();
var fm = new FluentMappingBulder(ms);
```

igor-tkachev: On the other hand we can use 
`new FluentMappingBulder(new())`

MaceWindu (approved):

sdanyliv (approved):

Source/LinqToDB/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -182,8 +182,9 @@ public Expression BuildExpression(IBuildContext context, Expression expression,
 									var prevCount  = ctx.SelectQuery.Select.Columns.Count;
 									var expression = ctx.BuildExpression(ma, 0, context.enforceServerSide);
 
-									if (expression.NodeType == ExpressionType.Extension && expression is DefaultValueExpression
-																						&& ma.Expression?.NodeType == ExpressionType.Parameter)
+									if (expression.NodeType == ExpressionType.Extension &&
+									    expression is DefaultValueExpression &&
+									    ma.Expression?.NodeType == ExpressionType.Parameter)
```
viceroypenguin, line 187:
```suggestion
									if (expression is DefaultValueExpression { NodeType = ExpressionType.Extension } &&
									    ma.Expression?.NodeType == ExpressionType.Parameter)
```

Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs
```diff
@@ -197,7 +197,7 @@ void SetLoadWithBindings(Type objectType, ParameterExpression parentObject, List
 
 							Builder.AssociationPath.Push(Tuple.Create(new AccessorMember(ma), (IBuildContext)this, (List<LoadWithInfo[]>?)loadWith));
 
-							ex = BuildExpression(ma, 1, parentObject);
+							ex = BuildExpression(ma, 1, parentObject, true);
```
viceroypenguin, line 200:
```suggestion
							ex = BuildExpression(ma, 1, parentObject, enforceServerSide: true);
```

viceroypenguin (approved):
PR Comment should be changed to "Fixes #3840", so that it will close the associated issue.

Issue #3847, opened 2022-11-28
Dictionary not includeable in Postgresql in linq2db


jogibear9988:
See the Test in the Pull request, will not work in linq2db with postgres

MaceWindu:
What exactly doesn't work? I've run test locally and it passed for all enabled databases

jogibear9988:
Mhm, I created invalid SQL in our Database. Of cause I simplified the Test case. So maybe it does not reproduce. I'll look, my problem atm is I cold not run the linq2db tests any more. I'll wait for the Net7 Bugfix release, then I'll test again.

MaceWindu:
status?

jogibear9988:
I'll look tomorrow

MaceWindu:
Closing for now as existing tests work and no other tests provided.

PR #3849, opened 2022-11-28
test for #3847

jogibear9988:
#3847


MaceWindu:
@jogibear9988 , I've updated test to use data, but still don't see any issue

jogibear9988:
Test seems to work, so issue seems to be gone or never been there.

@MaceWindu 
Could we merge this Test so we have it tested?

PR #3530, opened 2022-04-28
Refactoring options.

sdanyliv:
Fix #471
Fix #472

Work still in progress.

- [x] Copy static configuration to options
- [ ] Add `UseMiniProfiler`
- [ ] Add `UseNodaTime`
- [ ] Cleanup code
- [ ] Fix documentation

PR #3852, opened 2022-12-01
Fix C# init-only support for older TFMs

MaceWindu:
Critical issue discovered during https://github.com/linq2db/linq2db/pull/3530 testing.

Workaround to https://github.com/dotnet/runtime/issues/79082

### DIFF
--- a/Source/Shared/IsExternalInit.cs
+++ b/Source/Shared/IsExternalInit.cs
@@ -1,4 +1,6 @@
-﻿#if !NET5_0_OR_GREATER
+﻿// don't use conditional compilation
+// see https://github.com/dotnet/runtime/issues/79082
+
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
@@ -17,4 +19,3 @@ namespace System.Runtime.CompilerServices
 	{
 	}
 }
-#endif


viceroypenguin:
How are we hitting this particular issue? Seems like an overly complicated setup in the first place. As long as we provide both pre/post `IsExternalInit` TFMs of l2db, that should be the extent of our requirement to handle.

MaceWindu:
linq2db.aspnet is ns2.0-only

viceroypenguin:
Seems like the proper fix would be to add additional TFMs to l2db.aspnet between ns2.0 and net7.0. Regardless, LGTM

viceroypenguin (approved):

MaceWindu:
it will not solve situations when user has mixed build setup

Shane32:
By the way, we stopped using init-only properties (or provided workarounds in the form of constructors) over at GraphQL.NET when we discovered it currently requires a newer C# language version than would be default for .NET Framework applications. There’s no workaround either, besides bumping the language version in the consuming application, which is not always an acceptable option for large companies.

viceroypenguin:
Not that this is a recommendation for l2db, because I think l2db should have broader support, but I've simply dropped support for below .net31 on most of my libraries.

Shane32:
> Not that this is a recommendation for l2db, because I think l2db should have broader support, but I've simply dropped support for below .net31 on most of my libraries.

Ditto for Core, although .NET Framework 4.6.1 and newer are still supported by MS. In addition, just FYI, the ASP.Net Core 2.1 libraries are also officially supported by MS when used with .NET Framework, as newer versions of .NET Core do not work with .NET Framework. I recently added support to GraphQL.NET Server for those frameworks (and use them).

Shane32:
> @Shane32 that's actually good point to be careful with new features. In this case we are safe - even with old targets users still can create records as they are just classes

👍

Shane32:
For reference:
- https://github.com/graphql-dotnet/graphql-dotnet/issues/3322
- https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-framework
- https://dotnet.microsoft.com/en-us/platform/support/policy/aspnetcore-2.1
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/init#metadata-encoding

viceroypenguin:
> > Not that this is a recommendation for l2db, because I think l2db should have broader support, but I've simply dropped support for below .net31 on most of my libraries.
> 
> Ditto for Core, although .NET Framework 4.6.1 and newer are still supported by MS. In addition, just FYI, the ASP.Net Core 2.1 libraries are also officially supported by MS when used with .NET Framework, as newer versions of .NET Core do not work with .NET Framework. I recently added support to GraphQL.NET Server for those frameworks (and use them).

Actually, due to an oddity in how netfx 3.5 was handled, it's technically under msft support until 2028. Go figure. Definitely not going to offer support that far back though.

PR #3855, opened 2022-12-03
Implement IAsyncDisposable for DataContextTransaction

Shane32:
This adds an `IAsyncDisposable` implementation for `DataContextTransaction`s.

Note: `DataConnectionTransaction` already implements `IAsyncDisposable`

### DIFF
--- a/Source/LinqToDB/DataContextTransaction.cs
+++ b/Source/LinqToDB/DataContextTransaction.cs
@@ -11,6 +11,11 @@ namespace LinqToDB
 	/// </summary>
 	[PublicAPI]
 	public class DataContextTransaction : IDisposable
+#if NATIVE_ASYNC
+		, IAsyncDisposable
+#else
+		, Async.IAsyncDisposable
+#endif
 	{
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/DataContextTransaction.cs
```diff
@@ -11,6 +11,11 @@ namespace LinqToDB
 	/// </summary>
 	[PublicAPI]
 	public class DataContextTransaction : IDisposable
+#if NATIVE_ASYNC
+		, IAsyncDisposable
```
viceroypenguin, line 15: Any reason we can't add `using Async;` to the top of the file and avoid the ifdef here?

Shane32: The code here matches elsewhere in the library.  I can post another PR to change both places.

viceroypenguin: Sounds good. Thanks!

viceroypenguin (approved):

PR #3892, opened 2022-12-24
Add optimistic lock extensions

### LINKED ISSUE #553: Concurrency checking

MaceWindu:
Fix #553

based on code, provided by @sdanyliv.

Provides:
- `UpdateOptimistic[Async]` extensions
- `DeleteOptimistic[Async]` extensions
- `OptimisticLockPropertyAttribute` mapping attribute with 3 built-in value generation strategies
-  `OptimisticLockPropertyBaseAttribute` mapping attribute for custom value generation strategy, not covered by `OptimisticLockPropertyAttribute`. User should  inherit from `OptimisticLockPropertyBaseAttribute` and override `GetNextValue` method

PR driven by `linq2db.Identity` migration to main repo. We have optimistic lock helpers in old repo and I want to improve them and expose to all users


Tests/Linq/Linq/ConcurrencyTests.cs
```diff
@@ -0,0 +1,609 @@
...
+				: base(default)
+			{
+			}
+
+			public override LambdaExpression GetNextValue(ColumnDescriptor column, ParameterExpression record)
+			{
+				return Expression.Lambda(
+					Expression.Constant(Guid.NewGuid().ToString("N")),
+					record);
+			}
```
MaceWindu, line 40: open question: is it good enough interface or we should introduce something more user-friendly?

MaceWindu:
Will rework it a bit. Remove incorrect timestamp-based key generation and add guid-based one

jods4:
It is a really nice addition for sure! Some feedback:

1. I wish we had simpler (public) APIs to create these kinds of "higher-level" concerns, although it's much harder to design such an API and I have nothing precise to propose right now.
The point is that this is not the only concern of this style we could want. Other examples: soft delete (replace delete with an update on a specific column + auto. filter `delete = 0` on every source); multi-tenancy (auto. filter `tenantId = xxx` on every source); light auditing (auto insert/update columns such as `createdBy`/`modifiedBy`).
These are examples amongst many and it would be crazy good if we had a "framework" to build all of these and more easily (also by end-users).

2. For the internal design, it might be better to operate as sql AST transformations rather than C# Expressions. 
Partly because Expressions are more complex to manipulate but mostly because it's a lot more performant. It creates less allocation by skipping all C# Expressions; + it accelerates the costly Expression -> sql AST transform as it keeps the Expression smaller.
We still need to include the required information somewhere in the source expression... maybe we could use a single custom Expression such as `Concurrent [ tableExpression ]` to mark in Expression tree the things that then need to be converted to SQL AST later.

3. Naming: _Concurrent_ is generally understood as the "problem" and it has several strategies to address it: _Last Write Wins_, _Optimistic_, _Pessimistic_. What is implemented here is the optimistic strategy so it might be more descriptive to call it `DeleteOptimistic` and `UpdateOptimistic`?

4. There are 2 variations of optimistic concurrency: one based on a "version" field as implemented here, another that simply compares all (with possible exclusions) fields. This latter strategy misses ABA cases but can be useful when working with legacy schema that you may not change. It'd be nice if this extension supported them -- or at least had a design how such support could be added later if someone asks for it.

5. The extension methods provide no check on the returned value? 
I don't see why someone would do a concurrent update and not check whether it was successful or not... 
By far the most common use case will be to check the number of rows updated and throw if 0.
This should be built-in so that users don't forget to add it with every call.

6. API is a bit restrictive: you can customize nothing in `DeleteConcurrent`. 
At first glance it may look ok, but for more complex use cases you might want to add extra filters, for example to enforce security (access rights), or multi-tenancy.
Yes, the error would be misleading (concurrency exception yet the cause is security) but it's only something that happens if you tamper with the application, it couldn't happen normally. It's the same as passing a non-existing id (concurrency vs not found).
Maybe something like this should be possible?
```csharp
public void DeleteFolder(Folder folder)
{
  db.Folders
    .Where(CanAccess)
    .DeleteOptimistic(folder); 
}
```

MaceWindu:
@jods4 

> 1. I wish we had simpler (public) APIs to create these kinds of "higher-level" concerns...
> 2. For the internal design, it might be better to operate as sql AST transformations rather than C# Expressions...

Agree, but that's not something I planned for now. Maybe @sdanyliv have some design ideas. Anyway, we can update it in v6.

> 3. it might be more descriptive to call it DeleteOptimistic and UpdateOptimistic?

Done

> 4. There are 2 variations of optimistic concurrency: one based on a "version" field as implemented here, another that simply compares all (with possible exclusions) fields. 

No plans to implement it for now without user request

> 5. The extension methods provide no check on the returned value?

No strong opinion here. I would expect people using optimistic lock smart enough to check operation result.

Some notes from me:
- rownumber not supported by ClickHouse. Not our problem, but it makes any checks on our side useless/wrong if user decide to use it with CH
- not a big fan of exception

> 6. API is a bit restrictive: you can customize nothing in DeleteConcurrent.

added both update and delete over `IQueryable`

Source/LinqToDB/Mapping/VersionBehavior.cs
```diff
@@ -0,0 +1,36 @@
...
+		/// <summary>
+		/// Uses <see cref="Guid.NewGuid()"/> value.
+		/// Supported column types:
+		/// <list type="bullet">
+		/// <item><see cref="System.Guid"/></item>
+		/// <item><see cref="string"/> using <see cref="Guid.ToString()"/></item>
+		/// <item><see cref="T:byte[]"/> using <see cref="Guid.ToByteArray()"/></item>
+		/// </list>
+		/// </summary>
+		Guid
```
viceroypenguin, line 34: prevent future changed line when adding more fields.
```suggestion
		Guid,
```

viceroypenguin (dismissed):

sdanyliv (approved):

PR #3900, opened 2023-01-03
Refactor work with attributes

MaceWindu:
Main goal of this PR:
- separate mapping attributes and non-mapping attributes. Mapping attributes are attributes resolved through mapping schema, non-mapping attributes resolved through reflection
- optimize work with attributes to better use caching. Currently caching support is very limited
- cleanup and improve related API surface to remove not needed/not used functionality and fix usability issues with API

Changes:
- moved `string? Configuration` property to base `MappingAttribute` class. Removes need in internal `IConfigurationProvider` interface and `configGetter` delegates in `MappingSchema` methods. Also it adds `Configuration` support to mapping attributes that were missing it before
- enforce `MappingAttribute` generic constrain on `IMetadataReader` and `MappingSchema` methods
- moved existing attribute cache from `ReflectionExtensions` to separate `AttributeExtensions` class and ban runtime methods to read attributes. Prevously a lot of places were reading attributes directly ignoring cache
- implement similar cache for `MappingAttribute` and use it in: `AttributeReader`, `FluentMetadataReader`
- Fluent mappings now require call to `Build()` method to add all configured mappings to mapping schema
- renamed `Configuration.Linq.EnableAutoFluentMapping` setting to `Configuration.Linq.EnableContextSchemaEdit` to better reflect option behavior and change defaults from `true` to `false`
- update `MappingSchema` logic to not reset schema id when additional mapping information generated based only on information, already registered in schema (enum default/nullability information, conversion expressions generation for registered conversions) as it will trigger exceptions from locked mapping schemas
- updated connection factory delegate to get `DataOptions` as parameter

Other internal optimizations:
- add `StringBuilder` pooling
- reintroduce combined `MappingSchema` cache
- removed error message generation from `EvaluationContext` as it used very-very rarely but introduce a lot of string allocations for everyone

PR intriduce several breaking changes to fluent mapping and metadata reader implementation behavior:
- removed registration of `SystemComponentModelDataAnnotationsSchemaAttributeReader` and `SystemDataLinqAttributeReader` metadata providers from default configuration. Few people use it but readers called for everyone
- fluent mappings now added to mapping schema explicitly when user calls `Build()` method on fluent mapping builder
- for same reason following code will not work anymore:
```cs
var ms = new MappingSchema();

// THIS LINE should be moved after Build call
// because now mappings added to schema only after Build() call
db.AddMappingSchema(ms);

var builder = ms.GetFluentMappingBuilder();

// configure fluent mappings
//  ...

builder.Build();
```

fixed code:
```cs
var ms = new MappingSchema();


var builder = ms.GetFluentMappingBuilder();

// configure fluent mappings
//  ...

builder.Build();

// add mapping schema to connection after mappings configured
db.AddMappingSchema(ms);
```

In any case this is not a good way to configure mappings and should be avoided- mappings should be configured once and shared `MappingSchema` instance should be passed to `DataConnection`/`DataContext` on creation

- `IMetadataReader.GetAttributes<T>` methods could be called with `MappingAttribute` as type parameter (base class for all mapping attributes) and method implementation should return all supported mapping attributes, applicable to specific type/member

Old code:
```cs
public T[] GetAttributes<T>(Type type)  where T : MappingAttribute
{
    if (typeof(T) == typeof(TableAttribute))
    {
        // create and return TableAttribute instance if it is defined for "type"
    }
}
```

New code:
```cs
public T[] GetAttributes<T>(Type type)  where T : MappingAttribute
{
    if (typeof(T).IsAssignableFrom(typeof(TableAttribute)))
    {
        // create and return TableAttribute instance if it is defined for "type"
    }

   // if reader supports more attribute types, they also should be returned if they are assignable to T
}
```


Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -612,20 +612,25 @@ await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
...
 		{
-			var members = typeAccessor.Members.Where(m => m.HasSetter).ToArray();
+			var indexMap = new Dictionary<int, MemberAccessor>();
 
 			// Use attribute labels if any exist.
-			var indexMap = (from m in members
-				let a = m.GetAttributes<ResultSetIndexAttribute>()
-				where a != null
-				select new { Member = m, a[0].Index }).ToDictionary(e => e.Index, e => e.Member);
+			foreach (var m in typeAccessor.Members.Where(m => m.HasSetter))
```
viceroypenguin, line 618: Consistency: Recommending using either LINQ or code filtering. That is, either the above query (or it's equivalent in fluent notation), or move the `if (m.HasSetter)` inside the foreach instead of `.Where()`.

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -612,20 +612,25 @@ await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
...
+				if (attr != null)
+					indexMap.Add(attr.Index, m);
+			}
 
 			if (indexMap.Count == 0)
 			{
 				// Use ordering of properties according to reflection.
-				for (var i = 0; i < members.Length; i++)
+				var i = 0;
+				foreach (var m in typeAccessor.Members.Where(m => m.HasSetter))
```
viceroypenguin, line 630: Same

Source/LinqToDB/DataProvider/SqlServer/SystemDataSqlServerAttributeReader.cs
```diff
@@ -17,188 +21,184 @@ namespace LinqToDB.Metadata
...
-								}
-
-								attrs = new[] { (T)(Attribute)new Sql.ExpressionAttribute(ex) { ServerSideOnly = true } };
+								var name = key.memberInfo.DeclaringType!.Name.ToLowerInvariant();
+								name = name.StartsWith("sql") ? name.Substring(3) : name;
+
+								ex = string.Format(
+									"{0}::{1}({2})",
+									name,
+									((dynamic)attr).Name ?? key.memberInfo.Name,
```
viceroypenguin, line 123: `dynamic` is notoriously slow due to dynamic callsite lookup. Perhaps the `propertyof` could be cached by type and used to get the value instead? Also, may not be called often enough to matter; if so, then decline this and related comments.

Source/LinqToDB/DataProvider/SqlServer/SystemDataSqlServerAttributeReader.cs
```diff
@@ -17,188 +21,184 @@ namespace LinqToDB.Metadata
...
+									string.Join(", ", ps.Select((_, i) => '{' + i.ToString() + '}')));
 							}
 							else
 							{
-								attrs = Array<T>.Empty;
+								ex = string.Format(
+									"{{0}}.{0}({1})",
+									((dynamic)attr).Name ?? key.memberInfo.Name,
```
viceroypenguin, line 130: Same

Source/LinqToDB/DataProvider/SqlServer/SystemDataSqlServerAttributeReader.cs
```diff
@@ -17,188 +21,184 @@ namespace LinqToDB.Metadata
...
-								else
-								{
-									attrs = Array<T>.Empty;
-								}
-							}
-							else
+							if (attr != null)
 							{
-								attrs = Array<T>.Empty;
+								var ex = $"{{0}}.{((dynamic)attr).Name ?? key.memberInfo.Name}";
```
viceroypenguin, line 148: Same

MaceWindu:
oracle baselines changes are valid. I've removed `Guid`-typed field ignore as we added support for this type some time ago

Source/LinqToDB/DataProvider/SqlServer/SystemDataSqlServerAttributeReader.cs
```diff
@@ -17,188 +21,184 @@ namespace LinqToDB.Metadata
...
-				if (attrs.Length == 1)
+				var res = (T[])_cache.GetOrAdd((memberInfo, _sqlUserDefinedTypeAttribute), static key =>
 				{
-					var c = attrs[0];
-					var n = ((dynamic)c).Name ?? memberInfo.GetMemberType().Name;
+					var c = FindAttribute(key.memberInfo.GetMemberType(), key.attributeType);
+
+					if (c != null)
+					{
+						var n = ((dynamic)c).Name ?? key.memberInfo.GetMemberType().Name;
```
viceroypenguin, line 167: Same

Source/LinqToDB/Expressions/MemberHelper.cs
```diff
@@ -51,6 +51,7 @@ public override int GetHashCode()
 		/// <exception cref="ArgumentException">Only simple, non-navigational, member names are supported in this context (e.g.: x =&gt; Sql.Property(x, \"SomeProperty\")).</exception>
 		public static MemberInfo GetMemberInfo(LambdaExpression func)
 		{
+
```
viceroypenguin, line 54: ??

Source/LinqToDB/Mapping/EntityDescriptor.cs
```diff
@@ -150,24 +151,15 @@ TableOptions IEntityChangeDescriptor.TableOptions
...
-				if (mappingAttrs.Length > 0)
-					return true;
-
-				currentType = currentType.BaseType;
-			}
-
-			return false;
+			return TypeAccessor.Type.BaseType != null
+				? MappingSchema.HasAttribute<InheritanceMappingAttribute>(TypeAccessor.Type.BaseType)
+				: false;
```
viceroypenguin, line 156:
```suggestion
			return TypeAccessor.Type.BaseType != null
				&& MappingSchema.HasAttribute<InheritanceMappingAttribute>(TypeAccessor.Type.BaseType);
```

Source/LinqToDB/Mapping/EntityMappingBuilder.cs
```diff
@@ -383,8 +324,7 @@ public EntityMappingBuilder<TEntity> HasIdentity<TProperty>(Expression<Func<TEnt
 				func,
 				false,
 				 _    => new IdentityAttribute(Configuration),
-				(_,a) => {},
-				a => a.Configuration);
+				(_,a) => {});
```
viceroypenguin, line 327: I believe you can do this now (maybe not?):
```suggestion
				(_,_) => {});
```

Source/LinqToDB/Mapping/ColumnDescriptor.cs
```diff
@@ -43,43 +43,38 @@ public ColumnDescriptor(MappingSchema mappingSchema, EntityDescriptor entityDesc
...
+			PrimaryKeyOrder   = columnAttribute?.PrimaryKeyOrder   ?? 0;
+			IsDiscriminator   = columnAttribute?.IsDiscriminator   ?? false;
+			SkipOnEntityFetch = columnAttribute?.SkipOnEntityFetch ?? false;
+			DataType          = columnAttribute != null ? columnAttribute.DataType : dataType.Type.DataType;
+			DbType            = columnAttribute != null ? columnAttribute.DbType   : dataType.Type.DbType;
+			CreateFormat      = columnAttribute?.CreateFormat;
+
+			if (columnAttribute == null || (columnAttribute.DataType == DataType.Undefined || columnAttribute.DataType == dataType.Type.DataType))
 			{
-				columnAttribute = new ColumnAttribute();
```
MaceWindu, line 53: changes here required to address removal of this unnecessary `ColumnAttribute` instance creation. Previously it was the reason why we didn't introduced all those optimizations earlier: below we modified attribute properties even if attribute came from mapping schema, which resulted in mapping schema corruption as now attributes cached.

Source/LinqToDB/Expressions/MemberHelper.cs
MaceWindu: ¯\_(ツ)_/¯

viceroypenguin (approved):

viceroypenguin (approved):

MaceWindu:
Without today's fixes (using memory use fixes from yesterday, new benchmark results will be available tomorrow):
- even less memory use than version from master
- `ConcurrentBenchmark` is 10 times faster (completely unexpected)

sdanyliv (dismissed):

MaceWindu:
also check release notes https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#release-500-preview-2

NuGet/README.T4.md
```diff
@@ -71,4 +71,4 @@ There are multiple reasons for it:
 - T4 host use x86 process for Visual Studio 2019- and x64 process for Visual Studio 2022+. This creates issues with native providers as they are platform-dependent (ODBC and OleDb providers, various native ADO.NET providers).
 - It's very hard to maintain big code generation framework, written in T4 templates.
 
-Those are just several most major issues current T4 templates have.
+Those are just several most major issues T4 templates have.
```
viceroypenguin, line 74:
```suggestion
These are the most significant issues with T4 templates.
```

Source/LinqToDB/Common/Internal/IdentifierBuilder.cs
```diff
@@ -18,14 +18,22 @@ namespace LinqToDB.Common.Internal
 	/// Internal infrastructure API.
 	/// Provides functionality for <see cref="IConfigurationID.ConfigurationID"/> generation.
 	/// </summary>
-	public sealed class IdentifierBuilder
+	public sealed class IdentifierBuilder : IDisposable
 	{
+		readonly ObjectPool<StringBuilder>.RentedElement _pooledElement;
+		StringBuilder? _stringBuilder;
```
viceroypenguin, line 24: make `StringBuilder` non-null and `readonly`, and use a `bool _disposed` flag to simplify the code.

viceroypenguin (commented):
If possible, I'd like to see the `Pools.StringBuilder` changes done as a separate PR; it is a separate (useful!) change from refactoring attribute work. It would be easier to see the changes specific to each (pooled SB vs attributes) if the PRs were separate.

Issue #3902, opened 2023-01-10
is parameter name too long to be cutted?


pomesqb:
the code:
```
D.ConnLinq.GetTable<TPPSLT>()
                    .Where(c => c.TPPSLT_TYPE == '4' && c.TPPSLT_KIND_ID == "AAA")
                    .Set(c => c.TPPSLT_QUOTE_DURATION_MULTI, 10)
                    .Set(c => c.TPPSLT_USER_ID, "IamHandsome")
                    .Update();
```

the class:
```
public class TPPSLT
    {
        [PrimaryKey]
        public virtual char TPPSLT_TYPE { get; set; }
        [PrimaryKey]
        public virtual string TPPSLT_KIND_ID { get; set; }
        public virtual decimal TPPSLT_QUOTE_DURATION_MULTI { get; set; } 
        public virtual string TPPSLT_USER_ID { get; set; }
    }
```
the linq2db auto generate output SQL, i find the last character "I" in parameter @TPPSLT_QUOTE_DURATION_MULTI  was been cutted.
and it becomes  @TPPSLT_QUOTE_DURATION_MULT, so there raise an error.
```
DECLARE @TPPSLT_QUOTE_DURATION_MULTI Decimal(2, 0)
SET     @TPPSLT_QUOTE_DURATION_MULTI = 10
DECLARE @TPPSLT_USER_ID UniVarChar(11) -- String
SET     @TPPSLT_USER_ID = 'IamHandsome'

UPDATE
	[TPPSLT]
SET
	[c_1].[TPPSLT_QUOTE_DURATION_MULTI] = @TPPSLT_QUOTE_DURATION_MULT,
	[c_1].[TPPSLT_USER_ID] = @TPPSLT_USER_ID
FROM
	[TPPSLT] [c_1]
WHERE
	[c_1].[TPPSLT_TYPE] = '4' AND [c_1].[TPPSLT_KIND_ID] = 'AAA'`
```

```
Exception message:There is no host variable corresponding to the one specified by the PARAM datastream. This means that this variable '@TPPSLT_QUOTE_DURATION_MULTI' was not used in the preceding DECLARE CURSOR or SQL command.
```

### Environment details

`Linq To DB` version: 4.4.0
DataProvider: AdoNetCore.AseClient 0.19.2
Database (with version): Adaptive Server Enterprise/16.0 SP03
Operating system: Window10
.NET Version: .Net Standard 2.0

MaceWindu:
#3878

[...]

PR #3908, opened 2023-01-11
Sybase parameter length issue

igor-tkachev:
#3902


igor-tkachev:
Quick fix. Should work and do not break anything. To make it more reliable, we need deep parameter assignment refactoring.

PR #3919, opened 2023-01-23
Add SqlBuilder pooling

MaceWindu:
Extracted into separate PR from #3900


jods4:
Given our extensive use of local StringBuilders, it makes me wonder if it would be even better to use a `ValueStringBuilder` instead, e.g. 
https://source.dot.net/#System.Private.CoreLib/src/libraries/Common/src/System/Text/ValueStringBuilder.cs,1cc7544769bf04a4

`System.Memory` nuget package enables us to use Spans back to .net 4.5, so maybe it's doable.

viceroypenguin (commented):
Subject to @jods4 comments already, LGTM

PR #3921, opened 2023-01-23
Association fixes

### LINKED ISSUE #3658: CanBeNull is not respected in association without keys

MaceWindu:
Fix #3658

Replaces #3657

Fixes:
- respect `CanBeNull` for association with custom predicate
- respect associations in select for `Count` scalar queries
- don't remove "unused" **INNER** join for `Count` scalar query with association field access

### DIFF
--- a/Source/LinqToDB/Linq/Builder/AssociationHelper.cs
+++ b/Source/LinqToDB/Linq/Builder/AssociationHelper.cs
@@ -119,8 +119,8 @@ namespace LinqToDB.Linq.Builder
 
 				if (expressionPredicate != null)
 				{
-					shouldAddDefaultIfEmpty = true;
-					shouldAddCacheCheck = true;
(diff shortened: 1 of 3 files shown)


MaceWindu:
original issues fixed. Let's see what it will break :smile:

sdanyliv (approved):

PR #3930, opened 2023-01-26
Break fluent mapping configuration explicitly

MaceWindu:
Otherwise users will not notice that they need to update their setup code...

Also it removes builder API from  data context. It was promoting context-level mapping configuration approach, which is not recommended due to performance concerns.

+ add locked mapping schema guard to TempTable APIs as it could use fluent mappings...


viceroypenguin (dismissed):

sdanyliv (approved):

MaceWindu:
I will try to redesign TempTable internals to not introduce breaking changes. Will be in RC or if it will not take long - in current preview

PR #3939, opened 2023-01-29
Add new connection setup options

### LINKED ISSUE #2137: Using ILinqToDBSettings with SqlCredential
**Current usage of `ILinqToDBSettings`**
I use `ILinqToDBSettings` as easy way to inject multiple connection strings (as described here https://github.com/linq2db/linq2db#net-core) from top-level project and switch them dynamically. This way I don't need to pass a particular connection string all way down per each query because all strings are stored internally by `linq2db`.

**Plain strings**
The only problem they are stored as plain strings, so can be viewed by `Process Hacker 2` for example. A solution might be to use `SqlCredential` class on instantiating `DataConnection`.

**Not passing manually**
However, I see no way to store `SqlCredential` objects in `ILinqToDBSettings` object, the same way it currently stores plain connection strings. It means I need to pass a `SqlCredential` per each query manually all way down. Given amount of different queries, passing the `SqlCredential` for each is so tedious and tough, really hope there is more elegant way.

**A failed attempt**
Since `ILinqToDBSettings.ConnectionStrings` is of type `IEnumerable<IConnectionStringSettings>` I made this property computed (so extracting plain connection strings from secured `SqlCredential` objects each call). Unfortunately, `ILinqToDBSettings.ConnectionStrings` is called just once, at time of the first query, then it is saved as cached plain strings. 
Is there a way to not save value of `ILinqToDBSettings.ConnectionStrings` internally and compute the getter value per each query instead? Hence, the connection strings will exist in memory as list of secured `SqlCredential` objects only, not plain strings, and will be converted to a plain string each query.

MaceWindu:
fix #2137 (provides easy option to setup connection)

PR adds two new connection options:
- `UseBeforeConnectionOpened`
- `UseAfterConnectionOpened`
to support connection configuration without need to define interceptor explicitly (implementation based on interceptors).

Distinctions from interceptors:
- it is easier to implement from user's POV as you don't need to implement new class for insterceptors
- user could specify only sync action for both sync and async code pathes if your callback doesn't have blocking code or define both sync and async actions
- option used also for provider detection logic (and clickhouse bulk copy) where we create connection. Explicit interceptors are not called from those places

Other changes:
- `ConnectionEventData.DataConnection` is not nullable as e.g. in provider detector we don't have `DataConnection` instance (as only new configuration actions will be called with null value, impact is not big to existing implementations)
- add `ConnectionOptions` parameter to data provider factory delegeate
- reactored provider detection logic to work with `ConnectionOptions`


MaceWindu:
@linq2db guys, need your opinion:
1. do we want it
2. should we add interceptors support to all places where we create connection (provider detection and so on)

MaceWindu:
@linq2db any feedback?

viceroypenguin:
I haven't generally had a need for interceptors, so my opinion isn't super strong. That said, I think functions on the `DataOptions` are better than an `IInterceptor` in general. Like you said, no need to write our own class to meet an interface spec, and quick lambdas are easier to write as well. I'm all for this change.

sdanyliv (approved):

PR #3942, opened 2023-01-29
Resubmit https://github.com/linq2db/linq2db/pull/3933

MaceWindu:
- introduce explicit breaking changes to fluent configuration API
- fix temp table custom mappings support


MaceWindu:
Reverted and submitted again as  test baselines doesn't look right

MaceWindu:
It would be better to remove fluent mapping support from temp table API in future, but at least it will not break context mapping schema...

PR #3946, opened 2023-01-31
Failing test case for #3945

### LINKED ISSUE #3945: Bad SQL generated for INSERT w/ CTE + JOIN
I'm gonna link a PR with a failing test case.

The issue is that when doing an INSERT from a query that contains a CTE + a JOIN, linq2db wraps the query in SELECT *.
Oracle doesn't like that SELECT * and fails with ORA-00918, Column ambiguously defined (which makes no sense to me, but whatever).
That SELECT * serves no purpose and can be removed, the query then executes successfully.

Generated SQL structure:
```sql
INSERT INTO ...
SELECT * FROM (  -- This confuses Oracle
  WITH cte (..)
  SELECT ..
  FROM cte ..
  JOIN parent ..
)
```

### Environment details
`Linq To DB` version: *4.3.0*
Database (with version): *Oracle 19*
ADO.NET Provider (with version): *ODP.NET Core*
.NET Version: *.NET 6.0*

jods4:
Fix #3945 

Summary:
- remove not needed SELECT * wrapper around SELECT with CTE for insert queries
- fix DB2 CTE generation location
- added basic heuristics to detect that CTE created by `GetCte` API is not recursive
- "fixed" unnecessary column generation hack (whole hack will be gone in v6)

### DIFF
--- a/Source/LinqToDB/DataProvider/DB2/DB2SqlBuilderBase.cs
+++ b/Source/LinqToDB/DataProvider/DB2/DB2SqlBuilderBase.cs
@@ -15,6 +15,8 @@ namespace LinqToDB.DataProvider.DB2
 
 	abstract partial class DB2SqlBuilderBase : BasicSqlBuilder<DB2Options>
 	{
+		public override bool CteFirst => false;
+
 		protected DB2SqlBuilderBase(IDataProvider? provider, MappingSchema mappingSchema, DataOptions dataOptions, ISqlOptimizer sqlOptimizer, SqlProviderFlags sqlProviderFlags)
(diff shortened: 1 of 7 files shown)


MaceWindu:
Test results:
- all tested DBs generate 4 columns 
- firebird and oracle providers also generate `SELECT *` wrapper

sdanyliv (dismissed):

viceroypenguin (dismissed):
Cleaning up the common `BuildInsertQuery` makes a lot of sense. LGTM

MaceWindu:
Arghh, one test failed...

PR #3950, opened 2023-02-03
Add [Source] to Merge Output

### LINKED ISSUE #3643: [SQL Server] MergeWithOutputAsync possible to add Output for [Source]?
Currently in MergeWithOutputAsync it is only possible to output:
- action: "inserted" or "updated"
- result [INSERTED]
- result [DELEETED]

In the SQL Server it is also possible to output [Source], so I am wondering would it be possible to add something like this?
image

So it could look something like this. Required change is only for `MergeWithOutputAsync`.
```c#
await using var tempTable = await toughComDatabase.CreateTempTableAsync(input, cancellationToken: ct)
	.ConfigureAwait(false);

connection.Source
	.Merge()
	.Using(tempTable)
	.On(static (x, y) => x.Name == y.Name)
	.InsertWhenNotMatched(static x => new Record
	{
		Name = x.Name,
		Prop1 = x.Prop1,
		Prop2 = x.Prop2
	})
	.UpdateWhenMatched(static (_, y) => new Record
	{
		Prop1 = y.Prop1,
		Prop2 = y.Prop2
	})
	.MergeWithOutputAsync((action, deleted, inserted, source) =>
	{
		inserted.PK,
		source.GroupId
	});
```
And the generated SQL might look like this
```sql
MERGE INTO [tblTarget] [Target]
USING (
	SELECT
		[t1].[GroupId],
		[t1].[Name],
		[t1].[Prop1],
		[t1].[Prop2]
	FROM
		[tempdb]..[#tmpSource] [t1]
) [Source]
(
	[GroupId],
	[Name],
	[Prop1],
	[Prop2]
)
ON ([Target].[Name] = [Source].[Name])

WHEN NOT MATCHED THEN
INSERT
(
	[Name],
	[Prop1],
	[Prop2]
)
VALUES
(
	[Source].[Name],
	[Source].[Prop1],
	[Source].[Prop2]
)

WHEN MATCHED THEN
UPDATE
SET
	[Target].[Prop1] = [Source].[Prop1],
	[Target].[Prop2Keys] = [Source].[Prop2]
OUTPUT
	[INSERTED].[IdentityPK],
	[Source].[GroupId]
;
```

If it is possible to consider this feature, let me know if some details are needed.

### Environment details
linq2db version: 4.1.0
Database Server: SQL Server
Database Provider: Microsoft.Data.SqlClient
Operating system: Windows
.NET Framework: .NET 6

igor-tkachev:
Fix #3643


viceroypenguin (approved):

Source/LinqToDB/Linq/Builder/MergeBuilder.cs
```diff
@@ -57,11 +64,26 @@ protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, Meth
...
+				var actionFieldContext   = new SingleExpressionContext(null, builder, actionField, selectQuery);
+				var deletedTableContext  = new TableBuilder.TableContext(builder, selectQuery, deletedTable);
+				var insertedTableContext = new TableBuilder.TableContext(builder, selectQuery, insertedTable);
+
+				TableBuilder.TableContext? sourceTableContext = null;
 
-				if (kind == MergeKind.MergeWithOutput)
+				if (kind is MergeKind.MergeWithOutputSource or MergeKind.MergeWithOutputIntoSource)
+				{
+					var sourceTable = new SqlTable(args[1])
```
MaceWindu, line 75: it doesn't look right. We should use already existing source context instead of fake table. `Source` is not a special table, but alias for merge statement source:
- while we use hardcoded `Source` name for this alias, it could be any name
- more important, if source is a query, we generate columns in source based on used fields in different parts of merge statement and if user will reference unused anywhere else column in output - it will generate incorrect sql

MaceWindu (changes requested):
I've added failing test

Source/LinqToDB/Linq/Builder/MergeBuilder.cs
igor-tkachev: I tried. It was my first attempt. This solution does not change any existing logic. Other way requires some questionable modifications. I can do it again, so you will be able to compare them.

MaceWindu (approved):

PR #3955, opened 2023-02-06
Downgrade abstractions packages

### LINKED ISSUE #3953: [V5]Lower dependency versions for microsoft abstraction libs
Linq2db uses DependencyInjection and Logging abstractions from the MS extensions namespace. It is generally not the best practive to aggresively upgrade those packages in library projects as they are pretty much backwards compatible back to v1 and might create conflicts to consumer projects. Linq2db only uses IServiceCollection and ILoggerFactory / ILogger , you can stay to a very early version of abstractions for both or at least stick to v6.

MaceWindu:
Fix #3953


viceroypenguin (dismissed):

MaceWindu:
https://bugs.mysql.com/bug.php?id=109670

image

sdanyliv (approved):

PR #3960, opened 2023-02-09
preserve member ordering for derived types

### LINKED ISSUE #3959: Columns in SELECT statements are reversed when entity class inherits from a base class
I'm using the 5.0.0-preview2.

Just noticed something weird, not really a bug but it is annoying when analyzing SQL queries...

When the entity type inherits from another type, the columns are reversed in the SELECTs, here's what I mean:

```c#
    [Table(Name = "BsContact", IsColumnAttributeRequired = false)]
    public class Contact
    {
        [PrimaryKey]
        public int ContactId { get; set; }
        public string ContactCode { get; set; }
        public int? Age { get; set; }
        public DateTime? DateBirth { get; set; }
    }
```
generates this:
```SQL
SELECT
        [t1].[ContactId],
        [t1].[ContactCode],
        [t1].[Age],
        [t1].[DateBirth]
FROM
        [BsContact] [t1]
```

but this mapping:

```c#
    public class BaseType
    {

    }

    [Table(Name = "BsContact", IsColumnAttributeRequired = false)]
    public class Contact : BaseType
    {
        [PrimaryKey]
        public int ContactId { get; set; }
        public string ContactCode { get; set; }
        public int? Age { get; set; }
        public DateTime? DateBirth { get; set; }
    }
```
generates this:
```SQL
SELECT
        [t1].[DateBirth],
        [t1].[Age],
        [t1].[ContactCode],
        [t1].[ContactId]
FROM
        [BsContact] [t1]
```

Not really sure why it works with no base type, but the reverse order seems to originate from this method called by `TypeAccessor<T>`, it returns members in reverse order because it uses `AddFirst` and not `AddLast`. Is there any reason for that?

https://github.com/linq2db/linq2db/blob/0b8f1e08c173f12ab29d8d413e11f3bba3d4a476/Source/LinqToDB/Extensions/ReflectionExtensions.cs#L46

Thanks in advance

guillaume86:
fix #3959

### DIFF
--- a/Source/LinqToDB/Extensions/ReflectionExtensions.cs
+++ b/Source/LinqToDB/Extensions/ReflectionExtensions.cs
@@ -29,25 +29,35 @@ namespace LinqToDB.Extensions
 		public static MemberInfo[] GetPublicInstanceValueMembers(this Type type)
 		{
 			var members = type.GetMembers(BindingFlags.Instance | BindingFlags.Public)
-				.Where(m => m.IsFieldEx() || m.IsPropertyEx() && ((PropertyInfo)m).GetIndexParameters().Length == 0);
+				.Where(m => m.IsFieldEx() || m.IsPropertyEx() && ((PropertyInfo)m).GetIndexParameters().Length == 0)
+				.ToArray();
 
 			var baseType = type.BaseType;
 			if (baseType == null || baseType == typeof(object) || baseType == typeof(ValueType))
-				return members.ToArray();
+				return members;
+			
+			// in the case of inheritance, we want to:
+			//  - list base class members first
+			//  - remove shadowed members (new modifier)
+			//	- preserve the order of GetMembers() inside the same type declared type
 
-			var results = new LinkedList<MemberInfo>();
-			var names   = new HashSet<string>();
+			var results = new List<MemberInfo>(members.Length);
+			var seen = new HashSet<string>();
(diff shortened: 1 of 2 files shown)


jods4 (approved):

MaceWindu:
Something wrong with FluentMappingTests.Issue291Test1Attr and FluentMappingTests.Issue291Test2Attr tests

itemType should be before my_other_col if I see it correctly

MaceWindu:
otherwise baselines look good

guillaume86:
> Something wrong with `FluentMappingTests`: `Issue291Test1Attr` and `Issue291Test2Attr` tests
> 
> `itemType` should be before `my_other_col` if I see it correctly

I don't really get what's going on, it's impacting ClickHouse for TFMs < net7.0, but not Sqlite somehow?

guillaume86:
It must be something in the way the FluentMappingBuilder interacts with the other metadata, not sure what but maybe a data structure with unspecified ordering (ConcurrentDict / MemoryCache) changed its internals in net7.0.

I can stabilize the tests by declaring the discriminator in the fluent mapping instead of an Attribute. Do you want me to make the change?

MaceWindu:
if there will be no drift in baselines - let's just ignore it

MaceWindu (approved):

viceroypenguin (approved):

PR #3962, opened 2023-02-10
Added missing CreateTempTable overrides

igor-tkachev:
CreateTempTable overrides to support anonymous types Fluent Mapping.

```c#
using var tmp = db.CreateTempTable(
    new[] { new { Name = "John" } },
    m => m
        .HasTableName("TempTable")
        .Property(p => p.Name)
            .HasLength(20)
            .IsNotNull());
```


sdanyliv:
Do we need another overload? I think better to add another optional parameter and mention breaking change in release notes.

Source/LinqToDB/TempTable.cs
```diff
@@ -338,6 +364,32 @@ protected TempTable(ITable<T> table, EntityDescriptor? tableDescriptor)
...
+			IEnumerable<T>    items,
+			EntityDescriptor  tableDescriptor,
+			BulkCopyOptions?  options           = default,
+			string?           tableName         = default,
+			string?           databaseName      = default,
+			string?           schemaName        = default,
+			string?           serverName        = default,
+			TableOptions      tableOptions      = default,
+			CancellationToken cancellationToken = default)
+		{
```
MaceWindu, line 389: I think overloads with `EntityDescriptor` parameter should be internal

Source/LinqToDB/TempTable.cs
```diff
@@ -81,6 +82,31 @@ public class TempTable<T> : ITable<T>, ITableMutable<T>, IDisposable, IAsyncDisp
...
+		public TempTable(IDataContext db,
+			IEnumerable<T>   items,
+			EntityDescriptor tableDescriptor,
+			BulkCopyOptions? options      = default,
+			string?          tableName    = default,
+			string?          databaseName = default,
+			string?          schemaName   = default,
+			string?          serverName   = default,
+			TableOptions     tableOptions = default)
+			: this(db, tableDescriptor, items, options, tableName, databaseName, schemaName, serverName, tableOptions)
```
MaceWindu, line 106: mark internal?

MaceWindu (changes requested):

Source/LinqToDB/TempTable.cs
igor-tkachev: We do not need it at all.

viceroypenguin (approved):

viceroypenguin:
@sdanyliv I think I'd rather have the overload rather than the optional parameter, but that's just imho

viceroypenguin (approved):

MaceWindu (dismissed):

MaceWindu:
sql 2019 should be disabled for tests as we use case-sensitive collation for sql 2019 tests which conflicts with tempdb collation

PR #3964, opened 2023-02-10
Add support for custom setters for associations storage members

### LINKED ISSUE #3961: Support custom setters for storage members of associations
Hi, 

My use case is a progressive migration from Linq-to-SQL to LinqToDB.
In L2SQL, the generated code for associations follow this pattern:

```c#
        [global::System.Data.Linq.Mapping.TableAttribute(Name="dbo.Orders")]
	public partial class Order : INotifyPropertyChanging, INotifyPropertyChanged
        {
                private EntitySet<OrderDetail> _OrderDetails;
		private EntityRef<Customer> _Customer;

		public Order()
		{
			this._OrderDetails = new EntitySet<OrderDetail>(new Action<OrderDetail>(this.attach_OrderDetails), new Action<OrderDetail>(this.detach_OrderDetails));
			this._Customer = default(EntityRef<Customer>);
			OnCreated();
		}
                
		[global::System.Data.Linq.Mapping.AssociationAttribute(Name="FK_Order_Details_Orders", Storage="_OrderDetails", ThisKey="OrderID", OtherKey="OrderID", DeleteRule="NO ACTION")]
		public EntitySet<OrderDetail> OrderDetails
		{
			get
			{
				return this._OrderDetails;
			}
			set
			{
				this._OrderDetails.Assign(value);
			}
		}

		[global::System.Data.Linq.Mapping.AssociationAttribute(Name="FK_Orders_Customers", Storage="_Customer", ThisKey="CustomerID", OtherKey="CustomerID", IsForeignKey=true)]
		public Customer Customer
		{
			get
			{
				return this._Customer.Entity;
			}
			set
			{
				Customer previousValue = this._Customer.Entity;
				if (((previousValue != value) 
							|| (this._Customer.HasLoadedOrAssignedValue == false)))
				{
					this.SendPropertyChanging();
					if ((previousValue != null))
					{
						this._Customer.Entity = null;
						previousValue.Orders.Remove(this);
					}
					this._Customer.Entity = value;
					if ((value != null))
					{
						value.Orders.Add(this);
						this._CustomerID = value.CustomerID;
					}
					else
					{
						this._CustomerID = default(string);
					}
					this.SendPropertyChanged("Customer");
				}
			}
		}
        }
```

As you can see, the storage member is different from the visible member type. and obviously that causes an error when trying to use `.LoadWith()`:
```
  Message: 
    System.ArgumentException : Expression of type 'Northwind.Customer' cannot be used for assignment to type 'System.Data.Linq.EntityRef`1[Northwind.Customer]'

  Stack Trace: 
    Expression.Assign(Expression left, Expression right)
    TableContext.SetLoadWithBindings(Type objectType, ParameterExpression parentObject, List`1 exprs)
    TableContext.BuildDefaultConstructor(EntityDescriptor entityDescriptor, Type objectType, Tuple`2[] index)
    TableContext.BuildTableExpression(Boolean buildBlock, Type objectType, Tuple`2[] index)
    TableContext.BuildQuery(Type tableType, TableContext tableContext, ParameterExpression parentObject)
    TableContext.BuildQuery[T](Query`1 query, ParameterExpression queryParameter)
    PassThroughContext.BuildQuery[T](Query`1 query, ParameterExpression queryParameter)
    FirstSingleContext.BuildQuery[T](Query`1 query, ParameterExpression queryParameter)
    ExpressionBuilder.Build[T]()
    Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, Expression expr)
    Query`1.GetQuery(IDataContext dataContext, Expression& expr, Boolean& dependsOnParameters)
    ExpressionQuery`1.GetQuery(Expression& expression, Boolean cache, Boolean& dependsOnParameters)
    IQueryProvider.Execute[TResult](Expression expression)
    Queryable.First[TSource](IQueryable`1 source)
    QueryableTests.TestLoadWith() line 72
```
Type converters are not a good solution here (they are not used in that part of the code anyway) because I'd like to avoid unnecessary allocations and use `EntityRef<T>.Entity = value` and `EntitySet<T>.SetSource(value)` respectively.

After some reading in the code, I think I could implement it by adding an `Expression SetExpression` and matching `string SetExpressionMethod` in the AssociationAttribute class, and use it in `SetLoadWithBindings` to replace the assignment if available:
https://github.com/linq2db/linq2db/blob/6e5c436e8debd24b74e27880b9628edc925e140d/Source/LinqToDB/Linq/Builder/TableBuilder.TableContext.cs#L216-L220
Is there a better way to achieve what I'm trying to do? Or would you accept a pull request adding this functionality?

Thanks in advance.

guillaume86:
I'm not sure about the performance aspect, do you think I should cache the 
```
var setMethod = descriptor.GetSetterMethod(storageMember.Type, ex.Type)!;
```
or is it useless because the resulting expression is cached at a higher level?

Edit: fixes #3961

### DIFF
--- a/Source/LinqToDB/Mapping/AssociationAttribute.cs
+++ b/Source/LinqToDB/Mapping/AssociationAttribute.cs
@@ -88,7 +88,7 @@ namespace LinqToDB.Mapping
 		/// <para>
 		/// <example>
 		/// <code>
-		/// var Expression&lt;Func&lt;SomeEntity, IDataContext, IQueryable&lt;SomeOtherEntity&gt;&gt;&gt; associationQuery;
+		/// Expression&lt;Func&lt;SomeEntity, IDataContext, IQueryable&lt;SomeOtherEntity&gt;&gt;&gt; associationQuery;
 		/// <para />
 		/// associationQuery = (e, db) =&gt; db.GetTable&lt;SomeOtherEntity&gt;().Where(se =&gt; se.Id == e.Id);
 		/// </code>
@@ -124,22 +124,22 @@ namespace LinqToDB.Mapping
 		/// </example>
 		/// </para>
 		/// </summary>
-		public string? SetExpressionMethod { get; set; }
+		public string? AssociationSetterExpressionMethod { get; set; }
 
 		/// <summary>
-		/// Specifies a set expression. If is set, it will be used to set the storage member when using LoadWith().
+		/// Specifies a setter expression. If is set, it will be used to set the storage member when using LoadWith().
 		/// Action takes two parameters: the storage member and the value to assign to it.
 		/// <para>
 		/// <example>
 		/// <code>
-		/// var Expression&lt;Action&lt;SomeContainerType,SomeOtherEntity&gt;&gt; setContainerValue;
+		/// Expression&lt;Action&lt;SomeContainerType,SomeOtherEntity&gt;&gt; setContainerValue;
 		/// <para />
 		/// setContainerValue = (container, value) =&gt; container.Value = value;
 		/// </code>
 		/// </example>
 		/// </para>
 		/// </summary>
-		public Expression? SetExpression { get; set; }
+		public Expression? AssociationSetterExpression { get; set; }
 
 		internal bool?      ConfiguredCanBeNull;
 		/// <summary>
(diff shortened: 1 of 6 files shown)


guillaume86:
After a quick test, it seems caching would be useful, but I'm not sure it's worth it as it would only speed up the compilation of queries that are not in cache (first time running usually).

MaceWindu:
I think we should rename new options from `Setter` to `AssociationSetter` as they are for associations only

guillaume86:
> I think we should rename new options from `Setter` to `AssociationSetter` as they are for associations only

Ok I'll get to it

Source/LinqToDB/Mapping/AssociationAttribute.cs
```diff
@@ -103,6 +103,44 @@ public AssociationAttribute()
...
+		/// </summary>
+		public string? SetExpressionMethod { get; set; }
+
+		/// <summary>
+		/// Specifies a set expression. If is set, it will be used to set the storage member when using LoadWith().
+		/// Action takes two parameters: the storage member and the value to assign to it.
+		/// <para>
+		/// <example>
+		/// <code>
+		/// var Expression&lt;Action&lt;SomeContainerType,SomeOtherEntity&gt;&gt; setContainerValue;
```
viceroypenguin, line 135:
```suggestion
		/// Expression&lt;Action&lt;SomeContainerType,SomeOtherEntity&gt;&gt; setContainerValue;
```

viceroypenguin (approved):

MaceWindu (approved):

guillaume86:
Thanks for the quick replies/feedback :)

PR #3966, opened 2023-02-11
ClickHouse does not support EXISTS, switch to IN

igor-tkachev:


viceroypenguin (approved):

MaceWindu (changes requested):
That's not correct. ClickHouse supports EXISTS, but doesn't support correlated subqueries https://github.com/ClickHouse/ClickHouse/issues/6697

MaceWindu:
My point is:
- user should write queries taking into account this behavior
- or: we should implement set of optimizations in ClickHouseSqlOptimizer to rewrite queries that use correlated columns when it is possible

igor-tkachev:
The problem is when you use `expr.Cotains(subQuery)' l2db generates EXISTS in the way that ClickHouse does not support. If l2db generates IN, it works just fine. So, there is no way for user to write queries taking into account ClickHouse capabilities.

igor-tkachev:
BTW, maybe it's a good idea to always generate IN for scalar expression and EXISTS for objects. Another option is to and an 'Option.PrefereExistsThenIn' (or something), so user will be able to control it explicitly.

viceroypenguin:
I would say `IN` is the more natural operator for scalars, so I'm not opposed to switching to `IN` by default and using `EXISTS` only for non-scalars.

Source/LinqToDB/Linq/Builder/ContainsBuilder.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq.Expressions;
+using LinqToDB.Extensions;
```
viceroypenguin, line 3: Move inside namespace with other l2db usings.

viceroypenguin (dismissed):

viceroypenguin (dismissed):

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlOptimizer.cs
```diff
@@ -244,6 +244,9 @@ static SqlFunction ConvertFunc(SqlFunction func)
 					}
 				}
 
+				case SqlFunction(_, "CASE", [_, SqlValue(true), SqlValue(false)]) f :
+					return new SqlFunction(f.SystemType, f.Name, f.Parameters[0], new SqlValue(1), new SqlValue(0));
```
MaceWindu, line 248: Why it was added?

Source/LinqToDB/Linq/Builder/ContainsBuilder.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq.Expressions;
+using System.Security.Cryptography.X509Certificates;
```
MaceWindu, line 3: randomly selected using kindly added by Visual Studio

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlOptimizer.cs
igor-tkachev: This code is not supported by Octonica provider

```sql
SELECT
	CASE
		WHEN (NOT EXISTS(
			SELECT
				*
			FROM
				Child c_1
			WHERE
				c_1.ParentID <= toInt32(3)
		))
			THEN true -- toInt32(1)
		ELSE false        -- toInt32(0)
	END
```

```
Octonica.ClickHouseClient.Exceptions.ClickHouseException : The type "Bool" is not supported.
```

```toInt32(1)``` solves the problem.

Comment (igor-tkachev) [reply Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlOptimizer.cs]: Only ClickHouse.Client fully support it.

File: Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlOptimizer.cs [hunk 4]
```diff
@@ -244,6 +244,9 @@ static SqlFunction ConvertFunc(SqlFunction func)
 					}
 				}
 
+				case SqlFunction(_, "CASE", [_, SqlValue(true), SqlValue(false)]) f when SqlProviderFlags.IsProjectionBoolSupported is false:
+					return new SqlFunction(f.SystemType, f.Name, f.Parameters[0], new SqlValue(1), new SqlValue(0));
```
MaceWindu, line 248:
```suggestion
					return new SqlFunction(f.SystemType, f.Name, f.Parameters[0], new SqlValue((byte)1), new SqlValue((byte)0));
```

MaceWindu (approved):

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlOptimizer.cs
MaceWindu: better use byte, as `boolean` is mapped to `UInt8`. Actually boolean support already added to Octonica provider, but we don't update it as it currently breaks bulk copy for booleans (awaiting for fix release)

MaceWindu (approved):

PR #3967, opened 2023-02-12
DataOptions: fix equality operators nullability

MaceWindu:
Discovered during ef.core extension migration:

- `DataOptions` equality operators doesn't work with null values
- `UseDB` configuration methods should have overloads without connection string to allow provider configuration without specifying connection string

### DIFF
--- a/Source/LinqToDB/DataOptions.cs
+++ b/Source/LinqToDB/DataOptions.cs
@@ -160,14 +160,16 @@ namespace LinqToDB
 			return ((IConfigurationID)this).ConfigurationID;
 		}
 
-		public static bool operator ==(DataOptions t1, DataOptions t2)
+		public static bool operator ==(DataOptions? t1, DataOptions? t2)
 		{
+			if (ReferenceEquals(t1, t2))
+				return true;
+			if (t1 is null || t2 is null)
+				return false;
+
 			return t1.Equals(t2);
 		}
 
-		public static bool operator !=(DataOptions t1, DataOptions t2)
-		{
-			return !t1.Equals(t2);
-		}
+		public static bool operator !=(DataOptions? t1, DataOptions? t2) => !(t1 == t2);
 	}
 }
--- a/Source/LinqToDB/Compatibility/System/Data/Linq/Binary.cs
+++ b/Source/LinqToDB/Compatibility/System/Data/Linq/Binary.cs
@@ -43,27 +43,15 @@ namespace System.Data.Linq
 
 		public static bool operator ==(Binary? binary1, Binary? binary2)
 		{
-			if (binary1 is null && binary2 is null)
+			if (ReferenceEquals(binary1, binary2))
 				return true;
 			if (binary1 is null || binary2 is null)
 				return false;
-			if (ReferenceEquals(binary1, binary2))
-				return true;
 
-			return binary1.EqualsTo(binary2);
+			return binary1.Equals(binary2);
 		}
 
-		public static bool operator !=(Binary? binary1, Binary? binary2)
-		{
-			if (binary1 is null && binary2 is null)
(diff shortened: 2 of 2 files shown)


Source/LinqToDB/DataOptions.cs
```diff
@@ -160,13 +160,27 @@ public override int GetHashCode()
...
+			if (t1 is null || t2 is null)
+				return false;
+			if (ReferenceEquals(t1, t2))
+				return true;
+
 			return t1.Equals(t2);
 		}
 
-		public static bool operator !=(DataOptions t1, DataOptions t2)
+		public static bool operator !=(DataOptions? t1, DataOptions? t2)
```
viceroypenguin, line 175: Better to abstract `!=` as `!(t1 == t2)`, so that logic between the two cannot accidentally get mixed up.

Source/LinqToDB/Extensions/DataOptionsExtensions.Provider.cs
```diff
@@ -22,7 +22,25 @@ public static partial class DataOptionsExtensions
 		#region UseSqlServer
 
 		/// <summary>
-		/// Configure connection to use specific SQL Server provider, dialect and connection string.
+		/// Configure connection to use specific SQL Server provider, dialect and provider-specific options.
```
viceroypenguin, line 25: Changes to this file do not appear to be related to nullability. Perhaps better in a separate PR?

viceroypenguin (changes requested):

Source/LinqToDB/Extensions/DataOptionsExtensions.Provider.cs
```diff
@@ -22,7 +22,25 @@ public static partial class DataOptionsExtensions
...
+		/// </summary>
+		/// <param name="options">Instance of <see cref="DataOptions"/>.</param>
+		/// <param name="provider">SQL Server provider to use.</param>
+		/// <param name="dialect">SQL Server dialect support level.</param>
+		/// <param name="optionSetter">Optional <see cref="SqlServerOptions"/> configuration callback.</param>
+		/// <returns>The builder instance so calls can be chained.</returns>
+		public static DataOptions UseSqlServer(this DataOptions options,
+			SqlServerVersion  dialect                              = SqlServerVersion. AutoDetect,
```
igor-tkachev, line 33: Does it make any sense? What is the use case?

MaceWindu: To allow provider options configuration for ef.core integration. Connection string is already known there

https://github.com/linq2db/linq2db.EntityFrameworkCore/pull/293/files#diff-e669fdecf3f6f7caa0689c39fa9a7914c75472bfafbac50a9cb6981401ad0c2dR64

Source/LinqToDB/DataOptions.cs
```diff
@@ -160,13 +160,27 @@ public override int GetHashCode()
...
 		}
 
-		public static bool operator ==(DataOptions t1, DataOptions t2)
+		public static bool operator ==(DataOptions? t1, DataOptions? t2)
 		{
+			if (t1 is null && t2 is null)
+				return true;
+			if (t1 is null || t2 is null)
+				return false;
+			if (ReferenceEquals(t1, t2))
```
jods4, line 169: You can put `ReferenceEquals` first and drop the first check.
`ReferenceEquals` will be true if both `t1` and `t2` are both `null`.

Source/LinqToDB/Extensions/DataOptionsExtensions.Provider.cs
MaceWindu: #3970

PR #3969, opened 2023-02-13
Move EntityDescriptorCreatedCallback to options

### LINKED ISSUE #1592: MappingSchema.EntityDescriptorCreatedCallback should be static
Right now MappingSchema.EntityDescriptorCreatedCallback is defined per-MappingSchema instance, which makes it non-trivial to configure, because user should attach callback to current data connection mapping schema instance.

Attaching it to DefaultSchema will not work, because it is not used directly - we copy schemas from it to actual mapping schema, used by connection in MappingSchema constructor.

There are several ways to fix it:
- make callback application-wide static, so it will be called always and then users could inspect mapping schema parameter, if he wants to perform action only for specific schemas
- derive callbacks when we aggregate mapping schemas. Not a big fan of it - in case of multiple schemas having callbacks we need to call all of them
- move callback to DataConnection

I would prefer first approach.

Also see: https://github.com/linq2db/linq2db/issues/1074#issuecomment-460057547

### Environment details
linq2db version: *2.6.4*

### LINKED ISSUE #1822: EntityDescriptorCreatedCallback not working
I tried to use EntityDescriptorCreatedCallback to use snake_case names.

But it never triggered.

I see the same issues here and tried different ways including to set it inside DataBaseConnection Constructor with the same result.

image

MaceWindu:
Fix #1592
Fix #1822

With this fix user could define:
- application-wide callback using `MappingSchema.EntityDescriptorCreatedCallback` static property (was instance before fix)
- context-level option using `WithOnEntityDescriptorCreated`/`UseOnEntityDescriptorCreated` configuration extensions


Tests/Linq/UserTests/Issue1128Tests.cs
```diff
@@ -49,10 +49,10 @@ public void TestEntityDescriptor()
 		{
 			var ms = SetFluentMappings();
 
-			var ed1 = ms.GetEntityDescriptor(typeof(FluentBase));
-			var ed2 = ms.GetEntityDescriptor(typeof(FluentDerived));
-			var ed3 = ms.GetEntityDescriptor(typeof(AttributeBase));
-			var ed4 = ms.GetEntityDescriptor(typeof(AttributeBase));
+			var ed1 = ms.GetEntityDescriptor(typeof(FluentBase), null);
```
igor-tkachev, line 52: Create either default parameter or overload.

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1700,28 +1701,31 @@ public StringComparer ColumnNameComparer
...
 
 		/// <summary>
 		/// Returns mapped entity descriptor.
 		/// </summary>
 		/// <param name="type">Mapped type.</param>
+		/// <param name="onEntityDescriptorCreated">Action, called when new descriptor instance created.
+		/// When set to <c>null</c>, <see cref="EntityDescriptorCreatedCallback" /> callback used.</param>
 		/// <returns>Mapping descriptor.</returns>
-		public EntityDescriptor GetEntityDescriptor(Type type)
+		public EntityDescriptor GetEntityDescriptor(Type type, Action<MappingSchema, IEntityChangeDescriptor>? onEntityDescriptorCreated)
```
igor-tkachev, line 1719: Make parameter default or create overload without parameter.

igor-tkachev (changes requested):

Source/LinqToDB/Mapping/MappingSchema.cs
MaceWindu: I thought about it - it's useful for tests but in library code we don't pass `null` (and shouldn't)

igor-tkachev: ... and compatibility.

igor-tkachev (approved):

viceroypenguin (approved):

sdanyliv (approved):

Issue #2798, opened 2021-01-27
Oracle - Linq2db creates to long alias for Column name in SQL


jogibear9988:
We have a Database where we have a long Column Name:

"OutfeedTransportOrderPosition"

but linq2db adds "_1" in the sql but this fails in oracle cause 30chars is max size

![image](https://user-images.githubusercontent.com/364896/105967767-86e57180-6086-11eb-9139-6d65dad57de8.png)

jogibear9988:
problem is we use Orcale 11, and the longer identifiers are only possible in orcale v12

MaceWindu:
Currently we don't validate length of generated ids. I would prefer it fixed in generic way: max-length property for provider + length check in identifier generation logic

jogibear9988:
At the moment we generate the Parameternames at OptimizationContext.CorrectParamName, is this correct?
How should we access QueryProvider from here? (For Provider specific setting)

MaceWindu:
Probably not gonna pick it up for next release, so I will add implementation notes here:
- we should implement identifier trimming for autogenerated identifiers: alias, parameter name, identity sequence and trigger names
- as trimming logic is not straightforward for some databases it should be possible to override it to avoid blocking issues for users. For that we will adopt interceptors approach from ef.core: https://docs.microsoft.com/en-us/ef/core/logging-events-diagnostics/interceptors by introducing ISqlGenerationIterceptor with 4 initial operations (alias, parameter, sequence and trigger name generators)

Identifier limit vary per database not only by length but also by measurement unit. E.g.:
- in characters
- in bytes in specific encoding
- in bytes in database encoding
Also for some of databases documentation is vague and lack all requred information, so it should be covered with tests to detect and test actual limits.

Per-db limits:
- access : 64 characters (https://support.microsoft.com/en-us/topic/access-specifications-0cf3c66f-9cf2-4e32-9568-98c1025bb47c?ui=en-us&rs=en-us&ad=us)
- db2 luw : 128 bytes in database codepage... (https://www.ibm.com/support/knowledgecenter/SSEPGG_10.1.0/com.ibm.db2.luw.sql.ref.doc/doc/r0001029.html) + read codepage https://knowledgebase.progress.com/articles/Article/3218
- db2 zos : 128 utf8 bytes, 30 for column (https://www.ibm.com/support/knowledgecenter/SSEPEK_10.0.0/sqlref/src/tpc/db2z_limits.html)
[...]

PR #3977, opened 2023-02-17
Fix type conversion for association storage/setter expressions with different type

### LINKED ISSUE #3975: Convert the association value to the correct type when using a Storage or SetterExpression which use a different type
When using a storage member or a setter expression that is defined with a different type that the member type, linq2db should use the correct type conversions. 

At the moment, the value is always converted to the member type, but in the case of an AssociationSetterExpression which accepts a different type, or a Storage member with a different type, that conversion is not correct.

Here's a test that expose this issue (paste it in AssociationTests):

```
		// at the moment it must be generic because linq2db will infer entity type from generic arguments
		sealed class ChildrenContainer<T> : IEnumerable<T> where T : Child
		{
			public List<T>? Value;

			public IEnumerator<T> GetEnumerator()
			{
				return ((IEnumerable<T>)Value!).GetEnumerator();
			}

			public void SetValue(IEnumerable<T> value)
			{
				Value = value.ToList();
			}

			IEnumerator IEnumerable.GetEnumerator()
			{
				return ((IEnumerable)Value!).GetEnumerator();
			}
		}

		[Table("Parent")]
		[UsedImplicitly]
		sealed class ListAssociationSetterExpressionTestClass
		{
			[Column] public int ParentID;

			ChildrenContainer<Child> _children = new();

			[Association(ThisKey = "ParentID", OtherKey = "ParentID", Storage = "_children", AssociationSetterExpressionMethod = nameof(SetChildrenValue))]
			public ChildrenContainer<Child> Children
			{
				get => _children;
				set => throw new InvalidOperationException();
			}

			public static Expression<Action<ChildrenContainer<Child>, IEnumerable<Child>>> SetChildrenValue()
			{
				return static (ChildrenContainer<Child> container, IEnumerable<Child> value) => container.SetValue(value);
			}
		}

		[Test]
		public void ListAssociationSetterExpressionTest([DataSources] string context)
		{				
			{
				// we want to make sure the conversion is not possible because we want to bypass
				// that conversion if the setter value parameter type (IEnumerable<Child> in this case)
				// does not match the member type (ChildrenContainer<Child> in this case)
				Assert.Throws<LinqToDB.Common.LinqToDBConvertException>(() =>
					db.MappingSchema.ChangeType(new List<Child>(0), typeof(ChildrenContainer<Child>)),
					"List<Child> should not be convertible to ChildrenContainer<Child>");

				var value = db.GetTable<ListAssociationSetterExpressionTestClass>().LoadWith(x => x.Children).First();

				Assert.That(value.Children, Is.Not.Null);
			}
		}
```

I'm going to post a PR to solve this, I'm creating this issue to follow the procedure and get an issue number for my branch.

Hope this is OK, see you soon in the PR discussion.

guillaume86:
fix #3975

### DIFF
--- a/Build/Azure/pipelines/templates/build-job.yml
+++ b/Build/Azure/pipelines/templates/build-job.yml
@@ -17,7 +17,7 @@ jobs:
     displayName: 'Install .NET'
     inputs:
       packageType: sdk
-      version: 7.x
+      version: 7.0.102
 
   - task: PowerShell@2
     inputs:
--- a/Build/Azure/pipelines/templates/build-vars.yml
+++ b/Build/Azure/pipelines/templates/build-vars.yml
@@ -7,6 +7,6 @@ variables:
   build_configuration: 'Azure'
   assemblyVersion: 5.0.0
   ${{ if eq(variables['Build.SourceBranchName'], 'release') }}:
-    packageVersion: 5.0.0-rc.2
+    packageVersion: 5.0.0
   ${{ if ne(variables['Build.SourceBranchName'], 'release') }}:
     packageVersion: 5.0.0
--- a/Build/Azure/pipelines/templates/test-workflow-linux.yml
+++ b/Build/Azure/pipelines/templates/test-workflow-linux.yml
@@ -22,7 +22,7 @@ steps:
   displayName: 'Install .NET 7'
   inputs:
     packageType: sdk
-    version: 7.x
+    version: 7.0.102
   condition: and(variables.title, succeeded())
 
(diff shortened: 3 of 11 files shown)


MaceWindu:
tests failed

MaceWindu:
Actually it's just too long alias name #2798

Could you rename `ListAssociationSetterExpressionTestClass` to something shorter?

guillaume86:
Sure will do

MaceWindu (approved):

PR #3982, opened 2023-02-20
Simplify IMetadataReader.GetAttributes API

### LINKED ISSUE #3981: Purpose of generic argument in IMetadataReader.GetAttributes<T>
Is there any puprpose in the generic attributte in IMetadataReader.GetAttributes<T> after the attriutes refactoring? Unless I missed something the framework internally only calls IMetadataReader.GetAttributes<MappingAttribute> and I don't see IMetadataReader as part of the public API. Even so it's a minor breaking change, the consumer can simply filter the results with. This would simplify the implementation of provider implementations and remove a lot of type checking.

MaceWindu:
Fix #3981

After IMetadataReader refactoring [here](#3900) `GetAttributes` methods always called using `MappingAttribute` as type parameter, which makes it unnecessary and confusing.

### DIFF
--- a/Source/LinqToDB/Metadata/AttributeInfo.cs
+++ b/Source/LinqToDB/Metadata/AttributeInfo.cs
@@ -44,7 +44,7 @@ namespace LinqToDB.Metadata
 										member,
 										Expression.Constant(Converter.ChangeType(k.Value, mtype), mtype));
 								})),
-							typeof(Attribute)));
+							typeof(MappingAttribute)));
 
 					_func = expr.CompileExpression();
 				}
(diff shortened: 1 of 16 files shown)


Source/LinqToDB/Metadata/AttributeInfo.cs
```diff
@@ -7,6 +7,7 @@ namespace LinqToDB.Metadata
 {
 	using Common;
 	using Extensions;
+	using LinqToDB.Mapping;
```
viceroypenguin, line 10:
```suggestion
	using Mapping;
```

viceroypenguin (dismissed):

PR #3985, opened 2023-02-21
Fix copy-paste errors in InputWithOutput* APIs

### LINKED ISSUE #3983: The InsertWithOutputIntoAsync method does not work correctly
Hi, I wanted to call the InsertWithOutputIntoAsync method with the following request:

```c#
 await source.InsertWithOutputIntoAsync(
    targetTable,
    valueFromTargetTable=> valueFromTargetTable,
    otherTable,
    valueFromTargetTable=> new OtherClass
            {
                Id = valueFromTargetTable.Id, 
                Status = 1
            });
```

But I caught an exception when calling such code:

```
System.InvalidCastException : Unable to cast object of type 'LinqToDB.Linq.Table`1[OtherTable]' to type 'System.Linq.IQueryable`1[TargetTable]'
```

When I started looking at the source code, what I saw was that the method has this code `(IQueryable<TTarget>)outputTable`
link to the source:
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/LinqExtensions.Insert.cs/#L300

Why do you need a type conversion? The semantics of the method allows you to call the code that I have.

MaceWindu:
Fix #3983

### DIFF
--- a/Source/LinqToDB/LinqExtensions.Insert.cs
+++ b/Source/LinqToDB/LinqExtensions.Insert.cs
@@ -358,7 +358,7 @@ namespace LinqToDB
 				Expression.Call(
 					null,
 					MethodHelper.GetMethodInfo(InsertWithOutputInto, target, setter, outputTable, outputExpression),
-					query.Expression, Expression.Quote(setter), ((IQueryable<TTarget>)outputTable).Expression,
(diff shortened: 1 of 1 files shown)


viceroypenguin (approved):
All `outputTable` parameters look correct. Do we need to do a similar review of `UpdateWithOutput*` and `DeleteWithOutput*`?

MaceWindu:
update/delete extensions looks correct

PR #3989, opened 2023-02-22
Visual Studio 17.5 build fixes

MaceWindu:
Will be merged after Azure updated to 17.5

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -36,6 +36,8 @@
 
 		<!--required for SkipLocalInit-->
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+		<AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
 	</PropertyGroup>
 
 	<PropertyGroup Label="Code Analysis Setup">
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -38,6 +38,7 @@
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
 
 		<AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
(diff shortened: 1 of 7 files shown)


igor-tkachev (approved):

Directory.Build.props
```diff
@@ -36,6 +36,8 @@
 
 		<!--required for SkipLocalInit-->
 		<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+		<AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
```
drewnoakes, line 40: For build acceleration to work properly with this solution, you must configure your projects to produce a reference assembly.

```suggestion
		<AccelerateBuildsInVisualStudio>true</AccelerateBuildsInVisualStudio>
		<ProduceReferenceAssembly>true</ProduceReferenceAssembly>
```

We are investigating turning this on by default in newer versions of MSBuild, but until then this will help your incremental build performance considerably.

MaceWindu: Thanks!

jods4: > We are investigating turning this on by default in newer versions of MSBuild, but until then this will help your incremental build performance considerably.

According to MS documentation, `ProduceReferenceAssembly` is true by default for projects targeting .net 5+?
I know in linq2db we have targets < net50 but your remark sounds like it is never on by default with current MSBuild? I'm interested for other projects.

drewnoakes: That's right, but build acceleration needs to apply to all target frameworks, because VS builds them all in one hit. The projects here have .NET Framework targets, for which this is (currently) not the default.

That's mostly a quirk of history and tooling support. However the tools are now compatible, so we are hoping to set `ProduceReferenceAssembly` for all projects, regardless of target framework.

PR #3990, opened 2023-02-22
Fix #3986

### LINKED ISSUE #3986: Incorrect SQL generated over 'Contains' in array having null element
### Describe your issue

Consider the following table 
```sql
create table issue
(
    number        integer,
    number_string text
);
```
with the following data:
```sql
insert into issue(number, number_string)
VALUES (0, null), (1, 'one'), (2, 'two');
```
and with the following class generated:
```c#
[Table("issue", Schema = "public")]
	public partial class Issue
	{
		[Column("number"       )] public int?    Number       { get; set; } // integer
		[Column("number_string")] public string? NumberString { get; set; } // text
	}

```

### Steps to reproduce

Consider the following test:

```c#
        [Test]
        public async Task IssueTest()
        {
            var dataConnection = new LinqToDB.Data.DataConnection();

            var strings = new[] { null, "two" };

            var result = await dataConnection.GetTable<Issue>().Where(i => i.Number == 1 && strings.Contains(i.NumberString)).ToListAsync();

            var lastQuery = dataConnection.LastQuery;

            TestContext.WriteLine(lastQuery);

            Assert.AreEqual(0, result.Count);
        }
```

The condition over the table 'Issue' should obviously lead to a query that returns no rows. But that's not the case, the record (0, null) is actually returned. 
The bug is pretty clear, below is the SQL generated by linq2db and we can see that the necessary brackets are missed there.

```sql
SELECT
	i."number",
	i.number_string
FROM
	"public".issue i
WHERE
	i."number" = 1 AND i.number_string IN ('two') OR i.number_string IS NULL
```

### Environment details

`Linq To DB` version: *4.4.1*

Database (with version): *Postgres 14.1*

jods4:
Fix #3986

`OR xx IS NULL` should be wrapped with parens like `xx IN () OR xx IN ()` is, as it might be part of a larger AND expression.

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlBuilder.cs
@@ -2690,7 +2690,7 @@ namespace LinqToDB.SqlProvider
 			 		}
 				}
 
-				if (multipleParts && !hasNull)
+				if (multipleParts)
 					StringBuilder.Insert(len, "(").Append(')');
 			}
 		}
--- a/Tests/Linq/Linq/ContainsTests.cs
+++ b/Tests/Linq/Linq/ContainsTests.cs
@@ -274,5 +274,60 @@ namespace Tests.Linq
 			Value3,
 			Value4,
 		}
+
+		private static readonly string?[][] _issue3986Cases1 = new string?[][]
+		{
+			new string?[] { null, "Ko" },
+			new string?[] { "Ko", null },
+			new string?[] { null, "Ko", null },
+			new string?[] { "Ko", null, null },
+			new string?[] { null, null, "Ko" },
+			new string?[] { "123", null, "Ko" },
+			new string?[] { "123", "Ko", null },
+			new string?[] { null, "123", "Ko" },
+			new string?[] { null, null },
+			Array.Empty<string?>()
+		};
+
+		[Test]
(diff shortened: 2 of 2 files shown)


viceroypenguin (approved):

MaceWindu:
That's something for SqlOptimizer to do. I'm currently trying to implement it in SqlOptimizer and I will fail - we will use SqlBuilder fix for release

jods4:
I thought my fix was incorrect because of the `WithNull` clause that always add a `OR` but it turns out it's ok.
This is not obvious, so here's the full explanation why we don't need more parens:

A positive (`IN`) case that has multiple clauses, or contains null, or has WithNull true will end up like the following expression, which is fine to wrap in parens in case it's connected with `AND` to something else (fixes #3986):
```sql
x IN (..) OR x IN (..) OR x IS NULL
```

A negative (`NOT IN`) will codegen like one of those:
```sql
-- When null is part of the list:
x NOT IN (..) AND x NOT IN (..) AND x IS NOT NULL
-- When WithNull is true:
x NOT IN (..) AND x NOT IN (..) OR x IS NULL
```
Notice the second case is a mix of `AND`/`OR`, which made me think I might have introduced a bug with this fix.
But as `OR` has a lower precedence than `AND` it parses correctly as `(x NOT IN (..) AND x NOT IN (..)) OR x IS NULL`.
These expression can then be wrapped in parens (esp. the second one which has `OR` precedence and won't combine correctly with an outside `AND`.

viceroypenguin (approved):

PR #3991, opened 2023-02-22
Move IN optimizations to SQL Optimizer

### LINKED ISSUE #3986: Incorrect SQL generated over 'Contains' in array having null element

MaceWindu:
Fix #3986

also includes fix from #3990

### DIFF
--- a/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
+++ b/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
@@ -2448,40 +2448,37 @@ namespace LinqToDB.SqlProvider
 							return new SqlPredicate.InList(keys[0], null, p.IsNot, values);
 						}
 
+						var sc = new SqlSearchCondition();
+
+						foreach (var item in items)
 						{
-							var sc = new SqlSearchCondition();
+							var itemCond = new SqlSearchCondition();
 
-							foreach (var item in items)
+							foreach (var key in keys)
 							{
-								var itemCond = new SqlSearchCondition();
+								var field    = ExpectsUnderlyingField(key);
+								var cd       = field.ColumnDescriptor;
+								var sqlValue = mappingSchema.GetSqlValueFromObject(cd, item!);
+								//TODO: review
+								var cond = sqlValue.Value == null ?
+									new SqlCondition(false, new SqlPredicate.IsNull  (field, false)) :
+									new SqlCondition(false, new SqlPredicate.ExprExpr(field, SqlPredicate.Operator.Equal, sqlValue, null));
 
-								foreach (var key in keys)
-								{
-									var field    = ExpectsUnderlyingField(key);
-									var cd       = field.ColumnDescriptor;
-									var sqlValue = mappingSchema.GetSqlValueFromObject(cd, item!);
-									//TODO: review
-									var cond = sqlValue.Value == null ?
-										new SqlCondition(false, new SqlPredicate.IsNull  (field, false)) :
-										new SqlCondition(false, new SqlPredicate.ExprExpr(field, SqlPredicate.Operator.Equal, sqlValue, null));
-
-									itemCond.Conditions.Add(cond);
-								}
-
-								sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.Expr(itemCond), true));
+								itemCond.Conditions.Add(cond);
 							}
 
-							if (sc.Conditions.Count == 0)
-								return new SqlPredicate.Expr(new SqlValue(p.IsNot));
-
-							if (p.IsNot)
-								return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
-
-							return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
+							sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.Expr(itemCond), true));
 						}
-					}
 
-					if (p.Expr1 is SqlObjectExpression expr)
+						if (sc.Conditions.Count == 0)
+							return new SqlPredicate.Expr(new SqlValue(p.IsNot));
+
+						if (p.IsNot)
+							return new SqlPredicate.NotExpr(sc, true, Precedence.LogicalNegation);
+
+						return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
+					}
+					else if (p.Expr1 is SqlObjectExpression expr)
 					{
 						var parameters = expr.InfoParameters;
 						if (parameters.Length == 1)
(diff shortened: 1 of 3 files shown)


MaceWindu:
Disable whitespaces changes before review. GH diff goes crazy otherwise

Shane32:
I was just noticing the title of this PR and wondering if there’s a chance to add optimizations for IN with lists of certain types. Specifically, number types and guids. By using string split the query within SQL Server can be cached and speed should not suffer (since essentially it is string-split already). It also allows to provide lists that would otherwise exceed the maximum SQL query length.  I write my own code to implement this within my repositories but I think it would be a great feature to add to linq2db by default. Also wondering if this is something that would fit in principle with the project. 

I also have similar optimizations for lists of strings but I use a TVP which is not available without schema changes to the database.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2525,6 +2522,53 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
 
 						return new SqlPredicate.Expr(sc, Precedence.LogicalDisjunction);
 					}
+					else if (p.WithNull != null && paramValue.ProviderValue is not string)
+					{
+						List<ISqlExpression>? newValues = null;
+						var idx                         = 0;
+						var hasNull                     = false;
+
+						foreach (var value in items)
```
jods4, line 2531: This is a pretty obscure loop for something that, if I understand correctly, boils down to (I agree the impure Where is unusual):

```csharp
items
  .Where(x => 
  { 
    bool isNull = x == null;
    hasNull |= isNull;
    return isNull;
  })
  .Select(x => new SqlValue(x))
  .ToList();
  
  // Then change the check (newValues == null) to (newValues.Count == 0)
  ```
  
  Alternatively: 
  ```csharp
  var newValues = new List<SqlValue>(items.TryGetNonEnumeratedCount(out int len) ? len : 8);
  foreach (var value in items)
  {
    if (value == null) hasNull = true;
    else newValues.Add(new SqlValue(value));
  }
```

jods4:
@Shane32 if we go this route, note that Oracle accepts an array-typed parameter for `x IN :param`, which would be a similar optimisation.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2525,6 +2522,53 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
...
+									if (idx > 0)
+									{
+										newValues = new List<ISqlExpression>();
+										newValues.AddRange(items.Cast<object>().Take(idx).Select(value => new SqlValue(value)));
+									}
+								}
+							}
+							else if (hasNull)
+							{
+								newValues ??= new List<ISqlExpression>();
```
viceroypenguin, line 2547: Is it possible for `newValues` to be `null` on this line?

viceroypenguin (approved):
lgtm

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
MaceWindu: yes, e.g. for this test case: `new string?[] { null, "Ko" },`

hasNull set for first element, but `newValues` created only on second element

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
MaceWindu: it saves list instance creation for edge cases when list contains only nulls, so I would prefer to leave it as it.

(And in any case I don't plan to spend much time now on this PR as I don't want it to block release)

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2525,6 +2522,53 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
...
+								newValues.Add(new SqlValue(value));
+							}
+
+							idx++;
+						}
+
+						if (hasNull)
```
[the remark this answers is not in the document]
MaceWindu: What is missing? I know it is not equivalent as:
- it allows sql optimizer to invert `NOT(contains)` condition
- I generate simplier SQL for edge cases:
  - only nulls in list converted to just `x IS [NOT] NULL`
  - only one not-null value in list converted to `x =/<> value`

igor-tkachev (approved):

sdanyliv (dismissed):

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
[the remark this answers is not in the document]
jods4: @MaceWindu ok I see you made a new commit after my initial review and you noticed the differences around `WithNull = true`.
The new code after the 2nd commit seems now to be in-line with `BasicSqlBuilder`, so I'm marking this conversation as resolved.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2523,53 +2523,75 @@ public virtual ISqlPredicate ConvertInListPredicate(MappingSchema mappingSchema,
...
+				if (val == null)
+				{
+					if (!hasNull)
+					{
+						hasNull = true;
+						if (idx > 0)
+						{
+							newValues = new List<ISqlExpression>();
+							var i = 0;
+							foreach (var item in items)
```
jods4, line 2556: We're enumerating `items` twice, and in a nested fashion.
This should be ok practically always (e.g. with arrays, lists...) but some users might pass enumerables that are costly and don't expect us to enumerate twice.

We might debate if that's just bad usage of linq2db, but given this is easy to avoid (see my suggested alternatives for this part) I would say we should do so and be less surprising for users.

PR #4005, opened 2023-03-02
Metrics

### LINKED ISSUE #3405: OpenTracing (or DiagnosticSource) support
Hi, are there any plans to add opentracing or diagnosticsource support to linq2db? Similar to [EfCore](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Contrib.Instrumentation.EntityFrameworkCore) (this is a wrapper around DiagnosticSource though)

igor-tkachev:
Fix #3405

Added simple API compatible with discussed frameworks.

```c#
public interface IActivity : IDisposable
{
}
```

```c#
public static class ActivityService
{
	public static IActivity? Start(ActivityID metric)
	{
		return _factory?.Invoke(metric);
	}

	static Func<ActivityID,IActivity>? _factory;

	public static void AddFactory(Func<ActivityID,IActivity>? factory)
	{
		_factory = factory;
	}
}
```

```c#
public enum ActivityID
{
    QueryProviderExecuteT,
    QueryProviderExecute,
    QueryProviderGetEnumeratorT,
    QueryProviderGetEnumerator,
    GetQueryTotal,
        GetQueryFind,
            GetQueryFindExpose,
            GetQueryFindFind,
        GetQueryCreate,
            Build,
                BuildSequence,
                    BuildSequenceCanBuild,
                    BuildSequenceBuild,
                ReorderBuilders,
                BuildQuery,
                    FinalizeQuery,
        GetIEnumerable,
    ExecuteQuery,
    ExecuteQueryAsync,
    ExecuteElement,
    ExecuteElementAsync,
    ExecuteScalar,
    ExecuteScalarAsync,
    ExecuteNonQuery,
    ExecuteNonQueryAsync,
    ExecuteNonQuery2,
    ExecuteNonQuery2Async,
    ExecuteScalar2,
    ExecuteScalar2Async,
        BuildSql,
        OnTraceInternal,

        CommandExecuteScalar,
        CommandExecuteScalarAsync,
        CommandExecuteReader,
        CommandExecuteReaderAsync,
        CommandExecuteNonQuery,
        CommandExecuteNonQueryAsync,
}
```


MaceWindu:
if we collect metrics, maybe it's time to look at opentelementry support? https://github.com/linq2db/linq2db/issues/3405

Source/LinqToDB/Tools/Metric.cs
```diff
@@ -0,0 +1,49 @@
...
+	{
+		public Metric(string name)
+		{
+			Name = name;
+		}
+		public string    Name      { get; }
+		public Stopwatch Stopwatch { get; } = new();
+		public int       CallCount { get; private set; }
+
+		public IDisposable Start()
```
jods4, line 19: This is neither thread-safe nor re-entrant because a single Stopwatch is shared per metric.

It will fail to work properly if a metric is started in nested calls.
For instance, if a builder calls into `BuildSequence` recursively (and I'm pretty sure some do).

jods4 (changes requested):

Shane32:
See:
- https://github.com/dotnet/aspnetcore/blob/main/src/Shared/ValueStopwatch/ValueStopwatch.cs

Shane32:
> if we collect metrics, maybe it's time to look at opentelementry support? #3405

We just added OpenTelemetry support in GraphQL.NET. DiagnosticSource has enough methods in .NET 5+ to add telemetry without a dependency. So we added the code behind ifdefs.

OpenTelemetry is rather new, we figured, and not so likely that someone would attempt to use OpenTelemetry with prior versions of .NET. The alternative is to take a dependency on the System.Diagnostics.DiagnosticSource nuget package

MaceWindu:
Yep, sorry. I meant `DiagnosticSource` support as abstraction

Source/LinqToDB/Tools/Metric.cs
igor-tkachev: Do we run any tests in parallel?

jods4: It is also not re-entrant.
"For instance, if a builder calls into BuildSequence recursively (and I'm pretty sure some do)."

jods4:
@igor-tkachev You are basically adding a CPU profiler in the codebase manually (counting calls in time spent (incl. descendants)).

That's a lot of noise, esp. with the `#ifdef`... how are you gonna use it? Would there be a way to use a real profiler instead?

For example if you're gonna look at this manually, if you have VS Professional (or Ultimate), you can profile unit tests individually.

If this becomes some automated stats in our CI, I kind of wish we could use a Profiler library or AOP to achieve a similar thing without having to modify our code so much in every place that could be of interest.

igor-tkachev:
@jods4 There will be some noise, of cause.
These metrics are not for profiling. When I need to profile l2db, I prefer to create a project with a specific case I am going to profile. Also I do not need any linq2db profiling code in my other projects.

The idea of this code is to compare different solutions for optimization. It is not working yet, but we should get reports like the following:

```
+---------------------------+------------------+-----------+------------------+
| Name                      | Elapsed          | CallCount | TimePerCall      |
+---------------------------+------------------+-----------+------------------+
| IQueryProvider.Execute<T> | 00:00:09.6755964 |        80 | 00:00:00.1209449 |
| IQueryProvider.Execute    | 00:00:00         |         0 | 00:00:00         |
|   OnTraceInternal         | 00:00:00.3502372 |       200 | 00:00:00.0017511 |
|   GetQuery                | 00:00:00.7787958 |        80 | 00:00:00.0097349 |
|     Find                  | 00:00:00.3175412 |        80 | 00:00:00.0039692 |
|       Expose              | 00:00:00.1014083 |        80 | 00:00:00.0012676 |
|       Find                | 00:00:00.2161111 |        80 | 00:00:00.0027013 |
|     Create                | 00:00:00.4612266 |        76 | 00:00:00.0060687 |
|       Build               | 00:00:00.3923171 |        76 | 00:00:00.0051620 |
|         BuildSequence     | 00:00:00.0131293 |       228 | 00:00:00.0000575 |
|           CanBuild        | 00:00:00.0005601 |       487 | 00:00:00.0000011 |
|           Build           | 00:00:00.0125966 |       228 | 00:00:00.0000552 |
|   GetIEnumerable          | 00:00:00         |         0 | 00:00:00         |
|   ExecuteElement          | 00:00:08.8949492 |        80 | 00:00:00.1111868 |
| Total                     | 00:00:17.4146554 |         1 | 00:00:17.4146554 |
+---------------------------+------------------+-----------+------------------+
```

For example, we are discussing `CanBuild` performance [here](https://github.com/linq2db/linq2db/pull/3913). If we add metrics to existing code and new one, we will be able to compare performance and make better decision. What I see from the report above, elapsed time of the `CanBuild` method is one thousandth of the `GetQuery` method. Of cause, this is example for just a couple of tests only and we need to get it for all the tests.

igor-tkachev:
@Shane32 Do not you think that adding telemetry permanently to linq2db will not kill performance permanently as well? :)

Source/LinqToDB/Tools/Metric.cs
igor-tkachev: You right, will redo.

igor-tkachev: fixed

Source/LinqToDB/Tools/Metric.cs
```diff
@@ -0,0 +1,56 @@
...
+
+		private int      _callCount;
+		public  int      CallCount => _callCount;
+
+		public Watcher Start()
+		{
+			return new(this);
+		}
+
+		public void Stop(Stopwatch stopwatch)
```
viceroypenguin, line 28: Three recommendations:
* Make `Stop` a `private` method, so that it is only called by `Watcher`s.
* Make `Watcher` an `internal`/`private` class, so that it is an implementation detail rather than a public contract
* Make `_elapsed` a `long`, so that you can use `Interlocked.Add()`, and change `Elapsed => new TimeSpan(_elapsed)`

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -182,6 +182,10 @@ public static ParameterExpression GetDataContextParam()
 
 		public Query<T> Build<T>()
 		{
+#if METRICS
+			using var m = LinqToDB.Tools.Metrics.Build.Start();
```
viceroypenguin, line 186: I wonder if it would make sense to move the `#if METRICS` to `Metric.Start()`, to return a value in `METRICS` and `null` otherwise. If `Start()` is marked `AggressiveInlining`, then it should have negligible perf impact in Release, but be able to keep the code cleaner vis-a-vis having these `#if`s all over the place.

Shane32:
> @Shane32 Do not you think that adding telemetry permanently to linq2db will not kill performance permanently as well? :)

`ActivitySource.StartActivity` returns `null` if there are no attached listeners.  Also, there is `activity.IsAllDataRequested` that can be checked before any tags requiring significant CPU work are added.  So essentially the telemetry is only recording data on an as-needed basis.

Within GraphQL.NET we did not build OpenTelemetry support directly into the codebase but rather implemented it as a plugin.  This is not really the recommended approach, and I have no problem building the OpenTelemetry support directly into the codebase.  But to make it configurable, it would have required breaking API changes, which we were not prepared to do at this time.

I would endorse adding telemetry directly to the linq2db codebase.  So long as the source has not been enabled, it should not affect performance at all.  And if the source is enabled, we can assume the user wants the telemetry regardless of the performance impact.  Finally, if the telemetry is sampled, as is often the case, we will not be expending any excessive CPU time on data that will be discarded.

Shane32:
Another common implementation pattern is to provide a separate nuget package with an OpenTelemetry plugin.  For example, see the NuGet package [Npgsql.OpenTelemetry](https://www.nuget.org/packages/Npgsql.OpenTelemetry).  The plugin can contain the required dependency on [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource), making it compatible with .NET Framework or older .NET Core use cases.

We considered this for GraphQL.NET but figured that if someone wanted OpenTelemetry support for .NET Framework or .NET Core 3.1, they can just copy the 100 lines of code necessary to implement it (most of which are comments).  We may review this decision in the future.

igor-tkachev:
> Another common implementation pattern is to provide a separate nuget package with an OpenTelemetry plugin. For example, see the NuGet package [Npgsql.OpenTelemetry](https://www.nuget.org/packages/Npgsql.OpenTelemetry).

I thought about this solution and I like it more.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
jods4: @viceroypenguin Yeah, I've given some thoughts how calling sites could be simplified, too. I'm not fan of those `#if .. #endif` everywhere.

My idea was to use `[Conditional("METRICS")]` on the methods, but sadly I don't see a way to make it fully work with `using`.

`using` introduces a `try..finally` block and in my tests, even if they are empty they are not removed by compiler. It adds quite a few ASM instructions.

Not using `using` is an option, of course, but it is inconvenient when you want to time methods that have a `return` because you can't stop the timer after the return statement.

This is why I asked if we couldn't use AOP or some kind of profiler to instrument the code instead.
Ideas being:
- for AOP we'd just put an attribute on methods that must be timed, and that attribute could be `[Conditionnal("METRICS")]` so it wouldn't even exist outside tests.
- if we used a profiler we may indicate the methods we want to time with a filter/selector/list.

jods4:
@Shane32 if we introduce `DiagnosticSource` we'll need to give some thoughts what users may want vs what @igor-tkachev wants to benchmark. 
I don't think it makes a lot of sense to publicly expose all internal steps, although that's the granularity we need to benchmark linq2db internally.

So if we go `DiagnosticSource`, I assume we would create some public sources (for things like query compilation, execution, etc); and some private sources (for all the internal bits)?

Shane32:
> @Shane32 if we introduce `DiagnosticSource` we'll need to give some thoughts what users may want vs what @igor-tkachev wants to benchmark. I don't think it makes a lot of sense to publicly expose all internal steps, although that's the granularity we need to benchmark linq2db internally.
> 
> So if we go `DiagnosticSource`, I assume we would create some public sources (for things like query compilation, execution, etc); and some private sources (for all the internal bits)?

Offhand, I’d say distributed tracing isn’t designed for profiling. One concern I’d have is that it may be possible to enable all sources within OpenTelemetry. (I’m not sure). If so then end users would receive excessive tracing in their logs. 

Also, it’s possible that with the amount of tracing needed for profiling, the overhead for distributed tracing outweighs the cpu time in many of the short methods. (I’m not sure)

We can also consider the use of metrics instead of distributed tracing for this profiling data. 

Here is some MS links:
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/
- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/compare-metric-apis

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -5,6 +5,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using LinqToDB.Tools;
+
```
viceroypenguin, line 9: Move inside `namespace LinqToDB.Data` to match existing style.

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -543,9 +549,12 @@ await using (DataProvider.ExecuteScope(this))
...
 
-			var dr = result.HasValue
-				? result.Value
-				: await CurrentCommand!.ExecuteReaderAsync(commandBehavior, cancellationToken).ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
+			DbDataReader? dr;
+
+			using (ActivityService.Start(ActivityID.CommandExecuteReaderAsync))
+				dr = result.HasValue
+					? result.Value
+					: await CurrentCommand!.ExecuteReaderAsync(commandBehavior, cancellationToken).ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
```
viceroypenguin, line 557:
```suggestion
			using (ActivityService.Start(ActivityID.CommandExecuteReaderAsync))
			{
				dr = result.HasValue
					? result.Value
					: await CurrentCommand!.ExecuteReaderAsync(commandBehavior, cancellationToken).ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
			}
```

Source/LinqToDB/Data/DataConnection.QueryRunner.cs
```diff
@@ -10,6 +10,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using LinqToDB.Tools;
+
```
viceroypenguin, line 14: Move inside `namespace LinqToDB.Data` to match existing style.

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -830,13 +834,15 @@ public void DisposeCommand()
...
 
 			var result = _commandInterceptor.ExecuteNonQuery(new (this), command, Option<int>.None);
 
-			return result.HasValue
-				? result.Value
-				: command.ExecuteNonQuery();
+			using var m = ActivityService.Start(ActivityID.CommandExecuteNonQuery);
+				return result.HasValue
+					? result.Value
+					: command.ExecuteNonQuery();
```
viceroypenguin, line 845:
```suggestion
			using var m = ActivityService.Start(ActivityID.CommandExecuteNonQuery);
			{
				return result.HasValue
					? result.Value
					: command.ExecuteNonQuery();
			}
```

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -1049,9 +1057,12 @@ protected virtual DataReaderWrapper ExecuteReader(CommandBehavior commandBehavio
...
 
-			var rd = result.HasValue
-				? result.Value
-				: _command!.ExecuteReader(commandBehavior);
+			DbDataReader? rd;
+
+			using (ActivityService.Start(ActivityID.CommandExecuteReader))
+				rd = result.HasValue
+					? result.Value
+					: _command!.ExecuteReader(commandBehavior);
```
viceroypenguin, line 1065:
```suggestion
			using (ActivityService.Start(ActivityID.CommandExecuteReader))
			{
				rd = result.HasValue
					? result.Value
					: _command!.ExecuteReader(commandBehavior);
			}
```

Source/LinqToDB/Data/RetryPolicy/RetryingDbCommand.cs
```diff
@@ -4,6 +4,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using LinqToDB.Tools;
+
```
viceroypenguin, line 8: Move inside `namespace LinqToDB.Data.RetryPolicy` to match existing style.

Source/LinqToDB/Linq/ExpressionQuery.cs
```diff
@@ -7,6 +7,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using LinqToDB.Tools;
+
```
viceroypenguin, line 11: Move inside `namespace LinqToDB.Linq` to match existing style.

Source/LinqToDB/Linq/QueryRunner.cs
```diff
@@ -10,6 +10,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using LinqToDB.Tools;
+
```
viceroypenguin, line 14: Move inside `namespace LinqToDB.Linq` to match existing style.

Source/LinqToDB/Tools/ActivityService.cs
```diff
@@ -0,0 +1,19 @@
...
+	public static class ActivityService
+	{
+		public static IActivity? Start(ActivityID metric)
+		{
+			return _factory?.Invoke(metric);
+		}
+
+		static Func<ActivityID,IActivity>? _factory;
+
+		public static void SetFactory(Func<ActivityID,IActivity>? factory)
```
viceroypenguin, line 14: Is this for public to access, or just internal testing?

Tests/Base/TestBase.cs
viceroypenguin, line 1: Perhaps should be in same PR as `BaselinesWriter.cs` change

Tests/Linq/Linq/WhereTests.cs
```diff
@@ -1836,7 +1836,14 @@ sealed class WhereWithString
 		[Test]
 		public void CaseOptimization([DataSources] string context)
 		{
-			using (var db = GetDataContext(context))
+			List<int>? list = null;
+
+			if (list is [..])
```
viceroypenguin, line 1841: Dead code, or purpose?

viceroypenguin (changes requested):
I have serious concerns about the changes in linq2db.baselines. The data in Metrics.txt is too precise - every PR will now have 3k+ files changed due to the metrics.txt data changing. Perhaps we can reduce the level of precision to only 3-4 digits on each timestamp, so that the normal variances of processing don't create enormous PRs?

igor-tkachev:
Yes, I do not like it either. I think we can keep only one or two metrics for SqlServer and some other configuration.

Source/LinqToDB/Tools/ActivityService.cs
igor-tkachev: public

Source/LinqToDB.Tools/Activity/ActivityHierarchy.cs
```diff
@@ -0,0 +1,80 @@
...
+						_parent._children.Add(this);
+				}
+			}
+			else if (!(activityID is ActivityID.QueryProviderExecuteT or ActivityID.QueryProviderGetEnumeratorT))
+			{
+			}
+		}
+
+#pragma warning disable RS0030
+		[ThreadStatic]
```
jods4, line 47: Do we have a guarantee the metrics code will only ever be used on synchronous code?

If someone attempts to benchmark an async operation (which would look simple enough with your `using` approach), this shouldbe `AsyncLocal`.

Source/LinqToDB.Tools/Activity/StatActivity.cs
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		void Stop(Stopwatch stopwatch)
+		{
+			Interlocked.Add(ref _elapsedTicks, stopwatch.ElapsedTicks);
+		}
+
+		void Stop(TimeSpan time)
+		{
+			lock (this)
```
jods4, line 37: Why two storages `_elapsed` and `_elapsedTicks` for a single property `Elapsed`? 
Why the lock when everything else is lock-free with `Interlocked`? 

`TimeSpan` has a `Ticks` property, can't you reuse `Interlocked.Add(ref _elapsedTicks, stopwatch.ElapsedTicks)`?
Given the `StatActivity` is used across threads, it also feels like there could be a bug in mixing calls to `Stop(Stopwatch)` and `Stop(Timestamp)` as they don't share the same tally field, and only one is exposed by `Elapsed`.

Source/LinqToDB/Tools/ActivityService.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+
+namespace LinqToDB.Tools
+{
+	public static class ActivityService
+	{
+		public static IActivity? Start(ActivityID activityID)
+		{
```
jods4, line 8: I like the idea of erasing the metrics code as much as possible from our published build.
Should we have a compilation symbol `METRICS` and when `#if !METRICS` simplify this to:
```cs
public static IActivity? Start(ActivityID _) => null;
```

I know your code also returns `null` when there's no factory registered, but the JIT compiler cannot know that.

With simple `return null` implementation, this will get inlined, the JIT will still the using is always null so no Dispose call required, etc. There will be practically no overhead left.

Compare in this [SharpLab](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQAUEEsC2UECeAwgPYB2yMCcAxjCQsgDQwhTI4A+AAgEwCMAWABQI7gGYABFjIwICMFBoRJASQCCdLADcsMApJBqAIlmQAHEsijAANhBEBvAL4ixUvpM0wdeg0Y0tXX0nEUlwyQlpWTUAE0lHSQBzCBgAbklXYQjJMIio7199AAoZGGlYgEo88McanLjJAF4KtPqs+qjuABZJUwsrCGLKhKyO4S7+ADZI3i8gvwBleV1lUOz8j2m1QuCCAH5JRZh8GFKYpR891SrmgD5JMjhbWzaN8M6tmcCrv0OAWVS2Bo53KlyKBBuIyaDzIEAA7vNfiVwdcqm9xl05tx+AB2XLCOrvSIeXoAFRQMAAojhzCVqsSiQ18gAGSTafCSAD6zSREOWCFWEAAdMdTsVurxKm8clkcp9IuTKQB5YAAKwgdGG9SZzMibI5CG5vN2SxWWGUwsBVAtEqlMoiYyAA) the ASM disassembly of `StartEmpty` vs `StartObject`.

Source/LinqToDB/Tools/IActivity.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace LinqToDB.Tools
+{
+	public interface IActivity : IDisposable
```
jods4, line 5: Do we need this interface? It has no method, so `ActivityService.Start()` could just as well return `IDisposable`.

Source/LinqToDB.Tools/Activity/ActivityHierarchy.cs
igor-tkachev: accepted

Source/LinqToDB.Tools/Activity/StatActivity.cs
igor-tkachev: `Stopwatch` Linux version does not work correctly. It shows completely insane values.

Source/LinqToDB/Tools/IActivity.cs
igor-tkachev: I am sure 999% we will extend it.

Source/LinqToDB/Tools/ActivityService.cs
igor-tkachev: I did some optimizations, but I cant do it in this way, as I already need this functionality for my work project.

Source/LinqToDB.Tools/Activity/StatActivity.cs
jods4: I undertstood that. I suggest you rewrite this overload like so:
```cs
void Stop(TimeSpan time) => Interlocked.Add(ref _elapsedTicks, time.Ticks);`
```
If you do that, you don't need locking anymore and you can remove `_elapsed` field and simplify `Elapsed` property.

Source/LinqToDB.Tools/Activity/ActivityHierarchy.cs
```diff
@@ -0,0 +1,91 @@
...
+				}
+			}
+			else if (!(activityID is ActivityID.QueryProviderExecuteT or ActivityID.QueryProviderGetEnumeratorT))
+			{
+			}
+		}
+
+#if NET45
+#pragma warning disable RS0030
+		// TODO: Change to AsyncLocal<T> after removing FW 4.5 support.
```
MaceWindu, line 48: Copy `AsyncLocal` `net45` implementation from `Compatibility` folder in `linq2db` project

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,375 @@
...
+				m.Name,
+				m.Elapsed,
+				m.CallCount,
+				TimePerCall = m.CallCount switch
+				{
+					0 => TimeSpan.Zero,
+					1 => m.Elapsed,
+					_ => new TimeSpan(m.Elapsed.Ticks / m.CallCount)
+				},
+				Percent = FixUnixPercentFormat($"{(m.CallCount == 0 ? (decimal?)null : m.Elapsed.Ticks / totalTime),7:P}")
```
MaceWindu, line 270: is it needed with invariant culture formatting?

Source/LinqToDB/Async/AsyncDbConnection.cs
```diff
@@ -53,55 +55,102 @@ public virtual string ConnectionString
...
-		public virtual Task OpenAsync(CancellationToken cancellationToken) => Connection.OpenAsync(cancellationToken);
+		public virtual void Open()
+		{
+			using var _ = ActivityService.Start(ActivityID.ConnectionOpen);
+			Connection.Open();
+		}
+
+		public virtual Task OpenAsync(CancellationToken cancellationToken)
+		{
+			using var _ = ActivityService.Start(ActivityID.ConnectionOpenAsync);
```
MaceWindu, line 66: here and in similar places: does it have any value to collect metrics for non-awaited tasks?

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -191,14 +201,23 @@ public Query<T> Build<T>()
...
+			using var m = ActivityService.Start(ActivityID.BuildSequence);
+
 			buildInfo.Expression = buildInfo.Expression.Unwrap();
 
 			var n = _builders[0].BuildCounter;
 
 			foreach (var builder in _builders)
 			{
-				if (builder.CanBuild(this, buildInfo))
+				var mc       = ActivityService.Start(ActivityID.BuildSequenceCanBuild);
```
MaceWindu, line 212: I would prefer to use `using`/`finally` here and in several similar places. `IActivity` implementation provided by user with unknown disposal logic.

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -378,6 +379,8 @@ protected virtual async Task<int> ExecuteNonQueryAsync(CancellationToken cancell
 				result = await _commandInterceptor.ExecuteNonQueryAsync(new (this), CurrentCommand!, result, cancellationToken)
 					.ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
 
+			using var m = ActivityService.Start(ActivityID.CommandExecuteNonQueryAsync);
```
MaceWindu, line 382: is it correct in some other places we start activity only for `Execute` branch for ternary expression

MaceWindu (changes requested):

Tests/Base/Tests.Base.csproj
```diff
@@ -3,6 +3,7 @@
 	<PropertyGroup>
 		<AssemblyName>linq2db.Tests.Base</AssemblyName>
 		<RootNamespace>Tests</RootNamespace>
+		<LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
igor-tkachev: fixed

Source/LinqToDB/Tools/IActivity.cs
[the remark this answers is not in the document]
igor-tkachev: fixed

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
igor-tkachev: yes, different behavior for Win and Lin

Source/LinqToDB.Tools/Activity/ActivityHierarchy.cs
```diff
@@ -0,0 +1,84 @@
...
+				{
+					var p = _parent._children[^1];
+
+					if (p._activityID == _activityID)
+						p._count++;
+					else
+						_parent._children.Add(this);
+				}
+			}
+			else if (activityID is not (ActivityID.QueryProviderExecuteT or ActivityID.QueryProviderGetEnumeratorT))
```
MaceWindu, line 42: not needed branch/missing code?

Source/LinqToDB.Tools/Activity/ActivityHierarchy.cs
```diff
@@ -0,0 +1,84 @@
...
+						.Append(' ', indent)
+						.Append(a._name);
+
+					if (a._count > 1)
+						sb.Append($" ({a._count})");
+
+					sb.AppendLine();
+
+					foreach (var c in a._children)
+						Print(c, indent + 2);
```
MaceWindu, line 79: maybe allow user to provide indent size?

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
MaceWindu, line 1: xml-docs missing on public api

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,379 @@
...
+				GetIEnumerable                      = new("  GetIEnumerable"),
+
+				ExecuteTotal                        = new("Execute",
+					ExecuteQuery                    = new("  Execute Query"),
+					ExecuteQueryAsync               = new("  Execute Query Async"),
+					ExecuteElement                  = new("  Execute Element"),
+					ExecuteElementAsync             = new("  Execute Element Async"),
+					ExecuteScalar                   = new("  Execute Scalar"),
+					ExecuteScalarAsync              = new("  Execute Scalar Async"),
+					ExecuteScalar2                  = new("  Execute Scalar 2"),
```
MaceWindu, line 37: `2` here in in other places is not clear. Could we improve those names somehow (or clarify in corresponding member xml-doc)?

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,379 @@
...
+				CommandExecuteReaderAsync,
+				CommandExecuteNonQuery,
+				CommandExecuteNonQueryAsync,
+
+				OnTraceInternal                     = new("    OnTraceInternal"),
+				Materialization                     = new("    Materialization"),
+
+				GetSqlText                          = new("  GetSqlText"),
+
+				null!
```
MaceWindu, line 153: maybe add `// Total, set below` comment

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,379 @@
...
+				ActivityID.ConnectionDispose               => ConnectionDispose,
+				ActivityID.ConnectionDisposeAsync          => ConnectionDisposeAsync,
+				ActivityID.ConnectionBeginTransaction      => ConnectionBeginTransaction,
+				ActivityID.ConnectionBeginTransactionAsync => ConnectionBeginTransactionAsync,
+				ActivityID.TransactionCommit               => TransactionCommit,
+				ActivityID.TransactionCommitAsync          => TransactionCommitAsync,
+				ActivityID.TransactionRollback             => TransactionRollback,
+				ActivityID.TransactionRollbackAsync        => TransactionRollbackAsync,
+				ActivityID.TransactionDispose              => TransactionCommit,
+				ActivityID.TransactionDisposeAsync         => TransactionCommitAsync,
```
MaceWindu, line 233: wrong return values

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,379 @@
...
+				{
+					0 => TimeSpan.Zero,
+					1 => m.Elapsed,
+					_ => new TimeSpan(m.Elapsed.Ticks / m.CallCount)
+				},
+				Percent = FixUnixPercentFormat($"{(m.CallCount == 0 ? null : m.Elapsed.Ticks / totalTime),7:P}")
+			})
+			.ToDiagnosticString();
+
+			static string FixUnixPercentFormat(string str)
```
MaceWindu, line 277: maybe we shouldn't mess with system-provided format? It depends on used system culture and could be overrided by user in both windows/linux

MaceWindu (changes requested):
Added some review comments, will finish review tomorrow

Source/LinqToDB/Tools/IActivity.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+using JetBrains.Annotations;
+
+namespace LinqToDB.Tools
+{
+	/// <summary>
+	/// Represents a user defined operation with context to be used for Activity Service events.
```
MaceWindu, line 8: `user-defined`

Source/LinqToDB/Tools/ActivityService.cs
```diff
@@ -0,0 +1,99 @@
...
+	/// Provides API to register factory methods that return an Activity object or <c>null</c> for provided <see cref="ActivityID"/> event.
+	/// </summary>
+	[PublicAPI]
+	public static class ActivityService
+	{
+		internal static Func<ActivityID,IActivity?> Start { [MethodImpl(MethodImplOptions.AggressiveInlining)] get; private set; } = static _ => null;
+
+		static IActivity? StartImpl(ActivityID activityID)
+		{
+			var factory = _factory!;
```
MaceWindu, line 19: I would prefer if we handle null here and return null/throw explicit exception

Source/LinqToDB/AsyncExtensions.cs
```diff
@@ -374,5 +375,16 @@ async IAsyncEnumerator<T> IAsyncEnumerable<T>.GetAsyncEnumerator(CancellationTok
 
 		#endregion
 
+#if NATIVE_ASYNC
```
MaceWindu, line 378: 👍

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -378,9 +379,12 @@ protected virtual async Task<int> ExecuteNonQueryAsync(CancellationToken cancell
 				result = await _commandInterceptor.ExecuteNonQueryAsync(new (this), CurrentCommand!, result, cancellationToken)
 					.ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
 
-			return result.HasValue
-				? result.Value
-				: await CurrentCommand!.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(Common.Configuration.ContinueOnCapturedContext);
+			await using (ActivityService.StartAndConfigureAwait(ActivityID.CommandExecuteNonQueryAsync))
```
MaceWindu, line 382: Interceptors case is interesting. Should we include interceptor call into activity scope or just ignore time spent it it eve if interceptor provides result value like it is implemented now?

Source/LinqToDB/ITableT.cs
```diff
@@ -10,6 +10,7 @@ namespace LinqToDB
 	/// <typeparam name="T">Record mapping type.</typeparam>
 	[PublicAPI]
 	public interface ITable<out T> : IExpressionQuery<T>
+		// TODO: IT: WTF? it should be 'class'.
```
MaceWindu, line 13: we can review it in v6, not earlier

Source/LinqToDB/Linq/QueryRunner.cs
```diff
@@ -78,6 +79,8 @@ public sealed class ReaderMapperInfo
 
 			public T Map(IDataContext context, IQueryRunner queryRunner, DbDataReader dataReader, ref ReaderMapperInfo mapperInfo)
 			{
+				var a = Configuration.TraceMaterializationActivity ? ActivityService.Start(ActivityID.Materialization) : null;
```
MaceWindu, line 82: Could you create tracing article in docs repo based on examples doc and incude information about new configuration options?

MaceWindu (changes requested):
remaining part review

MaceWindu (dismissed):

Source/LinqToDB.Tools/Activity/StatActivity.cs
viceroypenguin: @igor-tkachev ping to fix this. I agree, you can use an `Interlocked.Add()` here for better behavior.

Source/LinqToDB/Async/AsyncDbConnection.cs
```diff
@@ -53,71 +55,159 @@ public virtual string ConnectionString
...
+			var a = ActivityService.StartAndConfigureAwait(ActivityID.ConnectionOpenAsync);
+
+			if (a is null)
+				return Connection.OpenAsync(cancellationToken);
+
+			return CallAwaitUsing(a, Connection, cancellationToken);
+
+			static async Task CallAwaitUsing(IAsyncDisposableEx activity, DbConnection connection, CancellationToken token)
+			{
+				await using (activity)
```
viceroypenguin, line 75: All `await using` should use `await using (activity.ConfigureAwait(Common.Configuration.ContinueOnCapturedContext))`. This will match existing and correct behavior regarding returning to correct async context.

Source/LinqToDB/Tools/ActivityService.cs
```diff
@@ -0,0 +1,102 @@
...
+			var activities = new IActivity?[list.Length];
+
+			for (var i = 0; i < list.Length; i++)
+				activities[i] = ((Func<ActivityID, IActivity>)list[i])(activityID);
+
+			return new MultiActivity(activities);
+		}
+
+#if NATIVE_ASYNC
+		sealed class AsyncDisposableWrapper(ConfiguredAsyncDisposable configured) : IAsyncDisposable
```
viceroypenguin, line 38: This wraps a configured disposable, but removes the vital information necessary for the calling code to configure properly. If the calling code `await`s on an `AsyncDisposableWrapper`, then the awaiter that they wait on is the state machine in `DisposeAsync()` instead of the configured awaiter that informs the continuation to skip the context reset.

igor-tkachev: Yeh. `await using` probably the stupidest part of C# design.

Source/LinqToDB.Tools/Activity/StatActivity.cs
```diff
@@ -0,0 +1,66 @@
...
+			Interlocked.Add(ref _elapsedTicks, stopwatch.ElapsedTicks);
+		}
+
+		void Stop(TimeSpan time)
+		{
+#pragma warning disable CA2002 // Do not lock on objects with weak identity
+			lock (this)
+				_elapsed += time;
+#pragma warning restore CA2002 // Do not lock on objects with weak identity
+		}
```
viceroypenguin, line 36:
```suggestion
		void Stop(TimeSpan time) 
			=> Interlocked.Add(ref _elapsedTicks, time.Ticks);
```

viceroypenguin (changes requested):
last change before my approval.

Source/LinqToDB.Tools/Activity/StatActivity.cs
igor-tkachev: Lets see if works for Linux.

viceroypenguin (dismissed):

Source/LinqToDB.Tools/Activity/ActivityStatistics.cs
```diff
@@ -0,0 +1,518 @@
...
+		static StatActivitySum Total;
+
+		static StatActivitySum Interceptors;
+		static StatActivitySum CommandInterceptor;
+		static StatActivitySum ConnectionInterceptor;
+		static StatActivitySum DataContextInterceptor;
+		static StatActivitySum EntityServiceInterceptor;
+		static StatActivitySum UnwrapDataObjectInterceptor;
+
+		static IStatActivity[] All;
```
sdanyliv, line 516: Why not mark everything as read-only?

sdanyliv (dismissed):
Approving, **squash and merge** but you have to resolve conflicts with `refactor/parsing` branch.

MaceWindu (approved):

viceroypenguin (approved):

Issue #4009, opened 2023-03-03
Update with xmin doesn't use in sql query 


twilightrus:
I have this code:

```
var transfer = await connection.Transfers.FirstAsync();

var results = await connection.Transfers
                .Where(x => x.xmin == transfer.xmin && transfer.RecipientAddress == "#account-1")
                .Set(x => x.Amount, 300)
                .UpdateAsync();
```

Resulting query:
```
DECLARE @Amount Numeric -- Decimal
      SET     @Amount = 300

      UPDATE
        "Transfer"
      SET
        "Amount" = :Amount
      WHERE
        1 = 0
```

xmin has value after FirstAsync(), but on update this ignored

twilightrus:
I want to implement optimistic concurrency using xmin. Is there a way to do this out of the box?

MaceWindu:
1. you can mark your xmin column with
```cs
[OptimisticLockProperty(VersionBehavior.Auto)]
```

and use update/delete extensions for optimistic lock https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#optimistic-lock-extensions

2. I need to see your `Transfers` mappings to understand why such filter generated.

twilightrus:
I have more questions:

1. Is it possible to configure Optimistic Concurrency with fluent instead of an attribute?

2. UpdateOptimisticAsync() does not make it possible to set only the required fields for updating and updates everything. Can the fields be set manually?

MaceWindu:
> Is it possible to configure Optimistic Concurrency with fluent instead of an attribute?

use `HasAttribute` fluent api method

twilightrus:
> > Is it possible to configure Optimistic Concurrency with fluent instead of an attribute?
> 
> use `HasAttribute` fluent api method

Okay, what about setting only certain fields on UpdateOptimisticAsync?

MaceWindu:
There is no such API currently. I could add something like `IQueryable<T> OptimisticFilter(this IQueryable<T> source, T record)` so you could combine it with regular `Update` API

twilightrus:
> There is no such API currently. I could add something like `IQueryable<T> OptimisticFilter(this IQueryable<T> source, T record)` so you could combine it with regular `Update` API

It would be nice. Will wait!
Thanks

PR #4015, opened 2023-03-06
add OptimisticFilter extension

MaceWindu:
Requested here #4009

leverage use of optimistic lock with linq api

### DIFF
--- a/Build/BannedSymbols.txt
+++ b/Build/BannedSymbols.txt
@@ -77,3 +77,11 @@ M:System.Reflection.Module.GetCustomAttributes(System.Boolean); Use AttributesEx
 M:System.Reflection.Module.GetCustomAttributes(System.Type,System.Boolean); Use AttributesExtensions.GetAttributes<TAttribute>(bool inherit) extension
 M:System.Reflection.ParameterInfo.GetCustomAttributes(System.Boolean); Use AttributesExtensions.GetAttributes<TAttribute>(bool inherit) extension
 M:System.Reflection.ParameterInfo.GetCustomAttributes(System.Type,System.Boolean); Use AttributesExtensions.GetAttributes<TAttribute>(bool inherit) extension
+M:System.Attribute.IsDefined(System.Reflection.Assembly,System.Type); Use AttributesExtensions.HasAttribute<TAttribute>(bool inherit) extension
+M:System.Attribute.IsDefined(System.Reflection.Assembly,System.Type,System.Boolean); Use AttributesExtensions.HasAttribute<TAttribute>(bool inherit) extension
+M:System.Attribute.IsDefined(System.Reflection.MemberInfo,System.Type); Use AttributesExtensions.HasAttribute<TAttribute>(bool inherit) extension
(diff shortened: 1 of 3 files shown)


jods4:
Bikeshedding: I'm not a fan of the name.
Before I looked at the code, it didn't cross my mind that it included a WHERE on the PK.

Do we have an API to look up objects by PK?
Given we have `Update`, `UpdateOptimistic` and `Delete`, `DeleteOptimistic`; it would make sense to name theses `Find` and `FindOptimistic` (assuming we have a `Find` that adds a WHERE by PK).

`Find` if the method doesn't exist yet, is also up for discussion, it's not a great name...

jods4:
I think there are opportunities for refactoring here.

Every call to `MakeConcurrentFilter` (there are 3) is preceded by:
```csharp
var objType = typeof(T);
var ed      = dc.MappingSchema.GetEntityDescriptor(objType, dc.Options.ConnectionOptions.OnEntityDescriptorCreated);
```
Then `ed` is passed is last parameter of `MakeConcurrentFilter`.

-> This code should be in `MakeConcurrentFilter`, which should take the `dc` instead of `ed`.

Once that change is made, there's so little in Delete/Update methods that they can be rewritten as shorthands for what #4009 asked:
```csharp
public static int DeleteOptimistic<T>(this IQueryable<T> source, T obj)
  where T : class
  => source.FindOptimistic(obj).Delete();
```

(I also removed the null checks, but those will be done by `FindOptimistic`)

MaceWindu:
both variants supported https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB.Tools/DataModel/Model/Entities/FindTypes.cs

MaceWindu:
@jods4 so, what we are doing here (except optimization)? Should we change name?

MaceWindu:
Key could be composite and I really don't want to invest too many time into this feature :)

jods4:
@MaceWindu yes, I thought about that and figured: if your key is composite, just use the entity syntax or an explicit `Where`.

If you don't want to invest too much time, I don't think we need to implement everything right now.
We just need to be sure we're ok with naming and that it would be consistent if we expand the API later.

Personally, I think `WhereKeyOptimistic` is my favorite name so far.

The rest could be added later, and I think we can make `WhereKey<T>` work with both entities and values.
For a start, it would be 2 different overloads: `WhereKey<T>(this IQueryable<T> src, T entity)` and `WhereKey<T, K>(this IQueryable<T> src, K key)`.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -241,9 +241,7 @@ public static int DeleteOptimistic<T>(this IQueryable<T> source, T obj)
 			if (source == null) throw new ArgumentNullException(nameof(source));
 			if (obj    == null) throw new ArgumentNullException(nameof(obj));
```
jods4, line 243: @MaceWindu For simple methods that simply forward a call to another, which will then null-check the same parameters, don't you think it's ok to remove the checks?

jods4 (approved):

PR #4016, opened 2023-03-06
Do not remove LEFT JOIN if there is table hint.

sdanyliv:
Wrong JOIN optimization with Temporal Tables.

### DIFF
--- a/Source/LinqToDB/SqlProvider/JoinOptimizer.cs
+++ b/Source/LinqToDB/SqlProvider/JoinOptimizer.cs
@@ -1228,6 +1228,9 @@ namespace LinqToDB.SqlProvider
 			if (join.JoinType != JoinType.Left)
 				return false;
 
+			if (join.Table.Source is SqlTable { SqlQueryExtensions.Count: > 0 })
+				return false;
+
 			var found = SearchForFields(null, join);
 
 			if (found == null)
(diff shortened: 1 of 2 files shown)


viceroypenguin (approved):

jods4:
@sdanyliv out of curiosity, what's the underlying issue?
I don't see why it would be wrong to remove the LEFT JOIN in the test case you created?

Is it a cardinality problem with temporal ranges (your test case uses AS OF, so only it should return only 1 row)?

MaceWindu (approved):

sdanyliv:
@jods4. with temporal tables here can be duplicates of Primary Key. Removing joins is not an option in that case. We can simplify restriction for `TemporalTableAsOf`, but not for `TemporalTableAll`. Decided to disable optimization for any hints right now.

MaceWindu:
yep, join with temporal table could produce multiple records (not in this test)

sdanyliv:
It is question to current hint implementation. I've just noticed that we have issue with this one and in production it may lead in problems. Since I have no time to work with this right now - disabled optimization for tables with hints. Will add comment about that when tests are passed.

jods4:
@sdanyliv I see this was merged, please don't forget to add a comment so we remember why this was added and could potentially refactor it in the future.

jods4:
@sdanyliv Understood, no problem if you need a fix for a bug in production.

Just for long-term maintenance I think we need to remember to add a test that would fail if this is removed and a comment to explain for what specific case we keep LEFT JOIN with hints.

PR #4021, opened 2023-03-09
Fluent mappings generator

### LINKED ISSUE #945: [Scaffold] Generate mappings using fluent mapper
Hi,
I did not found T4-template for generation models with fluent pattern for MsSql. Does such exist? if not, can you help to create it

MaceWindu:
Fix #945

- add new scaffold option to customize metadata generation: `--metadata` option with values: `none`, `attributes`, `fluent`
- automatically pass generated mapping schema to context in generated context constructors (for fluent mapping and `PostgreSQL` tuples mapping)
- [PostgreSQL] fix error when scaffold includes `NpgsqlTypes.NpgsqlInterval` type
- [SQL CE] fix duplicate foreign key columns #695 (looks like it wasn't fixed properly before)
- fixed equality generation for nullable `System.Data.SqlTypes.Sql*` struct types and `SqlHierarchyId`
- fixed equality generation for DB2 custom struct types (e.g. `DB2Int32`) including nullable types
- fixed defaults rendering in `--find-methods` command help, add `none` option to disable generation of `Find` methods

Also refactored data model generator to use main/additional schema build context instead of passing data through set of parameters, fields and callbacks...


MaceWindu:
generation example https://github.com/linq2db/linq2db/blob/8a3977730a1dcb3bddd7fcfd25e5a92dbcc9d21c/Tests/Tests.T4/Cli/Fluent/PostgreSQL/TestDataDB.cs

PR #4023, opened 2023-03-09
Fix test dependencies TFMs

MaceWindu:
because linq2db referenced from tests directly and as transitive dependency of other projects we end up with wrong linq2db binary in tests on build causing errors like this:

```
OneTimeSetUp: System.MissingMethodException : Method not found: 'LinqToDB.DataProvider.ClickHouse.ClickHouseOptions LinqToDB.DataProvider.ClickHouse.ClickHouseOptions.<Clone>$()'.
```


MaceWindu:
It will not harm to add those TFMs to nugets, but I want to postpone it till v6 where we will cleanup old TFMs

sdanyliv (dismissed):

viceroypenguin (approved):

PR #4036, opened 2023-03-18
Improve interfaces support

### LINKED ISSUE #4031: Expression 'c.Id' is not a Field.' for derived DBO with separate base assembly
### Describe your issue

I have a group of tables with similar data structure. DBOs for them are placed in a shared assembly. To write single method that modifies any of that tables, I've created an interface, then derived class from both DBO and that interface, then made a generic that modifies data through that interface. If DBO, interface and derived class are placed in the same assembly, everything works ok. If they are located in different assemblies, exception is thrown when trying to build update query string.

```
Unhandled exception. LinqToDB.Linq.LinqException: Expression 'c.Id' is not a Field.
   at LinqToDB.Linq.Builder.TableBuilder.TableContext.ConvertToSql(Expression expression, Int32 level, ConvertFlags flags)
   at LinqToDB.Linq.Builder.TableBuilder.TableContext.ConvertToIndex(Expression expression, Int32 level, ConvertFlags flags)
   at LinqToDB.Linq.Builder.SubQueryContext.ConvertToSql(Expression expression, Int32 level, ConvertFlags flags)
   at LinqToDB.Linq.Builder.ExpressionBuilder.ConvertToSql(IBuildContext context, Expression expression, Boolean unwrap, ColumnDescriptor columnDescriptor, Boolean isPureExpression)
   at LinqToDB.Linq.Builder.ExpressionBuilder.ConvertCompare(IBuildContext context, ExpressionType nodeType, Expression left, Expression right)
   at LinqToDB.Linq.Builder.ExpressionBuilder.ConvertPredicate(IBuildContext context, Expression expression)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSearchCondition(IBuildContext context, Expression expression, List`1 conditions)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildWhere(IBuildContext parent, IBuildContext sequence, LambdaExpression condition, Boolean checkForSubQuery, Boolean enforceHaving)
   at LinqToDB.Linq.Builder.WhereBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.UpdateBuilder.Set.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.UpdateBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Linq.Builder.ExpressionBuilder.Build[T]()
   at LinqToDB.Linq.Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, Expression expr)
   at LinqToDB.Linq.Query`1.GetQuery(IDataContext dataContext, Expression& expr, Boolean& dependsOnParameters)
   at LinqToDB.Linq.ExpressionQuery`1.GetQuery(Expression& expression, Boolean cache, Boolean& dependsOnParameters)
   at LinqToDB.Linq.ExpressionQuery`1.System.Linq.IQueryProvider.Execute[TResult](Expression expression)
   at LinqToDB.LinqExtensions.Update[T](IUpdatable`1 source)
   at Bad.Program.Confirm[T](Int32 id) in D:\TestCase\Bad\Program.cs:line 20
   at Bad.Program.Main(String[] args) in D:\TestCase\Bad\Program.cs:line 14
```

### Steps to reproduce

Here is sample solution
https://disk.yandex.ru/d/D7MlU_8FHMbaIw

Project Good contains DBO class 
```c#
    [Table("Test")]
    class MyItem
    {
        [Column]
        [PrimaryKey]
        public int Id { get; set; }

        [Column]
        public bool Ok { get; set; }
    }
```

interface 
```c#
    interface IConfirmable
    {
        int Id { get; set; }
        bool Ok { get; set; }
    }
```

and derived class 
```c#
    class MyItemConfirmable : MyItem, IConfirmable
    {
    }
```

Also here is generic function 

```c#
        {
            DataConnection.AddOrSetConfiguration("default", "Server=.;Database=Ts1;Trusted_Connection=True;encrypt=false", "SqlServer");
            Confirm<MyItemConfirmable>(5);

        }
        static void Confirm<T>(int id) where T : class, IConfirmable
        {
            using var db = new DataConnection("default");
            db.GetTable<T>().Where(c => c.Id == id).
                Set(c => c.Ok, true)
                .Update();
            Console.WriteLine(db.LastQuery);
        }
```
Everything works as expected.

Now, project Bad contains derved MyItemConfirmable only and links to project Shared, containing MyItem, IConfirmable. When invoking Update in generic function here, exception is thrown.

### Environment details

`Linq To DB` version: 5.0.0

Database (with version): SQL Server 2017

ADO.NET Provider (with version): SqlServer

Operating system: Windows 10

.NET Version: .Net 5

MaceWindu:
Fix #4031

Original issue caused by C# compiler adding explicit getter/setter implementations without property implementation to class inheriting implementation from base class, located in another assembly.
I've added fix, but not happy with it so will appreciate better heuristics proposed.

Additional tests showed that we also doesn't handle inherited explicit interface implementations, so it makes sense to try and fix it too.

Additionally, added guard against static interface members

### DIFF
--- a/Source/LinqToDB/Reflection/TypeAccessorT.cs
+++ b/Source/LinqToDB/Reflection/TypeAccessorT.cs
@@ -51,8 +51,8 @@ namespace LinqToDB.Reflection
 			else
 			{
 				// load properties takin into account explicit interface implementations
-				var interfacePropertiesList = new List<PropertyInfo>();
-				var interfaceProperties     = new HashSet<(Type? declaringType, string name, Type type)>();
+				var interfacePropertiesList = new List<PropertyInfo?>();
+				var interfaceProperties     = new Dictionary<(Type? declaringType, string name, Type type), int>();
 
 				// as interface could be re-implemented multiple times, we should track which inteface property accessors
 				// are not mapped yet and reduce this list walking inheritance hierarchy from top to base type
@@ -100,8 +100,8 @@ namespace LinqToDB.Reflection
 							if ((pi.GetMethod == null || (methods.TryGetValue(pi.GetMethod, out var ifaceGetters) && RemoveAll(unmappedAccessors, ifaceGetters))) &&
 								(pi.SetMethod == null || (methods.TryGetValue(pi.SetMethod, out var ifaceSetters) && RemoveAll(unmappedAccessors, ifaceSetters))))
 							{
+								interfaceProperties.Add((pi.DeclaringType, pi.Name, pi.PropertyType), interfacePropertiesList.Count);
 								interfacePropertiesList.Add(pi);
-								interfaceProperties.Add((pi.DeclaringType, pi.Name, pi.PropertyType));
 							}
 						}
 					}
@@ -113,10 +113,16 @@ namespace LinqToDB.Reflection
 				}
 
 				foreach (var mi in type.GetPublicInstanceValueMembers())
-					if (mi is not PropertyInfo pi || !interfaceProperties.Contains((pi.DeclaringType, pi.Name, pi.PropertyType)))
-						_members.Add(mi);
+				{
+					if (mi is PropertyInfo pi && interfaceProperties.TryGetValue((pi.DeclaringType, pi.Name, pi.PropertyType), out var idx))
+						interfacePropertiesList[idx] = null;
 
-				_members.AddRange(interfacePropertiesList);
+					_members.Add(mi);
+				}
+
+				foreach (var pi in interfacePropertiesList)
+					if (pi != null)
+						_members.Add(pi);
 			}
 
 			// ObjectFactory
--- a/Source/LinqToDB/Reflection/TypeAccessorT.cs
+++ b/Source/LinqToDB/Reflection/TypeAccessorT.cs
@@ -30,7 +30,7 @@ namespace LinqToDB.Reflection
 					Expression<Func<T>> mi;
 
 					if (type.IsAbstract) mi = () => ThrowAbstractException();
-					else                     mi = () => ThrowException();
+					else                 mi = () => ThrowException();
 
 					var body = Expression.Call(null, ((MethodCallExpression)mi.Body).Method);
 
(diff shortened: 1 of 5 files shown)


MaceWindu:
Cannot say I'm happy with it, but we are limited by current reflection API capabilities and at least it improves situation compared to older releases.

Also because public members discovery logic for types with interfaces changed, column order in queries also changed for entities that mix interface/non-interface properties. If you think it's important to preserve old order, I could try to look into it.

Shane32:
> Also because public members discovery logic for types with interfaces changed, column order in queries also changed for entities that mix interface/non-interface properties. If you think it's important to preserve old order, I could try to look into it.

I believe I do mix interface/non-interface properties, but the order does not matter to me.  Sure it might change a few tests (as I compare the generated SQL in certain circumstances), but that's not an issue.

Shane32 (approved):

igor-tkachev (approved):

Source/LinqToDB/Reflection/TypeAccessorT.cs
```diff
@@ -113,10 +113,16 @@ static TypeAccessor()
 				}
 
 				foreach (var mi in type.GetPublicInstanceValueMembers())
-					if (mi is not PropertyInfo pi || !interfaceProperties.Contains((pi.DeclaringType, pi.Name, pi.PropertyType)))
-						_members.Add(mi);
+				{
+					if (mi is PropertyInfo pi && interfaceProperties.TryGetValue((pi.DeclaringType, pi.Name, pi.PropertyType), out var idx))
+						interfacePropertiesList[idx] = null;
```
viceroypenguin, line 118: why not `interfacePropertiesList.Remove(idx)`?

MaceWindu: I would expect erasing value is more perfomant than List.RemoveAt

sdanyliv (approved):

viceroypenguin (approved):

PR #4062, opened 2023-04-11
FOR UPDATE MySql hint

### LINKED ISSUE #4055: Cannot attach hint to select part of INSERT FROM SELECT query
### Discussed in https://github.com/linq2db/linq2db/discussions/4054

<sup>Originally posted by **danillykk** April  5, 2023</sup>
I'm migrating from BLToolkit to Linq2db and trying to rebuild the InsertIfNotExist query because Linq2db builds it differently.

There is the following table
```
[Table("TestTable")]
public class TestTable
{
	[PrimaryKey]
	public int Id { get; set; }

	[Column("OtherNaming"), NotNull]
	public string Data { get; set; }
		
	[Column("timestampUpdated"), DataType(DataType.Timestamp)]
	public DateTime TimestampUpdated { get; set; }
}
```

I want to make the following request (as with BLToolkit)
```
// Declare block
... 
INSERT INTO `TestTable`
(
	`Id`,
	`OtherNaming`,
	`timestampUpdated`
)
SELECT
	1 as `c1`,
	'Something data' as `c2`,
	@p1 as `c3`
FROM
	(
		SELECT
			Count(*) as `c4`
		FROM
			`TestTable` `a`
		WHERE
			`a`.`Id` = 1
	) `item`
WHERE
	`item`.`c4` = 0 AND TRUE FOR UPDATE
```
a fake field appeared to add FOR UPDATE keyword in BLToolkit
```
[SqlProperty("TRUE FOR UPDATE", ServerSideOnly = true)]
private static bool ForUpdate => true;
```
but Linq2db builds its query differently.
```
...
	) `item`
WHERE
	Count(*) = 0 AND
	TRUE FOR UPDATE = 1 AND
	TRUE FOR UPDATE IS NOT NULL AND
	`a`.`Id` = 1
```

I'm trying to build it on Linq2db
```
...
private readonly ParameterExpression parameterIn = Expression.Parameter(typeof(CountHolder), "input");
private class CountHolder
{
	public int Count { get; set; }
}
...

var query = table.Where(a => a.Id == 1);
var insert = () => new TestTable
	{
		Id = 1,
		Data = "Something data",
		TimestampUpdated = DateTime.UtcNow
	});
var existenceCheckQuery = query.GroupBy(item => new { }).Select(item => new CountHolder {Count = item.Count()});
var commonQuery = existenceCheckQuery.AsSubQuery().Where(item => item.Count == 0).AsMySql().ForUpdateHint();
var setterExpression = Expression.Lambda<Func<CountHolder, T>>(insertSetter.Body, parameterIn);
var insertedCount = commonQuery.Insert(DataConnection.GetTable<T>(), setterExpression);

```
The problem is that the FOR UPDATE keyword doesn't appear at the end of the query.

I also tried to solve the question with the field
```
[Sql.Property(ServerSideOnly = true, ExpectExpression = true, Expression = "TRUE FOR UPDATE", CanBeNull = false, IsPredicate = true)]
private static bool ForUpdate => true;
```
with the following change
```
...
var commonQuery = existenceCheckQuery.AsSubQuery().Where(item => item.Count == 0 && ForUpdate)
...
```
but I get an unnecessary comparison at the end of the query
```
... AND TRUE FOR UPDATE = 1
```

I have no idea how to keep the old behavior</div>

igor-tkachev:
Fix #4055


Build/Azure/scripts/clickhouse.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-docker run -d --name clickhouse --ulimit nofile=262144:262144 -p 8123:8123 -p 9000:9000 -p 9004:9004 -p 9005:9005 -e CLICKHOUSE_USER=testuser -e CLICKHOUSE_PASSWORD=testuser clickhouse/clickhouse-server:latest
```
MaceWindu, line 3: revert

Source/LinqToDB/Linq/Builder/ParametersContext.cs
```diff
@@ -281,7 +281,20 @@ ParameterAccessor PrepareConvertersAndCreateParameter(ValueTypeExpression newExp
 					}
 					else
 					{
-						newExpr.ValueExpression = ColumnDescriptor.ApplyConversions(MappingSchema, newExpr.ValueExpression, newExpr.DataType, null, true);
+						// Try GetConvertExpression<.., DataParameter>() first.
```
MaceWindu, line 284: this belongs to wrong PR

Tests/Base/Tools/SettingsReader.cs
```diff
@@ -41,6 +41,26 @@ void Merge(TestSettings settings1, TestSettings settings2)
 					if (!settings1.Connections.ContainsKey(connection.Key))
 						settings1.Connections.Add(connection.Key, connection.Value);
 
+				if (settings2.Providers is not null)
```
MaceWindu, line 44: what's this change about?

MaceWindu (approved):
resolve conflicts

MaceWindu (dismissed):

viceroypenguin (approved):

PR #4072, opened 2023-04-16
DataConnection tracing improvement

igor-tkachev:
Adds the DataConnection.DefaultOnTraceConnection property.


Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -467,38 +467,38 @@ public  bool   IsMarsEnabled
...
 		/// Writes the trace out using <see cref="WriteTraceLineConnection"/>.
 		/// </summary>
-		void OnTraceInternal(TraceInfo info)
+		static void OnTraceInternal(TraceInfo info)
 		{
 			switch (info.TraceInfoStep)
 			{
 				case TraceInfoStep.BeforeExecute:
-					WriteTraceLineConnection(
+					info.DataConnection.WriteTraceLineConnection(
```
viceroypenguin, line 490: I'd probably prefer to see `var dc = info.DataConnection` at the top of this method and all of the references use the local variable rather than the full name each time. Technically would save one pointer deref each time it's access in this method, and there are some JIT perfs it can do as a local variable instead of as a property ref.

viceroypenguin (approved):

MaceWindu (approved):

Issue #4066, opened 2023-04-14
ClickHouse query hints/extensions 


jljordaan:
### Describe your issue

The Roadmap [mentions](https://github.com/linq2db/linq2db/wiki/Releases-and-Roadmap#additional-features) that feedback is welcomed on which custom query hints/extensions should be implemented for the **ClickHouse** provider.

May I suggest that the [FINAL](https://clickhouse.com/docs/en/sql-reference/statements/select/from#final-modifier) table modifier is supported via a custom `.AsClickHouse().UseFinalTableHint()` table hint extension.

Having this functionality will enable our team to start using linq2db in our production environment to perform queries against our OLAP database.

MaceWindu:
@igor-tkachev FYI

igor-tkachev:
Will they ever stop doing it to us?

MaceWindu:
We don't have hints API for ClickHouse yet. I decided to not implement them from start, as ClickHouse has miriad of SQL extensions and it is not clear which of them are useful and I don't have so many time to waste :)

igor-tkachev:
Is there any ClickHouse hint related documentation?

jljordaan:
Hi @MaceWindu & @igor-tkachev 

The time and effort you guys put into this project is very much appreciated!

I couldn't find an explicit list of the hints supported, but the [SELECT Query Syntax](https://clickhouse.com/docs/en/sql-reference/statements/select) does show what I think might be all the various supported hints.

I do understand that it does not makes sense to implement all the ClickHouse SQL extensions from the get-go. Would it maybe be possible to only support a `.AsClickHouse()` database specific API that could then be extended/implemented by end users to add their own hints etc. as they require? I might be on the wrong track here - so please guide me.

PR #4075, opened 2023-04-17
ClickHouse query hints/extensions

igor-tkachev:
#4066


Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
```diff
@@ -0,0 +1,236 @@
...
+				Expression.Call(
+					null,
+					MethodHelper.GetMethodInfo(TableHint, table, hint),
+					table.Expression, Expression.Constant(hint))
+			);
+
+			return new ClickHouseSpecificTable<TSource>(newTable);
+		}
+
+//		/// <summary>
```
viceroypenguin, line 59: This much commented code is auspicious in it's presence. Should it be removed, uncommented, ??  At the very least, a `// TODO:` should be added to indicate why this code is present, but commented out.

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
viceroypenguin, line 1: This change has nothing to do with CH. It should be introduced on a separate PR; either one related to MySql or one specific to making this change.

igor-tkachev: Then it is better not to do anything at all because I do not have much time for that. 
Unfortunately, we only have a choice: either we make small changes (formatting, fixing naming conventions) as we go, or we accumulate all this garbage, increase technical debt  and entropy of the universe.

MaceWindu:
there is a buch of failing tests

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
```diff
@@ -0,0 +1,246 @@
...
+		static Expression<Func<IClickHouseSpecificTable<TSource>,IClickHouseSpecificTable<TSource>>> FinalInScopeHintImpl2<TSource>()
+			where TSource : notnull
+		{
+			return table => TableHint(table, Table.Final);
+		}
+
+		/// <summary>
+		/// Adds <b>FINAL</b> modifier to FROM Clause of all the tables in the method scope.
+		/// </summary>
+		[ExpressionMethod(ProviderName.ClickHouse, nameof(FinalInScopeHintImpl))]
```
MaceWindu, line 222: It's not clear from description and test how it should work.
ClickHouse documentation defines FINAL as modifier to FROM clause which could be table/subquery/table_function but method description implies it will be applied to tables in subquery, which doesn't sound correct

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
viceroypenguin: You could put it on PR #4062 - it would make much more sense there.

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.tt
```diff
@@ -0,0 +1,57 @@
...
+namespace LinqToDB.DataProvider.ClickHouse
+{
+	public static partial class ClickHouseHints
+	{
+<#
+	GenerateJoinHint("Outer");
+	GenerateJoinHint("Semi");
+	GenerateJoinHint("Anti");
+	GenerateJoinHint("Any");
+	GenerateJoinHint("AsOf");
```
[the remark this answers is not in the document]
igor-tkachev: done

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
igor-tkachev: This specific overload will be applied only to tables in the method scope. We do not have a method for subqueries in scope.

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
```diff
@@ -21,11 +21,23 @@ public static class Table
...
+			public const string GlobalSemi  = Global + " SEMI";
+			public const string GlobalAnti  = Global + " ANTI";
+			public const string GlobalAny   = Global + " ANY";
+			public const string GlobalAsOf  = Global + " ASOF";
+			public const string All         = "ALL";
+			public const string AllOuter    = All + " OUTER";
+			public const string AllSemi     = All + " SEMI";
+			public const string AllAnti     = All + " ANTI";
+			public const string AllAny      = All + " ANY";
+			public const string AllAsOf     = All + " ASOF";
```
MaceWindu, line 40: It doesn't make sense to me.

ClickHouse documentation defines 4 independent groups of join modifiers which doesn't match current implementation

[`[GLOBAL] [ANY|ALL|ASOF] [INNER|LEFT|RIGHT|FULL|CROSS] [OUTER|SEMI|ANTI] JOIN`](https://clickhouse.com/docs/en/sql-reference/statements/select)

MaceWindu: looks like CH documentation is out-of-sync. Let me check which one is correct...

MaceWindu: join is more recent, but in any case I will rise documentation issue

MaceWindu: filled https://github.com/ClickHouse/clickhouse-docs/issues/1059

sdanyliv: One note can we transform these hints from string to classes? I really don't like that extensions accepts everything that is string. I would like to see code suggestions.

igor-tkachev: > It doesn't make sense to me.

Do you have any alternatives in mind?

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
igor-tkachev: We do not have such extension for now. TablesInScope was added to simplify assigning extension to all the tables in scope. The SubqueriesInScope extension is kind of ambiguous as we may generate subqueries or not depending on many reasons. If you think we really need it, I can add it.

MaceWindu:
Could we have tests against DB to see SQL in baselines and to validate it against ClickHouse?

MaceWindu (approved):

PR #4086, opened 2023-04-23
Fix eager load auto-transaction

### LINKED ISSUE #4057: LoadWith() One To Many : InvalidOperationException: There is already an open DataReader
We should look wether we have issue here. It doesn't sound right.

### Discussed in https://github.com/linq2db/linq2db/discussions/4056

<sup>Originally posted by **mxmissile** April  5, 2023</sup>
I'm obviously doing something wrong here or I dont understand how `LoadWith()` works. I'm getting

`System.InvalidOperationException: There is already an open DataReader associated with this Connection which must be closed first.
`
```
Parent? parent; 

await using (var db = new DataContext(options))
{
   parent = await db.GetTable<Parent>()
                .LoadWith(x => x.Children)
                .FirstOrDefaultAsync(x => x.Id == 208243);
} // exception is thrown here

Console.WriteLine(parent.Children.Count()); 
```

And my poco classes:

```
[Table("parents")]
    public class Parent
    {
        [PrimaryKey, Identity, Column("parent_id")]
        public int Id { get; set; }

        [Column] 
        public string Title { get; set; }

        [Association(
            ThisKey = nameof(Id),
            OtherKey = nameof(Child.ParentId))]
        public IEnumerable<Child> Children { get; set; } = null!;
    }

    [Table("children")]
    public class Child
    {
        [PrimaryKey, Identity, Column("child_id")]
        public int Id { get; set; }

        [Column("parent_id")] 
        public int ParentId { get; set; }
    }
```

 How can I eager load `Parent.Children`?

UPDATE: I'm not sure why, but using `DataConnection` instead of `DataContext` solved the issue, I think it has something to do with `Async`.</div>

MaceWindu:
Fix #4057

There were at least two bugs:
- for async enumeration we were disposing connection before data reader (in sync code disposal were done in different place with correct order) which raised exception in SqlClient when it tried to execute `Rollback Transaction` command on connection with already open data reader
- transaction rollback is actually another bug, affecting also sync code: because we open eager load transaction not on `DataContext` but on underlying `DataConnection` - `DataContext` with `KeepConnectionAlive = false` option set (default) doesn't know anything about it and close underlying connection after first eager load query for async code or don't close it at all for sync code (that's why there were no exception for sync code)
- also there were small issues with missing delegate unsubscribe + improved some async code
- [UPDATE] another issue with DataContext: context-id generation was incorrect, leading to false cache hits

In general it is caused by:
1. lack of `DataContext` testing. In general we test only `DataConnection` and remote context
2. access to `DataConnection` functionality directly, ignoring `DataContext` needs

### DIFF
--- a/Tests/Model/Extensions.cs
+++ b/Tests/Model/Extensions.cs
@@ -6,11 +6,9 @@ namespace Tests.Model
 	{
 		public static DataConnectionTransaction? BeginTransaction(this ITestDataContext context)
 		{
-			if (context is DataConnection)
-				return ((DataConnection)context).BeginTransaction();
-			//else if (context is ServiceModelDataContext)
-			//	((ServiceModelDataContext)context).BeginBatch();
-			return null;
+			return context is DataConnection dc
+				? dc.BeginTransaction()
+				: null;
 		}
 	}
 }
--- a/Tests/Model/Extensions.cs
+++ b/Tests/Model/Extensions.cs
@@ -4,12 +4,13 @@ namespace Tests.Model
 {
 	public static class Extensions
 	{
-		public static void BeginTransaction(this ITestDataContext context)
+		public static DataConnectionTransaction? BeginTransaction(this ITestDataContext context)
 		{
 			if (context is DataConnection)
-				((DataConnection)context).BeginTransaction();
+				return ((DataConnection)context).BeginTransaction();
 			//else if (context is ServiceModelDataContext)
 			//	((ServiceModelDataContext)context).BeginBatch();
+			return null;
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


Tests/Model/Extensions.cs
```diff
@@ -4,12 +4,13 @@ namespace Tests.Model
...
 	{
-		public static void BeginTransaction(this ITestDataContext context)
+		public static DataConnectionTransaction? BeginTransaction(this ITestDataContext context)
 		{
 			if (context is DataConnection)
-				((DataConnection)context).BeginTransaction();
+				return ((DataConnection)context).BeginTransaction();
 			//else if (context is ServiceModelDataContext)
 			//	((ServiceModelDataContext)context).BeginBatch();
+			return null;
```
viceroypenguin, line 13:
```csharp
return context is DataConnection dc
  ? dc.BeginTransaction()
  : null;
```

viceroypenguin (dismissed):

igor-tkachev (approved):

viceroypenguin (approved):

PR #4093, opened 2023-04-29
Fix interfaces inheritance support

### LINKED ISSUE #3034: Interfaces inheritance
I'm doing some generic work on several unrelated entities with common code, thanks to interfaces.

linq2db works well enough with interfaces but it seems to ignore members that come from interfaces hierarchies.

For example if you declare this model:
```cs
interface IA
{
  int Id { get; set; }
}

interface IB : IA
{
 string Name { get; set; }
}
```

If you then write Linq that refers to `Id` through `IB`, such as:
```cs
public void ChangeName(int id, string name)
{
  return db.GetTable<IB>().TableName("MyTable")
    .Where(x => x.Id == id)
    .Set(x => x.Name, x => newName)
    .Update();
}
```

It compiles but at runtime linq2db complains that `x.Id` is not a mapped property of `IB`.

MaceWindu:
Fix  #3034

Cherrypick from #4036

### DIFF
--- a/Source/LinqToDB/Extensions/ReflectionExtensions.cs
+++ b/Source/LinqToDB/Extensions/ReflectionExtensions.cs
@@ -28,6 +28,9 @@ namespace LinqToDB.Extensions
 
 		public static MemberInfo[] GetPublicInstanceValueMembers(this Type type)
 		{
+			if (type.IsInterface)
+				return GetInterfacePublicInstanceValueMembers(type);
+
 			var members = type.GetMembers(BindingFlags.Instance | BindingFlags.Public)
 				.Where(m => m.IsFieldEx() || m.IsPropertyEx() && ((PropertyInfo)m).GetIndexParameters().Length == 0)
 				.ToArray();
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Extensions/ReflectionExtensions.cs
```diff
@@ -62,6 +65,35 @@ public static MemberInfo[] GetPublicInstanceValueMembers(this Type type)
...
 
+		private static MemberInfo[] GetInterfacePublicInstanceValueMembers(Type type)
+		{
+			var members = type
+				.GetMembers(BindingFlags.Instance | BindingFlags.Public)
+				.Where(m => m.IsFieldEx() || m.IsPropertyEx() && ((PropertyInfo)m).GetIndexParameters().Length == 0);
+
+			var interfaces = type.GetInterfaces();
+			if (interfaces.Length == 0)
+				return members.ToArray();
```
viceroypenguin, line 76: Should this be possible since `type` is already known to be an Interface?

viceroypenguin (dismissed):

Source/LinqToDB/Extensions/ReflectionExtensions.cs
MaceWindu: Yes, for interface it returns inherited interfaces

igor-tkachev (approved):

PR #4094, opened 2023-04-29
Support readonly interface properties with writeable implmentation in query materialization

### LINKED ISSUE #4082: Interface members are not mapped in some cases
### Describe your issue

Interface members are not mapped in some cases and default value is returned. The issue is similar to https://github.com/linq2db/linq2db/issues/3034 but there is no exception, just incorrect data returned from the database.

### Steps to reproduce

Prerequisites:
- Interface contains get-only members.
- IQueryable\<TEntity\> is casted to IQueryable\<TInterface\>
- Extension methods such as .Where() or .Take() are called on instance of IQueryable\<TInterface\>

```c#
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;

var db = new DataConnection(LinqToDB.ProviderName.SqlServer2019,
  @"Server=.\;Database=Test;Trusted_Connection=True;Enlist=False;TrustServerCertificate=True");

IQueryable<IIdentifiable> q = db.GetTable<UserAccount>();

q = q.Where(u => u.ID != 0); // This line is required to trigger the bug

 // Displays 0 instead of the actual account ID
Console.WriteLine(q.First().ID);

public interface IIdentifiable
{
    int ID 
    { 
        get; // addition of setter fixes the issue
    }
}

[Table(Name = "[dbo].[UserAccounts]")]
public class UserAccount : IIdentifiable
{
    [PrimaryKey, Identity]
    public int ID { get; set; }

    [Column]
    public string Email { get; set; } = string.Empty;
}
```

### Environment details

`Linq To DB` version: 5.1.1

Database (with version): SQL Server 2019

ADO.NET Provider (with version): Microsoft.Data.SqlClient 4.1.0

Operating system: Windows 10

.NET Version: .NET7

MaceWindu:
Fix #4082

Cherrypick from #4036

### DIFF
--- a/Source/LinqToDB/Data/RecordReaderBuilder.cs
+++ b/Source/LinqToDB/Data/RecordReaderBuilder.cs
@@ -1,17 +1,18 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.Common;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 
 namespace LinqToDB.Data
 {
-	using System.Data.Common;
 	using Expressions;
 	using Linq;
 	using Linq.Builder;
-	using LinqToDB.Common;
+	using Common;
+	using Extensions;
 	using Mapping;
 	using Reflection;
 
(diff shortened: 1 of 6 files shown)


Source/LinqToDB/Data/RecordReaderBuilder.cs
```diff
@@ -122,16 +123,18 @@ IEnumerable<ReadColumnInfo> GetReadIndexes(EntityDescriptor entityDescriptor)
...
 		Expression BuildDefaultConstructor(EntityDescriptor entityDescriptor, Type objectType)
 		{
-			var members = new List<(ColumnDescriptor column, ConvertFromDataReaderExpression expr)>();
+			var members = new List<(ColumnDescriptor column, MemberInfo storage, ConvertFromDataReaderExpression expr)>();
 			foreach (var info in GetReadIndexes(entityDescriptor))
 			{
-				if (info.Column.Storage != null ||
-					  info.Column.MemberAccessor.MemberInfo is not PropertyInfo pi ||
-					  pi.GetSetMethod(true) != null)
+				var cd = info.Column;
```
viceroypenguin, line 129: Performance nit: is there any way to evaluate all of these `if`s and assign a variable that would be passed into `members.Add()`? This repeated `ValueTuple` creations might increase the amount of IL for this method. (x3, see EnumerableContext.cs and TableBuilder.TableContext.cs as well). In fact, with three implementations, would it make sense to abstract this? (if not, that's fine).

viceroypenguin (approved):

Source/LinqToDB/Data/RecordReaderBuilder.cs
MaceWindu: regarding abstraction - I've discussed it with @sdanyliv - in new parser it's already moved to single place, so I decided to leave it as it for now.

viceroypenguin: 👍🏼

viceroypenguin (approved):

igor-tkachev (approved):

Issue #4098, opened 2023-05-01
GroupBy clause is different depending on presence of .AsCte()


brd-mikek:
I have this situation when SQL is not generated correctly around GROUP BY clause.
If I use .AsCte() on subquery, the SQL looks correct.

See program attached to reproduce the problem.
Also, see the screenshot with generated SQL.

![Linq2DbProblem](https://user-images.githubusercontent.com/77951000/235388275-ebf216de-05cf-40b7-a164-4ec1396f54c0.png)

[Linq2DbProblem.zip](https://github.com/linq2db/linq2db/files/11362958/Linq2DbProblem.zip)

MaceWindu:
As another workaround you can use AsSubQuery() instead of Cte()

brd-mikek:
Hi guys!
When are you planning to release this?

MaceWindu:
It was released as part of 5.2.1 release last week

PR #4100, opened 2023-05-01
Fix aggregates tracking in joined subqueries

MaceWindu:
https://github.com/linq2db/linq2db/issues/4098


MaceWindu:
Found multiple places where I can "fix" this specific issue, but only @sdanyliv knows which one is correct :cry:

MaceWindu:
having 3 join builders doesn't help

viceroypenguin (approved):

PR #4129, opened 2023-05-19
Fix builds with VS 17.6

MaceWindu:
If it will cause issues with CI, we will merge it after Azure update to 17.6


MaceWindu:
btw, local testing (and I suspect Azure after it updated) probably broken in 17.6 for us (https://github.com/microsoft/vstest/issues/4467). Test discovery just crashes, I think I managed to see tests only once after 10+ discoveries

igor-tkachev (approved):

PR #4131, opened 2023-05-21
[T4] Fix identifiers normalization

### LINKED ISSUE #4058: ttmodels: ConvertToCompilable is running after the KeyWords checks
### Describe your issue

When a ttmodel is generating the model, it first checks for KeyWord.Contains, appending a `@` if it returns true, and only then calls ConvertToCompilable routine. Now, ConvertToCompilable may be overriden by the user to fix up things by the user... Except now the name is already "tainted" by this `@` prepended to it. Even more, that `@` is actually preventing the NormalizeName routine to uppercase the first letter!

If both ConvertToCompilable and NormalizeName would be called with the original name, the identifier may ended up not being a keyword, so not needing that extra `@` prepended.

Now, maybe that KeyWords check should be made either inside `ConvertToCompilable` or after this or `ToValidName` call?

### Steps to reproduce

Create a database with a table named after a keyword (for example `event`)
Set in the ttmodel configuration the following settings:
```
NormalizeNames = true;
NormalizeNamesWithoutUnderscores = true;
ToValidName = (name, under) =>
{
  var result = ToValidNameDefault(name, under);
  return result;
};
ConvertToCompilable = (name, under) =>
{
  var result = ConvertToCompilableDefault(name, under);
  return result;
}
```
Place the breakpoints on both `return result;`
Run the tt tool in debug mode
Observe that the class name is still `@event` and not `Event`
Observe how both `ToValidName` and `ConvertToCompilable` receives the actual name already with the `@` prepended.

### Environment details

`Linq To DB` version: 5.1.1

Database (with version): MySQL (version irrelevant)

ADO.NET Provider (with version): MySql.Data (8.0.32.1)

Operating system: Windows 11

.NET Version: 6.0

MaceWindu:
Fix #4058

Previously we applied identifier normalization and C# keyword escaping in different order:
- normalize -> escape
- escape -> normalize
- escape [without normalization]

This PR:
- enables same logic for all identifiers: normalize -> escape
- adds options to redefine/restore old behavior for backward compatibility:
  - `NormalizeName` delegate to redefine this logic
  - `NormalizeParameterName=true` to disable procedure/function parameter names normalization
  - `NormalizeProcedureColumnName=true` to disable procedute/table function result table column names normalization
- fixes T4 templates for Oracle (was broken by v5 changes to `OracleTools`)

(didn't regenerate DB2/Informix baselines as it is PITA to setup NETFX providers for them and they are currently not configured in my environment)


viceroypenguin (commented):
Why do some convert `ALLCAPS` => `aLLCAPS`, while others convert `ALLCAPS` to `allcaps` for the parameter names?

MaceWindu:
Could you link second case?

MaceWindu:
That's how it works in T4 currently and we hardly will change it (users could define own normalization logic if they want). When ALLCAPS, it is not lowercased. When ALL_CAPS, it converted to pascal case.

Tests/Tests.T4/Databases/Firebird.generated.cs
```diff
@@ -90,13 +97,13 @@ public partial class OutRefEnumTestResult
 		#region OutRefTest
 
 		[Sql.TableFunction(Name="OutRefTest")]
-		public ITable<OutRefTestResult> OutRefTest(int? ID, int? IN_INPUTOUTPUTID, string? STR, string? IN_INPUTOUTPUTSTR)
+		public ITable<OutRefTestResult> OutRefTest(int? iD, int? inInputoutputid, string? sTR, string? inInputoutputstr)
```
viceroypenguin, line 100: @MaceWindu correction on my question: why `ALLCAPS` => `aLLCAPS`, but `SCREAM_CASE` => `screamCase`? If `SCREAM_CASE` goes to `screamCase`, I'd expect `ALLCAPS` to go to `allcaps`.

viceroypenguin: > That's how it works in T4 currently and we hardly will change it (users could define own normalization logic if they want). When ALLCAPS, it is not lowercased. When ALL_CAPS, it converted to pascal case.

But converting to `camelCase` here is a change that users may not be expecting...

MaceWindu: not my logic, not gonna change it :laughing:  as it is here for a long time

MaceWindu: it applied to parameters ony, so hardly anyone will notice and it doesn't affect build in general

MaceWindu: Anyway, we can add special handling of this case for parameters, but I want to hear more opinions first

viceroypenguin: 😀 works for me! haha

viceroypenguin (approved):

PR #4133, opened 2023-05-21
Support column converters for Execute[Async](sql)

### LINKED ISSUE #4043: linq2db HasConversion doesn't work with Connection.ExecuteAsync
I have same code:

MappingBuilder:
```
        builder.HasTableName("Sendings");
        builder.Property(x => x.Params).HasDbType("jsonb").HasConversionFunc(JsonConvert.SerializeObject, y =>
        {
            return JsonConvert.DeserializeObject<SendingParams>(y)!;
        });
```

Query:
```
var sending = await Connection.ExecuteAsync<Sending>(@"SELECT * FROM ""Sendings"" WHERE ""Params""::json->>'Comment'='test'", CancellationToken.None);
```

Exception:
```
LinqToDB.Common.LinqToDBConvertException: Cannot convert value '{"Nonce": 12, "Amount": 1000000, "Comment": "test", "GasLimit": 2142, "GasPrice": 1000000000, "ToAddress": "cosmos1xwy7k0ffpf6lzlr6zrl20cj63v3gnqret5epm2", "FromAddress": "cosmos15h0zj4lwftstl
r8gufv59setxuv9y4yt75fczn", "TransferAll": false}: System.String' to type 'Domain.Sendings.SendingParams'
 ---> System.InvalidCastException: Invalid cast from 'System.String' to Domain.Sendings.SendingParams'.

```

Conversion doesn't work. For non raw sql queries conversion works.

MaceWindu:
Fix #4043

### DIFF
--- a/Source/LinqToDB/Data/CommandInfo.cs
+++ b/Source/LinqToDB/Data/CommandInfo.cs
@@ -1583,8 +1583,8 @@ namespace LinqToDB.Data
 							new ConvertFromDataReaderExpression(type, idx, null, dataReaderExpr).Reduce(dc, dr));
 				}
 
-				return CreateObjectReader<T>(context.dataConnection, context.dataReader, (dc, dr, type, idx, dataReaderExpr) =>
-					new ConvertFromDataReaderExpression(type, idx, null, dataReaderExpr).Reduce(dc, dr));
+				return CreateObjectReader<T>(context.dataConnection, context.dataReader, (dc, dr, type, idx, dataReaderExpr, conv) =>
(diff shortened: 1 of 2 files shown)


MaceWindu:
There are two tests, that test record materialization using constructor with parameters. In general case, we don't know which parameter mapped to which column as names could be misleading, which means we cannot discover column converter.

Anyways, this is not important case, as for such situations user already could add required conversion to constructor logic.

viceroypenguin (approved):

igor-tkachev (approved):

PR #4152, opened 2023-06-01
Update JetBrains.Annotations.cs

sirduke:
PublicAPIAttribute misses "Inherited = false" flag which causes false negative analysis from ReSharper/Rider like described in https://youtrack.jetbrains.com/issue/RSRP-492409
MeansImplicitUseAttribute and InstantHandleAttribute are updated to be in line with native attributes used by ReSharper/Rider


MaceWindu (approved):
Thanks!

igor-tkachev (approved):

PR #4166, opened 2023-06-14
Optimized ConvertExpressionImpl method

### LINKED ISSUE #3273: Use COALESCE instead of NVL for nested ??
This is not a bug, everything works fine. It's a small SQL codegen improvement.

linq2db uses `nvl` to translate `??`, which is nice because it's shorter than SQL standard `coalesce`.

The trick with `nvl` is that it only accepts 2 parameters, whereas `coalesce` accepts many.

In our codebase, we have a query that computes an error message based on a list of checks, it looks like this:
image

That's 7 `??` chained. 
The resulting SQL is of course 7 nested nvls: `nvl(.., nvl(.., nvl(.., /*4 more*/)))))))`.
It works well enough, I just think it'd be easier to read if it was a single `coalesce(.., .., .., .., .., .., ..)`.

So my suggestion is to convert nested `??` expressions into a single `coalesce` instead of nested `nvl`.
This expression tree: 
image
Could be converted into `coalesce(A, B, C, D, E)` (i.e. a depth-first traversal).

A related Oracle SQL codegen you could have is convert patterns `a == null ? c : d` (and `a != null`) into `nvl2(a, d, c)` (resp. a, c, d).
nvl2 is another Oracle proprietary function that returns the 2nd arg if the first is not null, and the 2rd arg is the first is null.

ddas09:
Fix #3273

Optimized ConvertExpressionImpl method to use nvl function for coalesce expressions with two parameters.

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
@@ -146,7 +146,10 @@ namespace LinqToDB.DataProvider.Oracle
 				{
 					case "Coalesce":
 					{
-						return ConvertCoalesceToBinaryFunc(func, "Nvl");
+						if (func.Parameters.Length == 2)
+							return ConvertCoalesceToBinaryFunc(func, "Nvl");
+							
+						return func;
 					}
 					case "Convert"        :
 					{


jods4 (approved):

viceroypenguin (approved):

MaceWindu:
Thanks!

PR #4173, opened 2023-06-20
#4172: Fixes issue with empty string check in Oracle

### LINKED ISSUE #4172: Issue when using NotEqual check with Enums that map to empty string in Oracle
We're facing issue when trying to use NotEqual check with Enums that map to empty string ("").
Earlier Linq2Db used to just compare the value to NULL, but now it's comparing it to empty string and adding (OR [col] IS NULL) to the predicate.
It works for when we're checking for Equal to empty string but fails for Not Equal to empty string

### Scenario 1 : Checking with "Equal" (Works)
```c#
var data = (
 from pe in db.USER
 where pe.ROLE == Role.Unknown
 select pe).ToList();
```

SQL Generated
```sql
SELECT
	pe.ROLE
FROM
	"USER" pe
WHERE
	(pe.ROLE = '' OR pe.ROLE IS NULL)
```

### Scenario 2 : Checking with "NotEqual" (Doesn't work)
```c#
// Doesn't work	
var data = (
 from pe in db.USER
 where pe.ROLE != Role.Unknown
 select pe).ToList();
```

SQL Generated
```sql
SELECT
	pe.ROLE
FROM
	"USER" pe
WHERE
	(pe.ROLE <> '' OR pe.ROLE IS NULL)
```

### Environment details

`Linq To DB` version: 5.2.1

Database (with version): Oracle 12

.NET Version: 6

[LinqToDbOracleNullValueIssue.zip](https://github.com/linq2db/linq2db/files/11796321/LinqToDbOracleNullValueIssue.zip)

divyanshbhatia1:
Update Oracle11SqlOptimizer to add null check based on operator

Fix #4172

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
@@ -90,7 +90,17 @@ namespace LinqToDB.DataProvider.Oracle
 							{
 								var sc = new SqlSearchCondition();
 								sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.ExprExpr(expr.Expr1, expr.Operator, expr.Expr2, null), true));
-								sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr2, false), true));
+
+								if (expr.Operator == SqlPredicate.Operator.Equal)
+								{
+									// When checking Equal to Empty String, adding 'OR [col] IS NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr2, false), true));
+								}
+								else if (expr.Operator == SqlPredicate.Operator.NotEqual)
+								{
+									// When checking NotEqual to Empty String, adding 'AND [col] IS NOT NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr2, true), false));
+								}
 								return sc;
 							}
 						}
@@ -102,7 +112,17 @@ namespace LinqToDB.DataProvider.Oracle
 							{
 								var sc = new SqlSearchCondition();
 								sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.ExprExpr(expr.Expr1, expr.Operator, expr.Expr2, null), true));
-								sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, false), true));
+
+								if (expr.Operator == SqlPredicate.Operator.Equal)
+								{
+									// When checking Equal to Empty String, adding 'OR [col] IS NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, false), true));
+								}
+								else if (expr.Operator == SqlPredicate.Operator.NotEqual)
+								{
+									// When checking NotEqual to Empty String, adding 'AND [col] IS NOT NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, true), false));
+								}
 								return sc;
 							}
 						}
--- a/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
@@ -119,7 +119,7 @@ namespace LinqToDB.DataProvider.Oracle
 								// else add 'OR [col] IS NULL'
 								sc.Conditions.Add(new(
 								  isNot: false,
-								  new SqlPredicate.IsNull(expr.Expr2, isNot: isNotEqual),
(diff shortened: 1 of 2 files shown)


MaceWindu (changes requested):
We need test(s), that demonstrate this behavior

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -4277,5 +4277,79 @@ public void Issue3740Test2([IncludeDataSources(TestProvName.AllOracle12Plus)] st
...
+		}
+
+		[Test]
+		public static void Issue4172Test1([IncludeDataSources(TestProvName.Oracle12Managed)] string context)
+		{
+			using var db = GetDataConnection(context);
+
+			try
+			{
+				using var users = db.CreateTable<ISSUE4172TABLE>();
```
MaceWindu, line 4308: Use CreateLocalTable helper instead of try/catch with delete

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -4277,5 +4277,79 @@ public void Issue3740Test2([IncludeDataSources(TestProvName.AllOracle12Plus)] st
...
+		}
+
+		[Test]
+		public static void Issue4172Test2([IncludeDataSources(TestProvName.Oracle12Managed)] string context)
+		{
+			using var db = GetDataConnection(context);
+
+			try
+			{
+				using var users = db.CreateTable<ISSUE4172TABLE>();
```
MaceWindu, line 4334: use CreateLocalTable

MaceWindu (changes requested):

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -90,7 +90,17 @@ public override ISqlPredicate ConvertPredicateImpl(ISqlPredicate predicate, Conv
...
+								if (expr.Operator == SqlPredicate.Operator.Equal)
+								{
+									// When checking Equal to Empty String, adding 'OR [col] IS NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr2, false), true));
+								}
+								else if (expr.Operator == SqlPredicate.Operator.NotEqual)
+								{
+									// When checking NotEqual to Empty String, adding 'AND [col] IS NOT NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr2, true), false));
+								}
```
jods4, line 103: This changes the behaviors of operators >= and <=. 
It's a breaking change for `Code >= ""`, admittedly an edge case but there's no good reason to break this.

I suggest to simplify to:
```csharp 
bool isNotEqual = expr.Operator == SqlPredicate.Operator.NotEqual;

sc.Conditions.Add(new(
  isNot: false, 
  new SqlPredicate.IsNull(expr.Expr2, isNot: isNotEqual),
  isOr: !isNotEqual)
);
```

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -102,7 +112,17 @@ public override ISqlPredicate ConvertPredicateImpl(ISqlPredicate predicate, Conv
...
+								if (expr.Operator == SqlPredicate.Operator.Equal)
+								{
+									// When checking Equal to Empty String, adding 'OR [col] IS NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, false), true));
+								}
+								else if (expr.Operator == SqlPredicate.Operator.NotEqual)
+								{
+									// When checking NotEqual to Empty String, adding 'AND [col] IS NOT NULL'
+									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, true), false));
+								}
```
jods4, line 125: Same as comment on `expr.Expr2 == ""`

jods4 (changes requested):

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -113,16 +113,16 @@ public override ISqlPredicate ConvertPredicateImpl(ISqlPredicate predicate, Conv
...
-									// When checking NotEqual to Empty String, adding 'AND [col] IS NOT NULL'
-									sc.Conditions.Add(new SqlCondition(false, new SqlPredicate.IsNull(expr.Expr1, true), false));
-								}
+								bool isNotEqual = expr.Operator == SqlPredicate.Operator.NotEqual;
+
+								// Add 'AND [col] IS NOT NULL' when checking Not Equal to Empty String,
+								// else add 'OR [col] IS NULL'
+								sc.Conditions.Add(new(
+								  isNot: false,
+								  new SqlPredicate.IsNull(expr.Expr2, isNot: isNotEqual),
```
jods4, line 122: Copy-paste mistake. It should be `expr.Expr1` on this side.

jods4 (approved):

MaceWindu (approved):

PR #4175, opened 2023-06-23
Fix MergeWithOutputInto with conditional in output expression

poizan42:
This makes the test included in the commit work instead of throwing a NotImplementedException.

I had a hard time figuring out what IsExpression is actually supposed to do. It asks whether an expression is an expression (?), and the comments are less than helpful. It's something about whether the (linq) expression can be turned into a valid SQL expression in the current context? - but then what does it mean when expression is null?

I figured the worst that can happen is that incorrect sql is produced now where it previously crashed with NotImplementedException. It at least makes this case work, and can't affect anything that didn't crash before.

### DIFF
--- a/Source/LinqToDB/Linq/Builder/SingleExpressionContext.cs
+++ b/Source/LinqToDB/Linq/Builder/SingleExpressionContext.cs
@@ -61,17 +61,9 @@ namespace LinqToDB.Linq.Builder
 
 		public IsExpressionResult IsExpression(Expression? expression, int level, RequestFor requestFlag)
 		{
-			if (requestFlag != RequestFor.Field)
+			if (requestFlag != RequestFor.Field || expression == null)
 			{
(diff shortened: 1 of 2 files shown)


poizan42:
The test-all and test-all (Tests: Linux MariaDB) checks are broken?

MaceWindu:
just glith, restarted failed run

poizan42:
Didn't help, now test-all just failed after 7 minutes without logging anything

MaceWindu:
ignore it, looks like mariadb docker image is broken or something changed with its setup

MaceWindu:
probably caused by https://github.com/MariaDB/mariadb-docker/issues/512

MaceWindu (approved):

viceroypenguin (approved):

igor-tkachev (approved):

PR #4177, opened 2023-06-24
Fix string.IsNullOrEmpty mapping

### LINKED ISSUE #4138: string.IsNullOrEmpty does not match C# version of string.IsNullOrEmpty for MS SQL Server (actually it matches string.IsNulOrWhitespace))
### Describe your issue

when you use in a query function like **!string.IsNullOrEmpty** it is converted to SQL like
SELECT
	Count(*)
FROM
	[Data$t11] [auto968]
WHERE
	NOT ([auto968].[text] IS NULL OR Len([auto968].[text]) = 0)

and according to [https://learn.microsoft.com/en-us/sql/t-sql/functions/len-transact-sql?view=sql-server-ver16] **LEN** Returns the number of characters of the specified string expression, **excluding trailing spaces**. It means that string of space characters will not match the predicate. Please use DATALEN or fix it in a different proper way.

### Steps to reproduce

```c#
GetTable().Where (_=> !string.IsNullOrEmpty(_.text)).Count()
```

### Environment details

`Linq To DB` version: *4.3.0*

Database (with version): *SQL Server 2019*

ADO.NET Provider (with version): *Microsoft.Data.SqlClient 4.1.0*

.NET Version: *net4.8*

MaceWindu:
Fix #4138

### DIFF
--- a/Tests/Linq/Linq/StringFunctionTests.cs
+++ b/Tests/Linq/Linq/StringFunctionTests.cs
@@ -1449,6 +1449,28 @@ namespace Tests.Linq
 			}
 		}
 
+		public sealed class IsNullOrEmptyTable
+		{
+			public int     Id    { get; set; }
+			public string? Value { get; set; }
+		}
+
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/4138")]
+		public void IsNullOrEmpty3([DataSources] string context)
+		{
+			using var db = GetDataContext(context);
+			using var t  = db.CreateLocalTable(new[]
+			{
+				new IsNullOrEmptyTable() { Id = 1, Value = "   " },
+				new IsNullOrEmptyTable() { Id = 2, Value = ""    },
+			});
+
+			var results = (from p in t where string.IsNullOrEmpty(p.Value)).ToList();
+
+			Assert.That(results.Count, Is.EqualTo(1));
+			Assert.That(results[0].Id, Is.EqualTo(2));
+		}
+
 		[Table]
 		sealed class CollatedTable
 		{


MaceWindu:
I don't get why some DBs think they have right to mess with user data interpretation....

viceroypenguin (commented):
Is this a breaking change? Should we be concerned with people who have already written code based on `string.Length` and adapted to the differences between c# and their database on how `string.Length` is interpreted?

MaceWindu:
If people rely on old behavior of Length, it could be breaking change for them (MSSQL/SQLCE users only)

Still I think it is a bug that should be fixed as:
- for .net mappings we try to emulate .net behavior when possible
- for db-behavior users should use explicit mappings for database functions

viceroypenguin:
If waiting till v6, I'm good with it

igor-tkachev (approved):

igor-tkachev:
I do not like this change at all as we fix very exotic case and break performance all other.
Lets keep previous behavior by default and add some settings to switch it to this implementation.

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -393,11 +403,41 @@ public static TTo From<TFrom>(TFrom obj)
...
 		}
 
+		/// <summary>
+		/// Returns string length in characters by calling corresponding database function.
+		/// Compared to <see cref="Length(string?)"/> function, it tries to calculate length in same way as <see cref="string.Length"/> property
+		/// for databases with non-standard length functions.
+		/// E.g. see LEN function notes for SQL CE or MSSQL databases.
+		/// Known issues with empty string:
+		/// <list type="bullet">
+		/// <item>Oracle treats '' as NULL and returns NULL instead of 1</item>
```
jods4, line 419: "instead of 0"

jods4:
I'm with @igor-tkachev on this one, I don't like this change.

1. **This is a massive breaking change for people who have fixed-length string columns in their DB such as `char(50)`. This is common in legacy schemas and is the motivation why `LEN` would not count trailing spaces.**

2. I see linq2db as a thin predictable mapper over SQL, even if in some specific cases the behaviour of SQL and C# differ.
These differences cannot all be efficiently erased, users have to be conscious of them and SQL semantics. Examples include: three-valued logic, null comparisons, empty string is null in Oracle, different datatypes range (e.g. `DateTime`), etc.
I don't like creating complex, inefficient expression for minor edge cases, and depriving users from an easy access to straightforward SQL code. 

It's surprising that `"abc".Length` does not convert to `LEN('abc')`.

I think the best way to proceed is to keep everything as-is and provide `FullLength()` as a SQL extension for SQL Server provider. If SQL Server users don't want the native behaviour, they can use this new function instead.

igor-tkachev:
Also I would add a method (configuration property) to switch default conversion to this new way.

igor-tkachev:
> Did we get a lot of feedback on this? #4138 was opened recently but this behaviour has been in linq2db + Sql Server for years.

This is the point. I have been using Sql Server for 25 years now and have never experienced this before.

PR #4178, opened 2023-06-24
Fix MariaDB 11 docker setup

MaceWindu:


grooverdan:
Appoligies for triggering the need to perform this change. Thanks for supporting MariaDB in LINQ to DB

MaceWindu:
@grooverdan no problem 😃

PR #4182, opened 2023-06-27
Fix table attributes ignored for output table when using MergeWithOutputInto

poizan42:
Referring to the outputTable using Expression.Constant results in TableAttributeBuilder not being used. Unfortunately TableBuilder.BuildSequence just returns a TableBuilder+TableContext which only bases its SqlTable on the original type without taking modified attributes into account. So in order to make the modified attributes take effect the full expression from outputTable must be used rather than just embedding it in a ConstantExpression (alternatively should the logic in TableBuilder be changed so the SqlTable is generated with the changed attributes)

### DIFF
--- a/Source/LinqToDB/LinqExtensions.Merge.cs
+++ b/Source/LinqToDB/LinqExtensions.Merge.cs
@@ -1099,7 +1099,7 @@ namespace LinqToDB
 					null,
 					Methods.LinqToDB.Merge.MergeWithOutputInto.MakeGenericMethod(typeof(TTarget), typeof(TSource), typeof(TOutput)),
 					currentQuery.Expression,
-					Expression.Constant(outputTable),
+					((IQueryable<TOutput>)outputTable).Expression,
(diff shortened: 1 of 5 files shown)


MaceWindu:
hmm, I see we have same code for `Delete/Update/Insert[OutputInto]` APIs. Could you update them too?

MaceWindu:
I would prefer it fixed in this PR, but it's not a big deal if you want it as separate one

poizan42:
@MaceWindu I couldn't find anywhere else an outputTable is embedded in a ConstantExpression, but I have added tests for checking handling of tables with options/name attribute changes for all *WithOutputInto overloads

viceroypenguin (approved):

poizan42:
Something went wrong with the database server. Can I restart the run myself?

MaceWindu:
restarted failed test job

igor-tkachev (approved):

PR #4196, opened 2023-07-07
Working on IN/EXISTS conversion.

igor-tkachev:
`IN`/`NOT IN` should support C# semantic for nullable.


jods4:
> `IN`/`NOT IN` should support C# semantic for nullable.

It does (and it's a mess). Can you provide more details what you're trying to fix?

Looking quickly at the code, it looks like you are tweaking `IN (SELECT ...)`, which I think is not currently supported as C# semantics. 
Are you trying to turn `x IN (SELECT y FROM ..)` into `EXISTS(SELECT 1 FROM (SELECT y FROM ..) WHERE y IS NOT DISTINCT FROM x)`?
It would be nice to describe the transformations you're adding in the PR description for an easy first feedback before squinting at the C# code (BTW I'm not a fan of using deconstruction to pattern match `SqlCondition`, because `SqlCondition` is unintuitive with its double bool parameters; property matching works the same and has explicit names for those bools).

Please make sure your transformation doesn't run when option `CompareNullsAsValues` is false.

MaceWindu (changes requested):
small remarks

Tests/Linq/Linq/InSubqueryTests.cs
```diff
@@ -110,9 +110,165 @@ public void ContainsExprTest([DataSources] string context, [Values(true, false)]
...
 		public void ContainsNullTest([DataSources] string context, [Values(true, false)] bool preferExists)
 		{
-			using var db = GetDataContext(context, o => o.WithOptions<LinqOptions>(lo => lo with { PreferExistsForScalar = preferExists }));
+			using var db = GetDataContext(context, o => o.WithOptions<LinqOptions>(lo => lo.WithPreferExistsForScalar(preferExists)));
 
 			_ = db.Parent.Select(c => c.Value1).Contains(null);
 		}
+
+		[Test]
+		public void NutNull_In_NotNull_Test([DataSources] string context, [Values(true, false)] bool preferExists)
```
jods4, line 119: For all new tests involving nulls, I'd like to see them run with and without `CompareNullsAsValues`.
It's important for me that when `CompareNullsAsValues` is false nothing else than a straightforward one-to-one translation to SQL happens.

igor-tkachev:
> It does (and it's a mess). Can you provide more details what you're trying to fix?

Currently we generate `Contains` as `EXISTS`. One of the reason is that it is much easier to implement C# semantic with `EXISTS`. So, we do not need to generate 'IN (SELECT * ...` at all. Unfortunately, some providers, such as ClickHouse, do not support `EXISTS` in some cases, but support `IN`. Also some people expect to see `IN` instead of `EXISTS` when they use `Contains`. Some of them even believe that `IN` performs better.

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -217,6 +217,17 @@ public static LinqOptions WithEnableContextSchemaEdit(this LinqOptions options,
 			return options with { EnableContextSchemaEdit = enableContextSchemaEdit };
 		}
 
+		/// <summary>
+		/// Depending on this options linq2db generates different SQL for <c>sequence.Contains(value)</c>.<br/>
```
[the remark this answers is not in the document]
igor-tkachev: fixed

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -413,6 +424,17 @@ public static DataOptions UseEnableContextSchemaEdit(this DataOptions options, b
 			return options.WithOptions<LinqOptions>(o => o with { EnableContextSchemaEdit = enableContextSchemaEdit });
 		}
 
+		/// <summary>
+		/// Depending on this options linq2db generates different SQL for <c>sequence.Contains(value)</c>.<br/>
```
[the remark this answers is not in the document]
igor-tkachev: fixed

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2139,14 +2138,81 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
 				return newElement!;
 
 			newElement = RunOptimization(element, optimizationContext, this, mappingSchema, dataOptions, !withConversion,
-				static (visitor, e) =>
+				(visitor, e) =>
```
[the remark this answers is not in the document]
igor-tkachev: fixed

jods4:
> Currently we generate `Contains` as `EXISTS`. One of the reason is that it is much easier to implement C# semantic with `EXISTS`. So, we do not need to generate `IN (SELECT)` at all. Unfortunately, some providers, such as ClickHouse, do not support `EXISTS` in some cases, but support `IN`. Also some people expect to see `IN` instead of `EXISTS` when they use `Contains`. Some of them even believe that `IN` performs better.

Yeah, in most RDBMS that I know of, `IN (SELECT)` **should** lead to identical execution plan as `EXISTS` (i.e. a semi-join), emphasis on **should**. I have many horror stories with Oracle where seemingly innocent changes in SQL can lead to different execution plans or even bugs.

Beacuse of this, as a user I like having fine control over the generated SQL.
I also think that as much as possible, the generated SQL should be predictable and close to C# function, ideally a 1-to-1 mapping.
So for me:
- `Any()` generates `EXISTS`;
- `Sql.In()` or `Contains()` should generate `IN`;
- As an exception to previous point, when `CompareNullsAsValues` is true, if the list is an `IQueryable`, you can generate `EXISTS` instead of `IN (SELECT)` because it's the only reasonable way to implement the C# semantics.

Tests/Linq/Linq/InSubqueryTests.cs
```diff
@@ -29,13 +29,13 @@ where c.ParentID.In(db.Parent.Select(p => p.ParentID))
 		}
 
 		[Test]
-		public void InNullTest([DataSources] string context, [Values(true, false)] bool preferExists)
+		public void InConstTest([DataSources] string context, [Values(true, false)] bool preferExists)
```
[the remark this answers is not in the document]
igor-tkachev: I added more tests that implement full coverage of NULL. This test can be actually removed.

Tests/Linq/Linq/InSubqueryTests.cs
igor-tkachev: done

igor-tkachev:
This is exactly how it works with this change now. Also we have `LinqOptions.PreferExistsForScalar = false' that you can use to switch to previous behavior.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
...
+							return ConvertInSubquery(ex1, false, col, subQuery, isOr);
+
+						case SqlCondition(false, SqlPredicate.InSubQuery({CanBeNull: true} ex1, true, {Select.Columns : [{CanBeNull: false} col]} subQuery), var isOr)
+							when !visitor.Context.Optimizer.SqlProviderFlags.DoesNotSupportCorrelatedSubquery :
+							return ConvertInSubquery(ex1, true, col, subQuery, isOr);
+					}
+
+					static IQueryElement ConvertInSubquery(ISqlExpression ex1, bool isNot, SqlColumn col, SelectQuery subQuery, bool isOr)
+					{
+						var newQuery = subQuery.Convert((subQuery,col,ex1), static (v,e) =>
```
jods4, line 2200: Instead of passing `subQuery`, pass `subQuery.Where` to context, as this is what you are interested in.
Then it's probably more efficient to test `e == v.Context.Where` before casting e to `SqlWhereClause`.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
...
+						{
+							if (e is SqlWhereClause w && w == v.Context.subQuery.Where)
+							{
+								var wc = new SqlWhereClause(new SqlSearchCondition(w.SearchCondition.Conditions));
+								wc.SearchCondition.Conditions.Add(new(
+									false,
+									new SqlPredicate.ExprExpr(v.Context.col.Expression, SqlPredicate.Operator.Equal, v.Context.ex1, true)));
```
jods4, line 2207: Could you use `SqlPredicate.IsDistinct(expr1, isNot: true, expr2)` ? 
This should pick up the best SQL based:
- on `expr1` and `expr2` nullability (i.e. `=` if at least one expr. is not nullable);
- on your RDBMS (e.g. `<=>` or `is not distinct from` where avail.).

With this in place, I think you can make less pattern matching in `OptimizeElement` and always convert `InSubquery` with this function.
This rule should be ok for all cases (when `CompareNullsAsValues` is true and `DoesNotSupportCorrelatedSubquery` is false):
`x [NOT] IN (SELECT y) --> [NOT] EXISTS (SELECT 1 WHERE y IS NOT DISTINCT FROM x)`
I would do this regardless of nullability of `x` and `y`. If they're not both nullable the sql will just be:
`[NOT] EXISTS (SELECT 1 WHERE x = y)`
which should have the same execution plan in most RDBMS as the `[NOT] IN` expression.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
...
 
-					if (ne is ISqlPredicate pred1)
-						ne = visitor.Context.Optimizer.OptimizePredicate(pred1, visitor.Context.OptimizationContext.Context, visitor.Context.DataOptions);
+					switch (ne)
+					{
+						case ISqlPredicate pred1:
+							ne = visitor.Context.Optimizer.OptimizePredicate(pred1, visitor.Context.OptimizationContext.Context, visitor.Context.DataOptions);
+							break;
+
+						case SqlCondition(true, SqlSearchCondition([(false, SqlPredicate.InSubQuery({CanBeNull: true}, false, {Select.Columns : [{CanBeNull: false}]}) insq, var isOr1) cond]), var isOr)
```
jods4, line 2154: What guarantees that it will always be `SqlCondition(isNot: true, SqlSearchCondition [ SqlCondition (isNot: false, InSubQuery(isNot: false))])` and not any other combination of `isNot` such as (false, false, true) ?

These `SqlCondition` are horrible to work with, maybe if there's normalization before this step it would be nice to point out the logic in a comment.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
...
+							when visitor.Context.Optimizer.SqlProviderFlags.DoesNotSupportCorrelatedSubquery:
+							return new SqlCondition(false, new SqlSearchCondition().Expr(insq.Expr1).IsNull.Or.Add(new (true, insq, isOr1)), isOr);
+
+						case SqlCondition(false, SqlSearchCondition([(false, SqlPredicate.InSubQuery({CanBeNull: true}, false, {Select.Columns : [{CanBeNull: true} col]} subQuery) insq, _) cond]), var isOr)
+							when visitor.Context.Optimizer.SqlProviderFlags.DoesNotSupportCorrelatedSubquery :
+						{
+							var newQuery = subQuery.Convert((subQuery,col.Expression), static (v, e) =>
+							{
+								if (ReferenceEquals(e, v.Context.Expression))
```
jods4, line 2163: I understand the goal of replacing `select x` by `select 1` here, but I'm wary of a sub-expression being cached and re-used here, i.e. that the instance `v.Context.Expression` could be found not only in SELECT but also in WHERE or some other part of the query.

I would check the parent to ensure we're modifying `insq` and not part of the query.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
...
+								}
+
+								return e;
+							});
+
+							return new SqlCondition(
+								false,
+								new SqlSearchCondition()
+									.Expr(insq.Expr1).IsNull.   And.Expr(new SqlValue(1)).InSubQuery(newQuery).Or
+									.Expr(insq.Expr1).IsNotNull.And.Add(cond), isOr);
```
jods4, line 2182: This is turning `x? in (select y?)` into:
```
x is null and exists(select 1 where y is null)
or
x is not null and x in exists(select y)
```

This is all correct but it is a massive change compared to `x in (select y)`.
I'm good with that because I don't use `CompareNullsAsValues`; just wanted to point it out.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -2142,11 +2141,78 @@ protected virtual ISqlExpression ConvertFunction(SqlFunction func)
 				static (visitor, e) =>
 				{
 					var ne = e;
+
 					if (ne is ISqlExpression expr1)
 						ne = visitor.Context.Optimizer.OptimizeExpression(expr1, visitor);
 
-					if (ne is ISqlPredicate pred1)
-						ne = visitor.Context.Optimizer.OptimizePredicate(pred1, visitor.Context.OptimizationContext.Context, visitor.Context.DataOptions);
+					switch (ne)
```
jods4, line 2148: Maybe I lack caller context: I don't see what makes us not run this code when `CompareNullsAsValues` is false?

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
igor-tkachev: Are you sure it is going to work?

jods4: Are these not equivalent?
```cs
var newQuery = subQuery.Convert((subQuery, col, ex1), static (v, e) => 
{
  if (e is SqlWhereClause w && w == v.Context.subQuery.Where)
  {
    var wc = new SqlWhereClause(new SqlSearchCondition(w.SearchCondition.Conditions));
    //...
  }
});

// vs

var newQuery = subQuery.Convert((subQuery.Where, col, ex1), static (v, e) => 
{
  if (ReferenceEquals(e, v.Context.Where))
  {
    var wc = new SqlWhereClause(new SqlSearchCondition(v.Context.Where.SearchCondition.Conditions));
    //...
  }
});
```
I don't see in which case they work differently.
If there's no difference, the second one is more performant for sure.

jods4: Although, looking at this again, two things bother me:
- I don't like looking for sub-expressions by reference, because sub-expressions could be reused in multiple places (even though it's unlikely in this case -- it's just a bad pattern);
- We do a full Convert visitor just to change the WHERE of the SubQuery instance we already have.

Can't you replace that Convert with just a new SubQuery?
(we need a nicer API for that 🙁)
```cs
var newWhere = new SqlWhereClause();
//... same logic as in Converter

var newQuery = new SqlQuery();
newQuery.Init(
  subQuery.Select,
  subQuery.From,
  newWhere,
  subQuery.GroupBy,
  // ...
);
```

igor-tkachev: Lets see if it works.

igor-tkachev: We do not need a new API as we can just rewrite only WHERE clause. I tried to find any related issues and do not see them so far. So, lets try.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
igor-tkachev: I like the idea. However, lets create a new PR to implement it.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
igor-tkachev: It covered by tests.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
igor-tkachev: All other combinations work fine. This is just a heuristic.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
jods4: That doesn't really answer the question!
I don't like that when I look at this code, I am not aware that this runs only under specific options.

Looking around, I have the impression that it is at the time of `Contains` translation that the `CompareNullsAsValues` flag is considered. This would pass simple tests, _but_ it is not satisfactory.

I could create a similar expression tree in other ways and I still don't want these transformations to take place.

As this is just for C# semantics, please wrap them in a big `if (CompareNullsAsValues)`. It will be obvious, and run faster when using SQL semantics as no pattern matching needs to be attempted.

Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
igor-tkachev: It is not working.

jods4: Ok. It's surprising, can you explain what's the difference between the two?

igor-tkachev: The problem is that for this test we call optimization two times. First time we modify WHERE clause of this subQuery and just ignore it. Second time, we do the same modification of the WHERE clause of the same subQuery and add the same additional condition to that. As a result we have doubled condition. 

Current implementation ignores the whole subQuery first time and second time we create new subQuery with single condition. 

I do not like the fact that we call optimizations multiple times either, but this is subject for another PR.

jods4: ok, I kind of see the problem. Tricky.

Wouldn't be an issue if we used `new SubQuery().Init(...)` instead of a Convert visitor.

MaceWindu (approved):

viceroypenguin (approved):

PR #4201, opened 2023-07-11
Fix association fields mapping.

igor-tkachev:
```c#
[Association(ThisKey = nameof(DisTypeID), OtherKey = nameof(DisTypeTable.DisTypeID), CanBeNull = false)]
```

if `DisTypeID` and `DisTypeTable.DisTypeID` have different types, an exception is generated.

Worked before.

### DIFF
--- a/Source/LinqToDB/Common/ConvertBuilder.cs
+++ b/Source/LinqToDB/Common/ConvertBuilder.cs
@@ -16,7 +16,7 @@ namespace LinqToDB.Common
 
 	public static class ConvertBuilder
 	{
-		internal static readonly MethodInfo _defaultConverter = MemberHelper.MethodOf(() => ConvertDefault(null!, typeof(int)));
+		internal static readonly MethodInfo DefaultConverter = MemberHelper.MethodOf(() => ConvertDefault(null!, typeof(int)));
 
 		static object ConvertDefault(object value, Type conversionType)
 		{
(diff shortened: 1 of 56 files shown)


Source/LinqToDB/Common/ConvertBuilder.cs
```diff
@@ -16,7 +16,7 @@ namespace LinqToDB.Common
 
 	public static class ConvertBuilder
 	{
-		static readonly MethodInfo _defaultConverter = MemberHelper.MethodOf(() => ConvertDefault(null!, typeof(int)));
+		internal static readonly MethodInfo _defaultConverter = MemberHelper.MethodOf(() => ConvertDefault(null!, typeof(int)));
```
MaceWindu, line 19: `_defaultConverter ` -> `DefaultConverter `

MaceWindu (changes requested):

MaceWindu:
Access tests fail

Source/LinqToDB/Common/ConvertBuilder.cs
igor-tkachev: fixed

Tests/Linq/UserTests/AssociationTests.cs
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		[Table]
+		class JurTable
+		{
+			[Column(Length = 2), PrimaryKey, NotNull]  public string JurCode { get; set; } = null!;
+		}
+
+		[Test]
+		public void Test([DataSources] string context)
```
MaceWindu, line 44: it would be better to execute query against database so we can validate generated sql and have baselines

MaceWindu (changes requested):

Tests/Linq/UserTests/AssociationTests.cs
igor-tkachev: fixed

MaceWindu (approved):

MaceWindu:
btw, enum join condition looks awful...

PR #4203, opened 2023-07-11
BUG: Using hint in subquery.

igor-tkachev:
The following code

```c#
from c in db.Child
select new
{
	cc =
	(
		from c1 in db.Child
		.AsSqlServer().WithNoLock()
		where c1.ParentID == c.ParentID select c1
	).Count()
}
```

generates an exception

```
value(Tests.Model.TestDataConnection).GetTable().AsSqlServer().TableHint("NoLock").Where(c1 => (c1.ParentID == c.ParentID)).Count()'
cannot be converted to SQL.
```

### DIFF
--- a/Source/LinqToDB/Common/Internal/MethodCache.cs
+++ b/Source/LinqToDB/Common/Internal/MethodCache.cs
@@ -2,11 +2,11 @@
 using System.Collections.Concurrent;
 using System.Reflection;
 
-using LinqToDB.Expressions;
-using LinqToDB.Extensions;
-
 namespace LinqToDB.Common.Internal
 {
+	using Expressions;
+	using Extensions;
+
 	static class MemberCache
 	{
 		static MemberCache()
@@ -19,8 +19,9 @@ namespace LinqToDB.Common.Internal
 			_cache.Clear();
 		}
 
-		static readonly ConcurrentDictionary<MemberInfo,Info> _cache   = new();
-		static readonly Info                                  _default = new();
+		static readonly ConcurrentDictionary<MemberInfo, Info> _cache           = new();
+		static readonly Info                                   _defaultInfo     = new();
+		static readonly Info                                   _isQueryableInfo = new() { IsQueryable = true };
 
 		public static Info GetMemberInfo(MemberInfo member)
 		{
@@ -32,16 +33,16 @@ namespace LinqToDB.Common.Internal
 				{
 					var attrs = m.GetAttribute<IsQueryableAttribute>();
 
-					return attrs != null ? new() { IsQueryable = true } : _default;
+					return attrs != null ? _isQueryableInfo : _defaultInfo;
 				});
 			}
 
-			return _default;
+			return _defaultInfo;
 		}
 
 		public class Info
 		{
-			public bool IsQueryable;
+			public bool IsQueryable { get; init; }
 		}
 	}
 }
--- a/Source/LinqToDB/Common/Internal/MethodCache.cs
+++ b/Source/LinqToDB/Common/Internal/MethodCache.cs
@@ -2,6 +2,8 @@
 using System.Collections.Concurrent;
 using System.Reflection;
 
+using LinqToDB.Linq;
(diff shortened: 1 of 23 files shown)


Source/LinqToDB/Common/Internal/MethodCache.cs
```diff
@@ -0,0 +1,47 @@
...
+				});
+			}
+
+			return _default;
+		}
+
+		public class Info
+		{
+			public bool IsQueryable;
+		}
```
viceroypenguin, line 45:
```suggestion
		class Info
		{
			public bool IsQueryable { get; set; }
		}
```

Source/LinqToDB/Common/Internal/MethodCache.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		public static Info GetMemberInfo(MemberInfo member)
+		{
+			if (member is MethodInfo { IsGenericMethod: true } mi)
+			{
+				var gm = mi.GetGenericMethodDefinitionCached();
+
+				return _cache.GetOrAdd(gm, static m =>
+				{
+					var attrs = m.GetAttribute<IsQueryableAttribute>();
```
MaceWindu, line 34: use `HasAttribute` API instead of `GetAttribute` + length check

Source/LinqToDB/Common/Internal/MethodCache.cs
```diff
@@ -0,0 +1,48 @@
...
+	using Extensions;
+
+	static class MemberCache
+	{
+		static MemberCache()
+		{
+			//Query.CacheCleaners.Enqueue(ClearCache);
+		}
+
+		public static void ClearCache()
```
MaceWindu, line 17: not called by linq2db and unreachable to user as cache class is internal. I propose to make class public

Source/LinqToDB/Common/Internal/MethodCache.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Reflection;
+
+namespace LinqToDB.Common.Internal
+{
+	using Expressions;
+	using Extensions;
+
+	static class MemberCache
+	{
+		static MemberCache()
```
MaceWindu, line 12: not needed? or cache cleaners registration should be enabled?

MaceWindu (changes requested):
Could you provide more details what wasn't working properly?

Source/LinqToDB/Expressions/InternalExtensions.cs
```diff
@@ -313,7 +313,8 @@ public static bool IsQueryable(this MethodCallExpression method, bool enumerable
 				enumerable && type == typeof(Enumerable) ||
 				type == typeof(LinqExtensions) ||
 				type == typeof(DataExtensions) ||
-				type == typeof(TableExtensions);
+				type == typeof(TableExtensions) ||
+				MemberCache.GetMemberInfo(method.Method).IsQueryable;
```
igor-tkachev, line 317: The fix.

Source/LinqToDB/Common/Internal/MethodCache.cs
igor-tkachev: I am not sure we need it at all.

Source/LinqToDB/Common/Internal/MethodCache.cs
igor-tkachev: done

Source/LinqToDB/Common/Internal/MethodCache.cs
igor-tkachev: done

Source/LinqToDB/Common/Internal/MethodCache.cs
igor-tkachev: fixed

MaceWindu (approved):

MaceWindu (approved):

PR #4204, opened 2023-07-11
BUG: GroupBy with subquery.

igor-tkachev:
The following code 

```c#
from pmp in
(
	from pmp  in db.Child
	group pmp by pmp.ParentID into g
	select g.Key
)
from pmp1 in db.Child
select new { pmp1.ChildID }
```

generates wrong SQL

```sql
SELECT
	[pmp1].[ChildID]
FROM
	[Child] [pmp],
	[Child] [pmp1]
GROUP BY
	[pmp].[ParentID]
```

should be 

```sql
SELECT
	[pmp1].[ChildID]
FROM
	(
		SELECT
			[t1].[ParentID]
		FROM
			[Child] [t1]
		GROUP BY
			[t1].[ParentID]
	) [pmp],
	[Child] [pmp1]
```

### DIFF
--- a/Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
+++ b/Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
@@ -1508,13 +1508,18 @@ namespace LinqToDB.SqlQuery
 		{
 			CorrectCrossJoinQuery(_selectQuery);
 
-			for (var i = 0; i < _selectQuery.From.Tables.Count; i++)
-			{
-				var table = OptimizeSubQuery(_selectQuery, _selectQuery.From.Tables[i], true, false, isApplySupported, true, null);
+			var tableSources = _selectQuery.From.Tables;
 
-				if (table != _selectQuery.From.Tables[i])
+			for (var i = 0; i < tableSources.Count; i++)
+			{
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/SqlQuery/SelectQueryOptimizer.cs
```diff
@@ -1508,21 +1508,26 @@ void OptimizeSubQueries(bool isApplySupported)
 		{
 			CorrectCrossJoinQuery(_selectQuery);
 
-			for (var i = 0; i < _selectQuery.From.Tables.Count; i++)
+			var tableSources = _selectQuery.From.Tables;
+
+			for (var i = 0; i < tableSources.Count; i++)
 			{
-				var table = OptimizeSubQuery(_selectQuery, _selectQuery.From.Tables[i], true, false, isApplySupported, true, null);
+				if (tableSources.Count > 1 && tableSources[i] is { Source: SelectQuery { GroupBy.IsEmpty: false }})
```
igor-tkachev, line 1515: Fix.

sdanyliv:
I would suggest to use `AassertQuery` to validate grouping result.

MaceWindu (dismissed):
Could you specify what was broken? And I will agree with @sdanyliv - we should have assertions as from time to time we discover tests that doesn't fail but generated sql doesn't match intended logic

MaceWindu (approved):

PR #4217, opened 2023-07-20
Nullable bool comparison

igor-tkachev:


Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1606,19 +1606,21 @@ ISqlPredicate ConvertCompare(IBuildContext? context, ExpressionType nodeType, Ex
...
+			if (op is SqlPredicate.Operator.Equal or SqlPredicate.Operator.NotEqual)
 			{
 				bool?           value      = null;
 				ISqlExpression? expression = null;
 				var             isNullable = false;
+
 				if (IsBooleanConstant(left, out value))
 				{
-					isNullable = typeof(bool?) == left.Type || r.CanBeNull;
+					isNullable = value == null && (typeof(bool?) == left.Type || r.CanBeNull);
```
jods4, line 1618: It seems to me this extra check has no effect.

If `value` is null, then the next `if (value != null && ...)` block is not entered and the value of `isNullable` is never read.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -1606,19 +1606,21 @@ ISqlPredicate ConvertCompare(IBuildContext? context, ExpressionType nodeType, Ex
...
 					expression = r;
 				}
 				else if (IsBooleanConstant(right, out value))
 				{
-					isNullable = typeof(bool?) == right.Type || l.CanBeNull;
+					isNullable = value == null && (typeof(bool?) == right.Type || l.CanBeNull);
```
jods4, line 1623: Same here, for the same reason as previous comment.

Tests/Linq/Linq/WhereTests.cs
```diff
@@ -1991,5 +1991,58 @@ public void Issue1767Test2([DataSources(false)] string context)
...
+			AssertQuery(
+				from t in tt
+				where t.Value == true
+				select t);
+
+			var sql = GetCurrentBaselines();
+
+			Assert.That(sql,
+				Contains.Substring("[t].[Value] = 1").And.Not.
+				Contains("[t].[Value] = 1 AND [t].[Value] IS NOT NULL"));
```
jods4, line 2020: I don't quite like checking syntax because it's fragile. 
I'd rather test that the outcome for every possible case amongst true/false/null, is correct.

The problem with syntax is that it's easily gonna break:
- when we use this test with another DB provider that doesn't use `[t]` as name quotes;
- when the sql formatting changes and line breaks are inserted
- when for some reason the sql generation changes and inverts the AND order.
- when we upgrade to a better syntax such as `t.Value IS NOT FALSE` (supported in dbs such as Postgres or Oracle 23).

Tests/Linq/Linq/WhereTests.cs
```diff
@@ -1991,5 +1991,58 @@ public void Issue1767Test2([DataSources(false)] string context)
...
 			}
 		}
+
+		class NullableBoolClass
+		{
+			public bool? Value;
+		}
+
+		[Test]
+		public void NullableBoolTest1([DataSources(TestProvName.AllSybase)] string context)
```
jods4, line 2001: Why test Sybase only? 
If you run the query and check the results, we could verify this works in all providers.
It would be a good thing given bools can be represented very differently depending on provider.

igor-tkachev: Except Sybase. Sybase does not allow null values for BIT columns.

Directory.Packages.props
```diff
@@ -46,7 +46,7 @@
 		<PackageVersion Include="Net.IBM.Data.Db2"                                      Version="7.0.0.200"     />
 		<PackageVersion Include="Net.IBM.Data.Db2-lnx"                                  Version="7.0.0.200"     />
 		<PackageVersion Include="Net.IBM.Data.Db2-osx"                                  Version="7.0.0.200"     />
-		<PackageVersion Include="Octonica.ClickHouseClient"                             Version="2.2.9"         />
+		<PackageVersion Include="Octonica.ClickHouseClient"                             Version="2.3.2"         />
```
MaceWindu, line 49: don't update Octonica provider. It is fixed on 2.2.9 as newer versions contain regressions

MaceWindu (changes requested):
not ready, tests failing

Tests/Linq/Linq/NullableBoolTests.cs
```diff
@@ -0,0 +1,287 @@
...
+
+			AreEqual(
+				value switch { null => [data[1], data[2]], true => [data[0], data[2]], false => [data[0], data[1]] }
+				,
+				from t in tt
+				where t.Value != value
+				select t);
+
+			if (value is not null)
+				Assert.That(LastQuery, Contains.Substring("IS NULL"));
```
jods4, line 202: Can you please add tests (and support if needed) for `CompareNullsAsValue: false`?

I'd like C# `t.Value != true` to compile to SQL `value <> True`, like every other type does without `CompareNullsAsValue`.

igor-tkachev: It needs more development as `bool` comparison is hardly hardcoded by some reason.

Directory.Packages.props
```diff
@@ -47,7 +47,7 @@
...
```
MaceWindu, line 50: Still not reverted. versions > 2.2.9 are broken

Tests/Linq/Linq/NullableBoolTests.cs
jods4: ok it needs dev. but why did you resolve the comment, what's the plan?

please do not release half the feature. If you do, then implementing `CompareNullsAsValue: false` later will be a breaking change (which we _really_ want to make).

igor-tkachev: We have two topics here:
- `CompareNullsAsValue: false` - I added this parameter to the test.
- `t.Value != true` - no plans until new parser release. Anyway, this is out of scope of this PR. You can create new issue/PR to remember it.

jods4: Ok 👍
No problem with adding comparisons to bools later, I just wanted to ensure that whatever is in the PR works in all configurations.

igor-tkachev: You can always add any additional tests to any PRs you are interested in.
Can I resolve it now? :)

jods4: I'm sorry I don't have much time right now to contribute.
I see `NotNullableTest` tests `compareNullsAsValues` but it's not meaningful as the source is non-nullable so there's no change in codegen between the two modes.

Can please you add `compareNullsAsValues` to this test (`ValueTest`)  so that nullable values and differences in codegen are covered? After that the PR is good for me 👍

sdanyliv: Why `t.Value != true` to `value <> True`? I know that I have implemented code which always translates to equal operator, in this case `value == False`

jods4: @sdanyliv for bools sure the outcome is indeed the same, so is there any reason why you made that flip?
I like that the SQL is a straightforward translation of my code, as much as possible: if I write `x != true` I expect `x <> True` in SQL and if I write `x == false` I expect `x = False`, although as I said: it doesn't matter much as both expressions are equivalent.

Note that similar transformations do not apply to `IS` operators. RDBMS that supports boolean types usually also support `IS [NOT] [TRUE|FALSE|UNKNOWN]` (where `NULL` is often supported instead of `UNKNOWN`).
In this context, `IS NOT TRUE` is **not** equivalent to `IS FALSE`.

Tests/Linq/Linq/NullableBoolTests.cs
```diff
@@ -0,0 +1,289 @@
...
+{
+	[TestFixture]
+	public class NullableBoolTests : TestBase
+	{
+		record NullableBoolClass
+		{
+			public bool? Value;
+		}
+
+		const string ProvidersThatDoNotSupportNullableBool = $"{TestProvName.AllAccess},{TestProvName.AllSybase},{ProviderName.ClickHouseOctonica}";
```
MaceWindu, line 18:
```suggestion

```suggestion

		// octonica blocked by https://github.com/Octonica/ClickHouseClient/issues/56
		const string ProvidersThatDoNotSupportNullableBool = $"{TestProvName.AllAccess},{TestProvName.AllSybase},{ProviderName.ClickHouseOctonica}";
```
				const string ProvidersThatDoNotSupportNullableBool = $"{TestProvName.AllAccess},{TestProvName.AllSybase},{ProviderName.ClickHouseOctonica}";
```

File: Tests/Linq/Linq/NullableBoolTests.cs [hunk 9]
```diff
@@ -0,0 +1,292 @@
...
+	public class NullableBoolTests : TestBase
+	{
+		record NullableBoolClass
+		{
+			public bool? Value;
+		}
+
+```suggestion
+
+		// octonica blocked by https://github.com/Octonica/ClickHouseClient/issues/56
```
MaceWindu, line 20: added comment to unblock testing when Octonica provider becames usable again

MaceWindu: and yes, this specific issue fixed (partially) in 2.2.10, but more issues that cannot be workarounded were introduced

Tests/Linq/Linq/NullableBoolTests.cs
MaceWindu: I'm not sure I like this discussion :) If we have more cases to cover in future,  it makes sense to add tests now and mark them with `ActiveIssue`

MaceWindu (dismissed):

Tests/Linq/Linq/NullableBoolTests.cs
igor-tkachev: > Why `t.Value != true` to `value <> True`? I know that I have implemented code which always translates to equal operator, in this case `value == False`

I think user should get what he writes. If user writes `Field != false`, this is his business and he should get the same SQL. The magic you added can be unwanted, should not be default, and should be implemented only with compatibility flag.

MaceWindu (approved):

MaceWindu:
@linq2db/developers any reviews pending or it could be merged?

viceroypenguin (approved):

PR #4146, opened 2023-05-27
Implement parameter names normalization

MaceWindu:
Fix #4144
Fix #3902

Rewrite parameter name generation to generate unique names with valid characters and length. As side-effect removes a lot of unnecessary parameter juggling code.

PR #4230, opened 2023-07-30
Add tests and fix for UniqueParametersNormalizer

### LINKED ISSUE #4254: Parameters name conflict when using Expressions with PostgreSQL
Good day!

It seems there is some kind of query parameters name conflict when heavelly using Expression. I used parameters `userId` and `now` in expression and helpers method and I got exception:

```
Exception message:
`Npgsql.PostgresException: '42601: syntax error at or near ":"`
```

When I change params to `userId2` and `now2` in `Expression` or `GetQuery` method this starts to work.

Current workaround for me: ensure all parameters have unique names. This is not robust solution, but for a moment it works.

### Steps to reproduce

```c#

using LinqToDB;
using LinqToDB.Data;
using LinqToDB.DataProvider.PostgreSQL;
using LinqToDB.Mapping;

static Expression<Func<MediaItem, MediaItemSearchSharedResult>> MediaItemSearchSharedResultProjection(DateTime now)
{
   // If change now -> now2 query will work
    return x => new MediaItemSearchSharedResult
    {
        CategoryIds = x.Categories.Select(y => y.CategoryId).ToList(),
        IsUnvisited = x.UserShare.Any(y => y.ExpiresAt > now)
    };
}

static IQueryable<MediaItem> GetQuery(DbContext context, Guid userId, DateTime now)
{
    return context.GetTable<MediaItem>()
        .Where(x =>
            x.UserShare.Any(y => y.UserId == userId && y.ExpiresAt > now) ||
            x.UserShare.Any(y => y.CreatedById == userId && y.ExpiresAt > now));
}

    .UsePostgreSQL("connection string", PostgreSQLVersion.v15)
    .UseMappingSchema(LinqToDB.Mapping.MappingSchema.Default);

{
    using var context = new DbContext(options);

    var now = DateTime.Now;
    var userId = Guid.NewGuid();

    var result = GetQuery(context, userId, now)
            .Select(MediaItemSearchSharedResultProjection(now))
            .ToList();
}

{
    using var context = new DbContext(options);

    var now = DateTime.Now;
    var userId = Guid.NewGuid();

    var result = GetQuery(context, userId, now)
            .Select(MediaItemSearchSharedResultProjection(now))
            .ToList();
}

Console.WriteLine("Done");

[Table("media_item_to_media_item_categories")]
internal sealed class MediaItemToMediaItemCategory
{
    [Column("id")]
    [PrimaryKey]
    public Guid Id { get; set; }

    [Column("category_id")]
    public Guid CategoryId { get; set; }

    [Column("media_item_id")]
    public Guid MediaItemId { get; set; }
}

[Table("media_item_user_share")]
internal sealed class MediaItemUserShare
{
    [Column("id")]
    [PrimaryKey]
    public Guid Id { get; set; }

    [Column("media_item_id")]
    public Guid MediaItemId { get; set; }

    [Column("created_by_id")]
    public Guid CreatedById { get; set; }

    [Column("user_id")]
    public Guid UserId { get; set; }

    [Column("expires_at")]
    public DateTime ExpiresAt { get; set; }
}

[Table("media_items")]
internal sealed class MediaItem
{
    [Column("id")]
    [PrimaryKey]
    public Guid Id { get; set; }

    [Association(ThisKey = nameof(Id), OtherKey = nameof(MediaItemToMediaItemCategory.MediaItemId))]
    public IList<MediaItemToMediaItemCategory> Categories { get; set; } = new List<MediaItemToMediaItemCategory>();

    [Association(ThisKey = nameof(Id), OtherKey = nameof(MediaItemUserShare.MediaItemId))]
    public IList<MediaItemUserShare> UserShare { get; set; } = new List<MediaItemUserShare>();
}

public sealed class MediaItemSearchSharedResult
{
    public IList<Guid> CategoryIds { get; set; } = new List<Guid>();

    public bool IsUnvisited { get; set; }
}

internal sealed class DbContext : DataConnection
{
    public DbContext(DataOptions options) : base(options)
    {
        (this as IDataContext).CloseAfterUse = true;

        CommandTimeout = (int)TimeSpan.FromMinutes(5).TotalSeconds;
    }
}
```

Wrong SQL seen in logs:
```sql
2023-08-22 16:34:08.050 +07 [17180] ERROR:  syntax error at or near ":" at character 281
2023-08-22 16:34:08.050 +07 [17180] STATEMENT:  SELECT

		key_data_result.id,

		detail.category_id

	FROM

		(

			SELECT DISTINCT

				x.id

			FROM

				media_items x

			WHERE

				(EXISTS(

					SELECT

						*

					FROM

						media_item_user_share y

					WHERE

						x.id = y.media_item_id AND y.user_id = $1 AND

						y.expires_at > :now

				) OR EXISTS(

					SELECT

						*

					FROM

						media_item_user_share y_1

					WHERE

						x.id = y_1.media_item_id AND y_1.created_by_id = $1 AND

						y_1.expires_at > :now

				))

		) key_data_result

			INNER JOIN media_item_to_media_item_categories detail ON key_data_result.id = detail.media_item_id
```

### Environment details

`Linq To DB` version: 5.2.2

Database (with version): PostgreSQL 15

ADO.NET Provider (with version): Npgsql 7.0.4

Operating system: Windows 10

.NET Version: 7.0

Shane32:
Fix #4254

This PR contains the following changes to `UniqueParametersNormailzer`:
- Fixes exception when attempting to shorten duplicated input names that are longer than `MaxLength` (default of 50 chars) - (see [here](https://github.com/linq2db/linq2db/pull/4146#discussion_r1278475938))
- Patches an infinite loop in the case where `MaxLength` is very short and the loop cannot find a unique parameter name
- Optimizes the `MakeValidName` method to use `stackalloc` rather than using the `ObjectPool` to pull and use a `StringBuilder` instance.  Although perhaps there are no allocations either way (except for the final string allocation), it should be faster, as there are no locks.  For older TFMs `stackalloc` is not used, as it requires a NuGet reference to `System.Memory`, and `string.ToCharArray()` is used, which causes an allocation.  (It still should be faster than an object pool for short strings, based on my experiences with GraphQL.NET, but I did not run benchmarks for this exact scenario.)
- Moved `MakeValidName` prior to the loop, eliminating the requirement that the method not add characters (or else infinite loop), but assuming that for any valid parameter name, any truncated version is also valid
- Added a great variety of unit tests

Additional notes:
- The code assumes that `DefaultName` and `MaxLength` are valid (e.g. where `DefaultName` is not `null`, an empty string, or a string which is not a valid parameter name).  Tests are in place to ensure that an infinite loop or stack overflow exception do not occur in those cases, but otherwise the results are undefined.
- The implementation converts `null` to the value of `DefaultName`.  If this is expected behavior, why is the return type nullable?  Perhaps because `NoopParameterNormalizer` returns the input value without changes (including returning `null` for `null`).  So then should not `UniqueParametersNormalizer` return `null` when the input is `null`?  Or we could ask why is the input value nullable?  It would seem that the behavior of `null` should be consistent and defined in the interface -- either it should return `null` for `null` or it should return a string for `null`.  Just a thought; this PR does not address that, and it is likely that there is a valid reason here.
- Any reason `Normalize` is not marked `virtual`?  Not a big deal since a derived class can implement `IQueryParametersNormalizer` on the derived class, but I'm not sure why that choice was made whereas most of the other methods are `protected virtual`.

Lmk if it is preferred to revert any of these changes, or to post them as separate PRs.


MaceWindu (approved):
👍

MaceWindu:
> The implementation converts null to the value of DefaultName. If this is expected behavior, why is the return type nullable? Perhaps because NoopParameterNormalizer returns the input value without changes (including returning null for null). So then should not UniqueParametersNormalizer return null when the input is null? Or we could ask why is the input value nullable? It would seem that the behavior of null should be consistent and defined in the interface -- either it should return null for null or it should return a string for null. Just a thought; this PR does not address that, and it is likely that there is a valid reason here.

Yep, because:
- input could be null at least according to existing anntations and it is not easy to track down wether we can get null here or not
- output is null also only because of `NoopParameterNormalizer`. This is acceptable only for this normalizer instance because it is used with providers which doesn't use provider names. Feel free to improve annotations if you have ideas how to do it :)

> Any reason Normalize is not marked virtual? Not a big deal since a derived class can implement IQueryParametersNormalizer on the derived class, but I'm not sure why that choice was made whereas most of the other methods are protected virtual.

No big reason, just don't see value in virtual here currently

Shane32:
Looks like some of the tests failed on the "checkout test baselines" step.  Is there a way to re-run the failed tests?

MaceWindu:
nope, currently our workflow not restartable if no baselines modified + we have storage issue with oracle 21/23 images, so they will not run anyways. You can run specific db pipeline (e.g. `/azp run test-oracle`), but as I said, it will fail for oracle

Shane32:
> The implementation converts null to the value of DefaultName ...

> ... Feel free to improve annotations if you have ideas how to do it :)

Well, we could add `[return: NotNullIfNotNull(nameof(originalName))]` - but it is not needed for the only use case, so I think it's good as-is.

igor-tkachev (approved):

Tests/Linq/DataProvider/UniqueParametersNormalizerTests.cs
```diff
@@ -0,0 +1,402 @@
...
+				_maxLength = maxLength;
+				_defaultName = defaultName;
+			}
+			protected override int MaxLength => _maxLength;
+			protected override string DefaultName => _defaultName;
+		}
+
+		// Ensures that "search" is always passed to IQueryParametersNormalizer.Normalize as the originalName
+		[Test]
+		public async Task CalledWithCorrectNames([DataSources(false)] string context)
```
Shane32, line 243: This test currently fails.  The normalizer should be called with "search" as the argument 3 times, but instead for the 3rd select query in the union, it is called with "search_1", so the generated names are "search", "search_1" and "search_1_1".

Tests/Linq/DataProvider/UniqueParametersNormalizerTests.cs
```diff
@@ -0,0 +1,402 @@
...
+					select row1.Id)
+				.Union(
+					from row1 in dbTable1
+					where row1.Field1!.StartsWith(search)
+					select row1.Id);
+		}
+
+		// Ensures that subsequent executions of the same query execute the same SQL
+		[Test]
+		public async Task ExecutesDeterministically([DataSources(false)] string context)
```
Shane32, line 274: This test currently fails.  Any query should execute the same way for subsequent executions (regardless of the behavior of the normalizer).  Here the second execution results in different SQL.

MaceWindu: I will look into it

MaceWindu: should be fixed now

MaceWindu (dismissed):
last commit probably will introduce some baselines noise, but all should work now

Shane32:
My test does not work on MySQL due to:

> You cannot refer to a TEMPORARY table more than once in the same query.

https://dev.mysql.com/doc/refman/8.0/en/temporary-table-problems.html

This would be proper, correct: `[DataSources(false, TestProvName.AllMySql)]`?

Shane32:
Looking at the baselines, many parameter names have been cleaned up, but also in a number of spots there are duplicated parameters.  Could it be that linq2db was deduplicating parameters previously and now isn't?

Shane32:
These tests are failing for all Oracle versions:
- XmlTableTest1
- XmlTableTest2
- XmlTableTest3
- XmlTableTest7
- XmlTableTest9

Errors look like this:

```
Error Message:
   System.ArgumentException : Value does not fall within the expected range.
DataConnection: Total Execution Time (Completed): 00:00:00.0012149. Rows Count: 0.

Stack Trace:
   at Oracle.ManagedDataAccess.Client.OracleParameter.set_Value(Object value)
   at LinqToDB.DataProvider.DataProviderBase.SetParameter(DataConnection dataConnection, DbParameter parameter, String name, DbDataType dataType, Object value) in D:\a\1\s\Source\LinqToDB\DataProvider\DataProviderBase.cs:line 396
   at LinqToDB.DataProvider.Oracle.OracleDataProvider.SetParameter(DataConnection dataConnection, DbParameter parameter, String name, DbDataType dataType, Object value) in D:\a\1\s\Source\LinqToDB\DataProvider\Oracle\OracleDataProvider.cs:line 243
   at LinqToDB.Data.DataConnection.QueryRunner.CreateParameter(DataConnection dataConnection, DbCommand command, SqlParameter parameter, SqlParameterValue parmValue, Boolean forGetSqlText) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.QueryRunner.cs:line 313
   at LinqToDB.Data.DataConnection.QueryRunner.GetParameters(DataConnection dataConnection, PreparedQuery pq, IReadOnlyParameterValues parameterValues, Boolean forGetSqlText) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.QueryRunner.cs:line 283
   at LinqToDB.Data.DataConnection.QueryRunner.CreateExecutionQuery(DataConnection dataConnection, IQueryContext context, IReadOnlyParameterValues parameterValues, Boolean forGetSqlText) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.QueryRunner.cs:line 192
   at LinqToDB.Data.DataConnection.QueryRunner.SetQuery(IReadOnlyParameterValues parameterValues, Boolean forGetSqlText) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.QueryRunner.cs:line 323
   at LinqToDB.Linq.QueryRunnerBase.SetCommand(Boolean forGetSqlText) in D:\a\1\s\Source\LinqToDB\Linq\QueryRunnerBase.cs:line 72
   at LinqToDB.Data.DataConnection.QueryRunner.ExecuteReader() in D:\a\1\s\Source\LinqToDB\Data\DataConnection.QueryRunner.cs:line 622
   at LinqToDB.Linq.QueryRunner.ExecuteQuery[T](Query query, IDataContext dataContext, Mapper`1 mapper, Expression expression, Object[] ps, Object[] preambles, Int32 queryNumber)+MoveNext() in D:\a\1\s\Source\LinqToDB\Linq\QueryRunner.cs:line 400
   at Tests.DataProvider.OracleTests.XmlTableTest9(String context) in D:\a\1\s\Tests\Linq\DataProvider\OracleTests.cs:line 2112
```

MaceWindu:
@Shane32 could you add/verify regression test for #4254?

Shane32:
Probably. Let me take a look.

Shane32:
@MaceWindu Done.  I only use MS SQL, but I can verify that the added test fails on MS SQL under the `master` branch (similar error to the reported one) and passes under this branch.  The test is set up as `[IncludeDataSources(TestProvName.AllPostgreSQL95Plus, TestProvName.AllSqlServer2008Plus)]` so we can see if it passes with postgres also.

To replicate the user's scenario most closely, the data connection is closed and reopened multiple times during the test.  Due to this I am using non-temporary tables, named with a prefix of `issue_4254_`, which are dropped at the end of the test.  It is likely enough that I can simplify the test to use a single connection, and then use temporary tables/etc.  Let me know if I should change it - thanks!

MaceWindu:
@Shane32 thanks!

MaceWindu (approved):

MaceWindu (changes requested):
note to myself: check why deduplication of parameters doesn't work for some baselines

MaceWindu (approved):

PR #4232, opened 2023-07-30
Disable long parameter names for oracle12 dialect

### LINKED ISSUE #4219: Wrong Nameing Strategy for Oracle 12c Release <2
Linq2Db uses Longer Identifiers for Oracle 12, but these are only available in Orcale 12c Release 2 (see: https://oracle-base.com/articles/12c/long-identifiers-12cr2), releases before still need to use the Shorter names (from oracle 11)

MaceWindu:
Fix #4219

We could reenable them when we will add more dialects to Oracle provider. Parameter names longer 30 characters doesn't sound right thing to require new dialect support.

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/Oracle11ParametersNormalizer.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/Oracle11ParametersNormalizer.cs
@@ -1,6 +1,6 @@
 ﻿namespace LinqToDB.DataProvider.Oracle
 {
-	public sealed class Oracle11ParametersNormalizer : Oracle12ParametersNormalizer
+	public sealed class Oracle11ParametersNormalizer : Oracle122ParametersNormalizer
 	{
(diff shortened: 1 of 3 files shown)


Shane32 (approved):

viceroypenguin (approved):

igor-tkachev (approved):

jods4:
Features are available based on `OracleProvider`, which I believe reflects the ADO.NET provider name and package version, and an `OracleVersion` passed to the provider that reflects the Oracle server version.

Is this not a matter of creating `OracleVersion.v12_2` and only enabling the long parameter names if version is >= v12_2?

MaceWindu:
@jods4 , yes, we need to have `OracleVersion.v12_2` or higher to enable this feature which we don't have now. Increase in allowed parameter name length from 30 to 50 doesn't look attractive enough to add new OracleVersion entry.

PR #4233, opened 2023-07-30
Array parameter value logging

### LINKED ISSUE #4122: SQL Trace with string array parameter
I'm trying to debug a query that has a string array parameter but trace is not outputting array values just System.String[]

```
trce: LinqToDB[0]
      --  PostgreSQL.9.5 PostgreSQL
      DECLARE @services -2147483629 -- Object
      SET     @services = System.String[]
      DECLARE @take Integer -- Int32
      SET     @take = 1000
``` 
My guess is that ToString was used but I cannot override ToString of an array, and in my case it needs to be an array, not a list.

MaceWindu:
Fix #4122


igor-tkachev (approved):

Shane32 (dismissed):

jods4:
Bikeshedding: I think the default value of logging 50 values per array is kind of a lot (esp. compared to the current release 0).
I'd rather settle to a lower default such as 8, it's enough to get an idea what's in an array parameter and it has a lower impact.
If someone needs to log a lot of elements the option can be changed manually.

viceroypenguin:
> Bikeshedding: I think the default value of logging 50 values per array is kind of a lot (esp. compared to the current release 0). I'd rather settle to a lower default such as 8, it's enough to get an idea what's in an array parameter and it has a lower impact. If someone needs to log a lot of elements the option can be changed manually.

I agree. This matches behavior in other systems, like xUnit, which show only a portion of the array for descriptive purposes.

MaceWindu:
\azp run test-all

viceroypenguin (approved):

PR #4251, opened 2023-08-19
Update dependencies and fix analyzers

MaceWindu:
- update dependencies
- review new/updated analyzers messages, fix valid errors (if not breaking change), disable slow/unmportant analyzers

There are some nice finds:
- [CA1508](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1508) found this bug #4176


.editorconfig
```diff
@@ -164,28 +164,103 @@ dotnet_diagnostic.CA3076.severity = none # CA3076: Insecure XSLT Script Executio
...
+dotnet_diagnostic.CA1043.severity = none # CA1043: Use integral or string argument for indexers
 dotnet_diagnostic.CA1051.severity = none # CA1051: Do not declare visible instance fields
+dotnet_diagnostic.CA1054.severity = none # CA1054: URI parameters should not be strings
+dotnet_diagnostic.CA1062.severity = none # CA1062: Validate arguments of public methods
+dotnet_diagnostic.CA1063.severity = none # CA1063: Implement IDisposable correctly
+dotnet_diagnostic.CA1065.severity = none # CA1065: Do not raise exceptions in unexpected locations
 dotnet_diagnostic.CA1067.severity = none # CA1067: Override Equals when implementing IEquatable
 dotnet_diagnostic.CA1068.severity = none # CA1068: CancellationToken parameters must come last
 dotnet_diagnostic.CA1069.severity = none # CA1069: Enums should not have duplicate values
+dotnet_diagnostic.CA1303.severity = none # CA1303: Do not pass literals as localized parametersdotnet_diagnostic.CA1307.severity = none # CA1307: Specify StringComparison for clarity
```
viceroypenguin, line 193: accident

sdanyliv (approved):

igor-tkachev (approved):

Source/LinqToDB.Tools/CodeModel/Types/Parser/TypeParser.cs
```diff
@@ -2,6 +2,8 @@
 using System.Collections.Generic;
 using System.Text;
 
+using LinqToDB.Extensions;
```
viceroypenguin, line 5: Move inside `using` to match current style.

Source/LinqToDB.Tools/DataModel/DataModelGenerator.Functions.Procedures.cs
```diff
@@ -66,7 +66,7 @@ private static void BuildStoredProcedure(IDataModelGenerationContext context, St
 			{
 				// if procedure resultset schema load failed, generate error pragma
 				if (context.Options.GenerateProceduresSchemaError)
```
viceroypenguin, line 68: add braces here for clarity

Source/LinqToDB/Common/EnumerableHelper.cs
```diff
@@ -6,7 +6,10 @@
 
 namespace LinqToDB.Common
 {
+	// TODO: v6: make static
```
viceroypenguin, line 9: is it really going to break anything if you make it static now? no one will `new`ing one of these, and if they are, we don't really care?

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -734,7 +734,9 @@ public virtual void Close()
 
 		#region Command
 
+#pragma warning disable CA2213 // Disposable fields should be disposed
```
viceroypenguin, line 737: any reason not to throw a `_command?.Dispose()` in the dispose method?

Source/LinqToDB/DataContext.cs
```diff
@@ -262,7 +262,9 @@ public  int   CommandTimeout
 		/// <summary>
 		/// Underlying active database connection.
 		/// </summary>
+#pragma warning disable CA2213 // Disposable fields should be disposed
```
viceroypenguin, line 265: again, any reason not to throw a `_dataConnection?.Dispose()` in the dispose?

Source/LinqToDB/DataProvider/Access/AccessOleDbSchemaProvider.cs
```diff
@@ -8,14 +8,15 @@ namespace LinqToDB.DataProvider.Access
...
+	using Extensions;
 	using SchemaProvider;
 
 	sealed class AccessOleDbSchemaProvider : AccessSchemaProviderBase
 	{
-		private const OleDbProviderAdapter.ColumnFlags COUNTER_OR_BIT = OleDbProviderAdapter.ColumnFlags.MayBeNull
-			| OleDbProviderAdapter.ColumnFlags.IsFixedLength
-			| OleDbProviderAdapter.ColumnFlags.WriteUnknown
-			| OleDbProviderAdapter.ColumnFlags.MayDefer;
+		//private const OleDbProviderAdapter.ColumnFlags COUNTER_OR_BIT = OleDbProviderAdapter.ColumnFlags.MayBeNull
```
viceroypenguin, line 16: why keep and comment this code?

Source/LinqToDB/DataProvider/BulkCopyReader.cs
```diff
@@ -20,8 +20,12 @@ namespace LinqToDB.DataProvider
 
 	public class BulkCopyReader<T> : BulkCopyReader, IAsyncDisposable
 	{
+#if !NATIVE_ASYNC
+#pragma warning disable CA2213 // Disposable fields should be disposed
```
viceroypenguin, line 24: is CA2213 too sensitive?

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseProviderAdapter.cs
```diff
@@ -147,23 +147,32 @@ public static ClickHouseProviderAdapter GetInstance(ClickHouseProvider provider)
 			{
 				if (_octonicaAdapter == null)
 					lock (_octonicaSyncRoot)
+						// https://github.com/dotnet/roslyn-analyzers/issues/1649
```
viceroypenguin, line 150: add braces on all of these for clarity

Source/LinqToDB/DataProvider/Informix/InformixBulkCopy.cs
```diff
@@ -204,19 +204,23 @@ await using (enumerator.ConfigureAwait(Configuration.ContinueOnCapturedContext))
 			ITable<T> table, DataOptions options, IEnumerable<T> source, CancellationToken cancellationToken)
 		{
 #if NATIVE_ASYNC
-			await using (new InvariantCultureRegion(null))
+#pragma warning disable CA2000 // Dispose objects before losing scope
```
viceroypenguin, line 207: is CA2000 too sensitive here?

Source/LinqToDB/DataProvider/SQLite/SQLiteProviderAdapter.cs
```diff
@@ -131,15 +131,19 @@ public static SQLiteProviderAdapter GetInstance(string name)
 			{
 				if (_systemDataSQLite == null)
 					lock (_systemSyncRoot)
+#pragma warning disable CA1508 // Avoid dead conditional code
```
viceroypenguin, line 134: add braces here for clarity

Source/LinqToDB/DataProvider/SqlServer/SqlServerProviderAdapter.cs
```diff
@@ -119,15 +119,19 @@ public static SqlServerProviderAdapter GetInstance(SqlServerProvider provider)
 			{
 				if (_systemAdapter == null)
 					lock (_sysSyncRoot)
+#pragma warning disable CA1508 // Avoid dead conditional code
```
viceroypenguin, line 122: add braces for clarity

Source/LinqToDB/DataProvider/Sybase/SybaseProviderAdapter.cs
```diff
@@ -75,15 +75,19 @@ public static SybaseProviderAdapter GetInstance(string name)
 			{
 				if (_nativeInstance == null)
 					lock (_nativeSyncRoot)
+#pragma warning disable CA1508 // Avoid dead conditional code
```
viceroypenguin, line 78: add braces for clarity

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -486,7 +487,9 @@ public void SetExpression(LambdaExpression expression)
 			{
 				if (_members == null)
 					lock (_memberSync)
+#pragma warning disable CA1508 // Avoid dead conditional code
```
viceroypenguin, line 490: add braces for clarity

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System;
+﻿#pragma warning disable CA1862 // Prefer using StringComparer to perform a case-insensitive comparison
```
viceroypenguin, line 1: is this still necessary with the switch to `IndexOfEx()`?

Source/LinqToDB/Linq/ReflectionHelper.cs
```diff
@@ -11,7 +11,9 @@ namespace LinqToDB.Linq
 
 	sealed class ReflectionHelper
 	{
+#pragma warning disable CA1052 // Static holder types should be Static or NotInheritable
```
viceroypenguin, line 14: `ReflectionHelper` is `internal`, so `Expressor` should not have any external consumers. This one can definitely be made private without breaking concerns.

Source/LinqToDB/Common/EnumerableHelper.cs
MaceWindu: I don't mind to make it static, but people here usually opposing breaking changes in non-major release 😄 , so I decided to postpone it

Source/LinqToDB/Data/DataConnection.cs
MaceWindu: will recheck, but I think we don't have issues here

Source/LinqToDB/DataProvider/BulkCopyReader.cs
MaceWindu: It fails to detect non-trivial disposal logic (cannot blame it )

Source/LinqToDB/Linq/Expressions.cs
MaceWindu: it is triggered also for calls, used in expressions (main pain point of CA analyzers to forget exclude `Expression`-scoped code from analysis)

Source/LinqToDB/Linq/ReflectionHelper.cs
MaceWindu: It is used as base class below to resove generic argument... Refactored it

Source/LinqToDB/Data/DataConnection.cs
MaceWindu: Checked - nothing wrong here, we use DisposeCommand API to dispose it, so analyzer fails to detect it being disposed

Source/LinqToDB/DataContext.cs
MaceWindu: it is actually disposed, but in Close method, called by Dispose

Source/LinqToDB/DataProvider/Informix/InformixProviderAdapter.cs
```diff
@@ -170,15 +170,19 @@ public static InformixProviderAdapter GetInstance(string name)
 			{
 				if (_ifxAdapter == null)
 					lock (_ifxSyncRoot)
+#pragma warning disable CA1508 // Avoid dead conditional code
 						_ifxAdapter ??= CreateIfxAdapter();
+#pragma warning restore CA1508 // Avoid dead conditional code
```
Shane32, line 175: Maybe just ignore CA1508 throughout, if it is known that it doesn't apply for multithreaded code?

Source/LinqToDB/DataProvider/Access/AccessOleDbSchemaProvider.cs
MaceWindu: Wanted to keep it for now in case if I will try to fix it later. But agree, OLE DB schema APIs are broken and with OLEDB+ODBC schema merging in scaffold tool we don't need it anymore. Will remove

Source/LinqToDB/DataProvider/MySql/MySqlBulkCopy.cs
```diff
@@ -151,7 +151,14 @@ public MySqlBulkCopy(MySqlDataProvider provider)
...
 			foreach (var batch in EnumerableHelper.Batch(source, options.MaxBatchSize ?? int.MaxValue))
 			{
+#pragma warning disable CA2000 // Dispose objects before losing scope
 				var rd = new BulkCopyReader<T>(dataConnection, columns, batch);
+#pragma warning restore CA2000 // Dispose objects before losing scope
+#if NATIVE_ASYNC
+				await using var _ = rd.ConfigureAwait(Configuration.ContinueOnCapturedContext);
+#else
+				await using var _ = rd;
+#endif
```
Shane32, line 161: I don't know that it's any better, but this is probably valid:

```suggestion
				await using var rd = new BulkCopyReader<T>(dataConnection, columns, batch)
#if NATIVE_ASYNC
					.ConfigureAwait(Configuration.ContinueOnCapturedContext);
#else
					;
#endif
```

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
```diff
@@ -344,8 +344,8 @@ public override StringBuilder Convert(StringBuilder sb, string value, ConvertTyp
 				case ConvertType.NameToQueryTable     :
 				case ConvertType.NameToProcedure      :
 					// https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
-					if (value.Contains('`'))
-						value = value.Replace("`", "``");
+					if (value.ContainsEx('`'))
+						value = value.ReplaceEx("`", "``");
```
Shane32, line 348: Rather than change every use of Contains to ContainsEx (and similar), why do we not simply ignore the warning for the entire solution?  `Contains` by definition "performs an ordinal (case-sensitive and culture-insensitive) comparison" when no argument is specified - which is exactly what `ContainsEx` does.  Are there ANY places where we need to perform culture-sensitive comparisons?  Do we really need an analyzer to remind us to perform culture-sensitive comparisons when 99% of the use cases for this library are ordinal-based comparisons?

Source/LinqToDB/DataProvider/Informix/InformixProviderAdapter.cs
MaceWindu: I prefere to keep it. MS fixed a lot of issues with it last month (it was unusable before) and now it helped to find a lot of small flaws and one bug (which we already know about, but anyway).
I'm fine with having suppressions around double-lock pattern (and I hope MS will add support for it to analyzer in future)

Source/LinqToDB/DataProvider/MySql/MySqlSqlBuilder.cs
MaceWindu: Makes sense, will try to change it

Shane32:
I'm about halfway through reviewing this PR, and I see a constant theme with these changes:
- The analyzer fails to recognize double checked locking
- The analyzer assumes we prefer culture-specific string comparisons

As such, I suggest the solution is to disable those rules, not to add code throughout to disable the warning.  The first issue isn't a potential bug fix, it's just a suggestion for simplifying the code; an ignored warning cannot hurt the project.  And as for the second, if we don't want culture-specific string comparisons, we don't want them - simple.

I would leave the `IDisposable` warnings enabled.  Usually we are skipping them where we are using an internal wrapper class that we know shouldn't be disposed.  There's also some complications around `await using` that are trigging false positives unfortunately.  However, in the majority of cases, I would expect that we are disposing of created objects properly, and missing a call to `Dispose` would be a bug (in which case the analyzer would notify us).  So I would guess that the analyzer is doing it's job most of the time.

Shane32:
> The first issue isn't a potential bug fix, it's just a suggestion for simplifying the code; an ignored warning cannot hurt the project.

@MaceWindu points out that the rule can and has found bugs; that's fine with me if the rule is serving a useful purpose.

Source/LinqToDB/DataProvider/Informix/InformixProviderAdapter.cs
MaceWindu: I've referenced it first message: we had duplicate null check over same parameter and it highlighted second condition as unreachable

MaceWindu: I've referenced it first message: we had duplicate null check over same parameter and it highlighted second condition as unreachable

Source/LinqToDB.Tools/Schema/LegacySchemaProvider.cs
```diff
@@ -396,9 +397,9 @@ private void ParseTable(TableSchema table, SchemaObjects loadedObjects)
 
 			if (hasPrimaryKey)
 			{
-				var pkColumns = table.Columns.Where(c => c.IsPrimaryKey);
+				var pkColumns = table.Columns.Where(c => c.IsPrimaryKey).ToArray();
```
viceroypenguin, line 400: technically, `ToList()` may have lower memory usage, since it does not require the array to be trimmed after loading the data.

Source/LinqToDB/Common/DbDataType.cs
```diff
@@ -137,7 +137,7 @@ public override int GetHashCode()
 			{
 				var hashCode = (SystemType != null ? SystemType.GetHashCode() : 0);
 				hashCode     = (hashCode * 397) ^ (int) DataType;
-				hashCode     = (hashCode * 397) ^ (DbType    != null ? DbType.GetHashCode()          : 0);
+				hashCode     = (hashCode * 397) ^ (DbType    != null ? DbType.GetHashCode()        : 0);
```
viceroypenguin, line 140: vertical spacing issue

Source/LinqToDB/Expressions/TypeMapper/TypeMapper.cs
```diff
@@ -1,17 +1,17 @@
...
 
 namespace LinqToDB.Expressions
 {
-	using System.Collections;
-	using System.Diagnostics.CodeAnalysis;
-	using System.Threading;
-	using System.Threading.Tasks;
 	using Common;
-	using LinqToDB.Extensions;
+		using Extensions;
```
viceroypenguin, line 14: Is this correctly aligned?

viceroypenguin (approved):

viceroypenguin (approved):

viceroypenguin (dismissed):

PR #4252, opened 2023-08-20
SqlOptions and IgnoreConstantExpressionInOrderBy option

igor-tkachev:
If `true` linq2db will ignore any constant expressions in ORDER BY clause.


Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -1476,5 +1508,103 @@ public static DataOptions UseBulkCopyWithoutSession(this DataOptions options, bo
...
+		public static SqlOptions WithGenerateFinalAliases(this SqlOptions options, bool generateFinalAliases)
+		{
+			return options with { GenerateFinalAliases = generateFinalAliases };
+		}
+
+		#endregion
+
+		#region DataOptions.SqlOptions
+
+		/// <summary>
```
viceroypenguin, line 1536: Incorrect comment for this property.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -644,7 +660,11 @@ public static class Sql
 			/// </code>
 			/// </example>
 			/// </summary>
-			public static bool GenerateFinalAliases { get; set; }
+			public static bool GenerateFinalAliases
```
viceroypenguin, line 663: should `AssociationAlias` move into `SqlOptions` as well?

Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
```diff
@@ -90,15 +90,15 @@ protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, Meth
...
 			}
 
-
 			if (!isContinuousOrder && !builder.DataContext.Options.LinqOptions.DoNotClearOrderBys)
 				sequence.SelectQuery.OrderBy.Items.Clear();
 
 			foreach (var expr in sql)
 			{
 				// we do not need sorting by immutable values, like "Some", Func("Some"), "Some1" + "Some2". It does nothing for ordering
+				// IT: Actually it does. See ORDER BY ordinal position.
```
viceroypenguin, line 99: I thought most DBMS had relegated `ORDER BY` ordinal position to the "we support it, but strongly recommend you don't use it" bucket. Is this something we should be allowing/encouraging (if we haven't supported it already)?

viceroypenguin (changes requested):

Source/LinqToDB/Common/Configuration.cs
igor-tkachev: No. Its impossible.

Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
igor-tkachev: We do not have to learn people what the best/worst practices are. We just need to provide functionality.

BTW. This feature worked before.

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
igor-tkachev: fixed

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
```diff
@@ -566,5 +566,13 @@ static bool HasFinal(SqlQueryExtension ext)
 					;
 			}
 		}
+
+		protected override void BuildExpressionForOrderBy(ISqlExpression expr)
+		{
+			if (!DataOptions.SqlOptions.IgnoreConstantExpressionInOrderBy && expr.SystemType == typeof(int) && expr is SqlValue(var value))
```
sdanyliv, line 572: It is bad to move it here. It should be done in Optimizer.

sdanyliv (changes requested):

sdanyliv:
Option should be `AllowPositionedOrderBy` and should be by default false. When option is disabled throw exception when `OrderByBuilder` detected constant value with explanation that `AllowPositionedOrderBy` can allow this possibility along with warning that `linq2db` do not guarantee columns position especially in if query reused later on subselect.

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
igor-tkachev: The optimizer is no longer called from this code. But at the same time, any integer constant is ultimately translated into toInt32(X)

igor-tkachev:
> Option should be `AllowPositionedOrderBy` and should be by default **false**. When option is disabled throw exception when `OrderByBuilder` detected constant value with explanation that `AllowPositionedOrderBy` can allow this possibility along with warning that `linq2db` do not guarantee columns position especially if query reused later on subselect.

I agree with name. I do not agree with default value. If you do not use this feature, you do not case. If you use it, you are going to be very surprised why linq2db does not generate expected SQL.

Like I was, when old functionality stopped working.

sdanyliv:
@igor-tkachev, it is potential buggy approach I have disabled that many years ago, so you only one who use that.

igor-tkachev:
@sdanyliv, it is standard SQL functionality. It is not our job to teach people good or bad manners and wipe their broken noses.

If nobody use it, why did you disable it? Nobody complained before, nobody complained after, and I guess no one said "thank you" to you for saving his life.

jods4:
It is standard SQL, but in linq2db it is a very dangerous feature that can very easily backfire.

You have no control over the generated SELECT, in fact linq2db tries as much as possible to execute the projection in C# and it rewrites the SELECT to do so.

Consider this example:
```cs
string MergePossiblyEmpty(params string[] values)
{
  return string.Join(" ", values.Where(s => !s.IsNullOrEmpty()));
}

var query = 
  from u in db.User
  select new 
  {
   FullName = MergePossiblyEmptyStrings(u.First, u.Middle, u.Last),
   u.Age,
  }
  orderby 2;

query.ToList();
```

If you expected a list sorted by Age, you're in for a surprise because the query will look like this:
```sql
select first, middle, last, age
from [User]
order by 2 -- oups, middle instead of age
```

This requires knowledge of linq2db internals to understand, so I don't think it's fair to assume users should know.

> If nobody use it, why did you disable it? Nobody complained before, nobody complained after, and I guess no one said "thank you" to you for saving his life.

Generally speaking, there's a good reason to remove features that nobody uses.
linq2db codebase is not pretty: it's large, complex, contains bugs and it's hard to maintain and evolve.
I have tried to refactor several aspects of linq2db only to be stopped in my tracks by edgy features of dubious value.
So I guess you can say the devs complain?

In this specific case it's about removing a `IsConstant(sql)` call so maybe it's not that much of a deal.

Something to also consider for rarely used features that would be likely to generate issues in github if someone did is: can you do it in user-land, without support in linq2db?
The answer here is yes: 
```cs
[SqlExpression("{0}", IsPure = false, ServerSideOnly = true)]
int ByIndex(int i) { }

// This would not be treated as a constant and generates ORDER BY 1
var query.OrderBy(_ => ByIndex(1));
```

So given this, is it really worth adding to linq2db core?

igor-tkachev:
It is not dangerous at all if you do not use this feature. You do not need to disable it to not use it. Just don't use it.

igor-tkachev:
We have compatibility flag for this feature. I think in the release notes we need to provide a list of settings for backward compatibility. Easy-peasy. Everybody happy!

jods4:
> It is not dangerous at all if you do not use this feature. You do not need to disable it to not use it. Just don't use it.

Sorry, but this is nonsense.

viceroypenguin:
@igor-tkachev I agree with @jods4 here. In general, we should default to fail-safe rather than fail-dangerous. A query that involves a positional order-by is dangerous, because it has the potential to fail unexpectedly, due to a minor change. This is the reason that most dbms have updated their stance on positional order-by to "allowed, but not recommended". 

Positional order-by are not common among the developers I work with and talk to; many don't even know that it exists. For many of them, using a constant for the order-by would be a mistake, rather than an intentional decision.

As such, our position should be that they are _not_ allowed, unless you explicitly turn it on; instead of the alternative where they are allowed, unless you explicitly turn it off. Especially since there is even a workaround that does not involve modifying l2db code, as shared by jods above.

Shane32:
> @igor-tkachev I agree with @jods4 here. In general, we should default to fail-safe rather than fail-dangerous. A query that involves a positional order-by is dangerous, because it has the potential to fail unexpectedly, due to a minor change. This is the reason that most dbms have updated their stance on positional order-by to "allowed, but not recommended".
> 
> Positional order-by are not common among the developers I work with and talk to; many don't even know that it exists. For many of them, using a constant for the order-by would be a mistake, rather than an intentional decision.
> 
> As such, our position should be that they are _not_ allowed, unless you explicitly turn it on; instead of the alternative where they are allowed, unless you explicitly turn it off. Especially since there is even a workaround that does not involve modifying l2db code, as shared by jods above.

Agree on all points and views @viceroypenguin .

Linq2db is supposed to generate SQL that typically matches C# semantics.  `Queryable.OrderBy` is not a positional order-by afaik.  It returns "an `IOrderedQueryable<T>` whose elements are sorted according to a key" where the `keySelector` is "a function to extract a key from an element.".  It's designed to mimic `Enumerable.OrderBy`.  This is true for many other areas of LINQ to SQL conversion.  I understand that sometimes it's not exactly the same, but as a general rule of thumb it is so.  The most logical way to support positional orderby is to support something like `.OrderBy(x => Sql.Position(1))`, in my opinion.  This reasoning is particularly compelling when the position of selected keys within the select query is not guaranteed by linq2db; simply updating the linq2db nuget package could cause silent bugs in the code that are only known at runtime when the order of the results are wrong.  But with `.OrderBy(x => Sql.Position(1))` (or similar) the user has explicitly asked for that behavior and it is up to them to ensure their code is correct when upgrading to a new version of linq2db.

Note that if `.OrderBy(x => 1)` currently supports positional order-by, then I would suggest only changing the behavior when the next major version is released, attempting to retain backwards compatibility as much as possible within a major version.

igor-tkachev:
Sorry, guys, I got COVID, lets continue a little bit later.

Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
```diff
@@ -90,15 +90,15 @@ protected override IBuildContext BuildMethodCall(ExpressionBuilder builder, Meth
...
 				//
-				if (QueryHelper.IsConstant(expr.Sql))
+				if (builder.DataOptions.SqlOptions.IgnoreConstantExpressionInOrderBy && QueryHelper.IsConstant(expr.Sql))
```
MaceWindu, line 101: This enables constants of any type, not just integer.

MaceWindu (changes requested):
1. I agree that this option should be disabled by default as it is unsafe
2. It would be nice to have separate AST element `SqlOrdinal { Ordinal : int }` so we can avoid `SqlValue` magic, e.g. in `OrderByBuilder` and in `ClickHouseSqlBuilder`
3. with (2) better name for `IgnoreConstantExpressionInOrderBy` option should be something like `EnableOrdinalsSupport` (even if we add support to `order by` here, ordinals also supported by some databases in other places like `group by`)

Source/LinqToDB/Linq/Builder/OrderByBuilder.cs
igor-tkachev: Yes. This is legitimate behavior. It is not about `integer` only. This is about using constants in `ORDER BY`.

viceroypenguin: I know for mssql it's a semantic error to use any constant other than an `int` in an `order by`. Do other sql systems allow constants in the `order by` other than `int`? Does it have a meaning for any of them? An `int` refers to the column ordinal by which to sort, what does it mean to order by 3.5, for example?

I would argue for limiting to just `int`, given the purpose of allowing a constant column selector.

```sql
-- mssql
select *
from sys.tables
order by 3.5

-- error: 
-- Msg 408, Level 16, State 1, Line 3
-- A constant expression was encountered in the ORDER BY list, position 1.
```

igor-tkachev: Some providers can support different constants.
I think if someone enable this feature, now it is his responsibility.

viceroypenguin (approved):

PR #4259, opened 2023-08-28
Add compilation support for Visual Studio 2022 users

Shane32:
Currently, when attempting to compile on a VS 2022 machine that does not have VS 2019 installed, an error is shown as follows:

> Error MSB3644 The reference assemblies for .NETFramework,Version=v4.5 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks

There is no developer pack available for .NET Framework 4.5.  However, if VS 2019 is installed on the same machine, the project compiles fine.  Following the link above refers you to [this webpage](https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/reference-assemblies?WT.mc_id=dotnet-35129-website) which instructs you how to use the NuGet package within this change to target .NET Framework 4.5.  

This change allows VS 2022 to download the reference assemblies allowing compilation on machines that do not have VS 2019 installed.

Note that I used the NuGet Package Manager to add the reference.  As it has `<PrivateAssets>all</PrivateAssets>` I assume it will not add dependencies.  Perhaps we can review the built package to be sure.


Shane32:
There are multiple projects that require this change so I moved it to Directory.build.props.  It should not harm any projects that do not require it.

MaceWindu (approved):

viceroypenguin (approved):

PR #4268, opened 2023-09-11
Change WriteTraceLineConnection accessibility level to protected

Metadorius:
Reasoning: the property is supposed to be overridable (why would a global static one be overridable then?) but for some reason it's not outside of the base class which is often inheritable.


MaceWindu (approved):
Makes sense

MaceWindu:
Thanks!

PR #3994, opened 2023-02-23
[WIP] Timespan support

jogibear9988:
Fix #3993
Fix #4306
Fix #4308
Fix #2950

---

Open Questions/Tasks

- Should we change the default type in Mapping Schema for TimeSpan to int64. I've done it, but is this okay?
- Can I access the Column Type from my builder? Cause I want also to support for example the usage of a Bigint Column in Postgres. I've done it now like this way: var dt = ((SqlField)((SqlExpression)timeSpan).Parameters[0]).ColumnDescriptor.DataType; but sure this is not the correct way

PR #4297, opened 2023-10-10
Support Guid ToString in all databases, add test for all DBs

### LINKED ISSUE #4295: Override how Guid.ToString for Sqlite is implemented in SQL
Is there a way, so I could override, how ToString() of a guid will be converted to SQL?

In Sqlite I need it to generate SQL like this:

        substr(hex(id), 1, 8)
        || '-' || substr(hex(id), 9, 4)
        || '-' || substr(hex(id), 13, 4)
        || '-' || substr(hex(id), 17, 4)
        || '-' || substr(hex(id), 21, 12)

jogibear9988:
Fix #4295

### DIFF
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -18,6 +18,7 @@ namespace LinqToDB
 	using Linq;
 	using SqlQuery;
 	using LinqToDB.Common;
+	using System.Diagnostics.CodeAnalysis;
 
 	[PublicAPI]
 	public static partial class Sql
@@ -282,7 +283,9 @@ namespace LinqToDB
 		[Expression(PN.SapHana, "Upper(Cast({0} as NVarChar(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.Sybase, "Upper(Convert(NVarChar(36), {0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Extension("", BuilderType = typeof(GuidToStringBuilder), PreferServerSide = true)]
+#if NET30_OR_GREATER || NETSTANDARD2_1_OR_GREATER
 		[return: NotNullIfNotNull(nameof(guid))]
+#endif
 		public static string? GuidToNormalizedString(Guid? guid)
 		{
 			return guid == null ? null : guid.ToString();
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -274,7 +274,7 @@ namespace LinqToDB
 		[Expression(PN.SQLite, "lower((substr(hex({0}), 7, 2) || substr(hex({0}), 5, 2) || substr(hex({0}), 3, 2) || substr(hex({0}), 1, 2) || '-' || substr(hex({0}), 11, 2) || substr(hex({0}), 9, 2) || '-' || substr(hex({0}), 15, 2) || substr(hex({0}), 13, 2) || '-' || substr(hex({0}), 17, 4) || '-' || substr(hex({0}), 21, 12)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.Access, "LCase(Mid(CStr({0}), 2, 36))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.PostgreSQL, "(Cast({0} as VarChar(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
-		[Expression(PN.MariaDB, "Lower(Cast({0} as CHAR(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.MariaDB10, "Lower(Cast({0} as CHAR(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.MySql, "Lower(Cast({0} as CHAR(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.Informix, "Lower(To_Char({0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
 		[Expression(PN.DB2, "lower((substr(hex({0}), 7, 2) || substr(hex({0}), 5, 2) || substr(hex({0}), 3, 2) || substr(hex({0}), 1, 2) || '-' || substr(hex({0}), 11, 2) || substr(hex({0}), 9, 2) || '-' || substr(hex({0}), 15, 2) || substr(hex({0}), 13, 2) || '-' || substr(hex({0}), 17, 4) || '-' || substr(hex({0}), 21, 12)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -259,8 +259,9 @@ namespace LinqToDB
 			public void Build(ISqExtensionBuilder builder)
 			{
 				var para = builder.GetExpression(0);
-				
-				builder.ResultExpression = PseudoFunctions.MakeConvert(SqlDataType.String, SqlDataType.Guid, para);
+				var toType = builder.Mapping.GetDbDataType(typeof(string));
+
+				builder.ResultExpression = PseudoFunctions.MakeMandatoryCast(para!, toType, SqlDataType.Guid);
 			}
 		}
 
--- a/Source/LinqToDB/Sql/Sql.cs
+++ b/Source/LinqToDB/Sql/Sql.cs
@@ -253,6 +253,24 @@ namespace LinqToDB
 			return Guid.NewGuid();
 		}
 
+		sealed class GuidToStringBuilder : IExtensionCallBuilder
+		{
+			public void Build(ISqExtensionBuilder builder)
+			{
+				var para = builder.GetExpression(0);
+				
+				builder.ResultExpression = PseudoFunctions.MakeConvert(SqlDataType.String, SqlDataType.Guid, para);
+			}
+		}
+
+		[CLSCompliant(false)]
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/DataProvider/SQLite/SQLiteSqlOptimizer.cs
```diff
@@ -151,6 +151,11 @@ public override ISqlExpression ConvertExpressionImpl(ISqlExpression expression,
 							return new SqlFunction(func.SystemType, "DateTime", func.Parameters[1]);
 						}
 
+						if (ftype == typeof(string) && func.Parameters[1].SystemType == typeof(Guid))
```
sdanyliv, line 154: It is really bad place to do that. Ask @MaceWindu for details.

jogibear9988: @MaceWindu could you help me?

jogibear9988:
Should I add "Guid.ToString" in Expressions.cs and create a Builder wich does nothing in Standard and in Sqlite I implement the logic. Same as I've done for TimeSpan in this: https://github.com/linq2db/linq2db/pull/3994/files#diff-0381ba7e238d0f1809d47075f54b06e6c5fb0dfd9c7f0ed1017188c0eab15e38
Would this be a better solution?

MaceWindu:
I'm a bit lost here. What you need to achieve? If you need guid.ToString mapping to SQL you should implement it in mappings (Expressions.cs probably).

Another thing that looks incorrect is that proposed mapping generates upper-case guid. It is common to use lowercased guid string and default mapping should generate lowercased guid.

jogibear9988:
> I'm a bit lost here. What you need to achieve? If you need guid.ToString mapping to SQL you should implement it in mappings (Expressions.cs probably).
> 
> Another thing that looks incorrect is that proposed mapping generates upper-case guid. It is common to use lowercased guid string and default mapping should generate lowercased guid.

I'd like to be able to use for example guid.ToString().Contains("aa") in LINQ.
I can change to a lowercase. And I look to move to expressions

igor-tkachev:
We already have it for other types. See `Expressions.cs` line 555.

jogibear9988:
@igor-tkachev 
changed code to use expression.cs and a converter in Sql Helper class

jogibear9988:
The test 3791 does not work for ClickHouse with this change

ClickHouse.Client.ClickHouseServerException : Code: 53. DB::Exception: Can't infer common type for joined columns: OtherId: Nullable(String) at left, a_Association.Id: UUID at right. There is no supertype for types String, UUID because some of them are String/FixedString and some of them are not. (TYPE_MISMATCH) (version 23.12.2.59 (official build))



		public void Issue3791Test([DataSources] string context)
		{
			var ms = new MappingSchema();
			ms.SetConvertExpression<Guid, string>(a => a.ToString());
			ms.SetConvertExpression<string, Guid>(a => Guid.Parse(a));

			using var table = db.CreateLocalTable<Issue3791Table>();
			using var _     = db.CreateLocalTable<Issue3791GuidTable>();

			table.LoadWith(a => a.Association).ToList();
		}


but for me it's not clear, when the convertExpression is used. Do I need to change smth. or need this to be fixed inside of linq2db?

jogibear9988:
Is the  ms.SetConvertExpression<Guid, string> used before my handling of ToSting in Expressions.cs?

Tests/Linq/UserTests/Issue4295Tests.cs
```diff
@@ -0,0 +1,62 @@
...
+	public class Issue4295Tests : TestBase
+	{
+		[Table]
+		public class InfeedAdvicePositionDTO
+		{
+			[Column] public Guid Id { get; set; }
+		}
+
+		[Test]
+		public void TestGuidToString([IncludeDataSources(TestProvName.AllSQLite, TestProvName.AllSqlServer)] string context)
```
jods4, line 24: shouldn't we test for all RDBMS?

jogibear9988:
I'll add test for every DBMS, and remove the Mapping in Expression.cs.
I thought it would only be used on explicit "ToString" calls, but it also seems to be used when Guid is mapped to a string, or compared inside of Database with a string column.
This maybe then would break.
I rename the function to "GuidToNormalizedString".
When we have the possiblity to create own additions to "Expressions.cs" as we said in the Regex Pull, everybody could map for themself

jogibear9988:
all tests work, ready for review

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -263,6 +263,40 @@ public static Guid NewGuid()
...
+		[Expression(PN.Oracle, "(substr(rawtohex({0}), 7, 2) || substr(rawtohex({0}), 5, 2) || substr(rawtohex({0}), 3, 2) || substr(rawtohex({0}), 1, 2) || '-' || substr(rawtohex({0}), 11, 2) || substr(rawtohex({0}), 9, 2) || '-' || substr(rawtohex({0}), 15, 2) || substr(rawtohex({0}), 13, 2) || '-' || substr(rawtohex({0}), 17, 4) || '-' || substr(rawtohex({0}), 21, 12))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.ClickHouse, "upper(toString({0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.SapHana, "Upper(Cast({0} as NVarChar(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.Sybase, "Upper(Convert(NVarChar(36), {0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Extension("", BuilderType = typeof(GuidToStringBuilder), PreferServerSide = true)]
+		public static string? GuidToNormalizedString(Guid? guid)
```
viceroypenguin, line 295:
```suggestion
		[return: NotNullIfNotNull(nameof(guid))]
		public static string? GuidToNormalizedString(Guid? guid)
```

viceroypenguin (dismissed):

viceroypenguin (dismissed):

jogibear9988:
1 test failed, but it's unrelated to the pull req, cause the Code from the pull req is only used in the new test...

image

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -264,6 +266,40 @@ public static Guid NewGuid()
...
+		[Expression(PN.Oracle, "(substr(rawtohex({0}), 7, 2) || substr(rawtohex({0}), 5, 2) || substr(rawtohex({0}), 3, 2) || substr(rawtohex({0}), 1, 2) || '-' || substr(rawtohex({0}), 11, 2) || substr(rawtohex({0}), 9, 2) || '-' || substr(rawtohex({0}), 15, 2) || substr(rawtohex({0}), 13, 2) || '-' || substr(rawtohex({0}), 17, 4) || '-' || substr(rawtohex({0}), 21, 12))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.ClickHouse, "upper(toString({0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.SapHana, "Upper(Cast({0} as NVarChar(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.Sybase, "Upper(Convert(NVarChar(36), {0}))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Extension("", BuilderType = typeof(GuidToStringBuilder), PreferServerSide = true)]
```
igor-tkachev, line 296: Also add Sql Server conversion as `CAST({0} AS char(36))`.

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -585,7 +585,6 @@ public void SetInfo(MappingSchema mappingSchema)
 			{ MT<uint?   >(() => ((uint?)    0) .ToString()!), N(() => L<uint?,    string>((uint?    p0) => p0!.Value.ToString() )) },
 			{ MT<ulong?  >(() => ((ulong?)   0) .ToString()!), N(() => L<ulong?,   string>((ulong?   p0) => p0!.Value.ToString() )) },
```
igor-tkachev, line 587: Should we add specific `Guid.ToString()` mapping and use your function?
But in this case it should be lower-cased.

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -264,6 +266,40 @@ public static Guid NewGuid()
...
+
+		/// <summary>
+		/// Converts a Guid to a normalized string in the format <c>XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXX</c>.
+		/// </summary>
+		/// <param name="guid">The guid to convert.</param>
+		/// <returns>The guid formatted to a string.</returns>
+		[CLSCompliant(false)]
+		[Expression(PN.SQLite, "(substr(hex({0}), 7, 2) || substr(hex({0}), 5, 2) || substr(hex({0}), 3, 2) || substr(hex({0}), 1, 2) || '-' || substr(hex({0}), 11, 2) || substr(hex({0}), 9, 2) || '-' || substr(hex({0}), 15, 2) || substr(hex({0}), 13, 2) || '-' || substr(hex({0}), 17, 4) || '-' || substr(hex({0}), 21, 12))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.Access, "Mid(CStr({0}), 2, 36)", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
+		[Expression(PN.PostgreSQL, "Upper(Cast({0} as VarChar(36)))", IsNullable = IsNullableType.IfAnyParameterNullable, PreferServerSide = true)]
```
igor-tkachev, line 287: Why `Upper`? In most cases the conversion is done to lowercase by default.

igor-tkachev (changes requested):

Source/LinqToDB/Linq/Expressions.cs
jogibear9988: No. I had it inside and removed it. Because for example in some tests there is already a Join of some tables against a string representation of guid. And this will then break cause every database converted the Guid to string differently. I had it inside in my first commits and removed it cause of this. Everybody who do not yet use String Guids and does not depend on the legacy behaviour can add the Expression for the ToString themselves.

I also had it first lowercase and changed to uppercase cause I thought it was more common (it was the default in SQLServer when casting). But now most of the other databses need an Upper. I think, I find lowercase nicer, what do you think? I can change it.

Source/LinqToDB/Sql/Sql.cs
jogibear9988: will do

Source/LinqToDB/Sql/Sql.cs
jogibear9988: changed it

sdanyliv:
Superseeded by https://github.com/linq2db/linq2db/pull/4828

PR #4323, opened 2023-10-24
Improve `IAsyncEnumerable` usage

viceroypenguin:
This PR does two things:
* Updates `InsertWithOutputAsync`, `UpdateWithOutputAsync`, and `DeleteWithOutputAsync` to return `IAsyncEnumerable` properly. This is a binary-compatible change, as the existing methods are kept. However, these methods are now marked with `[Obsolete]` to encourage source migration to the newer `IAsyncEnumerable` method. 
* Adds `QueryToAsyncEnumerable` in places where `QueryToArrayAsync` and `QueryToListAsync` are provided. This allows consumers to use `IAsyncEnumerable` if desired. No obsoletions are planned here, though consumers are allowed and encouraged to upgrade if the query has many rows. 
  > [!NOTE]
  > This change only applies to `NATIVE_ASYNC`.

## 💣 Breaking Change
* The new versions of `XxxWithOutputAsync` are not source compatible, so this is a v6.0 breaking change. Consumers will need to add `.ToListAsync()` or similar to address this change. This is to restore consistency between `InsertWithOutputAsync`/`UpdateWithOutputAsync`/`DeleteWithOutputAsync` with `MergeWithOutputAsync`; this consistency has been chosen to use the more flexible and modern approach.

### DIFF
--- a/Source/LinqToDB/Data/CommandInfo.cs
+++ b/Source/LinqToDB/Data/CommandInfo.cs
@@ -256,6 +256,30 @@ namespace LinqToDB.Data
 			}
 		}
 
+#if NETSTANDARD2_1PLUS
+		/// <summary>
+		/// Executes command asynchronously and apply provided action to each record, mapped using provided mapping function.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="objectReader">Record mapping function from data reader.</param>
+		/// <param name="cancellationToken">Asynchronous operation cancellation token.</param>
+		/// <returns>Async sequence of records returned by the query.</returns>
+		public async IAsyncEnumerable<T> QueryToAsyncEnumerable<T>(Func<DbDataReader, T> objectReader, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+		{
+			await DataConnection.EnsureConnectionAsync(cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext);
+
+			InitCommand();
+
+			await using ((DataConnection.DataProvider.ExecuteScope(DataConnection) ?? EmptyIAsyncDisposable.Instance).ConfigureAwait(Configuration.ContinueOnCapturedContext))
+			{
+				var rd = await DataConnection.ExecuteDataReaderAsync(GetCommandBehavior(), cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext);
+				await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
+					while (await rd.DataReader!.ReadAsync(cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext))
+						yield return objectReader(rd.DataReader!);
+			}
+		}
+#endif
+
 		#endregion
 
 		#region Query
@@ -457,9 +481,9 @@ namespace LinqToDB.Data
 								if (isFaulted)
 									throw;
 
-								isFaulted    = true;
+								isFaulted = true;
 								objectReader = GetObjectReader2<T>(DataConnection, reader, CommandText, additionalKey);
-								result       = objectReader(reader);
+								result = objectReader(reader);
 							}
 
 							action(result);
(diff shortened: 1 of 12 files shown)


Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -470,6 +498,62 @@ await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
...
+			InitCommand();
+
+			await using ((DataConnection.DataProvider.ExecuteScope(DataConnection) ?? EmptyIAsyncDisposable.Instance).ConfigureAwait(Configuration.ContinueOnCapturedContext))
+			{
+#if NETSTANDARD2_1PLUS
+				var rd = await DataConnection.ExecuteDataReaderAsync(GetCommandBehavior(), cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext);
+				await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
+#else
+				using (var rd = await DataConnection.ExecuteDataReaderAsync(GetCommandBehavior(), cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext))
+#endif
```
Shane32, line 521: ditto

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -275,6 +275,21 @@ public static Task<T[]> QueryToArrayAsync<T>(this DataConnection connection, Fun
...
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="objectReader">Record mapping function from data reader.</param>
+		/// <param name="sql">Command text.</param>
+		/// <returns>Async sequence of records returned by the query.</returns>
+		public static IAsyncEnumerable<T> QueryToAsyncEnumerable<T>(this DataConnection connection, Func<DbDataReader, T> objectReader, string sql)
+		{
+			return new CommandInfo(connection, sql).QueryToAsyncEnumerable(objectReader);
+		}
```
Shane32, line 290: A number of these methods do not accept a `[EnumeratorCancellation]` token and pass it on to the `QueryToAsyncEnumerable` method.  While I know that the `.WithCancellation` extension method can be used, for user convenience, it would seem that we should add the cancellation token as an optional parameter whenever possible.

Source/LinqToDB/Data/DataReaderAsync.cs
```diff
@@ -206,6 +229,13 @@ public Task QueryForEachAsync<T>(T template, Action<T> action, CancellationToken
 			return QueryForEachAsync(action, cancellationToken);
 		}
 
+#if NATIVE_ASYNC
+		public IAsyncEnumerable<T> QueryForEachAsync<T>(T template, CancellationToken cancellationToken)
```
Shane32, line 233:
```suggestion
		public IAsyncEnumerable<T> QueryForEachAsync<T>(T template, [EnumeratorCancellation] CancellationToken cancellationToken = default)
```

Is this the proper method name?  First, I'm not sure it should end in 'Async', and second, the other overloads here are overloads of the same method name without the template, correct?  So shouldn't this be named `QueryToAsyncEnumerable<T>(T template, ...)` ?

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -256,6 +256,34 @@ await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
...
```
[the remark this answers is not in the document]
viceroypenguin: This is because `await using` doesn't work properly with net472. The `#if NETSTANDARD2_1PLUS` is copied from the other lines in this file that have the same behavior.

Source/LinqToDB/Data/DataConnectionExtensions.cs
viceroypenguin: `[EnumeratorCancellation]` is a moniker _specifically_ for the state-machine of the implicit `async IAsyncEnumerable<T>`. If you are simply returning an `IAsyncEnumerable<T>`, then `[EnumeratorCancellation]` has no impact on the method.

`.WithCancellation` is generally the preferred method of passing a token to an `IAsyncEnumerable<T>`, because then the `IAsyncEnumerable<T>` can be passed around as necessary and the token provided later. 

I'm going to argue for leaving these as is.

Source/LinqToDB/Data/DataReaderAsync.cs
viceroypenguin: Whoops. Missed one. Fixed in #732badc (will be uploaded when possible)

Source/LinqToDB/Data/CommandInfo.cs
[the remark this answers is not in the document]
Shane32: Ok, so should we be using `await using` on line 273 ?

Source/LinqToDB/Data/DataConnectionExtensions.cs
Shane32: Perhaps we should do the opposite - hide the cancellation token from being exposed in the methods where we now have `[EnumeratorCancellation]` ?  (By putting the implementation in a private method.)

Shane32: For reference, EF Core’s AsAsyncEnumerable has no exposed cancellation token

https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.entityframeworkqueryableextensions.asasyncenumerable?view=efcore-7.0

Source/LinqToDB/Data/CommandInfo.cs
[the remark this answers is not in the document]
viceroypenguin: Interesting, I didn't notice that. After reviewing, it looks like it's because `DataReaderWrapper` does not implement `IAsyncDisposable` in netfx, only in netcoreapp3.1+. Maybe @MaceWindu can explain?

Source/LinqToDB/Data/DataConnectionExtensions.cs
viceroypenguin: > Perhaps we should do the opposite - hide the cancellation token from being exposed in the methods where we now have `[EnumeratorCancellation]` ? (By putting the implementation in a private method.)

Works for me. I'll take a look in a bit.

viceroypenguin: Fixed in #9d5bd15 (will be uploaded when possible)

MaceWindu:
@viceroypenguin what happened to it?

PR #4327, opened 2023-10-25
In Subquery tests

jods4:
Tests for #4196 are poorly written.
They don't test actual possible combinations of values and expected results, and as such it was missed that several of the tests actually don't test what they should.

For instance, the first test that I modified in this PR verifies if the generated SQL contains `IS [NOT] NULL` or not, but because the columns used in the query are all non-nullables, nothing relevant is tested.

After I use nullable columns instead, the test fail, as it should: when `CompareNullsAsValues` is false, the new transform is still applied even though it should not.


igor-tkachev (approved):

igor-tkachev (approved):

Tests/Linq/Linq/InSubqueryTests.cs
```diff
@@ -7,102 +8,135 @@
...
+		}
+
 		[Test]
-		public void InTest([DataSources] string context, [Values(true, false)] bool preferExists, [Values(true, false)] bool compareNullsAsValues)
+		public void InTest([DataSources] string context, [Values] bool preferExists, [Values] bool compareNullsAsValues)
 		{
 			using var db = GetDataContext(context, o => o.WithOptions<LinqOptions>(lo => lo with { PreferExistsForScalar = preferExists, CompareNullsAsValues = compareNullsAsValues }));
 
-			var q =
+			AssertTest(db, preferExists, true,
```
jods4, line 44: is it normal that parameter `compareNullsAsValue` is hard-coded `true` when the test is parameterized to run both `true` and `false`? Shouldn't this be the parameter?

jods4: (there are similar instances below)

igor-tkachev (approved):

viceroypenguin (approved):

igor-tkachev (approved):

MaceWindu (approved):

PR #4345, opened 2023-11-27
Migrate to own docker test images

MaceWindu:
Once again sql server testing is broken because people decide to delete images from dockerhub (no idea why), so we finally move to own images

- [x] SQL Server 2005 Image
- [x] SQL Server 2008 Image
- [x] SQL Server 2012 Image
- [x] SQL Server 2014 Image
- [x] SQL Server 2016 Image
- [x] SQL Server 2017 Image (win)
- [x] SQL Server 2019 Image (win)
- [x] SQL Server 2022 Image (win)
- [ ] Sybase ASE Image (linux)
- [x] Firebird 5RC1 Image (linux)
- [ ] fix memory limit issues with new SAP HANA docker release

All SQL Sever images use latest version SQL Server Express with Advanced Services with database engine + full text search features installed.

SQL Server 2005 doesn't support TLS 1.2+, so it require protocols configuration on host (which is not recommended):
- for S.D.S: TLS 1.1 should be enabled
- for M.D.S: TLS 1.1 should be enabled AND TLS 1.2/1.3 disabled
- note that it is not possible to disable TLS for SQL Server as it is always used for login phase even for non-encrypted connection


MaceWindu:
Plan to merge today to unblock sql server testing. Also will try to tune new sap hana image (not enough memory on CI with new image and SAP removed old image).
Sybase/FB5 updates will be done later

viceroypenguin (dismissed):

PR #4350, opened 2023-12-10
Removing SqlCondition from building filters,

sdanyliv:
Well, removed an optimization pain - `SqlCondition`

No more 
```cs
searchCondition.Add(new SqlCondition(isNot: false, predicate, isOr: false);
```
Just
```cs
searchCondition.Add(predicate);
```


jods4:
So glad to see this go away! I had started working on similar simplifications in the SqlFactory branch (which I'll have to start over when the new parser is done).
The double, not always useful, bools were so annoying and actually the cause of more than one bug in the codebase.
It's a lot of meticulous work (I know, I tried) so thanks a lot!
🎉🎉🎉

PR #4352, opened 2023-12-12
Make association element can be subclass of type that implements IEnu…

Ethen-Wang:
Make association element can be subclass of type that implements `IEnumerable<T>`


sdanyliv (changes requested):
Please add regression Unit test.

Without Unit Test this change will be lost when we switch to upcoming release, which has everything rewritten in this area.

viceroypenguin:
Closing for now; @Ethen-Wang, you may reopen once you have made new changes against the latest mainline.

PR #4362, opened 2023-12-22
Move F# support to separate library

MaceWindu:
This will allow us:
- implement pluggable support (intead of being enabled and run always for all users)
- use normal code instead of reflection
- it will be much easier to add better F# support in future

TODO:
- [x] finalize API
- [x] add nuget
- [x] support non-column constructor parameters mapping by records
- [x] support records in projections
- [ ] convert `RecordReaderBuilder` to new mapping logic

### DIFF
--- a/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
+++ b/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
@@ -9,4 +9,4 @@ module Methods =
     /// <remarks>Currently it is limited to record types support, but we plan to add more support in future releases.</remarks>
     [<Extension>]
     let UseFSharp(options : DataOptions) =
-        options.UseInterceptor FSharpExpressionInterceptor.Instance
+        options.UseInterceptor FSharpEntityBindingInterceptor.Instance
--- a/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
+++ b/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
@@ -5,6 +5,8 @@ open System.Runtime.CompilerServices
 
 [<Extension>]
 module Methods =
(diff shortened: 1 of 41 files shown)


Source/LinqToDB/Interceptors/ExpressionInterceptor.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+using LinqToDB.Reflection;
+
+namespace LinqToDB.Interceptors.Internal
+{
+	public abstract class ExpressionInterceptor : IExpressionInterceptor
```
sdanyliv, line 9: Need better name.

Source/LinqToDB/Interceptors/IExpressionInterceptor.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+using LinqToDB.Reflection;
+
+namespace LinqToDB.Interceptors.Internal
+{
+	public interface IExpressionInterceptor : IInterceptor
```
sdanyliv, line 9: Need better name.

Source/LinqToDB.FSharp/DataOptionsExtensions.fs
```diff
@@ -0,0 +1,10 @@
+﻿namespace LinqToDB.FSharp
+
+open LinqToDB
+open System.Runtime.CompilerServices
+
+[<Extension>]
+module Methods =
+    [<Extension>]
+    let UseFSharpRecords(options : DataOptions) =
```
sdanyliv, line 9: Maybe more generalized name `UseFSharp`. I think it will be more interceptors.

PR #4370, opened 2024-01-16
Update solution TFMs

### LINKED ISSUE #3935: Drop .NET Core 3.1 from releases and testing
Cleanup task for v6 release

### LINKED ISSUE #3937: Drop .NET Framework 4.5/4.6.1 support
bump lowest supported version to currently supported 4.6.2

Cleanup task for v6

MaceWindu:
Fix #3935
Fix #3937

Removed old TFMs. General idea is to have 4 TFMs for releases and testing (3):
- .NET Framework oldest supported TFM : `4.6.2` is enough for us, so we replace `net45`, `net46` and `net472` with single `net462`
- .NET oldest supported runtime: `net6.0` for now
- .NET newest supported runtime: `net8.0` for now to support new features
- `netstandard2.0` for use from ns2.0 projects and pre-v6 .NET runtimes

Same changes applied to tests, so you will need to update test config configurations:
- `NETFX` for .net framework tests
- `NET6.0` and `NET8.0` for .net tests

Also created separate baselines branch for v6 development to track baselines changes without conflicts with master.


Directory.Build.props
```diff
@@ -14,6 +14,8 @@
 		<Nullable>enable</Nullable>
 		<WarningLevel>9999</WarningLevel>
 		<ErrorReport>prompt</ErrorReport>
+		<!--1573: missing xmldoc-->
+		<!--1591: missing xmldoc-->
 		<NoWarn>1573;1591</NoWarn>
```
viceroypenguin, line 19: I would probably prefer we unchecked this, and added specific disables for places where we don't have time to document yet.

Directory.Build.props
```diff
@@ -58,19 +60,15 @@
 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
 	</PropertyGroup>
 
-	<PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'netcoreapp3.1' OR '$(TargetFramework)' == 'net6.0' OR '$(TargetFramework)' == 'net7.0'">
-		<DefineConstants>NETSTANDARD2_1PLUS;$(DefineConstants)</DefineConstants>
-	</PropertyGroup>
-
-	<ItemGroup>
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'net6.0' ">
```
viceroypenguin, line 63: FYI PolySharp can be included on net8 as well.

Redist/CLI/net8.0/dotnet-linq2db.win-x64.exe
viceroypenguin, line 1: what does adding the exe's to the source control do for us?

viceroypenguin (approved):

Shane32:
Does this drop .NET Standard 2.0 as a build target?  I'm actively using linq2db with multiple old frameworks.  I think it's essential to retain a build target of .NET Standard 2.0, even if such use is unsupported.  This matches both Microsoft's recommendation, and their behavior for most of their nuget packages.

viceroypenguin:
> Does this drop .NET Standard 2.0 as a build target?  I'm actively using linq2db with multiple old frameworks.  I think it's essential to retain a build target of .NET Standard 2.0, even if such use is unsupported.  This matches both Microsoft's recommendation, and their behavior for most of their nuget packages.

@Shane32 I'm curious what TFM you're working with that is not covered by net462 and net6? As far as I'm aware, the only active support Microsoft is keeping for ns2.0 is for source generators; and they should never touch a database anyway. The only older versions of net that are still under support are net471+ and net3.5. We haven't supported net35 in forever, and net462 is well past eol, so that should be good enough for most consumers.

Shane32:
Well, I have one very large project that is stuck on .NET Core 3.1, which uses linq2db extensively.  Hopefully it will see an upgrade to .NET 8 later this year, but for various reasons it would take a lot of manpower to upgrade.  I also manage very large number of other projects in active use which run on ASP Classic, .NET 2.0, .NET 4.5, and .NET Core 2.1.  (They don't use linq2db, just saying.)  Just because a framework isn't supported by MS doesn't mean that there are not valid uses for it.  For example, there are also various other frameworks besides .NET Core and .NET Framework, although in general use of those frameworks has been going down since .NET Core was released.  Another problem with not building for .NET Standard is that now any other packages that depend on linq2db cannot be built for .NET Standard.  I have some currently-private nuget packages that fall into this category.

I cannot express how strongly I recommend building for .NET Standard 2.0, in accordance with Microsoft's recommendations.

Shane32:
https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-standard-not-deprecated

> .NET Standard not deprecated
>
> .NET Standard is still needed for libraries that can be used by multiple .NET implementations. We recommend you target .NET Standard in the following scenarios:
>
> - Use netstandard2.0 to share code between .NET Framework and all other implementations of .NET.
> - Use netstandard2.1 to share code between Mono, Xamarin, and .NET Core 3.x.

viceroypenguin:
Obviously, this is just my opinion, but I'm going to say that ns2, netcoreapp3.1, etc. should be dropped anyway. The existing versions of l2db will continue to be available, obviously, for older consumers, but future developments should be done on modernizing versions of .net.

Even EFC has dropped support for older versions; the last version to support ns2.0 (and netfx) is v3.1.32, released December '22. 

One of the things you can see in this PR is how much the code is simplified by removing support for versions of .net that do not have native async. There is only so much I can recommend keeping support for versions of .net that are out of support.

Shane32:
> Even EFC has dropped support for older versions; the last version to support ns2.0 (and netfx) is v3.1.32, released December '22.

Entity Framework Core and ASP.Net Core have always followed .NET and are basically release as part of the same framework just in a separate package. They don’t support .NET Framework at all and were not intended to.  This is one of the great benefits of linq2db in that it is fully cross platform. Or has been up until this PR. 

> One of the things you can see in this PR is how much the code is simplified by removing support for versions of .net that do not have native async.

Is it so much different when we still are retaining support for .Net Framework? I’ll take a look…

I might feel differently if we could completely abandon the older technologies but as we have to support .NET Framework anyway it seems like it should be a minimal amount of effort to build for .NET Standard also.  At least that’s been my experience with the other libraries I support.

Shane32:
> There is only so much I can recommend keeping support for versions of .net that are out of support.

I'm only recommending that we _build_ for .NET Standard 2.0, not _support_ old versions of .NET.  I have no problem adding language to that effect to the readme or adding build warnings, etc.  MS doesn't _support_ using their nuget packages on old .NET frameworks, but they still _build_ for it.

Shane32:
I also don't mind dropping build targets for .NET Framework versions prior to .NET 4.6.2.  But .NET Standard 2.0 should have roughly the same framework footprint as .NET Framework 4.6.2.

Directory.Build.props
MaceWindu: Unfortunatelly there are too many of them :)

Redist/CLI/net8.0/dotnet-linq2db.win-x64.exe
MaceWindu: cli tool supports option to run scaffold using specific architecture for native providers (e.g. access). For that we need exe to run child process. Maybe there are better options exist, but don't really want to touch it :)

Shane32:
> Taking into account that in any case I expected `netstandard2.0` question arise soon after release, I will restore ns2.0 support

Thanks!!

sdanyliv (dismissed):

Shane32 (approved):

igor-tkachev:
Also `netstandard2.0` is used by Roslyn Source Generators if we plan to use them in the future (for AOT?).

viceroypenguin:
> Also `netstandard2.0` is used by Roslyn Source Generators if we plan to use them in the future (for AOT?).

Per discussion with the roslyn team, linq2db absolutely does not in any way belong in a source generator. I've considered this, and it was pointed out (rightfully) that it should be done as a build-step instead of a source generator. 

Max time to complete source generation is 10ms, with no async allowed. As such, db calls (and frankly IO of any kind) are strictly forbidden per performance expectations.

igor-tkachev:
> Max time to complete source generation is 10ms, with no async allowed. As such, db calls (and frankly IO of any kind) are strictly forbidden per performance expectations.

We have a lot of code that can be used without accessing database. For example SQL generation. Who can prevent us from replacing methods containing linq with methods containing already generated SQL?

PR #4374, opened 2024-01-18
Removed 'with' keyword usage.

igor-tkachev:
The use of the `with` keyword causes some problems for some developers in some environments.

### DIFF
--- a/Source/LinqToDB/Common/Configuration.cs
+++ b/Source/LinqToDB/Common/Configuration.cs
@@ -124,16 +124,16 @@ namespace LinqToDB.Common
 		/// annotations in [Column], [Association], or [Nullable].
 		/// </summary>
 		/// <remarks>Defaults to false.</remarks>
-		public static bool UseNullableTypesMetadata 
-		{ 
+		public static bool UseNullableTypesMetadata
+		{
 			get => _useNullableTypesMetadata;
-			set 
+			set
 			{
-				// Can't change the default value of "false" on platforms where nullable metadata is unavailable.				
+				// Can't change the default value of "false" on platforms where nullable metadata is unavailable.
 				if (value) Mapping.Nullability.EnsureSupport();
 				_useNullableTypesMetadata = value;
 			}
-		}	
+		}
 
 		public static class Data
 		{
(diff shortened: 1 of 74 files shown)


viceroypenguin:
> The use of the `with` keyword causes some problems for some developers in some environments.

Do you have an example of when it affects a developer?

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -188,7 +188,7 @@ public static bool PreloadGroups
 				set
 				{
 					if (Options.PreloadGroups != value)
```
jods4, line 190: Ok so maybe `with` in generated source code causes issues for people using older C# versions/runtimes (is that the issue?), but why remove it from main library (internal) code compiled by us?

igor-tkachev:
I had a personal request from @sdanyliv. He has troubles regularly to debug our project with the following reason

image

I have no clue why it happens. I do not have such problem.

It is still possible to use `with` keyword. I just removed using it from source code and tests, so it should not use generated `Clone` method.

If you know how to fix it, lets discussed it and discard this PR as I do not like it either.

jods4:
That's quite strange! Hard to say without more context and information, or maybe some clue/pattern.

`ClickHouseOptions` is a `record`, so it should have a `<Clone>$` special method emitted by C# compiler.
`with` calls this `<Clone>$` method and I don't see how/why it wouldn't be found...

On which runtime does it happen? Old or any?
Is mocking or anything special involved?
Are we sure the right DLL is loaded? The tests aren't loading an old DLL lingering around that contains a non-record `ClickHouseOptions`?

Otherwise I have no clue.

MaceWindu:
We don't need it. Issue caused by incorrect build configuration when same project referenced as transient dependency through other projects with different TFMs. We had this issue before and fixed it by configuring TFMs for our projects. Now it arised again temporary in parser branch (I expect to fix it today after merge with tfm update branch).

Issue occurs under following condition:
- project build against library with TFM1 but due to multi-TFM transient dependencies build system puts library with TFM2 to bin folder
- TFM1 or TFM2 is pre-net5.0

While it should in general work, microsoft runtime team did a bad job here which results in such exception. While roslyn team check for presense of marker types by name (as it's just markers, no need to go into berserk mode here), runtime team decided to be a smartcookie and ensure that marker type is exactly same type which where used during build. It doesn't make sense, but it's how it is done so we should live with it.

Taking into account that we could have different API surface for builds with different TFMs we should leave it as it to not mask build configuration issues but encourage people to fix their build.

PR #4377, opened 2024-01-19
Fix nuget package vulnerability.

igor-tkachev:


MaceWindu:
references from nuspec files should be also updated

PR #4381, opened 2024-01-21
Add [Pure] to options.

igor-tkachev:
Adding [Pure] to Option API.


Source/LinqToDB/Extensions/DataOptionsExtensions.Provider.cs
```diff
@@ -814,6 +861,71 @@ public static partial class DataOptionsExtensions
 		}
 #endif
 
+		/// <summary>
```
MaceWindu, line 864: Remove new methods. I plan to resolve inconsistencies in provider configuration APIs here in v6 and it's better to not introduce new conflicts where it wasn't requested by users

MaceWindu (changes requested):
maybe don't

Source/LinqToDB/Extensions/DataOptionsExtensions.Provider.cs
igor-tkachev: Done. Do not forget.

MaceWindu (approved):

PR #4382, opened 2024-01-22
Support for one way converter.

### LINKED ISSUE #4368: How to define a one way converter?
i.e., converters only take effect from .net to DB,
and converters only take effect from DB to .Net.
Currently, the converters work both ways.

igor-tkachev:
Fix #4368.
It is possible to convert data differently depending on source/destination (FromDatabase/ToDatabase) type.


Source/LinqToDB/Common/ConvertInfo.cs
```diff
@@ -98,16 +98,37 @@ public override int GetHashCode()
...
+						return _fromDatabaseExpressions ??= new();
+				case ConversionType.ToDatabase:
+					lock (_sync)
+						return _toDatabaseExpressions ??= new();
+				default:
+					throw new ArgumentOutOfRangeException(nameof(conversionType), conversionType, null);
+			}
 		}
 
-		public void Set(DbDataType from, DbDataType to, LambdaInfo expr)
```
viceroypenguin, line 108: source and binary breaking change. leave the old method in place, as a proxy to the new one with a default `conversionType` value. (note: for all deleted methods in this file, x6)

Source/LinqToDB/Common/Converter.cs
```diff
@@ -115,34 +115,34 @@ static bool ToBoolean(char ch)
...
-		/// Converts value to <paramref name="conversionType"/> type.
+		/// Converts value to <paramref name="toConvertType"/> type.
 		/// </summary>
 		/// <param name="value">Value to convert.</param>
-		/// <param name="conversionType">Target conversion type.</param>
+		/// <param name="toConvertType">Target conversion type.</param>
 		/// <param name="mappingSchema">Optional mapping schema.</param>
+		/// <param name="conversionType">Conversion type. See <see cref="ConversionType"/> for details.</param>
 		/// <returns>Converted value.</returns>
-		public static object? ChangeType(object? value, Type conversionType, MappingSchema? mappingSchema = null)
```
viceroypenguin, line 124: source and binary breaking change. leave the old method in place, as a proxy to the new one with a default `conversionType` value. (note: for all deleted methods in this file, x2)

viceroypenguin (changes requested):

Source/LinqToDB/Common/ConvertInfo.cs
igor-tkachev: `ConvertInfo` is not a public class.

Source/LinqToDB/Common/Converter.cs
igor-tkachev: Are you sure about source breaking change?

viceroypenguin: > Are you sure about source breaking change?

ah, nope. copy/paste from the other file. just a binary breaking change.

igor-tkachev: Do you really upgrade nugets without recompiling your projects? Or maybe you know someone?

Source/LinqToDB.Tools/CodeModel/Types/WellKnownTypes.cs
```diff
@@ -479,7 +479,7 @@ public static class Common
 				public static IType Converter { get; } = Parser.Parse(typeof(Converter));
 
 				/// <summary>
-				/// <see cref="Converter.ChangeTypeTo{T}(object?, MappingSchema?)"/> method reference.
+				/// <see cref="Converter.ChangeTypeTo{T}(object?, MappingSchema?,ConversionType)"/> method reference.
```
MaceWindu, line 482: Adding extra parameter requires changes to method call generation for both methods

Source/LinqToDB/Common/ConversionType.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace LinqToDB.Common
+{
+	public enum ConversionType
```
MaceWindu, line 5: Add xml-doc

Source/LinqToDB/Common/ConvertInfo.cs
```diff
@@ -118,54 +139,82 @@ static void Set(ConcurrentDictionary<DbDataType,ConcurrentDictionary<DbDataType,
...
+				case ConversionType.ToDatabase:
+				{
+					if (_toDatabaseExpressions != null &&
+						_toDatabaseExpressions.TryGetValue(from, out var dic) && dic.TryGetValue(to, out var li))
+						return li;
+					break;
+				}
+			}
+
+			{
```
MaceWindu, line 162: remove brackets

MaceWindu (changes requested):

Source/LinqToDB.Tools/CodeModel/Types/WellKnownTypes.cs
igor-tkachev: This parameter has a default value. Generated code should be fine.

Source/LinqToDB/Common/ConvertInfo.cs
igor-tkachev: It will make this code not compliable as these brackets limit the scope of some local variables.

MaceWindu (approved):

Source/LinqToDB/Common/Converter.cs
viceroypenguin: The issue is not for direct consumers - only source compatibility is necessary for them, because they will recompile when upgrading. The issue is for intermediate consumers: App -> Lib -> L2db.  If App upgrades to a new version of l2db, but lib does not, then binary compatibility is important for lib to keep working. Examples would be: https://github.com/akkadotnet/Akka.Persistence.Sql, https://github.com/OrchardCoreContrib/OrchardCoreContrib, https://github.com/Synnotech-AG/Synnotech.Linq2Db, etc.. That said, these methods are _probably_ not directly referenced by such libraries, so it's probably fine.

PR #4392, opened 2024-01-26
Add `Sql.Regexp`

### LINKED ISSUE #698: Use Regex in Query
Do we support Regex in Linq (converted to sql)?

mysql:
https://stackoverflow.com/questions/18780194/how-to-regex-in-a-sql-query

sqlite:
https://stackoverflow.com/questions/5071601/how-do-i-use-regex-in-a-sqlite-query

postgres:
https://www.postgresql.org/docs/9.0/static/functions-matching.html

oracle:
https://docs.oracle.com/cd/B12037_01/appdev.101/b10795/adfns_re.htm

sap hana:
http://help-legacy.sap.com/saphelp_hanaplatform/helpdata/en/b4/0d483dd34d47aa9cc89b4d8a6e617e/content.htm?frameset=/en/dc/f1045ce51d45119cfe8ba17cf9da4f/frameset.htm&current_toc=/en/2e/1ef8b4f4554739959886e55d4c127b/plain.htm&node_id=113

mssql:
http://devnambi.com/2016/sql-server-regex/

jogibear9988:
supersedes #1077 

Fixes #698


Source/LinqToDB/Sql/Sql.Strings.cs
```diff
@@ -286,5 +286,20 @@ protected override SqlExpression TruncateExpression(ISqlExpression value, ISqlEx
...
 		#endregion
+
+		#region Regex
+
+		[Expression(PN.MySql, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[Expression(PN.SQLite, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)] // only possible with custom user function, see TableFunctionTests.Regex
+		[Expression(PN.PostgreSQL, "{0} ~ {1}", ServerSideOnly = true, IsPredicate = true)]
+		[Expression(PN.Oracle, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[Expression(PN.SapHana, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[Expression(PN.SqlServer, "dbo.regex({0}, {1})", ServerSideOnly = true, IsPredicate = true)] // supported via https://github.com/infiniteloopltd/SQLServerRegex
```
viceroypenguin, line 297: I would remove this line; sql server should throw an exception, and if consumers want to use regex, they can add the library specifically. Having this will create a false expectation for the consumer, if they look at the failed query and see `dbo.regex()` in the querystring, and wonder why they don't have the `regex` method.

jogibear9988:
in #1077 you wrote I should create an extension lib.

I could do this, but could I also add this expression https://github.com/linq2db/linq2db/pull/4392/files#diff-2a6a9163861fbdc4015c6f32a3373c85469cb401552aea366b2d2c07ede63fb5R1198
in an external lib?

I would like to have the support for Regex in Query.
And should this be an extra lin in linq2db, or should it be my own?

Source/LinqToDB/Sql/Sql.Strings.cs
```diff
@@ -286,5 +286,20 @@ protected override SqlExpression TruncateExpression(ISqlExpression value, ISqlEx
...
+		public static bool RegExpIsMatch(string text, string expression)
```
viceroypenguin, line 298: Add xmldocs for this method.

viceroypenguin (changes requested):

sdanyliv:
I'm still against using `new Regex()` in LINQ queries. Don't. Each database engine has it's own regular expression dialect non comparable with .NET `Regex` . We discussed that several times!

Introduce only `Sql.Regexp()` function instead `RegExpIsMatch` and do not create Expression mapping

jogibear9988:
It's okay for me, if I could support it in a extension, so I could add my own "Expressions" so it works on my system with std regex

jogibear9988:
Yes, I wanted to now, is there a way for me, to add the line I did in Expressions.cs later in my usercode?

So that in my code we still could user the "Regexp" object.

We mostly use simple Regexes, wich would worknearly  in all DBs.
(and in sqlserver and sqlite they would in any case, cause there C# code is called)

Source/LinqToDB/DataProvider/SqlServer/SqlServerConfiguration.cs
```diff
@@ -10,5 +10,13 @@ public static bool GenerateScopeIdentity
 			get => SqlServerOptions.Default.GenerateScopeIdentity;
 			set => SqlServerOptions.Default = SqlServerOptions.Default with { GenerateScopeIdentity = value };
 		}
+
+		/// <summary>
+		/// You can specify a UserDefinedFunction wich is called to run a regex against your data.
```
jods4, line 15: typo in "wich".

Source/LinqToDB/Sql/Sql.Strings.cs
jods4: I tend to agree, I wouldn't commit core linq2db functions to custom libraries.
What if MS adds Regex support in the future (quite possible) and it's not exactly the same dialect / not 100% compatible with this lib?

If that lib is popular enough to be included in main linq2db lib, I'd put it on a specific static class: `InfiniteLoop.Regex(..)`.

Source/LinqToDB/DataProvider/SqlServer/SqlServerConfiguration.cs
```diff
@@ -10,5 +10,13 @@ public static bool GenerateScopeIdentity
...
+		/// where {0} is the text and {1} is the expression.
+		/// Look at https://github.com/infiniteloopltd/SQLServerRegex for a sample
+		/// in this case the value would be "dbo.regex({0}, {1})"
+		/// </summary>
+		public static string? RegexUserDefinedFunctionName { get; set; }
```
jods4, line 20: Do we really want to have such specific options? It's basically exactly the same level of abstraction as `[Expression]` but for a single specific function.
We could imagine similar options for many other SQL libraries -- and I really don't think we want to do that.
Moreover, I'd rather reserve the "official" linq2db regex function to a possible future "native" regex function in SQL Server (should it ever happen).

I'd rather have: 
- Docs/fully working example code to create such a static function with `[Expression("dbo.regex({0}, {1})")]`.
- Shipping such functions in different namespaces that can be imported  (`using LinqToDB.InfiniteLoop`) and/or different libraries.

jogibear9988: I'd move the complete code to an external library, if someone could tell me, how I could accomplish the add of the C# regex class to Expressions.cs, so I could use that in my queries

jods4: @jogibear9988 This PR is currently doing a `(string, string)` extension, isn't it?
Do you mean you'd like to be able to handle `Regex.IsMatch(..)`?

jogibear9988: yes, I'd like to use regex.ismatch in my queries

viceroypenguin: I'm gonna be honest, that's terrifying.

jogibear9988: why?
users can create complex queries at our frontends wich also support regex, and then we create expressions wich then are run agains database queries or sometimes IEnumerable queries.

jogibear9988:
efcore seem to support this for postgres:

https://www.npgsql.org/efcore/mapping/translations.html#string-functions

image

jogibear9988:
and also in sqlite they seem to support them:

image

https://github.com/dotnet/efcore/issues/28078

jogibear9988:
They also seem to automaticly map a regex function for sqlite (if you'd like to)

https://github.com/dotnet/efcore/issues/18845

Source/LinqToDB/DataProvider/SqlServer/SqlServerConfiguration.cs
viceroypenguin: I'm not opposed to `Sql.Regexp()`, because that is explicit that we're sending strings to the db server for consumption by the db engine to interpret as it does. 

That said, in the case of pgsql, `Sql.Regexp()` only covers one case of using regex.
image

I'm opposed to `Regex.IsMatch()` because the interpretation of the regex is different between .net and pgsql. For example, this is a somewhat simple and valid .net regex, with syntax highlighting indicating it is valid. However, when translated to pgsql, it would use the pgsql semantics which would not recognize `(<?name>)` syntax or the `\k<name>` syntax, and reject it. 
image

This would be, I think, unexpected behavior for the consumer, and we have no way of documenting it _on the `Regex.IsMatch()`_ method, because it is not our method. 

I understand that pgsql has added it to their translation support, and sqlite too apparently, but I am not as certain that linq2db should do it.

jogibear9988: In my oppinion, a user would not expect that all regexes work, but I also don't see a problem, if they not work, they fail.
I showed that also EF Added them to Sqlite and also for Cosmos.

If it will not be accepted, I would like to have a way, that I can add support for it myself. (And this is the thing I don't know how, cause I can not add anything to Expressions.cs)

Shane32: My two cents: in general I agree that Sql.Regex provides better clarity that the regular expression is processed by the database engine and is not the .NET dialect. This is always a gray area perhaps, such as how `==` is typically case insensitive in SQL and case sensitive in C#. Here I think using Sql.Regex is appropriate, but it is not illogical for Regex.IsMatch to be mapped either.

I also agree that linq2db should have global extension points available so users can map Regex.IsMatch or any other function to SQL as they choose.

Shane32: Also, if EF Core maps Regex.IsMatch then I’d probably go that route instead.

sdanyliv: Regex.IsMatch mapped only by PostgreSQL provider, and I do not think that it was good idea.

jods4: @jogibear9988 having specific options for each built-in methods is not a maintainable approach.
What if someone wants to map `Regex.Split(..)` as well and maybe have instance methods handled, and...

But hear me out, I think we could approach this in a more extensible way.

Let's have a `Register` method (not the best name, just for the sake of discussion) that works like `[Expression]` but for any method not under your control.

```cs
// Expression must be a lambda that consists of a single method call or property (static or instance).
// sql is a pattern like [Expression]: {0} is this (if any), {1}, {2} are parameters
void Register(Expression expr, string sql);
```

Internally linq2db can use it to register all default translations (SQL standard).
```cs
// linq2db initialization
Register((string s) => s.ToLower(), "lower({0})");
Register(() => DateTime.Now, "current_timestamp");
```

DataProviders should inherit the default and have an opportunity to add/remove/modify them:
```cs
// Oracle provider
Register(() => DateTime.Now, "sysdate");
```

And as a 3rd layer, this function should be public so that users can add/remove/modify their provider with custom library functions or opiniated preferences.
Maybe best to expose this as a callback in options so that it's only accessible at startup?
```cs
interface ISqlExtensions
{
  void Register(Expression, string);
  void Unregister(Expression);
}

record DataProviderOptions
{
  public Action<ISqlExtensions>? Extensions = null;
}

// In your own code
var options = new DataProviderOptions
{
  Extensions = x => 
  {
    x.Register(() => Regex.IsMatch("", ""), "dbo.regex({1}, {2})");
  },
};
```

jogibear9988: I think the functionality in Expression.cs (where I at the moment register the regex.match) should be acessible in any way. So that it works as I do at the moment.
I think your approche is to limited.

jogibear9988: > Regex.IsMatch mapped only by EF Core PostgreSQL provider, and I do not think that it was good idea.

For me it looks like, they have also mapped for Cosmos and SqLite

jogibear9988: I think also, everybody knows (or needs to know), that querys against objects are different than agains a database.
Also if I have a JOIN and it returns null, and I filter for a joined sub property, this completely works in the database but not against objects. 
So everybody would understand, regex in a database behave differently.

And also at the moment, we do not support every method or property of a datatype, only some, so this is not an issue.

jogibear9988: But as I also saind, if we don't want to do this change, it is okay.

I'm only looking for a way, how I can extend Expressions.cs default... from outside.
Should we add a method? Problem is, that once initialized, adding something to it, does not have an effect. So it needs refactoring

jods4: > I think your approche is to limited.

Can you elaborate? Seems much more generic and flexible to me. In any case, it covers `dbo.regex({1}, {2})` that you're wanting in this PR.

Here's why I think your approach is limited and worst:
- You introduce a new option `RegexUserDefinedFunctionName` that will only work for `Regex.IsMatch`. No other function is supported. Right now there's PR #3624 that's adding a lot of ugly code to support NodaTime that would benefit from a generic way to do this.
- `RegexUserDefinedFunctionName` is static. What if I want a different config for two different DB instances? How could we easily run tests in parallel?

IMHO this design does not scale to more functions/libraries and is not easily maintainable long-term. 
It moves inside linq2db core something that could trivially be done in user-land.

jods4: > I'm only looking for a way, how I can extend Expressions.cs default... from outside.
Should we add a method? Problem is, that once initialized, adding something to it, does not have an effect. So it needs refactoring

I think this is the way to go and it's basically my proposal above.
Yes, it's a non-trivial refactoring but I think it is the best option long-term. Only issue right now is that there's the large @sdanyliv parser refactoring going on. We need to merge this before doing other large changes.

jogibear9988: okay for me

jods4: BTW will gladly help with this. It's been on my mind for a while.

sdanyliv:
It is Roji's decision because he created PostgreSQL provider, and probably ported to SQLite. And I think he is wrong here.

jogibear9988:
I don't think so ;-) , but I'll wait till we have a possibility to add my own mappings for Standard methods.
I also need it for the Guid.ToString call in my other pull request.

jogibear9988:
So, I now removed the functionality from Linq2db and added a sample in the Tests.
So maybe we can merge this, it test if regex registering is possible and usable.
I this it's usefull as a information.

viceroypenguin:
@jogibear9988 looks like you remove both the expression and the `Sql.Regexp` method. You will need to restore the `Sql.` method.

jogibear9988:
> @jogibear9988 looks like you remove both the expression and the `Sql.Regexp` method. You will need to restore the `Sql.` method.

No, I moved everything to an extra file. Cause for me it only makes sense if SqlServer is also supported. So I only added it to the tests as a sample. If someone likes to use it, he could copy the file from the tests. Or maybe I create a extra Nuget for Linq2db -> For Example Linq2DB.Regex, wich then maybe also contains the assembly for sqlserver

jogibear9988:
So we can merge this, as a sample, or close the pull req.
As I can now add the support from outside it's okay

viceroypenguin:
Given that the code doesn't build without changes to the `Sql.` class, I'll go ahead and close it.

jogibear9988:
@viceroypenguin now I understand, it should build. This was wrong, I'll fix it

Tests/Linq/UserTests/SqlRegex.cs
```diff
@@ -0,0 +1,75 @@
...
+				new SQLiteFunctionAttribute() { Name = "REGEXP", Arguments = 2, FuncType = FunctionType.Scalar },
+				new MyRegEx());
+			}
+		}
+
+		/// <summary>
+		/// For SQLServer support you need to register the function from
+		/// https://github.com/infiniteloopltd/SQLServerRegex
+		/// </summary>
+		public static void AddRegexSupport()
```
MaceWindu, line 53: I would prefer to avoid explicit registrations. Wether we register it automatically or don't map members at all

Tests/Linq/UserTests/SqlRegex.cs
```diff
@@ -0,0 +1,75 @@
...
+			{
+				return Regex.IsMatch(Convert.ToString(args![1]!)!, Convert.ToString(args![0]!)!);
+			}
+		}
+		public static void EnableSqliteRegex(DataConnection connection)
+		{
+			EnableSqliteRegex(connection.Connection);
+		}
+
+		public static void EnableSqliteRegex(DbConnection connection)
```
MaceWindu, line 32: we don't need it. User could register function from his code, if he needs it

Tests/Linq/UserTests/Issue698Tests.cs
```diff
@@ -0,0 +1,51 @@
...
+	{
+		[Table]
+		public class InfeedAdvicePositionDTO
+		{
+			[Column] public int Id { get; set; }
+			[Column] public string? Text { get; set; }
+		}
+
+		[Test]
+		public void Test698([IncludeDataSources(TestProvName.AllSQLite, TestProvName.AllPostgreSQL, TestProvName.AllOracle)] string context)
```
MaceWindu, line 30: all supported databases and APIs should be tested

MaceWindu (changes requested):

Tests/Linq/UserTests/SqlRegex.cs
```diff
@@ -0,0 +1,75 @@
...
+		/// </summary>
+		/// <param name="text">The string to Test against.</param>
+		/// <param name="expression">The Regular Expression. (Datbase specific)</param>
+		/// <returns></returns>
+		[LinqToDB.Sql.Expression(ProviderName.MySql, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.SQLite, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.PostgreSQL, "{0} ~ {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.Oracle, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.SapHana, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.SqlServer, "dbo.regex({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
```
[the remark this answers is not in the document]
jogibear9988: it's only a sample class in the tests. but I can remove

Tests/Linq/UserTests/SqlRegex.cs
jogibear9988: look at efcore, there EF does the registration for you

MaceWindu:
FYI https://github.com/linq2db/linq2db/blob/9b1094cc5513b0a5dd6253aedce2b47b3456700d/Tests/Linq/Linq/MemberTranslatorTests.cs

jogibear9988:
@MaceWindu whats the advantage to use this instead of Linq.Expressions.MapMember?

MaceWindu:
Check some basic notes here https://github.com/linq2db/linq2db/wiki/Linq-To-DB-6#member-translators

jogibear9988:
SQLServer 2024 seems to include Regexp support directly: https://www.sqlservercentral.com/articles/announcing-sql-server-2025

Tests/Linq/UserTests/SqlRegex.cs
```diff
@@ -0,0 +1,81 @@
...
+		[LinqToDB.Sql.Expression(ProviderName.Oracle, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.SapHana, "{0} REGEXP {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.DB2, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.DB2LUW, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.DB2zOS, "REGEXP_LIKE({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.Firebird, "{0} SIMILAR TO {1}", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.Informix, "regex_match({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.ClickHouse, "match({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		[LinqToDB.Sql.Expression(ProviderName.SqlServer, "dbo.regex({0}, {1})", ServerSideOnly = true, IsPredicate = true)]
+		public static bool IsMatch(string text, string expression)
```
jods4, line 76: Add `[StringSyntax(StringSyntaxAttribute.Regex)] ` on string regex parameters?
It tells VisualStudio to offer syntax support (coloring, completions, bracket highlighting, etc.)

The regex flavor might not be 100% for all RDBMS but it's better than nothing at all?

PR #4397, opened 2024-01-27
Fix Oracle BulkCopy parameters

### LINKED ISSUE #4385: BulkInsert DateOnly columns fails
### Describe your issue

We tried to migrate some of our C# properties to `DateOnly`, as support was recently added in #3502 .

C# property is typed as `DateOnly`, Oracle column is `date` (includes time but that's the best fit...), this is the default mapping in linq2db Oracle provider.

Reading data into `DateOnly` properties, sending normal updates with `DateOnly` parameters: work flawlessly! 😃 

But when doing a BulkInsert, the `DateOnly` value is not converted to `DateTime` before being put into `OracleParameter.Value`. As ODP.NET has no support for `DateOnly` values yet, it then tries a generic conversion during query execution that fails.

I guess the path that extracts parameter values in BulkInsert is not going through all the mappings of normal statements.

We observe this with Oracle "INSERT INTO + array parameter values" strategy, not sure if it reproduces with other strategies (multiple INSERT or native BulkCopy).

@vinceroypenguin did you check/test BulkInsert when you did #3502?

### Environment details

`Linq To DB` version: *5.3.2*

Database (with version): *Oracle 19*

ADO.NET Provider (with version): *Oracle.ManagedDataAccess.Core 3.21.130*

.NET Version: *.net 8*

jods4:
Fix #4385

### DIFF
--- a/Source/LinqToDB/DataProvider/Oracle/OracleBulkCopy.cs
+++ b/Source/LinqToDB/DataProvider/Oracle/OracleBulkCopy.cs
@@ -362,18 +362,21 @@ namespace LinqToDB.DataProvider.Oracle
 
 		bool Execute(MultipleRowsHelper helper, List<object> list)
 		{
+			var valueConverter = new BulkCopyReader.Parameter();
+
 			for (var i = 0; i < helper.Columns.Length; i++)
 			{
-				var column   = helper.Columns[i];
-				var dataType = column.DataType == DataType.Undefined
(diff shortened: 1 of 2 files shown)


jods4:
Thanks @MaceWindu ! 🙏

viceroypenguin (approved):

igor-tkachev (approved):

PR #4399, opened 2024-01-27
Concurrency issue in DataConnection options

### LINKED ISSUE #4398: Concurrency issue in DataConnection options
One of our tests failed:

```
System.AggregateException : One or more errors occurred.
----> System.ServiceModel.FaultException : System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
at System.Collections.Generic.Dictionary2.ValueCollection.Enumerator.MoveNext()    at LinqToDB.Common.OptionsContainer1.Apply[TA](TA obj) in D:\a\1\s\Source\LinqToDB\Common\OptionsContainer.cs:line 120
at LinqToDB.DataOptions.Apply(DataConnection dataConnection) in D:\a\1\s\Source\LinqToDB\DataOptions.cs:line 120
at LinqToDB.Data.DataConnection..ctor(DataOptions options) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.cs:line 424
at LinqToDB.Data.DataConnection..ctor(String configurationString) in D:\a\1\s\Source\LinqToDB\Data\DataConnection.cs:line 53
at LinqToDB.Remote.LinqService.CreateDataContext(String configuration) in D:\a\1\s\Source\LinqToDB\Remote\LinqService.cs:line 57
at Tests.Remote.ServerContainer.TestLinqService.CreateDataContext(String configuration) in D:\a\1\s\Tests\Base\Remote\ServerContainer\TestLinqService.cs:line 12
at LinqToDB.Remote.LinqService.ExecuteReader(String configuration, String queryData) in D:\a\1\s\Source\LinqToDB\Remote\LinqService.cs:line 322
at LinqToDB.Remote.Wcf.WcfLinqService.LinqToDB.Remote.Wcf.IWcfLinqService.ExecuteReader(String configuration, String queryData) in D:\a\1\s\Source\LinqToDB.Remote.Wcf\WcfLinqService.cs:line 62

```

https://dev.azure.com/linq2db/linq2db/_build/results?buildId=12299&view=ms.vss-test-web.build-test-results-tab&runId=266981&resultId=106614&paneView=debug

igor-tkachev:
Closes #4398

### DIFF
--- a/Source/LinqToDB/Common/OptionsContainer.cs
+++ b/Source/LinqToDB/Common/OptionsContainer.cs
@@ -96,6 +96,8 @@ namespace LinqToDB.Common
 			return Find<TSet>() ?? defaultOptions;
 		}
 
+		readonly object _sync = new ();
+
 		/// <summary>
 		/// Returns options set by set type <typeparamref name="TSet"/>. If options doesn't contain specific options set, it is created and added to options.
 		/// </summary>
@@ -103,13 +105,15 @@ namespace LinqToDB.Common
 		/// <returns>
 		/// Returns options set by set type <typeparamref name="TSet"/>. If options doesn't contain specific options set, it is created and added to options.
 		/// </returns>
+		[Pure]
 		public virtual TSet Get<TSet>()
 			where TSet : class, IOptionSet, new()
 		{
 			if (Find<TSet>() is { } set)
 				return set;
 
-			_sets = new (_sets ?? []) { [typeof(TSet)] = set = new() };
+			lock (_sync)
+				_sets = new (_sets ?? []) { [typeof(TSet)] = set = new() };
 
 			return set;
 		}
--- a/Source/LinqToDB/Common/OptionsContainer.cs
+++ b/Source/LinqToDB/Common/OptionsContainer.cs
@@ -109,7 +109,7 @@ namespace LinqToDB.Common
 			if (Find<TSet>() is { } set)
 				return set;
 
-			(_sets ??= new())[typeof(TSet)] = set = new();
+			_sets = new (_sets ?? []) { [typeof(TSet)] = set = new() };
 
 			return set;
 		}
--- a/Source/LinqToDB/Common/OptionsContainer.cs
+++ b/Source/LinqToDB/Common/OptionsContainer.cs
@@ -58,7 +58,7 @@ namespace LinqToDB.Common
 			return ReferenceEquals(original, options) ? (T)this : WithOptions(options);
 		}
 
-		Dictionary<Type,IOptionSet>? _sets;
+		volatile Dictionary<Type,IOptionSet>? _sets;
 
 		/// <summary>
 		/// Provides access to option sets, stored in current options object.


Source/LinqToDB/Common/OptionsContainer.cs
```diff
@@ -109,7 +109,7 @@ public virtual TSet Get<TSet>()
 			if (Find<TSet>() is { } set)
 				return set;
 
-			(_sets ??= new())[typeof(TSet)] = set = new();
+			_sets = new (_sets ?? []) { [typeof(TSet)] = set = new() };
```
jods4, line 112: You need a memory barrier here, otherwise it's possible for concurrent code to see the new `_sets` pointer it sees the initialized `Dictionary` fields (write re-ordering).

Use `Thread.VolatileWrite(ref _sets, ...)` or declare `volatile _sets`.

viceroypenguin, line 112: this definitely fixes so there's no race condition in manipulating the dictionary, but a) are we concerned about memory/perf issue in creating a new one every time? and b) are we concerned about potential loss of assignment if two threads are creating separate dictionaries at the same time (aka last assignment wins, and doesn't have the `typeof(TSet)` from the previous one)?

jods4: This is not thread-safe lock-free code for two reasons:
- I mentioned above that the "publish" of new reference is not a "releasing write", it lacks a memory barrier.
- You are right: "read `_sets`, modify, then write" will lose data if there are concurrent changes.

If we want to keep things lock-free, the solution to both problems at once is to use `Interlocked.CompareExchange` to pulbish the new dictionary, ensuring that the overwritten reference is the one that was read and modified (and looping if there was a concurrent change).

RE memory usage, if we have concerns `ImmutableDictionary` could be an option with better perf.

igor-tkachev: Performance should not be an issue as this is a very rare case and all the option initializations should be done once at the beginning.
Last assignment wins - agreed.

viceroypenguin: I'd probably prefer to see `ConcurrentDictionary<>` be used. Perhaps a `Lazy<ConcurrentDictionary<>>` if we want to protect initialization.

jods4: > Last assignment wins - agreed.

What?
This is not assignment to the same option (which I would agree).
It might be different options and one would simply be lost -- why not fix this?

Source/LinqToDB/Common/OptionsContainer.cs
```diff
@@ -96,20 +96,24 @@ public TSet FindOrDefault<TSet>(TSet defaultOptions)
...
+		[Pure]
 		public virtual TSet Get<TSet>()
 			where TSet : class, IOptionSet, new()
 		{
 			if (Find<TSet>() is { } set)
 				return set;
 
-			(_sets ??= new())[typeof(TSet)] = set = new();
+			lock (_sync)
+				_sets = new (_sets ?? []) { [typeof(TSet)] = set = new() };
```
viceroypenguin, line 116: I like this better, but with a `lock`, we don't need a new assignment every time.
```suggestion
			lock (_sync)
				(_sets ??= new())[typeof(TSet)] = set = new();
```

Source/LinqToDB/Common/OptionsContainer.cs
igor-tkachev: Fixed.

viceroypenguin (changes requested):
if we did `ConcurrentDictionary<Type, IOptionSet> _sets = [];` and `_sets.GetOrAdd(typeof(TSet), _ => new());`, then we don't need to worry about concurrency issues.

Source/LinqToDB/Common/OptionsContainer.cs
igor-tkachev: `ConcarrentDictionary` is good if read/write ration is 90/10 or less. Here we are going to have 1M+/1.

jods4: @viceroypenguin `ConcurrentDictionary<>` is pretty heavy. 
As options are only modified a few times at startup and then used a lot as a readonly structure, I tend to think a plain `Dictionary` or `SortedList` is most efficient here.
I'd favor `ConcurrentDictionary` if there was sustained concurrent modifications.

Source/LinqToDB/Common/OptionsContainer.cs
jods4: Yeah the approach with a lock is safe as long as we publish a new `Dictionary` and don't mutate the existing one (which would require a lock on readers as well). 👍

jods4 (approved):

viceroypenguin (approved):

Source/LinqToDB/Common/OptionsContainer.cs
igor-tkachev: > What?

This code works for non-standard options that are not predifined. Such as specific provider options. If we do not have it in Option Container, we add it once. It can happen if we run the following code in parallel:

```c#
DefaultDataOptions = DefaultDataOptions.WithOption<PostgreSQLOptions>(...)
DefaultDataOptions = DefaultDataOptions.WithOption<SqlServerOptions>(...)
```

viceroypenguin: The new `lock` code does. The previous version did not. Either `PostgreSQLOptions` or `SqlServerOptions` would get added, but not both, because they were assigned in two different dictionaries and only one got saved to `_sets`. The current version avoids that concern though.

Source/LinqToDB/Common/OptionsContainer.cs
igor-tkachev: If we do not assign it every time, we need to use lock everywhere we read `_sets` which is much worse.
Again, this code will be called once for Option Container and all its copies. Once for each exotic option for the entire application run-time.

jogibear9988: but do we really need "volatile"?

MaceWindu (approved):

PR #4406, opened 2024-01-30
Fix registration by `AddLinqToDBContext<TContext, TContextImplementation>`

### LINKED ISSUE #4326: `AddLinqToDBContext<TContext, TContextImplementation>` could register incorrect `DataOptions` type
See https://github.com/linq2db/linq2db/discussions/4324#discussioncomment-7382490

We should:
- use `DataOptions<TContextImplementation>`
- probe and register `DataOptions<TContext>` as fallback for backward compatibility if `DataOptions<TContextImplementation>` constructor missing

MaceWindu:
Fix #4326

### DIFF
--- a/Source/LinqToDB.AspNet/ServiceConfigurationExtensions.cs
+++ b/Source/LinqToDB.AspNet/ServiceConfigurationExtensions.cs
@@ -158,32 +158,48 @@ namespace LinqToDB.AspNet
 			where TContextImplementation : TContext, IDataContext
 			where TContext: IDataContext
 		{
-			var hasTypedConstructor = HasTypedContextConstructor<TContextImplementation>();
+			var constructorType = HasTypedContextConstructor<TContextImplementation, TContext>();
 
 			serviceCollection.TryAdd(new ServiceDescriptor(typeof(TContext), typeof(TContextImplementation), lifetime));
-			serviceCollection.TryAdd(new ServiceDescriptor(typeof(DataOptions<TContext>),
-				provider => new DataOptions<TContext>(configure(provider, new DataOptions())),
-				lifetime));
 
-			if (!hasTypedConstructor)
+			if (constructorType == 1)
+			{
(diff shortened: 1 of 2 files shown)


Source/LinqToDB.AspNet/ServiceConfigurationExtensions.cs
```diff
@@ -158,32 +158,48 @@ public static class ServiceConfigurationExtensions
...
+			{
 				serviceCollection.TryAdd(new ServiceDescriptor(typeof(DataOptions),
 					provider => configure(provider, new DataOptions()), lifetime));
+			}
 
 			return serviceCollection;
 		}
 
-		static bool HasTypedContextConstructor<TContext>() where TContext : IDataContext
+		static int HasTypedContextConstructor<TContextImplementation, TContext>()
```
viceroypenguin, line 186: can we return an `enum` here, even one local to this class? just for clarity of the different return values

Source/LinqToDB.AspNet/ServiceConfigurationExtensions.cs
```diff
@@ -158,32 +158,48 @@ public static class ServiceConfigurationExtensions
...
-			var hasTypedConstructor = HasTypedContextConstructor<TContextImplementation>();
+			var constructorType = HasTypedContextConstructor<TContextImplementation, TContext>();
 
 			serviceCollection.TryAdd(new ServiceDescriptor(typeof(TContext), typeof(TContextImplementation), lifetime));
-			serviceCollection.TryAdd(new ServiceDescriptor(typeof(DataOptions<TContext>),
-				provider => new DataOptions<TContext>(configure(provider, new DataOptions())),
-				lifetime));
 
-			if (!hasTypedConstructor)
+			if (constructorType == 1)
```
viceroypenguin, line 165: would this be better as a switch?

viceroypenguin (approved):

PR #4407, opened 2024-01-30
Support primary constructor parameters as query parameters

### LINKED ISSUE #4359: Ensure primary constructor parameters supported as parameters by all providers
We generate parameter names (and aliases) based on identifiers in user code. For a long time we had a good trust in them as compiler will prevent users from using bad names and we didn't cared about "non-conventional" names generation using metadata.

All changed with C# 12 and primary constructor parameters. With this feature compiler leaks real backing field names to expressions and e.g. it result in bad parameter names like `@<myvalue>`.

We need to add test for such case and ensure all providers escape/normalize parameter names properly for at least `<` and `>` characters.

MaceWindu:
Fix #4359


MaceWindu:
looks like there is nothing to fix already

PR #4411, opened 2024-02-05
Use version-less baselines folders

MaceWindu:
Try to remove TFM from baselines as different TFMs should produce same baselines. Should reduce amount of duplicate baselines by 60%


viceroypenguin:
Why net70 instead of net80?

MaceWindu:
because it's master, it doesn't have net80 testing yet

PR #4420, opened 2024-02-07
Add a concurrency check to inform users of invalid use

### LINKED ISSUE #4405: Add Concurrency Check with explicit exceptions
Per the discussion in #4404, we should add an explicit concurrency check, similar to EFC, which would inform the user of invalid behavior. We can use a similar class to https://github.com/dotnet/efcore/blob/main/src/EFCore/Infrastructure/Internal/ConcurrencyDetector.cs

viceroypenguin:
Adds a concurrency detector, which will throw when the `DataConnection` has been used on multiple threads or in simultaneous situations incorrectly.

Fixes #4405

### DIFF
--- a/Tests/Linq/Common/ConcurrencyTests.cs
+++ b/Tests/Linq/Common/ConcurrencyTests.cs
@@ -1,0 +1,49 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+
+using LinqToDB;
+using LinqToDB.Linq;
+
+using NUnit.Framework;
+
+namespace Tests.Common
+{
+	[TestFixture]
+	public class ConcurrencyTests : TestBase
+	{
+		[Test]
+		public void SequentialUsagesDoNotThrow([DataSources] string context)
+		{
+			using var db = GetDataContext(context);
+
+			_ = db.Parent.ToList();
+			_ = db.Parent.ToList();
+		}
+
+		[Test]
+		public async Task ConcurrentUsageMultipleThreadsThrows([DataSources] string context)
+		{
+			using var db = GetDataContext(context);
+
+			await using var enum1 = db.Parent.AsAsyncEnumerable().GetAsyncEnumerator();
+			await enum1.MoveNextAsync();
+
+			Assert.ThrowsAsync<LinqException>(async () =>
+				await Task.Run(() => db.Parent.ToListAsync()));
+		}
+
+		[Test]
+		public void ConcurrentUsageSameThreadThrows([DataSources] string context)
+		{
+			using var db = GetDataContext(context);
+
+			using var enum1 = db.Parent.GetEnumerator();
+			enum1.MoveNext();
+
+			using var enum2 = db.Parent.GetEnumerator();
+
+			Assert.Throws<LinqException>(
+				() => enum2.MoveNext());
+		}
+	}
+}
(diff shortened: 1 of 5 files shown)


Tests/Linq/Common/ConcurrencyTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+			await using var enum1 = db.Parent.AsAsyncEnumerable().GetAsyncEnumerator();
+			await enum1.MoveNextAsync();
+
+			Assert.ThrowsAsync<LinqException>(async () =>
+				await Task.Run(() => db.Parent.ToListAsync()));
+		}
+
+		[Test]
+		public void ConcurrentUsageSameThreadThrows([DataSources] string context)
```
MaceWindu, line 36: Why this test fail? MARS-like queries are valid scenario for most of providers except AFAIK pgsql and mysql.
Real issue is thread-safety

MaceWindu: BTW, this will be a big breaking change for users

jods4: This is a good catch @MaceWindu .
In this test, there is no concurrency in `DbConnection` calls, only overlapping queries.
This is valid on some RDBMS: for example it should work on Oracle, or on MSSQL with MARS enabled.

MaceWindu: Same thread test case is when async operation started without await and another operation followed on same context.

MaceWindu: more details on mars support is here : https://github.com/linq2db/linq2db/pull/2728:

`some providers doesn't suppoort multiple data readers in any form on single connection (mysql.data, mysqlconnector, npgsql, sql server client in non-mars mode)`

viceroypenguin: > Why this test fail? MARS-like queries are valid scenario for most of providers except AFAIK pgsql and mysql. Real issue is thread-safety

Good catch - I wasn't thinking of MARS. `ConcurrentUsageSameThread` shoudl be `DoesNotThrow()` - will correct.

Tests/Linq/Common/ConcurrencyTests.cs
```diff
@@ -26,24 +26,22 @@ public async Task ConcurrentUsageMultipleThreadsThrows([DataSources] string cont
...
+			_ = await enum1.MoveNextAsync();
 
-			Assert.ThrowsAsync<LinqException>(async () =>
+			_ = Assert.ThrowsAsync<LinqException>(async () =>
 				await Task.Run(() => db.Parent.ToListAsync()));
 		}
 
 		[Test]
-		public void ConcurrentUsageSameThreadThrows([DataSources] string context)
+		public void ConcurrentUsageSameThreadDoesNotThrow([DataSources] string context)
```
jods4, line 36: Does it also work on multiple threads?
Threading is orthogonal to concurrency: the key criteria is that no two calls to a single linq2db connection should happen concurrently.

- It is ok to call linq2db from 2 different threads, even on overlapping query/fetches, as long as each call completes/is awaited without concurrency.
- It is an error to call linq2db from a single thread, if a former async call was not awaited.

viceroypenguin: > It is an error to call linq2db from a single thread, if a former async call was not awaited.

`Thread` is used here as a shorthand. If you look at the concurrency check, it uses an `AsyncLocal<>` to manage concurrent calls, so same thread will still throw because it will pick up via being a different async context. 

> It is ok to call linq2db from 2 different threads, even on overlapping query/fetches, as long as each call completes/is awaited without concurrency.

Technically checked in the first unit test.

PR #4433, opened 2024-02-25
CI build tuning

MaceWindu:
- detect nuget generation errors
- switch to `dotnet build` for builds

Looks like something broken in roslyn editorconfig processing in VS 17.9[.1] as it doesn't suppress some IDExxxx errors. If dotnet build has same issue, I will fix/suppress them in code


MaceWindu:
maybe related to #72094, looking into it

MaceWindu:
Ok, that's it. For now I will remove conflicting code-styles. Taking into account that code styles now enable corresponding analyzers we should enable such analyzers explicitly and fix them in future.

PR #4459, opened 2024-03-24
PostgreSQLBulkCopy - replaced GetNativeType by pre calculated npgsqlType

### LINKED ISSUE #4445: PostgreSQLBulkCopy performance degradation for method BulkCopyType.ProviderSpecific
Good afternoon!

After switching to .NET 8 bugs have been fixed in the `ProviderSpecificCopySyncImpl` method https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs#L139

Bugs have been fixed by adding lines:
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs#L167
https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs#L170

Both lines use the `GetNativeType` method - this method is incredibly slow. And it is called in a loop for each column and for each row.

In the line https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs#L170 you can swap the check order to reduce the number of GetNativeType calls:
`if (value is DateTimeOffset dto && _provider.GetNativeType(dataType.DbType) == NpgsqlProviderAdapter.NpgsqlDbType.TimeTZ)`

The `NormalizeTimeStamp` method also calls `GetNativeType'.

At the moment, inserting 10,000 rows with 15 columns (four of them require calling `NormalizeTimeStamp`) via `BulkCopyType.MultipleRows` is about 3 times faster than `BulkCopyType.ProviderSpecific`. At the moment, `BulkCopyType.ProviderSpecific` cannot be used in production - it always works slower.

`BulkCopyType.MultipleRows` spends a lot of time and memory generating SQL strings, and `BulkCopyType.ProviderSpecific` spends 99% of the time calling `GetNativeType` 200,000 times.

As a good solution, you can calculate `GetNativeType` before loops, because it makes no sense to check the column type for each row.

AndreyShipunov:
Fix #4445

Good afternoon.

`NormalizeTimeStamp` now accepts the parameter `NpgsqlProviderAdapter.NpgsqlDbType?`, the responsibility for calculating it has been transferred to the following methods:
 - `setParameter` - as before, GetNativeType(data type.DbType) is simply called without parameters.
 - `ProviderSpecificCopySyncImpl<T>` - now gets `NpgsqlProviderAdapter.NpgsqlDbType?` from the `npgsqlTypes` array.
 - `ProviderSpecificCopyImplAsync<T>` - now gets `NpgsqlProviderAdapter.NpgsqlDbType?` from the `npgsqlTypes` array.
 - `ProviderSpecificCopyImplAsync<T> with NATIVE_ASYNC` - now gets `NpgsqlProviderAdapter.NpgsqlDbType?` from the `npgsqlTypes` array.

The `ProviderSpecificCopy*` methods already had a calculated `npgsqlTypes` array, so I just deleted the `GetNativeType` calls. The main difference is that `npgsqlTypes` is calculated as `GetNativeType(DbType, true)`, and in the `ProviderSpecificCopy*` methods it was called as `GetNativeType(DbType, false)`.  Unfortunately, I do not know if this change is critical.

Half of the tests are not work locally in Rider, both before and after these changes. And in the pull request, I can't run tests. Therefore, I am not sure that my solution is working.

It would be ideal to build a nuget package from this branch and I could check the fix. https://github.com/linq2db/linq2db/issues/4445 , but now azure does not push packages from the branch.

I'm sorry if I didn't fully figure it out somewhere.

### DIFF
--- a/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
+++ b/Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
@@ -164,10 +164,9 @@ namespace LinqToDB.DataProvider.PostgreSQL
 					for (var i = 0; i < columns.Length; i++)
 					{
 						// DBNull.Value : https://github.com/npgsql/npgsql/issues/5330
-						var value = _provider.NormalizeTimeStamp(pgOptions, columns[i].GetProviderValue(item!) ?? DBNull.Value, columnTypes[i]);
-						var dataType = columnTypes[i];
+						var value = _provider.NormalizeTimeStamp(pgOptions, columns[i].GetProviderValue(item!) ?? DBNull.Value, columnTypes[i], npgsqlTypes[i]);
 
-						if (_provider.GetNativeType(dataType.DbType) == NpgsqlProviderAdapter.NpgsqlDbType.TimeTZ && value is DateTimeOffset dto)
+						if (value is DateTimeOffset dto && npgsqlTypes[i] == NpgsqlProviderAdapter.NpgsqlDbType.TimeTZ)
 						{
 							// https://github.com/npgsql/npgsql/issues/5332
 							// reset date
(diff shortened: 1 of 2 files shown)


AndreyShipunov:
@MaceWindu , sorry, can you run tests again.

AndreyShipunov:
The tests on Oracle 12 look unstable. The changes in the branch affect only the provider for PostgreSQL. And in the first run, the Oracle tests were passed.

AndreyShipunov:
May be we shoud re run tests?

MaceWindu (approved):
Thanks!

MaceWindu:
> May be we shoud re run tests?

No need, Oracle indeed has unstable behavior

PR #3510, opened 2022-04-18
Working on IExpressionInterceptor

sdanyliv:

PR #4479, opened 2024-04-15
Added support for IQueryExpressionInterceptor, IMemberTranslator via DataOptions.

sdanyliv:
Replacement for https://github.com/linq2db/linq2db/pull/3510


Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -330,10 +331,10 @@ IEnumerable<T> ReadEnumerator<T>(DataReaderWrapper rd, IExecutionScope? scope, b
 
 						DbDataReader reader;
 
-						if (((IDataContext)DataConnection).UnwrapDataObjectInterceptor is {} interceptor)
+						if (DataConnection is IInterceptable<IUnwrapDataObjectInterceptor> { Interceptor: not null } interceptable)
```
viceroypenguin, line 334: same for others
```suggestion
if (DataConnection is IInterceptable<IUnwrapDataObjectInterceptor> { Interceptor: { } interceptor })
{
  using (...)
    reader = interceptor...
}
```

Source/LinqToDB/Remote/RemoteDataContextBase.Interceptors.cs
```diff
@@ -10,7 +10,8 @@ public abstract partial class RemoteDataContextBase :
 		IInterceptable<IDataContextInterceptor>,
 		IInterceptable<IEntityServiceInterceptor>,
 		IInterceptable<IUnwrapDataObjectInterceptor>,
-		IInterceptable<IEntityBindingInterceptor>
+		IInterceptable<IEntityBindingInterceptor>,
+		IInterceptable<IQueryExpressionInterceptor>
```
viceroypenguin, line 14:
```suggestion
		IInterceptable<IQueryExpressionInterceptor>,
```

Source/LinqToDB/Data/CommandInfo.cs
sdanyliv: Good point will fix that.

viceroypenguin (approved):

Source/LinqToDB/Data/DataConnection.Async.cs
```diff
@@ -138,18 +139,18 @@ public async Task EnsureConnectionAsync(CancellationToken cancellationToken = de
 
 				if (_connection.State == ConnectionState.Closed)
 				{
-					if (_connectionInterceptor != null)
+					if (((IInterceptable<IConnectionInterceptor>)this).Interceptor != null)
```
viceroypenguin, line 142: could you do:
```cs
if (this is IInterceptable<IConnectionInterceptor> { Interceptor: { } interceptor })
```
like the other pattern matching ones?

Source/LinqToDB/Common/OptionsContainer.cs
```diff
@@ -116,13 +116,5 @@ public virtual TSet Get<TSet>()
 
 			return set;
 		}
-
-		protected void Apply<TA>(TA obj)
```
igor-tkachev, line 120: What's wrong with this code?

igor-tkachev (changes requested):
Is this PR about Interceptors or Options? I see a lot of questionable Options refactoring. Can you split it, so we will be able to discuss it separately?

Source/LinqToDB/Common/OptionsContainer.cs
sdanyliv: Tight coupling with DataConnection, DataContext, RemoteDataContext. Options should know nothing about these classes. Also current code has strange unpredictable side effects. 

Removing all of this strange stuff made code readable, simpler, predictable and performant.

igor-tkachev: Then split this PR, so it will be easier to revert it later, when I will need to extend it.

sdanyliv: Reverted this change, will create new PR for that after merging with `version_6` branch.

MaceWindu (approved):

Issue #4475, opened 2024-04-11
Task-spawning code refactoring


MaceWindu:
We should review/fix/simplify places where we create and run Tasks, e.g. 

https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/AsyncExtensions.cs#L30-L68

this leads to exceptions like:
```
Message: 'Start may not be called on a task that has completed.',
    Stack: '   at System.Threading.Tasks.Task.Start(TaskScheduler scheduler)
   at LinqToDB.AsyncExtensions.GetTask[T](Func`1 func, CancellationToken token)
   at LinqToDB.AsyncExtensions.ToArrayAsync[TSource](IQueryable`1 source, CancellationToken token)
```

https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/LinqExtensions.cs#L732-L735

this is not necessary, as we should always work with `IQueryProviderAsync`

MaceWindu:
> I believe your exception arises when calling Start() on a task that was created with a cancellation token that is already cancelled when you call Start().

yes, and while it is completely valid scenario, it adds one to much useless exception to ignore when we already have Task/Operation cancelled duo...

> ... ContinueOnCapturedContext ...

it was added in times when we had little idea about this stuff to just please everyone :). Probably it's a good idea to get rid of it in v6 finally

MaceWindu:
All sub-tasks are done

PR #3401, opened 2022-01-13
Refactoring LINQ Query parsing.

sdanyliv:
Well, trying to simplify things and avoid any workarounds.

### List of fixed issues (incomplete)

- fix #417
- fix #649
- fix #680
- fix #1644
- fix #2264
- fix #2461
- fix #2511
- fix #2948
- fix #3150
- fix #3346
- fix #3586
- fix #3699
- fix #4284 
- fix #4390

### Tasks todo:

- [x] Ensure that CTE do not hold `DataContext` instance in cache.
- [x] MySql provider needs version support. Window Functions are introduced from version 8.0
- [x] Sybase provider needs version support. Window Functions are not supported in ASE
- [x] Firebird provider needs version support. Window Functions are introduced from version 3.0

### New features:

- There is new way how to define `GetTable` call  for `TableFunctionAtrribute` - `TableFromExpression`
```cs
[Sql.TableFunction(Name="GetParentByID")]
public ITable<Parent> GetParentByID(int? id)
{
    return _ctx.TableFromExpression(() => GetParentByID(id));
}
```
Instead of: 
```cs
[Sql.TableFunction(Name="GetParentByID")]
public ITable<Parent> GetParentByID(int? id)
{
    var methodInfo = typeof(FunctionsOld).GetMethod("GetParentByID", new [] {typeof(int?)})!;

    return _ctx.GetTable<Parent>(this, methodInfo, id);
}
```
### Breaking changes:

- SQL AST refactored.

PR #4486, opened 2024-04-21
Improve `IAsyncEnumerable` usage

viceroypenguin:
This PR does two things:
* Updates `InsertWithOutputAsync`, `UpdateWithOutputAsync`, and `DeleteWithOutputAsync` to return `IAsyncEnumerable` properly. This is a binary-compatible change, as the existing methods are kept. However, these methods are now marked with `[Obsolete]` to encourage source migration to the newer `IAsyncEnumerable` method. 
* Adds `QueryToAsyncEnumerable` in places where `QueryToArrayAsync` and `QueryToListAsync` are provided. This allows consumers to use `IAsyncEnumerable` if desired. No obsoletions are planned here, though consumers are allowed and encouraged to upgrade if the query has many rows. 

## 💣 Breaking Change
* The new versions of `XxxWithOutputAsync` are not source compatible, so this is a v6.0 breaking change. Consumers will need to add `.ToListAsync()` or similar to address this change. This is to restore consistency between `InsertWithOutputAsync`/`UpdateWithOutputAsync`/`DeleteWithOutputAsync` with `MergeWithOutputAsync`; this consistency has been chosen to use the more flexible and modern approach.

Replaces #4323


Shane32:
Just for clarity when reviewing this PR, I'd like to point out the difference between the old and new signatures:

```cs
// old
public static Task<TOutput[]> DeleteWithOutputAsync<TSource,TOutput>(
    this IQueryable<TSource>           source,
    Expression<Func<TSource, TOutput>> outputExpression,
    CancellationToken                  token = default)

// new methods
public static IAsyncEnumerable<TOutput> DeleteWithOutputAsync<TSource,TOutput>(
    this IQueryable<TSource>           source,
    Expression<Func<TSource, TOutput>> outputExpression)

public static Task<TOutput[]> DeleteWithOutputAsync<TSource, TOutput>(
    this IQueryable<TSource> source,
    Expression<Func<TSource, TOutput>> outputExpression,
    CancellationToken token)
```

This means that the following code will immediately break for v6 due to the cancellation token not being specified:

```cs
// old syntax; immediately broken with v6
var ret = await query.DeleteWithOutputAsync(x => x.Id);

// required fix to use obsolete method
var ret = await query.DeleteWithOutputAsync(x => x.Id, default);
```

As specified, however, the change is binary compatible - already-compiled code will continue to function with v6, as the optional parameter is built into the caller's compiled code and so will match on the obsolete method.

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -254,6 +254,36 @@ await using (rd.ConfigureAwait(Configuration.ContinueOnCapturedContext))
...
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="objectReader">Record mapping function from data reader.</param>
+		/// <returns>Async sequence of records returned by the query.</returns>
+		public IAsyncEnumerable<T> QueryToAsyncEnumerable<T>(Func<DbDataReader, T> objectReader)
+		{
+			return Impl(objectReader);
+
+			async IAsyncEnumerable<T> Impl(Func<DbDataReader, T> objectReader, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+			{
+				await DataConnection.EnsureConnectionAsync(cancellationToken).ConfigureAwait(Configuration.ContinueOnCapturedContext);
```
jods4, line 269: Didn't you have another PR open to clean up `ConfigureAwait`?

The creep of needing a `Configuration` instance everywhere async to do `ConfigureAwait(ContinueOnCapturedContext)` when it's actually against standard .net patterns and should just be `ConfigureAwait(false)` is real.

Hopefully with upcoming major version number (for the parser rewrite?) we can obsolete this config.

viceroypenguin: No, definitely wasn't me.

Source/LinqToDB/LinqExtensions.Delete.cs
```diff
@@ -74,6 +72,32 @@ public static IEnumerable<TSource> DeleteWithOutput<TSource>(this IQueryable<TSo
...
+		/// Database support:
+		/// <list type="bullet">
+		/// <item>SQL Server 2005+</item>
+		/// <item>Firebird 2.5+ (prior version 5 returns only one record; database limitation)</item>
+		/// <item>PostgreSQL</item>
+		/// <item>SQLite 3.35+</item>
+		/// <item>MariaDB 10.0+ (doesn't support multi-table statements; database limitation)</item>
+		/// </list>
+		/// </remarks>
+		[Obsolete("Will be removed in l2db 7.0")]
```
Shane32, line 95: I suggest a longer obsolete message that explains to the developer the replacement syntax when they are fixing their code.

```suggestion
		[Obsolete("Will be removed in linq2db 7.0. Please remove the cancellation token and use await foreach or ToListAsync instead.")]
```

Not sure exactly, but something like that.

viceroypenguin: I would argue it doesn't really matter, since the only way to call this method is: 
```cs
LinqExtensions.DeleteWithOutputAsync(dbContext, ...);
```
since the `this` has been removed.

Source/LinqToDB/Data/CommandInfo.cs
jods4: Right, it was @MaceWindu in #4475.

Source/LinqToDB/LinqExtensions.Delete.cs
Shane32: So it’s only there for binary compatibility at all then?

sdanyliv:
@viceroypenguin , check mergeability with this PR https://github.com/linq2db/linq2db/pull/3401

Source/LinqToDB/LinqExtensions.Delete.cs
viceroypenguin: Correct

viceroypenguin:
> @viceroypenguin , check mergeability with this PR #3401

@sdanyliv no merge conflicts; and shouldn't have any impact on #3401.

sdanyliv (approved):

igor-tkachev (dismissed):

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -1177,6 +1252,19 @@ public static Task<T[]> QueryToArrayAsync<T>(this DataConnection connection, str
...
 
+		/// <summary>
+		/// Executes command asynchronously and returns array of values.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="sql">Command text.</param>
+		/// <param name="parameter">Command parameter.</param>
+		/// <returns>Async sequence of records returned by the query.</returns>
+		public static IAsyncEnumerable<T> QueryToAsyncEnumerable<T>(this DataConnection connection, string sql, DataParameter parameter)
```
MaceWindu, line 1263: Why we have this overload for linq2db-mapped method?

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -1460,6 +1570,20 @@ public static Task<T[]> QueryToArrayAsync<T>(this DataConnection connection, T t
...
+		/// <summary>
+		/// Executes command asynchronously and returns array of values of specified type.
+		/// </summary>
+		/// <typeparam name="T">Result record type.</typeparam>
+		/// <param name="connection">Database connection.</param>
+		/// <param name="template">This value used only for <typeparamref name="T"/> parameter type inference, which makes this method usable with anonymous types.</param>
+		/// <param name="sql">Command text.</param>
+		/// <param name="parameters">Command parameters.</param>
+		/// <returns>Async sequence of records returned by the query.</returns>
+		public static IAsyncEnumerable<T> QueryToAsyncEnumerable<T>(this DataConnection connection, T template, string sql, params DataParameter[] parameters)
```
MaceWindu, line 1582: and why for template-based specification it is only 2 overloads?

Source/LinqToDB/LinqExtensions.Delete.cs
MaceWindu, line 95: let's not invent another name for Linq To DB 🥲  Please use `Linq To DB 7` here

MaceWindu (changes requested):
minor updates needed

PR #4522, opened 2024-05-30
Do not allow EXISTS optimization to remove columns from UNION queries.

sdanyliv:


viceroypenguin (approved):

MaceWindu (changes requested):
Actually no, access baselines shows bad query. Could we improve test somehow to fail in such cases?

PR #4542, opened 2024-06-19
Improve sequence building

jods4:
This is a #3913 redone on latest v6 master branch.
I'm gonna copy-paste my original description below:

----

### Limitations of current design
Today `ExpressionBuilder` linearly goes through a long list (~70) of `ISequenceBuilder` to find which builder can handle an `Expression`. Each builder is probed by a call to `CanBuild`, in sequence. 

**This is horribly not performant.** If your query uses one of the last builders -- or simply the expression has no builder, which is sometimes a normal case -- you have to go through 70+ checks!! They are virtual calls, many builders repeatedly perform the same casts or checks, and not all builders perform quick, cheap tests. Of course this is only gonna get worse as we add new builders.

The current design attempts at improving performance by dynamically sorting the list to put the most used builders at the front.
This is a good idea but it has two drawbacks. 
1. It adds complexity for the book-keeping of how many times each builder is used. This is mutable concurrent code and requires locks. The amortized perf increases (probably) but the overall perf is worse.
2. It is not that effective. The list is re-sorted only when the first item would change. I assume this is done to avoid constantly reordering the list if 2 builders fight for a specific spot (although that can happen for 1st place!!). 
So once the most used builder is found, nothing changes anymore, even if a builder that becomes commonly used is still way down the list at the moment. 
3. As I said before, "not finding a builder" is sometimes an expected use-case and this always go through the whole list.

### New strategy
Finding a builder now starts with a big `switch` statement on `NodeType` (an int). C# optimizes this depending on the number and distribution of items, it could be a hard-coded binary search (ifs) or a single indirect jump based on a table.

Then a `CanBuild` method is called, but notice: 
1. At this stage, only 1 or 2 (rarely 3) candidates remain.
2. `CanBuild` is now a static method. Short ones will be inlined, which is very likely as most have been simplified thanks to the fact that we know what node to expect when `CanBuild` is called. Most of them are as simple as `=> true`, `=> call.IsQueryable()` or `=> call.IsAsyncExtension()`

One notable case is `ExpressionType.Call` because we have _many_ (most) builders for this node type.
This is handled by a second `switch`, this time on `MethodCallExpression.Method.Name`, which C# will compile into a static hashtable. This is then handled as other node types. In 90% of cases there's only a single candidate left.

There are other benefits, such as clarity. The book-keeping and re-sorting code is gone.
Before it was difficult to see which builder handles what, which is particularly important as they must all be mutually exclusive!
Now you just need to peek at the switch tables to find what is handled, by what.
- Did you know that 3 different builders handle the `Join` method?
- I found a builder that is actually unused! I deleted `CountBuilder` (methods `Count` and `LongCount` are actually handled by `AggregationBuilder`).

### Source generators
I'm not crazy and I didn't write the big switches by hand, that would be too error-prone.
Instead, I added a C# Source Generator to the project, that looks for attributes like `BuildsExpression` or `BuildsMethodCall` on builders and implements `FindBuilderImpl` automatically.

> **Note**
> Tips for source generators:
> - Consuming a generator in VS generally works nicely. VS will go to definitions in generated sources and you can also see the files in Solution Explorer, under References > Analyzers.
> - If you modify a generator your changes apply to Build immediately but as of the current version of VS 2022, IDE will continue to use the one it has loaded. The only way to refresh IDE after modifying a generator is to close and reload VS.
> - To help when working on a generator, you can use `<EmitCompilerGeneratedFiles>` in csproj to save files generated during build on disk (inside `obj`). I left this property, commented, in the csproj PR.

As generated code is not committed to git, I uploaded the source generated by this PR here for your convenience:
https://gist.github.com/jods4/6b145327f3c7563836c357984b723caf

### Code patterns
```csharp
// BuildsExpression is picked up by the source generator.
// It takes the list of ExpressionType this builder may handle
[BuildsExpression(ExpressionType.Lambda, ExpressionType.Constant)]
class MyBuilder : ISequenceGenerator
{
  // This static method will be called from generated code.
  // `expr` is guaranteed to be of the types specified in attributes.
  // Keep this short and efficient!
  public static bool CanBuild(Expression expr, BuildInfo info, ExpressionBuilder builder)
    => true;
}

// BuildsMethodCall is picked up by the source generator.
// It takes the list of method names that this builder may handle
[BuildsMethodCall("Take", "Skip")]
class MethodBuilder : ISequenceGenerator // or more commonly : MethodCallBuilder
{
  // This static method will be called from generated code.
  // `call` is guaranteed to be for a method whose name is amongst BuildsMethodCall arguments.
  // Keep this short and efficient!
  public static bool CanBuildMethod(MethodCallExpression call, BuildInfo info, ExpressionBuilder builder)
    => true;
}
```

### Special cases
A few builders may handle _any_ method call. You can do this simply by using `[BuildsExpression(ExpressionType.Call)]`, they will be probed after any method-specific builder `[BuildsMethodCall]`. 
Today this inculdes `MethodChainBuilder`, `QueryExtensionBuilder`, `TableBuilder`. Those builders tend to look for attributes on methods, e.g. `[Extension]` or `[Table]`.

Some builders take advantage of having different, simpler checks based on what method or expression type matched.
You can have multiple attributes on a single builder, and use the named property `CanBuildName` to direct the check to a different method for some cases. Many sync/async builders do that, for example `ContainsBuilder`:
```csharp
[BuildsMethodCall("Contains")]
[BuildsMethodCall("ContainsAsync", CanBuildName = nameof(CanBuildAsyncMethod))]
sealed class ContainsBuilder : MethodCallBuilder
{
	public static bool CanBuildMethod(MethodCallExpression call, BuildInfo info, ExpressionBuilder builder)
		=> call.IsQueryable() && call.Arguments.Count == 2;

	public static bool CanBuildAsyncMethod(MethodCallExpression call, BuildInfo info, ExpressionBuilder builder)
		=> call.IsAsyncExtension() && call.Arguments.Count == 3;
}
```

`ContextRefBuilder` is unique: it seems that it could potentially handle any expression if `BuilderInfo.Parent` is null. For this case I added `[BuildsAny]`. Builders with this attribute will be probed after everything else failed. 
The signature of the static method is `CanBuild(BuildInfo info, ExpressionBuilder builder)`.

### Risks and breaking changes
Everything is internal: there's no public API change, so no breaking change.

A lot of files were touched because there is 70+ builders but 95% of changes are very mechanical. I think there's still room for improvements but I wanted to keep this PR straightforward, as it's fairly large. 
If all tests pass, I'd say the risk of regression is low.


jods4:
@MaceWindu Tests look good.
There are 2 failures: 
- One is the currently known issue of `ORA-01000: maximum open cursors exceeded` during one of Oracle 19 tests.
- The other is SAP HANA test suite timing-out after one hour.

Neither of those is related to SQL query generation (this PR).

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -23,75 +24,22 @@ internal sealed partial class ExpressionBuilder : IExpressionEvaluator
...
-			new HasUniqueKeyBuilder        (),
-			new MultiInsertBuilder         (),
-			new TagQueryBuilder            (),
-			new EnumerableBuilder          (),
-			new QueryExtensionBuilder      (),
-			new QueryNameBuilder           (),
-		};
+			return FindBuilderImpl(info, this) is {} builder
+				? builder 
+				: throw new LinqException("Sequence '{0}' cannot be converted to SQL.", info.Expression);
```
sdanylivsoftserve, line 37: It should not throw. New parser tries other ways if it is not sequence. Or name of this method should be changed.

jods4: Look higher in the code, it is hidden by 60 lines of removed builders but there are two methods:
- `ISequenceBuilder FindBuilder()`, which returns a builder or throws;
- `bool TryFindBuilder(out ISequenceBuilder)` which returns a boolean indicating if a matching builder was found or not.

The new code uses these two methods based on whether the old code was throwing or not.
Behaviour should be unchanged (and all tests are passing).

Source/LinqToDB/Linq/Builder/EnumerableBuilder.cs
```diff
@@ -35,32 +30,23 @@ public bool CanBuild(ExpressionBuilder builder, BuildInfo buildInfo)
...
 			static bool CanBuildMemberChain(Expression? expr)
 			{
-				if (expr == null)
-					return true;
-
-				if (expr.NodeType == ExpressionType.MemberAccess)
-					return CanBuildMemberChain(((MemberExpression)expr).Expression);
-
-				return expr.NodeType == ExpressionType.Constant;
+				while (expr is { NodeType: ExpressionType.MemberAccess })
```
sdanylivsoftserve, line 45: Do you think we do not need recursion here?

jods4: The function performs exactly the same operation before/after my change.
A better name for the function (I was hesitant to rename) is: `IsConstantMemberExpression(expr)`.
An expr is a CME if (1) it's a ConstantExpression; or (2) it's a MemberExpression off of a CME.

The code before was evaluating exactly this recursive definition.
The code after is an equivalent loop-based rewrite: walking up the MemberExpression chain until it ends (on a ConstantExpression?).

If you're lucky and JIT tail call optimizations kick in, the two perform the same but that's unreliable.
Otherwise the tight loop uses constant stack space (vs O(N)) and performs much less instructions than recursion.

I also find easier to understand the high-level purpose when reading the loop (unwrap MemberExpression and look if it started on a constant) than figuring out the recursion, but renaming would help too.

Note that I kept the `expr == null` check but I have no idea how the base expression of `MemberExpression` could be `null`??? If you know I think it's worth a comment.

jods4: Well I did one second of research in .net source code and `MemberExpression.Expression` is explicitly marked nullable, so that's a case for sure. I believe that covers static member access?

viceroypenguin: Yes, Expression is `null` in the case of static properties and fields.

sdanyliv (approved):
I like it.

viceroypenguin (changes requested):
Am working on rewriting source generator according to correct patterns.

Source/CodeGenerators/CodeGenerators.csproj
```diff
@@ -0,0 +1,17 @@
...
+	<IncludeBuildOutput>false</IncludeBuildOutput>
+    <Nullable>enable</Nullable>
+	<LangVersion>latest</LangVersion>
+	<ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally> 
+	<EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.10.0" PrivateAssets="all" />
```
MaceWindu, line 14: What's wrong with it using CPM?

Source/LinqToDB/Linq/Builder/ContextParser.cs
```diff
@@ -5,15 +5,11 @@ namespace LinqToDB.Linq.Builder
 {
 	using SqlProvider;
 
+	[BuildsMethodCall("GetContext")]
```
MaceWindu, line 8: @sdanyliv do we need it? I see it used by `LinqExtensions.GetContext` but that method never called

MaceWindu: I propose to remove both contextparser and method with field

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -24,75 +25,22 @@ internal sealed partial class ExpressionBuilder : IExpressionEvaluator
...
```
MaceWindu, line 38: Wrap expression into new expression pretty-print helper `SqlErrorExpression.PrepareExpressionString(info.Expression)`?

MaceWindu: It was lost on merge. See line 409  below

MaceWindu:
Wonder what changed so it affected generation of couple of aliases in baselines. In any case it doesn't look worrying

Source/CodeGenerators/CodeGenerators.csproj
jods4: To be honest I don't fully remember, it was many months ago I first wrote this.
As this project is entirely internal and standalone, I believe I didn't want to tie it to the other deliverables.

jods4:
@MaceWindu 
> Wonder what changed so it affected generation of couple of aliases in baselines. In any case it doesn't look worrying

That's weird indeed, it shouldn't have had any impact on generated SQL, except if multiple builders were competing for the same expression.

More likely that something was missed during the rebase :(

viceroypenguin:
@jods4 If you want to look, see here: https://github.com/linq2db/linq2db.baselines/pull/1240/files. Probably not worth researching too deeply, because the SQL is functionally equivalent.

Source/LinqToDB/Linq/Builder/TableBuilder.cs
```diff
@@ -10,9 +10,11 @@ namespace LinqToDB.Linq.Builder
 	using Mapping;
 	using Reflection;
 
+	[BuildsExpression(ExpressionType.Call, ExpressionType.Constant, ExpressionType.MemberAccess)]
```
MaceWindu, line 13: Why we have such filter?

I see `FindBuildContext` works only with `ExpressionType.Call`

MaceWindu: Also can we shortcircuit known supported methods to this builder?

jods4:
> @jods4 If you want to look, see here: https://github.com/linq2db/linq2db.baselines/pull/1240/files. Probably not worth researching too deeply, because the SQL is functionally equivalent.

Ok, well there were some pretty substantial changes in `AggregationBuilder` that I had to merge... given that all these differences seem to come from aggregates I'm gonna guess this is were the difference comes from. If it's working fine I'm not gonna dig deeper.

Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
```diff
@@ -132,42 +132,11 @@ static AggregationType GetAggregationType(MethodCallExpression methodCallExpress
...
-			var definition = GetAggregateDefinition(methodCall, builder.MappingSchema);
-
-			if (definition != null)
-			{
-				if (methodCall.Arguments.Count > 0)
-				{
-					if (builder.IsSequence(new BuildInfo(buildInfo, methodCall.Arguments[0])))
-						return true;
-				}
-			}
```
MaceWindu, line 149: I see this whole branch with related code removed. @sdanyliv was it needed as there is no changes to tests except couple of aliases?

Source/LinqToDB/Linq/Builder/EnumerableBuilder.cs
```diff
@@ -10,20 +10,15 @@ namespace LinqToDB.Linq.Builder
...
+		public static bool CanBuild(Expression expr, BuildInfo info, ExpressionBuilder builder)
 		{
-			var expr = buildInfo.Expression;
-
 			if (expr.NodeType == ExpressionType.NewArrayInit)
-			{
 				return true;
-			}
 
 			if (expr.NodeType == ExpressionType.Call)
```
MaceWindu, line 23: Looks like whole if with _containsMethodInfos could be removed?

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -183,17 +128,6 @@ public Query<T> Build<T>()
 
 			var sequence = BuildSequence(new BuildInfo((IBuildContext?)null, Expression, new SelectQuery()));
 
-			if (_reorder)
-			{
-				using var mr = ActivityService.Start(ActivityID.ReorderBuilders);
```
MaceWindu, line 188: Looks like we can remove `ActivityID.ReorderBuilders` enum value

Source/LinqToDB/Linq/Builder/TableBuilder.cs
jods4: Why `Constant` and `MemberAccess`? Probably they were handled by this builder when I first wrote this code, and v6 has removed them. I agree that the rebased version can only handle `Call` so we should remove those attributes.

Know methods: not completely, because at the end of `FindBuildContext`, any method that has `TableFunctionAttribute` applied is handled by this.
We _could_ have the `BuildsMethod` attribute with the known ones + a quick `CanBuildKnown() => true`, combined with a `BiuldExpression(Call)` and an alternative `CanBuildCall() => CheckAttribute()`. Not sure if it's worth the effort, but doable. The largest benefit IMHO is that when looking at the generated code we would see the known method names.

MaceWindu: 1. That's what i suspected
2. I would say if we can do it easily, why not do it

Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
jods4: @MaceWindu I realised that `GetAggregateDefinition` always returns `null`, here's the code of the now-removed `GetAggregateDefinition()`: 
```csharp
public static Sql.ExpressionAttribute? GetAggregateDefinition(MethodCallExpression methodCall, MappingSchema mapping)
{
  var function = methodCall.Method.GetExpressionAttribute(mapping);
  return function != null  && function is not Sql.ExtensionAttribute && (function.IsAggregate || function.IsWindowFunction) 
    ? function 
    : null;
}
```

Observe that `GetExpressionAttribute` returns `Sql.ExtensionAttribute?`, so `function != null && function is not Sql.ExtensionAttribute` is always false (what else could it be?), hence the returned value is always `null`.

I proceeded to remove that method, and all dead code that called into it.

Was it a bug? I don't think so, just dead code.
Methods having `ExtensionAttribute` with `IsAggregate` or `IsWindowFunction` are handled by `MethodChainBuilder`, look here: https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/Builder/MethodChainBuilder.cs#L16-L22

Source/LinqToDB/Linq/Builder/EnumerableBuilder.cs
jods4: Indeed that must be a changed from v6 branch. 
The current code of `CanBuild()` can never be true for a `Call` expression, in fact I did not even put `Call` in the builder attributes. So that branch is useless and should be removed.

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
jods4: Indeed!

Source/LinqToDB/Linq/Builder/AggregationBuilder.cs
MaceWindu: Yeah, we discussed it with @sdanyliv - it was duplicate code, so it is fine to remove it

MaceWindu (approved):

PR #4547, opened 2024-06-24
Null comparison changes, rebased #3537

### LINKED ISSUE #3535: ExprExpr is a mess under CompareNullsAsValues

jods4:
Fix #3535

See #3537 for all details, this is a rebase on v6 `master`.
The rebase was rather messy due to the many changes that have been made in v6 since 2022, let's see how the tests fail.

Full discussion can be found in #3537, here's the quick summary of this PR contents:

**Tests and fix for #3535**, where some ExprExpr reductions are incorrect when comparing nulls as values.

**Change of behavior for `compareNullsAsValues = false`:** comparisons to a non-constant `null`, for example a local variable, should not be converted into `is null` predicates.
To avoid breaking consumers, a new option is introduced:

- New option `CompareNulls` is an enum:
  - `LikeCSharp` works like `true` in v5.
  - `LikeSql` is the new mode, where comparisons are never modified (SQL semantics throughout); literal `== null` is the only exception that intuitively maps to `is null`.
  - `LikeSqlExceptParameters` behaves like `false` in v5, for compatibility. It's mostly `LikeSql` but comparison to client-evaluated constants that are null are converted into `is null`.
- Option `CompareNullsAsValues` remains to ease migration but is marked Obsolete. It maps to the new options `LikeCSharp` and `LikeSqlExceptParameters`.


Source/LinqToDB/Common/Configuration.cs
```diff
@@ -283,9 +283,14 @@ public static bool OptimizeJoins
 			}
 
 			/// <summary>
-			/// If set to true nullable fields would be checked for IS NULL in Equal/NotEqual comparisons.
+			/// If set to LikeCSharp nullable fields would be checked for IS NULL in Equal/NotEqual comparisons.
```
viceroypenguin, line 286:
```suggestion
			/// If set to <see cref="CompareNulls.LikeCSharp" /> nullable fields would be checked for IS NULL in Equal/NotEqual comparisons.
```
(and similar for other references to values from the enum)

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -283,9 +283,14 @@ public static bool OptimizeJoins
...
+			/// If set to LikeSql comparisons are compiled straight to equivalent SQL operators, which consider
+			/// nulls values as not equal.
+			/// LikeSqlExceptParameters is a backward compatible option that works mostly as LikeSql, 
+			/// but sniffs parameters values and changes = into IS NULL when parameters are null.
```
viceroypenguin, line 290: Use `<c></c>` to mark code blocks, like `<c>IS NULL</c>`

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -283,9 +283,14 @@ public static bool OptimizeJoins
...
+			/// Comparisons to literal null are always compiled into IS NULL.
 			/// This affects: Equal, NotEqual, Not Contains
-			/// Default value: <c>true</c>.
+			/// Default value: <c>LikeCSharp</c>.
```
viceroypenguin, line 293: Use `<see />` instead of `<c>` here

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -315,15 +320,22 @@ public static bool OptimizeJoins
...
+				set 
 				{
-					if (Options.CompareNullsAsValues != value)
-						Options = Options with { CompareNullsAsValues = value };
+					if (Options.CompareNulls != value)
+						Options = Options with { CompareNulls = value };
 				}
 			}
+			
+			public static bool CompareNullsAsValues
```
viceroypenguin, line 333: Add `[EditorBrowsable(EditorBrowsableState.None)]` (or `Advanced`, but I think `None` is better for `[Obsolete]` properties) to hide accidental use.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -315,15 +320,22 @@ public static bool OptimizeJoins
...
-						Options = Options with { CompareNullsAsValues = value };
+					if (Options.CompareNulls != value)
+						Options = Options with { CompareNulls = value };
 				}
 			}
+			
+			public static bool CompareNullsAsValues 
+			{
+				get => CompareNulls == CompareNulls.LikeCSharp;
+				[Obsolete("Use CompareNulls instead: true maps to LikeCSharp and false to LikeSqlExceptParameters")]
```
viceroypenguin, line 336: Why is this on the `set` and not the property as a whole?

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -2262,7 +2263,7 @@ Expression GenerateConstructorComparison(SqlGenericConstructorExpression leftCon
 						if (trueValue.ElementType  == QueryElementType.SqlValue &&
 						    falseValue.ElementType == QueryElementType.SqlValue)
 						{
-							var withNullValue = compareNullsAsValues
+							var withNullValue = compareNulls == CompareNulls.LikeCSharp
```
viceroypenguin, line 2266:
```suggestion
							var withNullValue = compareNulls is CompareNulls.LikeCSharp
```
(and others)

Source/LinqToDB/Linq/Translation/ProviderMemberTranslatorDefault.cs
```diff
@@ -226,7 +226,11 @@ protected bool ProcessGetValueOrDefault(ITranslationContext translationContext,
 		protected virtual ISqlExpression? TranslateConvertToBoolean(ITranslationContext translationContext, ISqlExpression sqlExpression, TranslationFlags translationFlags)
 		{
 			var sc = new SqlSearchCondition();
-			var predicate = new SqlPredicate.ExprExpr(sqlExpression, SqlPredicate.Operator.Equal, new SqlValue(0), translationContext.DataOptions.LinqOptions.CompareNullsAsValues)
+			var predicate = new SqlPredicate.ExprExpr(
+				sqlExpression,
+				SqlPredicate.Operator.Equal, 
+				new SqlValue(0),
+				translationContext.DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeCSharp ? true : null)
```
viceroypenguin, line 233:
```suggestion
			var predicate = new SqlPredicate
				.ExprExpr(
					sqlExpression,
					SqlPredicate.Operator.Equal, 
					new SqlValue(0),
					translationContext.DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeCSharp ? true : null)
```

Source/LinqToDB/SqlQuery/SqlPredicate.cs
```diff
@@ -321,39 +321,49 @@ ISqlPredicate MakeWithoutNulls()
...
+							return search;
+						}
+						default:
+							return this;
 					}
-					else
+				}
+				else
+				{
+					if (Operator is Operator.Equal)
```
viceroypenguin, line 357: should this be converted to a `switch`

sdanyliv: Please do not convert everything to switch. It is not converted specially to be more clear and debuggable.

viceroypenguin: I'll challenge you this case though.

```cs
if (Op is Equal)
{
}
else if (Op is NotEqual)
{
}
else
{
}
```
vs 
```cs
switch (op)
{
  case Equal: break;
  case NotEqual: break;
  default: break;
}
```
should have the same debuggability, and be _more_ clear.

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -30,27 +31,25 @@ public override SqlStatement TransformStatement(SqlStatement statement, DataOpti
...
-					//
-					if ((expr.Operator == SqlPredicate.Operator.Equal          ||
-						 expr.Operator == SqlPredicate.Operator.NotEqual       ||
-						 expr.Operator == SqlPredicate.Operator.GreaterOrEqual ||
-						 expr.Operator == SqlPredicate.Operator.LessOrEqual) && expr.WithNull == true)
+					var (a, op, b, withNull) = (SqlPredicate.ExprExpr)element;
+
+					// See ConvertPredicateImpl, we transform empty strings "" into null-handling expressions
+					if (withNull != null ||
+						(dataOptions.LinqOptions.CompareNulls != CompareNulls.LikeSql &&
```
igor-tkachev, line 47: It looks like you removed some optimizations.

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -137,9 +138,9 @@ public static LinqOptions WithOptimizeJoins(this LinqOptions options, bool optim
 		/// </code>
 		/// </example>
 		[Pure]
```
igor-tkachev, line 140: Do not remove it. Make it `Obsolete`.

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -359,9 +360,9 @@ public static DataOptions UseOptimizeJoins(this DataOptions options, bool optimi
...
-		public static DataOptions UseCompareNullsAsValues(this DataOptions options, bool compareNullsAsValues)
+		public static DataOptions UseCompareNulls(this DataOptions options, CompareNulls compareNulls)
 		{
```
igor-tkachev, line 364: Do not remove it. Make it `Obsolete`.

igor-tkachev (changes requested):

MaceWindu:
@jods4 status?

jods4:
@MaceWindu sorry, recently I've not had a great time at work, so I don't feel like coding when I come home.
I'll try to take a look this week so that we can finalize and merge.

sdanyliv:
`LikeCSharp` ? Why not `LikeCLR`?

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
jods4: @viceroypenguin as much as I like C# pattern matching, it doesn't make any difference here, does it? Same number of chars in source, same compiled bytecode?

Source/LinqToDB/Common/Configuration.cs
jods4: @viceroypenguin This was at the very beginning and my thought was that I wanted to discourage users to set this option, but still be able to read it in our code. Then I noticed with the new options we mustn't read the global config anymore anyway, so you're right the whole thing is deprecated.

Source/LinqToDB/Common/Configuration.cs
jods4: I thought this would be implied by `[Obsolete]` slightly disappointed by MS that it's not!

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
jods4: Good catch, I thought the new Options were added to 6.0 but they already shipped in 5.x.

I'm putting back an [Obsolete] WithCompareNullsAsValues.

jods4:
@sdanyliv Sure, let me rename.

Source/LinqToDB/DataProvider/Oracle/Oracle11SqlOptimizer.cs
jods4: It has changed, yes. In commit 2840efac I fixed Oracle empty strings handling (it was buggy).

This method indicates whether compilation is dependent on actual parameter values.
The transformation is made in OracleSqlExpressionConvertVisitor.ConvertExprExprPredicate.
So basically the condition here must be the same, and it is.

You can check if the new transformation looks correct to you in OracleSqlExpressionConvertVisitor.cs.
The explanation why this expression is what it is can be found in comments on lines 39-43.
It's not intuitive but after reviewing one more time it looks ok to me.

Source/LinqToDB/Common/CompareNulls.cs
```diff
@@ -0,0 +1,14 @@
+namespace LinqToDB.Common
+{
+	/// <summary>Defines how null values are compared in compiled queries.</summary>
```
MaceWindu, line 3: I would prefer to avoid use of `compiled query` as it means something else (Compiled Query API). Better to use `generated SQL`

Source/LinqToDB/Common/CompareNulls.cs
```diff
@@ -0,0 +1,14 @@
...
+	public enum CompareNulls
+	{
+		/// <summary>CLR semantics: nulls values are equal.</summary>
+		/// <note>This option may add significant complexity in generated SQL, possibly preventing index usage.</note>
+		LikeClr,
+		/// <summary>SQL semantics: nulls values are unknown.</summary>
+		LikeSql,
+		/// <summary>SQL semantics, except null parameters are treated like null constants (they compile to "IS NULL").</summary>
+		LikeSqlExceptParameters,
```
MaceWindu, line 12: maybe add c#+SQL examples to comments. Now they expect from users to have good understanding what they do 😄 . Also examples are always easier to understand

MaceWindu (changes requested):
baselines:

- OrmBattleTests.DateTimeTest : not sure how to feel about this change, looks suspicious
- Issue3148Tests.TestDefaultExpression_10 : doesn't look right @sdanyliv ?

Source/LinqToDB/Common/CompareNulls.cs
jods4: I have copied the example from configuration API + added a little bit more notes in comments.

jods4:
Linux Oracle 19 is unrelated, flaky failure `ORA-01000: maximum open cursors exceeded`

Source/LinqToDB/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -22,75 +23,83 @@ public OracleSqlExpressionConvertVisitor(bool allowModify) : base(allowModify)
...
-					}
+					return CompareToEmptyString(a, op);
 				}
-
-				if (expr.Expr2.SystemType == typeof(string)                             &&
-				    expr.Expr2.TryEvaluateExpression(EvaluationContext, out var value2) && value2 is string string2)
+				
+				if (a.SystemType == typeof(string)                             &&
+				    a.TryEvaluateExpression(EvaluationContext, out var aValue) && 
+					aValue is string { Length: 0 })
```
viceroypenguin, line 58: why not 
```suggestion
					aValue is "")
```

jods4: I think it's a mechanical rewriting of previous code.

Old code did it in two steps, first `if (value2 is string string2)` then later `if (string2.Length == 0)`.
It just got mindlessly merged into a single `is string { Length: 0 }` check.

Replacing with `aValue is ""` would be exactly the same. This actually occurs 4 times in this file, twice here and twice on lines 30-37.

Fun fact I was looking at the JIT disasembly and although the logic is exactly the same, the quality of `is ""` is lower. JIT actually tests if the pointer is null twice for some reason:
https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjgPYBGAVgKYDGwA/NgJ4Dcp5PZW2apUoAbbAFkkACkIAPAJTE+5fgHZssgHTtsASwjYARIe5oVAX2XKBQ0RIQyNikmZXk46rTv24kABmJsABl6ADsAc2AACxBsAPNTC3RzIA==

As a result asm code for `is ""` is 9 bytes longer than `is { Length: 0 }`.
I doesn't matter though, we can change if you want. If I had written this code from scratch, I'd probably have written `is ""`, it's more natural.

viceroypenguin: Shorter ASM is good enough reason for me to support `is { Length: 0 }`. Leave it, imo.

jods4: @viceroypenguin for what it's worth, MS seems to have fixed that redundant null check in .net 9.0
https://github.com/dotnet/runtime/issues/105191

MaceWindu:
@jods4 , could you please add this regression test to oracletests.cs to see if it is fixed (require some changes to `UseCompareNullsAsValues`)

```cs
#region Issue 4163
[Test]
public void Issue4163Test([IncludeDataSources(true, TestProvName.AllOracle)] string context, [Values] bool compareNullsAsValues)
{
	using var db = GetDataContext(context, o => o.UseCompareNullsAsValues(compareNullsAsValues));
	using var tb = db.CreateLocalTable(Issue4163TableForCreate.Data);

	var cnt = db.GetTable<Issue4163Table>().Where(r => r.Method != PaymentMethod.Unknown).Count();

	Assert.That(cnt, Is.EqualTo(2));
}

[Table("Issue4163Table")]
sealed class Issue4163TableForCreate
{
	[Column(CanBeNull = true)] public PaymentMethod Method { get; set; }

	public static readonly Issue4163TableForCreate[] Data = new[]
	{
		new Issue4163TableForCreate() { Id = 1, Method = PaymentMethod.Unknown },
		new Issue4163TableForCreate() { Id = 2, Method = PaymentMethod.Cheque },
		new Issue4163TableForCreate() { Id = 3, Method = PaymentMethod.EFT },
	};
}

[Table("Issue4163Table")]
sealed class Issue4163Table
{
	[Column(CanBeNull = false)] public PaymentMethod Method { get; set; }

	public static readonly Issue4163Table[] Data = new[]
	{
		new Issue4163Table() { Id = 1, Method = PaymentMethod.Unknown },
		new Issue4163Table() { Id = 2, Method = PaymentMethod.Cheque },
		new Issue4163Table() { Id = 3, Method = PaymentMethod.EFT },
	};
}

enum PaymentMethod
{
	[MapValue("")]
	Unknown,

	[MapValue("C")]
	Cheque,

	[MapValue("E")]
	EFT
}
#endregion
```

jods4:
@MaceWindu Here you go.

/azp run test-all

MaceWindu:
ok (not really), test fails...

MaceWindu:
```sql
    SELECT
	COUNT(*)
FROM
	"Issue4163Table" r
```

optimizer removed condition, because it was reduced to `true`

jods4:
@MaceWindu The logic is simple but the resolution might not be.

The field is declared as a non-nullable enum.

1. The source condition is `Method != Enum.Unknown`
2. Mapping to strings turns into `Method != ""`
3. Oracle empty strings turns into `Method IS NOT NULL`
4. Optimizer sees "non-nullable field is not null" and simplifies to `true`

Mapped enums here are a very special case, but it's fair to assume that not all Oracle users always mark their string fields nullable, so that bug could easily surface in other situations.

One possibily solution could be for Oracle provider to always consider any field of type string as intrinsically nullable. This isn't strictly true as you can declare a `varchar` column as `not null` and that simply implies you can't store the empty string in it. But it's better to produce slightly less optimized SQL that works rather than SQL that does not.

MaceWindu:
Will try to change tests first

MaceWindu:
Ok, I've updated tests to not force not-null and now there are two issues.

1. `LikeClr` test case doesn't work with remote context

expected
```sql
SELECT
   	COUNT(*)
FROM
   	"Issue4163Table" r
WHERE
   	r."Method" IS NOT NULL
```
but was
```sql
SELECT
   	COUNT(*)
FROM
   	"Issue4163Table" r
WHERE
   	(r."Method" IS NOT NULL OR r."Method" IS NULL)
```

something lost during ast de/serialization probably

2. We infer nullability from enum map values, but it doesn't handle empty string as null. Will check if we can configure it easily for oracle

MaceWindu:
No idea for now what to do with test failing over remote context. We use different query optimization approach for it and it doesn't work well in this case. Maybe @sdanyliv will come with some solution later. I will disable test for now

MaceWindu (approved):

PR #4553, opened 2024-07-01
Regression tests for existing issues (part 2)

### LINKED ISSUE #3311: Bad query: JOIN for correlated subquery instead of APPLY
### Steps to reproduce
Consider this simple query:
```c#
class Src 
{
  public int N { get; set; }
}

var query = 
  from x in db.GetTable<Src>()
  from apply in db.SelectQuery(() => Sql.AsSql(x.N + 1))
  select apply;
```

It generates the following invalid SQL (this example for Oracle but the problem is likely not Oracle-specific):
```sql
SELECT
        t1."c1"
FROM
        "Src" x,
        (
                SELECT
                        x.N + 1 as "c1"
                FROM SYS.DUAL
        ) t1
```
It created a CROSS APPLY (with comma `,` syntax) between `x` and `t1` but notice that the subquery is correlated to `x` because it references `x.N`.

Joins must be uncorrelated. 
The correct SQL for a correlated sub-query is an APPLY (in this case CROSS APPLY).

### Workaround
You can get a correct query, albeit an `OUTER APPLY` instead of `CROSS APPLY` simply by adding `DefaultIfEmpty()`:
```csharp
var query = 
  from x in db.GetTable<Src>()
  from apply in db.SelectQuery(() => Sql.AsSql(x.N + 1)).DefaultIfEmpty()
  select apply;
```
Working SQL:
```sql
SELECT
        t1."c1"
FROM
        "Src" x
                OUTER APPLY (
                        SELECT
                                x.N + 1 as "c1"
                        FROM SYS.DUAL
                ) t1
```

### Environment details
linq2db version: **master** branch
Database Server: **Oracle**
Database Provider: **ODP.NET**
Operating system: **Win 10**
.NET Framework: **.net 5**

### LINKED ISSUE #3665: Enh: support expressions that evaluates to IEnumerable in FROM
This works:
```csharp
var values = Enum.GetValues<MyEnum>();

var query = 
  from x in db.MyTable
  from y in values
  select x.A + y;
``` 

Although `Enum.GetValues<MyEnum>()` can be evaluated server-side, this query can't be translated:
```csharp
var query = 
  from x in db.MyTable
  from y in Enum.GetValues<MyEnum>()
  select x.A + y;
```
It says _"Sequence 'GetValues()' cannot be converted to SQL."_

MaceWindu:
Tests to check for possible regressions/issues:

- [x] Issue3365Test : two providers generate unsupported SQL (should be disabled or fixed?), one (Sybase) - generates parser error
- [x] Issue3822Test : access and sybase generate `ArgumentException : must be reducible node`
- [x] Issue3226Test1/3/4 : some dbs fail due to unsupported SQL (disable?), or due to parser error (ClickHouse)
- [x] Issue3295Test1 : sybase - parser error, oracle - typing error (mapping issue?)
- [x] Issue3311Test2 - optimization of query doesn't work for SQL CE and ClickHouse
- [x] Issue3311Test3 - generates invalid sql except sap hana and firebird4+
- [x] ~~Issue3372Test1 - potential optimization? unnecessary column generated~~
- [x] Issue3665Test - Issue3665Test1 worked in v5, now doesn't work (regression). 2/4 still doesn't work

Fix #3311
Fix #3665


Tests/Linq/Linq/EagerLoadingTests.cs
```diff
@@ -1920,6 +1920,94 @@ public void Issue4060([DataSources] string context)
...
+			t1
+				.OrderBy(x => x.Values.Sum(y => y.Value))
+				.Select(x => new {
+					Id = x.Id,
+					Text = x.Text,
+					Summary = x.Values.Select(y => new { Total = y.Value }).AsEnumerable()
+				})
+				.ToList();
+		}
+
```
MaceWindu, line 1991: ActiveIssue should be:

- removed if those providers work
- disabled if providers not supported
- or issue detail failed if they fail still for another fixable reason

MaceWindu (commented):
baselines regressions:

- Issue1455Test1/Issue1455Test2 test baselines introduced unnecessary aggragate query to subqueries. Would be nice to fix it

MaceWindu:
Baselines regression fix will be separate PR, I will review remainig `ActiveIssue`s soon

Source/LinqToDB/Linq/Builder/ExpressionBuilder.SqlBuilder.cs
```diff
@@ -185,7 +185,7 @@ public bool IsSupportedSubquery(IBuildContext parent, IBuildContext context, out
 				SqlJoinedTable? fakeJoin = null;
 
 				// add fake join there is no still reference
-				if (null == clonedParentContext.SelectQuery.Find(e => e is SelectQuery sc && sc == clonedContext.SelectQuery))
+				if (clonedParentContext.SelectQuery.From.Tables.Count > 0 && null == clonedParentContext.SelectQuery.Find(e => e is SelectQuery sc && sc == clonedContext.SelectQuery))
```
viceroypenguin, line 188: Awfully long - can we break this up over multiple lines?
```suggestion
				if (clonedParentContext.SelectQuery.From.Tables.Count > 0 
					&& clonedParentContext.SelectQuery.Find(e => e is SelectQuery sc && sc == clonedContext.SelectQuery) == null)
```

Source/LinqToDB/Linq/Builder/ExpressionBuilder.cs
```diff
@@ -187,7 +187,7 @@ public Query<T> Build<T>()
 
 					if (queryInfo.Statement.SelectQuery != null)
 					{
-						if (!SqlProviderHelper.IsValidQuery(queryInfo.Statement.SelectQuery, null, null, false, DataContext.SqlProviderFlags, out var errorMessage))
+						if (!SqlProviderHelper.IsValidQuery(queryInfo.Statement.SelectQuery, null, null, null, DataContext.SqlProviderFlags, out var errorMessage))
```
viceroypenguin, line 190: can we add parameter names to the three `null` parameters?

viceroypenguin (approved):

PR #4554, opened 2024-07-01
Refactored DefaultIfEmptyContext usage.

sdanyliv:


Source/LinqToDB/Linq/Builder/AllJoinsBuilder.cs
```diff
@@ -74,10 +72,12 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
 				}*/
 
 				result.SetAlias(condition.Parameters[0].Name);
-				return BuildSequenceResult.FromContext(result);
 			}
 
-			return BuildSequenceResult.FromContext(sequence);
+			if (joinType == JoinType.Left || joinType == JoinType.Full)
```
viceroypenguin, line 77:
```suggestion
			if (joinType is JoinType.Left or JoinType.Full)
```

Source/LinqToDB/Linq/Builder/AllJoinsLinqBuilder.cs
```diff
@@ -66,11 +66,11 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
 			}
 
 			if (joinType == JoinType.Right || joinType == JoinType.Full)
-				outerContext = new DefaultIfEmptyBuilder.DefaultIfEmptyContext(buildInfo.Parent, outerContext, outerContext, null, false);
+				outerContext = new DefaultIfEmptyBuilder.DefaultIfEmptyContext(buildInfo.Parent, outerContext, outerContext, null, false, false);
```
viceroypenguin, line 69: with multiple true/false values, my preference would be to break the parameters up over multiple lines and `name: ` the null, true, false values the other places this function is called.

PR #4569, opened 2024-07-15
Regression tests for existing issues (part 3) 

MaceWindu:
Next batch of issue tests

Tests with issues:

- Issue4253Test : regression, now linq2db fails to parse query with `NotImplementedException`
- SelectCompositePropertyMapped_Class_OnlyRequredColumns, SelectCompositePropertyMapped_Struct_OnlyRequredColumns : not a regression, but would be nice to fix
- Issue4520Test : regressed in preview.1. Currently fails for Firebird 2.5, Informix, Oracle, SAP HANA, SQL CE and SQL SERVER
- Issue4347Test2, Issue4347Test3 : regression since 5.4.1, Issue4347Test1 - original issue, also fails (all 3 fail in parser)

More bad tests:

- Issue2815Test: regression in v6 (exception). feature itself is not implemented yet and generate less effective sql
- Issue3631Test3: looks like test now works, but not for all providers. Following providers generate invalid SQL: sybase, sqlserver, sqlce, access. Probably some could be fixed
- Issue2494Test1/Issue2494Test2 : tests are strange, but they were fixed in preview.1 and broken now
- Issue4508Test : fixed in preview.1, now regressed in master
- Issue3782Test* : review required wether some/all failing tests should work or not
- Issue4012Test: regression in master, worked in v5 and v6 preview.1

and more regresssions from last batch of tests:
- class:Issue3340Tests.Test : regression from v5
- Issue3807Test14, Issue3807Test1 : regression from v5 (other Issue3807Test* tests didn't worked before)
- Issue3250Test1/Issue3250Test2 : regression from v5
- Issue4220Test: not regression as it wasn't worked in v5, but now it fails for informix (only) with `InvalidCastException : Unable to cast object of type 'ConcreteA' to type 'ConcreteB'.`
- Issue4222Test1/Issue4222Test2 : regression from v5
- Issue3560Test1/Issue3560Test3 : regression from v5 (Issue3560Test4 didn't worked in v5)
- Issue1855Test : regression from v5 for mysql and oracle, sqlite.ms (classic didn't worked)
- Issue4588Test : need to check if access and sybase failures to generate sql are valid

another set of regressions:
- Issue4607Test
- Issue4613Test
- Issue4596Test
- Issue4654Tests.Test

Included fixes to help new tests pass:

- ~fix NRE in transaction disposal in new tracing tags feature~ moved to separate PR
- enabled RECURSIVE keyword for CLickHouse CTE (available since 24.4 release) and enabled CH in some tests with recursive CTE


MaceWindu:
moved to https://github.com/linq2db/linq2db/pull/4676

PR #4577, opened 2024-07-17
Improve Task.Run()/CancellationToken code

viceroypenguin:
Part of #4475


MaceWindu (changes requested):
I see code still assume Provider could be not IQueryProviderAsync. Remove it here or you plan to submit another PR?

MaceWindu:
If there is no `IQueryProviderAsync`, then it is not linq2db query provider and calling it with Linq2db API will result in error from provider, in this case `System.Linq.EnumerableExecutor`:
```
System.InvalidOperationException: 'There is no method 'Delete' on type 'LinqToDB.LinqExtensions' that matches the specified arguments'
```

Instead of trying to execute broken code we can throw exception with descriptive message that linq2db API used over non-linq2db query.

PS: if there are valid cases I'm not aware of - we could preserve it

viceroypenguin:
@MaceWindu fair enough - let's do a sep PR for that.

MaceWindu (approved):

PR #4591, opened 2024-07-30
Fix latest VS compatibility

MaceWindu:


jods4:
@MaceWindu wouldn't it be better to remove the rule that requires xml doc comments on public API for the code generator project, rather than adding empty comments on everything?

MaceWindu:
https://github.com/dotnet/roslyn/issues/41640 😄

jods4:
😮 that looks like a mess!

MaceWindu:
Anyway, need to submit another pr, looks like roslyn cannot locate generators from artifacts folder now...

Shane32:
Personally, I would not insert empty xml comments just to remove unnecessary usings.  Unnecessary usings does not hurt any code.  Empty XML comments does.  Either (a) when you want to require XML comments, they are not flagged as errors, or (b) they mask auto-inherited comments.  Further, I'm not sure why there is any valid reason to add a public member without an XML comment.  It only takes a few seconds to write an applicable XML comment, and ChatGPT can draft code comments for entire classes extremely quickly.  Besides documentation being lacking for linq2db in general, one of the biggest issues with the linq2db codebase was the lack of commenting the internal workings.  Without comments, usually effective at the method level, it becomes very difficult to understand the purpose and logic of the code flow, severely limiting my and other contributors' ability to provide bug fixes.  I've not been involved much recently with v6, so hopefully we have much more comments now.  I've spent extensive time with other open-source projects to write XML comments for all public API members (and often internal members), but I have not had the spare time to try to tackle linq2db.  Just my two cents.

PR #4594, opened 2024-08-01
Version 6 regressions

MaceWindu:
- `ManyAssociationEmptyCheck(1|2)`: `entity.Items =/!= null` condition fails to convert to SQL
- `IssueCteDuplicateColumn` : bad CTE definition `WITH(c1, c2) AS (select c1, c2, c2)`


MaceWindu:
Moved to #4676

PR #4595, opened 2024-08-03
Move linq2db.EntityFrameworkCore to this repo

MaceWindu:
- [x] move source code
- [x] move tests
- [x] migrate nuget generation code
- [x] move test connection configuration to config files
- [x] add test baselines generation
- [x] integrate with CI: build
- [x] integrate with CI: nugets build and publish
- [x] integrate with CI: testing
- [x] migrate to current linq2db version

Moved to separate tasks:

- [ ] review supported providers list
- [ ] add efcore 9 preview support
- [ ] import issues from old repo with regression tests
- [ ] fix regressions: TestCreateTempTable, TestChangeTrackerTemporaryTable : `SqlException : Table not found for 't7[Id:3].IsDeleted => t8.IsDeleted'`
- [ ] fix regressions: TestFunctions : debug bugcheck in placeholder handling
- [ ] bug (regression?) : TestDeleteFrom test generates bad SQL for SQLite and MySQL


Source/LinqToDB.EntityFrameworkCore/Compat/Polyfills.cs
viceroypenguin, line 1: If we're going to add this, should we add it for all projects? Maybe don't need to update existing code to use this in this PR, but if it's available, it odesn't have to be moved in the next PR.

MaceWindu: that's the plan, but not for this PR

MaceWindu:
Open for review (still need to add tests to CI)
net9 and extra providers/tests will go as separate tasks later

Tests/EntityFrameworkCore.FSharp/Tests.EntityFrameworkCore.FSharp.fsproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	<Import Project="..\linq2db.BasicTestProjects.props" />
+	
+	<PropertyGroup>
+		<TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
MaceWindu, line 5: F# not tested for EF 3.1 as there is no F# support for it

Source/LinqToDB.EntityFrameworkCore/LinqToDBForEFTools.ContextOptionsBuilderExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+			Action<LinqToDBContextOptionsBuilder>? linq2dbOptionsAction = null)
+			=> UseLinqToDB(optionsBuilder, linq2dbOptionsAction);
+
+		/// <summary>
+		/// Registers custom options related to LinqToDB provider.
+		/// </summary>
+		/// <param name="optionsBuilder"></param>
+		/// <param name="linq2dbOptionsAction">Custom options action.</param>
+		/// <returns></returns>
+		public static TContext UseLinqToDB<TContext>(
```
MaceWindu, line 30: One of few changes: return type changed from `DbContextOptionsBuilder` to `TContext` to allow fluent configuration without loosing context type

Source/LinqToDB/DataProvider/ProviderDetectorBase.cs
```diff
@@ -12,6 +12,8 @@ abstract class ProviderDetectorBase<TProvider,TVersion>
 		where TProvider   : struct, Enum
 		where TVersion    : struct, Enum
 	{
+		readonly bool _hasVersioning;
```
MaceWindu, line 15: provider detection logic updated to skip version detection if provider doesn't support versioning to avoid unnecessary work and errors from detection logic. E.g. in that case it was caused by connection string missing from data options, which was fine as we don't really need CS for test to work.

Source/LinqToDB/Expressions/SqlPlaceholderExpression.cs
```diff
@@ -19,7 +19,7 @@ public sealed class SqlPlaceholderExpression : Expression
 
 		public SqlPlaceholderExpression(SelectQuery? selectQuery, ISqlExpression sql, Expression path, Type? convertType = null, string? alias = null, int? index = null, Expression? trackingPath = null)
 		{
-			#if DEBUG
+#if BUGCHECK
```
MaceWindu, line 22: With new parser we have some debug-only extra code to detect errors. I've moved this code behing BUGCHECK symbol and enabled it for Azure configuration, used by CI tests. This will allow us to catch potential errrors in future.

Source/Shared/Extensions.cs
```diff
@@ -1,5 +1,23 @@
 ﻿#if NETFRAMEWORK || NETSTANDARD2_0
-using System.Runtime.CompilerServices;
+
+namespace System
+{
+	internal static class StringExtensions
```
MaceWindu, line 5: This is something I plan to expand to main lib in separate PR: emulate missing runtime APIs for old TFMs to avoid unnecessary #if-defs polluting our codebase

Tests/Base/BaselinesManager.cs
```diff
@@ -0,0 +1,45 @@
...
+using LinqToDB.Data;
+
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+
+namespace Tests
+{
+	using Model;
+
+	public static class BaselinesManager
```
MaceWindu, line 16: as we use different base class for EF.Core tests but still use some shared functionality like baselines generation and connection configuration files - that functionality was moved from TestBase class to separate classes.

Tests/Base/BaselinesWriter.cs
```diff
@@ -15,7 +15,7 @@ public sealed class BaselinesWriter
 
 		static string? _context;
 
-		internal static void Write(string baselinesPath, string baseline)
+		internal static void Write(string baselinesPath, string baseline, string? providerSuffix)
```
MaceWindu, line 18: ef test baselines saved to separate folder with ".EF" suffix to avoid naming conflicts (and make them discoverable)

Tests/Base/DatabaseUtils.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.IO;
+
+using NUnit.Framework;
+
+namespace Tests
+{
+	public static class DatabaseUtils
+	{
+		public static void CopyDatabases()
```
MaceWindu, line 10: Another piece of shared test logic extracted from TestBase

Tests/Base/TestBase.Asserts.cs
```diff
@@ -0,0 +1,242 @@
...
+using LinqToDB.Tools;
+using LinqToDB.Tools.Comparers;
+
+using NUnit.Framework;
+
+using Tests.Model;
+
+namespace Tests
+{
+	partial class TestBase
```
MaceWindu, line 18: remaining TestBase code moved to partial class files based on it's category to reduce main file size and make functionality more discoverable

Tests/EntityFrameworkCore/ContextTestBase.cs
```diff
@@ -0,0 +1,167 @@
...
+
+namespace LinqToDB.EntityFrameworkCore.Tests
+{
+	internal static class TestContextTracker
+	{
+		// cannot add it to ContextTestBase as it will have separate instance per-TContext
+		public static readonly Dictionary<string, Type> LastContexts = new ();
+	}
+
+	public abstract class ContextTestBase<TContext> : TestBase
```
MaceWindu, line 38: base class for EF.core tests. Instantiated with base EF context class as generic type parameter and implements:
- provider connection configuration
- context database creation. We create separate database, based on main linq2db connection string with extra suffix to database name to avoid linq2db database removal by EF

Tests/Linq/DataProvider/Types/ClickHouseTypeTests.cs
```diff
@@ -644,6 +644,7 @@ public async ValueTask TestString([ClickHouseDataSources(false)] string context)
 			}
 		}
 
+		[ActiveIssue("Update to new JSON type implementation")]
```
MaceWindu, line 647: unrelated change, CH introduced new JSON data type instead of old and ugly one

Source/LinqToDB/Expressions/SqlPlaceholderExpression.cs
viceroypenguin: Do we build and test both with and without `BUGCHECK`? I just want to ensure that the bits that get deployed are covered by CI - don't expect there to be any differences here, but better safe than sorry.

viceroypenguin: Reading the code under `BUGCHECK`, I'm comfortable (for now) with not doing a separate CI just for `Release`; that said, I would like to make sure we give extra attention during code reviews in the future for any code under `BUGCHECK` to ensure that it is only checks and does not have any material impact on the behavior of any object.

Source/LinqToDB.EntityFrameworkCore/LinqToDB.EntityFrameworkCore.csproj
```diff
@@ -0,0 +1,39 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<AssemblyName>linq2db.EntityFrameworkCore</AssemblyName>
+		<RootNamespace>LinqToDB.EntityFrameworkCore</RootNamespace>
+		<TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
+
+		<DocumentationFile>$(ArtifactsPath)\bin\$(MSBuildProjectName)\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
```
viceroypenguin, line 8: Is this not automatic already with `<GenerateDocumentationFile />`?

viceroypenguin (commented):
lgtm - mostly nitpicks and questions

.runsettings
```diff
@@ -6,5 +6,7 @@
 	<NUnit>
 		<AssemblySelectLimit>1000000</AssemblySelectLimit>
 		<ShadowCopyFiles>false</ShadowCopyFiles>
+		<!--<Verbosity>5</Verbosity>-->
```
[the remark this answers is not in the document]
MaceWindu: Was playing with log level for nunit trying to debug netfx assembly load issues. Looks like the only way to do it is to use debugger as nunit still doesn't log anything usefull for those cases...

Source/LinqToDB.EntityFrameworkCore/LinqToDB.EntityFrameworkCore.csproj
MaceWindu: We use this approach everywhere, not only here. Don't remember details - probably default path is not what we want

viceroypenguin: Fair enough. If I can ever find time, I'll take a look at it on a sep PR. I'm good with this for now for consistency.

viceroypenguin (approved):

Issue #4602, opened 2024-08-16
Association key not found when using custom IMetadataReader


boginw:
### Describe your issue

I'm using a custom `IMetadataReader` and `DynamicColumnsStore` to add associations without the `Column`- and `Association`-attributes. Inserting, updating, deleting, and querying a single entity works fine, but if I involve the association, I the exception listed below. Is there something I'm missing?

Exception

```
/home/bogi/projects/linq2db-metadata-bug/bin/Debug/net8.0/linq2db-metadata-bug
Unhandled exception. LinqToDB.Linq.LinqException: Association key 'ParentId' not found for type 'ChildClass.
   at LinqToDB.Linq.Builder.AssociationHelper.CreateAssociationQueryLambda(ExpressionBuilder builder, AccessorMember onMember, AssociationDescriptor association, Type parentOriginalType, Type parentType, Type objectType, Boolean inline, Boolean enforceDefault, List`1 loadWith, Boolean& isLeft)
   at LinqToDB.Linq.Builder.AssociationHelper.BuildAssociationInline(ExpressionBuilder builder, BuildInfo buildInfo, TableContext tableContext, AccessorMember onMember, AssociationDescriptor descriptor, Boolean inline, Boolean& isOuter)
   at LinqToDB.Linq.Builder.TableBuilder.TableContext.FindContextExpression(Expression expression, Int32 level, Boolean forceInner, Boolean throwExceptionForNull)
   at LinqToDB.Linq.Builder.TableBuilder.TableContext.IsExpression(Expression expression, Int32 level, RequestFor requestFlag)
   at LinqToDB.Linq.Builder.ExpressionContext.IsExpression(Expression expression, Int32 level, RequestFor requestFlag)
   at LinqToDB.Linq.Builder.ExpressionBuilder.<>c.<CheckSubQueryForWhere>b__126_0(CheckSubQueryForWhereContext context, Expression expr)
   at LinqToDB.Expressions.VisitFuncVisitor`1.Visit(Expression expr)
   at LinqToDB.Expressions.VisitFuncVisitor`1.Visit(Expression expr)
   at LinqToDB.Expressions.ExpressionExtensions.Visit[TContext](Expression expr, TContext context, Func`3 func)
   at LinqToDB.Linq.Builder.ExpressionBuilder.CheckSubQueryForWhere(IBuildContext context, Expression expression, Boolean& makeHaving)
[...]

PR #4608, opened 2024-08-28
Make AssociationHelper take DynamicColumns into consideration

boginw:
This PR refers to #4602 where I mentioned an issue when using a combination of a custom `IMetadataReader` and `DynamicColumnsStore` to hold association keys. The issue being that `AssociationHelper` didn't take Dynamic Columns into account. I hope that this solves the issue.

**N.B:** Since I'm running Ubuntu, I am unable to run the test to completion (even with Mono). However, with the use of Mono, I can run everything that doesn't talk directly to a database. Before the fix in `AssociationHelper.cs` the test fails with: 
```
Association key 'ParentID' not found for type 'Tests.Linq.DynamicColumnsTests+DynamicChild
```

And after the fix it fails with (on my machine): 
```
OleDb is not implemented
```

This seems to confirm that the issue in #4602 is gone.

Any help confirming that everything else still works properly would be greatly appreciated! :)


boginw:
I was able to use the NuGet artifacts from the CI pipeline to confirm that this does fix the issue I was experiencing :)

MaceWindu:
Ignore those failures:
- oracle fails randomly due to provider bug
- clickhouse test fails due to changes to recent ClickHouse release

MaceWindu:
Yes, https://github.com/linq2db/linq2db/pull/4595 contains fixes for recent CI issues

boginw:
Nice, thanks for the quick response :) Are there any blockers for merging?

MaceWindu:
Yes, we currently work on some preview.1 regressions I want to merge first. After that I will merge this one

boginw:
Sounds great, much appreciated :)

MaceWindu (approved):

MaceWindu:
Thanks!

PR #4609, opened 2024-08-29
Move UserDataProviders.json.template to root

Shane32:
Based on the contributing.md file:

> Repository already contains pre-configured `UserDataProviders.json.template` configuration with basic setup for SQLite-based testing and all you need is to rename it to `UserDataProviders.json`, add connection string for other databases you want to test.

As such it would seem that both files should be in the same folder.  The latter file is not included but the sln points to the file as if it existed in the repo root.  This seems correct as the `DataProviders.json` file is also in the repo root.  However, the template is not in the root.

This PR moves the template to the repo root, so as described in the readme, the file can be simply renamed/copied and then edited.


linq2db.sln
```diff
@@ -302,17 +302,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "scripts", "scripts", "{49A3
 		Build\Azure\scripts\mac.docker.sh = Build\Azure\scripts\mac.docker.sh
 		Build\Azure\scripts\mac.informix14.sh = Build\Azure\scripts\mac.informix14.sh
 		Build\Azure\scripts\mac.mariadb11.sh = Build\Azure\scripts\mac.mariadb11.sh
-		Build\Azure\scripts\mac.mysql57.sh = Build\Azure\scripts\mac.mysql57.sh
 		Build\Azure\scripts\mac.mysql.sh = Build\Azure\scripts\mac.mysql.sh
+		Build\Azure\scripts\mac.mysql57.sh = Build\Azure\scripts\mac.mysql57.sh
```
Shane32, line 306: The rest of these changes were performed automatically by VS.  Looks like it was sorting the list.

PR #4616, opened 2024-09-14
[EFCore] Regression tests for existing issues

MaceWindu:
List of failing tests:

- `ToolsTests.TestCreateTempTable` + `ToolsTests.TestChangeTrackerTemporaryTable` : v6 regression
- `ToolsTests.TestFunctions` : v6 regression
- `ToolsTests.TestDeleteFrom` : could be regression, test was enabled for more providers and fails there with bad SQL
- `IssueTests.Issue4603Test` : v6 regression. `ArgumentOutOfRangeException` due to set operator column count mismatch
- `IssueTests.Issue4570Test` : could be regression as it now fails with another exception: `ArgumentException : must be reducible node` (but works with EF 3.1)
- `IssueTests.Issue3174Test` : could be v6 regression. While original issue is not expected to befixed, now it fails with `SqlException : Table not found for 't91[Id:31].Id => t92.Id'.`
- `IssueTests.Issue4012Test` : v6 regression. `LinqToDBException : Expression 'construct Parent[Full] ... is not an association`
- `IssueTests.Issue4626Test2`: v6 regression. original issue was fixed, but `StringAggregate` test case now produce stack overlow
- `IssueTests.Issue4666Test`: partial v6 regression, now it throws `ArgumentException : Property 'System.String Type1EntityProp' is not defined for type 'LinqToDB.EntityFrameworkCore.Tests.Models.IssueModel.Issue4666BaseEntity' (Parameter 'property')`


MaceWindu:
moved to https://github.com/linq2db/linq2db/pull/4676

Issue #4528, opened 2024-06-06
Power only accept doubles


jods4:
## Problem
linq2db implementation of Power only accepts doubles and not decimals, unlike all other Math functions such Log or Round.

`Sql.Power`: https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Sql/Sql.cs#L1339
`Math.Pow`: https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Linq/Expressions.cs#L1014

## Consequences
This can lead to subtle inaccuracies when using constants.
For example, consider expression `x * Sql.Power(10, 2)` when field `x` is `8.7345`:

In Oracle, when arguments are double, generated code is `x * Power(10d, 2d)`. Notice the `d` suffix which means the type is a `BINARY_DOUBLE`. Float arithmetic then leads to value `873.449999999999` and by default Oracle/linq2db will truncate this into `NUMBER` type with value `873.4499` not exactly the expected value.

If we had `decimal` overload the SQL would be `x * Power(10, 2)`. Now Oracle performs the computation using `NUMBER` type and the result is exactly `837.45`.

## Suggestion
Add `decimal` overload in the two places linked above.

riktimmondal:
Hi.
Raise a PR https://github.com/linq2db/linq2db/pull/4635.
Please review.

PR #4635, opened 2024-09-25
Add decimal overload function of power

riktimmondal:
Fix for #4528.


MaceWindu:
Looks good, thanks! Please add test for new overload here https://github.com/linq2db/linq2db/blob/master/Tests/Linq/Linq/MathFunctionTests.cs#L193-L200

riktimmondal:
@MaceWindu Added test. Please review.

Tests/Linq/Linq/MathFunctionTests.cs
```diff
@@ -191,14 +191,23 @@ from t in from p in    Types select Math.Min(p.MoneyValue, 5) where t != 0 selec
 		}
 
 		[Test]
-		public void Pow([DataSources(ProviderName.SQLiteMS)] string context)
+		public void Pow1([DataSources(ProviderName.SQLiteMS)] string context)
```
MaceWindu, line 194: don't rename existing tests as it will mess with our test baselines. Just name new test e.g. `PowDecimal`

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -1012,6 +1012,7 @@ interface ISetInfo
 			{ M(() => Math.Log10          (0)),  N(() => L<double,double>       ((double p)          => Sql.Log10  (p)!   .Value )) },
 
 			{ M(() => Math.Pow        (0,0) ), N(() => L<double,double,double>    ((double x,double y) => Sql.Power(x, y)!.Value )) },
+			{ M(() => Math.Pow        (0,0) ), N(() => L<decimal,decimal,decimal>    ((decimal x,decimal y) => Sql.Power(x, y)!.Value )) },
```
MaceWindu, line 1015: I looked at list of available overloads and see that Math.Pow has only `double` version, which means we cannot map it for double. Please remove this line as it will override previous mapping

Tests/Linq/Linq/MathFunctionTests.cs
```diff
@@ -191,14 +191,23 @@ from t in from p in    Types select Math.Min(p.MoneyValue, 5) where t != 0 selec
...
 					from t in from p in    Types select Math.Floor(Math.Pow((double)p.MoneyValue, 3)) where t != 0 select t,
 					from t in from p in db.Types select Math.Floor(Math.Pow((double)p.MoneyValue, 3)) where t != 0 select t);
 		}
 
+		[Test]
+		public void Pow2([DataSources(ProviderName.SQLiteMS)] string context)
+		{
+			using (var db = GetDataContext(context))
+				AreEqual(
+					from t in from p in Types select Math.Floor((decimal)(8.7345*Math.Pow((double)p.MoneyValue, 2))) where t != 0 select t,
```
MaceWindu, line 207: As discussed above, because we don't have decimal version of Math.Pow, test should use Sql.Power directly (`MoneyValue` is decimal, so it will call proper overload):

```cs
Math.Floor(Sql.Power(p.MoneyValue, 3))
```

MaceWindu (changes requested):

MaceWindu:
@jods4, am I correct about having only Sql.Power for decimal? I don't see how we can map Math.Pow here without ton of ugly code. Also Math.Pow doesn't handle null values, which makes it even more useless

Tests/Linq/Linq/MathFunctionTests.cs
riktimmondal: Understood

Tests/Linq/Linq/MathFunctionTests.cs
riktimmondal: Understood

Source/LinqToDB/Linq/Expressions.cs
riktimmondal: Understood

MaceWindu:
There are build errors https://dev.azure.com/linq2db/linq2db/_build/results?buildId=13791&view=results

You can alter test in this way

```cs
Math.Floor(Sql.Power(p.MoneyValue, 3)!.Value)
```

Tests/Linq/Linq/MathFunctionTests.cs
```diff
@@ -199,6 +199,15 @@ from t in from p in    Types select Math.Floor(Math.Pow((double)p.MoneyValue, 3)
 					from t in from p in db.Types select Math.Floor(Math.Pow((double)p.MoneyValue, 3)) where t != 0 select t);
 		}
 
+		[Test]
+		public void PowDecimal([DataSources(ProviderName.SQLiteMS)] string context)
```
MaceWindu, line 203: Changes to disable ASE testing. Not worth fixing for now
```suggestion
		// Sybase: https://stackoverflow.com/questions/25281843
		[Test]
		public void PowDecimal([DataSources(ProviderName.SQLiteMS, TestProvName.AllSybase)] string context)
```

MaceWindu:
- for sybase failure I already proposed to disable test
- sql server failures could be ignored. Caused by recent change to CI images [#899](https://github.com/microsoft/mssql-docker/issues/899) and will be handled by separate PR
- oracle/firebird - unrelated failures, should go away after re-run

MaceWindu (approved):

PR #4637, opened 2024-09-27
CI Fixes

MaceWindu:
As ubuntu 22.04 kernel update broke existing sql server 2017/2019 images, I think it's a good time to look into 24.04 migration and replacing some 3rd-party images with our own.

- [x] Switch to own SQL 2017 Linux image
- [x] Switch to own SQL 2019 + FTS Linux image
- [ ] Switch to own Sybase ASE Linux image with UTF-8 support enabled
- [x] Run analyzers and examples only for default pipeline
- [x] Bump STJ version to avoid build nagging


MaceWindu:
Merging now to unblock testing, remaining tasks will go into new PR

PR #4691, opened 2024-10-19
ToLookupAsync implementation

### LINKED ISSUE #4689: Implement ToLookupAsync
requires implementation of ToLookupAsync

igor-tkachev:
Fix #4689


Source/LinqToDB/AsyncExtensions.cs
```diff
@@ -416,5 +443,153 @@ public static async Task<Dictionary<TKey,TElement>> ToDictionaryAsync<TSource,TK
...
+		/// <returns>
+		/// A <see cref="ILookup{TKey,TElement}"/> that contains keys and values.
+		/// The values within each group are in the same order as in <c>source</c>.
+		/// </returns>
+		public static Task<ILookup<TKey,TElement>> ToLookupAsync<TSource,TKey,TElement>(
+			this IQueryable<TSource> source,
+			Func<TSource,TKey>       keySelector,
+			Func<TSource,TElement>   elementSelector,
+			CancellationToken        token = default)
+			where TKey : notnull
```
Shane32, line 522: Shouldn't the key and element selectors here be expressions so that only the necessary columns can be passed to the database engine?  For example:

```cs
var usersByLastName = await db.Users.ToLookupAsync(x => x.LastName, x => x.FirstName, token);
```

If there's no `elementSelector` then it may make sense to use a `Func<>` for the `keySelector`, but with this signature, the only reasonable way to write the query is like this:

```cs
var usersByLastName = await db.Users
    .Select(x => new { x.FirstName, x.LastName })
    .ToLookupAsync(x => x.LastName, x => x.FirstName, token);
```

Seems a bit redundant.  I'd also argue that `ToDictionary` methods should change in a similar fashion.

igor-tkachev: I think by design ToXXX(Async) methods do not imply any changes to the query results. They should only transform an enumeration into a collection.

viceroypenguin (approved):

Source/LinqToDB/AsyncExtensions.cs
Shane32: Ok, that makes sense.  Not what I would do, but I understand the logic, and as stated above, it does match EF's design (for better or worse).

Shane32 (approved):

MaceWindu (approved):

PR #4749, opened 2024-12-20
Fix additional discovered regressions

MaceWindu:
- re-enable tests for ClickHouse after it is fixed on CH side
- fix query conditions merge
- more fixes to come


MaceWindu:
Let's merge it so I continue regression testing

PR #4750, opened 2024-12-21
Fix regressions

MaceWindu:
- fix mapping regression in evaluator
- fix binary operation type inference


MaceWindu:
Another batch of fixes, will be merged tomorrow

Source/LinqToDB/SqlQuery/SqlBinaryExpressionHelper.cs
```diff
@@ -0,0 +1,339 @@
...
+using System.Collections.Generic;
+
+namespace LinqToDB.SqlQuery
+{
+	internal static class SqlBinaryExpressionHelper
+	{
+		private static readonly Dictionary<(Type left, string operation, Type right), Type> _operationTypes = new()
+		{
+			// +
+			{ (typeof(byte), "+", typeof(byte   )), ((byte)1 + (byte   )1).GetType() },
```
viceroypenguin, line 11: personally i prefer the following syntax:
```suggestion
			[(typeof(byte), "+", typeof(byte   )] = ((byte)1 + (byte   )1).GetType(),
```

makes it more obvious where the key and value are, especially in complex syntax like this.

viceroypenguin (approved):

PR #4771, opened 2025-01-08
Fix aggregation query optimization regression.

sdanyliv:


Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -1313,6 +1316,101 @@ public static bool IsAggregationFunction(IQueryElement expr)
...
+				IsAggregation = false;
+				IsWindow      = false;
+				HasReference  = false;
+			}
+
+			[return : NotNullIfNotNull("element")]
+			public override IQueryElement? Visit(IQueryElement? element)
+			{
+				// if we already found aggregation or window function, we can stop
+				if (IsAggregation && IsWindow && HasReference)
```
MaceWindu, line 1340: shouldn't it be `||`?

MaceWindu (changes requested):

Source/LinqToDB/SqlQuery/QueryHelper.cs
```diff
@@ -1313,6 +1316,101 @@ public static bool IsAggregationFunction(IQueryElement expr)
...
+			}
+
+			public void Cleanup()
+			{
+				IsAggregation = false;
+				IsWindow      = false;
+				HasReference  = false;
+			}
+
+			[return : NotNullIfNotNull("element")]
```
MaceWindu, line 1336:
```suggestion
			[return : NotNullIfNotNull(nameof(element))]
```

Source/LinqToDB/SqlQuery/QueryHelper.cs
sdanyliv: No, exactly all should be true.

PR #4772, opened 2025-01-09
Fix ToString() inside concat not translated.

sdanyliv:


Source/LinqToDB/SqlProvider/SqlExpressionOptimizerVisitor.cs
```diff
@@ -977,25 +977,49 @@ protected override IQueryElement VisitIsNullPredicate(SqlPredicate.IsNull predic
...
+					}
+				}
+				else if (unwrapped is SqlFunction func)
 				{
-					var sc = new SqlSearchCondition();
-					sc.Add(condition.Condition.MakeNot());
-					sc.AddIsNull(condition.TrueValue);
-					return Visit(sc.MakeNot(predicate.IsNot));
+					// We can extend to more parameters, but it's not clear if it's needed
+					if (func.Parameters.Length == 1 && func is { IsAggregate: false, IsPure: true })
```
viceroypenguin, line 1004:
```suggestion
					if (func is { IsAggregate: false, IsPure: true, Parameters.Length: 1 })
```

viceroypenguin (approved):

MaceWindu (commented):
There are several potential issues with baselines to review:

- `UpdateTests.TestUpdateFromJoin` and some others could simplify IS NULL operand even more (this could be done in separate PR)
- I don't think `DateValue` type conversion function for Access require null check, but this should be tested (e.g. `TypesTests.DateTime1` test)
- probably correct, but need to check Access changes for `TestIssue4415_Test5` baselines

MaceWindu:
Update:
- `TestUpdateFromJoin` optimization improved and fixed incorrect sql condition evaluation logic
- `DateValue` also fails on `null`, no changes required
- `TestIssue4415_Test5` baselines are correct

MaceWindu (dismissed):

MaceWindu (approved):

MaceWindu:
I would like to review all SqlFunction/SqlExpression constructors later for nullability annotactions. Not for this PR

PR #4786, opened 2025-01-22
Add DataContextModifier option to ScaffoldCommand

johnjuuljensen:
Hey

In my ongoing quest to get my scaffolding exactly like I want it, I now have the need for changing the access modifier of the generated Context class.

I've added a property to the scaffolding dataModel that enables this.
When not specified it does as before.

I'm not entirely sure where or if it would be appropriate to add this new option to tests.


MaceWindu (approved):

MaceWindu:
Thanks!

PR #4792, opened 2025-01-26
Add option to generate static init method in scaffolding for data context

johnjuuljensen:
When needing to modify the generated ContextSchema further before first use, I've previously had to have a static Initialize method that would do those modifications, or do it in InitDataContext with a guard clause.

This change simplifies things a bit.

The generated db context looks like this:
```
	{
		static DbContext()
		{
			var builder = new MegaFluentMappingBuilder(ContextSchema);

			builder
				.Entity<ActionDetail>()
                --8<--- Mapping --8<---

            builder.Build();
			StaticInitDataContext(); <---- Call static init
        }

        --8<--- Constructors etc. --8<---

		static partial void StaticInitDataContext();
		partial void InitDataContext();
	}
```

And my partial implementation becomes as simple as this:
```
public partial class DbContext {
    public Common.Id<Client> CurrentClientId { get; } = new Common.Id<Client>( 4 );

    static partial void StaticInitDataContext() {
        ContextSchema.ConfigureForId();
    }
}
```

This and my PR related to configuring FluentMappingBuilder type solves my issues and are minimally invasive, but I don't feel entirely confident that they're the best general solution. 
Very briefly explained I need to configure ContextSchema and I need to intercept EntityMappingBuild to be able to add Query Filters.

Adding StaticInitDataContext solved the config of ContextSchema issue, but further static init options could eliminate the need for configuring the FluentMappingBuilder type (other PR).

Some options I've considered:

(For any of these to be relevant FluentMappingBuilder would need to virtual methods, at least for Entity<>)

A StaticPreInit could be useful if it was allowed to return a FluentMappingBuilder as a reference argument, but for that to be useful FluentMappingBuilder would need to have virtual methods for overloading.

A static partial CreateFluentMappingBuilder which actually returns a FluentMappingBuilder is also an option, but that'd require C# 9 and I'm not sure what the lower target for linq2db is? When not enabled the static constructor would create FluentMappingBuilder as it does today, but when enabled the other partial context implementation could do so. 

Yet another option, if C# version is no issue, is to have a `static partial FluentMappingBuilder FluentMappingBuilder { get; }` with a default implementation if nothing else is configured. Once again FluentMappingBuilder would need to be virtual.


johnjuuljensen:
This solution is working well for me, so I'm promoting from draft.
I'll be happy to make any changes needed.

Source/LinqToDB.Scaffold/Metadata/FluentMetadataBuilder.cs
```diff
@@ -550,6 +550,18 @@ void IMetadataBuilder.Complete(IDataModelGenerationContext context)
...
+			// additional initialization logic
+
+			if ( context.Options.GenerateStaticInitDataContextMethod )
+			{
+				var staticInitDataContext = context.MainDataContextPartialMethods
+					.New(context.AST.Name(DataModelConstants.CONTEXT_STATIC_INIT_METHOD))
+						.SetModifiers(Modifiers.Static | Modifiers.Partial);
+
+				context.StaticInitializer.Append( context.AST.Call( context.ContextReference, staticInitDataContext.Method.Name ) );
+			}
```
viceroypenguin, line 564: Style fix
```suggestion
			if (context.Options.GenerateStaticInitDataContextMethod)
			{
				var staticInitDataContext = context.MainDataContextPartialMethods
					.New(context.AST.Name(DataModelConstants.CONTEXT_STATIC_INIT_METHOD))
					.SetModifiers(Modifiers.Static | Modifiers.Partial);

				context.StaticInitializer.Append(
					context.AST.Call(context.ContextReference, staticInitDataContext.Method.Name)
				);
			}
```

viceroypenguin (approved):

MaceWindu (approved):

viceroypenguin (approved):

Issue #3052, opened 2021-06-14
How do I loop through all tables to set query filter automatically?


ww9:
Hi! Instead of having to set Tenantid query filter individually for each table, I'd like to set it for all tables that have TenantId column.

Like this (see comments in code):

```c#

public partial class MyDb : LinqToDB.Data.DataConnection
{
public long currentTenantId = 0;

partial void InitMappingSchema()
{
	// Instead of having to specify query filter for each table like this:
	var builder = MappingSchema.GetFluentMappingBuilder();
	builder.Entity<MyTable>().HasQueryFilter<MyDb>((q, db) => q.Where(e => e.TenantId == currentTenantId));
	builder.Entity<MyTable2>().HasQueryFilter<MyDb>((q, db) => q.Where(e => e.TenantId == currentTenantId));

	// I would like to set it for all tables that have "TenantId" columns. But I can't get this to work:
	var schema = DataProvider.GetSchemaProvider().GetSchema(this);
	foreach (var table in schema.Tables)
	{
		if (table.Columns.Where(c => c.ColumnName == "TenantId").FirstOrDefault() != null)
		{
			builder.Entity<table.GetType()>().HasQueryFilter<MyDb>((q, db) => q.Where(e => e.TenantId == currentTenantId));
		}
	}
}
}
```

### Environment details
linq2db version: *Release 4.0.0-preview.1*
Database Server: *MariaDB 10.4*
Database Provider: *MySql.Data 8.0.25*
Operating system: *Windows 10 x64*
.NET Framework: *.NET 5.0*

sdanyliv:
Try this sample. It registers all entities which are part of MyDb class. For sure constructor should be corrected for your needs.
```cs
public partial class MyDb : LinqToDB.Data.DataConnection
{
    public long currentTenantId = 1;

    public MyDb()
    {
        AddMappingSchema(InitializedMappingSchema);
    }

    private static MethodInfo RegisterTenantMethodInfo =
        MemberHelper.MethodOfGeneric(() => RegisterTenant<int>(null));

    private static MappingSchema InitializedMappingSchema = InitMappingSchema();

    public static MappingSchema InitMappingSchema()
    {
        var ms      = new MappingSchema();
        var builder = ms.GetFluentMappingBuilder();

        var props = typeof(MyDb).GetProperties().Where(p => typeof(ITable<>).IsSameOrParentOf(p.PropertyType));
        foreach (var prop in props)
        {
[...]

PR #4793, opened 2025-01-26
Add custom FluentMappingBuilderType option to ScaffoldCommand

johnjuuljensen:
Introduce a new option, `FluentMappingBuilderType`, to the `ScaffoldCommand` configuration, allowing users to specify a custom type for the `FluentMappingBuilder` in the generated DataContext.
When not specified FluentMappingBuilder is used like it currently is.

I need to add QueryFilterAttribute to many entities, like described here:https://github.com/linq2db/linq2db/issues/3052, but I'm not happy with using reflection to get the types.
So instead I've made this change which, when enabled, allows me to create my own FluentMappingBuilder, either by inheriting or by replacing, and have it used in the static constructor instead of plain FluentMappingBuilder .

It looks like this:
```
	{
		static DbContext()
		{
			var builder = new MyFluentMappingBuilder(ContextSchema); <---  Using custom type

			builder
				.Entity<SomeEntity>()
         ------- 8< ---------------------
```
and this:
```
class MyFluentMappingBuilder : FluentMappingBuilder {
    public MegaFluentMappingBuilder( MappingSchema mappingSchema ): base(mappingSchema) {}

    new public EntityMappingBuilder<T> Entity<T>( string? configuration = null ) {
        var entityBuilder = base.Entity<T>( configuration );
        if (typeof(T).IsAssignableTo(typeof(IHasTenantId))) {
            entityBuilder = AddTenantIdQueryFilter( entityBuilder );
        } else if ( typeof( T ).IsAssignableTo( typeof( IIsGlobal ) ) ) {

        ---------- 8< -------------------

        return entityBuilder;
    }
```

When inheriting I have to use the new modifier on the Entity method because it's not virtual, but since the static constructor uses the exact type it still works as intended.

If MyFluentMappingBuilder.Entity was virtual and I had some way of intercepting its construction this wouldn't be needed. (See other PR about static init method)

--------------------------------

I'm also working on an extended version of this, which is a little more metaprogrammy, but eliminates the need for building the filter expressions programmatically. It's only slightly more invasive, and I'll submit another PR with that once it's done. 

The extension is to add a type discriminator to the Entity<> function(s) of  FluentMappingBuilderType like so:
```
class MyFluentMappingBuilder : FluentMappingBuilder {
    public EntityMappingBuilder<T> Entity<T>( IHasTenantId? typeDisciminator , string? configuration = null ) where T: IHasTenantId {
        return base.Entity<T>().HasQueryFilter( ( T e, DbContext db ) => e.TenantId == db.CurrentTenantId );
    }

    public EntityMappingBuilder<T> Entity<T>( IIsInternal? typeDisciminator, string? configuration = null ) where T: IIsInternal {
        return base.Entity<T>().HasQueryFilter( ( T e, DbContext db ) => false );
    }

    public EntityMappingBuilder<T> Entity<T>( IIsGlobal? typeDisciminator, string? configuration = null ) where T : IIsGlobal {
        return base.Entity<T>( configuration );
    }

    [Obsolete]
    public EntityMappingBuilder<T> Entity<T>( object? typeDisciminator, string? configuration = null ) {
        return base.Entity<T>();
    }


    [Obsolete]
    new public EntityMappingBuilder<T> Entity<T>( string? configuration = null ) {
----- 8< ----
}
``` 

(Note: I use Obsolete and warnings as errors to identify mapping that have no tenancy relationship defined)

The mapping then looks like:
```
	{
		static DbContext()
		{
			var builder = new MyFluentMappingBuilder(ContextSchema);

			builder
				.Entity<TypeWithTenant>(typeDisciminator: default( TypeWithTenant) )
					.HasAttribute(new TableAttribute("tablename"))
					.Member(e => e.Id)

			builder
				.Entity<GlobalType>( typeDisciminator: default( GlobalType) )
                                        .HasAttribute(new TableAttribute("anothertable"))
					.Member(e => e.Id)
```

Without this I have to build the filterexpression dynamically as in issue 3052, but with this it's generated by the compiler and I get all the benefits, like find all references and rename support etc.

For this to work it's not enough to make FluentMappingBuilder.Entity<> virtual since the overload resolution requires the exact type, so I have to provide my own type.


johnjuuljensen:
This PR is made redundant by https://github.com/linq2db/linq2db/pull/4794, at least for my case.

I still feel that either this or making FluentMappingBuilder virtual could be useful, but maybe that's best solved if someone has concrete issues they need solved.

johnjuuljensen:
I have no need for this atm. and I don't want it taking up maintainers time.

PR #4794, opened 2025-01-26
Scaffolding - add entity type discriminator to fluent mapping

johnjuuljensen:
Introduce a new option, FluentEntityTypeDiscriminators, to the ScaffoldCommand configuration, which changes how the fluent mapping is built.
When not specified fluent mapping is built like it currently is.

It's enabled by providing one or more method names which will be called immediately after Entity<>.
These methods must either be present on the EntityMappingBuilder<T> or more likely as extension methods to it.
The methods must take a type discriminator argument which will be default(T) in the invocation.

I need to add QueryFilterAttribute to many entities, like described here:https://github.com/linq2db/linq2db/issues/3052, but I'm not happy with using reflection to get the types.

This PR makes my previous PR related to overriding the FluentMappingBuilder type redundant.

The extension methods could for instance look like:
```
static class FluentMappingBuilderExt {

    internal static EntityMappingBuilder<T> HasTenancyRelationShip<T>( this EntityMappingBuilder<T> builder, IHasTenantId? _) where T : IHasTenantId=>
        builder.HasQueryFilter( ( T e, DbContext db ) => e.TenantId== db.CurrentTenantId );

    internal static EntityMappingBuilder<T> HasTenancyRelationShip<T>( this EntityMappingBuilder<T> builder, IIsInternal? _ ) where T : IIsInternal =>
        builder.HasQueryFilter( ( T _, DbContext _ ) => false );

    internal static EntityMappingBuilder<T> HasTenancyRelationShip<T>( this EntityMappingBuilder<T> builder, IIsGlobal? _) where T : IIsGlobal =>
        builder;

    [Obsolete]
    internal static EntityMappingBuilder<T> HasTenancyRelationShip<T>( this EntityMappingBuilder<T> builder, object? _) {
        throw new NotImplementedException();
    }
}
```
(Note: I use Obsolete and warnings as errors to identify mapping that have no tenancy relationship defined)

The mapping then looks like:
```
	{
		static DbContext()
		{
			var builder = new MyFluentMappingBuilder(ContextSchema);

			builder
				.Entity<TypeWithTenant>()
				.HasTenancyRelationShip<TypeWithTenant>(default(TypeWithTenant)!)
					.HasAttribute(new TableAttribute("tablename"))
					.Member(e => e.Id)

			builder
				.Entity<GlobalType>()
				.HasTenancyRelationShip<GlobalType>(default(GlobalType)!)
                                        .HasAttribute(new TableAttribute("anothertable"))
					.Member(e => e.Id)
```

Without this I have to build the filterexpression dynamically as in issue 3052, but with this it's generated by the compiler and I get all the benefits, like find all references and rename support etc.


johnjuuljensen:
This latest solution is working very well for me. I currently register two discriminator methods, one for HasTenancyRelationShip and another one which registers the type in another part of the system (for a couple of purposes but one is OpenAPI documentation).

Disciminating on interfaces that has static abstract members has allowed me to eliminate all reflection.

Whether this feature is generally useful or not is unclear to me, so I'll leave it at draft pending feedback.
I do however think that these techniques will play a very important role in  AOT scenarios, eliminating reflection without resorting to source generators.

johnjuuljensen:
This solution is still working well for me, so I'm promoting from draft.
I'll be happy to make any changes needed.

MaceWindu (approved):
I've renamed option to `fluent-entity-type-helpers`

Issue #4266, opened 2023-09-04
Wrong query results when using Extention (query cache problem)


feeleen:
Trying to write my own Extention to generate subquery from array of int.
came up with solution below, but faced problem with query cache.

```csharp
using (DbContext db = new MyTest.DbContext())
{
	var persons = new List<int> { 1, 2, 3 };
	var query = from p in db.GetTable<Person>()
				where Sql.Ext.In(p.BusinessEntityID, persons)
				select p.BusinessEntityID;

	var result1 =  query.ToList(); // returns 1,2,3 Ok

    persons.AddRange(new int[] { 4, 5 });
    var query2 = from p in db.GetTable<Person>()
                where Sql.Ext.In(p.BusinessEntityID, persons)
                select p.BusinessEntityID;

    var result2 =  query2.ToList(); // returns 1,2,3 Not Ok, must be 1,2,3,4,5
}
```

Extention code:

```csharp
public static class Extentions
{
       [Sql.Extension(ProviderName.SqlServer, "{field} IN (select * from {values})", IsPredicate = true, BuilderType = typeof(InAny_ExpressionItemBuilder), ServerSideOnly = true)]
       public static bool In<T>(this Sql.ISqlExtension ext, [ExprParameter] T field, [SqlQueryDependent] IEnumerable<T> values) where T : struct, IEquatable<int>
       {
            throw new NotImplementedException();
       }
}

public class InAny_ExpressionItemBuilder : Sql.IExtensionCallBuilder
{
    public void Build(Sql.ISqExtensionBuilder builder)
    {
        var parameterName = (builder.Arguments[2] as MemberExpression)?.Member.Name ?? "p";

        var values = builder.GetValue<System.Collections.IEnumerable>("values")
                ?.OfType<int>()
                .ToArray();

        if (values == null)
        {
            throw new ArgumentNullException("values", "Values for \"In/Any\" operation should not be empty");
        }
        
        DataTable dataTable = new DataTable("IntTableType");
        dataTable.Columns.Add("Id", typeof(int));

        foreach (var x in values.Distinct())
        {
            var newRow = dataTable.Rows.Add();
            newRow[0] = x;
        }

        dataTable.AcceptChanges();

        var param = new SqlParameter(new DbDataType(dataTable?.GetType() ?? typeof(object), "IntTableType"), parameterName, dataTable);

        builder.AddParameter("values", param);
    }
}
```

[...]

PR #4810, opened 2025-02-04
Fix for #4266

feeleen:
Fix for #4266


MaceWindu:
That's not a correct fix, it basically disables query caching, which you can achive for specific query by wrapping it into
```cs
using (NoLinqCache.Scope())
```

feeleen:
@MaceWindu 

Then, as far as I can see, another possible way to fix that, is to replace the constant in Expression with new instanse of array.
Something like this

MaceWindu:
Problem here is that we don't have proper support for array/tvp as scalars yet, so simple changes will not make them works.
I have one idea, how to make your query work, let me check it first

MaceWindu:
I have workaround for you. Let's close this PR and I will add woraround to issue

PR #4092, opened 2023-04-25
Fix outer apply columns nullability tracking

Shane32:
Fix #4090 

The queries here were mostly copied from the tests for #3462 and although are not quite the same as in #4090, effectively demonstrate the bug.  Note that only when the nested OUTER APPLY has no data does the bug manifest itself, and only depending on the order of the columns selected.

PR #4814, opened 2025-02-04
Fix for issue #4813

Shane32:
Fixes #4813

Issue #4813, opened 2025-02-04
Selecting dictionaries don't provide consistent results depending on select order


Shane32:
## Issue

When selecting dictionaries to materialize related tables, whether the first related table is materialized defines whether any other related tables are materialized.  It's easier to understand from the examples below:

## Example

```cs
var queryWorksCorrectly =
    from r in _db.Returns
    where r.ReshipSaleId == null
    select new {
        r.Id,
        r.OriginalSaleId,                               // contains value
        OriginalSale = _db.Sales                        // returns object
            .Where(x => x.Id == r.OriginalSaleId)
            .Select(x => new { x.Id })
            .FirstOrDefault(),
        r.ReshipSaleId,                                 // is null
        ReshipSale = _db.Sales                          // returns null
            .Where(x => x.Id == r.ReshipSaleId)
            .Select(x => new { x.Id })
            .FirstOrDefault(),
    };
var retWorksCorrectly = await queryWorksCorrectly.Take(1).ToListAsync();
Console.WriteLine(JsonSerializer.Serialize(retWorksCorrectly, _jsonSerializerOptions));

var queryfailsNotNull =
    from r in _db.Returns
    where r.ReshipSaleId == null
    select new
    {
        r.Id,
        r.OriginalSaleId,                               // contains value
        OriginalSale = _db.Sales                        // returns object
            .Where(x => x.Id == r.OriginalSaleId)
            .Select(x => new Dictionary<string, object> { { "Id", x.Id } })
            .FirstOrDefault(),
        r.ReshipSaleId,                                 // is null
        ReshipSale = _db.Sales                          // INCORRECTLY returns object
            .Where(x => x.Id == r.ReshipSaleId)
            .Select(x => new Dictionary<string, object> { { "Id", x.Id } })
            .FirstOrDefault(),
    };
var retFailsNotNull = await queryfailsNotNull.Take(1).ToListAsync();
Console.WriteLine(JsonSerializer.Serialize(retFailsNotNull, _jsonSerializerOptions));

var queryfailsNull =
    from r in _db.Returns
    where r.ReshipSaleId == null
    select new {
        r.Id,
        r.ReshipSaleId,                                 // is null
        ReshipSale = _db.Sales                          // returns null
            .Where(x => x.Id == r.ReshipSaleId)
            .Select(x => new Dictionary<string, object> { { "Id", x.Id } })
            .FirstOrDefault(),
        r.OriginalSaleId,                               // contains value
[...]

Issue #4607, opened 2024-08-27
6.x issue using InsertWithInt64Identity with MemberInitExpression with interface property


Shane32:
When executing this code with 6.0.0-preview-1 for SQL Server the insert fails that "x.Deleted cannot be mapped to SQL":

```cs
interface IHasDeleted
{
    bool Deleted { get; set; }
}
class SomeTable : IHasDeleted
{
    public string Name { get; set; }
    public bool Deleted { get; set; }
}

await dbContext.GetTable<SomeTable>().InsertWithInt64IdentityAsync(() => new SomeTable {
    Name = "Joe",
    Deleted = true,  // <--- where this MemberBindExpression's Member property points to the interface property not the class property
});
```

This can only occur with expressions created by code, most likely, but it did work in 5.x.

Is this intentional?  Should it be fixed?  Of course I can modify my code to compensate if this is intentional or will not be fixed.  If not intentional, I can provide further details, write a test, etc.  Seems like it's pretty easy with reflection to remap the interface property to the correct property on the class.

PR #4815, opened 2025-02-05
Add removed issue-4090 tests

Shane32:
I was having an issue in my application that was indicating a regression in 6.0.0-preview-3 for the issue that #4092 tests for.  But for some reason the assertions in test #4092 were removed, asserting only the SQL generation rather than the results.  This PR adds the assertions back.

### DIFF
--- a/Tests/Linq/UserTests/Issue4090Tests.cs
+++ b/Tests/Linq/UserTests/Issue4090Tests.cs
@@ -277,9 +277,83 @@ namespace Tests.UserTests
 							Id2 = t2.Id2,
 						})
 						.FirstOrDefault()
-			});
+				});
 
-			AssertQuery(query);
+			// assert the generated SQL
+			var ret = AssertQuery(query);
+
+			// assert the result (more important than the SQL for these tests)
+			Assert.That(ret, Has.Length.EqualTo(7));
+
+			Assert.Multiple(() =>
+			{
+				Assert.That(ret[0].Name3!, Is.EqualTo("Child21"));
+				Assert.That(ret[0].Value2, Is.Not.Null);
+			});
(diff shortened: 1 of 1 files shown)


sdanyliv:
Strange that `AssertQuery` do not cover your cases. Anyway looks like possible issue should be fixed by https://github.com/linq2db/linq2db/pull/4814

MaceWindu:
@Shane32 what issue do you have? your tests pass and AssertQuery is not sql assertion, but query result assertion

Shane32:
> what issue do you have?

@MaceWindu @sdanyliv I have multiple applications built around a complex dynamic data service layer that builds queries at runtime.  These queries rely largely on support of CROSS APPLY joins and materialization of queries that return `Dictionary` types.  Over time, I've had multiple materialization issues which don't show up in SQL analysis.  A recent example is #4813 -- the problem did not occur with the SQL but with the materialization of the results.  Other issues I've encountered relate to the atypical Expression generated by my code (e.g. #4607) and simply generate an exception if they fail, while others like #4228 have to do with the generated SQL.

In this case, issue #4090 was also an issue with materialization, and a SQL assertion would not necessarily demonstrate that there was an issue.  So I'd like to maintain the assertions on the result set for #4090, verifying that the materialized results are correct, rather than (or in addition to) verifying what SQL was executed.

If `AssertQuery` does assert the results have not changed, that's fine as well.

MaceWindu:
Yes, there is no need for additional asserts, you basically assert results twice here: manually and with AssertQuery

MaceWindu:
They are not stored, they are eveluated on client and compared to database evaluation. For this reason, it cannot be used for queries with non-evaluable on client parts

PR #4825, opened 2025-02-10
Migrate configuration code to linq2db.Compat entirely

### LINKED ISSUE #4822: Move `LinqToDBSection` and related classes to `linq2db.Compat`
This code currently exists in linq2db. It should rest in linq2db.Compat going forward.

Exposed by #4821 

💣 Will be breaking change for netfx; though linq2db.Compat provides usable path forward.

viceroypenguin:
Fixes #4822


igor-tkachev (changes requested):
Do not remove it from the framework version. `ConfigurationManager` remains the primary way to configure applications in .NET FW and it does not require any dependences. You can move only the source code, though there isn't much need for it.

MaceWindu (changes requested):
Agree with @igor-tkachev here: we shouldn't break netfx behavior

MaceWindu:
let's not do it

PR #4839, opened 2025-02-19
InsertOrUpdate throws exception

igor-tkachev:
The following code has two issues:

```c#
[Test]
public void MergeSubquery([MergeDataContextSource(false)] string context)
{
	using var db  = GetDataConnection(context);

	db.BeginTransaction();

	using var tmp = db.CreateTempTable(
		"MergeTemp",
		[new { ID = 1, Name = "John" }],
		mb => mb
			.Property(t => t.ID)
				.IsPrimaryKey()
			.Property(t => t.Name)
				.HasLength(20));

	tmp.InsertOrUpdate(
		() => new
		{
			ID   = (from t in tmp where t.Name == "John" select t.ID).Single(),
			Name = "John II"
		},
		s => new { s.ID, s.Name });
}
```

1. Subquery in `InsertOrUpdate` generates an exception. Worked before.
2. `InsertOrUpdate` ignores fluent mapping. I am afraid this is architectural issue. We have expression preprocessing - method `ExposeAndPrepareExpression`, that removes fluent mapping info from expression.


igor-tkachev:
Quick fix for the fluent mapping is easy. It is enough (it is regular way). 

```c#
		public static TempTable<T> CreateTempTable<T>(
			this IDataContext               db,
			string                          tableName,
			IEnumerable<T>                  items,
			Action<EntityMappingBuilder<T>> setTable,
			BulkCopyOptions?                options      = default,
			string?                         databaseName = default,
			string?                         schemaName   = default,
			string?                         serverName   = default,
			TableOptions                    tableOptions = TableOptions.IsTemporary)
			where T : class
		{
			if (setTable == null) throw new ArgumentNullException(nameof(setTable));

			var tempTableDescriptor = GetTempTableDescriptor(db.MappingSchema, db.Options, setTable);

			if (db is DataConnection dc)
				dc.AddMappingSchema(tempTableDescriptor.MappingSchema);

			return new TempTable<T>(db, tempTableDescriptor, items, options, tableName, databaseName, schemaName, serverName, tableOptions);
		}
```

Mapping schema created for entityDescriptor should be just added to DataContext. This example fixes this, but of cause needs some work.

Source/LinqToDB/Linq/ExpressionCacheHelpers.cs
```diff
@@ -12,9 +12,15 @@ public static bool ShouldRemoveConstantFromCache(ConstantExpression node, Mappin
...
-				if (!mappingSchema.IsScalarType(node.Value.GetType()))
+				var valueType = node.Value.GetType();
+
+				// Handling UseTableDescriptor case.
+				if (valueType == typeof(EntityDescriptor))
+					return false;
+
+                if (!mappingSchema.IsScalarType(node.Value.GetType()))
 				{
-					if (node.Value is Array or FormattableString)
```
viceroypenguin, line 17: whitespace change

Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
```diff
@@ -2454,9 +2456,17 @@ bool MoveOuterJoinsToSubQuery(SelectQuery selectQuery, bool processMultiColumn)
 						    join.JoinType == JoinType.Left       ||
 						    join.JoinType == JoinType.CrossApply)
 						{
+							bool? isSingeRecord = null;
```
viceroypenguin, line 2459: spelling

viceroypenguin (approved):

viceroypenguin (approved):

MaceWindu (dismissed):
looks good, but still need to review baselines

MaceWindu (approved):
InsertOrReplace for Informix re-introduced casts around parameters, but it is not functional change and could be addressed later

PR #4842, opened 2025-02-21
Fluent mapping issue.

igor-tkachev:


Source/LinqToDB/TableExtensions.cs
```diff
@@ -97,15 +97,18 @@ internal static IDataProvider GetDataProvider<T>(this ITable<T> table)
 		}
 
 		// internal API
-		internal static DataConnection GetDataConnection<T>(this ITable<T> table)
+		internal static DataConnection? GetDataConnection<T>(this ITable<T> table, bool throwException = true)
```
viceroypenguin, line 100: i generally don't like this api style. i'd prefer a `TryGetDataConnection()` which returns `null` if it can't get one, etc. that fits better with bcl behaviors (parse vs tryparse, etc.)

viceroypenguin (dismissed):

Source/LinqToDB/TableExtensions.cs
igor-tkachev: I absolutely agree. This is right approach for public API. This one is internal and I will need to copy/paste throwing exception code a few times.

viceroypenguin: Could easily do?
```cs
		internal static DataConnection? TryGetDataConnection<T>(this ITable<T> table)
			where T : notnull
		{
			if (table.DataContext is DataConnection dataConnection)
				return dataConnection;
			if (table.DataContext is DataContext dataContext)
				return dataContext.GetDataConnection();

			return null;
		}

		internal static DataConnection GetDataConnection<T>(this ITable<T> table)
			where T : notnull
		{
			if (TryGetDataConnection(table) is { } cn)
				return cn;
		
			throw new ArgumentException($"Data context must be of {nameof(DataConnection)} or {nameof(DataContext)} type.", nameof(table));
		}
```

Source/LinqToDB/Extensions/DataOptionsExtensions.cs
```diff
@@ -777,6 +777,22 @@ public static DataOptions UseMappingSchema(this DataOptions options, MappingSche
 			return options.WithOptions<ConnectionOptions>(o => o with { MappingSchema = mappingSchema });
 		}
 
+		/// <summary>
+		/// Defines additional mapping schema to use with DataOptions.
```
MaceWindu, line 781: Let's add here and to UseMappingSchema more information about difference between them:
- UseMappingSchema : add information that it replaces all previous registrations by both UseMappingSchema and UseAdditionalMappingSchema
- UseAdditionalMappingSchema : add information and it combines mapping information with schemas added before by UseMappingSchema or UseAdditionalMappingSchema

Source/LinqToDB/TableExtensions.cs
MaceWindu: I agree it is better to have two APIs instead of supporessing NRT errors

Source/LinqToDB/Data/DataConnectionExtensions.cs
```diff
@@ -2407,7 +2412,16 @@ public static BulkCopyRowsCopied BulkCopy<T>(this ITable<T> table, BulkCopyOptio
 
 			using var _ = ActivityService.Start(ActivityID.BulkCopy);
 
-			var dataConnection = table.GetDataConnection();
+			var dataConnection = table.GetDataConnection(options.BulkCopyType != BulkCopyType.RowByRow);
+
+			if (dataConnection == null)
+			{
+				return new RowByRowBulkCopy().BulkCopy(
```
MaceWindu, line 2419: with this implementation we should remove `virtual` from BasicBulkCopy.RowByRowCopy[Async] methods to avoid pootential issues in future

MaceWindu (changes requested):

Source/LinqToDB/TableExtensions.cs
MaceWindu: another option is to add `[DoesNotReturnIf(true)]` to boolean parameter. It is not exactly what we need here, but technically without `never` type in C# it fits our needs

igor-tkachev: It's even worse, now I have to decide which method to call depending on some conditions.

viceroypenguin: Correct. That's a benefit. Because now you can properly think about which one you should be using, instead of hiding it behind a `!`.

Source/LinqToDB/Data/DataConnectionExtensions.cs
igor-tkachev: `RowByRowBulkCopy` is an `internal sealed` class. What can happen?

Source/LinqToDB/IDataContext.cs
```diff
@@ -126,5 +126,12 @@ public interface IDataContext : IConfigurationID, IDisposable, IAsyncDisposable
 		/// Gets initial value for database connection configuration name.
 		/// </summary>
 		string?                       ConfigurationString         { get; }
+
+		/// <summary>
+		/// Adds mapping schema to current context.
+		/// </summary>
+		/// <param name="mappingSchema">Mapping schema to add.</param>
+		/// <returns>Returns current context instance.</returns>
+		IDataContext AddMappingSchema(MappingSchema mappingSchema);
```
MaceWindu, line 135: I propose to return `void` here. No big value in returning context in general, especially when it downtyped to interface.

Source/LinqToDB/Data/DataConnectionExtensions.cs
MaceWindu: Here we create `RowByRowBulkCopy` (basically `BasicBulkCopy`) instance instead of provider-specific instance. It calls base implementation of `virtual RowByRowCopy[Async]` method.

This is fine, but problem is that any provider-specific overloads to `RowByRowCopy[Async]` will not be called which could lead to confusion/design gap. I think best solution is to remove virtual from `RowByRowCopy[Async]` methods as I doubt it needs provider-specific overload anyways - base logic is complete.

MaceWindu (changes requested):

PR #4844, opened 2025-02-23
New Window Functions syntax. DistinctBy translation support.

sdanyliv:
### Implemented Features
- Added support for `DistinctBy`.
- Implemented several window functions for evaluating proposed syntax.

---

### Current Implementation (preliminary, subject to changes, frame boundaries not yet implemented)

```csharp
var query =
    from t in db.SomeTable
    let wnd = Sql.Window.DefineWindow(f => f.PartitionBy(t.Field1).OrderBy(t.Field2))
    select new
    {
        RN1 = Sql.Window.RowNumber(f => f.PartitionBy(t.Field1).OrderBy(t.Field2)),
        RN2 = Sql.Window.RowNumber(f => f.UseWindow(wnd)),

        Sum1 = Sql.Window.Sum(t.Field, f => f.PartitionBy(t.Field1).OrderBy(t.Field2)),
        Sum2 = Sql.Window.Sum(t.Field, f => f.UseWindow(wnd)),
        SumFiltered = Sql.Window.Sum(t.Field, f => f.Filter(t.Field > 10).PartitionBy(t.Field1).OrderBy(t.Field2))
    };
```

### Aggregate Functions Examples

```csharp
var query =
    from t in db.SomeTable
    group t by t.Category into g
    select new
    {
        PercentileCont = g.PercentileCont(0.5, (e, f) => f.OrderBy(e.Field1))
    };
```

```csharp
var aggregated = db.SomeTable.PercentileCont(0.5, (e, f) => f.OrderBy(e.Field1));
```

---

#### COUNT

| SQL Syntax                               | Proposed C# Syntax                                         |
|------------------------------------------|------------------------------------------------------------|
| `COUNT() OVER(...)`                      | `Sql.Window.Count(f => f.PartitionBy(x.Some))`             |
| `COUNT(x.Some) OVER(...)`                | `Sql.Window.Count(x.Some, f => f.PartitionBy(x.Some))`     |
| `COUNT(DISTINCT x.Some) OVER(...)`       | Possible variations include:<br>1. `Sql.Window.CountDistinct(x.Some, f => f.PartitionBy(x.Some))`<br>2. `Sql.Window.Count(f => f.DistinctBy(x.Some).PartitionBy(x.Some))`<br>3. `Sql.Window.Count(a => a.Distinct(x.Some), f => f.PartitionBy(x.Some))` |

#### SUM, AVG, etc.

| SQL Syntax                               | Proposed C# Syntax                                         |
|------------------------------------------|------------------------------------------------------------|
| `SUM(x.Some) OVER(...)`                  | `Sql.Window.Sum(x.Some, f => f.PartitionBy(x.Some))`       |
| `SUM(DISTINCT x.Some) OVER(...)`         | Possible variations include:<br>1. `Sql.Window.SumDistinct(x.Some, f => f.PartitionBy(x.Some))` (Note: multiple overloads required)<br>2. `Sql.Window.Sum(a => a.Distinct(x.Some), f => f.PartitionBy(x.Some))` |

#### Recommended syntax (preferred for clarity and extensibility):

```csharp
Sql.Window.Count(x.Some, f => f.PartitionBy(x.Some))
Sql.Window.Count(a => a.Distinct(x.Some), f => f.PartitionBy(x.Some))

Sql.Window.Sum(x.Some, f => f.PartitionBy(x.Some))
Sql.Window.Sum(a => a.Distinct(x.Some), f => f.PartitionBy(x.Some))
```

---

#### SQL Frame Syntax:
```sql
[ROWS | RANGE | GROUPS] frame_start [frame_exclusion]
[ROWS | RANGE | GROUPS] BETWEEN frame_start AND frame_end [frame_exclusion]

frame_exclusion:
    EXCLUDE CURRENT ROW
    EXCLUDE GROUP
    EXCLUDE TIES
    EXCLUDE NO OTHERS
```

#### SQL Examples:
```sql
-- Rows Example
SUM(x.Some) OVER (
    ORDER BY x.Some
    ROWS BETWEEN 1 PRECEDING AND CURRENT ROW
)

-- Range Example
SUM(x.Some) OVER (
    ORDER BY x.Some
    ROWS BETWEEN 1 PRECEDING AND UNBOUNDED FOLLOWING
)
```

#### Current Proposed C# Syntax:

Currently proposed (open to discussion):
```csharp
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RowsBetween.Value(1).And.Current)
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RangeBetween.Value(1).And.Unbounded)
```

Alternate possibilities:
```csharp
// Alternative simplified syntax
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RowsBetween.Value(1).Current)
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RangeBetween.Value(1).Unbounded)

// More verbose but explicit syntax
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RowsBetween(b => b.Value(1), b => b.Current))
Sql.Window.Sum(x.Some, f => f.OrderBy(x.Some).RangeBetween(b => b.Value(1), b => b.Current))
```

---

### Seeking Feedback

Please share your thoughts on the proposed syntaxes and indicate your preference, especially considering clarity, consistency, and extensibility.


jods4:
@linq2db/developers As we're introducing new apis for this complex and often-used feature, I think this deserves a thorough discussion with the whole team to try and land the best DX possible. This is a complex API surface for sure, there could be many different ideas.

It's also good to design in a way where we're able to cover the complete SQL standard regarding windows, so I'll start by summarizing Postgre docs on the topic (as Postgre is probably the most standard-compliant DB that I know of).

First a quick note: window functions apply to _aggregate-like_ functions, not _aggregates._
They have a lot in common but the key differences IMHO is that _aggregates_ have `ALL | DISTINCT` support and built-in `ORDER BY` (incl. `WITHIN GROUP` stuff).
Here's the [Aggregates syntax from Postgre](https://www.postgresql.org/docs/17/sql-expressions.html#SYNTAX-AGGREGATES):
```
aggregate_name (expression [ , ... ] [ order_by_clause ] ) [ FILTER ( WHERE filter_clause ) ]
aggregate_name (ALL expression [ , ... ] [ order_by_clause ] ) [ FILTER ( WHERE filter_clause ) ]
aggregate_name (DISTINCT expression [ , ... ] [ order_by_clause ] ) [ FILTER ( WHERE filter_clause ) ]
aggregate_name ( * ) [ FILTER ( WHERE filter_clause ) ]
aggregate_name ( [ expression [ , ... ] ] ) WITHIN GROUP ( order_by_clause ) [ FILTER ( WHERE filter_clause ) ]
```
I think we should keep them in our mind but they could also have a different "entry" api?

Now here's the [Postgre syntax for window functions](https://www.postgresql.org/docs/17/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS). First, what's an _aggregate-like_ function?
```
function_name ([expression [, expression ... ]]) [ FILTER ( WHERE filter_clause ) ] OVER window_name
function_name ([expression [, expression ... ]]) [ FILTER ( WHERE filter_clause ) ] OVER ( window_definition )
function_name ( * ) [ FILTER ( WHERE filter_clause ) ] OVER window_name
function_name ( * ) [ FILTER ( WHERE filter_clause ) ] OVER ( window_definition )
```
- It's a function with zero (`*` such as `count(*)`), one, or several (e.g. `string_agg` separator) parameters;
- It can have a `FILTER` applied (so do aggregates).
- It's applied `OVER` a window: either **inline** `(window_definition)` or that references **a previously named window** `window_name`.

The ability to reference a `window_name` has a big impact in that we need to be able to define a window without a function, and refer to it, but syntax-wise it's the same. It translates into a clause `WINDOW window_name AS ( window_definition )` in the statement.

Here's the `window_definition` syntax (same inline or as `WINDOW`):
```
[ existing_window_name ]
[ PARTITION BY expression [, ...] ]
[ ORDER BY expression [ ASC | DESC | USING operator ] [ NULLS { FIRST | LAST } ] [, ...] ]
[ frame_clause ]
```
- `existing_window_name` can be used to _extend_ a named `WINDOW` definition, e.g. change it's ORDER BY, PARTITION BY or frame. I am not familiar with this, my understanding is that whatever you specify fully replaces the equivalent clause in the existing window. E.g. `WINDOW w1 as (PARTITION BY department ORDER BY salary DESC), w2 as (w1 ORDER BY salary ASC)`.
- `PARTITION BY` is just a list of expressions to use as partition keys
- `ORDER BY` is our old friend. It'd be good to check if we can unify `ORDER BY`s from other places, to have consistent `NULLS FIRST` support for example. I'm not familiar with the `USING operator` clause, I assume you can define custom comparisons in Postgre?

Finally `frame_clause` indicates which siblings rows are part of the aggregate for each selected row:
```
{ RANGE | ROWS | GROUPS } frame_start [ frame_exclusion ]
{ RANGE | ROWS | GROUPS } BETWEEN frame_start AND frame_end [ frame_exclusion ]
```
- Three modes: a `RANGE` of actual values, a number of `ROWS` or a number of sibling `GROUPS`, where a group is defined by `ORDER BY` equivalence (so `GROUPS` can't be used without `ORDER BY`).
- A start and optionnally an end.
- A clause to exclude specific rows.

Frames are: 
```
UNBOUNDED PRECEDING
offset PRECEDING
CURRENT ROW
offset FOLLOWING
UNBOUNDED FOLLOWING
```

Note that offset is usually an integer, but when the frame is defined as `RANGE` it can be other types, e.g. an `INTERVAL` if the window is ordered on a `DATE` (e.g. `RANGE '7 days' PRECEDING`).

Exclusion are: 
```
EXCLUDE CURRENT ROW
EXCLUDE GROUP
EXCLUDE TIES
EXCLUDE NO OTHERS
```

jods4:
@sdanyliv You'll have to let me know what is possible with the new parser or not. 
Here I'm really thinking only as an end-user, what syntax I would ideally like best.

Here's the current PR syntax from a random test case:
```cs
Sql.Window.Sum(t.IntValue, w => w.PartitionBy(t.CategoryId).OrderBy(t.Id))
```

I am not a big fan of nesting the window definition inside the aggregate function, nor using lots of lambdas, esp. in a LINQ expression where I think they wouldn't be required?

Here's one idea: what if we started with the window and ended with the aggregate?
I know it doesn't follow exactly the SQL order, but as much as I like have linq2db look like a literal SQL DSL, sometimes a few changes just make it more usable from C#...

```cs
from t in table
// Can we use fluent api without lambdas to define the window?
let w = Sql.Window("optional_name").PartitionBy(t.Country)
select new
{
  // Can we end with aggregate?
  // This works naturally for existing windows:
  sumByCountry = w.Sum(t.Salary),
  // Also work to modify a named window?
  sumByAge = w.PartitionBy(t.Age).Sum(t.Salary),
  // Or define a new window
  localWindowSum = Sql.Window.PartitionBy(t.Country).Sum(t.Salary),
}
```
What do you guys think?

igor-tkachev:
An issue with the proposed `Sql.Window()` is that it conflicts with the existing `Sql.Window` property.
Instead of `Sql.Window()`, we could use something like `Sql.Window.Define()`, `Sql.Window.Create()`, or `Sql.Window.New()`.

sdanyliv:
@jods4, try applying your syntax to small test classes first. The issue here is that it doesn’t guarantee a correct window function definition. Some functions require a partition, while others require `OrderBy`. For instance, `PercentileCont` mandates a single `OrderBy`, and the argument of `OrderBy` determines the function’s return type.  

I’m also not a big fan of using a nested lambda inside a window function, but in this case, it’s the most capable approach since the builder is strictly constrained to a specific window function.  

Keep in mind that lambda syntax greatly simplifies translation. Once we identify the `RowNumber` function, everything is encapsulated within its definition.  

Regarding window definition, I might be mistaken, but I don’t see much benefit in defining a window the way you suggested. Check the `RowNumberWithMultiplePartitionsWithDefineWindow` test—it demonstrates how I define such a window. There’s no name assigned yet, and it may not even be necessary. In fact, we could automatically detect duplicate window definitions and incorporate them into the query if it significantly improves SQL performance.

jods4:
@igor-tkachev 
> An issue with the proposed Sql.Window() is that it conflicts with the existing Sql.Window property.

Wasn't it introduced for this release? I just looked in 5.1.1 completion and I don't see it.
Anyway we can adjust, e.g. we can start off of `.Window` property as you suggested, it's not a blocker.

@sdanyliv 
> The issue here is that it doesn’t guarantee a correct window function definition.

This is a good first design question. Is it an explicit goal to enforce 100% valid syntax?
In general I agree it's a good thing to have, but if it turns out to be very tricky, I think we can shift some blame for writing valid SQL to users and they have to test.

But let's talk about some concrete examples, as it's interesting. You mention `PERCENTILE_CONT`. This is part of _ordered set aggregate functions_ and their syntax is actually different.
As you mention they take a single order in `WITHIN GROUP (ORDER BY ..)`, not inside `OVER (..)`.

In fact in Postgre, these functions DO NOT support windows! It seems to be the simple approach for these would be a dedicated two args function:
```cs
// T PercentileCont<T>(double fraction, T withinGroupOrderBy)
Sql.PercentileCont(0.34, table.Salary)
```

Things get more fun if you look at other DB such as MSSQL. `PERCENTILE_CONT` is very similar to Postgre (forced usage of `WITHIN GROUP`) _but MSSQL accepts a window, containing only a `PARTITION BY`:
```
-- MSSQL syntax
PERCENTILE_CONT ( numeric_literal )
    WITHIN GROUP ( ORDER BY order_by_expression [ ASC | DESC ] )
    OVER ( [ <partition_by_clause> ] )
```

Per-provider syntax is not enforceable in our API. For the MSSQL extension, I personally would suggest re-using the same `PercentileCont` signature as above at the end of a window:
```cs
Window.PartitionBy(table.Department).PercentileCont(0.34, table.Salary)
```

Now this is not perfect:
- It would fail on RDBMS where this is not supported (e.g. Postgre). I don't think there's anything we can do about it.
- It would fail on MS SQL if you add invalid window clauses, such as referring to a named window (this would always fail anyway in MSSQL < 2022) or if you add `ORDER BY`. Quite frankly, I would be ok with this: you get the query you asked for, and it didn't make sense.

I would need to lookup how the standard describes these functions. Maybe the `PARTITION BY` clause is not even a proper "window" per se, in which case I would suggest an overload with additional parameters rather than using `Window` here.

@sdanyliv can you think of other significant cases where using the window first would allow invalid SQL, that are significant enough that we shouldn't allow them?

> Regarding window definition, I might be mistaken, but I don’t see much benefit in defining a window the way you suggested.

Are you referring to the name usage only, or something else?
This was an idea to provide an optional name in generated SQL, I think we have a similar feature for CTE?
```cs
// WINDOW w as (PARTITION BY t.Department)
let w = Window.PartitionBy(t.Department)

// WINDOW dept as (PARTITION BY t.Department)
let w = Window.WithName("dept").PartitionBy(t.Department)
```
I don't feel strongly about this, we can always add later or not at all. As I said, I was mostly inspired by similar CTE option.

jods4:
I did a quick search on `PERCENTILE_CONT`. Postgre is the outlier here.
MSSQL, Redshift, Snowflake, Oracle, MariaDB all define this function as both an aggregate and a window function -- and they all use the exact same syntax: when it's used as a window function only `PARTITION BY` seems supported by RDBMS.

If we really wanted to enforce this at syntax level (personally not convinced) we could do it with "clever" extensions methods.
If we `Window` is a `IUnorderedWindow : IWindow` and becomes `IOrderedWindow : IWindow` after `.OrderBy()`, then `PercentileCont` can be an extension method on `IUnorderedWindow` only.

Likewise frames (`RANGE | ROWS | GROUP`) only make sense on ordered window so could be extensions of `IOrderedWindow`.

It should be noted that if we want to go to the very end of this, `RANGE` is only valid when there's a single `ORDER BY` expression, but this is something that can be done as well. In fact, because of how `ThenBy` can only be proposed after an `OrderBy` we already need some differentiation in static typing based on number of columns ordered by. And if we want to type the range strictly, the type of that first expression must be reflected in that static type.

sdanyliv:
@jods4, I'm open to your API proposal—mainly consisting of extension methods and interfaces—which would require minimal rework on my end.  

1. Window functions should reuse a predefined window.  
2. They should return the correct type.  
3. The compiler should prevent defining an invalid window unless a predefined one is used.  

If you can implement this for several complex window functions, I'll review and compare. This part is quite tricky...  

Leave PercentileCont behind, It is realy strange function and you are wrong with syntax, it is aggregation function with OVER PART and actually it perfectly fits to proposed API. Check sample in this PR comment.

jods4:
@sdanyliv I like this initiative and I agree with you the current syntax is clumsy.
I especially don't like the `ToValue()` call either, which is why I had the idea of putting the _function_ at the end.

I think my proposal is not perfect either, I'm just trying to put out as many ideas as possible so we can pick the best one before publishing them.

> Leave PercentileCont behind, It is realy strange function and you are wrong with syntax,

What do you mean I'm wrong with syntax? I have check 6 different DB docs, the following was quoted straight from MS documentation. With exception of Postgre, which doesn't support `OVER` at all, all RDBMS uses `WITHIN GROUP` instead of `OVER` for the `ORDER BY`: 
```
PERCENTILE_CONT ( numeric_literal )
    WITHIN GROUP ( ORDER BY order_by_expression [ ASC | DESC ] )
    OVER ( [ <partition_by_clause> ] )
```

Compared approaches from point 5:
```cs
// Current linq2db, I think this is worse option
RN1 = Sql.Ext.RowNumber().Over().PartitionBy(t.Field1).OrderBy(t.Field2).ToValue(),
// Current PR, using lambda inside analytical function
RN1 = Sql.Window.RowNumber(f => f.PartitionBy(t.Field1).OrderBy(t.Field2)),
// What I had proposed in comment above:
RN1 = Sql.Window.PartitionBy(t.Field1).OrderBy(t.Field2).RowNumber()
```

On this example, the PR and my proposal are very similar, I personally like the 3rd option better that 2nd because there's no nesting or lambda.

**Can you tell me what you think are "hard" examples so that I can try to see if they could fit in a different proposal?**

For _ordered set aggregates_ `mode`, `percentile_cont`, `percentile_disc`, I would handle them as above: with 2 mandatory arguments (`fraction` + `order expression`). If the provider supports using them as windowed functions with an `OVER (partition)` I would chain them after a window.

Note that some DB (e.g. Postgre and Oracle) also have very weird _Hypothetical-Set Aggregates_ like `rank(args) within group (order by sorted_args)`. Althought these are never window functions, their treatment of `order by` would be off and inconsistent with what I proposed for `percentile_cont` (because they have multiple values and should support `nulls first` etc. 😞

sdanyliv:
> `RN1 = Sql.Window.PartitionBy(t.Field1).OrderBy(t.Field2).RowNumber()`

This one is bad idea ever :)

When you want `RowNumber`, you type `RowNumber` then think about window. 

Fail example, with filter:
`Sql.Window.Filter(t.Some < 10).PartitionBy(t.Field1).OrderBy(t.Field2)....` oops `RowNumber` is not psossible because of `Filter`.

jods4:
I'm curious what other think about it. 

In C# functions naturally come at the end. For example: `list.GroupBy(x => key).Sum()`. First you define by what you want to group, then you compute sums and the return type is defined by the last function called.

So yes it's backward from SQL, but it feels somewhat natural in C# to define what you want to partition by first, then aggregate over those partitions. Another example is how "FROM/SELECT" are inverted in C# compared to SQL.

And I'm gonna add that I never know where those functions actually are (on `Sql`, on `Sql.Ext`, on `OracleSql`, on my own extension class? depends...), so if they are always extensions method on `Sql.Window` they might actually be easier to discover once you know you should start with the window, always).

The `FILTER` point is a good one.
BTW `FILTER` is not actually part of the window. In your syntax, you muddy the water a bit by putting everything in the same lambda. 
We should be able to do `SUM(salary) FILTER (WHERE gender = 'F') OVER namedWindow`.
So I assume you support this: `Sql.Sum(t.Salary, f => f.Filter(t.Gender == 'F').UseWindow(namedWindow))` ?

What I like about your way is that for the cost of one lambda it actually extends naturally to functions that take `FILTER` or complex `WITHIN GROUP`, it generalizes quite well.

Some attempts at fitting `FILTER` in my proposal:

Put it before the function, after the window: `window.Filter(x.criteria > 5).Sum(x.value)`.
This feels similar to how queries are built in LINQ, it looks a bit like a `.Where(..)`.
You are correct that only _aggregates_ can use `FILTER`, not _window functions_. So we would need to type this with some `IFilteredWindow`.

Other idea, we could think of `FILTER` as part of the aggregate function itself (as it isn't part of the window), so have it as an overload parameter for each aggregate function? `Sum<T>(T value)` and `Sum<T>(T value, Func<T, bool> filter)`?
`window.Sum(x.value, x.criteria > 5)`.

jods4:
I think I need more time to think about aggregates options (even without window). I'm not super happy with my proposals, I'm starting to think yours is more flexible in that regard.

It would be nice to have a generic way to combine aggregate options, and they can get quite complex.
For example, this is valid (no window):
```sql
string_agg(DISTINCT name, ',' ORDER BY name NULLS LAST) FILTER (age > 18) 
```

I assume with your proposal it is the following?
```cs
Sql.StringAgg(t.Name, ",", f => f.Distinct().OrderBy(t.Name).NullsLast().Filter(t.Age > 18))
```

sdanyliv:
>I assume with your proposal it is the following?
> `Sql.StringAgg(t.Name, ",", f => f.Distinct().OrderBy(t.Name, NullsPosition.Last).Filter(t.Age > 18))`

Yes, but with a key difference. The `Aggregate` function requires the first parameter to be an `IEnumerable`, and references from the outer query cannot be used—only those within the lambda expression.

Here's an example:

```csharp
var query = 
   from t in db.Some
   group t by t.Category into g
   select new 
   {
      g.Key,
      Sum = g.Sum(e => e.Quantity), // Example of a standard aggregate
      Aggregated = g.StringAgg(e => e.Name, ",", (e, f) => 
         f.Distinct()
          .OrderBy(e.Name, NullsPosition.Last)
          .Filter(e.Age > 18))
   };
```

Alternatively, using an `IQueryable` extension:

```csharp
var result1 = db.Some.StringAgg(e => e.Name, ",", (e, f) => 
   f.Distinct()
    .OrderBy(e.Name, NullsPosition.Last)
    .Filter(e.Age > 18));

var result2 = await db.Some.StringAggAsync(e => e.Name, ",", (e, f) => 
   f.Distinct()
    .OrderBy(e.Name, NullsPosition.Last)
    .Filter(e.Age > 18));
```

For such aggregates, three functions need to be defined: one for `IEnumerable` extensions and two for `IQueryable`. I have tested this approach successfully with `PercentileCont`.

sdanyliv:
> Does the following work (aggregates without group-by)?
> ```cs
> db.Some.Select(x => new 
> {
>   MeanAge = Sql.Mean(x.Age),
>   MeanSalary = Sql.Mean(x.Salary),
> }) ;
> ```

Don't think that it is a good approach. Without group by it is easy to create invalid SQL: 
```cs
db.Some.Select(x => new 
{
  Age = x.Age,
  MeanAge = Sql.Mean(x.Age),
  MeanSalary = Sql.Mean(x.Salary),
}) ;
```
> What do you think of creating a generic `Aggregate(..)`

Such function already in this PR, called `AggregateExecute`, name is the subject to change.

jods4:
> Don't think that it is a good approach. Without group by it is easy to create invalid SQL.

True. Do we have a C# syntax to generate such SQL then? Was it the `.GroupBy(1)` trick (constant group by is removed from query)? I can't remember.

> Such function already in this PR, called AggregateExecute, name is the subject to change.

Ah, great minds think alike! 😆

sdanyliv:
> Do we have a C# syntax to generate such SQL then? Was it the .`GroupBy(1)` trick (constant group by is removed from query)? I can't remember.

Yes, grouping by constant will generate SQL without grouping but with two or more aggregates.

igor-tkachev:
> > `RN1 = Sql.Window.PartitionBy(t.Field1).OrderBy(t.Field2).RowNumber()`
> This one is bad idea ever :)

Agree. But we can add this option as an alternative, if the first option remains available.

sdanyliv:
> But we can add this option as an alternative, if the first option remains available.

Previous Windows functions will work, but without support priority.

igor-tkachev:
I suggest adding another option:

`Sql.Window.Count().a.Distinct(x.Some).Over(f => f.PartitionBy(x.Some))`

This would be more consistent with:

`Sql.Window.Sum().a.Distinct(x.Some).Over(f => f.PartitionBy(x.Some))`

and other similar patterns.

jods4:
@igor-tkachev 
How would you address the following two points with your proposal?

(1) How is your expression typed as its value? 
Does `.Over()` return a type appropriate for the target function?
If so, does it mean that the exact same syntax cannot be used when those functions are used as aggregates (no OVER clause)?

(2) Can you provide an example for more analytical functions, do you think the syntax scales nicely?
For instance, how would 
`string_agg(DISTINCT name, ',' ORDER BY name NULLS LAST) FILTER (age > 18)` 
look in C#, in a way that's somewhat re-usable / consistent across window and aggregate functions?

jods4:
@sdanyliv With the syntax currently proposed in this PR, what's your take on differentiating `ORDER BY` from inside `OVER` vs `WITHIN GROUP` vs inside function? It seems to me you want to use an `.OrderBy()` call for all these situations but they are actually slightly different.

This is a bit hypothetical at the moment because I don't know any DB engine that supports it, but the following does make sense:
```sql
string_agg(name, ',' ORDER BY age) OVER (
  PARTITION BY dept 
  ORDER BY code 
  GROUPS BETWEEN 1 PRECEDING AND 1 FOLLOWING
)
```
This extremely convoluted but it has sense:
`GROUPS BETWEEN` modifies what rows belong to the window, based on `ORDER BY code` result.
Then `ORDER BY age` says how these rows must be sorted when they are concatenated by `string_agg`.

If you try that with Postgre it doesn't say this is invalid. It says: _"aggregate ORDER BY is not implemented for window functions"_ (so you can't even use `string_agg` with `OVER(PARTITION BY)`).
I don't know of any DB that supports this today, but it could be supported in the future (MSSQL already supports STRING_AGG OVER() but only with PARTITION BY, no ORDER).

A similar logic/argument exists for functions that use `WITHIN GROUP ORDER BY`.

So I think it'd be nice to reflect this structure in our API, although I also don't want it to become overly verbose.

Proposal: should we add `.Over()`, always at the end, and precede the window definition (so it's clear we're not WITHIN GROUP)?
```cs
Count(x => x.Over(declaredWindow));
Sum(cost, x => x.Over.PartitionBy(color).OrderBy(date));
StringAgg(name, ", ", x => x.OrderBy(age).Over.OrderBy(dept).Groups(1, 1))
```

sdanyliv:
> I suggest adding another option:
>
>`Sql.Window.Count().Distinct(x.Some).Over(f => f.PartitionBy(x.Some))`
>`Sql.Window.Count().Over(f => f.PartitionBy(x.Some))`

That's why we rewrite window functions, they have almost similar syntax:
Let's brake this down from compiler perspective
```cs
var query = 
    from t in db.Some
    select new 
    {
         a = Sql.Window.Count(), // invalid but compilable
         b = Sql.Window.Count().Distinct(x.Some), // invalid but compilable
         valid = Sql.Window.Count().Distinct(x.Some) .Over(f => f.PartitionBy(x.Some)) 
    }
```
Pros:
1. Simple for reading and maybe typing

Const:
1. Function can be divided to invalid parts
2. Count of Interfaces, which should be introduced grows significantly and it is simplest function.
3. Parsing is complicated, We should start translation from `Over`. then collect everything  up to  `Sql.Window.Count()`
4. Well... many hidden problems which users may introduce..

sdanyliv:
@jods4, good point! I'm starting to think that the `Sql.Window` prefix might be too weak. What do you think about `Sql.Functions` instead?  

I also changed `.Over()` to a method instead of a property for consistency:  

```csharp
Sql.Functions.Count(x => x.Over(declaredWindow));  
Sql.Functions.Sum(cost, x => x.Over().PartitionBy(color).OrderBy(date));
```

Regarding aggregates, you missed how they should be defined—as an extension over `IEnumerable`:  

```csharp
var query =  
    from t in db.Some  
    group t by t.SomeId into g  
    select new  
    {  
        g.StringAgg(e => e.name, ", ", (e, x) => x.OrderBy(e.age).Over().OrderBy(e.dept).Groups(1, 1))  
    };
```

There are still open questions about `CountDistinct` and frame definition. `.Groups(1, 1)` is the simplest approach.

jods4:
@sdanyliv maybe I'm crazy but I'd like to circle back to my initial idea (put window/over first, like groupby) with a mix of your options builder.

I'm going to put the _window_ upfront (everything inside `OVER()`).
Everything that is either inside the function (such as `ALL|DISTINCT` in aggregates and `ORDER BY` in string_agg) or before `OVER` (i.e. `WITHIN GROUP` and `FILTER`), I'm going to put inside the function builder, too.

```cs
// First we'll have a Window class that extends IEnumerable
sealed class Window<T> extends IEnumerable<T>;

// A window can be defined with a builder. Here's a WINDOW clause:
from p in People
let wnd = Sql.Window(w => w.PartitionBy(p.Gender).OrderBy(p.Age))

// Aggregates are extensions of IEnumerable<T>
// They are configured with builder, too
from g in db.People.GroupBy(p => p.Age)
select new 
{
  Count = g.Count(), // basic usage: COUNT(*)
  Count = g.Count(e => e.Nationality, (a, _) => a.Distinct()), // COUNT(DISTINCT nationality)
  FilteredMax = g.Max(e => e.Siblings, (a, e) => a.Filter(e.Gender == 'F')) // MAX(siblings) FILTER (WHERE gender = 'F')
}

// Aggregates can be called on IQueryable through `Aggregate[Async]()`  (name up to bikeshed), 
// and for most useful cases we can define an shortcut method on IQueryable directly
db.People.AggregateAsync(e => e.Count(o => o.Distinct()));  // Any aggregate can be immediately executed
db.People.CountAsync(o => o.Distinct()); // Nice-to-have shortcut for count.

// Because Window is IEnumerable, aggregates can naturally be called on windows.
// Although I can't figure how to make Window generic in this context, so it still has to be a different overload :(
// The upside is that less lambdas are required.
from p in db.People
let wnd = Sql.Window() // OVER (): everything in one partition
select wnd.Max(p.Age, a => a.Filter(p.Gender == 'M'))  // MAX(age) FILTER (WHERE gender = 'M') OVER (wnd)

// Windows can also be declared on the spot or extended 
from p in db.People
select new 
{
  oldest = Sql.Window(w => w.PartitionBy(p.Gender)).Max(p.Age),  // MAX(age) OVER (PARTITION BY gender)
  youngest = Sql.Window(wnd, w => w.PartitionBy(p.Gender)).Min(p.Age),  // MIN(age) OVER (wnd PARTITION BY gender)
}

// Windowed functions are defined as extensions of Window, so they can't be used without one:
from p in db.People
select Sql.Window(w => w.PartitionBy(p.Gender).OrderBy(p.Name, Nulls.First))
  .Rank()  // RANK() OVER (PARTITION BY gender ORDER BY name NULLS FIRST)

// Finally, my hypothetical string_agg example that no DB supports today:
Sql.Window(w => w.OrderBy(p.birthday.Month).Groups(1, 1))
  .StringAgg(p.name, ",", o => o.OrderBy(p.name))
  // STRING_AGG(name, ',' ORDER BY name) 
  // OVER (ORDER BY month(birthday) GROUPS BETWEEN 1 PRECEDING AND 1 FOLLOWING)
```

jods4:
Okay small tweak on my previous idea that I think I like better:
```cs
// Use `Over()` to be closer to actual SQL code, make it fluent
select Sql.Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date)
  .Sum(x.Amount, o => o.Filter(x.Amount > 0))
  
// Declared windows can be fluent too (less lambdas) and use Window() to remain closer to SQL
let wnd = Sql.Window().PartitionBy(x.Color)  // WINDOW wnd AS (PARTITION BY color)
// Basic usage doesn't really require `Over()` but I think I like it.
// The main issue _without_ `Over()` is that it allows calling `wnd` fluent builder and that's ambiguous here,
// esp. considering the next example.
select Sql.Over(wnd).Max(x.Price) // MAX(price) OVER (wnd) 

// Windows can be modified when they're used:
select Sql.Over(wnd)
  .PartitionBy(x.Year)
  .Max(x.Price)  // MAX(price) OVER (wnd PARTITION BY year)
  
// Window declarations can be based off previous windows
let wnd2 = Sql.Window(wnd).OrderBy(x.Age) // WINDOW wnd2 AS (wnd ORDER BY age)
```

sdanyliv:
@jods4, 

### Summary:

#### Comparison of Syntax:

```cs
Sql.Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date)
  .Sum(x.Amount)
```
vs
```cs
Sql.Functions.Sum(x.Amount, f => f
  .Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date))
```
---
```cs
Sql.Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date)
  .Sum(x.Amount, o => o.Distinct().Filter(x.Amount > 0))
```
vs
```cs
Sql.Functions.Sum(x.Amount, f => f.Distinct().Filter(x.Amount > 0)
  .Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date))
```

#### Aggregates:

Currently, only I have defined them:

```cs
var query = 
    from t in db.Some
    group t by 1
    select g.StringAggregate(x => x.Name, (x, f) => f.OrderBy(x.Name).Over().PartitionBy(x.Department))
```

---

### Your Syntax:

#### Advantages:
1. No need for a dedicated `DefineWindow` function.
2. In most cases (90%), window functions do not require a lambda.

#### Disadvantages:
1. Translation becomes slightly more complex, though still feasible.
2. Function name appears at the end, which may affect readability.
3. Window functions have a distinct syntax compared to other functions.
4. Part of the window definition is outside the function, while some parts are inside a lambda, requiring extra effort.
5. Builder interfaces leak into the LINQ query.
6. Defining a frame boundary results in invalid syntax for `ROW_NUMBER`.

---

### Proposed Alternative:

#### Advantages:
1. Provides a unified syntax for defining both window and non-window functions, ensuring consistency even for aggregates.
2. Maintains a natural order of function definitions.
3. Simplifies translation and error handling.
4. Stronger compiler control over function definitions.
5. No leakage of builder interfaces.

#### Disadvantages:
1. Slightly more verbose due to the use of lambdas and parentheses.
2. Requires `DefineWindow` for reusable window definitions, though this could be beneficial if named definitions are needed.
3. ..... please define

I will define later list of API examples for all window functions and we can choose better syntax for definition.

jods4:
@sdanyliv some thoughts on pros/cons:

For my proposal I would add advantages:
3. Aggregate lambda is identical whether used on a groupby or on a window.
4. Concept of window (which set of rows are passed to function) is clearly separated from configuration of function (receiver) itself.
5. Function name at the end is consistent with groupby, and C#/LINQ in general (which is the opposite of your disadvantage 2, not quite agree on this one).
6. All windowed functions can be discovered as extensions of `Window`, incl. user-defined ones. 
7. Formatting tools such as CSharpier will format better:
```cs
// Your last example as formatted by Prettier:
// My proposal
Sql.Over()
  .PartitionBy(x.Country)
  .OrderBy(x.Date)
  .Sum(x.Amount, o => o.Distinct().Filter(x.Amount > 0))

// This PR
Sql.Functions.Sum(
  x.Amount, 
  f => 
    f.Distinct()
      .Filter(x.Amount > 0)
      .Over()
      .PartitionBy(x.Country)
      .OrderBy(x.Date)
)
```

Regarding disadvantages:
2. Debatable, I don't quite agree: see advantage 5 above.
3. Can you clarify what you mean? I don't get it.
4. I don't get this one either. The SQL window is defined in `OVER ()` and that part of the SQL spec is fully outside the function. What part of the window is "partly inside" and "requires extra effort"?
7. I am not sure why having a frame into a window makes `ROW_NUMBER` invalid? PostgreSQL 17 runs the following just fine:
`select row_number() over (order by 1 rows 1 preceding)`

About your proposal, regarding advantages:
(1.) I don't understand how having different options in the same aggregate function, depending on whether it's called on a group or a window, is "unified". In fact, I mention as advantage 3. above that I think my proposal is more consistent on this aspect. 
(4.) Do you have a concrete example? I am not aware of any aggregate-like functions that imposes requirements on its window. See my test of `ROW_NUMBER` above. (Conceptually / based on SQL standard: I have mentioned in this thread that I know no RDBMS that supports framing combined with ordered windowed functions -- e.g. `string_agg` -- but that's in theory possible and could be implemented one day.)

Regarding disadvantages:
2. If we want an explicitly named function with my syntax, it's as easy as `let w = Window("name")`, no real difference here.
3. Windowed functions are harder to find, esp. if not built-in. Are they on `Sql`, `Sql.Window`, `Sql.Functions`, what about provider-specific or user-defined ones?

sdanyliv:
@jods4, sorry, but is hard to follow your answer, add citations to numbers. Show more examples (full examples) than words. Define range in your proposed syntax. For example LAG, LEAD do not allow range definition.

jods4:
@sdanyliv no worry, there's a lot to process! I have edited my post above, hopefully it's easier to read now.

sdanyliv:
Postgres 17 introduces new functionality. Oracle and other database providers currently don't support frame definitions in these window functions.

The syntax you've provided for the Frame is the simplest case, defining unbounded ranges clearly:
```sql
OVER (ORDER BY t ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) EXCLUDE CURRENT ROW
```

Consider also this example below, which is syntactically valid from a compiler perspective but difficult to understand clearly:
```cs
var query = 
    from t in db.Some
    from ss in (
        from ss in db.Some2
        let o = Sql.Over()
        let partition = o.PartitionBy(ss.Some2Field)
        let ordering = partition.OrderBy(t.SomeField)
        select ordering
    )
    select new
    {
        sum = ss.RowsRange
                .Between.CurrentRow.And.Unbounded
                .Exclude.CurrentRow
                .Sum(x => x.SomeValue)
    };
```
Just want to mention that I don't like that we cannot control this process. Something may work, but something will fail - additional code, additional tests.

sdanyliv:
> Sql.Over() starts a window expression that MUST end with an aggregate-like function, i.e. the final expression value cannot be IOver (or similar).

I'm not sure about the compiler feature you're referring to, but the proposed implementation won't compile.

jods4:
I think that's what I meant: linq2db builder should refuse to compile invalid `Over()` expressions, such as the example you provide with `let` and subqueries.

sdanyliv:
@jods4, check updated comment.

I understand your point, but I'm still hesitant - I'm strongly inclined to prevent compiling bad functions. Let me first create a table of the functions we plan to implement, and then we'll discuss further.

igor-tkachev:
In SQL, the OVER clause is the key element that defines a Window Function, as specified in the official SQL documentation:
[Microsoft SQL Server - OVER Clause](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-over-clause-transact-sql?view=sql-server-ver16).

Based on this, our API should reflect this fundamental structure. So, our Window Function API should always start with `Sql.Window` and end with `.Over(...)`.

Everything else - Ranking Functions, Aggregate Functions, Analytic Functions - should be placed between these two constructs. This way, we can clearly separate regular Aggregate Functions from Window Aggregate Functions, avoiding confusion in usage.

According to the specification, the OVER clause consists of:

```sql
OVER ( 
       [ <PARTITION BY clause> ]
       [ <ORDER BY clause> ]
       [ <ROW or RANGE clause> ]
      )
```

This means that an API like:

```c#
Over().PartitionBy().OrderBy()
```

would be problematic because the entire construct must return a column value, and there's no clear way to bind the method chain to anything meaningful. Alternatively, we could make the chain end with something like:

```c#
Over().PartitionBy().OrderBy().Value()
```

but that introduces unnecessary complexity (say hello to Builder pattern), which may not be ideal in this case.

Given these constraints, we have two viable options:

Lambda-based configuration:

```c#
Over(x => x.PartitionBy(...).OrderBy(...).RowOrRange(...))
```

This approach is concise and consistent with other fluent APIs. It avoids ambiguity in interpreting Over(x => x), ensuring clarity in the function's purpose.

Multiple lambda overloads for PartitionBy, OrderBy, etc:

```c#
Over(x => x, x => x)
```

This would require multiple overloads to cover different cases. But this one introduces ambiguity:

```c#
Over(x => x) // Is this PartitionBy or OrderBy?
```

Resolving this ambiguity would either require additional overloads or explicit method naming, both of which complicate API usage.

Given the above considerations, the most reasonable and practical API design would be:

```c#
Sql.Window.[SqlFunction()].Over(x => x.[PartitionBy(...)].[OrderBy(...)].[RowOrRange(...)])
```

sdanyliv:
@igor-tkachev, let's look at corner cases
```cs
// SUM(DISTINCT some) OVER (...)
Sql.Window.Sum(t.Value, f => f.Distinct).Over(f => f.[PartitionBy(...)].[OrderBy(...)].[RowOrRange(...)])
```
With one lambda for everything
```cs
Sql.Window.Sum(t.Value, f => f.Distinct.Over().[PartitionBy(...)].[OrderBy(...)].[RowOrRange(...)])
```
With window definition
```cs
let wnd = Sql.Window.DefineWindow(f => f.[PartitionBy(...)].[OrderBy(...)].[RowOrRange(...)]);
sum = Sql.Window.Sum(t.Value, f => f.Distinct.Over(wnd))
```
Can you catch these cases?

igor-tkachev:
@sdanyliv 

A single lambda looks like a workable option but is less intuitive. As mentioned earlier, the OVER clause is the key element in defining Window Functions, so it makes sense to highlight it explicitly in the API.

There are already inconsistencies in your example:

```c#
f => f.Distinct.Over()
```

Here, `Distinct` has no direct relation to `Over`, so they should not be dependent on each other in the API. `Distinct` should be isolated at the level of `Sum`, while `Over` should remain a standalone method.

Keeping `Over` as an independent construct prevents such ambiguities and makes the API more intuitive and aligned with the SQL specification.

jods4:
Two funsies I encountered at work today, which should be accounted for in the syntax (those at least work in Oracle).

Postgre doesn't support this, but SQL standard defines an option `RESPECT NULLS` and `IGNORE NULLS` for lead, lag, first_value, last_value, and nth_value. This should be after the function but Oracle supports this syntax both inside (proprietary) and outside (standard). Example: `last_value(order_id) ignore nulls over (order by order_date)`.

I believe `FIRST` and `LAST` (aka `KEEP`) are Oracle-specific, see https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/FIRST.html. They let you apply an aggregate function to the set of values that ranks first or last (often a single row, unless you have ties). They are very similar in nature to `OVER` framing, but: 
1. You can use `FIRST` and `LAST` with or without `OVER` (usable in a normal aggregate / group by as well).
2. What it achieves actually can't be done with `OVER` frames (frames are always defined around current row, you can't say "absolute first")

It can be applied to MIN, MAX, SUM, AVG, COUNT, VARIANCE, or STDDEV.
Example: `min(id) keep (dense_rank last order by salary)`

jods4:
IMHO the first `[RESPECT | IGNORE] NULLS` should a fluent option when configuring `LEAD` and co.

For the second, I would be tempted to create a `.KeepFirstBy[Desc]().ThenBy[Desc]().Min(val)` on `IEnumerable` (for aggregates / group by) and on `Over()`. Still with the ideas that aggregate function should be last in C#, an example would be:
```csharp
Sql.Over()
  .PartitionBy(country)
  .KeepLastBy(year, Order.NullsLast).ThenByDesc(population)
  .Min(gdp)
  
// MIN(gdp) KEEP (DENSE_RANK LAST ORDER BY year NULLS LAST, population DESC) OVER (PARTITION BY country)
```

jods4:
@sdanyliv didn't we go over this? 
I agree that function-specific options should be inside a fluent option builder inside the call, like `.Count(id, x => x.Distinct())`.
What I would put in front is:
- `IEnumerable` for regular GROUP BY aggregates (nobody argued with this one)
- `Over()` for analytical windows
- `KeepFirst()` / `KeepLast()` for consistency, as they apply to all aggregates and windows. I actually just noticed this is how it works in linq2db v5 BTW.

jods4:
If you want a real-world complex query to compare syntax and results, I coded this today:

```cs
  db.PartialStatementFile
    .Merge()
    .Using(
      PendingFiles()
        .Where(x => x.LastPageId == null)
        .Select(x => new
        {
          x.Id,
          LastPageId = Sql.Ext.Max(x.Id)
            .KeepLast().OrderBy(x.IsLastPage)
            .Over().PartitionBy(x.BankSwift, x.AccountNumber, x.StatementId)
            .ToValue(),
          HasLastPage = Sql.Ext.Max(x.IsLastPage)
            .Over().PartitionBy(x.BankSwift, x.AccountNumber, x.StatementId)
            .ToValue(),
          MaxPage = Sql.Ext.Max(x.PageNumber)
            .Over().PartitionBy(x.BankSwift, x.AccountNumber, x.StatementId)
            .ToValue(),
          PageCount = Sql.Ext.Count()
            .Over().PartitionBy(x.BankSwift, x.AccountNumber, x.StatementId)
            .ToValue(),
        })
        .Where(x => x.HasLastPage && x.PageCount >= x.MaxPage)
    )
    .On((stmt, g) => stmt.Id == g.Id)
    .UpdateWhenMatched((stmt, g) => new()
    {
      LastPageId = g.LastPageId
    })
    .Merge();
```

Resulting query (slightly reformatted):
```sql
MERGE INTO PartialStatementFile Target
USING (
  SELECT
    x.Id,
    x.LastPageId
  FROM
    (
      SELECT
        f.Id,
        MAX(f.Id) KEEP (DENSE_RANK LAST ORDER BY f.IsLastPage) OVER(PARTITION BY f.BankSwift, f.AccountNumber, f.StatementId) as LastPageId
        MAX(f.IsLastPage) OVER(PARTITION BY f.BankSwift, f.AccountNumber, f.StatementId) as HasLastPage,
        MAX(f.PageNumber) OVER(PARTITION BY f.BankSwift, f.AccountNumber, f.StatementId) as MaxPage,
        COUNT(*) OVER(PARTITION BY f.BankSwift, f.AccountNumber, f.StatementId) as PageCount,
      FROM
        PartialStatementFile f
      WHERE
        Nvl2(f.MergeFileId, NULL, f.Received) < :before AND
        f.LastPageId IS NULL
  ) x
  WHERE
    x.HasLastPage = 1 AND x.PageCount >= x.MaxPage
) Source
ON (Target.Id = Source.Id)

WHEN MATCHED THEN UPDATE SET
  Target.LastPageId = Source.LastPageId
```

jods4:
Feature request based on example above: it would be great in RDBMS that don't support named WINDOW yet, if linq2db could emulate the feature, simply by repeating the OVER(..) definition.

You can see in my complex example above that I wrote multiple times
`.Over().PartitionBy(x.BankSwift, x.AccountNumber, x.StatementId)`

Even though Oracle doesn't support `WINDOW` yet, I would love to just do it once and then `.Over(wnd)`, code would be more readable.

jods4:
> we can't prevent the user from defining an incorrect window function.

What is an example of an incorrect window function?

> While chaining looks elegant, it comes with several drawbacks described several times.

I agree on some drawbacks, which is why I now think that not everything must be chained fluently.
For instance, although it's more nesting, I lean towards a builder inside `.Over()` instead of fluent syntax. For several reasons but mostly (1) to avoid misuse of windows that you've shown before, and (2) to disambiguate what belongs to `OVER` and what another options (I'm getting a feel there are many...).
`Sql.Over(w => w.PartitionBy(x.Color).OrderBy(x.Size)).Max(x.Cost)`

sdanyliv:
Superceded by https://github.com/linq2db/linq2db/pull/5468

PR #4852, opened 2025-02-28
ORDER BY is removed from subquery even when TOP or LIMIT/OFFSET are used

igor-tkachev:
See test.

`group by` itself has `order by`.

```sql
DECLARE @take Int -- Int32
SET     @take = 2

/* GroupBy */
SELECT TOP (@take)
	[t1].[Key_1],
	[t1].[Max_1]
FROM
	(
		SELECT
			MAX([g_1].[ChildID]) as [Max_1],
			[g_1].[ParentID] as [Key_1]
		FROM
			[Child] [g_1]
		GROUP BY
			[g_1].[ParentID]
	) [t1]
ORDER BY
	[t1].[Max_1] DESC
```

If it is used as subquery `order by` is lost.

```sql
/* GroupBy
Query */
SELECT
	[t].[ParentID],
	[t].[ChildID]
FROM
	[Child] [t]
WHERE
	[t].[ParentID] IN (
		SELECT
			[t1].[ParentID]
		FROM
			(
				SELECT TOP (2)
					[x].[ParentID]
				FROM
					[Child] [x]
				GROUP BY
					[x].[ParentID]
			) [t1]
	)
```

### DIFF
--- a/Source/LinqToDB/DataProvider/DataProviderBase.cs
+++ b/Source/LinqToDB/DataProvider/DataProviderBase.cs
@@ -70,6 +70,7 @@ namespace LinqToDB.DataProvider
 				RowConstructorSupport                = RowFeature.None,
 				IsWindowFunctionsSupported           = true,
 				IsDerivedTableOrderBySupported       = true,
+				IsOrderByAggregateFunctionSupported  = true,
 			};
 
 			SetField<DbDataReader, bool>    ((r,i) => r.GetBoolean (i));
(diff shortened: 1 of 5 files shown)


jods4:
I did not immediately get what is wrong.

The point here is not "grouping" but the `TOP 2`.
Usually `ORDER BY` is not allowed in subqueries, such as `IN` from this test case.
It's because there's a `TOP 2` that's it's allowed.

For example, here's what MS SQL says about it: 
https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16
> The ORDER BY clause isn't valid in views, inline functions, derived tables, and subqueries, unless either the TOP or OFFSET and FETCH clauses are also specified.

MaceWindu (approved):

PR #4855, opened 2025-03-03
TempTable.Insert method is broken.

igor-tkachev:
The following code stopped working:

```c#
class Table1
{
	public int ID;
}

class Table2
{
	public int       ID;
	public DateTime? Date;
}

[Test]
public void InsertTest([DataSources(false, TestProvName.AllClickHouse)] string context)
{
	using var t1 = db.CreateLocalTable<Table1>([new Table1 { ID = 10 }]);
	using var t2 = db.CreateLocalTable<Table2>();

	t2.Insert(from t in t1 select new Table2 { ID = t.ID });

	var result = t2.ToList();

	Assert.Multiple(() =>
	{
		Assert.That(result,         Has.Count.EqualTo(1));
		Assert.That(result[0].ID,   Is.EqualTo(10));
		Assert.That(result[0].Date, Is.Null);
	});
}
```

Works in 5.4.1.

### DIFF
--- a/Source/LinqToDB/TempTable.cs
+++ b/Source/LinqToDB/TempTable.cs
@@ -576,9 +576,7 @@ namespace LinqToDB
 		/// <returns>Number of records, inserted into table.</returns>
 		public long Insert(IQueryable<T> items)
 		{
-			var l = GenerateInsertSetter(items ?? throw new ArgumentNullException(nameof(items)));
-
-			var count = items.Insert(_table, l);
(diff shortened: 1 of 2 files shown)


viceroypenguin (approved):
Why did we have it as a separate setter in the first place?

igor-tkachev (commented):
I like this kind of fixes :)

MaceWindu (approved):

PR #4861, opened 2025-03-08
UseOptions

igor-tkachev:
A new method, `UseOptions`, has been introduced to allow temporary changes to `IDataContext` options, with automatic restoration afterward.

`IDataContext` options are immutable, meaning they cannot be changed once set. With this feature, it is now possible to override options within a specific execution context while ensuring they revert to their original state when no longer needed.

Several overloads are available to fit different use cases:

```c#
using (db.UseOptions(o => o
	.WithOptions<LinqOptions>    (co => co with { OptimizeJoins = false })
	.WithOptions<BulkCopyOptions>(bo => bo with { BulkCopyType = BulkCopyType.RowByRow })))
{
	// ...
}
```

```c#
using (db.UseOptions<DataContextOptions>(o => o with { CommandTimeout = 45 }))
{
	// ...
}
```

```c#
using (db.UseLinqOptions(o => o with { CompareNulls = param }))
{
	// ...
}
```

### DIFF
--- a/Source/LinqToDB/Common/IOptionSet.cs
+++ b/Source/LinqToDB/Common/IOptionSet.cs
@@ -1,4 +1,6 @@
-﻿using LinqToDB.Common.Internal;
+﻿using System;
+
+using LinqToDB.Common.Internal;
 
 namespace LinqToDB.Common
 {
@@ -13,5 +15,9 @@ namespace LinqToDB.Common
 	/// </summary>
 	public interface IOptionSet : IConfigurationID
 	{
+		/// <summary>
+		/// Gets the default options.
+		/// </summary>
+		IOptionSet Default { get; }
 	}
 }
--- a/Source/LinqToDB/Common/IReapplicable.cs
+++ b/Source/LinqToDB/Common/IReapplicable.cs
@@ -1,0 +1,9 @@
+﻿using System;
+
+namespace LinqToDB.Common
+{
+	interface IReapplicable<T>
+	{
+		Action? Apply(T obj, object? previousObject);
+	}
+}
--- a/Source/LinqToDB/Common/OptionsContainer.cs
+++ b/Source/LinqToDB/Common/OptionsContainer.cs
@@ -126,5 +126,39 @@ namespace LinqToDB.Common
 					if (item is IApplicable<TA> a)
 						a.Apply(obj);
 		}
+
+		protected Action? Reapply<TA>(TA obj, OptionsContainer<T> previousContainer)
+		{
+			if (_sets == null && previousContainer._sets == null)
+				return null;
+
+			Action? actions = null;
+
+			if (_sets != null)
+				foreach (var item in _sets)
+					if (item.Value is IReapplicable<TA> a)
+					{
+						IOptionSet? previousItem = null;
+
+						previousContainer._sets?.TryGetValue(item.Key, out previousItem);
+
+						var action = a.Apply(obj, previousItem);
+						if (action != null)
+							actions += action;
+					}
+
+			if (previousContainer._sets != null)
+				foreach (var item in previousContainer._sets)
+				{
+					if (item.Value is IReapplicable<TA> a && _sets?.ContainsKey(item.Key) == false)
+					{
+						var action = ((IReapplicable<TA>)item.Value.Default).Apply(obj, a);
+						if (action != null)
+							actions += action;
+					}
+				}
+
+			return actions;
+		}
 	}
 }
--- a/Source/LinqToDB/Data/DataConnection.Configuration.cs
+++ b/Source/LinqToDB/Data/DataConnection.Configuration.cs
@@ -10,6 +10,8 @@ using LinqToDB.Common.Internal;
 using LinqToDB.Configuration;
 using LinqToDB.Data.RetryPolicy;
 using LinqToDB.DataProvider;
+using LinqToDB.Interceptors;
+using LinqToDB.Interceptors.Internal;
 
 namespace LinqToDB.Data
 {
@@ -649,9 +651,57 @@ namespace LinqToDB.Data
 				}
 			}
 
+			// For ConnectionOptions we reapply only mapping schema and connection interceptor.
+			// Connection string, configuration, and data provider are not reapplyable.
+			//
+			public static Action Reapply(DataConnection dataConnection, ConnectionOptions options, ConnectionOptions? previousOptions)
+			{
+				if (previousOptions?.ConnectionInterceptor != null)
+					dataConnection.RemoveInterceptor(previousOptions.ConnectionInterceptor);
+
+				if (options.ConnectionInterceptor != null)
+					dataConnection.AddInterceptor(options.ConnectionInterceptor);
+
+				var mappingSchema = dataConnection.MappingSchema;
+
+				dataConnection.MappingSchema = dataConnection.DataProvider.MappingSchema;
+
+				if (options.MappingSchema != null)
+				{
+					dataConnection.AddMappingSchema(options.MappingSchema);
+				}
+				else if (dataConnection.Options.LinqOptions.EnableContextSchemaEdit)
+				{
+					dataConnection.MappingSchema = new (dataConnection.MappingSchema);
+				}
+
+				return () =>
+				{
+					if (options.ConnectionInterceptor != null)
+						dataConnection.RemoveInterceptor(options.ConnectionInterceptor);
+
+					if (previousOptions?.ConnectionInterceptor != null)
+						dataConnection.AddInterceptor(previousOptions.ConnectionInterceptor);
+
+					dataConnection.MappingSchema = mappingSchema;
+				};
+			}
+
 			public static void Apply(DataConnection dataConnection, RetryPolicyOptions options)
 			{
 				dataConnection.RetryPolicy = options.RetryPolicy ?? options.Factory?.Invoke(dataConnection);
+			}
+
+			public static Action Reapply(DataConnection dataConnection, RetryPolicyOptions options, RetryPolicyOptions? previousOptions)
+			{
+				var retryPolicy = dataConnection.RetryPolicy;
+
+				Apply(dataConnection, options);
+
+				return () =>
+				{
+					dataConnection.RetryPolicy = retryPolicy;
+				};
 			}
 
 			public static void Apply(DataConnection dataConnection, DataContextOptions options)
@@ -663,11 +713,57 @@ namespace LinqToDB.Data
 						dataConnection.AddInterceptor(interceptor);
 			}
 
+			public static Action Reapply(DataConnection dataConnection, DataContextOptions options, DataContextOptions? previousOptions)
+			{
+				var commandTimeout = dataConnection._commandTimeout;
+
+				dataConnection.CommandTimeout = options.CommandTimeout ?? -1;
+
+				if (previousOptions?.Interceptors != null)
+					foreach (var interceptor in previousOptions.Interceptors)
+						dataConnection.RemoveInterceptor(interceptor);
+
+				if (options.Interceptors != null)
+					foreach (var interceptor in options.Interceptors)
+						dataConnection.AddInterceptor(interceptor);
+
+				return () =>
+				{
+					dataConnection.CommandTimeout = commandTimeout ?? -1;
+
+					if (options.Interceptors != null)
+						foreach (var interceptor in options.Interceptors)
+							dataConnection.RemoveInterceptor(interceptor);
+
+					if (previousOptions?.Interceptors != null)
+						foreach (var interceptor in previousOptions.Interceptors)
+							dataConnection.AddInterceptor(interceptor);
+				};
+			}
+
 			public static void Apply(DataConnection dataConnection, QueryTraceOptions options)
 			{
-				if (options.OnTrace    != null) dataConnection.OnTraceConnection        = options.OnTrace;
-				if (options.TraceLevel != null) dataConnection.TraceSwitchConnection    = new("DataConnection", "DataConnection trace switch") {Level = options.TraceLevel.Value};
-				if (options.WriteTrace != null) dataConnection.WriteTraceLineConnection = options.WriteTrace;
+				dataConnection.OnTraceConnection        = options.OnTrace    ?? DefaultOnTraceConnection;
+				dataConnection.WriteTraceLineConnection = options.WriteTrace ?? WriteTraceLine;
+				dataConnection._traceSwitchConnection   = options.TraceLevel != null
+					? new ("DataConnection", "DataConnection trace switch") { Level = options.TraceLevel.Value }
+					: null;
+			}
+
+			public static Action Reapply(DataConnection dataConnection, QueryTraceOptions options, QueryTraceOptions? previousOptions)
+			{
+				var onTraceConnection        = dataConnection.OnTraceConnection;
+				var writeTraceLineConnection = dataConnection.WriteTraceLineConnection;
+				var traceSwitchConnection    = dataConnection._traceSwitchConnection;
+
+				Apply(dataConnection, options);
+
+				return () =>
+				{
+					dataConnection.OnTraceConnection        = onTraceConnection;
+					dataConnection.WriteTraceLineConnection = writeTraceLineConnection;
+					dataConnection._traceSwitchConnection   = traceSwitchConnection;
+				};
 			}
 		}
 	}
--- a/Source/LinqToDB/Data/DataConnection.cs
+++ b/Source/LinqToDB/Data/DataConnection.cs
@@ -811,7 +811,7 @@ namespace LinqToDB.Data
 		/// Gets or sets command execution timeout in seconds.
 		/// Negative timeout value means that default timeout will be used.
 		/// 0 timeout value corresponds to infinite timeout.
-		/// By default timeout is not set and default value for current provider used.
+		/// By default, timeout is not set and default value for current provider used.
 		/// </summary>
 		public  int   CommandTimeout
 		{
@@ -820,13 +820,16 @@ namespace LinqToDB.Data
 			{
 				if (value < 0)
 				{
-					// to reset to default timeout we dispose command because as command has no reset timeout API
-					_commandTimeout = null;
-					// TODO: that's not good - user is not aware that he can trigger blocking operation
-					// we should postpone disposal till command used (or redesign CommandTimeout to methods)
-					DisposeCommand();
+					if (_commandTimeout is not null)
+					{
+						// to reset to default timeout we dispose command because as command has no reset timeout API
+						_commandTimeout = null;
+						// TODO: that's not good - user is not aware that he can trigger blocking operation
+						// we should postpone disposal till command used (or redesign CommandTimeout to methods)
+						DisposeCommand();
+					}
 				}
-				else
+				else if (_commandTimeout != value)
 				{
 					_commandTimeout = value;
 					if (_command != null)
@@ -1539,5 +1542,39 @@ namespace LinqToDB.Data
 		}
 
 		IServiceProvider IInfrastructure<IServiceProvider>.Instance => ((IInfrastructure<IServiceProvider>)DataProvider).Instance;
+
+		/// <inheritdoc cref="IDataContext.UseOptions"/>
+		public IDisposable? UseOptions(Func<DataOptions,DataOptions> optionsSetter)
+		{
+			var prevOptions = Options;
+			var newOptions  = optionsSetter(Options) ?? throw new ArgumentNullException(nameof(optionsSetter));
+
+			if (((IConfigurationID)prevOptions).ConfigurationID == ((IConfigurationID)newOptions).ConfigurationID)
+				return null;
+
+			var configurationID = _configurationID;
+			var msID            = _msID;
+
+			Options          = newOptions;
+			_configurationID = null;
+			_msID            = 0;
+
+			var action = Options.Reapply(this, prevOptions);
+
+			action += () =>
+			{
+				Options          = prevOptions;
+
+#if DEBUG
+				_configurationID = null;
+				_msID            = 0;
+#else
+				_configurationID = configurationID;
+				_msID            = msID;
+#endif
+			};
+
+			return new DisposableAction(action);
+		}
 	}
 }
--- a/Source/LinqToDB/DataContext.cs
+++ b/Source/LinqToDB/DataContext.cs
@@ -658,6 +658,42 @@ namespace LinqToDB
 				}
 			}
 
+			// For ConnectionOptions we reapply only mapping schema and connection interceptor.
+			// Connection string, configuration, and data provider are not reapplyable.
+			//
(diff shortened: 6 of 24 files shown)


igor-tkachev:
> I don't like this API. UseOptions may change data provider on the fly and probably has a lot of other sideeffects. If you need some specific feature, we can discuss.

I get the concern, but `UseOptions` doesn’t allow changing everything on the fly. Things like switching the data provider or other options that fundamentally change `IDataContext` won’t have any effect. The main goal here is to tweak settings that affect SQL generation and execution modes within the current context. So it’s more about flexibility in query behavior rather than altering core `IDataContext` functionality.

sdanyliv:
I wouldn't introduce it this way. Instead, you can use the Builder pattern to restrict modifications to only the allowed options. While direct `options` passing may seem flexible, it often leads to maintainability issues down the road. The API should be self-descriptive.

igor-tkachev:
Introducing a separate Builder pattern just to restrict modifications would add unnecessary complexity. Users will need to learn yet another part of an already feature-rich API, making it harder to adopt. Instead, we can rely on clear documentation and enforce restrictions at runtime by throwing exceptions when unsupported options are modified.

sdanyliv:
We can't ensure clear documentation when additional options are introduced. However, users might discover undocumented features that work and start relying on them. Later, if we make changes, everything undocumented could stop working. I want to prevent users from making mistakes due to a lack of documentation. Additionally, a strictly limited builder would be more precise than simply modifying the options class.

Exceptions are useful, but why not prevent mistakes entirely by designing the builder in a way that makes them impossible? If something isn’t allowed, it simply won’t compile—this is far better than letting users try unsupported options and fail at runtime.

igor-tkachev:
Our API is already complex enough, and we shouldn’t make it even harder to use. The proposed `UseOptions` method follows the existing options-setting API, so users don’t have to learn something entirely new.

We already have a Builder-based API - Merge API, and to be honest, I can't use it without constantly referring to the documentation, examples, or even the source code. All those "_prevent mistakes entirely by designing the builder_" arguments don’t help in practice. Complex APIs that builders tend to become are a bad idea. Reusing the existing, familiar API is a good one.

As for user mistakes, in the latest update, I’ve added the necessary checks to ensure that invalid modifications throw an exception. I wouldn’t have even done that, but maybe this will give you more peace of mind :)

igor-tkachev:
> I find that interesting, because I find the current version of the Merge API to be incredibly intuitive. The intellisense based on limited types means that the next available method(s) are immediately available and obvious to select which one should be used.

Oh, that’s probably because I personally took part in refining the Merge API to make it as polished as it is! :)

But that doesn’t mean it’s simple. Like any API built using the Builder pattern, it comes with an additional learning curve, is often inconsistent with existing APIs, and isn’t always as intuitive. It also adds complexity to an already feature-rich library API.

Sometimes, a Builder is necessary - Merge API is a good example. It’s complex, but I don’t see any alternative for it. However, making the entire library follow this pattern would be a huge mistake.

In the case of `UseOptions`, there’s simply no need for this approach. The existing options API is already familiar to users, and we don’t gain anything by introducing an extra API. It would just make things harder for no real benefit.

sdanyliv:
Changes to options should be strictly prohibited. I’m not happy with how options have been handled, and I want to restrict new parts until proper refactoring is done. If that's not possible, then it's better to avoid introducing `UseOptions` altogether.

viceroypenguin:
I would prefer @sdanyliv's proposal for how to build the new options, yes.

igor-tkachev:
Added additional tests to cover all unwanted option changes.

MaceWindu:
schema id logging change should go to separate pr or reverted as it makes impossible to review baselines

Source/LinqToDB/Common/IOptionSet.cs
```diff
@@ -13,5 +15,9 @@ namespace LinqToDB.Common
 	/// </summary>
 	public interface IOptionSet : IConfigurationID
 	{
+		/// <summary>
+		/// Gets the default options.
+		/// </summary>
+		IOptionSet Default { get; }
```
MaceWindu, line 21: I've checked `Default` implementation and use and see a lot of inconsistency around it.
It returns either `private empty`, `public empty`, `public default` or options from `Common.Configuration`: would be nice if we can rid of empty/default duplication here and have only `public Default` (backed either by `Common.Configuration` or `_empty` value) including options init in DataOptions (lines 70-75, especially taking into account that some code there reference to `Common.Configuration` which is implementation detail of specific option set and shouldn't be used here)

Source/LinqToDB/Common/IReapplicable.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace LinqToDB.Common
+{
+	interface IReapplicable<T>
+	{
+		Action? Apply(T obj, object? previousObject);
```
MaceWindu: It looks like we can change type from `object` to `IOptionSet` if we inherit interface from `IOptionSet`

Source/LinqToDB/Common/Internal/IdentifierBuilder.cs
```diff
@@ -84,8 +83,19 @@ public IdentifierBuilder Add(object? data)
 
 		public IdentifierBuilder Add(Delegate? data)
 		{
-			_sb.Value.Append(CultureInfo.InvariantCulture, $".{data?.Method}");
+			_sb.Value
+				.Append('.')
+				.Append(GetObjectID(data?.Method))
+				;
```
MaceWindu: Add(GetObjectID(data?.Method))

Source/LinqToDB/Common/Internal/IdentifierBuilder.cs
```diff
@@ -84,8 +83,19 @@ public IdentifierBuilder Add(object? data)
...
+			return this;
+		}
 
+		public IdentifierBuilder Add(Type? data)
+		{
+			_sb.Value
+				.Append('.')
+				.Append(GetObjectID(data))
+				;
 			return this;
```
MaceWindu: Add(GetObjectID(data))

MaceWindu: and maybe in other places where transform data? Like Add(object?)

Source/LinqToDB/Common/OptionsContainer.cs
```diff
@@ -126,5 +126,51 @@ protected void Apply<TA>(TA obj)
...
+					{
+						IOptionSet? previousItem = null;
+
+						previousContainer._sets?.TryGetValue(item.Key, out previousItem);
+
+						if (!ReferenceEquals(a, previousItem))
+						{
+							var action = a.Apply(obj, previousItem);
+							if (action != null)
+								actions += action;
```
MaceWindu: could be simplified to
`actions += a.Apply(obj, previousItem);`
also below and DataOptions Reapply methods

Source/LinqToDB/Data/DataConnection.Configuration.cs
```diff
@@ -533,59 +533,79 @@ public static void Apply(DataConnection dataConnection, ConnectionOptions option
...
+						throw new LinqToDBException("DataProvider was not specified");
+					}
+					case (_,               _,                   {} provider, _,               {} connection, _,              _) :
+					{
+						SetConnection(provider, connection);
+						break;
+					}
+					case (_,               _,                   _,           {} providerName, {} connection, _,              _) :
+					{
+						var provider = GetDataProviderEx(providerName, connection.ConnectionString);
```
MaceWindu: it is not a good idea to read ConnectionString property from connection. Many providers mask credentials in returned string, which will lead to strange errors. We should treat this case as no connection string provided

Source/LinqToDB/Data/DataConnection.Configuration.cs
```diff
@@ -533,59 +533,79 @@ public static void Apply(DataConnection dataConnection, ConnectionOptions option
...
+						break;
+					}
+					case (_,               _,                   {} provider, _,               _,             {} transaction, _) :
+					{
+						SetTransaction(provider, transaction);
+						break;
+					}
+					case (_,               _,                   _,           {} providerName, _,             {} transaction, _) :
+					{
+						var provider = GetDataProviderEx(providerName, transaction.Connection?.ConnectionString ?? throw new LinqToDBException("Transaction.Connection is null"));
```
MaceWindu: same here, we shouldn't access ConnectionString property on connection

Source/LinqToDB/Data/DataConnection.Configuration.cs
```diff
@@ -649,6 +674,89 @@ IAsyncDbConnection WrapConnection(DbConnection connection)
...
+				}
+
+			}
+
+			public static Action? Reapply(DataConnection dataConnection, ConnectionOptions options, ConnectionOptions? previousOptions)
+			{
+				// For ConnectionOptions we reapply only mapping schema and connection interceptor.
+				// Connection string, configuration, data provider, etc. are not reapplyable.
+				//
+				if (options.ConfigurationString       != previousOptions?.ConfigurationString)       throw new LinqToDBException("Configuration string cannot be changed.");
```
MaceWindu: Maybe use something like `"Option 'NAME' cannot be changed for context dynamically."` or something like that? Context of change is not clear from message

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -770,6 +770,7 @@ public Action<string,string,TraceLevel> WriteTraceLineConnection
 		/// Gets underlying database connection, used by current connection object, or opens new.
 		/// </summary>
 		// TODO: Remove in v7
+		// What is the point to remove this API?
```
MaceWindu, line 773: It violates expected behavior for property getters to not have side effects by opening connection when it is not opened yet. This resulted in issues for both users and us before when it was used without taking it into account. Instead if it we introduce two separate APIs to get existing connection with and without creation.

MaceWindu (commented):
part 3

Source/LinqToDB/DataContext.cs
```diff
@@ -778,6 +784,63 @@ public static void Apply(DataContext dataContext, ConnectionOptions options)
 				}
 			}
 
+			public static Action? Reapply(DataContext dataContext, ConnectionOptions options, ConnectionOptions? previousOptions)
+			{
+				// For ConnectionOptions we reapply only mapping schema and connection interceptor.
+				// Connection string, configuration, data provider, etc. are not reapplyable.
+				//
+				if (options.ConfigurationString       != previousOptions?.ConfigurationString)       throw new LinqToDBException("Configuration string cannot be changed.");
```
MaceWindu: same suggestion to improve message text

Source/LinqToDB/Data/DataConnection.Configuration.cs
```diff
@@ -667,15 +786,74 @@ public static void Apply(DataConnection dataConnection, DataContextOptions optio
...
 
+			public static Action? Reapply(DataConnection dataConnection, DataContextOptions options, DataContextOptions? previousOptions)
+			{
+				Action? action = null;
+
+				if (options.CommandTimeout != previousOptions?.CommandTimeout)
+				{
+					var commandTimeout = dataConnection._commandTimeout;
+
+					dataConnection.CommandTimeout = options.CommandTimeout ?? -1;
```
MaceWindu: CommandTimeout behavior changed and will throw exception for negative values. Use  ResetCommandTimeout method for timeout resetting

Source/LinqToDB/DataContext.cs
```diff
@@ -786,11 +849,94 @@ public static void Apply(DataContext dataContext, DataContextOptions options)
...
+
+			public static Action? Reapply(DataContext dataContext, DataContextOptions options, DataContextOptions? previousOptions)
+			{
+				Action? action = null;
+
+				if (options.CommandTimeout != previousOptions?.CommandTimeout)
+				{
+					var commandTimeout = dataContext._commandTimeout;
+
+					dataContext.CommandTimeout = options.CommandTimeout ?? -1;
```
MaceWindu: CommandTimeout behavior changed and will throw exception for negative values. Use  ResetCommandTimeout method for timeout resetting

Source/LinqToDB/DataProvider/SqlServer/SqlServerMappingSchema.cs
```diff
@@ -771,21 +771,21 @@ public SqlServer2022MappingSchema() : base(ProviderName.SqlServer2022, Instance)
...
+		public sealed class SqlServer2019MappingSchemaSystem   () : LockedMappingSchema(ProviderName.SqlServer2019 + SDS, new[] { Instance2019, SqlServerProviderAdapter.GetInstance(SqlServerProvider.SystemDataSqlClient   ).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2019MappingSchemaMicrosoft() : LockedMappingSchema(ProviderName.SqlServer2019 + MDS, new[] { Instance2019, SqlServerProviderAdapter.GetInstance(SqlServerProvider.MicrosoftDataSqlClient).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2022MappingSchemaSystem   () : LockedMappingSchema(ProviderName.SqlServer2022 + SDS, new[] { Instance2022, SqlServerProviderAdapter.GetInstance(SqlServerProvider.SystemDataSqlClient   ).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2022MappingSchemaMicrosoft() : LockedMappingSchema(ProviderName.SqlServer2022 + MDS, new[] { SqlServerProviderAdapter.GetInstance(SqlServerProvider.MicrosoftDataSqlClient).MappingSchema, Instance2022 }.Where(ms => ms                != null).ToArray()!);
```
MaceWindu: order is not same as for other chemas

Source/LinqToDB/DataProvider/SqlServer/SqlServerMappingSchema.cs
```diff
@@ -771,21 +771,21 @@ public SqlServer2022MappingSchema() : base(ProviderName.SqlServer2022, Instance)
...
+		public sealed class SqlServer2008MappingSchemaSystem   () : LockedMappingSchema(ProviderName.SqlServer2008 + SDS, new[] { Instance2008, SqlServerProviderAdapter.GetInstance(SqlServerProvider.SystemDataSqlClient   ).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2008MappingSchemaMicrosoft() : LockedMappingSchema(ProviderName.SqlServer2008 + MDS, new[] { Instance2008, SqlServerProviderAdapter.GetInstance(SqlServerProvider.MicrosoftDataSqlClient).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2012MappingSchemaSystem   () : LockedMappingSchema(ProviderName.SqlServer2012 + SDS, new[] { Instance2012, SqlServerProviderAdapter.GetInstance(SqlServerProvider.SystemDataSqlClient   ).MappingSchema }.Where(ms => ms != null).ToArray()!);
+		public sealed class SqlServer2012MappingSchemaMicrosoft() : LockedMappingSchema(ProviderName.SqlServer2012 + MDS, new[] { Instance2012, SqlServerProviderAdapter.GetInstance(SqlServerProvider.MicrosoftDataSqlClient).MappingSchema }.Where(ms => ms != null).ToArray()!);
```
MaceWindu: If we deside to lower priority of adapter schema over provider schema - we should do it for all providers that have adapter schema

Source/LinqToDB/IDataContext.cs
```diff
@@ -127,10 +127,48 @@ public interface IDataContext : IConfigurationID, IDisposable, IAsyncDisposable
...
 		string?                       ConfigurationString         { get; }
 
+		/// <summary>
+		/// Sets new options for current data context.
+		/// <para>
+		/// <b>Implements the Disposable pattern, which must be used to properly restore previous options.</b>
+		/// </para>
+		/// </summary>
+		/// <remarks>
+		/// For ConnectionOptions we reapply only mapping schema and connection interceptor. Connection string, configuration, data provider, etc. are not reapplyable.
```
MaceWindu: maybe use "update" instead of "reapply" in help texts?

Source/LinqToDB/Mapping/LockedMappingSchemaInfo.cs
```diff
@@ -24,5 +28,27 @@ public override void ResetID()
 				throw new LinqToDBException($"MappingSchema '{_mappingSchema.GetType()}' is locked.");
 			base.ResetID();
 		}
+
+		public bool Equals(LockedMappingSchemaInfo? other)
+		{
+			if (other is null)                return false;
+			if (ReferenceEquals(this, other)) return true;
+
+			return _mappingSchema.GetType() == other._mappingSchema.GetType();
```
MaceWindu, line 37: is it correct equality?

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -140,7 +148,7 @@ public MappingSchema(string? configuration, params MappingSchema[]? schemas)
 				var i = 0;
 				var j = 0;
 
-				schemaList[schemaInfo] = i++;
+				schemaList[schemaInfo] = -100;
```
MaceWindu: what is this magic number?

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -1295,14 +1309,19 @@ protected internal virtual int GenerateID()
...
 			lock (_syncRoot)
 			{
 				foreach (var s in Schemas)
 					idBuilder.Add(s.ConfigurationID);
 
-				var reader = Schemas[0].MetadataReader;
-				if (reader != null)
-					idBuilder.Add(reader.GetObjectID());
+//				var reader = Schemas[0].MetadataReader;
+//				if (reader != null)
```
MaceWindu: remove commented code?

Source/LinqToDB/Remote/RemoteDataContextBase.cs
```diff
@@ -635,12 +641,132 @@ public static void Apply(RemoteDataContextBase dataContext, ConnectionOptions op
 				}
 			}
 
+			public static Action? Reapply(RemoteDataContextBase dataContext, ConnectionOptions options, ConnectionOptions? previousOptions)
+			{
+				// For ConnectionOptions we reapply only mapping schema and connection interceptor.
+				// Connection string, configuration, data provider, etc. are not reapplyable.
+				//
+				if (options.ConfigurationString       != previousOptions?.ConfigurationString)       throw new LinqToDBException("Configuration string cannot be changed.");
```
MaceWindu: same suggestion to improve error message

Source/LinqToDB/Remote/RemoteDataContextBase.cs
```diff
@@ -635,12 +641,132 @@ public static void Apply(RemoteDataContextBase dataContext, ConnectionOptions op
...
+			_configurationID = null;
+
+			var action = Options.Reapply(this, prevOptions);
+
+			action += () =>
+			{
+				Options          = prevOptions;
+
+#if DEBUG
+				_configurationID = null;
```
MaceWindu: do we need those DEBUG branches or they could be removed?

Tests/Linq/Linq/ParameterTests.SqlServer.cs
```diff
@@ -806,5 +806,12 @@ private void SetupCustomTypes(MappingSchema ms, bool asDataParameter = false)
 				ms.SetConvertExpression<VarBinary?, DataParameter?>(v => v == null ? null : DataParameter.VarBinary(null, v.Value));
 			}
 		}
+
+		[Test, Explicit]
```
MaceWindu: I would prefer to avoid explicit (AKA non-existing) tests if they don't break CI runs. Can we remove Explicit attribute?

Source/LinqToDB/Common/IOptionSet.cs
igor-tkachev: fixed

Source/LinqToDB/Common/IReapplicable.cs
igor-tkachev: fixed

Source/LinqToDB/Common/Internal/IdentifierBuilder.cs
igor-tkachev: done

Source/LinqToDB/Common/Internal/IdentifierBuilder.cs
igor-tkachev: done

Source/LinqToDB/Common/OptionsContainer.cs
igor-tkachev: done

Source/LinqToDB/Data/DataConnection.Configuration.cs
igor-tkachev: done

Source/LinqToDB/Data/DataConnection.Configuration.cs
igor-tkachev: done

Source/LinqToDB/Data/DataConnection.Configuration.cs
igor-tkachev: fixed

Source/LinqToDB/DataProvider/SqlServer/SqlServerMappingSchema.cs
igor-tkachev: fixed

igor-tkachev: Adapter schema should have higher priority.

Source/LinqToDB/Mapping/LockedMappingSchemaInfo.cs
igor-tkachev: Yes. `LockedMappingSchemaInfo` is an internal class and used by unique provider/adapter schemas that have only single instance.

Source/LinqToDB/Mapping/MappingSchema.cs
igor-tkachev: Any negative number. What number do you like?

Source/LinqToDB/DataContext.cs
MaceWindu: This is not fixed

Source/LinqToDB/Data/DataConnection.Configuration.cs
MaceWindu: not fixed too

MaceWindu (changes requested):
two issues CommandTimeout not fixed

Source/LinqToDB/DataContext.cs
igor-tkachev: It is. This is outdated code.

MaceWindu (dismissed):

PR #4869, opened 2025-03-10
Re-enable sql server 2017 linux tests

MaceWindu:
Let's see if it works with recent update ... https://learn.microsoft.com/en-us/troubleshoot/sql/releases/linux/release-history-2017#AzureConnectFeaturePack


MaceWindu:
Nope

PR #4875, opened 2025-03-14
JOIN with nullable expression is not working

igor-tkachev:
The following code is not working 

```c#
from t2 in temp1
join t3 in temp2 on t2.ID equals t3.ID into gt3
from t3 in gt3.DefaultIfEmpty()
let Value = t3.Value ?? t2.Value
//let Value = Sql.AsNullable(t3.Value) ?? t2.Value
join t4 in temp3 on new { Value } equals new { t4.Value } into gt5
from t4 in gt5.DefaultIfEmpty()
select t4
```

If you uncomment the commented line, the code works.

Worked in 2.xxx

### DIFF
--- a/Tests/Base/TestBase.Asserts.cs
+++ b/Tests/Base/TestBase.Asserts.cs
@@ -21,9 +21,9 @@ namespace Tests
 {
 	partial class TestBase
 	{
-		protected void AreEqual<T>(IEnumerable<T> expected, IEnumerable<T> result, bool allowEmpty = false)
+		protected void AreEqual<T>(IEnumerable<T> expected, IEnumerable<T> result, bool allowEmpty = false, bool printData = false)
 		{
-			AreEqual(t => t, expected, result, EqualityComparer<T>.Default, allowEmpty);
+			AreEqual(t => t, expected, result, EqualityComparer<T>.Default, allowEmpty, printData : printData);
 		}
 
 		protected void AreEqual<T>(IEnumerable<T> expected, IEnumerable<T> result, Func<IEnumerable<T>, IEnumerable<T>> sort)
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/SqlQuery/SqlCoalesceExpression.cs
```diff
@@ -51,7 +51,10 @@ public override bool Equals(ISqlExpression other, Func<ISqlExpression, ISqlExpre
 
 		public override bool CanBeNullable(NullabilityContext nullability)
 		{
-			return Expressions.All(e => e.CanBeNullable(nullability));
+			if (Expressions.Length == 0)
+				return false;
+
+			return Expressions[^1].CanBeNullable(nullability);
```
MaceWindu, line 57: ??

`coalesce(nullble, nonnullable, nullable)`
- old logic: nullable=false
- new logic: nullable=true

sdanyliv:
@MaceWindu, wrong fix. Will continue analyse.

MaceWindu:
Remaining baselines issues:
- [ ] `Issue2815Test2` doesn't work for some DBs (bad translation)
- [x] `TestIssue2832` introduce join condition with `IS NULL` check for non-nullable field (`oudg.DatagroupId`). Probably because it is defined using `Where` and use wrong nullability context
- [ ] `Issue2478Tests.OuterApplyTest` add unnecessary nested subquery for second query, also test doesn't use OuterApply anymore for first query but use subquery, which doesn't look necessary
- [ ] `SetOperatorComplexTests.EagerDifferentDetails` adds condition that is not needed as it evaluates to true always
- [ ] `SetOperatorComplexTests.EagerSameDetails` adds `bookid is null` condition which always false
- [ ] `SetOperatorComplexTests.ExceptInheritance` sql improved, but also have same issue as `EagerSameDetails` where it has unnecessary null checks for BookId and Discriminator
- [ ] `SetOperatorComplexTests.IntersectInheritance`  same as `EagerSameDetails` - always false null check for book id added
- [ ] `SetOperatorComplexTests.UsingDictionary` has always false null check for discriminator
- [x] `Issue1838Tests.ConditionalTests` nullability for aggregate is not correct at least for ClickHouse
- [x] `OrmBattleTests.LeftJoinTest` adds IN NULL check for non-nullable `[a_Category].[CategoryID]` column on join
- [x] ~`Issue1983Tests.Test` valid null check for aggregate value disappeared`~ change is correct, improved SQL even more
- [x] `BooleanTests.Test` looks like regression, unnecessary bool conversion added

Identified issues:
- [ ] `Issue2815Test2` translated incorrectly for most of providers
- [ ] would be nice if Inheritance fiter with non-null discriminators convert left join to inner join (`SetOperatorComplexTests` tests)
- [x] joined subquery nullability context not correct in `SqlExpressionConvertVisitor`
- [x] fix nullability of joined table in join condition
- [ ] outer apply optmization and other optimizations doesn't work well together (`Issue2478Tests.OuterApplyTest`)

PR #4876, opened 2025-03-15
Add Testing build configuration and solution filter

MaceWindu:
Adds new build configuration Testing based on Debug configuration to include only `net9.0` TFM and projects, required to run .net 9 tests and efcore9 tests.

This will reduce time, wasted on build during local testing/development due to multiple configurations being build. E.g. on my machine it shows x4 times difference in build.

Also added slnf solution filter for new configuration. It doesn't add additional perfomace gain (maybe less memory used by VS) and just exclude projects, that are not built to avoid confusion/reduce visual noise.


MaceWindu:
And yes, TFMs should be specified in csproj, msbuild goes crazy when they set only in props

viceroypenguin (approved):

PR #4885, opened 2025-03-20
Use EnsureConnection() in ConfigurationID property of DataConnection and DataContext

### LINKED ISSUE #4883: Async query opens connection synchronously
### Description

During query execution the connection is opened synchronously even though the query execution is async. This happens when connection is provided as a part of DataOptions. 

In a constructor, Query class attempts to obtain ConfigurationID from DataContext, which in turn calls DataConnection.EnsureConnection() to obtain connection string. DataConnection.EnsureConnection() then opens the connection synchronously. There should be no need to open connection in this case as the code only needs the connection string.

### Reproduction Steps

```csharp
#r "nuget: linq2db, 5.4.1"
#r "nuget:Microsoft.Data.SqlClient/6.0.1"

using System.Data;
using System.Data.Common;
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;
using Microsoft.Data.SqlClient;

var sqlConn = new SqlConnection(@"Server=.;Database=Test;Trusted_Connection=True;TrustServerCertificate=True");
var connection = new DelegatingConnection(sqlConn);

var options = new DataOptions();
options = options.UseConnection(
    SqlServerTools.GetDataProvider(
        SqlServerVersion.v2019,
        SqlServerProvider.MicrosoftDataSqlClient),
    connection, disposeConnection: true);
    
var db = new DataConnection(options);

var _ = await db.GetTable<TestEntity>().FirstOrDefaultAsync();


[Table(Name = "[dbo].[TestEntities]")]
public class TestEntity
{
    [Column]
    public int ID { get; set; }
}


class DelegatingConnection(SqlConnection conn) : DbConnection
{
    private readonly SqlConnection conn = conn;

    public override string ConnectionString { get => conn.ConnectionString; set => conn.ConnectionString = value; }

    public override string Database => conn.Database;

    public override string DataSource => conn.DataSource;

    public override string ServerVersion => conn.ServerVersion;

    public override ConnectionState State => conn.State;

    public override void ChangeDatabase(string databaseName)
    {
        conn.ChangeDatabase(databaseName);
    }

    public override void Close()
    {
        conn.Close();
    }

    public override void Open()
    {
        Console.WriteLine($"Open() is called.\n{new System.Diagnostics.StackTrace()}");
        conn.Open();
    }

    public override Task OpenAsync(CancellationToken cancellationToken)
    {
        Console.WriteLine("OpenAsync() is called.");
        return conn.OpenAsync(cancellationToken);
    }

    protected override DbTransaction BeginDbTransaction(IsolationLevel isolationLevel)
    {
        return conn.BeginTransaction();
    }

    protected override DbCommand CreateDbCommand()
    {
        return conn.CreateCommand();
    }
}
```

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

5.4.1

### .NET version

9.0.2

### Database

SQL Server 2022

### Database provider

Microsoft.Data.SqlClient 6.0.1

### Operating system

Windows

### Configuration

_No response_

sksk571:
Use EnsureConnection(connect: false) in ConfigurationID to obtain connection string. This way connection is not opened synchronously during query generation.

Fixes #4883

### DIFF
--- a/Source/LinqToDB/Data/DataConnection.Configuration.cs
+++ b/Source/LinqToDB/Data/DataConnection.Configuration.cs
@@ -488,7 +488,7 @@ namespace LinqToDB.Data
 					using var idBuilder = new IdentifierBuilder();
 					_configurationID = idBuilder
 						.Add(_msID = ((IConfigurationID)MappingSchema).ConfigurationID)
-						.Add(ConfigurationString ?? ConnectionString ?? Connection.ConnectionString)
+						.Add(ConfigurationString ?? ConnectionString ?? EnsureConnection(connect: false).ConnectionString)
 						.Add(Options)
 						.Add(GetType())
 						.CreateID();
--- a/Source/LinqToDB/DataContext.cs
+++ b/Source/LinqToDB/DataContext.cs
@@ -125,7 +125,7 @@ namespace LinqToDB
 					_configurationID = idBuilder
 						.Add(_msID = ((IConfigurationID)MappingSchema).ConfigurationID)
 						// GetDataConnection :-/
-						.Add(ConfigurationString ?? ConnectionString ?? GetDataConnection().Connection.ConnectionString)
+						.Add(ConfigurationString ?? ConnectionString ?? GetDataConnection().EnsureConnection(connect: false).ConnectionString)
 						.Add(Options)
 						.Add(GetType())
 						.CreateID();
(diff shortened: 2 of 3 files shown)


viceroypenguin (approved):

igor-tkachev (approved):

MaceWindu (approved):
I will merge it for now so it will be a part of upcoming release. Still plan to review used approach in #4894

PR #4888, opened 2025-03-23
Left Join issue

igor-tkachev:
The following code is not working:

```c#
from p in db.Parent
join c in db.Child on p.Value1 equals c.ParentID into g
from c in g.DefaultIfEmpty()
where g == null
select p.ParentID;
```

```
LinqToDB.LinqToDBException : The LINQ expression 'g' could not be converted to SQL.
```

Worked in 5.4.1.

### DIFF
--- a/Source/LinqToDB/Linq/Query.cs
+++ b/Source/LinqToDB/Linq/Query.cs
@@ -590,19 +590,6 @@ namespace LinqToDB.Linq
 
 		internal static Query<T> CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, ref IQueryExpressions expressions)
 		{
-			var linqOptions = optimizationContext.DataContext.Options.LinqOptions;
-
-			if (linqOptions.GenerateExpressionTest)
-			{
-				var testFile = new ExpressionTestGenerator(dataContext).GenerateSource(expressions.MainExpression);
-
-				if (dataContext.GetTraceSwitch().TraceInfo)
-					dataContext.WriteTraceLine(
-						$"Expression test code generated: \'{testFile}\'.",
-						dataContext.GetTraceSwitch().DisplayName,
-						TraceLevel.Info);
-			}
-
 			var query = new Query<T>(dataContext);
 
 			try
@@ -623,7 +610,19 @@ namespace LinqToDB.Linq
 			}
 			catch (Exception)
 			{
-				if (!linqOptions.GenerateExpressionTest)
+				var linqOptions = optimizationContext.DataContext.Options.LinqOptions;
+
+				if (linqOptions.GenerateExpressionTest)
+				{
+					var testFile = new ExpressionTestGenerator(dataContext).GenerateSource(expressions.MainExpression);
+
+					if (dataContext.GetTraceSwitch().TraceInfo)
+						dataContext.WriteTraceLine(
+							$"Expression test code generated: \'{testFile}\'.",
+							dataContext.GetTraceSwitch().DisplayName,
+							TraceLevel.Info);
+				}
+				else
 				{
 					dataContext.WriteTraceLine(
 						"""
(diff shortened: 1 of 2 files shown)


sdanyliv:
@igor-tkachev, which SQL do you expec from `where g == null`. Looks like a bug in your query.

sdanyliv:
@igor-tkachev, translation of `GroupJoin` handling is completely rewritten and we need special complex "crutches" to make it work. We do not store information about LEFT JOIN or any join in this step, it is just GroupJoin result, which has no direct SQL tarnsaltion.

How about
1. Roslyn analyzer to find such problems in existing code
2. You can register `IQueryExpressionInterceptor` and correct expression tree for such "bloody" projects. Pattern is comlex, but doeable. The same query in Method syntax:
```cs
var q = db.Parent
	.GroupJoin(db.Child, p => p.Value1, c => c.ParentID, (p, g) => new { p, g })
	.SelectMany(t => t.g.DefaultIfEmpty(), (t, c) => new { t, c })
	.Where(t => t.t.g == null)
	.Select(t => t.t.p.ParentID);
```

viceroypenguin:
@igor-tkachev An alternative query that would be more explicitly correct that should work:
```cs
from p in db.Parent
where !db.Child.Any(c => c.ParentID == p.Value1)
select p.ParentID;
```

sdanyliv:
@viceroypenguin, it's not about writing the ideal query, but rather about handling problematic customer queries without needing to rewrite them or at least don't allow to compile.

viceroypenguin:
@sdanyliv my point is, I don't agree that writing `where g == null` is a bug; it's a valid query. We may choose not translate it, but it's not that ridiculous of a query to write.

sdanyliv:
> I don't agree that writing `where g == null` is a bug.

Modern compilers will warn about comparing a non-nullable `g` to null. What does it even mean for the collection to be null if `GroupJoin` doesn't permit that scenario?

Yes, I can rewrite `g == null` as `!g.Any()` to remove the LEFT JOIN, but I prefer not to open Pandora's box here - the behavior should remain explicitly defined.

viceroypenguin:
> What does it even mean for the collection to be null if `GroupJoin` doesn't permit that scenario?
Ah, good point. Then yes, I agree.

igor-tkachev:
@sdanyliv I understand that this is kind of crazy code. If we write analyzer to show this code, it would be OK. But even if you generate an appropriate exception with explanation, it will be very helpful as well.

sdanyliv:
Modified the exception message as follows. To avoid introducing specialized logic during predicate generation, the message is kept general:
```
LinqToDB.LinqToDBException : The LINQ expression 'g' could not be converted to SQL.
Additional details: 'Cannot use the collection from a GroupJoin as an expression. This typically occurs when attempting a LEFT JOIN and choosing the wrong property for comparison.'
```

Source/LinqToDB/Linq/Query.cs
```diff
@@ -621,9 +608,21 @@ internal static Query<T> CreateQuery(ExpressionTreeOptimizationContext optimizat
...
+
+				if (linqOptions.GenerateExpressionTest)
+				{
+					var testFile = new ExpressionTestGenerator(dataContext).GenerateSource(expressions.MainExpression);
+
+					if (dataContext.GetTraceSwitch().TraceInfo)
+						dataContext.WriteTraceLine(
+							$"Expression test code generated: \'{testFile}\'.",
+							dataContext.GetTraceSwitch().DisplayName,
+							TraceLevel.Info);
```
viceroypenguin:
```suggestion
					if (dataContext.GetTraceSwitch().TraceInfo)
					{
						dataContext.WriteTraceLine(
							$"Expression test code generated: \'{testFile}\'.",
							dataContext.GetTraceSwitch().DisplayName,
							TraceLevel.Info);
					}
```

viceroypenguin (approved):

MaceWindu (approved):

PR #4900, opened 2025-03-29
Address several API issues around IDataContext implementations

### LINKED ISSUE #4320: Add context disposed checks to public APIs of context classes where it is missing
See #4303

We have some public APIs, that doesn't check dispose status of context. We should review following classes and add missing checks for our context implementations:
- DataConnection
- DataContext
- RemoteContext

### LINKED ISSUE #4894: Review use of connection string in context id generation
https://github.com/linq2db/linq2db/pull/4885

1. probably we don't need to use connection string explicitly as it is either not specified explicitly or part of data options anyway
2. check why baselines added extra query to `Issue496Tests.Test1` and `Issue488Tests.Test1` baselines with PR changes

MaceWindu:
- [x] Fix #4320 : add missing disposed checks
- [x] Fix #4894
- [x] Obsolete/fix dangerous dbconnection access APIs on `DataConnection`
- [x] fix discovered multiple sync-over-async calls
- [x] obsolete `IsMarsEnabled` and  `GetConnectionInfo` APIs
- [x] obsolete (but still use for now)  `Configuration.Data.ThrowOnDisposed` option
- [x] review existing APIs on contexts for more obsoletion/removal candidates
- [x] add explicit connect APIs to data connection to replace Connection property obsoletion
- [x] add `UseTraceSwitch` `DataOptions` configuration API
- [x] add  `DataContext.SetKeepConnectionAlive[Async]` method instead of blocking property setter
- [x] Fix context instance stuck in mapper discovered by new use-after-dispose exceptions
- [x] **BREAKIN' CHANGE** `DataConnection.BeginTransaction*` will throw exception if transaction already exists instead of silently rolling it back
- [x] **BREAKIN' CHANGE** disallow setting negative command timeout to `DataConnection.CommadTimeout` to reset it. Instead introduce `ResetCommandTimeout[Async]()` APIs

Obsolete several APIs for various reasons, mostly because:
- they are not expected to be available to user
- they implement sync-over-async pattern
- perform changes (e.g. to context configuration) that should be done through `DataOptions`

Some of obsoletions are temporary as we don't plan to remove API itself, but only change it's visibility with v7 release.

### DIFF
--- a/Source/LinqToDB.EntityFrameworkCore/Internal/LinqToDBForEFQueryProvider.cs
+++ b/Source/LinqToDB.EntityFrameworkCore/Internal/LinqToDBForEFQueryProvider.cs
@@ -195,6 +195,9 @@ namespace LinqToDB.EntityFrameworkCore.Internal
 		IReadOnlyList<QuerySql> IExpressionQuery.GetSqlQueries(SqlGenerationOptions? options) =>
 			((IExpressionQuery)QueryProvider).GetSqlQueries(options);
 
+		Task<IReadOnlyList<QuerySql>> IExpressionQuery.GetSqlQueriesAsync(SqlGenerationOptions? options, CancellationToken cancellationToken) =>
+			((IExpressionQuery)QueryProvider).GetSqlQueriesAsync(options, cancellationToken);
+
 		#endregion
 	}
 }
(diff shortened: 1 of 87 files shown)


viceroypenguin (dismissed):

MaceWindu:
Added changes list to release notes https://github.com/linq2db/linq2db/wiki/Linq-To-DB-6#context-api-review

MaceWindu:
Still need to review baselines

MaceWindu:
baselines are good. Just a lot of async queries added, which were not logged before due to bug in async data connection calls

sdanyliv (approved):

PR #4909, opened 2025-04-07
Improved LoadWith handling.

### LINKED ISSUE #4797: Associations Issues Loading Nested `LoadWith`s
I'm having difficulty loading nested associations with the latest version. This is a regression of 6.0.0 since this works in 5.4.1.

I have a query that looks something like this:

```csharp
context
    .GetTable<Parent>()
    .LoadWith(
        x => x.Children,
        x => x.LoadWith(y => y.Parent, y => y.LoadWith(z => z.Children))
    );
```

Where each child has one parent and each parent can have multiple children.

The issue I'm having is that the first `.Children` is loaded, and their `.Parent` is as well, but the final `.Children` is not loaded.

I've provided some example code below. It works when using version `5.4.1`, but fails when using anything newer.

<details>
<summary>Full Example</summary>

```csharp
using System.Diagnostics;
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.DataProvider.SQLite;
using LinqToDB.Mapping;
using Microsoft.Data.Sqlite;

namespace linq2db_bug;

[Table(Name = "Parent")]
public class Parent
{
    [Column("Id"), PrimaryKey, Identity]
    public int Id { get; set; }

    [Association(ThisKey = nameof(Id), OtherKey = nameof(Child.ParentId))]
    public Child[]? Children { get; set; }
}

[Table(Name = "Child")]
public class Child
{
    [Column("Id"), PrimaryKey, Identity]
    public int Id { get; set; }

    [Column("ParentId")]
    public int? ParentId { get; set; }

    [Association(ThisKey = nameof(ParentId), OtherKey = nameof(linq2db_bug.Parent.Id))]
    public Parent? Parent { get; set; }
}

class Program
{
    public static async Task Main()
    {
        await using var context = GetDataContext();

        var parent = await context.InsertWithInt32IdentityAsync(new Parent());
        var child1 = await context.InsertWithInt32IdentityAsync(new Child { ParentId = parent });
        var child2 = await context.InsertWithInt32IdentityAsync(new Child { ParentId = parent });

        var query = context
            .GetTable<Parent>()
            .LoadWith(
                x => x.Children,
                x => x.LoadWith(y => y.Parent, y => y.LoadWith(z => z.Children))
            );

        var results = await query.ToArrayAsync();

        var nestedChildIds = results
            .Single()
            .Children!
            .First()
            .Parent!
            .Children // should be 2, but is null
            ?.Select(it => it.Id)
            .Order()
            .ToArray();

        Debug.Assert(nestedChildIds?.Length == 2);
        Debug.Assert(nestedChildIds?[0] == child1);
        Debug.Assert(nestedChildIds?[1] == child2);
    }

    private static DataConnection GetDataContext()
    {
        var connection = new SqliteConnection("Data Source=InMemoryDb;Mode=Memory;Cache=Shared");
        var provider = SQLiteTools.GetDataProvider();

        var options = new DataOptions().UseConnection(
            provider,
            connection,
            disposeConnection: true
        );

        var context = new DataConnection(options);
        context.CreateTable<Child>();
        context.CreateTable<Parent>();

        return context;
    }
}
```

</details>

sdanyliv:
Fixes #4797


Source/LinqToDB/Linq/Builder/AssociationHelper.cs
```diff
@@ -225,38 +224,34 @@ public static LambdaExpression CreateAssociationQueryLambda(
...
-			if (loadWith?.NextInfos != null)
-			{
-				var associationLoadWith = loadWith.NextInfos
-					.FirstOrDefault(li =>
-						MemberInfoEqualityComparer.Default.Equals(li.MemberInfo, association.MemberInfo));
 
-				associationLoadWith ??= loadWith.NextInfos
-					.FirstOrDefault(li =>
-						li.MemberInfo?.Name == association.MemberInfo.Name);
+				var associationLoadWith = found;
```
viceroypenguin, line 248: is there a reason to create a new variable here, when neither `associationLoadWith` nor `found` are used again after the `if()` on the next line?

viceroypenguin (approved):

viceroypenguin (approved):

MaceWindu (approved):

PR #4938, opened 2025-05-01
Update tests for #1294 and fix TableFunction parameters generation

### LINKED ISSUE #1294: Sql.TableFunction parametrization
I am using a custom sql table function for joins. It looks like this:

```c#
		[Sql.TableFunction(Name = "somePermissionFunction")]
		public ITable<EntityBase> GetPermissions(int userId, int subcatId)
		{
			var methodInfo = (typeof(TestDataConnection)).GetMethod("GetPermissions", new[] { typeof(int), typeof(int) });

			return _dataConnection.GetTable<EntityBase>(this, methodInfo, userId, subcatId);
		}

```

and after being used like:

```c#
var userId = 1;
var subcatId = 2;
var userId2 = 3;
var permissions = db.GetPermissions(userId, subcatId).Select(x => x.Id).Union(db.GetPermissions(userId2, subcatId).Select(x => x.Id));
var q = db.Users.Where(x => permissions.Contains(x.Id));
```

it's translated into the following sql:

```sql
SELECT
	-- fields
FROM
	[Users] [t1] WITH (NOLOCK)
WHERE
	EXISTS(
		SELECT
			*
		FROM
			(
				SELECT
					[t2].[Id] as [Id1]
				FROM
					[somePermissionFunction](1, 2) [t2]
				UNION
				SELECT
					[t3].[Id] as [Id1]
				FROM
					[somePermissionFunction](3, 2) [t3]
			) [t5]
	)
```

Is there a way to make those 1, 2, 3 into normal sql parameters?

MaceWindu:
Fix #1294


MaceWindu:
Tests show that use of constant value or `Sql.Contant` helper doesn't force literal parameters currently

Source/LinqToDB/DataExtensions.cs
```diff
@@ -78,8 +78,11 @@ public static ITable<T> GetTable<T>(
 
 				for (var i = 0; i < parameters.Length; i++)
 				{
-					var type = pis[i].ParameterType;
-					args.Add(Expression.Constant(parameters[i], (type.IsByRef ? type.GetElementType() : type)!));
+					var        type    = pis[i].ParameterType;
+					Expression argExpr = Expression.Constant(parameters[i], (type.IsByRef ? type.GetElementType() : type)!);
+					argExpr = SequenceHelper.WrapAsParameter(argExpr);
```
viceroypenguin, line 83: make `WrapAsParameter` an extension method to simplify this and avoid assigning the variable twice
```suggestion
					Expression argExpr = Expression
						.Constant(parameters[i], (type.IsByRef ? type.GetElementType() : type)!)
						.WrapAsParameter();
```

Source/LinqToDB/DataExtensions.cs
```diff
@@ -1504,8 +1507,7 @@ static Expression GenerateArray(object?[] arguments)
 					return Expression.Constant(null, typeof(object));
 				
 				var argumentType    = p.GetType();
-				var method          = Methods.LinqToDB.SqlParameter.MakeGenericMethod(argumentType);
-				var valueExpression = (Expression)Expression.Call(method, Expression.Constant(p, argumentType));
+				var valueExpression = SequenceHelper.WrapAsParameter(Expression.Constant(p, argumentType));
```
viceroypenguin, line 1510: Would also be simplified as an extension method

Issue #4917, opened 2025-04-15
"Connection string is not provided" exception when EF dbContext created using NpgsqlDataSource


denis-tsv:
### Description

net9.0
linq2db.EntityFrameworkCore 9.1.0-preview.4
Npgsql.EntityFrameworkCore.PostgreSQL 9.0.4

```cs
var services = new ServiceCollection();

services.AddDbContext<EfDbContext>(options =>
{
    var dataSource = new NpgsqlDataSourceBuilder("Host=localhost;Port=5432;Database=test;Username=postgres;Password=postgres;")
        .Build();
    options.UseNpgsql(dataSource);
});

LinqToDBForEFTools.Initialize();

var dbContext = services.BuildServiceProvider().GetRequiredService<EfDbContext>();

var dataConnection = dbContext.CreateLinqToDBConnection(); //Exception 
```

### Reproduction Steps

- Create EF dbContext using NpgsqlDataSource
- Call dbContext.CreateLinqToDBConnection(), exception will be thrown

### Exception details
```
System.InvalidOperationException: Connection string is not provided.
   at LinqToDB.DataProvider.ProviderDetectorBase`2.DetectServerVersion(ConnectionOptions options, TProvider provider)
   at LinqToDB.DataProvider.PostgreSQL.PostgreSQLProviderDetector.GetDataProvider(ConnectionOptions options, Provider provider, PostgreSQLVersion version)
   at LinqToDB.DataProvider.PostgreSQL.PostgreSQLTools.GetDataProvider(PostgreSQLVersion version, String connectionString)
   at LinqToDB.EntityFrameworkCore.LinqToDBForEFToolsImplDefault.CreatePostgreSqlProvider(PostgreSQLVersion version, String connectionString)
   at LinqToDB.EntityFrameworkCore.LinqToDBForEFToolsImplDefault.CreateLinqToDBDataProvider(EFProviderInfo providerInfo, LinqToDBProviderInfo provInfo, EFConnectionInfo connectionInfo)
   at LinqToDB.EntityFrameworkCore.LinqToDBForEFToolsImplDefault.<>c__DisplayClass5_0.<GetDataProvider>b__0(ProviderKey k)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
[...]

PR #4945, opened 2025-05-10
4940 fix postgres enums mappings

denis-tsv:
Is it possible to create value to sql

### DIFF
--- a/Source/LinqToDB.EntityFrameworkCore/EFCoreMetadataReader.cs
+++ b/Source/LinqToDB.EntityFrameworkCore/EFCoreMetadataReader.cs
@@ -343,6 +343,20 @@ namespace LinqToDB.EntityFrameworkCore
 						{
 							dataType = DbTypeToDataType(typeMapping.DbType.Value);
 						}
+						else if (typeMapping.GetType().Name == "NpgsqlEnumTypeMapping")
+						{
+							dataType = DataType.Enum;
+							
+							var labels = (Dictionary<object, string>)typeMapping.GetType().GetProperty("Labels")!.GetValue(typeMapping)!;
+							var enumDbType = prop.GetColumnType();
+							var typedLabels = labels.ToDictionary(kv => kv.Key, kv => $"'{kv.Value}'::{enumDbType}");
+							
+							//TODO add ValueToSqlConverter to result as attribute
+							// ms.SetValueToSqlConverter(prop.ClrType, (sb, _, v) =>
+							// {
+							// 	sb.Append(typedLabels[v]);
+							// });
+						}
 						else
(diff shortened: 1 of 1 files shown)


Source/LinqToDB.EntityFrameworkCore/EFCoreMetadataReader.cs
```diff
@@ -140,6 +140,20 @@ public MappingAttribute[] GetAttributes(Type type)
...
+			if (type.IsEnum)
+			{
+				var mapping = _mappingSource?.FindMapping(type);
+				if (mapping?.GetType().Name == "NpgsqlEnumTypeMapping")
+				{
+					var labels = (Dictionary<object, string>) mapping.GetType().GetProperty("Labels")!.GetValue(mapping)!;
+					var typedLabels = labels.ToDictionary(kv => kv.Key, kv => $"'{kv.Value}'::{mapping.StoreType}");
+
+					MappingSchema.Default.SetDataType(type, new SqlDataType(new DbDataType(type, DataType.Enum, mapping.StoreType)));
+					MappingSchema.Default.SetValueToSqlConverter(type, (sb, _, v) => sb.Append(typedLabels[v]));
```
sdanyliv, line 154: It is bad idea to change Default mapping schema.

denis-tsv: I found the way to do the same without changing of MappingSchema.Default

sdanyliv:
Please introduce unit test for regression handling.

viceroypenguin (approved):

denis-tsv:
PostgreSQL tests failed because of #4917
I fixed and checked tests at the local environment

viceroypenguin:
Upon consideration, I'm concerned about breaking changes, for consumers who have been using l2db.efc already with enums. What is the current behavior for them? Does it break, or just convert to integer?

viceroypenguin (approved):

denis-tsv:
This pull request doesn't affect enums mapped to integer values at database. Enums mapped ro integers will have [IntTypeMapping type](https://github.com/linq2db/linq2db/pull/4945/files#diff-64cde6986d0f512d3d0bdeb17e38cb140ee52f088a3dc96ff71989193a51880eR399), so pull request does't affects them.
Mapping changed only for enums mapped to [PostgreSQL enums](https://www.postgresql.org/docs/current/datatype-enum.html) which has NpgsqlEnumTypeMapping. linq2db generated invalid sql by default for temp tables, merge and bulk copy quieries before this pull request.

MaceWindu:
Thanks!

PR #4946, opened 2025-05-10
Remove IsTakeSupported from SqlProviderFlags

jods4:
`IsTakeSupported` was probably a legacy provider flags that was used before.
Currently all providers have `IsTakeSupported = true`, even the few that don't support TOP/TAKE/FETCH such as Oracle 11.

I'm noting that we have 60 flags remaining, that's a lot of bool / bytes. It would fit into just 8 bytes (a `long`) if we stored them as a bitfield. That would also simplify serialization (numbering 60 `Datamember` is no fun) as well as `Equals` and `GetHashCode` implementations.

In fact `GetHashCode()` impl. is currently horrible: `bool.GetHashCode()` simply returns 1 or 0. So all those chained `^` xor just compute whether there's an odd or even number of true flags ☹️


jods4:
There is 1 failure, which is unrelated. Concurrency test that fails to open connection to ClickHouse on Linux.

MaceWindu (approved):

PR #4947, opened 2025-05-10
Unbanned some integer apis

jods4:
For integers, I removed from banned api list:
- `.ToString` without format string
- `.Parse` and `.TryParse` without `NumberStyles`
- `StringBuilder.Append` and `StringBuilder.Insert`

Without a format or `NumberStyles`, .net formats integers without thousand separator, so culture doesn't matter.

I have updated a few trivial places `Append(CultureInfo.InvariantCulture, $"{value}")` as it's shorter code and it avoids boxing / allocations (at least on older .NET releases).


jods4:
There is 1 test failure, which is unrelated. It's a known flaky test (libdb2.so not found).

sdanyliv (approved):

MaceWindu (approved):

PR #4948, opened 2025-05-10
Remove ISqlBuilder

jods4:
Implementing `ISqlBuilder` is a extremely complex, as it should be able to convert any AST to SQL string.
`BasicSqlBuilder` is the only class that implements this interface and it's pretty certain that there will never be another one.
image

I have removed the interface and renamed `BasicSqlBuilder` to simply `SqlBuilder`.


sdanyliv:
Please don't. I need interface for the future refactoring.

jods4:
@sdanyliv okay then

Issue #4941, opened 2025-05-06
copy paste: Summaries of IsNotNull() <=> IsNullable


HypeillingerChillinger:
### Description

The summaries are identical for IsNotNull() and IsNullable()

### Reproduction Steps

-

### Linq To DB version

5.4.1

### .NET version

9.0.2

Akilaydin:
I’ve opened a [PR that addresses this](https://github.com/linq2db/linq2db/pull/4970) – please take a look

MaceWindu:
Resolved by #4970

PR #4969, opened 2025-06-13
Fixed IsNotNull method misleading summary (Issue #4941)

Akilaydin:
The summary tag for the IsNotNull() method was previously duplicating the description 
used by IsNullable(), which was misleading. This change updates the documentation 
to clearly state that IsNotNull() marks the column as NOT NULL, disallowing any NULL values.

See the [issue](https://github.com/linq2db/linq2db/issues/4941)for details


Akilaydin:
Sorry, didn't make the right branch before opening pull request. Fixed here https://github.com/linq2db/linq2db/pull/4970

PR #4974, opened 2025-06-18
fix linq2db.Tools dependency version linq2db.Scaffold nuspec

MaceWindu:


MaceWindu:
Verified fixed 
image

PR #4993, opened 2025-07-01
Remove unneceessary `bindingRedirect`s

viceroypenguin:
Formerly necessary for netfx projects, but are unnecessary since all projects have moved to SDK-style projects. Consumers may still need them, but that need is not impacted in any way by this change.


MaceWindu:
FYI, this was already done in https://github.com/linq2db/linq2db/pull/4875, but I don't mind to merge this first

Tests/Linq/app.config
```diff
@@ -1,12 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
-<configuration>
-	<runtime>
-		<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
MaceWindu: this one cannot be removed. it is the only redirect, not handled by build

MaceWindu (dismissed):

Tests/Linq/app.config
viceroypenguin: Sorry, got busy with other things this weekend; will look at other ways to fix in a separate PR later.

PR #5011, opened 2025-07-09
Feature. LEFT JOIN transformation

### LINKED ISSUE #5004: [Optimization] Implement left join conversion to inner join if query contains non-nullable filter over non-nullable field
For details check `SetOperatorComplexTests` with inheritance (https://github.com/linq2db/linq2db/pull/4875#issuecomment-3032154448)

sdanyliv:
Fix #5004

A `LEFT JOIN` can be converted to an `INNER JOIN` when the `WHERE` clause enforces a condition that guarantees non-null values from the joined table - such as checking for `NOT NULL` on a non-nullable column or comparing it to a constant.

Examples:

```sql
SELECT
    *
FROM Some
LEFT JOIN Joined j ON ...
WHERE j.Id IS NOT NULL
```

```sql
SELECT
    *
FROM Some
LEFT JOIN Joined j ON ...
WHERE j.Name = 'John'
```
Also it is possible to transform LEFT JOIN if it is followew by INNER JOIN with condition on not nullable column 
```sql
SELECT
    *
FROM Some
LEFT JOIN Joined j ON ...
INNER JOIN Some2 j2 ON j2.Id == j.NotNullable
```

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/SelectQueryOptimizerVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/SelectQueryOptimizerVisitor.cs
@@ -1908,16 +1908,20 @@ namespace LinqToDB.Internal.SqlQuery
 						return false;
 				}
 
-				if (!subQuery.Select.Columns.All(c =>
-						QueryHelper.UnwrapCastAndNullability(c.Expression) switch
-						{
-							SqlColumn or SqlField or SqlTable or SqlBinaryExpression => true,
-							SqlParameterizedExpressionBase e => !e.IsAggregate,
-							_ => false,
-						})
-					)
+				// Check that all columns in sub-query are allowed to move up
+				NullabilityContext? nullabilityContext = null;
+				foreach (var c in subQuery.Select.Columns)
 				{
-					return false;
+					var columnExpression = QueryHelper.UnwrapCast(c.Expression);
+					if (columnExpression is not (SqlField or SqlColumn))
+					{
+						nullabilityContext ??= NullabilityContext.GetContext(subQuery);
+						if (!c.Expression.CanBeNullable(nullabilityContext))
+						{
+							if (QueryHelper.IsDependsOn(selectQuery, c, [joinTable]))
+								return false;
+						}
+					}
 				}
 			}
 
--- a/Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
@@ -157,7 +157,7 @@ namespace LinqToDB.Internal.SqlQuery
 			public          bool                    DependencyFound;
 		}
 
-		static bool IsDependsOn(IQueryElement testedRoot, IQueryElement onElement, HashSet<IQueryElement>? elementsToIgnore = null)
+		public static bool IsDependsOn(IQueryElement testedRoot, IQueryElement onElement, HashSet<IQueryElement>? elementsToIgnore = null)
 		{
 			var ctx = new IsDependsOnElementContext(onElement, elementsToIgnore);
 
@@ -1713,6 +1713,14 @@ namespace LinqToDB.Internal.SqlQuery
 				else
 					break;
 			} while (true);
+
+			return expr;
+		}
+
+		public static ISqlExpression UnwrapCast(ISqlExpression expr)
+		{
+			while (expr is SqlCastExpression { Expression: var sqlCast })
+				expr = sqlCast;
 
 			return expr;
 		}
--- a/Tests/Linq/Linq/JoinTests.cs
+++ b/Tests/Linq/Linq/JoinTests.cs
@@ -587,111 +587,66 @@ namespace Tests.Linq
 		[Test]
 		public void GroupJoin9([DataSources(TestProvName.AllAccess, TestProvName.AllInformix)] string context)
 		{
-			using (var db = GetDataContext(context))
-				AreEqual(
-					Parent
-						.GroupJoin(
-							Parent,
-							x => new { Id = x.ParentID },
-							y => new { Id = y.ParentID },
-							(xid, yid) => new { xid, yid }
-						)
-						.SelectMany(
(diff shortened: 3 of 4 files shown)


Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
```diff
@@ -2104,6 +2109,92 @@ bool CorrectRecursiveCteJoins(SelectQuery selectQuery)
...
+						source = ExtractSource(exprExPredicate.Expr1) ?? ExtractSource(exprExPredicate.Expr2);
+						if (source != null)
+						{
+							if (!(exprExPredicate.Expr1 is SqlValue { Value: not null } || exprExPredicate.Expr2 is SqlValue { Value: not null }))
+							{
+								source = null;
+							}
+						}
+					}
+				}
```
viceroypenguin:
```suggestion
				if (predicate is SqlPredicate.IsNull { IsNot: true, Expr1: var predExpr })
				{
					source = ExtractSource(predExpr);
					if (source != null)
					{
						var nullability = NullabilityContext.GetContext(source as SelectQuery);
						if (nullability.CanBeNull(predExpr))
						{
							source = null; // cannot convert to inner join if source is nullable
						}
					}
				}
				else if (predicate is SqlPredicate.ExprExpr { Operator: SqlPredicate.Operator.Equal } exprExPredicate)
				{
					source = ExtractSource(exprExPredicate.Expr1) ?? ExtractSource(exprExPredicate.Expr2);
					if (source != null)
					{
						if (!(exprExPredicate.Expr1 is SqlValue { Value: not null } || exprExPredicate.Expr2 is SqlValue { Value: not null }))
						{
							source = null;
						}
					}
				}
```

Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
```diff
@@ -2104,6 +2109,92 @@ bool CorrectRecursiveCteJoins(SelectQuery selectQuery)
...
+
+			ISqlTableSource? ExtractSource(ISqlExpression expr)
+			{
+				if (expr is SqlColumn column)
+					return column.Parent;
+				if (expr is SqlField field)
+					return field.Table;
+				if (expr is ISqlTableSource ts)
+					return ts;
+				return null;
```
viceroypenguin:
```suggestion
				return expr switch
				{
					SqlColumn column   => column.Parent,
					SqlField field     => field.Table,
					ISqlTableSource ts => ts,
					_ => null,
				};
```

viceroypenguin (approved):

Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
sdanyliv: It is not relevant change. My variant will stop earlier in matching. In your case excution will continue on `else`. That's why I have not used pattern matching as you proposed.

Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
```diff
@@ -2104,6 +2109,92 @@ bool CorrectRecursiveCteJoins(SelectQuery selectQuery)
...
+
+				if (source != null)
+				{
+					var foundJoin = leftJoins.FirstOrDefault(j => j.Table.Source == source);
+					if (foundJoin != null)
+					{
+						leftJoins.Remove(foundJoin);
+						foundJoin.JoinType = JoinType.Inner;
+						foundJoin.IsWeak   = false;
+						leftJoins.Remove(foundJoin);
```
sdanyliv: Double remove. Should fix.

Source/LinqToDB/SqlQuery/SelectQueryOptimizerVisitor.cs
viceroypenguin: as long as it is not possible from a type system perspective (i.e. `SqlPredicate.ExprExpr` is a derived class of `SqlPredicate.IsNull`), the compiled code will be the same. once a type has been matched, it will assume that type is present for the rest of the `if`/`else if` chain, and not check the type again, even if the rest of the pattern fails matching.

sdanyliv: I had to check it. It would be great.

sdanyliv: Checked in sharp lab, looks like it is not right https://sharplab.io/#gist:c8589d9e85412b7b15560a92819f4b36

viceroypenguin: Ah, I'm thinking of `switch`es, where it will absolutely separate the arms based on type first, then do pattern matching inside each type. I wonder if it's due to the reuse of the parameter/variable (in if/elif), as opposed to sharing the expression (in switch). 

That said, if the types are `sealed`, it's a simple integer compare to find the type, so it's up to you if the cost of a check is worth it for the extra layer of indentation. 🤷🏼‍♂️

viceroypenguin (dismissed):

MaceWindu:
couple of tests failed

Source/LinqToDB/Internal/SqlQuery/SelectQueryOptimizerVisitor.cs
```diff
@@ -2103,7 +2113,152 @@ bool CorrectRecursiveCteJoins(SelectQuery selectQuery)
...
+				if (!hasStrictCondition)
+				{
+					for (var j = i + 1; j < joins.Count; j++)
+					{
+						var nextJoin = joins[j];
+						if (nextJoin.JoinType == JoinType.Inner)
+						{
+							if (IsStrictCondition(nullability, nextJoin.Condition, join.Table.Source))
+							{
+								//throw new InvalidOperationException("Test");
```
MaceWindu:
```suggestion
```

MaceWindu (approved):

PR #5020, opened 2025-07-15
Restore compatibility.

igor-tkachev:


Source/LinqToDB/DataProvider/DataProviderExtensions.cs
```diff
@@ -0,0 +1,171 @@
+﻿using System;
+using System.Data.Common;
+using System.Linq.Expressions;
+
+using LinqToDB.Interceptors;
+using LinqToDB.Internal.DataProvider;
+
+namespace LinqToDB.DataProvider
+{
+	public static class DataProviderExtensions
```
MaceWindu, line 10: those helpers should include call to cache clear to flush old mappings

sdanyliv, line 10: It’s recommended to move this method to an internal namespace and add a comment indicating that the API is subject to change without notice.

In version 7, we plan to rewrite the mapping system, and this method will definitely be removed.

igor-tkachev: I use it my projects a lot. Before we remove it, we will need to provide an alternative.

igor-tkachev: Done.

sdanyliv: I mean, move it under an `Internal` namespace to prevent others from discovering and relying on it.

Also, please provide a sample unit test for this method. That will help us design an appropriate replacement for this mapping in the future.

igor-tkachev: We just introduced the `Internal` namespace, and as I understand it, it should not be used by users at all. That said, I need this feature right now for my own projects. Once we redesign the mapping and have a proper alternative, we can safely remove it.
I'll add a test. Good idea.

MaceWindu (approved):

PR #5022, opened 2025-07-17
SQL Server decimal overflow test

igor-tkachev:
Adds a test that demonstrates an overflow issue when converting decimal values from SQL Server to C#. SQL Server supports a wider range for the `decimal` type (up to precision 38), while C# `decimal` is limited to 28–29 significant digits. This mismatch can cause overflows during data retrieval.


MaceWindu:
tests fail

MaceWindu (approved):

PR #5025, opened 2025-07-19
Statement Header and Footer support for Temp Table creation

AidenFuller:


AidenFuller:
Note, I wasn't entirely sure how to add tests to cover this in a provider-agnostic way so I didn't include any since it doesn't look like the coverage for ITable is extensive for the header/footer either, and it's basically just plugging into the existing API anyway. Happy to be advised on the best way to add tests for this though

igor-tkachev:
This approach feels inconvenient and leads to an overloaded and hard-to-maintain API, since every new parameter results in additional overloads everywhere. I suggest replacing all these overloads with a single method that accepts an options object (e.g., `CreateTempTableOptions` / `CreateTempTableParameters`), which would encapsulate all the parameters. This will keep the API clean, extensible, and much easier to support in the future.

AidenFuller:
Sure, that sounds reasonable enough

AidenFuller:
@igor-tkachev do you think it's reasonable to have API support for the existing methods exclusively in the form of extensions and discontinue the constructor versions in their current form of table/schema/database, etc. separated? Not sure how often users would be utilizing the constructor APIs `new TempTable<T>(db, ...)` versus the extensions `db.CreateTempTable<T>(....)`, so want to make sure

igor-tkachev:
I’d keep everything as is for compatibility and add a single method that takes a single object parameter (e.g., `CreateTempTableParameters`) to encapsulate all the arguments - for both the constructor and the extension method.

AidenFuller:
@igor-tkachev I've updated, however for the sake of simplicity I didn't try too hard to shoehorn everything into a single method. It didn't feel like a simple and easy to understand API when I tried to combine the IQueryable and IEnumerable overloads, for example, especially with the BulkCopyOptions only being applicable for the IEnumerable. I think it's more intuitive and simpler if those overloads are kept separate and the CreateTempTableOptions is just the wrapper for all of the properties specific to the table creation. I've also updated the existing overloads to redirect into the new overloads with CreateTempTableOptions so that the bulk of the logic is more concentrated into the places that are more maintainable and future-proof. 

It might make sense to broaden the scope of that object to be a `CreateTableOptions` rather than being specific to temp tables so that regular table initialization receives the same benefit.

What do you think?

AidenFuller:
@MaceWindu I've synchronized my fork to grab the latest master changes which fixed those build failures

Source/LinqToDB/CreateTempTableOptions.cs
viceroypenguin, line 1: The properties in this class should be either `{ get; init; }` with no constructor or `{ get; }` and set in the constructor. Having both constructor values and `{ get; set; }` to create a mutable type is problematic.

Source/LinqToDB/DataExtensions.TempTable.cs
```diff
@@ -116,16 +190,23 @@ public static TempTable<T> CreateTempTable<T>(
...
 		{
-			return new TempTable<T>(db, tableName, items, options, databaseName, schemaName, serverName, tableOptions);
+			return db.CreateTempTable(
+				new CreateTempTableOptions(
+					tableName   : tableName,
+					databaseName: databaseName,
+					schemaName  : schemaName,
+					serverName  : serverName,
+					tableOptions: tableOptions),
+				items, options);
```
viceroypenguin: Stylistic preference for all of these methods.
```suggestion
			return db.CreateTempTable(
				new CreateTempTableOptions(
					tableName:    tableName,
					databaseName: databaseName,
					schemaName:   schemaName,
					serverName:   serverName,
					tableOptions: tableOptions),
				items,
				options);
```

Source/LinqToDB/TempTable.cs
```diff
@@ -526,7 +594,7 @@ public static Task<TempTable<T>> CreateAsync(IDataContext db,
 			TableOptions          tableOptions      = default,
 			CancellationToken     cancellationToken = default)
 		{
-			return CreateAsync(db, null, items, tableName, databaseName, schemaName, action, serverName, tableOptions, cancellationToken);
+			return CreateAsync(db, null, new CreateTempTableOptions(tableName, databaseName, schemaName, serverName, tableOptions: tableOptions), items, action, cancellationToken);
```
viceroypenguin: Break this up into multiple lines for readability

viceroypenguin (commented):
stylistic nitpicks

MaceWindu (approved):

sdanyliv (approved):

MaceWindu:
@AidenFuller thanks for PR

PR #5032, opened 2025-07-20
[Cleanup] Polyfils/#if refactoring

MaceWindu:
- add SUPPORTS_DATEONLY symbol
- enable Lock polyfill
- enable ConcurrentDictionary.TryAdd polyfill

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -47,6 +47,7 @@
 
 		<!--feature flags-->
 		<DefineConstants Condition=" $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) ">SUPPORTS_COMPOSITE_FORMAT;$(DefineConstants)</DefineConstants>
+		<DefineConstants Condition=" $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0')) ">SUPPORTS_DATEONLY;$(DefineConstants)</DefineConstants>
 
 		<ArtifactsPath>$(MSBuildThisFileDirectory).build</ArtifactsPath>
 		<ArtifactsPivots>$(Configuration)\$(TargetFramework)\$(RuntimeIdentifier)</ArtifactsPivots>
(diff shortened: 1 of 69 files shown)


Source/LinqToDB/Internal/Async/AsyncDbConnection.cs
```diff
@@ -148,34 +146,28 @@ public virtual async ValueTask<IAsyncDbTransaction> BeginTransactionAsync(Isolat
...
 		{
-			if (Connection is IAsyncDisposable asyncDisposable)
-			{
-				var a = ActivityService.StartAndConfigureAwait(ActivityID.ConnectionDisposeAsync)?.AddQueryInfo(DataConnection, Connection, null);
+			var a = ActivityService.StartAndConfigureAwait(ActivityID.ConnectionDisposeAsync)?.AddQueryInfo(DataConnection, Connection, null);
 
-				if (a is null)
-					return asyncDisposable.DisposeAsync();
+			if (a is null)
+				return Connection.DisposeAsync();
```
sdanyliv, line 162: `Connection` is property. Maybe we should copy to local variable?

Source/LinqToDB/Internal/Async/AsyncDbTransaction.cs
```diff
@@ -96,26 +96,18 @@ public virtual void Dispose()
...
 		{
-			if (Transaction is IAsyncDisposable asyncDisposable)
-			{
-				var a = ActivityService.StartAndConfigureAwait(ActivityID.TransactionDisposeAsync)?.AddQueryInfo(DataConnection, DataConnection?.CurrentConnection, null);
+			var a = ActivityService.StartAndConfigureAwait(ActivityID.TransactionDisposeAsync)?.AddQueryInfo(DataConnection, DataConnection?.CurrentConnection, null);
 
-				if (a is null)
-					return asyncDisposable.DisposeAsync();
+			if (a is null)
+				return Transaction.DisposeAsync();
```
sdanyliv, line 102: `Transaction` is property. Maybe we should copy to local variable?

sdanyliv (dismissed):

Source/LinqToDB/Internal/Async/AsyncDbConnection.cs
MaceWindu: we can, but it will probably hide use issue as ADO classes are not expected to be used concurrently

PR #5065, opened 2025-08-01
Disable mapping of  ValueType's to scalar types by default

### LINKED ISSUE #5056: Composite object constructors support in LINQ regression
I upgraded from version 5.4.1 to 6 and I'm facing a problem where saving doesn't work if I create a nested object

**This worked in 5.4.1 and doesn't work in 6** 
I get an error The LINQ expression could not be converted to SQL.Expression
```
await db.Process
        .InsertAsync(() => new ProcessDb
        {
            Id = approvalProcess.Id,
            DocumentId = approvalProcess.DocumentId,
            ContextId = approvalProcess.ContextId,
            State = new StateDb
            {
                CompletedTasks = approvalProcess.CompletedTasks,
                CurrentTask = approvalProcess.CurrentTask,
                Step = approvalProcess.Step,
                Status = approvalProcess.Status,
                StartAt = approvalProcess.StartAt
            },
            ConcurrencyVersion = 1
        }, cancellationToken)
        .ConfigureAwait(false);
```
 
 Now, instead, I need to do this
 
```
   var state = new StateDb
  {
      CompletedTasks = approvalProcess.CompletedTasks,
      CurrentTask = approvalProcess.CurrentTask,
      Step = approvalProcess.Step,
      Status = approvalProcess.Status,
      StartAt = approvalProcess.StartAt
  };
  
  await db.Process
          .InsertAsync(() => new ProcessDb
          {
              Id = approvalProcess.Id,
              DocumentId = approvalProcess.DocumentId,
              ContextId = approvalProcess.ContextId,
              State = state,
              ConcurrencyVersion = 1
          }, cancellationToken)
          .ConfigureAwait(false);
```
 Please tell me is this the intended behavior or is it really a bug?

_Originally posted by @qpd-repos in https://github.com/linq2db/linq2db/discussions/5023#discussioncomment-13947023_

MaceWindu:
Related to #5056

- Fix #5056

New parser use information about scalar types better than old one, leading to situations where incorrect type attribution to scalar leads to errors.

To solve it we should stop treatening structs as scalars by default and require user to mark scalar types explicitly for structs too. This doesn't affect well-known scalar types and provider-specific types as they are registered as scalars in mapping schema already (with some fixes in this PR)

a bit late big change, but we need it to avoid regressions and unstable workarounds


MaceWindu:
Well, situation is following:  with more places using IsScalar check in new parser now we have following breaking changes already:
- reference types (classes), not marked as scalars, but used for scalar values (see original issue) now could produce errors without annotations
- value types (structs) being marked as scalars by default but used as structural types also could produice errors now (see test in this pr)

Looks like in any case there is a behavior change in new version and I think it will be less disturbing to ask users only to mark scalar types as a migration step to v6 (and it could be done beforehand with v5 codebase already)

sdanyliv (dismissed):

sdanyliv (approved):

PR #5071, opened 2025-08-04
Change ClickHouse.Client to ClickHouse.Driver

danielcrenna:
ClickHouse is officially supporting a fork of the original ClickHouse.Client, changing all namespaces including the assembly and package names to ClickHouse.Driver.

This breaks the linq2db integration on the ClickHouse side, as the library tries to resolve ClickHouse.Client.dll and other namespaces that have changed to use Driver nomenclature.

The commit should switch these references to the officially supported client.

The new driver repository is here: https://github.com/ClickHouse/clickhouse-cs

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseProviderAdapter.cs
+++ b/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseProviderAdapter.cs
@@ -210,15 +210,13 @@ namespace LinqToDB.Internal.DataProvider.ClickHouse
 			typeMapper.RegisterTypeWrapper<DriverWrappers.ClickHouseConnection             >(connectionType);
 			typeMapper.RegisterTypeWrapper<DriverWrappers.ClickHouseConnectionStringBuilder>(connectionStringBuilderType);
 			typeMapper.RegisterTypeWrapper<DriverWrappers.ClickHouseBulkCopy               >(bulkCopyType);
+			typeMapper.RegisterTypeWrapper<DriverWrappers.ClickHouseDecimal                >(decimalType);
+			typeMapper.FinalizeMappings();
 
 			var mappingSchema = new MappingSchema();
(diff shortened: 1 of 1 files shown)


viceroypenguin:
@danielcrenna unrelated nit: i'm seeing your commits show up as "Unverified" in github. This is because the commits are signed, but github does not have your public key, so it cannot confirm the signature. Would recommend you upload your public key to gh so that your commits show up as verified.

danielcrenna:
> @danielcrenna unrelated nit: i'm seeing your commits show up as "Unverified" in github. This is because the commits are signed, but github does not have your public key, so it cannot confirm the signature. Would recommend you upload your public key to gh so that your commits show up as verified.

This is fixed.

danielcrenna:
These errors look like file locking/flakiness in the harness, not sure I can diagnose that from here. 

https://github.com/linq2db/linq2db/pull/5071/checks?check_run_id=47348551271

Build/Azure/pipelines/templates/test-matrix.yml
```diff
@@ -814,8 +814,8 @@ jobs:
 # ClickHouse
         - name: ClickHouse Client
           title: ClickHouse Client
-          config_linux: clickhouse.client
-          config_macos: clickhouse.client
+          config_linux: ClickHouse.Driver
+          config_macos: ClickHouse.Driver
```
MaceWindu: those are names of configuration files above, so:
- it should use lowercase
- files Build/Azure/net90/clickhouse.client.json and Build/Azure/net80/clickhouse.client.json should be renamed

MaceWindu: also name/title properties could be changed to new name

MaceWindu:
Usually we preserve compatibility with old provider versions for two reasons:
- preserve compatibility for users
- if database provider has EFCore support - we must have compatibility with all EFCore versions we support. Currently it is EFC 3.1, 8 and 9

Looking at https://www.nuget.org/packages/EntityFrameworkCore.ClickHouse I see it has support for EF8 and 9.

Anyway, lets drop compatibility for now. If it will break somebody's code - we can restore it later.

MaceWindu (approved):

MaceWindu:
Thanks!

PR #5079, opened 2025-08-08
Enhance ConnectionOptions with Custom Options Dictionary for Advanced Provider Detection

nsarris:
This PR introduces a new `Custom` dictionary to the `ConnectionOptions` class, enabling more advanced and flexible provider detection scenarios.

### ✅ Key Changes

- **Custom Options Dictionary**  
  A new `Dictionary<string, object>` property has been added to `ConnectionOptions` to hold custom configuration values.

- **Improved Provider Detection**  
  Providers can now pass additional context—such as available mapping schemas—through the `Custom` dictionary. These can be set via `UseXXX` provider extension methods and consumed by the `ProviderDetector`.

- **Configuration ID Calculation**  
  The `Custom` dictionary is now included in the calculation of the `ConfigurationId`.  
  It is treated as a separate configuration segment and processed using the `IdentifierBuilder`. The resulting ID is integrated into the overall `ConnectionOptions` identifier.

### 💡 Motivation

Previously, the `ProviderDetector` relied solely on `ProviderType` and `ProviderVersion`, limiting extensibility.  
With the addition of the `Custom` dictionary, providers can now convey richer metadata to support more complex detection logic.

### DIFF
--- a/Source/LinqToDB/Data/ConnectionOptions.cs
+++ b/Source/LinqToDB/Data/ConnectionOptions.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Data.Common;
+using System.Linq;
 
 using LinqToDB.DataProvider;
 using LinqToDB.Interceptors;
@@ -64,7 +66,7 @@ namespace LinqToDB.Data
 		Func<ConnectionOptions, IDataProvider>?         DataProviderFactory       = default,
 		ConnectionOptionsConnectionInterceptor?         ConnectionInterceptor     = default,
 		Action<MappingSchema, IEntityChangeDescriptor>? OnEntityDescriptorCreated = default
-
+		
 		// If you add another parameter here, don't forget to update
 		// ConnectionOptions copy constructor and IConfigurationID.ConfigurationID.
 	)
@@ -90,7 +92,13 @@ namespace LinqToDB.Data
 			DataProviderFactory       = original.DataProviderFactory;
 			ConnectionInterceptor     = original.ConnectionInterceptor;
 			OnEntityDescriptorCreated = original.OnEntityDescriptorCreated;
+			Custom                    = original.Custom.ToDictionary(x => x.Key, x => x.Value);
 		}
+
+		/// <summary>
+		/// Custom options that can be setup in UseXXX methods of <see cref="DataOptionsExtensions"/> for extending <see cref="Internal.DataProvider.ProviderDetectorBase{TProvider}"/> auto detecion.
+		/// </summary>
+		public Dictionary<string, object> Custom { get; } = new();
 
 		int? _configurationID;
 		int IConfigurationID.ConfigurationID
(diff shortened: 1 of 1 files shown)


MaceWindu:
I would like to understand use case for it. Maybe we can have better solution

nsarris:
In the current architecture, the ProviderDetector serves as the central factory for creating DataProvider instances. All requests initiated through the UseXXX extension methods are funneled through the ProviderDetector. Additionally, static detectors are configured in DataConnection to enable automatic provider discovery when no explicit context is available.

DataProvider instances can currently receive supplemental metadata via DataOptions. However, these options are only utilized by the SQL generation pipeline and do not influence the DataProvider itself. This limitation arises because DataProvider instances are designed as long-lived dependencies—once instantiated, they are not dynamically reconfigured. Consequently, options that affect schema resolution, entity mapping, or other provider-specific behaviors cannot be applied after creation.

To address this limitation, it is necessary to supply additional contextual information at the provider detection stage. This ensures that ProviderDetector can construct DataProvider instances with a richer set of configuration options than just the current (Type, Version) combination.

Introducing a custom dictionary property in ConnectionOptions provides a flexible, non-breaking way to pass arbitrary key–value metadata through to the ProviderDetector, and ultimately, the DataProvider. This approach avoids altering method signatures for ProviderDetector.GetDataProvider or ProviderDetector.DetectProvider, which would constitute a breaking change to the public API.

This solution preserves backward compatibility while unlocking the ability to inject provider-specific schema and mapping configuration directly at connection creation time.

MaceWindu:
>> options that affect schema resolution, entity mapping, or other provider-specific behaviors cannot be applied after creation

I fail to see how those are provider-level options - could you provide some real-world examples to understand why it could be needed?

In any case I would like to avoid adding provider specific options to those option types. We have provider-specific option classes for that. I will check how it is feasible to pass DataOptions to provider detector so you can access provider-specific options object.

nsarris:
For the DB2i provider, there are two primary configuration parameters: Type (chosen from four available ADO providers) and Version (ranging from 7.1 to 7.4). In addition, there’s a choice between two mapping schemas—one that supports GUIDs in binary format and another that stores them as strings.

The original author of the provider also implemented automatic trimming of char types when reading values. I retained this behavior for backward compatibility, but I’m now considering making it optional.

These two provider-level options (and potentially other similar settings) need to be passed to the detector so it can construct provider instances with the appropriate constructor parameters. 

Passing the DataOptions class to the detector is a perfectly acceptable way to achiieve this, though it would introduce a small breaking change, as the current detection method only receives ConnectionOptions, ProviderType, and ProviderVersion.

Alternatively, custom providers could inherit from the base detector and extend it with a custom detection method. However, that approach somewhat undermines the purpose of offering a base detector with a shared, common interface.

MaceWindu:
I think we will reject this proposal.

I've checked how we call provider detector and the only place where we can pass external `DataOptions` is `UseXXX` configuration extensions which call `ProviderDetectorBase.CreateOptions` method. It doesn't look like it make much sense to change detector interfaces just to pass `DataOption` to detector.

You can implement your provider detection logic on your side and call it instead of `CreateOptions` call to provider detector from configuration helpers.

nsarris:
Understood. Thanks for the insights, I will follow your suggestion.

PR #5080, opened 2025-08-09
Changed the CustomFlags collection in SqlProviderFlags to HashSet

### LINKED ISSUE #5073: Custom Flags on SqlProvider Flags should be a HashSet
### Description

It's a minor performance improvement, but the Custom list of flags in SqlProviderFlags would only be used to look up values with contains. A HashSet makes more sense in that context than a List.

### New API

_No response_

nsarris:
Fix #5073

Updated the `CustomFlags` property in `SqlProviderFlags` to use `HashSet<string>` instead of `List<string>`, and removed the property setter. (Addresses #5073 )

### ✅ Key Changes

- Changed `CustomFlags` type from `List<string>` to `HashSet<string>`.
- Removed `CustomFlags` setter to guard against setting the property to `null`.
- Preserved collection mutability: consumers can still add or remove flags, but cannot replace the collection instance.

### 💡 Motivation

- **Performance:** `HashSet` provides faster membership checks (`Contains`) than `List`.
- **Equality Checks:** `HashSet` offers `SetEquals`, which is well-suited for property-level equality comparisons in `SqlProviderFlags`.
- **Safety:** Prevents `null` assignments by ensuring the collection is always instantiated in the constructor.
- **Code Quality:** Addresses prior internal comment noting that using `List` for `CustomFlags` was not the best choice.

### DIFF
--- a/Source/LinqToDB.Remote.HttpClient.Client/HttpClientLinqServiceClient.cs
+++ b/Source/LinqToDB.Remote.HttpClient.Client/HttpClientLinqServiceClient.cs
@@ -17,7 +17,7 @@ namespace LinqToDB.Remote.HttpClient.Client
 
 		LinqServiceInfo ILinqService.GetInfo(string? configuration)
 		{
-			var response = HttpClient.PostAsJsonAsync($"{requestUri}/{nameof(ILinqService.GetInfo)}", configuration, default).Result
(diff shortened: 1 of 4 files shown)


MaceWindu (commented):
I'm fine with this change. Let me add a test to ensure it works with remote contex...

MaceWindu:
hmm, I've created tests, but they show a lot of issues with flags transfer over remote context. both related and unrelated to this change. I will work on fixes for them...

MaceWindu (approved):

PR #5112, opened 2025-08-18
Fix incorrect order preserving

jogibear9988:


Source/LinqToDB/Internal/Linq/Builder/OrderByBuilder.cs
```diff
@@ -81,7 +81,7 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
 							break;
 					}
 
-					if (isValid && !prevSequence.SelectQuery.OrderBy.IsEmpty)
+					if (isValid && !prevSequence.SelectQuery.OrderBy.IsEmpty && sequence.SelectQuery != prevSequence.SelectQuery)
```
jogibear9988, line 84: I already thought of doing that, but I was not sure about the implications ;-)

PR #5129, opened 2025-09-10
Extract non-provider changes from YDB provider branch

MaceWindu:
To simplify review and test non-provider changes separately

Source changes:
- add `ConvertName.NameToCteName` conversion. For YDB we use TE as CTE and they share naming namespace with parameters, so we should be able to distinguish when we emit CTE name
- new linqpad packaging script
- ISO8601 interval generation helper for YDB interval type
- improvements to MultipleRows bulk copy helpers
- various data reader expression registration helpers improvements
- LIKE SQL generation improvements to allow custom typing of escape character as YDB requires specific literal type here
- improvements to enum values mapping in adapter wrappers
- fixes for CTE finalization in SQL optmizer
- remove setters for provider flags to value, equal with default flag value
- add join optimization flag to remove non-column equality expressions from join ON clause to WHERE clause. YDB supports only column equality in ON (also probably could be used by ClickHouse)
- `SameAsFirstOrSecondParameter` new nullability option for function/expression mappings
- other small changes

Test changes:
- as YDB requires table having primary key, many test tables updated to have primary key
- add separate test source attribute for databases with RECURSIVE CTE support (SAP HANA and YDB)
- add expected error attribute for databases without correlated subquery support (CH and YDB)
- add helper to detect if current provider returns rowcount for command (CH and YDB)
- distinguish providers that support simple column expression form providers with complex expression support in RETURNING clause (YDB supports only simple column expressions there)
- update some test data for dates to fall in supported by YDB range


Source/LinqToDB/Internal/SqlProvider/SqlProviderFlags.cs
```diff
@@ -525,6 +525,19 @@ public sealed class SqlProviderFlags
...
 
+		/// <summary>
+		/// When disabled, all conditions from INNER JOIN ON moved to WHERE except conjunction of equality predicates.
+		/// <code>
+		/// FROM T1 INNER JOIN T2 ON t1.field1 == t2.field1 AND t1.field2 == t2.field2 AND t1.field3 > 10
+		/// -- with flag:
+		/// FROM T1 INNER JOIN T2 ON t1.field1 == t2.field1 AND t1.field2 == t2.field2
+		/// WHERE t1.field3 > 10
+		/// </code>
+		/// Default: <c>false</c>.
```
viceroypenguin: discrepancy between comment and following implementation where default is `true`

MaceWindu:
Changes causing issues are moved to separate PRs to not block this one

MaceWindu:
I will try to split it into several smaller PRs to separate test changes from functional changes

MaceWindu:
Closing. Due to a lot of different changes with hard to atribute side-effects, it was splitted into several smaller PRs

PR #5131, opened 2025-09-11
Fix incorrect join optimization

MaceWindu:
Marking column as primary key breaks `TestFormattableInExpr` and `TestFormattableInExpr2` tests

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
@@ -700,6 +700,10 @@ namespace LinqToDB.Internal.SqlQuery
 			if (table1 == null || table2 == null)
 				return false;
 
+			// TODO: we should introduce better class hierarchy for tables
+			if (table1.GetType() != typeof(SqlTable) || table2.GetType() != typeof(SqlTable))
+				return false;
+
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
```diff
@@ -700,6 +700,10 @@ public static bool IsEqualTables([NotNullWhen(true)] SqlTable? table1, [NotNullW
 			if (table1 == null || table2 == null)
 				return false;
 
+			// TODO: we should introduce better class hierarchy for tables
+			if (table1.GetType() != typeof(SqlTable) || table2.GetType() != typeof(SqlTable))
```
viceroypenguin, line 704:
```suggestion
			if (table1 is not SqlTable || table2 is not SqlTable)
```

sdanyliv: `SqlRawSqlTable` and `SqlCteTable` both are `SqlTable`. Bad hierarchy.

viceroypenguin: Ah, right. 👍🏼

PR #5134, opened 2025-09-12
v6 regression: CTE

ili:
Wrong query generated in v6

v5.4.1 is OK :)

PS running tests on my project, some new issues would be in next few days, sorry but my kung-fu would be not enough now to fix em :( so only tests would be


sdanyliv:
Wow, welcome back. Will check it ASAP.

viceroypenguin (approved):

MaceWindu (approved):

PR #5135, opened 2025-09-13
v6 regression: Concat fom different sources + Count fails

ili:
Only `ConcatCountTest` fails, generates wrong SQL:
```sql
  SELECT
   COUNT(*)
  FROM
   (
    SELECT
    	*
    FROM
    	[Person] [t1]
    UNION ALL
    SELECT
    	*
    FROM
    	[Parent] [t2]
   ) [t3]
```

other tests are just added to check for similar issues


MaceWindu (approved):
My bad..

PR #5137, opened 2025-09-13
Add some test helpers/sources

MaceWindu:
Part of #4954 task

Shouldn't affect baselines, just adds some generalization to tests:
- test source attribute for providers with recursive CTE support (needed for YDB)
- correlated subquery support error attribute (needed for YDB)
- rowcount support helper method (needed for YDB)
- test sources for providers that support only simple (single column) OUTPUT expression (needed for YDB)
- add existing InsertOrUpdate and Merge attributes to tests that were missing them

### DIFF
--- a/Tests/Linq/Linq/AssociationTests.cs
+++ b/Tests/Linq/Linq/AssociationTests.cs
@@ -266,8 +266,9 @@ namespace Tests.Linq
 			}
 		}
 
+		[RequiresCorrelatedSubquery]
 		[Test]
-		public void StackOverflow2([DataSources(ProviderName.SqlCe, TestProvName.AllClickHouse)] string context)
+		public void StackOverflow2([DataSources(ProviderName.SqlCe)] string context)
 		{
 			using (var db = GetDataContext(context))
 				AreEqual(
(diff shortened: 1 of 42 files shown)


Tests/Linq/Linq/ConcurrencyTests.cs
```diff
@@ -148,14 +148,14 @@ public async ValueTask TestAutoIncrementAsync([DataSources] string context)
 			record.Stamp--;
 			record.Value = "value 3";
 			cnt = await db.UpdateOptimisticAsync(record);
-			Assert.That(cnt, Is.Zero);
+			if (!skipCnt) Assert.That(cnt, Is.Zero);
```
viceroypenguin, line 151: recommend this be two lines, not one (here and elsewhere)

Tests/Linq/Linq/AssociationTests.cs
```diff
@@ -266,17 +266,19 @@ public void StackOverflow1([NorthwindDataContext] string context)
 			}
 		}
 
+		[RequiresCorrelatedSubquery]
```
viceroypenguin, line 269: why are these above the `[Test]` in this file, but below in other files? recommend consistency.

PR #5144, opened 2025-09-14
v6 regression: nullable constant definition with as keyword fails

ili:
to bug or not to bug...

we can define nullable constant in two ways:
   * `(int?)1` - this is converted to SQL
   * `1 as int?` - this fails, and in geneal it is discussionable, those both are not the same, `as` is not `convert` it is [try_convert](https://learn.microsoft.com/ru-ru/sql/t-sql/functions/try-convert-transact-sql?view=sql-server-ver17) but in any way this is regression

### DIFF
--- a/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
@@ -2140,7 +2140,6 @@ namespace LinqToDB.Internal.Linq.Builder
 					break;
 				}
 
-				case ExpressionType.TypeAs:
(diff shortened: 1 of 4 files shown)


sdanyliv:
@ili, I decided to keep `Sql.AsSql(1 as int?)` as non-convertible. We do not support `try_convert` out of the box and never have. Making it work across multiple RDBMS is a separate (and likely large) task.

ili:
> @ili, I decided to keep `Sql.AsSql(1 as int?)` as non-convertible. We do not support `try_convert` out of the box and never have. Making it work across multiple RDBMS is a separate (and likely large) task.

This is good solution especially with exact exception! )

PR #5063, opened 2025-08-01
Fix custom type collection mapping

MaceWindu:
Fix #5057

Issue #5057, opened 2025-07-31
Cannot generate SQL in 6.0.0-rc.2 that was generated in 5.4.1


timbze:
### Description

See repro: https://github.com/timbze/Linq2Db4539/blob/master/Program.cs

5.4.1 did generate SQL properly but had a different error with this same code. I think the similar issue that was fixed in #4798 would make this work, if the SQL could be generated.

This specific thing I can fix by making my list to be Guid, but I don't understand why it should not work when the type is List<TenderId>.

```csharp
List<TenderId> tenderIds2 = new List<TenderId> {TenderId.From(Guid.NewGuid()), TenderId.From(Guid.NewGuid())};
var tender2 = await db
    .GetTable<Tender>()
    .Where(i => tenderIds2.Any(id => id == i.Id))
    .FirstOrDefaultAsync();
```

### Reproduction Steps

See code above.

### Exception details

```
Unhandled exception. LinqToDB.LinqToDBException: The LINQ expression could not be converted to SQL.
Expression:
<>c__DisplayClass0_0.tenderIds2
        .Any(id => id == i.Id)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.Build[T](IQueryExpressions& expressions)
   at LinqToDB.Internal.Linq.Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, IQueryExpressions& expressions)
   at LinqToDB.Internal.Linq.Query`1.GetQuery(IDataContext dataContext, IQueryExpressions& expressions, Boolean& dependsOnParameters)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetQuery(IQueryExpressions& expression, Boolean cache, Boolean& dependsOnParameters)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.LinqToDB.Internal.Async.IQueryProviderAsync.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken)
```

### Linq To DB version

6.0.0-rc.2

### .NET version

9.0.7

MaceWindu:
FYI https://github.com/linq2db/linq2db/wiki/Linq-To-DB-6#changes-to-scalar-types-detection

PR #4676, opened 2024-10-08
Refactored SQL generation. More bugfixes.

sdanyliv:
Fixed:

- Fix #940
- Fix #2787
- Fix #2897
- Fix #2943
- Fix #3599
- Fix #3663
- Fix #3807
- Fix #4060
- Fix #4163
- Fix #4253
- Fix #4322
- Fix #4347
- Fix #4431
- Fix #4520
- Fix #4539
- Fix #4544
- Fix #4556
- Fix #4570
- Fix #4586
- Fix #4596
- Fix #4607
- Fix #4610
- Fix #4613
- Fix #4626
- Fix #4654
- Fix #4667
- Fix #4714
- Fix #4717
- Fix #4723
- Fix #4728
- Fix #4729

Other changes

- enabled RECURSIVE keyword for CLickHouse CTE (available since 24.4 release) and enabled CH in some tests with recursive CTE
- added Access provider versioning to distinguish JET and ACE engines
- more improvements to nullability tracking
- updated default precision/scale mappings for `decimal` instead of db-specific `decimal` type defaults for MySQL/MariaDB to `(29,10)` (old default `(10, 0)`)
- providers without required SQL featues, e.g. `OUTER/LATERAL JOIN`s will throw corresponding exceptions for queries that require them instead of old behavior where it silently generated multiple queries or incorrect SQL
- `Update[Async]` overloads with table as first parameter were deprecated in favor of same overloads with lambda parameter
- [EF] change default dialect for PostgreSQL/SQLServer to `AutoDetect`
- oboleted `LinqException` and `SqlException` (especially anoying as it conflicts with `SqlClient` exception type) in favor of `LinqToDBException`
- Inplement `RESPECT NULLS` translation for FIRST_VALUE/LAST_VALUE in ClickHouse
- improvements to `Sql.IsDistinctFrom` support
- improvements to `BOOLEAN` type support in predicates

PR #5149, opened 2025-09-16
Fixed empty values selection for type-specific operations

### LINKED ISSUE #5147: Empty Query does not cast columns as data types which breaks some usages
### Description

When creating an empty query with `Enumerable.Empty<TEntity>().AsQueryable(db)`, the query that is generated is something like
```sql
SELECT NULL Prop1, NULL Prop2, NULL Prop3 WHERE 1=0
```
The issue with this is that it's not usable for type-specific operations, like SUM, AVG, etc.
It would be good if these selected properties could be cast as their respective scalar types, which would allow them to produce valid SQL when they're used in the place of a table.
```sql
SELECT CAST(NULL AS INT) Prop1, CAST(NULL AS NVARCHAR(200)) Prop2, CAST(NULL AS MONEY) Prop3 WHERE 1=0
```

### Reproduction Steps

```csharp
await using var db = new DataConnection(options);

await db.DropTableAsync<TestEntity>(tableOptions: TableOptions.DropIfExists);
await db.DropTableAsync<OtherTestEntity>(tableOptions: TableOptions.DropIfExists);
await db.CreateTableAsync<TestEntity>();
await db.CreateTableAsync<OtherTestEntity>();

var emptyQuery = Enumerable.Empty<TestEntity>().AsQueryable(db);

var testEntities = await emptyQuery
    .GroupBy(entity => new { entity.Prop1, entity.Prop2, entity.Value })
    .Select(group => new
    {
        Prop1 = group.Key.Prop1,
        Prop2 = group.Key.Prop2,
        Value = group.Sum(entity => entity.Value)
    })
    .ToListAsync();


// Helper classes
public class TestEntity
{
    public int Prop1 { get; set; }
    public int Prop2 { get; set; }
    public decimal Value { get; set; }
}
```

### Exception details

```
Unhandled exception. System.Data.SqlClient.SqlException (0x80131904): Operand data type NULL is invalid for sum operator.
   at System.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__126_0(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location ---
   at LinqToDB.Data.DataConnection.ExecuteReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.ExecuteReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.ExecuteDataReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.ExecuteDataReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.QueryRunner.ExecuteReaderAsync(CancellationToken cancellationToken)
   at LinqToDB.Internal.Linq.QueryRunner.BasicResultEnumerable`1.GetAsyncEnumerable(CancellationToken cancellationToken)+MoveNext()
   at LinqToDB.Internal.Linq.QueryRunner.BasicResultEnumerable`1.GetAsyncEnumerable(CancellationToken cancellationToken)+MoveNext()
   at LinqToDB.Internal.Linq.QueryRunner.BasicResultEnumerable`1.GetAsyncEnumerable(CancellationToken cancellationToken)+MoveNext()
   at LinqToDB.Internal.Linq.QueryRunner.BasicResultEnumerable`1.GetAsyncEnumerable(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetForEachAsync(Action`1 action, CancellationToken cancellationToken)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetForEachAsync(Action`1 action, CancellationToken cancellationToken)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetForEachAsync(Action`1 action, CancellationToken cancellationToken)
   at LinqToDB.Async.AsyncExtensions.ToListAsync[TSource](IQueryable`1 source, CancellationToken token)
```

### Regression?

_No response_

### Linq To DB version

5.4.1, 6.0.0-rc3

### .NET version

8

### Database

SQL Server 2019

### Database provider

System.Data.SqlClient

### Operating system

_No response_

### Configuration

_No response_

AidenFuller:
Fixes #5147


Tests/Linq/Linq/EnumerableSourceTests.cs
```diff
@@ -830,6 +830,26 @@ from t in table
 					table.GetCacheMissCount().ShouldBe(cacheMiss);
 			}
 		}
+		
+		[Test]
+		public void EmptyValuesWithTypeSpecificUsage([DataSources(TestProvName.AllClickHouse, TestProvName.AllAccess, ProviderName.DB2, TestProvName.AllSybase, TestProvName.AllSybase, TestProvName.AllInformix)] string context, [Values(1, 2)] int iteration)
```
AidenFuller: Not 100% sure if this is the right place to put this test, nor if the name of the test is acceptable. This definitely covers the issue I was seeing though, so tried to make it consistent given the fact that the test above this already covers the baseline functionality

sdanyliv: Looks good.

viceroypenguin:
@AidenFuller failing tests; ensure the failing tests are not affected by your change

AidenFuller:
@viceroypenguin looks like they are related to these changes, I'll have a look in the morning unless one of you guys spot a quick fix for it and end up doing it yourself in the meantime

AidenFuller:
@viceroypenguin I could be wrong, but I feel like this is actually an issue with the specific test that's failing ( #5057 / #5063 ). Should the TenderId scalar type have an underlying DataType defined? Adding `ms.SetDataType(typeof(TenderId), DataType.Guid)` is fixing the issue for me locally, but I'm not sure if this is correct and whether it's expected that all scalar types are should have an underlying data type

Edit: worst case is that we just revert to the existing behaviour only for properties/columns whose data types are undefined, but that feels like it could be confusing from an implementation perspective if things go wrong and you end up with a SQL error.

AidenFuller:
I've just updated the problematic test case to include the data type, as the more I look at it, the more I think it's definitely the right solution, and not that the functionality itself is broken. Let me know if there's a reason why this is not the case though.

viceroypenguin:
@AidenFuller based on <https://github.com/timbze/Linq2Db4539/blob/master/TenderId.cs>, that is not the correct solution. the test needs to reflect the code that actually worked in 5.4.1 and should continue to work. if you look at the test code that is provided, it does not provide such a `SetDataType()`. You will need to address this without changing the test code.

AidenFuller:
@viceroypenguin Fair enough. I missed the specific reason as to why the empty values was being used for SQLite in the way that it is in that test, and knowing that extra context now, it makes sense that the existing behaviour should work. I believe I've got a fix, just not sure if it matches the convention, because I'm not totally content that it's clean. Perhaps one of you have a suggestion for how it could be tweaked if you feel it's necessary.

Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteSqlBuilder.cs
```diff
@@ -219,7 +219,7 @@ protected override void BuildSqlValuesTable(SqlValuesTable valuesTable, string a
 
 				StringBuilder.AppendLine();
 				AppendIndent();
-				BuildEmptyValues(valuesTable);
+				BuildEmptyValues(valuesTable, false);
```
viceroypenguin:
```suggestion
				BuildEmptyValues(valuesTable, useTypedExpression: false);
```

AidenFuller:
Am I right to assume those CI pipeline failures are the result of an external issue (transient or otherwise) and not based on this PR's content?

MaceWindu:
Only test-all pipeline configured for external prs

Source/LinqToDB/DataProvider/PostgreSQL/PostgreSQLFactory.cs
```diff
@@ -23,7 +23,7 @@ public override IDataProvider GetDataProvider(IEnumerable<NamedValue> attributes
 				_               => PostgreSQLVersion.AutoDetect,
 			};
 
-			return PostgreSQLTools.GetDataProvider();
+			return PostgreSQLTools.GetDataProvider(version);
```
AidenFuller, line 26: @sdanyliv just an FYI, I boyscouted this change as part of this PR which I guess was intended to be included in #4676. Not sure if you care whether it's in a separate PR for scoping purposes.

Tests/Linq/Linq/EnumerableSourceTests.cs
viceroypenguin:
```suggestion
		public void EmptyValuesWithTypeSpecificUsage([DataSources(TestProvName.AllClickHouse, TestProvName.AllAccess, ProviderName.DB2, TestProvName.AllSybase, TestProvName.AllInformix)] string context, [Values(1, 2)] int iteration)
```

MaceWindu:
Thanks!

PR #5151, opened 2025-09-29
Add .NET 10 support

viceroypenguin:
EF Core 10 is not possible until Pomelo MySQL driver is updated


MaceWindu:
>  EF Core 10 is not possible until Pomelo MySQL driver is updated

pomelo tests just need to be disabled. It could take some time for them to release new version

Source/LinqToDB/Data/CommandInfo.cs
```diff
@@ -966,19 +966,19 @@ public async ValueTask<bool> MoveNextAsync()
 			}
 		}
 
-		Task<T[]> ReadAsArrayAsync<T>(DbDataReader rd, CancellationToken cancellationToken)
+		async Task<T[]> ReadAsArrayAsync<T>(DbDataReader rd, CancellationToken cancellationToken)
 		{
-			return new ReaderAsyncEnumerable<T>(this, rd).ToArrayAsync(cancellationToken: cancellationToken);
+			return await new ReaderAsyncEnumerable<T>(this, rd).ToArrayAsync(cancellationToken: cancellationToken).ConfigureAwait(false);
```
jods4: What's the goal with this kind of changes?

jods4: Mmm... reading further down I think I'm starting to get a clue.
.net 10 has introduced its own `ToArrayAsync` extension that we're calling instead... and I guess this is an attempt of adding `.ConfigureAwait(false)` that we had in our own `ToArrayAsync`?

If I'm correct, this still feels unnecessary to me, one reason being because it's a private method.

viceroypenguin: `.net10`'s `ToArrayAsync()` returns a `ValueTask` instead of a `Task`. probably a better solution here, was just getting through it to get to stable state before doing deeper dive into viability of changes.

Source/LinqToDB/Internal/Async/AsyncEnumerableExtensions.cs
```diff
@@ -29,7 +29,10 @@ static class AsyncEnumerableExtensions
 		///     The task result contains a <see cref="List{T}" /> that contains elements from the input sequence.
 		/// </returns>
 		public static async Task<List<T>> ToListAsync<T>(
-			this IAsyncEnumerable<T> source,
+#if !NET10_0_OR_GREATER
```
jods4: Initially confused by this change, I figured this is because .net 10 introduced its own extension methods on `IAsyncEnumerable`.

If so, why not remove the whole method, or even most of this file, from the .net 10 build?

I don't like that linq2db runs different code in different runtime, esp. when code has subtle differences that could be bug-inducing. `ToListAsync` and co. on .net 10 do not contain the `.ConfigureAwait(false)` calls that our methods have (and for a good reason: arguably we shouldn't have them either).
EDIT: even more so when the difference is not subtle, the return type is not the same, `ValueTask` vs `Task`. I really think we shouldn't have two different implementations based on runtime :(

Shouldn't we handle this as we do for other new C#/BCL features? 1. Remove the code completely from linq2db; 2. Reference the package-provided version in older .net runtimes (that package should be `System.Linq.AsyncEnumerable`, see also MS guidance here https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable).

viceroypenguin: > If so, why not remove the whole method, or even most of this file, from the .net 10 build?

Removing would be a binary-breaking change. If an existing consumer compiles against linq2db 5.4.1 and relies on this method, then when they upgrade, if this method is not available in the binary, then they get a `MethodNotFoundException`.

That said, I think in this case, we can probably forgo it due to the massive amount of other binary breaking changes they will encounter from 5.4.1 to 6.0.0 anyway.

For the record, we cannot do a method forward here due to different return type (`Task<>` vs `ValueTask<>`)

Source/LinqToDB/Data/CommandInfo.cs
jods4: @viceroypenguin ah, I missed that difference.

Ideally, it's probably better to update at least our private/internal methods to propagate the `ValueTask` but that might have a ripple effect on their callers...

`await` adds IL bloat and a small perf penality.

Source/LinqToDB/Internal/Async/AsyncEnumerableExtensions.cs
viceroypenguin: Oh. Transitive dependency.

App relies on Package A relies on linq2db. Package A compiles for net8.0 and uses the method. App compiles for net10.0 and imports the version of linq2db that doesn't include the method, MethodNotFoundException. 

Though if we change our return type to match, we could do the method forward.

Source/LinqToDB/Data/CommandInfo.cs
viceroypenguin: This particular method must be a `Task`, due to [`ReadMultipleResultSetsAsync()`](https://github.com/linq2db/linq2db/blob/cfa37b497c53d3ccf234ff07530afcdf7de95d38/Source/LinqToDB/Data/CommandInfo.cs#L1016-L1020), which cannot be polymorphic over types as a `ValueTask<>`

Source/LinqToDB/Internal/Async/AsyncEnumerableExtensions.cs
viceroypenguin: > MS guidance here https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable).

This is not viable in our context, as relying on that package will create package version conflicts on `net9.0` and lower TFMs due to their reliance on other System. and Microsoft. packages that will remain at the TFM version.

jods4: > Oh. Transitive dependency.
> 
> App relies on Package A relies on linq2db. Package A compiles for net8.0 and uses the method. App compiles for net10.0 and imports the version of linq2db that doesn't include the method, MethodNotFoundException.
> 
> Though if we change our return type to match, we could do the method forward.

As you said, linq2db 6.0 contains a ton of breaking changes anyway. In that scenario if App upgrades to linq2db 6 and one of its dependencies is built for linq2db 5 they will most likely encounter bugs anyway, so they shouldn't do that unless the dep. gets upgraded to 6 as well.

The large 6.0 release is our opportunity to get rid of this code, otherwise it will stick with us for a long time. I agree with your arguments, we shouldn't remove it in a minor release.

Source/LinqToDB/Data/CommandInfo.cs
jods4: Right... but the same comments apply to `ReadAsListAsync` and `ReadFirstOrDefaultAsync` and then they're all `ValueTask` and `ReadMultipleResultSetsAsync` can use a `ValueTask` instead of `Task` for its reflection invoke?

jods4: And if we go all the way wrt to my other comment (use `System.Linq.AsyncEnumerable` in all runtimes) then those one-liners are just useless and we should be able to directly put `AsyncEnumerable.ToArrayAsync()` in the reflection delegate?

viceroypenguin: > And if we go all the way wrt to my other comment (use System.Linq.AsyncEnumerable in all runtimes)

As stated in the other comment, this is not possible. Also they cannot use `ValueTask`, because `ValueTask` is not polymorphic.

jods4: In my comment above I pointed out that all 3 usages inside `ReadMultipleResultSetsAsync` reflection code become `ValueTask` so there's no mix of `Task` and `ValueTask` and it should possible? Or am I missing something else, what do you mean by `ValueTask` is not polymorphic?

viceroypenguin: Look specifically at lines 1016-1020 of `ReadMultipleResultSetsAsync`. Notice that `InvokeExt` is being used as a `Task`, not a `Task<>`. This way it can be `await`ed without knowing what generic type the `Task` is holding. Then `Invoke` is called using reflection, so that again, the holding type does not have to be known. 

This is not possible with `ValueTask`, since as a structure, `ValueTask<>` does not inherit from `ValueTask`.

jods4: Right, I missed the fact that `ReadMultipleResultSetsAsync` casts `Task<T>` into `Task`, which is impossible with struct values.

The lighter weight way to get `Task<T>` out of `ValueTask<T>` might be to just do `.AsTask()`, rather than using `async` machinery?

That being said, the (looped) upcast followed by reflection access to `Task<T>.Result` with `task.GetType().GetProperty("Result").GetValue(task, null)` is neither elegant nor performant. As the value is going to be cast into an `object` anyway for the `member.SetValue()` call, why don't we return `ValueTask<object>` (or `Task<object>`) for those methods?
Sure, it's less precise type-wise, but its consumption is reflection-free, which is much better.

What I'm suggesting:
```csharp
// For each of Read[AsArray|AsList|FirstOrDefault]Async,
// the minimal change is just to type the returned value as object in signature
async ValueTask<object> ReadAsArrayAsync<T>(rd, cancel)
{
  return await new ReaderAsyncEnumerable<T>(this, rd)
    .ToArrayAsync(cancel)
    .ConfigureAwait(false);
}

async Task<T> ReadMultipleResultSetsAsync(..)
{
  // ...
  var task = genericMethod.InvokeExt<ValueTask<object>>(this, [ rd, cancellationToken ]);
  var result = await task.ConfigureAwait(false);
  member.SetValue(result, value);
  // ...
}
```

viceroypenguin: Addressed

Source/LinqToDB/Internal/Async/AsyncEnumerableExtensions.cs
viceroypenguin: Addressed

Directory.Packages.props
MaceWindu, line 1: I would also recommend to test Npgsql 10 prerelease version. Sometimes our tests find regressions on major release

Tests/EntityFrameworkCore/Tests/ConvertorTests.cs
```diff
@@ -47,7 +47,8 @@ public void TestToList([EFDataSources] string provider)
 			var id = new Id<SubDivision, long>?(0L.AsId<SubDivision>());
 			var result2 = ctx.Subdivisions.ToLinqToDB().Where(s => s.Id == id).ToArray();
 
-			var ids = new[] {1L.AsId<SubDivision>(), 2L.AsId<SubDivision>(),};
+			// TODO: Undo and fix properly
```
viceroypenguin: TBD: Fix

Source/LinqToDB/Internal/Async/AsyncEnumerableExtensions.cs
MaceWindu: Could we just make this type available to .net < 10? Like polyfill

MaceWindu (dismissed):
let's wait till 10.0.0 nugets for .net and npgsql released to not break nuget builds

Build/Azure/pipelines/templates/nuget-job.yml
```diff
@@ -215,14 +215,6 @@ jobs:
     displayName: Build nugets for Azure Artifacts
     condition: and(succeeded(), ne(variables['Build.SourceBranchName'], variables['release_branch']))
 
-  - task: CmdLine@2
```
MaceWindu, line 181: this is not correct. You need to remove cmd parameter only

MaceWindu:
Waiting for
- [x] SDK release
- [ ] Npgsql 10

MaceWindu:
looks like npgsql regressed in final release

PR #5172, opened 2025-11-03
Fix ToString translation

### LINKED ISSUE #5152: Scalar Type Detection not working (6.0.0-rc.3)
### Description

In #4902 we already discussed problems when using custom classes mapped to a scalar type (nvarchar).
With your help this was fixed and worked fine in Linq2DB 5.x and 6.-Preview4.

Probably since the [change to Scalar Type Detection](https://github.com/linq2db/linq2db/wiki/Linq-To-DB-6#changes-to-scalar-types-detection) in RC.3 I now encounter multiple new problems with queries that previously worked.

It looks like Linq2DB introduces unnecessary CAST expressions in both SET and WHERE clauses, which leads to:
- invalid SQL (broken SET clause)
- performance issues (unnecessary casts preventing index usage)

**Expected Behavior**

Update statements should map directly to the nvarchar column without extra CAST.
ScalarType attribute should prevent Linq2DB from casting unnecessarily.

**Actual Behavior**
Problem 1: Broken SET clause

Produces invalid SQL:
```sql
UPDATE [SampleClass]
SET CAST([SampleClass].[MyString] AS NVarChar(4000)) = ...
```
Problems:
1. The CAST on the left side of SET is invalid SQL.
2. A cast isn’t even necessary because the column already has [Column(DataType = DataType.NVarChar, Length = 50)].

---
Problem 2: Even when we say, i am not allowed to call ToString() in the Setter any more, even then in the SQL generated contains casts in both WHERE and SET:
```sql
UPDATE
	[SampleClass]
   SET
	[SampleClass].[MyString] = CAST(Replace(Replace([SampleClass].[MyString], @param1, NULL), CAST(@param2 AS NVarChar(255)), NULL) AS NVarChar(Max)) 
	WHERE
	   CAST([SampleClass].[MyString] AS NVarChar(4000)) LIKE @param3 ESCAPE N'~'
```

### Reproduction Steps

I wrote a UnitTest to reproduce the problem in your Tests.Playground.
Just replace the TestTemplate.cs with the following code, and you should be able to reproduce the error.
Both Queries worked fine in LinqToDb 5.4.1 and created SQL-Queries without Casts to NvarChar.
The new "ScalarType" Attribute or 
```c#
			schema.AddScalarType(typeof(MySpecialBaseClass), null, true, DataType.NVarChar);
			schema.AddScalarType(typeof(MySpecialStringClass), null, true, DataType.NVarChar);
```
 don't change the Queries that are generated in my Environment.

```c#
using System;
using System.Diagnostics.CodeAnalysis;
using System.Linq;

using LinqToDB;
using LinqToDB.Mapping;

using NUnit.Framework;
using LinqToDB.Tools;

using Shouldly;

namespace Tests.Playground
{
	[ScalarType]
	public abstract class MySpecialBaseClass : IConvertible, IEquatable<MySpecialBaseClass>
	{
		[LinqToDB.ExpressionMethod(nameof(ValueCast))]
		public string Value { get; set; }

		static System.Linq.Expressions.Expression<Func<MySpecialBaseClass, string>> ValueCast()
		{
			return awl => awl.ToString();
		}

		protected MySpecialBaseClass(string value)
		{
			if (value == null)
			{
				throw new ArgumentNullException(nameof(value));
			}

			Value = value;
		}

		public static bool operator ==(MySpecialBaseClass? leftSide, string? rightSide)
			=> leftSide?.Value == rightSide;

		public static bool operator !=(MySpecialBaseClass? leftSide, string? rightSide)
			=> leftSide?.Value != rightSide;

		/// <inheritdoc />
		public override string ToString() => Value;

		/// <inheritdoc />
		public override int GetHashCode() => Value.GetHashCode();

		/// <inheritdoc />
		public override bool Equals(object? obj)
		{
			if (obj == null)
			{
				return false;
			}

			if (obj is string str)
			{
				return Value.Equals(str);
			}

			if (obj.GetType() == GetType())
			{
				return string.Equals(((MySpecialBaseClass)obj).Value, Value);
			}

			return base.Equals(obj);
		}

		/// <inheritdoc />
		public bool Equals(MySpecialBaseClass? obj)
		{
			if (obj == null && Value == null)
			{
				return true;
			}

			if (obj == null)
			{
				return false;
			}

			if (obj.GetType() != GetType())
			{
				// Falls die Auswahllistentypen nicht zusammen passen.
				return false;
			}

			return string.Equals(obj.Value, Value);
		}

		#region IConvertible
		public TypeCode GetTypeCode() => TypeCode.String;

		public string ToString(IFormatProvider? provider) => Value;

		/// <summary>
		/// if TargetType <see cref="MySpecialBaseClass"/> it returns targettype, else stringAuswahllisten object
		/// </summary>
		public object ToType(Type conversionType, IFormatProvider? provider)
		{
			if (conversionType.IsSubclassOf(typeof(MySpecialBaseClass))
				|| conversionType == typeof(MySpecialBaseClass))
			{
				return this;
			}

			return Value;
		}

		public bool     ToBoolean(IFormatProvider?  provider) { throw new NotImplementedException(); }
		public char     ToChar(IFormatProvider?     provider) { throw new NotImplementedException(); }
		public sbyte    ToSByte(IFormatProvider?    provider) { throw new NotImplementedException(); }
		public byte     ToByte(IFormatProvider?     provider) { throw new NotImplementedException(); }
		public short    ToInt16(IFormatProvider?    provider) { throw new NotImplementedException(); }
		public ushort   ToUInt16(IFormatProvider?   provider) { throw new NotImplementedException(); }
		public int      ToInt32(IFormatProvider?    provider) { throw new NotImplementedException(); }
		public uint     ToUInt32(IFormatProvider?   provider) { throw new NotImplementedException(); }
		public long     ToInt64(IFormatProvider?    provider) { throw new NotImplementedException(); }
		public ulong    ToUInt64(IFormatProvider?   provider) { throw new NotImplementedException(); }
		public float    ToSingle(IFormatProvider?   provider) { throw new NotImplementedException(); }
		public double   ToDouble(IFormatProvider?   provider) { throw new NotImplementedException(); }
		public decimal  ToDecimal(IFormatProvider?  provider) { throw new NotImplementedException(); }
		public DateTime ToDateTime(IFormatProvider? provider) { throw new NotImplementedException(); }
		#endregion
	}

	[ScalarType]
	public class MySpecialStringClass : MySpecialBaseClass
	{
		public const string Example = "Example";

		public static ISet<MySpecialStringClass> AllValues = new HashSet<MySpecialStringClass>
		{
			new MySpecialStringClass(Example)
		};

		public MySpecialStringClass(string value)
			: base(value)
		{
		}

		[return: NotNullIfNotNull(nameof(value))]
		public static implicit operator MySpecialStringClass?(string? value)
		{
			if (value == null)
			{
				return null;
			}

			return AllValues.FirstOrDefault(a => string.Equals(a.Value, value, StringComparison.OrdinalIgnoreCase)) ?? new MySpecialStringClass(value);
		}

		[return: NotNullIfNotNull(nameof(auswahlliste))]
		public static implicit operator string?(MySpecialStringClass? auswahlliste)
			=> auswahlliste?.Value;
	}

	[TestFixture]
	public class TestTemplate : TestBase
	{
		[Table]
		class SampleClass
		{

			[Column(DataType = DataType.NVarChar, Length = 50)] public MySpecialStringClass? MyString { get; set; }

			[Column] public string? NormalString { get; set; }
		}

		[Test]
		public void SampleSelectTest([IncludeDataSources(TestProvName.AllSQLite)] string context)
		{
			using var db     = GetDataContext(context);
			using var table  = db.CreateLocalTable<SampleClass>();

			// Problem 1: This produces an Update-Statement with a Broken Set Clause. This has 2 Problems. First a Cast on the Field in Set never works.
			// And Secondly a Cast isn't even necessary. We have an Attribute telling the Column is of Type NVarChar(50). And we have the ScalarType Attributes on the Base-Class and the implementing Class as well. 
			/* UPDATE
			   	[SampleClass]
			   SET
			   	CAST([SampleClass].[MyString] AS NVarChar(4000)) = ... */
			var queryBrokenSet = (from sample in table
			             // I Have to cast with ToString() or use .Value so I can use string.Contains.
			             where sample.MyString!.ToString().Contains(MySpecialStringClass.Example)
			             select sample)
			            .Set(sample => sample.MyString!.ToString(), // ToString() Or Value -> Broken Set Clause with cast. Works in Linq2db 5.
			                 sample => sample.MyString!.ToString().Replace(";" + MySpecialStringClass.Example, null).Replace(MySpecialStringClass.Example, null))
			            .Update();

			// Problem 2: This produces an Update-Statement with a correct Set Clause. But there are Casts everywhere, in the Where Statement and in the Value Part of the Set-Clause.
			// This is also a performance Problem, because the Casts prevent the use of Indexes and casts the whole nvarchar-column to nvarchar again.
			/* UPDATE
				[SampleClass]
			   SET
				[SampleClass].[MyString] = CAST(Replace(Replace([SampleClass].[MyString], @param1, NULL), CAST(@param2 AS NVarChar(255)), NULL) AS NVarChar(Max)) 
				WHERE
				   CAST([SampleClass].[MyString] AS NVarChar(4000)) LIKE @param3 ESCAPE N'~'
			 */
			var queryWithCasts = (from sample in table
			             // I Have to cast with ToString() or use .Value so I can use string.Contains.
			             where sample.MyString!.ToString().Contains(MySpecialStringClass.Example)
			             select sample)
			            .Set(sample => sample.MyString,
			                 sample => (MySpecialStringClass)sample.MyString!.ToString().Replace(";" + MySpecialStringClass.Example, null).Replace(MySpecialStringClass.Example, null))
			            .Update();
		}
	}
}

```

### Exception details

_No response_

### Regression?

5.4.1

### Linq To DB version

6.0.0-rc.3

### .NET version

.NET Framework 4.8

### Database

SQL Server 16.0

### Database provider

System.Data.SqlClient

### Operating system

Windows

### Configuration

ScalarType attribute is set on both base and implementing class. Column attribute specifies NVarChar(50).

MaceWindu:
Fix #5152

- don't translate ToString if source value is already database string
- don't generate CAST if types differ only by SystemType
- fix missing Oracle keyword escaping for `SAMPLE`
- generate expression conversion using converter functionality

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/Translation/ProviderMemberTranslatorDefault.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Translation/ProviderMemberTranslatorDefault.cs
@@ -98,7 +98,7 @@ namespace LinqToDB.Internal.DataProvider.Translation
 			// ToString called on custom type already mapped to text-based db type or string
 			if (fromType.IsTextType())
 			{
-				return objPlaceholder;
+				return objPlaceholder.WithType(typeof(string));
 			}
 
 			DbDataType toType;


Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.QueryBuilder.cs
```diff
@@ -571,7 +572,12 @@ public Expression ToReadExpression(
 
 					if (placeholder.Type != readerExpression.Type)
 					{
-						readerExpression = Expression.Convert(readerExpression, placeholder.Type);
+						var convertExpression = MappingSchema.GetConvertExpression(readerExpression.Type, placeholder.Type, false, true, ConversionType.FromDatabase);
+
+						if (convertExpression != null)
+							readerExpression = InternalExtensions.ApplyLambdaToExpression(convertExpression, readerExpression);
+						else
+							throw new InvalidOperationException($"No conversions defined from {readerExpression.Type} to {placeholder.Type}");
```
viceroypenguin:
```suggestion
						if (convertExpression is null)
							throw new InvalidOperationException($"No conversions defined from {readerExpression.Type} to {placeholder.Type}");

						readerExpression = InternalExtensions.ApplyLambdaToExpression(convertExpression, readerExpression);
```

viceroypenguin (dismissed):

Issue #5193, opened 2025-11-21
Name referenced from where it is out of scope in joined subquery in `6.0.0-rc.3`


thomas-hori:
### Description

`linq2db` 6 tries to reference a table column from where the name is out of scope (inside a nested query it is joining to that table, instead of in the `ON` clause), while `linq2db` 5 handles this correctly.

Workaround is to refactor the query expression.

CC @coudy.

### Reproduction Steps

Reproduction code with workarounds: https://github.com/megaslice/P1-linq2db-repros/blob/master/SpecificReproductions/JoinNameScopeTest.cs

SQL DDL to set up the initial state: https://github.com/megaslice/P1-linq2db-repros/blob/master/setup.sql

MSTest test suite with `linq2db6` and `linq2db5` build configurations, with this and other repros: https://github.com/megaslice/P1-linq2db-repros/tree/master

### Exception details

Test failure output with stack trace and logged SQL: https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db6.trx#L363-L422

### Regression?

Success in 5.4.1.9: https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db5.trx#L47-L118

### Linq To DB version

6.0.0-rc.3

### .NET version

9.0.11

### Database

SQL Server 2019

### Database provider

Microsoft.Data.SqlClient 6.1.3

### Operating system

Windows

Issue #5192, opened 2025-11-21
`InsertWithOutputAsync` on a query of an aggregate of a left join to a scalar function incorrectly deletes a `COALESCE` in `6.0.0-rc.3`


thomas-hori:
### Description

See title (reading the repro code might explain it a bit better).  This results in a database error due to inserting `NULL` into a `NOT NULL` column.

Workaround is to add a `Sql.AsNullable(…).

CC @coudy.

### Reproduction Steps

Reproduction code with workarounds: https://github.com/megaslice/P1-linq2db-repros/blob/master/SpecificReproductions/InsertOnNextCounterNullabilityTest.cs

SQL DDL to set up the initial state: https://github.com/megaslice/P1-linq2db-repros/blob/master/setup.sql

MSTest test suite with `linq2db6` and `linq2db5` build configurations, with this and other repros: https://github.com/megaslice/P1-linq2db-repros/tree/master

### Exception details

Test failure output with stack trace and logged SQL: https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db6.trx#L94-L154

### Regression?

Success in 5.4.1.9: https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db5.trx#L187

### Linq To DB version

6.0.0-rc.3

### .NET version

9.0.11

### Database

SQL Server 2019

### Database provider

Microsoft.Data.SqlClient 6.1.3

### Operating system

Windows

MaceWindu:
Thanks for a bunch of good reports!

Issue #5191, opened 2025-11-21
Full-text search (SQL Server) of multiple joined entire tables broken in `6.0.0-rc.3`


thomas-hori:
### Description

A sequence of `Sql.Ext.SqlServer().Contains(…)` calls joined with `||`, where each references an entire (joined) table, fails in linq2db `6.0.0-rc.3` but succeeds in linq2db 5.  Searches of inner-joined tables get omitted from the query, while those of left-joined tables error out and fail to generate SQL.

Workaround is to exhaustively pass in all of the full-text indexed columns instead of passing in the table object.

CC @coudy.

### Reproduction Steps

Reproduction code with workarounds: https://github.com/megaslice/P1-linq2db-repros/blob/master/SpecificReproductions/FullTextSearchesTest.cs

SQL DDL to set up the initial state: https://github.com/megaslice/P1-linq2db-repros/blob/master/setup.sql

MSTest test suite with `linq2db6` and `linq2db5` build configurations, with this and other repros: https://github.com/megaslice/P1-linq2db-repros/tree/master

### Exception details

Test failure output with stack trace (left join): https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db6.trx#L67-L93

Test failure output with stack trace and logged SQL code (inner join): https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db6.trx#L296-L336

### Regression?

Success in 5.4.1.9: https://github.com/megaslice/P1-linq2db-repros/blob/2949e1437841dab02bc4a2a4a9c7feba8b2cd06e/linq2db5.trx#L159-L160

### Linq To DB version

6.0.0-rc.3

### .NET version

9.0.11

### Database

SQL Server 2019

### Database provider

Microsoft.Data.SqlClient 6.1.3

### Operating system

Windows

Issue #5176, opened 2025-11-06
ClickHouse temp table inserts broken in 6.0.0-rc3


erictilley:
### Description

In 6.0.0-rc3, certain ClickHouse temp table inserts fail. This happens because the insert statement is incorrectly generated with the form `INSERT INTO ... DEFAULT VALUES SELECT ...`, rather than just `INSERT INTO ... SELECT ...`.

### Reproduction Steps

Reproduction code:

```
[Test]
public static async Task Test()
{
    var dataOptions = new DataOptions().UseClickHouse(CONNECTION_STRING, ClickHouseProvider.Octonica);
    await using var connection = new DataConnection(dataOptions);

    var tempRows =
        connection.FromSql<Projection1>("(with Rows as (select 1 as i1, 2 as i2) select * from Rows)");

    await using var tempTable = await connection.CreateTempTableAsync<Projection2>();

    await tempRows
        .Select(t1 => new Projection2(t1.i1))
        .InsertAsync(tempTable, row => row);
}

public sealed record Projection1(int i1, int i2);

public sealed record Projection2(int i);
```

SQL generated by Linq2DB:

```
CREATE TEMPORARY TABLE Projection2
(
	i Int32
)
ENGINE = Memory()

INSERT INTO Projection2 DEFAULT VALUES
SELECT
	*
FROM
	(with Rows as (select 1 as i1, 2 as i2) select * from Rows) t1
```

### Exception details

```
Octonica.ClickHouseClient.Exceptions.ClickHouseServerException : DB::Exception: Syntax error: failed at position 25 (DEFAULT) (line 1, col 25): DEFAULT VALUES
SELECT
	*
FROM
	(with Rows as (select 1 as i1, 2 as i2) select * from Rows) t1
. Expected one of: token, Dot, OpeningRoundBracket, FROM INFILE, SETTINGS, VALUES, FORMAT, SELECT, WITH
   at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
   at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteNonQueryAsync(CancellationToken cancellationToken)
   at LinqToDB.Data.DataConnection.ExecuteNonQueryAsync(CancellationToken cancellationToken)
[...]

PR #5197, opened 2025-11-21
Regression tests

MaceWindu:
- #5194
- #5193
- #5192
- #5191
- #5176


sdanyliv:
@MaceWindu, ClickHouse temp table works with current codebase.

PR #5213, opened 2025-11-26
possibility to disable convert of timespan member accesses (which only…

jogibear9988:
… works in some special cases), so outside conversion via ExtensionAttribute could be used

### DIFF
--- a/Source/LinqToDB/Internal/Linq/Builder/Visitors/ExposeExpressionVisitor.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/Visitors/ExposeExpressionVisitor.cs
@@ -623,33 +623,35 @@ namespace LinqToDB.Internal.Linq.Builder.Visitors
 				return expr;
 			}
 
-			if (node.Member.DeclaringType == typeof(TimeSpan) && node.Expression != null)
+			if (!DataContext.Options.SqlOptions.DisableBuiltInTimeSpanConversion)
 			{
-				switch (node.Expression.NodeType)
+				if (node.Member.DeclaringType == typeof(TimeSpan) && node.Expression != null)
 				{
-					case ExpressionType.Subtract:
-					case ExpressionType.SubtractChecked:
+					switch (node.Expression.NodeType)
+					{
+						case ExpressionType.Subtract:
+						case ExpressionType.SubtractChecked:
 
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Internal/Linq/Builder/Visitors/ExposeExpressionVisitor.cs
```diff
@@ -596,33 +596,36 @@ public Expression ConvertNull(MemberExpression expression)
...
-							case "TotalSeconds"     : datePart = Sql.DateParts.Second;      break;
-							case "TotalMinutes"     : datePart = Sql.DateParts.Minute;      break;
-							case "TotalHours"       : datePart = Sql.DateParts.Hour;        break;
-							case "TotalDays"        : datePart = Sql.DateParts.Day;         break;
-							default                 : return null;
-						}
+							switch (node.Member.Name)
+							{
+								case "TotalMilliseconds": datePart = Sql.DateParts.Millisecond; break;
+								case "TotalSeconds": datePart = Sql.DateParts.Second; break;
```
igor-tkachev, line 640: Keep formatting.

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -23,6 +23,12 @@ namespace LinqToDB.Common
 	[PublicAPI]
 	public static class Configuration
 	{
+		/// <summary>
+		/// Disables built-in datediff and timespanpart call conversions in ExposeExpressionVisitor to allow external conversion via ExtensionAttribute.
+		/// </summary>
+		[Obsolete("This API is planned for removal when pull #3994 is merged"), EditorBrowsable(EditorBrowsableState.Never)]
+		public static bool DisableLegacySqlBuilderDateDiffCalls;
```
igor-tkachev: All the settings related to SQL generation should be implemented as options. Move this flag to `Configuration.Sql` and add it to `SqlOptions` record.

igor-tkachev (changes requested):

Source/LinqToDB/Common/Configuration.cs
jogibear9988: I moved it, but in the record I could not hide it and also not mark it as obsolete

PR #5252, opened 2025-12-14
Vector t4 support

igor-tkachev:
T4 support for Vector type.

### DIFF
--- a/Tests/Linq/SchemaProvider/SqlServerTests.cs
+++ b/Tests/Linq/SchemaProvider/SqlServerTests.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Linq;
 
-using LinqToDB;
 using LinqToDB.Data;
 using LinqToDB.Internal.DataProvider.SqlServer;
 using LinqToDB.SchemaProvider;
@@ -32,8 +31,11 @@ namespace Tests.SchemaProvider
 			var jsonColumn = table.Columns.First(c => c.ColumnName == "jsonDataType");
 			var type       = preferProviderSpecificTypes && db.DataProvider is SqlServerDataProvider { Adapter.SqlJsonType: not null } ? typeof(SqlJson) : typeof(string);
 
-			Assert.That(jsonColumn.Length,     Is.Null);
-			Assert.That(jsonColumn.SystemType, Is.EqualTo(type));
+			using (Assert.EnterMultipleScope())
+			{
+				Assert.That(jsonColumn.Length,     Is.Null);
+				Assert.That(jsonColumn.SystemType, Is.EqualTo(type));
+			}
 		}
 
 		[Test]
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerSchemaProvider.cs
```diff
@@ -543,5 +543,26 @@ protected override string BuildTableFunctionLoadTableSchemaCommand(ProcedureSche
...
+			{
+				list.Add(new DataTypeInfo
+				{
+					TypeName         = "vector",
+					DataType         = typeof(float[]).FullName!,
+					CreateFormat     = "vector({0})",
+					CreateParameters = "length",
+					ProviderSpecific = true,
+					// 36 is the provider type code for the SQL Server 'vector' type (see https://learn.microsoft.com/en-us/sql/relational-databases/vector-database/vector-type?view=sql-server-ver17#data-type-mapping)
+					ProviderDbType   = 36
```
viceroypenguin, line 575:
```suggestion
					ProviderDbType   = 36,
```

viceroypenguin (dismissed):

NuGet/SqlServer/LinqToDB.SqlServer.ttinclude
```diff
@@ -62,9 +62,11 @@ void DoGenerateSqlServerFreeText()
 	);
 }
 
+static SqlServerProvider DefaultSqlServerProvider = SqlServerProvider.SystemDataSqlClient;
```
MaceWindu, line 65: 1. it will not work without adding MDS client files to nuget
2. what SDS client missing that we even need it?

Tests/Tests.T4/Databases/SqlServer.2025.generated.cs
```diff
@@ -0,0 +1,1404 @@
...
+		[Column("nvarchar_max_DataType",    DbType="nvarchar(max)",     Length=int.MaxValue),                  Nullable            ] public string?           NvarcharMaxDataType      { get; set; } // nvarchar(max)
+		[Column("varchar_max_DataType",     DbType="varchar(max)",      Length=int.MaxValue),                  Nullable            ] public string?           VarcharMaxDataType       { get; set; } // varchar(max)
+		[Column("varbinary_max_DataType",   DbType="varbinary(max)",    Length=int.MaxValue),                  Nullable            ] public byte[]?           VarbinaryMaxDataType     { get; set; } // varbinary(max)
+		[Column("xmlDataType",              DbType="xml"),                                                     Nullable            ] public string?           XmlDataType              { get; set; } // xml
+		[Column("jsonDataType",             DbType="json",              Length=-1),                            Nullable            ] public string?           JsonDataType             { get; set; } // json
```
MaceWindu: it makes sense to fix schema provider for JSON type to return null length

Tests/Tests.T4/Databases/SqlServer.2025.generated.cs
MaceWindu, line 1: we also should add CLI testing for sql2025

NuGet/SqlServer/LinqToDB.SqlServer.ttinclude
igor-tkachev: 1. I considered using MDS, and even tried it. Unfortunately, MDS requires copying several dozen assemblies needed for its dependencies. On the other hand, the user would only need to add a reference to a single assembly from the output folder to their T4 template.
2. SDS does not support `SqlJson` and `SqlVector` types. So, if user need it he can switch to MDS. By default T4 generates `float[]`.

Tests/Tests.T4/Databases/SqlServer.2025.generated.cs
igor-tkachev: ?

MaceWindu: `Length=-1`

igor-tkachev: done

MaceWindu:
we still need to add cli testing for those types and it probably will require changes to code

PR #5268, opened 2025-12-20
IN SubQuery with Distinct optimization.

igor-tkachev:
The following query

```c#
from p in db.Parent
where p.ParentID.In(db.Child.Select(c => c.ParentID).Distinct())
select p;
```

generates SQL:

```sql
SELECT
	[p].[ParentID],
	[p].[Value1]
FROM
	[Parent] [p]
WHERE
	[p].[ParentID] IN (
		SELECT
			[t1].[In_1]
		FROM
			(
				SELECT DISTINCT
					[c_1].[ParentID] as [In_1]
				FROM
					[Child] [c_1]
			) [t1]
	)
```

### DIFF
--- a/Tests/Linq/Linq/SubQueryTests.cs
+++ b/Tests/Linq/Linq/SubQueryTests.cs
@@ -1473,10 +1473,6 @@ namespace Tests.Linq
 					"SELECT",
 					"SELECT",
 					"DISTINCT"));
-
-//			var sql = query.GetWhere();
-//			Assert.That(sql.Predicates, Has.Count.EqualTo(1));
-//			Assert.That(sql.Predicates[0] is SqlPredicate.InSubQuery q &&, Has.Count.EqualTo(1));
 		}
 	}
 }


Tests/Linq/Linq/InSubqueryTests.cs
```diff
@@ -128,7 +128,7 @@ where UseInQuery(c.ParentID, compareNullsAsValues) && c.ParentID.In(db.Parent.Se
 		}
 
 		[Test]
-		public void InWithSkipTest([DataSources(TestProvName.AllSybase)] string context, [Values] bool preferExists, [Values] bool compareNullsAsValues)
+		public void InWithSkipTest([DataSources(TestProvName.AllSybase, TestProvName.AllAccess, TestProvName.AllSQLite)] string context, [Values] bool preferExists, [Values] bool compareNullsAsValues)
```
sdanyliv, line 131: So, you have not fixed issue, but disabled tests? Don't think that it is acceptable.

igor-tkachev: This is **new** tests for `Skip`. I disabled providers that do not support Skip in subqueries at all.

MaceWindu (approved):
I think it was one of those non-critical baselines regressions we postponed for after release

sdanyliv (approved):

PR #5269, opened 2025-12-20
Force ServerSideOnly=true only when explicitly stated for translated members

MaceWindu:
Pre-6.0.0 (including rc.3), code `DateTimeOffset.Now - TimeSpan.FromDays(1)` translated to single parameter, now it is translated to `SYSUTCDATETIME() - @p`

### DIFF
--- a/Source/LinqToDB/Internal/Linq/Builder/ExpressionTreeOptimizationContext.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/ExpressionTreeOptimizationContext.cs
@@ -395,8 +395,8 @@ namespace LinqToDB.Internal.Linq.Builder
 
 						if (method.DeclaringType != null)
 						{
-							var attribute = method.GetExpressionAttribute(_mappingSchema);
-							if (attribute != null && !attribute.PreferServerSide && attribute.IsPure && !method.IsServerSideOnly(_mappingSchema))
+							if (method.GetExpressionAttribute(_mappingSchema) is { PreferServerSide: false, IsPure: true }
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/Internal/Linq/Builder/ExpressionTreeOptimizationContext.cs
```diff
@@ -397,8 +395,8 @@ bool IsReadOnlyMethod(MethodInfo method)
 
 						if (method.DeclaringType != null)
 						{
-							var attributes = _mappingSchema.GetAttributes<Sql.ExpressionAttribute>(method.DeclaringType, method);
-							if (attributes.Length > 0 && !attributes.Any(a => a.PreferServerSide || a.ServerSideOnly || !a.IsPure))
+							var attribute = method.GetExpressionAttribute(_mappingSchema);
+							if (attribute != null && !attribute.PreferServerSide && attribute.IsPure && !method.IsServerSideOnly(_mappingSchema))
```
viceroypenguin:
```suggeestion
							if (method.GetExpressionAttribute(_mappingSchema) is { PreferServerSide: false, IsPure: true }
								&& !method.IsServerSideOnly(_mappingSchema))
```

MaceWindu:
```suggestion
							if (method.GetExpressionAttribute(_mappingSchema) is { PreferServerSide: false, IsPure: true }
								&& !method.IsServerSideOnly(_mappingSchema))
```

PR #5276, opened 2025-12-23
Corrected Decimal type calculation for Firebird and DB2.

### LINKED ISSUE #5272: InsertOrReplace fails with scale/precision error in merge statement
### Description

The generated Merge statement fails the a scale error when trying to insert/replace some decimal values 
Insert or Update appear to work fine with the same values.

(We are using the Firebird provider so unsure if just with this provider or if it could be others as well?)

Basic problem is that it changes the Precision/Scale in the [NOT] MATCHED parts, at the top it is (4,4) (which is good), further down it is (1,4) which is bad and likely will be causing the error.
```
DECLARE @ID Integer -- Int32
SET     @ID = 4500
DECLARE @MoneyValue Decimal(4, 4)
SET     @MoneyValue = 0.0005

MERGE INTO "LinqDataTypes" "t1"
USING (SELECT CAST(@ID AS Int) AS ID FROM rdb$database) "s" ON
(
	"t1".ID = "s".ID
)
WHEN MATCHED THEN
	UPDATE 
	SET
		"MoneyValue" = CAST(@MoneyValue AS Decimal(1, 4))
WHEN NOT MATCHED THEN
	INSERT
	(
		ID,
		"MoneyValue"
	)
	VALUES
	(
		CAST(@ID AS Int),
		CAST(@MoneyValue AS Decimal(1, 4))
	)
```

### Reproduction Steps

```
	var data = new LinqDataTypesMoney()
	{
		ID = id,
		MoneyValue = 0.0005m
	};
	db.Insert(data); //OK
	db.InsertOrReplace(data); //Fail
```

(will attach the full test code that I added Tests\Linq\DataProvider\FirebirdTests.cs)

### Exception details

```
 DataConnection: Error
Exception: FirebirdSql.Data.FirebirdClient.FbException
Message  : Dynamic SQL Error
SQL error code = -842
Scale must be between zero and precision
At line 9, column 39
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteNonQuery()
   at LinqToDB.Data.DataConnection.ExecuteNonQuery(DbCommand command) in C:\Prog\VS\linq2db-v6\Source\LinqToDB\Data\DataConnection.cs:line 1082
   at LinqToDB.Data.DataConnection.ExecuteNonQuery() in C:\Prog\VS\linq2db-v6\Source\LinqToDB\Data\DataConnection.cs:line 1119

Exception: FirebirdSql.Data.Common.IscException
Message  : Dynamic SQL Error
SQL error code = -842
Scale must be between zero and precision
At line 9, column 39
   at FirebirdSql.Data.Client.Managed.IResponseExtensions.HandleResponseException(IResponse response)
   at FirebirdSql.Data.Client.Managed.Version10.GdsDatabase.ReadResponse()
   at FirebirdSql.Data.Client.Managed.Version11.GdsStatement.Prepare(String commandText)
   at FirebirdSql.Data.FirebirdClient.FbCommand.Prepare(Boolean returnsSet)
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteCommand(CommandBehavior behavior, Boolean returnsSet)
   at FirebirdSql.Data.FirebirdClient.FbCommand.ExecuteNonQuery()
```

### Regression?

_No response_

### Linq To DB version

6.1.0

### .NET version

10.0

### Database

Firebird 5.0

### Database provider

FirebirdSql.Data.FirebirdClient

### Operating system

Windows

### Configuration

_No response_

sdanyliv:
Fixes #5272


Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3842,6 +3842,22 @@ protected virtual void BuildDataTypeFromDataType(DbDataType type, bool forCreate
...
+			var precision = DecimalHelper.GetPrecision(decValue);
+			var scale     = DecimalHelper.GetScale(decValue);
+
+			if (precision == 0)
+				precision = 1;
+
+			if (scale > precision)
+			{
+				precision = scale;
+			}
```
viceroypenguin:
```suggestion
			var scale     = DecimalHelper.GetScale(decValue);
			var precision = Math.Max(
				Math.Max(DecimalHelper.GetPrecision(decValue), 1),
				scale
			);
```

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -3842,6 +3842,22 @@ protected virtual void BuildDataTypeFromDataType(DbDataType type, bool forCreate
 				StringBuilder.Append(CultureInfo.InvariantCulture, $"({type.Precision}{InlineComma}{type.Scale})");
 		}
 
+		protected static DbDataType CorrectDecimalPrecision(DbDataType dataType, decimal decValue)
```
viceroypenguin: Can we make this `public` and do some unit tests on this method specifically?

sdanyliv: For which purpose tests? We already have integration test with DB2 and Firebird. Later this method will be moved to TypeMapping refactoring, which is planned for Verison 7

viceroypenguin: Easier confidence of the reliability of this method itself. 🤷🏼

PR #5288, opened 2025-12-31
Use `T->DataParameter` conversion to generate `Nullable<T>->DataParameter`  conversion

### LINKED ISSUE #5259: Simplify DataParameter mapping for nullable value types
### Description

We should use non-nullable converter if nullable not registered as we do in other places with such types already

### New API

https://github.com/linq2db/linq2db/issues/5254#issuecomment-3662418718

MaceWindu:
Fix #5259

### DIFF
--- a/Source/LinqToDB/Data/DataParameter.cs
+++ b/Source/LinqToDB/Data/DataParameter.cs
@@ -179,7 +179,7 @@ namespace LinqToDB.Data
 			return this;
 		}
 
-		internal DataParameter WithType(DbDataType type)
+		internal DataParameter SetType(DbDataType type)
 		{
 			DbDataType = type;
 			return this;
--- a/Source/LinqToDB/Mapping/MappingSchema.cs
+++ b/Source/LinqToDB/Mapping/MappingSchema.cs
@@ -883,7 +883,7 @@ namespace LinqToDB.Mapping
 							static (context, e) => e == context.oldParam ? Expression.Coalesce(context.newParam, context.defaultValue) : e),
 						Methods.LinqToDB.DataParameter.ClearValue,
 						Expression.Equal(p, Expression.Constant(null, p.Type))),
-					Methods.LinqToDB.DataParameter.WithType,
+					Methods.LinqToDB.DataParameter.SetType,
 					Expression.Constant(fromNullable)),
 				p);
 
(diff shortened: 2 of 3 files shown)


Source/LinqToDB/Expressions/MemberHelper.cs
```diff
@@ -199,6 +199,12 @@ public static MethodInfo MethodOf(Expression<Action> func)
 			return mi is PropertyInfo info ? info.GetGetMethod()! : (MethodInfo)mi;
 		}
 
+		public static MethodInfo MethodOf<T>(Expression<Action<T>> func)
+		{
+			var mi = GetMemberInfo(func);
+			return mi is PropertyInfo info ? info.GetGetMethod()! : (MethodInfo)mi;
```
viceroypenguin:
```suggestion
			return GetMemberInfo(func) switch
			{
				PropertyInfo info => info.GetGetMethod()!,
				MethodInfo info => info,
				_ => throw new LinqToDBException("??"),
			};
```

Source/LinqToDB/Mapping/MappingSchema.cs
```diff
@@ -966,6 +967,53 @@ static bool Simplify(ref DbDataType type)
...
+								static (context, e) => e == context.oldParam ? Expression.Coalesce(context.newParam, context.defaultValue) : e),
+							Methods.LinqToDB.DataParameter.ClearValue,
+							Expression.Equal(p, Expression.Constant(null, p.Type))),
+						Methods.LinqToDB.DataParameter.WithType,
+						Expression.Constant(from)),
+					p);
+
+				if (conversion.Lambda != conversion.CheckNullLambda)
+				{
+					throw new InvalidOperationException();
```
viceroypenguin: why IOE instead of linq2db exception?

MaceWindu: shouldn't happen, just-in-case guard

viceroypenguin (approved):

Source/LinqToDB/Data/DataParameter.cs
```diff
@@ -171,6 +171,20 @@ public DbDataType DbDataType
...
 
+		internal DataParameter ClearValue(bool clearValue)
+		{
+			if (clearValue)
+				Value = null;
+
+			return this;
+		}
+
+		internal DataParameter WithType(DbDataType type)
```
sdanyliv: Can we make it immutable? Otherwise change name to, for example, `SetType`

MaceWindu: There is no reason to make it immutable as we don't share instances and don't expect users to return shared instances from custom converter. I will rename method

PR #5258, opened 2025-12-16
Custom operators mapping improvements

MaceWindu:
- fix/add support for unary/binary operators mapping using `Sql.ExpressionAttribute`, `MemberExpressionAttribute` or `IMemberTranslator`
- fix v6 regressions where mapper loose custom operator calls: fix #5254, fix #5295

PR #5295, opened 2026-01-06
Implicit conversion bug

igor-tkachev:
Add regression test for incorrect `NULL` handling when mapping via implicit conversion operators. 

When a database column value is `NULL`, the mapper currently assigns `null` to the destination wrapper object itself, instead of creating the wrapper instance and passing `null` into it (i.e., `new ImplicitValue<T> { Value = null })`. 

This breaks wrapper semantics and produces different results compared to the same mapping performed in memory.


MaceWindu:
I will merge tests to https://github.com/linq2db/linq2db/pull/5258 as they track very similar if not the same issue

MaceWindu:
Closing, fix provided by #5258

PR #5301, opened 2026-01-09
Fix parameters detection

MaceWindu:


sdanyliv:
It would be better to have regression test.

PR #5314, opened 2026-01-28
Fix incorrect IgnoreFilters chaining.

sdanyliv:
Fix incorrect IgnoreFilters chaining.

### DIFF
--- a/Tests/Linq/UserTests/Issue1284Tests.cs
+++ b/Tests/Linq/UserTests/Issue1284Tests.cs
@@ -64,6 +64,7 @@ namespace Tests.UserTests
 			}
 		}
 
+		[ActiveIssue(Configuration = TestProvName.AllClickHouse, Details = "CH 26.1.1 regression")]
 		[Test]
 		public void TestCteInvalidMappingUnion([CteContextSource] string context)
 		{


Source/LinqToDB/Internal/Linq/Builder/TranslationModifier.cs
```diff
@@ -65,7 +65,15 @@ public TranslationModifier WithIgnoreQueryFilters(Type[]? ignoreQueryFilters)
...
 			}
 
-			return new TranslationModifier(InlineParameters, ignoreQueryFilters);
+			var newFilters = IgnoreQueryFilters == null
+				? ignoreQueryFilters
+				: ignoreQueryFilters == null
+					? null
+					: IgnoreQueryFilters.Length == 0 || ignoreQueryFilters.Length == 0
+						? []
+						: IgnoreQueryFilters.Union(ignoreQueryFilters).ToArray();
```
viceroypenguin: This is the translation of the ternaries. There may be a better set of patterns...
```suggestion
			var newFilters = (IgnoreQueryFilters, ignoreQueryFilters) switch
			{
				(null, null) => null,
				([], _) or (_, []) => [],
				(null, _) => ignoreQueryFilters,
				_ => IgnoreQueryFilters,
			};
```

PR #5316, opened 2026-01-29
Fix inheritance detection.

### LINKED ISSUE #4668: Inheritance Discriminator is not defined
I have a class with its own table. And also a class inherited from it, which is stored entirely in a separate table.

In the EF in EntityTypeBuilder it is done as follows:

`builder.HasBaseType((Type)null)`

In .net 6 linq2db.EntityFrameworkCore worked normally. But in .net 8 an error appeared "Inheritance Discriminator is not defined.."

The reason is that EFCoreMetadataReader contains the condition
`if (GetBaseTypeRecursive(e) == et && e.GetDiscriminatorValue() != null)`

In .net 8, `e.GetDiscriminatorValue()` always returns the class name in such cases, regardless of whether it needs to be used or not.

So, apparently, it is necessary to do some additional checks in order not to add `InheritanceMappingAttribute`.

sdanyliv:
Fixes #4668

### DIFF
--- a/Tests/EntityFrameworkCore/Tests.EntityFrameworkCore.EF9.csproj
+++ b/Tests/EntityFrameworkCore/Tests.EntityFrameworkCore.EF9.csproj
@@ -3,7 +3,7 @@
 	<Import Project="Tests.EntityFrameworkCore.props" />
 
 	<PropertyGroup>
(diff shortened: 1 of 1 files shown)


MaceWindu:
build fails

MaceWindu (approved):

PR #5331, opened 2026-02-06
Fix unwanted join condition nullability weakening

MaceWindu:


MaceWindu:
Conflicts with #4875

MaceWindu:
Replaced with #5341

PR #5332, opened 2026-02-08
Add missing `DateTime.UtcNow` translation mapping, fix found translation issues

### LINKED ISSUE #5325: PostgreSQL: Incorrect SQL generated when transform DateTime.UtcNow
### Description

linq2db currently transform expression like `DateTime.UtcNow` into inline value with incorrect value.

### Reproduction Steps

Consider code something like:
```
    var q = db.Entity1
      .Where(x => x.CreatedAt == DateTime.UtcNow);
    var sql = q.ToSqlQuery().Sql;

// sql now holds:
// WHERE
//	x.created_at = '2026-02-03 21:50:09.734'::timestamp
```

Query above generated at `2026-02-04T00:50:09.734` at MSK local time (+03:00 offset), e.g. linq2db pick's UTC datetime value and pretend it like local time, but that's incorrect. Correct and expected value is `'2026-02-03T21:50:09.734Z'::timestamptz`.

Also please note, what postgresql always store values in UTC, and `timestamptz` doesn't stores actual offset, it is matches to DateTime into 1-to-1, whenever DateTime is in UTC. Using DateTimeOffset is not an option, as it intended to store offset, but that's has no any sense in this scope. Also conversion/coercion from timestamp to timestamptz assumes what timestamp in local time, but this conversion will not work in general case, because client's local timezone, and server's timezone might be different, so this conversions should not occur in first place.

PS: Note, what passing DateTime as parameter have no issue.

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.0.1

### .NET version

10

### Database

PostgreSQL 18

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

MaceWindu:
Fix #5325

### DIFF
--- a/.runsettings
+++ b/.runsettings
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="UTF-8"?>
 <RunSettings>
 	<RunConfiguration>
-		<MaxCpuCount>0</MaxCpuCount>
+		<MaxCpuCount>1</MaxCpuCount>
 	</RunConfiguration>
 	<NUnit>
 		<AssemblySelectLimit>1000000</AssemblySelectLimit>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -137,7 +137,7 @@
 
 	<ItemGroup Label="Testing">
 		<PackageVersion Include="Shouldly"                                              Version="4.3.0"                        />
-		<PackageVersion Include="Microsoft.NET.Test.Sdk"                                Version="18.0.1"                       />
+		<PackageVersion Include="Microsoft.NET.Test.Sdk"                                Version="18.3.0"                       />
 		<PackageVersion Include="NUnit"                                                 Version="4.4.0"                        />
 		<PackageVersion Include="NUnit.Analyzers"                                       Version="4.11.2"                       />
 		<PackageVersion Include="NUnit3TestAdapter"                                     Version="5.2.0"                        />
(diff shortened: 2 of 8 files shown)


MaceWindu:
Baselines in general show improvements, there are several cases where we now generate upcasts like from short to int, but we never had code to omit them and they were missing only because of  incorrect Convert translation

Source/LinqToDB/Internal/DataProvider/PostgreSQL/Translation/PostgreSQLMemberTranslator.cs
```diff
@@ -240,7 +240,10 @@ ISqlExpression ToInterval(ISqlExpression numberExpression, string intervalKind)
 			protected override ISqlExpression? TranslateSqlCurrentTimestampUtc(ITranslationContext translationContext, DbDataType dbDataType, TranslationFlags translationFlags)
 			{
 				var factory = translationContext.ExpressionFactory;
-				return factory.Function(dbDataType, "timezone", factory.Value("UTC"), factory.Function(dbDataType, "now"));
+
+				// timezone('UTC', now()) returns timestamp without tz
+				// https://www.postgresql.org/docs/current/functions-datetime.html
+				return factory.Function(dbDataType.WithDataType(DataType.DateTime2), "timezone", factory.Value("UTC"), factory.Function(dbDataType.WithDataType(DataType.DateTimeOffset), "now"));
```
viceroypenguin, line 246:
```suggestion
				return factory.Function(
					dbDataType.WithDataType(DataType.DateTime2),
					"timezone",
					factory.Value("UTC"),
					factory.Function(
						dbDataType.WithDataType(DataType.DateTimeOffset),
						"now"
					)
				);
```

PR #5352, opened 2026-02-17
Fix ValueConverter handling for updating record.

### LINKED ISSUE #5351: 6.x regression: ValueConverter on bool
### Description

When a bool ValueConverter maps to `null`, linq2db 6 incorrectly tries to compare with a strict equality, which never works.

### Reproduction Steps

```csharp
// Here's a value converter that maps DB strings to bool:
public class YonConverterAttribute : ValueConverterAttribute
{
  public YonConverterAttribute(string yes, string? no)
  {
    ValueConverter = new ValueConverter<bool, string?>(
      b => b ? yes : no,
      s => s == yes,
      handleNulls: no == null
    );
  }
}

// Here's a DB model that contains a legacy char(1) column where 'X' means true and NULL means false.
public class Entity
{
  [YonConverter(yes: "X", no: null)]
  public bool Test { get; set; }
}

// Here's a query that is wrong with linq2db 6:
var q = db.GetTable<Entity>().Where(x => !x.Test);
```

The generated query is this:
```sql
-- :false = null
select * from Entity where Test = :false
```
Because `:false` is null, this will never return any row.
The correct translation is `where Test is null`.

In other statements such as UPDATE, I observed a `WHERE 1 = 0` translation, presumably because linq2db noticed that the predicate would always be false.

### Exception details

No exception, but SQL translation is incorrect

### Regression?

Worked in 5.x, broken since 6.0

### Linq To DB version

6.1

### .NET version

10

### Database

Oracle 19

### Database provider

ODP.NET 23.26.100

### Operating system

Windows

### Configuration

_No response_

sdanyliv:
Fixes #5351


Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
viceroypenguin, line 4055:
```suggestion
if (nodeType is ExpressionType.Equal or ExpressionType.NotEqual)
{
  // ..
}
```

viceroypenguin, line 4162:
```suggestion
var lOriginal = leftPlaceholder.Sql;
var rOriginal = rightPlaceholder.Sql;

var l = QueryHelper.UnwrapExpression(lOriginal, checkNullability: true);
var r = QueryHelper.UnwrapExpression(rOriginal, checkNullability: true);
```

PR #5356, opened 2026-02-19
Fix optimization of unused joins with nested joins

MaceWindu:
Issue existed for some time, but became more severe after recent fix to joins #5331

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
@@ -2825,7 +2825,7 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 					{
 						if (_providerFlags.IsCrossJoinSupported
 							&& (table.Joins.Count > (_providerFlags.IsCrossJoinSyntaxRequired ? 0 : 1)
-								|| !QueryHelper.IsDependsOnSource(selectQuery.Where, join.Table)))
+								|| !QueryHelper.IsDependsOnSource(selectQuery.Where, join.Table.Source)))
 						{
 							join.JoinType = JoinType.Cross;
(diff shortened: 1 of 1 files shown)


Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
viceroypenguin: I think this is what the copilot comment is referencing; I'd agree that if we're going to modify `j` here, we should be more explicit that modifying `j` is intentional. Especially since we set `sj = 0`, don't modify it again, and use it as an addition to `j` on line 1007.

MaceWindu: that's some "strange" code... I will try to find out what we really wanted to do here

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
```diff
@@ -2373,7 +2373,7 @@ bool MoveJoinConditionsToWhere(SqlStatement root, SqlTableSource left, SqlJoined
 						sources ??= QueryHelper.EnumerateAccessibleSources(join.Table).ToList();
 						move = !QueryHelper.IsDependsOnSources(predicate, sources);
 
-						if (!move && !QueryHelper.IsDependsOnSources(predicate, [left]))
+						if (!move && !QueryHelper.IsDependsOnSource(predicate, left))
```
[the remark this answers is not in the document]
MaceWindu: @copilot, so, what is your problem?

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
MaceWindu: We need a polyfill for InsertRange to avoid such magic code..

PR #5358, opened 2026-02-19
Corrected subquery nullability.

### LINKED ISSUE #5328: Unexpected LeftJoin Result After Upgrading (Maybe DefaultIfEmpty?)
### Description

We finally made the move from 5.4 to 6.1 and we're having issues with some of our queries that were using left join. We aren't receiving any errors, but values are missing in random places where left joins were used.

This is a example of a section of a large query that worked correctly in the previous version:
```
var result = 
    from Table1 in db.Table1
    from supplementTotal in (from Table2 in db.Table2.LeftJoin(t2 => t2.LinkedId == Table1.Id)
                                group Table2 by Table2.CntrctNumber into Table2Group
                                select new
                                {
                                    supplementAmount = Table2Group.Sum(can => can.Amount)
                                }).DefaultIfEmpty()
    from transactionsTotals in (from Table3 in db.Table3.LeftJoin(t3 => t3.LinkedId == Table1.Id)
                                group Table3 by Table3.CntrctNumber into Table3Group
                                select new
                                {
                                    amount1 = Table3Group.Where(dt => dt.Type == "Type1").Sum(d => d.Amount),
                                    amount2 = Table3Group.Where(dt => dt.Type == "Type2").Sum(d => d.Amount)
                                }).DefaultIfEmpty()
    select new TestEntity
    {
        Table1 = Table1,
        AddedAmounts = transactionsTotals.amount1 - supplementTotal.supplementAmount,
        Amount1 = transactionsTotals.amount1,
        Amount2 = transactionsTotals.amount2,
    };
```

The issue is `AddedAmounts` will return as null when there are no records for the supplementTotal left join. Since the value `null` is being used in the equation `AddedAmounts = transactionsTotals.amount1 - supplementTotal.supplementAmount,` it causes the entire thing to null out.

Things I've tried:
- Using the null coalescing operator: (`Table2Group.Sum(can => (decimal?)can.Amount) ?? 0`)
- messing with various Sql.X functions: `Sql.ToNullable(Table2Group.Sum(can => can.Amount)) ?? 0m`
- Providing the DefaultIfEmpty with a value `.DefaultIfEmpty(new { supplementAmount = 0m })`
- Ternary operator in the select `AddedAmounts = transactionsTotals.amount1 - (supplementTotal != null ? supplementTotal.supplementAmount : 0)` 

But nothing I do seems to affect the generated query which is always some form of (or sometimes with a Cross Apply instead):

```
    SELECT
	set_1."ROWID",
	transactionsTotals_1."Amount1" - (
		SELECT
			SUM(supplementTotal.Amount)
		FROM
			EX2 supplementTotal
		WHERE
			supplementTotal.LinkedId = set_1.Id AND
		GROUP BY
			supplementTotal.LinkedId
	),
	transactionsTotals_1."Amount1",
	transactionsTotals_1."Amount2"
FROM
	EX1 set_1
		OUTER APPLY (
			SELECT
				SUM(CASE
					WHEN transactionsTotals.Type = 'Type1' THEN transactionsTotals.Amount
					ELSE NULL
				END) as "Amount1",
				SUM(CASE
					WHEN transactionsTotals.Type = 'Type2' THEN transactionsTotals.Amount
					ELSE NULL
				END) as "Amount2"
			FROM
				EX3 transactionsTotals
			WHERE
				transactionsTotals.LinkedId = set_1.Id
			GROUP BY
				transactionsTotals.LinkedId
		) transactionsTotals_1
WHERE
	set_1.Id = :TestKey
```

For clarity you can see the result here:
Image
So Amount1 has a value, but the second selected column `(Amount1 - NULL)` is null since oracle treats it as an unknown. I need something like a way to wrap the entire supplementTotal in a NVL(..., 0)?

I tried looking flipping through the [v6 wiki ](https://github.com/linq2db/linq2db/wiki/Linq-To-DB-6) page but didn't see anything specific around LeftJoins and DefaultIfEmpty that seemed to match our case.

### Reproduction Steps

See example code above.

### Exception details

_No response_

### Regression?

5.4

### Linq To DB version

6.1

### .NET version

10

### Database

_No response_

### Database provider

Oracle

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5328

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/SelectQuery.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/SelectQuery.cs
@@ -339,7 +339,7 @@ namespace LinqToDB.Internal.SqlQuery
 
 		public SelectQuery CloneQuery()
 		{
-			return this.Clone(e => ReferenceEquals(e, this));
+			return this.Clone(e => e is not SqlCteTable);
 		}
 	}
 }
--- a/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
+++ b/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
@@ -422,7 +422,7 @@ namespace LinqToDB.Internal.DataProvider.ClickHouse
 		protected override void BuildCteBody(SelectQuery selectQuery)
 		{
 			var sqlBuilder = (ClickHouseSqlBuilder)CreateSqlBuilder();
-			sqlBuilder.BuildSql(0, new SqlSelectStatement(selectQuery), StringBuilder, OptimizationContext, Indent, false, null);
+			sqlBuilder.BuildSql(0, new SqlSelectStatement(selectQuery), StringBuilder, OptimizationContext, Indent, ColumnAliasMode.None, null);
 		}
 
 		protected override void BuildInsertQuery(SqlStatement statement, SqlInsertClause insertClause, bool addAlias)
--- a/Source/LinqToDB/Internal/DataProvider/DB2/DB2SqlBuilderBase.cs
+++ b/Source/LinqToDB/Internal/DataProvider/DB2/DB2SqlBuilderBase.cs
@@ -82,7 +82,15 @@ namespace LinqToDB.Internal.DataProvider.DB2
 			StringBuilder.AppendLine();
 		}
 
-		protected override void BuildSql(int commandNumber, SqlStatement statement, StringBuilder sb, OptimizationContext optimizationContext, int indent, bool skipAlias, NullabilityContext? nullabilityContext)
+		protected override void BuildSql(
+			int commandNumber,
+			SqlStatement statement,
+			StringBuilder sb,
+			OptimizationContext optimizationContext,
+			int indent,
+			ColumnAliasMode aliasMode,
+			NullabilityContext? nullabilityContext
+		)
 		{
 			var nullability = NullabilityContext.GetContext(statement.SelectQuery);
 
@@ -90,7 +98,7 @@ namespace LinqToDB.Internal.DataProvider.DB2
 			StringBuilder       = sb;
 			OptimizationContext = optimizationContext;
 			Indent              = indent;
-			SkipAlias           = skipAlias;
+			AliasMode           = aliasMode;
 
 			if (_identityField != null)
 			{
@@ -105,7 +113,7 @@ namespace LinqToDB.Internal.DataProvider.DB2
 				AppendIndent().Append('\t').AppendLine(OpenParens);
 			}
 
-			base.BuildSql(commandNumber, statement, sb, optimizationContext, indent, skipAlias, nullabilityContext);
+			base.BuildSql(commandNumber, statement, sb, optimizationContext, indent, aliasMode, nullabilityContext);
 
 			if (_identityField != null)
 				sb.AppendLine("\t)");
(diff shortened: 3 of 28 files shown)


MaceWindu (approved):

Source/LinqToDB/Internal/SqlQuery/SelectQuery.cs
```diff
@@ -217,19 +217,19 @@ internal set
 
 		public override QueryElementType ElementType => QueryElementType.SqlQuery;
 
-		public override bool CanBeNullable(NullabilityContext nullability)
+	public override bool CanBeNullable(NullabilityContext nullability)
+	{
+		if (!this.IsLimited() && !this.HasSetOperators && !this.HasGroupBy() && !this.HasHaving() && this.IsSingleColumn()
```
jods4, line 222: @MaceWindu @sdanyliv This is wrong.
Having a single column *is not a requirement* for non-nullable result.

Queries that use aggregation functions without GROUP BY always return a row, for example:
```sql
select 1, count(*), sum(qty) from Products where 1 = 0
-- 1, 0, null
```

sdanyliv: For subquery expression it should be always single column, just additional test.

jods4: I am also unsure why Set operators or HAVING influence the result?
`having count(*) > 0` would make the row disappear but doesn't change the nullability of result.
for instance, `having count(*) < 10` produces the exact same result.

jods4: @sdanyliv ah this code runs only for subqueries?

ok but then the "single column" requirement is surely guaranteed somewhere else, so this is superfluous code that still isn't meaningful?

and does it matter for places where a subquery could return several columns?
for example in an EXISTS statement (ok the selected value are not relevant but the query could be reused from somewhere else and have multiple columns); or when comparing a subquery to a tuple (ROW)?

sdanyliv: Nullability is checked when we test `COALESCE(SELECT x FROM .., y)` or `(SELECT x FROM ..) IS NULL`, ect. Nullability for the query that has several columns has no sense.

jods4: I see. 👍
For HAVING I assume the logic is that if it could produce no row in result you consider it nullable?
Note that `HasSetOperator` would be conservative then, as `UNION [ALL]` cannot remove rows from result.

jods4: > Nullability for the query that has several columns has no sense.

Just for curiosity: actually `is null` and `is not null` are defined for row expressions, but I don't know any RDBMS that supports applying them to subqueries.

A row expression can be compared to null:
```sql
select 1 from Product where (category, description) is null
-- equivalent to category is null and description is null
-- note that `is not null` isn't the negation but rather: category is not null and description is not null
```

And subqueries can return rows, e.g. in MariaDB, Postgres, Oracle...
```sql
select 1 from Product where (name, category) = (select name, category from Table)
-- works, as well as other comparisons like >
```

Interestingly of all RDBMS I've tried, none seem to support `IS NULL` on a subquery row.
Some (e.g. MariaDB) return an error _"row does not support isnull operator"_.
Postgres mysteriously doesn't return an error but doesn't produce a result either :D

Issue #5371, opened 2026-02-24
DataParameter name not used in executed query


darko1979:
### Description

Upgrading from linq2db 4.5.1 to 6.1.0 following query is not working:
```cs
var result = await db.GetTable<Item>()
    .Where(x => Sql.Expr<bool>("LOWER(\"value\") = LOWER(@p1)", new DataParameter("@p1", "ONE", DataType.NVarChar)))
    .ToArrayAsync();
```

### Reproduction Steps

This is the part of an example:
```cs
public partial class Item
{
    [Column("id", DataType = LinqToDB.DataType.Int32, Precision = 32, Scale = 0), PrimaryKey, NotNull] public int Id { get; set; } // integer
    [Column("value", DataType = LinqToDB.DataType.Text), NotNull] public string Value { get; set; } // text
}

var result = await db.GetTable<Item>()
    .Where(x => Sql.Expr<bool>("LOWER(\"value\") = LOWER(@p1)", new DataParameter("@p1", "ONE", DataType.NVarChar)))
    .ToArrayAsync();
```
using linq2db 4.5.1 it produced following query that works:
```sql
DECLARE @p1 Varchar(3) -- String
SET     @p1 = 'ONE'

SELECT
        x.id,
        x.value
FROM
        "public".item x
WHERE
        LOWER("value") = LOWER(@p1)
```
using linq2db 6.1.0 I get following query that fails:
```sql
DECLARE @p Varchar(3) -- String
SET     @p = 'ONE'

SELECT
        x.id,
        x.value
FROM
        public.item x
WHERE
        LOWER("value") = LOWER(@p1)
, DataConnection
Error: Error
Exception: Npgsql.PostgresException
Message  : 42703: column "p1" does not exist
```

### Linq To DB version

6.1.0

### .NET version

.NET 10.0

### Database

PostgreSQL 18

### Operating system

Windows 10

sdanyliv:
@darko1979, note that this expression is wrong and it works accidentally. We do not parse content of string
```cs
Sql.Expr<bool>("LOWER(\"value\") = LOWER(@p1)", new DataParameter("@p1", "ONE", DataType.NVarChar))
```
This one with `{0}` is correct, like `string.Format`:
```cs
Sql.Expr<bool>("LOWER(\"value\") = LOWER({0})", new DataParameter("@p1", "ONE", DataType.NVarChar))
```

PR #5372, opened 2026-02-24
test for issue #5371

jogibear9988:


jogibear9988:
I've tried to fix this. It seems to work, but I don't really like it

sdanyliv:
Supreceded by https://github.com/linq2db/linq2db/pull/5375

sdanyliv:
@jogibear9988, missed your tests. Anyway, fix was wrong.

PR #5376, opened 2026-02-26
docs: add package-local linq2db AI skill and Expert knowledge pack

igor-tkachev:
## Summary

Adds a package-local AI skill for linq2db and a versioned linq2db Expert knowledge pack derived from the same package-local sources.

The goal is to let coding agents use the installed NuGet package as the version-matched source of truth for this linq2db version instead of relying on online docs, older examples, or pretrained memory.

## Included changes

- add `Source/Skills/linq2db/SKILL.md` as the canonical AI agent entry point
- ship task-focused skill docs under `skills/linq2db/docs/` in the NuGet package
- keep `Source/LinqToDB/README.md` as the NuGet/package-facing discovery entry
- generate `skills/linq2db/docs/api.md` from package XML documentation for API discovery
- add custom XML-doc `<ai-tags />` / `<ai-tags-defaults />` metadata for generated API indexing without putting machine tags in human-facing `<remarks>` prose
- document provider setup, provider capabilities, mapping/configuration, CRUD/MERGE, temp tables, hints, custom SQL, interceptors, and translatable methods for agent use
- add and maintain `Source/Knowledge/linq2db-expert/` as a versioned Custom GPT / Expert upload pack derived from the skill docs and XML-doc
- add maintainer tooling and validation for regenerating the skill/API/knowledge artifacts, including Expert-pack source hash manifests and missing-doc bundle checks

## Source of truth

Package behavior and LinqToDB API guidance should be changed first in:

- `Source/Skills/linq2db/SKILL.md`
- `Source/Skills/linq2db/docs/**/*.md`
- XML documentation in public API source files
- generated `Source/Skills/linq2db/docs/api.md`

`Source/Knowledge/linq2db-expert/` is a derived/versioned artifact for hosted assistants and must not become a second source of truth.

## Reviewer notes

- The generated markdown artifacts are intentionally committed: `docs/api.md` and `Source/Knowledge/linq2db-expert/*`.
- Review focus should be on source docs, XML-doc comments, generators, packaging rules, and source-of-truth boundaries.
- Generated files should be checked for deterministic regeneration and upload/package layout regressions rather than reviewed as independent hand-authored guidance.
- `<ai-tags />` metadata is authored as custom XML-doc elements and rendered into generated docs as `AI metadata`; it should not appear as `AI-Tags:` text inside `<remarks>`.

## Validation performed locally

- `dotnet build Source\LinqToDB\LinqToDB.csproj -f net10.0 -c Release --no-restore /p:GenerateDocumentationFile=true /p:GenerateAiDocsOnBuild=true`
- `.agents\Build-LinqToDBExpert.ps1 -NoBuild -XmlDocPath .build\bin\LinqToDB\Release\net10.0\linq2db.xml -OutputRoot Source\Knowledge\linq2db-expert`
- `dotnet build Source\LinqToDB\LinqToDB.csproj --no-restore`
- `dotnet pack Source\LinqToDB\LinqToDB.csproj -c Release --no-restore`
- `git diff --check`
- package layout check for `skills/linq2db/SKILL.md`, skill docs, and bundled `linq2db.xml`

## Follow-up commit

Added an internal, enum-backed alternative to the XML-doc `<ai-tags />` custom tag:
`AiTagsAttribute`/`AiTagsDefaultsAttribute` (`Source/LinqToDB/Internal/Metadata/`), motivated by
@MaceWindu's review question about whether XML-doc-adjacent metadata is the right container.
Rationale, tradeoff comparison against a typed public attribute and a sidecar file, and current
status are in `.agents/ai-tags-attribute-design.md`.

`Build/GenerateApiDocs.ps1` now reads `AiTags`/`AiTagsDefaults` via `CustomAttributeData` off the
compiled assembly, falling back to the legacy XML-doc `<ai-tags />` path for members not yet
migrated - both mechanisms coexist during the transition. 7 representative members are converted
as a proof of concept (covering every tag pattern found across the ~747 existing instances);
mechanical conversion of the rest is not done yet.

## Follow-up commit (update)

The AI-Tags XML-doc → attribute migration described above is now complete: all ~777
`<ai-tags>`/`<ai-tags-defaults>` XML-doc declarations were mechanically converted to
`AiTagsAttribute`/`AiTagsDefaultsAttribute` via `.agents/scripts/Convert-AiTagsToAttributes.ps1`.
`Build/GenerateApiDocs.ps1` no longer has a legacy XML-doc fallback (removed, guarded by
`AssertNoLegacyAiTagXml`); `docs/api.md` is generated exclusively from compiled-assembly attribute
metadata.

Additional skill docs added: `docs/concurrency.md`, `docs/associations.md`, `docs/raw-sql.md`,
`docs/null-semantics.md`, `docs/parameters.md`.

`docs/custom-sql.md` renamed to `docs/extensions.md` and broadened to cover `IMemberTranslator` in
addition to `[Sql.Expression]`/`[Sql.Function]`/`[ExpressionMethod]`. `docs/configuration.md` was
audited against source: fixed an inaccurate `IRetryPolicy` method count and a fabricated
`TraceLevel` behavior table, added external-config-file guidance (JSON/`ILinqToDBSettings`, legacy
`app.config`/`linq2db.Compat`), and trimmed content duplicated with `interceptors.md`/
`extensions.md` out to those dedicated docs.


Source/LinqToDB/LinqExtensions/LinqExtensions.cs
```diff
@@ -22,8 +22,68 @@
...
+	///
+	/// <para>
+	/// <b>Provider contract:</b>
+	/// availability, supported syntax, and exact SQL semantics are defined by the configured provider.
+	/// When an API is not supported, behavior is provider-defined.
+	/// </para>
+	///
+	/// <para>
+	/// AI-Tags: Groups=QueryDirectives,NavigationLoading,DML,Merge,Helpers; Pipeline=ExpressionTree,SqlAST,SqlText; Provider=ProviderDefined;
+	/// </para>
```
viceroypenguin: These should not be publicly visible to regular consumers. Can we put them in another tag (i.e. not `<remarks />`), so they do not add visual noise for non-AI consumers?

igor-tkachev: You’re absolutely right. Ideally this shouldn’t add any visual noise for regular users. I completely understand the concern and, philosophically, I’m with you here. If we could make it invisible without side effects, I would happily do that.

The issue is that moving it to a custom XML tag doesn’t actually guarantee it will be hidden. Different tools render unknown tags inconsistently, and some AI agents may ignore non-standard elements entirely, at which point the whole idea loses its purpose.

Using a simple `AI-Tags:` line keeps it predictable, tooling-agnostic, easy to parse, and relatively unobtrusive. It’s a compact, standardized annotation that most users will simply ignore.

If proper mechanisms for AI-specific metadata become standardized in the future, we can migrate to them (very likely with the help of AI tooling itself). For now, this is the most stable way to keep the library LLM-friendly without introducing fragility.

viceroypenguin: > You’re absolutely right. Ideally this shouldn’t add any visual noise for regular users. I completely understand the concern and, philosophically, I’m with you here. If we could make it invisible without side effects, I would happily do that.

Can we please not use AI for our comments discussing this?

> Using a simple `AI-Tags:` line keeps it predictable, tooling-agnostic, easy to parse, and relatively unobtrusive. It’s a compact, standardized annotation that most users will simply ignore.

I disagree. I have not seen this used in any other OSS library, and as such, it would stand out significantly to me. If it is not necessary for other libraries, then I think we should avoid using it here as well. If there are examples of consistent patterns being used to encourage LLM context information, I would support that, but I don't buy that `AI-Tags:` has such a history or pattern.

NOTE: I agree that the changes made in the xmldocs here improve the quality of linq2db significantly, other than these `AI-Tags:` lines.

viceroypenguin (changes requested):
Reiterating: the `AI-Tags:` should be removed.

There is a new standard for providing information to LLMs for various libraries: "skills".

See:
* https://devblogs.microsoft.com/dotnet/extend-your-coding-agent-with-dotnet-skills/
* https://github.com/github/awesome-copilot

It would be better to develop a `SKILL.md`, similar to https://github.com/github/awesome-copilot/blob/main/skills/postgresql-optimization/SKILL.md.

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
igor-tkachev: I do not see a real problem here.

I do not even remember the last time I carefully read xml-docs for every method I use. Do you actually read through the full documentation of each method, so such lines will be distracting, or are you just being picky about it?

The `AI-Tags:` line is intentionally short and easy to ignore. It is just one extra line in `<remarks>`, it does not make the docs harder to understand. I would say some developers might actually appreciate this. It shows we care about real-world usage and about helping users avoid incorrect AI-generated suggestions.

As AI tools are becoming more common, without some hints they tend to hallucinate or suggest patterns that don’t apply to linq2db (`SaveChanges`, EF-style workflows, etc.).

Also, the argument that "other OSS libraries don’t do this" is not very convincing to me. There are quite a few things in linq2db that other libraries never did either.

AI-oriented documentation patterns are already starting to appear (for example `llm.txt` and `llm-full.txt`). We will add those as well. But in practice they have much less impact than xml-doc, because AI tools most often rely on what is directly attached to the API.

And as I said earlier, if a standard way to attach AI metadata appears, these tags can be removed immediately. That is not a problem at all. For now this is simply the most practical way to reduce incorrect AI-generated suggestions for linq2db users.

viceroypenguin: >  Do you actually read through the full documentation of each method, so such lines will be distracting, or are you just being picky about it?

Yes. I do read the documentation of methods when I'm using them. Not just that, I do so in-place, where it is hovered over the rest of my code. Having switched to the branch to review some of the docs, even just the large documentations like below will be incredibly distracting to me. That said, given that it is on a type that is rarely (if ever) used directly (used instead via extension methods), both the long documentation and the tags would be nominally acceptable.

image

This is closer to acceptable, both descriptive and concise. However, the extra line _does_ stand out, particularly because it sits between the relevant remarks documentation and the returns documentation.

image

If we wanted a better way to segregate methods to improve correctness, it would be better to introduce `public static class InsertExtensions;`, `public static class MergeExtensions;`. Functionally, this would be better for AI, general users, and us as devs because the methods would be organized rather than in a single large class split across multiple partial files. 

I would imagine part of the problem for AI in analyzing linq2db for extension methods is how many are in a single class with little additional context for them. 

> AI-oriented documentation patterns are already starting to appear (for example llm.txt and llm-full.txt). We will add those as well. 

My point is we should do as much as we can to focus on those, including the new Skills introduced by the dotnet team for this purpose, rather than introducing documentation that will be irritating to at least some users (myself included).

igor-tkachev:
@viceroypenguin, also, take a look at the Copilot review in this PR.

It pointed out quite a few inaccuracies, and interestingly almost all of them were in the **AI-Tags**, not in the regular user-facing xml-doc. And most of those comments are actually correct.

So it looks like Copilot understood those tags immediately and used them very effectively. In some cases even more reliably than the surrounding descriptive text.

Which is pretty much exactly what those annotations are meant for.

viceroypenguin:
@igor-tkachev I would argue the point that the xmldocs were enough to identify the inaccuracies in the `AI-Tags:`, which means that the xmldocs are enough to rely on, and the tags are not necessary.

I don't know what affects LLM context, so I don't know what will help, but would you be willing to attempt to break out `LinqExtensions` into multiple relevant extension classes to see if that helps with context instead?

igor-tkachev:
> There is a new standard for providing information to LLMs for various libraries: "skills".

I looked through the links you posted.

What they describe are examples of **knowledge bases for AI agents** (like `SKILL.md` files), not a formal standard for documenting libraries. They show how additional guidance can be provided to agents, but they do not define any rules for API documentation.

Also, I could not find anything there that says structured hints or annotations in API docs should not be used.

Those files are useful as **external guidance**, but they solve a different problem. In practice AI tools like Copilot mostly operate on **local context**: the current file, surrounding code, and documentation attached directly to the API. If guidance is not attached to the API, there is a good chance the model simply never sees it.

igor-tkachev:
> I don't know what affects LLM context, so I don't know what will help, but would you be willing to attempt to break out `LinqExtensions` into multiple relevant extension classes to see if that helps with context instead?

Splitting `LinqExtensions` into multiple classes is a completely different problem. That would affect the public API and discoverability, so it is not something we should change just to experiment with AI context behavior.

As for the documentation: yes, natural text helps, but **structured hints are much more reliable for LLMs**. Free-form descriptions can be interpreted in many ways, while short formal annotations are much easier to parse and act on.

LLMs typically operate on embeddings and token patterns. When information is expressed in a consistent and structured form, it produces much clearer semantic signals in the embedding space than loosely written natural text. That makes it easier for the model to reliably associate a method with its intended usage.

Placing those hints directly in **XML docs next to the API** also maximizes the chance they appear in the local context that tools like Copilot actually send to the model. It is a pragmatic solution for now, but currently one of the most effective ways to guide AI behavior with minimal impact on regular users.

viceroypenguin:
> public API and discoverability

Public API, yes. Discoverability, I disagree. Since the contents of that class are all extension methods, the discoverability of all methods is via the `IQueryable<>` (or in some cases, the l2db child interface) the methods are based on, not which class the method itself lives in.

And I'm not saying we need to unilaterally publish a version of l2db with these changes... A branch, similar to this one, can be created with said refactoring, and the ADO feed nuget could be fed to the LLM for testing purposes.

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
igor-tkachev: I understand your point about hover documentation and visual noise.

That said, the `AI-Tags` line is intentionally short and predictable, and in practice it’s a very small addition compared to the benefit it provides for AI tools. We already saw in this PR that Copilot parsed and used those tags quite effectively.

Regarding Skills: the examples you linked describe **agent skills**, which are essentially external knowledge bundles for AI agents (guidelines, workflows, best practices, etc.). They are meant to provide high-level context like *how to optimize PostgreSQL queries* or *how to approach a task*, not to describe the semantics of individual APIs.

In other words, Skills are closer to **knowledge bases or guides**, not API-level documentation.

> My point is we should do as much as we can to focus on those, including the new Skills introduced by the dotnet team for this purpose, rather than introducing documentation that will be irritating to at least some users (myself included).

Also, what the .NET team introduced there is an approach or pattern, not an established standard yet. If you look around, you’ll also find things like `llm.txt` and `llm-full.txt`. It seems the industry is still experimenting with different ways of providing context to LLMs.

From what I have seen so far, attaching short structured hints directly to the API (like AI-Tags) is currently one of the most effective ways to describe API semantics for AI agents. External files like `SKILL.md` can certainly be useful, but they solve a different problem.

PS: I also would not treat examples from the .NET team as some kind of ultimate authority here. I have worked with Microsoft guys on joint MS/IBM projects in the past and know quite a few of them personally. They are very smart people, but still people, and like everyone else they are experimenting in this space too.

to11mtm: @aaronontheweb I think this would be a good time to pitch https://github.com/Aaronontheweb/dotnet-skills as another thing to consider here 😇

igor-tkachev:
> Public API, yes. Discoverability, I disagree. 

On discoverability: extension methods are still surfaced through tooling together with their declaring types. Splitting them across multiple classes would change how they appear in documentation, navigation, and search results, so it does affect discoverability to some extent.

Anyway, this is really a separate discussion about API organization. It’s not directly related to the topic of this PR. If we want to explore refactoring `LinqExtensions`, it would be better to discuss it in a separate issue or PR.

igor-tkachev:
@copilot summarize this PR

Source/Skills/linq2db/docs/agent-antipatterns.md
```diff
@@ -0,0 +1,189 @@
...
+Data modifications on `db2` are committed immediately and are not affected by `tx.Rollback()`.
+
+**Correct pattern:**
+Pass the existing connection (or transaction) explicitly, or consolidate the work into one `DataConnection`:
+```cs
+using var db = new DataConnection(options);
+using var tx = db.BeginTransaction();
+
+// Correct: all work goes through the same connection
+db.Insert(new Order { ProductID = 1, Quantity = 10 });
```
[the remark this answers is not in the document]
igor-tkachev: It is an exaple

viceroypenguin:
Is this PR still necessary in light of #5473?

sdanyliv:
@viceroypenguin, this PR addresses the other side of the problem. We’re teaching Claude how to work with sources, while this PR teaches AI agents how to work with the library.

Shane32:
Looks great but can AI agents actually read these docs?  Do they read XML documentation?  Can it access files packaged within the NuGet package?  NuGet isn't like npm where the packages get installed to a subfolder of the application.  Does it know to do so?  Do we need to provide a sample skill to be installed by users?  There are a lot of agents now, each best for different purposes - I use Roo Code, GitHub Agent, and Codex depending on the exact scenario/code I'm working on.  We need to think about this and not just optimize for Claude Code.

MaceWindu:
Not used it, but at least context7 provides access to linq2db docs https://context7.com/linq2db/linq2db

MaceWindu:
### Open question: the XML-doc `AI-Tags` convention vs. existing docs / a structured alternative

This PR introduces machine-readable `AI-Tags` inside `<remarks>` on public API members — e.g. `AI-Tags: Group=Hints; HintType=Join; Execution=Deferred; Composability=Composable; Affects=SqlSemantics; Pipeline=ExpressionTree,SqlAST,SqlText; Provider=ProviderDefined;` — across the provider hint APIs, `DataConnection`, `DataOptions`, bulk-copy, and more. Before we lock this in as a convention, I'd like to settle **what problem it solves that isn't already covered by (a) good prose XML documentation or (b) a typed/structured alternative**, because the current shape carries a real cost.

**Core concern — audience collision.** `<remarks>` is a *human-facing* documentation element: it renders in IntelliSense tooltips, on the docfx / API-site pages, and ships verbatim in the NuGet `.xml` doc file. Embedding `AI-Tags: …` there exposes machine metadata to every human consumer of the API. On top of that the value is a free-text string — nothing validates the `Group` / `HintType` / `Execution` / `Composability` / … vocabulary, so it will drift — and it's hand-repeated per template.

**Which consumer reads the tags?** That's the question that decides the mechanism, and it maps onto the established real-world approaches:

- **An LLM reading docs offline** (the stated goal of the skill / knowledge pack) reads *prose*. Frontier models consume well-written `<summary>` / `<remarks>` / `<example>` and a docs site directly; a `Composability=Composable` taxonomy rarely beats a clear sentence. The lowest-friction "for AI" artifact here is an [`llms.txt`](https://llmstxt.org) generated from the existing docs — which this PR is already most of the way to producing via `api.md`.
- **Runtime tooling / a planner that branches on the values** wants a **typed attribute** (`[ApiSemantics(Execution = …, Composability = …)]` with enums). This is the .NET-idiomatic, compiler-validated, reflection-queryable channel — exactly what Semantic Kernel does (`[KernelFunction, Description(...)]`) and what `System.ComponentModel` / DataAnnotations / ASP.NET attributes have always done. It also keeps the tags out of human docs.
- **An agent that queries the surface programmatically** wants a **sidecar spec** keyed by doc-comment member IDs (the OpenAPI / `api.json` pattern) or an **MCP** server. Since this PR already generates `api.md` from the XML docs, a structured sidecar is a small step from there.

**The concrete open questions:**

1. Who is the intended consumer of `AI-Tags` — an offline LLM, runtime tooling, or a querying agent?
2. Given that consumer, what does the inline `<remarks>` string give us that prose docs + `llms.txt` (for an offline LLM) or a typed attribute / sidecar (for a programmatic consumer) don't — enough to justify shipping machine tags in the public, human-facing XML docs across every provider?

Not blocking the skill / knowledge-pack work — that's valuable. I'd just like the `AI-Tags`-in-`<remarks>` mechanism settled before it locks into the public doc surface. (Context: #5555 carries the same tags on the ClickHouse hints; raised there as well.)

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
igor-tkachev: all done

viceroypenguin (dismissed):
Alternative way of storing `<ai-tags />` to be non-visible to human consumers approved. Remainder of changes unreviewed.

igor-tkachev:
@MaceWindu Good question, and it deserves a real answer rather than "AI-Tags exist because the PR does them."

On why structure earns its keep over prose alone: as I noted upthread, when Copilot reviewed this PR, the inaccuracies it flagged were concentrated almost entirely in the `AI-Tags`, not in the regular XML-doc prose — and most of those flags were correct. That's not a hypothesis, it already happened in this PR's history: a controlled, grep-able key=value shape surfaced real doc bugs that free-text `<remarks>` prose describing the same members did not. `docs/api.md` is explicit that it's a search index, not something read sequentially (28k+ lines) — agents are told to grep `Search anchors:` lines for tokens like `Group=Hints`/`HintType=...` before reading prose. That's a retrieval/filtering problem, not a comprehension problem, and free-text prose is a weak filtering key precisely because the same fact ("executes immediately, not composable") can be phrased a dozen different ways across a dozen methods. A fixed vocabulary is what makes that grep reliable.

That said — I think you're right to push on the *container*, separately from the *need for structure*. Since `docs/api.md` is itself a generated bundle that these tags already get extracted into, the source of truth for the structured data doesn't have to be an XML-doc-adjacent custom tag. `api.md` itself is out as a source location — it's derived, not authored. But that leaves two real options worth comparing properly:

1. A typed attribute (compiler-validated vocabulary, reflectable without the shipped `.xml` doc file even being present) — at the cost of becoming real public API surface (`PublicAPI.Unshipped.txt`, per-TFM baselines, `CompatibilitySuppressions.xml`), which this repo otherwise treats as expensive to add.
2. An external sidecar file, hand-maintained separately from the member it describes — cheaper on the public-surface axis, but loses the co-location that currently makes a tag show up in the same diff hunk as the method it describes, which is presumably part of why it hasn't drifted so far.

I'd like to actually work through that tradeoff rather than assume the current XML-doc-tag shape is final. Will follow up with a proper comparison.

igor-tkachev:
@viceroypenguin all review threads are resolved, and the blocking requirements have been addressed. The AI metadata has been moved out of human-facing remarks into the dedicated <ai-tags /> XML-doc element, and the package-local SKILL.md structure is now the canonical agent entry point. The PR is ready for another review pass.

viceroypenguin (commented):
@igor-tkachev Please review the xmldoc for `DataConnection`, `DataConnectionExtensions.UseMemberTranslator`, and `DataConnectionExtensions.UseTracing`. Each of these has a user-visible "See <c>skills/...</>" that is not relevant for human consumers. These comments should be migrated to non-visible sources, or at the very least, commented as not-relevant to human users.

viceroypenguin (dismissed):

Issue #4325, opened 2023-10-25
Suggestion: BulkCopy with ON CONFLICT


darko1979:
I have a situation that is illustrated by following example:

```sql
CREATE TABLE IF NOT EXISTS "item"
(
  id INT NOT NULL,
  code VARCHAR(100) NOT NULL,
  CONSTRAINT pk_item PRIMARY KEY (id)
);
CREATE UNIQUE INDEX IF NOT EXISTS idx_item_code ON item USING btree (code);
```
Now I want to do a bulk insert by ignore duplicates on unique column "code" and it can be done like this:
```sql
INSERT INTO item(id,code)
VALUES (1,'1'),(2,'2'),(3,'2'),(4,'4')
ON CONFLICT DO NOTHING;
```
Existing BulkCopy method does not have an option to add ON CONFLICT at the end of statement.
Is it possible to add this option?

MaceWindu:
It is possible. PR is welcome, as we probably will not have it on our list soon.

igor-tkachev:
Is this specific database feature?

MaceWindu:
postgres and sqlite

igor-tkachev:
We can add a database specific option.

PR #5395, opened 2026-03-04
IgnoreConflicts option in BulkCopyOptions for MySql/MariaDB, PostgreSQL and SQLite

darko1979:
Implementation for feature suggestion #4325

### DIFF
--- a/Source/LinqToDB/CompatibilitySuppressions.xml
+++ b/Source/LinqToDB/CompatibilitySuppressions.xml
@@ -4569,39 +4569,4 @@
     <Right>lib/net10.0/linq2db.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
   </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:LinqToDB.Data.BulkCopyOptions.#ctor(System.Nullable{System.Int32},System.Nullable{System.Int32},LinqToDB.Data.BulkCopyType,System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.String,System.String,System.String,System.String,LinqToDB.TableOptions,System.Int32,System.Action{LinqToDB.Data.BulkCopyRowsCopied},System.Boolean,System.Nullable{System.Int32},System.Nullable{System.Int32},System.Boolean)</Target>
-    <Left>lib/net10.0/linq2db.dll</Left>
-    <Right>lib/net10.0/linq2db.dll</Right>
-    <IsBaselineSuppression>true</IsBaselineSuppression>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:LinqToDB.Data.BulkCopyOptions.#ctor(System.Nullable{System.Int32},System.Nullable{System.Int32},LinqToDB.Data.BulkCopyType,System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.String,System.String,System.String,System.String,LinqToDB.TableOptions,System.Int32,System.Action{LinqToDB.Data.BulkCopyRowsCopied},System.Boolean,System.Nullable{System.Int32},System.Nullable{System.Int32},System.Boolean)</Target>
-    <Left>lib/net8.0/linq2db.dll</Left>
-    <Right>lib/net8.0/linq2db.dll</Right>
-    <IsBaselineSuppression>true</IsBaselineSuppression>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:LinqToDB.Data.BulkCopyOptions.#ctor(System.Nullable{System.Int32},System.Nullable{System.Int32},LinqToDB.Data.BulkCopyType,System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.String,System.String,System.String,System.String,LinqToDB.TableOptions,System.Int32,System.Action{LinqToDB.Data.BulkCopyRowsCopied},System.Boolean,System.Nullable{System.Int32},System.Nullable{System.Int32},System.Boolean)</Target>
-    <Left>lib/net9.0/linq2db.dll</Left>
-    <Right>lib/net9.0/linq2db.dll</Right>
-    <IsBaselineSuppression>true</IsBaselineSuppression>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:LinqToDB.Data.BulkCopyOptions.#ctor(System.Nullable{System.Int32},System.Nullable{System.Int32},LinqToDB.Data.BulkCopyType,System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.String,System.String,System.String,System.String,LinqToDB.TableOptions,System.Int32,System.Action{LinqToDB.Data.BulkCopyRowsCopied},System.Boolean,System.Nullable{System.Int32},System.Nullable{System.Int32},System.Boolean)</Target>
-    <Left>lib/net462/linq2db.dll</Left>
-    <Right>lib/net462/linq2db.dll</Right>
-    <IsBaselineSuppression>true</IsBaselineSuppression>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:LinqToDB.Data.BulkCopyOptions.#ctor(System.Nullable{System.Int32},System.Nullable{System.Int32},LinqToDB.Data.BulkCopyType,System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.Nullable{System.Boolean},System.String,System.String,System.String,System.String,LinqToDB.TableOptions,System.Int32,System.Action{LinqToDB.Data.BulkCopyRowsCopied},System.Boolean,System.Nullable{System.Int32},System.Nullable{System.Int32},System.Boolean)</Target>
-    <Left>lib/netstandard2.0/linq2db.dll</Left>
-    <Right>lib/netstandard2.0/linq2db.dll</Right>
-    <IsBaselineSuppression>true</IsBaselineSuppression>
-  </Suppression>
 </Suppressions>


sdanyliv:
I think whe should introduce enum
```cs
public enum InsertConflictAction
{
     None,
     DoNothing,
     Update
}
```
And and name property `OnConflict`

darko1979:
> I think whe should introduce enum
> 
> ```cs
> public enum InsertConflictAction
> {
>      None,
>      DoNothing,
>      Update
> }
> ```
> 
> And and name property `OnConflict`

I agree it would be better but it is also more complicated to implement. I just need the DoNothing option.

igor-tkachev:
Is `OnConflictDoNothing` PostgreSQL-specific functionality? 

If an option is only applicable to PostgreSQL, it should be placed in `PostgreSQLOptions` rather than in the global options. Otherwise, we will gradually overload `BulkCopyOptions` with provider-specific flags, which contradicts the goal of keeping the common layer provider-agnostic.

Please move this flag to `PostgreSQLOptions`.

sdanyliv:
@igor-tkachev, historically `BulkCopyOptions` has almost everything from any providers. `PostgreSQLOptions` is not `BulkCopyOptions` and has different purpose.

igor-tkachev:
If `PostgreSQLOptions` is not intended for bulk copy–related settings, then what is `PostgreSQLOptions.BulkCopyType` doing there?

It already contains bulk copy–specific configuration. So the argument that `PostgreSQLOptions` has a completely different purpose doesn’t really work.

igor-tkachev:
I rechecked the current design.
Provider-specific options are used to define defaults, while the same setting in `BulkCopyOptions` has higher priority and overrides them.
If the intention is that all bulk copy–related settings (including provider-specific ones) must live in `BulkCopyOptions`, this should be explicitly documented, including precedence rules.
Otherwise we are mixing abstraction layers without clearly defining the contract.
But this is a more general documentation question, not something for @darko1979.

darko1979:
> Is `OnConflictDoNothing` PostgreSQL-specific functionality?
> 

[MySql / MariaDb](https://dev.mysql.com/doc/refman/8.4/en/insert.html) can ignore conflicts like this
```sql
INSERT IGNORE INTO table (id, name)
VALUES (1, 'Alice');
```
or update like this
```sql
INSERT INTO table (id, name)
VALUES (1, 'Alice')
ON DUPLICATE KEY UPDATE name = VALUES(name);
```

[SQLite](https://sqlite.org/lang_insert.html) can do this
```sql
INSERT OR IGNORE INTO table (id, name)
VALUES (1, 'Alice');
```
or this
```sql
INSERT INTO table (id, name)
VALUES (1, 'Alice')
ON CONFLICT(id) DO NOTHING;
```
[Oracle](https://docs.oracle.com/html/E26088_01/sql_elements006.htm#CHDEGDDG) can do this
```sql
INSERT /*+ IGNORE_ROW_ON_DUPKEY_INDEX(table, index_name) */
INTO table (id, name)
VALUES (1, 'Alice');

```

darko1979:
> Please move this flag to `PostgreSQLOptions`.

As I see it, options defined in `PostgreSQLOptions` are global PostgreSQL options and every provider specific options class has `BulkCopyType` that defines default `BulkCopy` mode. 
If I move this `OnConflictDoNothing` to ` PostgreSQLOptions` then this flag needs to be configured on connection level and not on `BulkCopy` operation level.
So, instead of using `BulkCopy` like this
```cs
   table.BulkCopy( new BulkCopyOptions { BulkCopyType = BulkCopyType.MultipleRows, OnConflictDoNothing = true }, rows);
```
we need to do this
```cs
using var dbConflict = GetDataContext(context, o => o.WithOptions(o.FindOrDefault(PostgreSQLOptions.Default) with { OnConflictDoNothing = true })) {
  table.BulkCopy( new BulkCopyOptions { BulkCopyType = BulkCopyType.MultipleRows }, rows);
}
```

darko1979:
I renamed this option to `IgnoreConflicts` and implemented it also for MySql/MariDB and SQLite

@sdanyliv is it better now? Can I mark this as "Ready for review" ?

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -147,7 +157,8 @@ public sealed record BulkCopyOptions
 		bool                        UseParameters          = default,
 		int?                        MaxParametersForBatch  = default,
 		int?                        MaxDegreeOfParallelism = default,
-		bool                        WithoutSession         = default
+		bool                        WithoutSession         = default,
+		bool?                       IgnoreConflicts        = default
```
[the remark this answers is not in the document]
darko1979: resolved in new commit

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -147,7 +158,8 @@ public sealed record BulkCopyOptions
...
```
MaceWindu: I think non-nullable enum looks better here as it will allow us to add at least UPDATE in future easier.
```cs
enum ConflictAction
{
Default,
Ignore
}
```

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -123,6 +124,16 @@ namespace LinqToDB.Data
...
 	/// Note that session-less connections cannot be used with session-bound functionality like temporary tables.
 	/// </param>
+	/// <param name="IgnoreConflicts">
+	/// When set to true, ignores conflicts during bulk copy operation.
+	/// Supported for following databases:
+	/// <list type="bullet">
+	/// <item>MySql/MariaDB (it will automatically fall back to <see cref="BulkCopyType.MultipleRows"/>)</item>
+	/// <item>PostgreSQL (it will automatically fall back to <see cref="BulkCopyType.MultipleRows"/>)</item>
+	/// <item>SQLite</item>
+	/// </list>
```
MaceWindu: As I requested to convert it to enum, implementation/support details should be moved to enum field help.

For each provider, which support specific action we should specify which SQL feature used to implement it as e.g. for MySQL `INSERT IGNORE` means "ignore any errors", not just conflicts.

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -157,6 +169,55 @@ public BulkCopyOptions() : this((int?)null)
 		{
 		}
 
+		[EditorBrowsable(EditorBrowsableState.Never)]
+		[Obsolete("Retained for binary compatibility; planned for removal in version 7")]
+		public BulkCopyOptions
```
MaceWindu, line 168: I'm fine with it, but I don't think we care much about BINARY compatibility

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
```diff
@@ -34,6 +34,9 @@ public MySqlBulkCopy(MySqlDataProvider provider)
 		protected override BulkCopyRowsCopied ProviderSpecificCopy<T>(
 			ITable<T> table, DataOptions options, IEnumerable<T> source)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return MultipleRowsCopy(table, options, source);
```
MaceWindu: no mode change on secondary options

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
```diff
@@ -50,6 +53,9 @@ protected override BulkCopyRowsCopied ProviderSpecificCopy<T>(
 		protected override async Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 			ITable<T> table, DataOptions options, IEnumerable<T> source, CancellationToken cancellationToken)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return await MultipleRowsCopyAsync(table, options, source, cancellationToken).ConfigureAwait(false);
```
MaceWindu: no mode change on secondary options

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
```diff
@@ -67,6 +73,9 @@ protected override async Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 		protected override async Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 			ITable<T> table, DataOptions options, IAsyncEnumerable<T> source, CancellationToken cancellationToken)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return await MultipleRowsCopyAsync(table, options, source, cancellationToken).ConfigureAwait(false);
```
MaceWindu: no mode change on secondary options

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
```diff
@@ -55,6 +62,9 @@ protected override Task<BulkCopyRowsCopied> MultipleRowsCopyAsync<T>(
 		protected override BulkCopyRowsCopied ProviderSpecificCopy<T>(
 			ITable<T> table, DataOptions options, IEnumerable<T> source)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return MultipleRowsCopy(table, options, source);
```
MaceWindu: no mode change on secondary options

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
```diff
@@ -64,6 +74,9 @@ protected override BulkCopyRowsCopied ProviderSpecificCopy<T>(
 		protected override Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 			ITable<T> table, DataOptions options, IEnumerable<T> source, CancellationToken cancellationToken)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return MultipleRowsCopyAsync(table, options, source, cancellationToken);
```
MaceWindu: no mode change on secondary options

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
```diff
@@ -73,6 +86,9 @@ protected override Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 		protected override Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
 			ITable<T> table, DataOptions options, IAsyncEnumerable<T> source, CancellationToken cancellationToken)
 		{
+			if (options.BulkCopyOptions.IgnoreConflicts == true)
+				return MultipleRowsCopyAsync(table, options, source, cancellationToken);
```
MaceWindu: no mode change on secondary options

Tests/Linq/DataProvider/SQLiteTests.cs
MaceWindu, line 1: Instead of 3 sets of tests, could we add one to BulkCopyTests and enabable it for 3 providers.

Source/LinqToDB/Data/BulkCopyOptions.cs
```diff
@@ -123,6 +124,16 @@ namespace LinqToDB.Data
 	/// Implemented only by ClickHouse.Driver provider. When set, provider-specific bulk copy will use session-less connection even if called over connection with session.
 	/// Note that session-less connections cannot be used with session-bound functionality like temporary tables.
 	/// </param>
+	/// <param name="IgnoreConflicts">
+	/// When set to true, ignores conflicts during bulk copy operation.
+	/// Supported for following databases:
+	/// <list type="bullet">
+	/// <item>MySql/MariaDB (it will automatically fall back to <see cref="BulkCopyType.MultipleRows"/>)</item>
+	/// <item>PostgreSQL (it will automatically fall back to <see cref="BulkCopyType.MultipleRows"/>)</item>
```
MaceWindu: Also per-provider it should specify supported bulk copy modes, no mode change should be done on our side - if user specify action where it is not supported - it is ignored.
Right now we change mode on our side in two cases:
1. when provider doesn't support specific mode we downgrade to supported one
2. in rare cases when we can detect that specific request will fail (for some buggy provider) we also downgrade it
In all other cases we use only options, supported for provider in specified mode.

This also important for situations where we want to add support for this option for provider-specific bulk copy. I've checked providers and currently only one provider has such support:
- MySqlConnector supports ignore/update action : https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySqlBulkLoaderConflictOption.cs

If you want, you could try to add support for it too

Source/LinqToDB/Data/BulkCopyOptions.cs
darko1979: created new enum ConflictAction and replaced bool IgnoreConflicts parameter

Source/LinqToDB/Data/BulkCopyOptions.cs
darko1979: descriptions added to new ConflictAction enum

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
darko1979: removed

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
darko1979: lines removed

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
darko1979: lines removed

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
darko1979: lines removed

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
darko1979: lines removed

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
darko1979: lines removed

Tests/Linq/DataProvider/SQLiteTests.cs
darko1979: tests moved to single file .\Tests\Linq\Linq\ConflictActionTests.cs

Source/LinqToDB/Data/BulkCopyOptions.cs
darko1979: description updated what option Ignore does for each supported provider and that only works if MultipleRows mode is used (for MySql and PostrgreSQL)

Source/LinqToDB/Data/BulkCopyOptions.cs
darko1979: removed, resolved with suppressions

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
```diff
@@ -34,6 +34,13 @@ public PostgreSQLBulkCopy(PostgreSQLDataProvider dataProvider)
 			_provider = dataProvider;
 		}
 
+		protected override string? GetMultipleRowsSuffix(BulkCopyOptions options)
+		{
+			if (options.ConflictAction == ConflictAction.Ignore)
+				return "\nON CONFLICT DO NOTHING";
```
MaceWindu:
```suggestion
				return $"{Environment.NewLine}ON CONFLICT DO NOTHING";
```

Source/LinqToDB/Data/BulkCopyOptions.cs
viceroypenguin, line 215: Add a third public constructor, without the `ConflictAction` parameter. Don't pass any default values and mark it `[EditorBrowsable(None)]` and `[Obsolete]`. This is to retain binary compatibility. See commit 040e5b170cb7bcab9aebbe99bee243f62f9048c0 for example

Source/LinqToDB/Internal/DataProvider/MySql/MySqlBulkCopy.cs
```diff
@@ -341,6 +341,13 @@ await TraceActionAsync(
 			return rc;
 		}
 
+		protected override string GetInsertInto(MultipleRowsHelper helper)
+		{
+			if (helper.Options.BulkCopyOptions.ConflictAction == ConflictAction.Ignore)
+				return $"INSERT IGNORE INTO {helper.TableName}";
+			return base.GetInsertInto(helper);
```
viceroypenguin: For easier adaptation in the future if necessary:

```suggestion
			return helper.Options.BulkCopyOptions.ConflictAction switch
			{
				ConflictAction.Ignore => $"INSERT IGNORE INTO {helper.TableName}",
				_                     => base.GetInsertInto(helper),
			};
```

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
```diff
@@ -34,6 +34,13 @@ public PostgreSQLBulkCopy(PostgreSQLDataProvider dataProvider)
 			_provider = dataProvider;
 		}
 
+		protected override string? GetMultipleRowsSuffix(BulkCopyOptions options)
+		{
+			if (options.ConflictAction == ConflictAction.Ignore)
```
viceroypenguin: Same as previous

Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteBulkCopy.cs
```diff
@@ -19,6 +19,13 @@ public class SQLiteBulkCopy : BasicBulkCopy
 		/// </remarks>
 		protected override int MaxSqlLength => 1000000;
 
+		protected override string GetInsertInto(MultipleRowsHelper helper)
+		{
+			if (helper.Options.BulkCopyOptions.ConflictAction == ConflictAction.Ignore)
```
viceroypenguin: Same

Source/LinqToDB/Data/BulkCopyOptions.cs
darko1979: I already did than, than @MaceWindu said that you (as in we) don't care about that, so I removed it and left only in suppressions. Should I add it again?

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLBulkCopy.cs
darko1979: added

Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteBulkCopy.cs
darko1979: added

Source/LinqToDB/DataOptionsExtensions.cs
```diff
@@ -1645,7 +1645,17 @@ public static BulkCopyOptions WithMaxDegreeOfParallelism(this BulkCopyOptions op
 		[Pure]
 		public static BulkCopyOptions WithWithoutSession(this BulkCopyOptions options, bool withoutSession)
 		{
-			return options with { WithoutSession = withoutSession };
+				return options with { WithoutSession = withoutSession };
```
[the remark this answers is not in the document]
darko1979: fixed

Source/LinqToDB/Data/ConflictAction.cs
```diff
@@ -0,0 +1,24 @@
...
+		/// </summary>
+		Default = 0,
+
+		/// <summary>
+		/// Ignore conflicts during bulk copy operation.
+		/// Supported for following databases:
+		/// <list type="bullet">
+		/// <item>MySql/MariaDB - uses INSERT IGNORE INTO statement (only works with <see cref="BulkCopyType.MultipleRows"/>)</item>
+		/// <item>PostgreSQL - adds ON CONFLICT DO NOTHING clause to INSERT INTO statement (only works with <see cref="BulkCopyType.MultipleRows"/>)</item>
+		/// <item>SQLite - uses INSERT OR IGNORE INTO statement</item>
```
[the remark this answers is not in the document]
darko1979: added

Tests/Linq/Linq/ConflictActionTests.cs
```diff
@@ -0,0 +1,117 @@
...
+
+using LinqToDB;
+using LinqToDB.Async;
+using LinqToDB.Common;
+using LinqToDB.Data;
+using LinqToDB.Internal.DataProvider.MySql;
+using LinqToDB.Mapping;
+using LinqToDB.SchemaProvider;
+using LinqToDB.Tools;
+using LinqToDB.Tools.Comparers;
```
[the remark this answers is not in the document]
darko1979: removed suggested

Tests/Linq/DataProvider/PostgreSQLTests.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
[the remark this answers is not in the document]
darko1979: reverted to original file

Source/LinqToDB/Data/BulkCopyOptions.cs
viceroypenguin: Yes, please add it again. @MaceWindu is apathetic to it, but I think it's important. 🙂

PR #5397, opened 2026-03-04
Issue/5396 - lateral join emulation

### LINKED ISSUE #5396: Emulate Lateral/Apply join support for Sqlite (or other providers)
### Description

Some Databases do not have lateral/apply join support, but it could be emulated by writing different querys.

### New API

_No response_

jogibear9988:


sdanyliv:
@jogibear9988, please stop experimenting with this. **linq2db** doesn’t require emulation of `OUTER APPLY` joins - it already has built-in support and handles it automatically.

jogibear9988:
> @jogibear9988, please stop experimenting with this. **linq2db** doesn’t require emulation of `OUTER APPLY` joins - it already has built-in support and handles it automatically.

Only wanted to see what works with this and what not.

The queries in CountBy, ExceptBy, IntersectBy and Index wich are not run for Sqlite cause of missing apply/lateral join work with this change. They require some sort of emulation. But for sure I'm not very experienced with this, so it's only a test, nothing wich can be merged.

jogibear9988:
The PR compiles and several SQLite tests pass, but there are cases where the rewrite can change query semantics and silently return incorrect results.

## Main concerns

### 1. Moving predicates out of an inner `LEFT JOIN` is not generally valid

`TryDecorrelateApplyJoinConditions` extracts correlated predicates from inner join conditions without checking the inner join type:

https://github.com/linq2db/linq2db/blob/d757a59cb540ef4820141b6532dd1ad6a925bbb5/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs#L1404-L1416

For example, this:

```sql
A OUTER APPLY (
    B LEFT JOIN C ON C.Key = A.Id
)
```

is effectively rewritten as:

```sql
A LEFT JOIN (
    B LEFT JOIN C ON TRUE
) ON C.Key = A.Id
```

These queries are not equivalent. If `A.Id = 1`, `B` contains a row, and `C` has no matching row, the original query preserves the `B` row with null values for `C`. The rewritten query can instead null out the entire inner result, including the `B` columns.

The tests covering this pattern only verify that the result is non-null and contains at least one row:

https://github.com/linq2db/linq2db/blob/d757a59cb540ef4820141b6532dd1ad6a925bbb5/Tests/Linq/Linq/ApplyJoinTests.cs#L20-L35

They should compare the complete result set, including unmatched rows and projected values.

### 2. A failed rewrite can leave the query tree partially modified

`ExtractCorrelatedPredicates` removes predicates immediately. If it later encounters an unsupported correlated predicate, it returns `false` without restoring the predicates already removed:

https://github.com/linq2db/linq2db/blob/d757a59cb540ef4820141b6532dd1ad6a925bbb5/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs#L1607-L1633

The same can happen when one inner join is processed successfully but a later join cannot be decorrelated. A `Try...` method should either analyze everything before mutating the query or roll back all changes when it returns `false`.

### 3. `IsApplyJoinEmulated` does not gate the new behavior

SQLite sets `IsApplyJoinEmulated = true`, but the optimizer checks only `!IsApplyJoinSupported`:

https://github.com/linq2db/linq2db/blob/d757a59cb540ef4820141b6532dd1ad6a925bbb5/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs#L1214-L1221

Consequently, the new rewrite is attempted for every provider without native `APPLY` support, not only SQLite. The new flag currently appears to affect serialization/equality only and does not control the optimizer.

### 4. Window-function rewriting is incomplete

`ExtendWindowPartitionByFromPredicates` handles only:

- A window function used directly as a selected expression.
- A window function immediately inside a binary expression.

https://github.com/linq2db/linq2db/blob/d757a59cb540ef4820141b6532dd1ad6a925bbb5/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs#L1669-L1696

Window functions nested inside casts, functions, conditions, or more complex expressions can therefore be decorrelated without receiving the required partition columns. That changes ranking/window semantics.

## CI status

The code builds, and the Windows SQLite and SQLiteExtras jobs passed. However, the full test run was not green: Firebird 2.5, Firebird 3, DB2, and Access MDB failed, after which many remaining jobs were cancelled.

It is not clear that every failure was caused by this PR, but the current CI run does not establish that the optimizer changes are safe across providers.

## Suggested direction

I think decorrelation can work for a carefully defined subset of queries, but it should not be treated as general `APPLY` emulation. At minimum, I would suggest:

1. Restricting predicate extraction to join types for which the transformation is proven equivalent.
2. Analyzing the complete query before mutating it.
3. Gating the behavior with the new provider flag.
4. Traversing and rewriting all window functions recursively.
5. Using result-comparison tests (`AssertQuery`) for every rewrite.
6. Adding tests for unmatched rows, null keys, duplicate keys, mixed supported/unsupported predicates, nested joins, aggregates, and complex window expressions.

As an experiment this is useful, but in its current form it can produce incorrect results and should not be merged.

PR #5201, opened 2025-11-24
General code modernization

viceroypenguin:
Changes:
- various code style updates (`switch expression`, pattern matching, etc)
- enable and use more runtime API polyfills
- merge `.editorconfig` files into single file
- use `Ordinal` comparison for strings in more places (string equality, hashed collections)
- use exception throw helpers
- get rid of `Tuple<>` types
- demand parenthesis for clarity

PR #5401, opened 2026-03-05
Add support for ClickHouse.Driver v1.0 provider

MaceWindu:


Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSchemaProvider.cs
```diff
@@ -227,5 +196,56 @@ private static (DataType dataType, Type? type) GetTypeMapping(string? dataType,
 			{ "Float64"   , (DataType.Double    , typeof(double        )) },
 		};
 		#endregion
+
+		#region Mappings
```
viceroypenguin, line 200: should we look at using similar mappings for other schema providers in the future?

MaceWindu: We must and I wanted to do it for a long time. I did it here, because with binary/string read behavior switch in new provider version, old code with raw data readers just doesn't work anymore

viceroypenguin: Create an issue and assign it to me for v6.3.0; I'll do it after #5201 comes in (will affect same code)

PR #5402, opened 2026-03-05
Add async code analyzers

MaceWindu:


Source/LinqToDB/Data/DataConnection.QueryRunner.cs
```diff
@@ -460,8 +460,7 @@ public static async Task<int> ExecuteNonQueryAsync(
 				var commandsParameters = GetParameters(dataConnection, preparedQuery, parameterValues);
 				var executionQuery     = new ExecutionPreparedQuery(preparedQuery, commandsParameters);
 
-				return await ExecuteNonQueryImplAsync(dataConnection, executionQuery, cancellationToken)
-					.ConfigureAwait(false);
+				return ExecuteNonQueryImplAsync(dataConnection, executionQuery, cancellationToken);
 			}
```
viceroypenguin, line 465: I would prefer not to make these kinds of changes. They remove the introduction of `ExecuteNonQueryAsync()` into the call-stack, making the relationship between the caller and `ExecuteNonQueryImplAsync()` non-obvious. Further, in net11, this will actually be an antipattern, reducing performance rather than improving it.

MaceWindu: I don't have strong feelings about one or another approach, but if it will mater for native async (I have no idea right now) we could address it everywhere before .net 11

MaceWindu: Reverted those changes

viceroypenguin (approved):

PR #5414, opened 2026-03-11
Improved UPDATE transformation for providers that do not support UPDATE FROM.

### LINKED ISSUE #5413: Subquery in update issues (v6 regression)
### Description

In specific (and unclear to me) circumstances, bad transformations of LINQ query generate invalid SQL query.
The reproduction is silly and full of constants, but please keep in mind it's a minimal reproduction reduced from a large and complex query.
**I could only reproduce the bug with Oracle provider, the same query generates correct SQL for Postgres for example.**

The provided test case is:

```csharp
db.Person
  .Where(x => x.ID == id)
  .Set(
  	x => Sql.Row(x.FirstName, x.LastName),
  	x => (
  	    from start in db.SelectQuery(() => 1)
  		from canChange in db.SelectQuery(() => x.ID > -1).DefaultIfEmpty()
  		select Sql.Row(
  			canChange ? "Tooth" : x.FirstName,
  			canChange ? "Fairy" : x.LastName
  		)
  	).Single()
  )
  .Update();
```

The generated SQL is:
```sql
UPDATE
  "Person"
 SET
  ("FirstName", "LastName") = (
          SELECT
                  CASE
                          WHEN canChange."c1" = 1 THEN 'Tooth'
                          ELSE "Person"."FirstName"
                  END,
                  CASE
                          WHEN canChange."c1" = 1 THEN 'Fairy'
                          ELSE "Person"."LastName"
                  END
          FROM
                  "Person" x_1
                          LEFT JOIN (
                                  SELECT
                                          CASE
                                                  WHEN x_1."PersonID" > -1 THEN 1
                                                  ELSE 0
                                          END as "c1"
                                  FROM SYS.DUAL
                          ) canChange ON 1=1
          WHERE
                  x_1."PersonID" = :id AND "Person"."PersonID" = x_1."PersonID"
  )
 WHERE
  EXISTS(
          SELECT
                  *
          FROM
                  "Person" x
          WHERE
                  x."PersonID" = :id AND "Person"."PersonID" = x."PersonID"
  )
```

First notice the weird `WHERE EXISTS` this would work but is completely unnecessary as `WHERE PersonID = :id` is the expected translation.

The same construct is found inside the subquery: linq2db has introduced an addition `FROM Person` which is linked to the update table in `WHERE` with the same pattern as the outside `WHERE EXISTS`.
This is again completely unnecessary and is usually not generated when the bug does not happen.

Now the bug is that inside the `LEFT JOIN`, linq2db referes to `x_1` which is not allowed in a join.
Because of this correlation an `OUTER APPLY` would be more correct and work, but there's a much better query.
Remember the `FROM Person` is totally unnecessary and unexpected: if `x_1` is replaced by `"Person"` (the updated table) then the query would work as the updated table is allowed to be referenced from inside a subquery.

The following translation would work and would feel natural based on C# query:

```sql
UPDATE Person
SET (FirstName, LastName) = (
  SELECT
          CASE
                  WHEN canChange.c1 = 1 THEN 'Tooth'
                  ELSE Person.FirstName
          END,
          CASE
                  WHEN canChange.c1 = 1 THEN 'Fairy'
                  ELSE Person.LastName
          END
  FROM (SELECT 1 FROM SYS.DUAL) start
  LEFT JOIN (
      SELECT
              CASE
                      WHEN Person.PersonID > -1 THEN 1
                      ELSE 0
              END as c1
      FROM SYS.DUAL
  ) canChange ON 1=1
)
WHERE Person.PersonID = :id
```

### Reproduction Steps

Test case provided in #5414

### Exception details

ORA error during execution because SQL query is invalid

### Regression?

Regression since 6.0

### Linq To DB version

6.2-dev.19204

### .NET version

10.0

### Database

Oracle 19

### Database provider

Managed ODP.NET

### Operating system

Windows

### Configuration

_No response_

jods4:
#5413


Source/LinqToDB.Scaffold/CompatibilitySuppressions.xml
viceroypenguin, line 1: Incorrect removal?

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerSqlOptimizer.cs
```diff
@@ -76,7 +76,7 @@ SqlUpdateStatement CorrectSqlServerUpdate(SqlUpdateStatement updateStatement, Da
 				// do not remove if there is other tables
 				if (QueryHelper.EnumerateAccessibleTables(updateStatement.SelectQuery).Take(2).Count() == 1)
 				{
-					if (RemoveUpdateTableIfPossible(updateStatement.SelectQuery, updateStatement.Update.Table, out removedTableSource))
+					if (RemoveUpdateTableIfPossible(updateStatement.SelectQuery, updateStatement.Update.Table, false, out removedTableSource))
```
viceroypenguin:
```suggestion
					if (RemoveUpdateTableIfPossible(updateStatement.SelectQuery, updateStatement.Update.Table, allowLeftJoin: false, out removedTableSource))
```

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1169,187 +1149,384 @@ protected SqlUpdateStatement GetAlternativeUpdate(SqlUpdateStatement updateState
...
 
-			// It covers subqueries also. Simple subquery will have sourcesCount == 2
-			if (QueryHelper.EnumerateAccessibleTableSources(updateStatement.SelectQuery).Any())
-			{
-				var sql = new SelectQuery { IsParameterDependent = updateStatement.IsParameterDependent  };
+			if (updateStatement.SelectQuery.HasGroupBy)
+				return false;
+
+			if (updateStatement.SelectQuery.From.Tables.Count != 1)
+				return false;
```
viceroypenguin, line 1131:
```suggestion
			if (updateStatement is not 
				{
					Update.Table: { },
					SelectQuery:
					{
						HasGroupBy: false,
						From.Tables.Count: 1,
					},
				})
			{
				return false;
			}
```

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1169,187 +1149,384 @@ protected SqlUpdateStatement GetAlternativeUpdate(SqlUpdateStatement updateState
...
+				throw new InvalidOperationException("Update table cannot be null");
 
-							rows.AddRange(GenerateRows(item.Column, item.Expression, replaceTree, innerTree, innerQuery));
-						}
+			// extracting values, which should use subquery
 
-						var sqlRow        = new SqlRowExpression(rows.ToArray());
-						var newUpdateItem = new SqlSetExpression(sqlRow, innerQuery);
+			var newItems          = new List<SqlSetExpression>();
+			var dependedArguments = new List<SqlSetExpression>();
```
viceroypenguin:
```suggestion
			var dependentArguments = new List<SqlSetExpression>();
```

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
```diff
@@ -188,6 +191,14 @@ protected internal override IQueryElement VisitSqlJoinedTable(SqlJoinedTable ele
 
 		protected internal override IQueryElement VisitSqlQuery(SelectQuery selectQuery)
 		{
+			// simplify select query expression
+			if (_correcting == null && _isExpression && selectQuery is { HasNoTables: true, IsSingleColumn: true, HasSetOperators: false, Where.SearchCondition.IsTrue: true, HasGroupBy: false, DoNotRemove: false })
+			{
+				var columnExpression = selectQuery.Select.Columns[0].Expression;
+				if (columnExpression is not SqlRowExpression)
+					return Visit(columnExpression);
+			}
```
viceroypenguin:
```suggestion
			if (_correcting == null
                && _isExpression
                && selectQuery is
                {
                    HasNoTables: true,
                    IsSingleColumn: true,
                    HasSetOperators: false,
                    Where.SearchCondition.IsTrue: true,
                    HasGroupBy: false,
                    DoNotRemove: false,
                    Select.Columns: [{ Expression: not SqlRowExpression and var expr }, ..],
                })
            {
					return Visit(columnExpression);
			}
```

Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
```diff
@@ -1824,7 +1821,42 @@ public static bool IsLimitedToOneRecord(SqlJoinedTable joinedTable)
...
+							if (leftSource == joinedSource && rightSource != joinedSource)
+								equalityKeys.Add(exprExpr.Expr1);
+							else if (rightSource == joinedSource && leftSource != joinedSource)
+								equalityKeys.Add(exprExpr.Expr2);
+						}
+					}
+
+					foreach (var key in keys)
+					{
+						if (key.All(k => equalityKeys.Contains(k)))
```
viceroypenguin:
```suggestion
						if (key.All(equalityKeys.Contains))
```

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1337,13 +1513,25 @@ protected void CorrectSetters(List<SqlSetExpression> setters, SelectQuery query)
 					{
 						if (SqlProviderFlags.RowConstructorSupport.HasFlag(RowFeature.UpdateLiteral))
 						{
+							// Provider supports row literals — collapse `SELECT v1, v2` back to `Row(v1, v2)`.
 							var rowValues = subQuery.Select.Columns.Select(c => c.Expression).ToArray();
 							item.Expression = new SqlRowExpression(rowValues);
 						}
+						// else: leave as SelectQuery; FlattenRowConstructors will expand into individual setters.
 					}
```
MaceWindu: **MIN004** — The defensive `LinqToDBException` mentions the expected row length but omits the actual RHS type. When this guard fires ("Any other shape is a logic error upstream"), diagnosis is harder without seeing what shape actually appeared.

Fix: include `item.Expression?.GetType().Name ?? "null"` in the exception message.

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1080,222 +1060,418 @@ protected static TElement RemapCloned<TElement>(
...
+
+			if (!tableSource.Joins.TrueForAll(j =>
+				    (j.JoinType is JoinType.Left or JoinType.OuterApply || (j.JoinType is JoinType.Inner && j.IsSubqueryExpression)) && QueryHelper.IsLimitedToOneRecord(j)))
 			{
-				var sqlTableSource = updateStatement.SelectQuery.From.Tables[0];
-				if (sqlTableSource.Source == updateStatement.Update.Table && sqlTableSource.Joins.Count == 0)
+				return false;
+			}
+
+			// Build a correlated subquery from the outer joins. The first join is folded into
```
MaceWindu: **SUG001** — The join-acceptance predicate in `TryMoveUpdateToSubQuery` inlines a non-trivial condition: three `JoinType` checks, an `IsSubqueryExpression` check, and a helper call. Pulling it into a named local function (e.g. `static bool CanFoldJoinIntoScalarSubquery(SqlJoinedTable j)`) would make the intent legible without re-reading the surrounding comment.

Additionally: consider whether `CrossApply` should also be accepted — when combined with `IsLimitedToOneRecord`, it has the same single-match semantics as `OuterApply`.

Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
```diff
@@ -1824,7 +1821,42 @@ public static bool IsLimitedToOneRecord(SqlJoinedTable joinedTable)
...
+							var leftSource  = ExtractSqlSource(exprExpr.Expr1);
+							var rightSource = ExtractSqlSource(exprExpr.Expr2);
+
+							if (leftSource == joinedSource && rightSource != joinedSource)
+								equalityKeys.Add(exprExpr.Expr1);
+							else if (rightSource == joinedSource && leftSource != joinedSource)
+								equalityKeys.Add(exprExpr.Expr2);
+						}
+					}
+
```
MaceWindu: **SUG002** — The uniqueness analysis in `IsLimitedToOneRecord(SqlJoinedTable)` uses `key.All(equalityKeys.Contains)` at the end, which is O(n*m). For composite primary keys with many columns and joins carrying many predicates, this becomes quadratic. Additionally, `List<T>.Contains` uses the default equality comparer — reference equality for `ISqlExpression` here — which is correct but should be documented.

Fix: replace `equalityKeys` with a `HashSet<ISqlExpression>` (default referential comparer) to document intent and make `Contains` O(1). Add a one-line comment stating reference equality is intentional.

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1080,222 +1060,418 @@ protected static TElement RemapCloned<TElement>(
 			yield return (target, source, null);
 		}
 
-		static IEnumerable<ISqlExpression> GenerateRows(
-			ISqlExpression                            target,
-			ISqlExpression                            source,
-			Dictionary<IQueryElement, IQueryElement>? mainTree,
-			Dictionary<IQueryElement, IQueryElement>? innerTree,
-			SelectQuery                               selectQuery)
+		protected SqlUpdateStatement GetAlternativeUpdate(SqlUpdateStatement updateStatement, DataOptions dataOptions, MappingSchema mappingSchema)
```
MaceWindu: **NIT001** — Inconsistent `InvalidOperationException` messages within the same class. `GetAlternativeUpdate` and `GetAlternativeUpdatePostgreSqlite` throw "Update.Table is required."; `ProcessUpdateItemsWithoutRows`, `ProcessUpdateItemsWithRows`, and `MakeUniversalUpdate` throw "Update table cannot be null". Pick one form and apply everywhere — drift across five sibling throws is easy to notice in this column-aligned file.

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
```diff
@@ -318,43 +338,58 @@ protected internal override IQueryElement VisitSqlQuery(SelectQuery selectQuery)
...
 			_currentSetOperator = saveCurrent;
+			_isExpression = saveIsExpression;
 
 			return newElement;
 		}
 
 		protected internal override IQueryElement VisitInSubQueryPredicate(SqlPredicate.InSubQuery predicate)
 		{
 			var saveInsubquery = _inSubquery;
+			var saveIsExpression = _isExpression;
```
MaceWindu: **NIT002** — Trailing whitespace on the blank line after `var saveIsExpression = _isExpression;` here and in several other new `_isExpression` save/restore blocks (`VisitSqlOrderByClause`, `VisitSqlGroupByClause`, `VisitSqlColumnExpression`, `VisitSqlSearchCondition`, `VisitExistsPredicate`). Not a formal rule, but existing code doesn't leave this pattern.

Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs
```diff
@@ -1013,6 +1013,22 @@ public static string SuggestTableSourceAlias(SelectQuery selectQuery, string ali
 			};
 		}
 
+		/// <summary>
+		/// Returns table source from specific expression. Usually from SqlColumn or SqlField.
+		/// <param name="expression">Expression to extract table source from.</param>
+		/// <returns>Table source associated with expression</returns>
+		/// </summary>
```
MaceWindu, line 1018: **MIN001** — The `<param>` and `<returns>` tags are placed inside the `<summary>` element. XML doc comments require these to be siblings of `<summary>`, not nested children — the current layout silently omits the param/return descriptions from generated docs and most IDE tooltips, and can trip CS1570 depending on analyzer settings. Still present on HEAD after being flagged in earlier review.

Fix: Close `</summary>` before the `<param>` and `<returns>` tags so they sit at the top level of the doc comment.

```suggestion
		/// <summary>
		/// Returns table source from specific expression. Usually from SqlColumn or SqlField.
		/// </summary>
		/// <param name="expression">Expression to extract table source from.</param>
		/// <returns>Table source associated with expression.</returns>
```

— ✓ Fixed in 83d743bb

Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs
```diff
@@ -1284,6 +1284,11 @@ protected internal override IQueryElement VisitExistsPredicate(SqlPredicate.Exis
 				}
 			}
 
+			if (predicate.SubQuery is { HasNoTables: true, HasSetOperators: false, Where.SearchCondition.IsTrue: true, HasGroupBy: false })
```
MaceWindu: **MIN002** — The guard only checks `HasNoTables`, `HasSetOperators`, `Where.SearchCondition.IsTrue`, and `HasGroupBy`. It ignores HAVING, TAKE, and SKIP — all of which can still make a tabless SELECT produce zero rows, in which case `EXISTS(...)` should return `false`, not `true`. `SELECT 1 HAVING 1=0`, `SELECT 1 FETCH FIRST 0 ROWS ONLY`, `SELECT 1 OFFSET 1 ROWS` are all zero-row constant-shape queries. Most providers are unlikely to generate such shapes internally, but the simplification is unconditional — one surprising upstream shape is enough to silently change query semantics.

Fix: Widen the guard to also require `HasHaving: false` and `IsLimited: false` (equivalently `Select.TakeValue is null && Select.SkipValue is null`).

```suggestion
			if (predicate.SubQuery is { HasNoTables: true, HasSetOperators: false, Where.SearchCondition.IsTrue: true, HasGroupBy: false, HasHaving: false, IsLimited: false })
```

— ✓ Fixed in 83d743bb

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
```diff
@@ -188,6 +191,23 @@ protected internal override IQueryElement VisitSqlJoinedTable(SqlJoinedTable ele
...
+				&& selectQuery is
+				{
+					HasNoTables: true,
+					IsSingleColumn: true,
+					HasSetOperators: false,
+					Where.SearchCondition.IsTrue: true,
+					HasGroupBy: false,
+					DoNotRemove: false,
+					Select.Columns: [{ Expression: not SqlRowExpression and var columnExpression }, ..],
+				})
```
MaceWindu, line 207: **MIN003** — Same class of gap as the EXISTS simplification. A tabless single-column SELECT is replaced by its column expression in expression position, but the pattern does not check `HasHaving`, `Select.HasModifier` (TAKE/SKIP/DISTINCT), or `HasOrderBy`. `SELECT TOP 0 x` and `SELECT x HAVING 1=0` are legitimate zero-row scalar subqueries; replacing them with `x` changes the evaluated value from NULL (empty scalar subquery) to `x` itself.

Fix: Add `HasHaving: false` and `Select.HasModifier: false` to the pattern so modifiers and HAVING prevent inlining.

```suggestion
			if (_correcting == null
				&& _isExpression
				&& selectQuery is
				{
					HasNoTables: true,
					IsSingleColumn: true,
					HasSetOperators: false,
					Where.SearchCondition.IsTrue: true,
					HasGroupBy: false,
					HasHaving: false,
					Select.HasModifier: false,
					DoNotRemove: false,
					Select.Columns: [{ Expression: not SqlRowExpression and var columnExpression }, ..],
				})
```

— ✓ Fixed in 83d743bb

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1080,222 +1060,418 @@ protected static TElement RemapCloned<TElement>(
...
+							if (IsDependedExceptedSource(value, updateStatement.Update.Table))
+								dependentArguments.Add(new SqlSetExpression(columnsRow.Values[i], value));
+							else
+								newItems.Add(new SqlSetExpression(columnsRow.Values[i], value));
+						}
 					}
+					else
+					{
+						newItems.Add(item);
+					}
```
MaceWindu, line 1319: **MIN004** — When the setter has a row-shaped column (`Column is SqlRowExpression`) but the RHS is neither an inline `SqlRowExpression` nor a `SelectQuery` with matching column count, `ProcessUpdateItemsWithRows` falls through to re-emitting the item verbatim. This branch is reached only after the guard above has already confirmed `IsDependedExceptedSource(item.Expression, Update.Table) == true` — meaning the RHS references another source. After `TryMoveUpdateToSubQuery` clears `SelectQuery.From.Tables`, the outer UPDATE no longer has access to that source, so the resulting setter dangles a reference to an unreachable source. At best this fails in `SqlProviderHelper.IsValidQuery`; at worst it silently generates broken SQL.

Fix: Either (a) treat the fallback as a dependent setter and push it through the `dependentArguments` path so it becomes a correlated subquery, or (b) throw `InvalidOperationException` to signal the shape isn't handled — mirroring the exception `FlattenRowConstructors` already throws for the same shape invariant.

— ✓ Fixed in 83d743bb

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -893,7 +873,7 @@ protected bool NeedsEnvelopingForUpdate(SelectQuery query)
 			return false;
 		}
 
-		protected bool RemoveUpdateTableIfPossible(SelectQuery query, SqlTable table, out SqlTableSource? source)
+		protected bool RemoveUpdateTableIfPossible(SelectQuery query, SqlTable table, bool allowLeftJoin, out SqlTableSource? source)
```
MaceWindu, line 881: **MIN005** — The new `allowLeftJoin` parameter is a breaking signature change on a `protected` API — yet no call site in the repo passes `true`. All internal callers and all three per-provider optimizer callers pass `false`. If the plan is to wire this in follow-up work, that's fine, but right now the parameter is a dead knob that forces every existing override and downstream subclass to adapt for no behavior change.

Fix: Either (a) identify and update the caller that intended `allowLeftJoin: true` in this PR, or (b) give the parameter a default value of `false` (`bool allowLeftJoin = false`) — the latter keeps the protected signature a non-breaking addition for subclasses and allows the associated `CompatibilitySuppressions.xml` CP0002 suppression to be dropped.

— ✓ Fixed in 83d743bb

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1080,222 +1060,418 @@ protected static TElement RemapCloned<TElement>(
...
+			if (IsDependedExceptedSource(updateStatement.SelectQuery.Where, updateStatement.Update.Table))
+				return false;
+
+			if (!tableSource.Joins.TrueForAll(j =>
+				    (j.JoinType is JoinType.Left or JoinType.OuterApply || (j.JoinType is JoinType.Inner && j.IsSubqueryExpression)) && QueryHelper.IsLimitedToOneRecord(j)))
 			{
-				var sqlTableSource = updateStatement.SelectQuery.From.Tables[0];
-				if (sqlTableSource.Source == updateStatement.Update.Table && sqlTableSource.Joins.Count == 0)
+				return false;
+			}
```
MaceWindu, line 1146: **MIN006** — The comment above explains that dropping the first join's `JoinType` is safe because a scalar subquery returning zero rows yields NULL, matching `LEFT` / `OUTER APPLY` semantics. That holds for `JoinType.Left` / `JoinType.OuterApply`, but the predicate also admits `JoinType.Inner && IsSubqueryExpression`. Original `INNER JOIN` semantics filter the outer UPDATE row when no match exists; after the rewrite, the outer UPDATE row is kept and the setter becomes NULL. This is a behavior change that's masked only because `IsSubqueryExpression` joins are synthesized from subquery-in-expression LINQ and usually have `IsLimitedToOneRecord` guaranteeing one match — but that isn't the same as guaranteeing "never zero matches".

Fix: Either (a) drop `JoinType.Inner` from the allow-list and restrict the subquery form to joins that truly tolerate a NULL-on-no-match result, or (b) extend the leading comment to spell out why an `INNER` join with `IsSubqueryExpression` is always guaranteed to match (e.g. because the subquery uses `DefaultIfEmpty` upstream) and add a `TODO` to revisit if that upstream contract ever relaxes.

— ✓ Fixed in 83d743bb

Tests/Linq/Update/UpdateFromTests.cs
```diff
@@ -484,6 +486,266 @@ public void UpdateTestAssociationSimpleAsUpdatable(
...
+		[Table]
+		sealed class UpdateSubquerySourceTable
+		{
+			[PrimaryKey] public int Id { get; set; }
+			[Column] public string? FirstName { get; set; }
+			[Column] public string? LastName { get; set; }
+		}
+
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/5413")]
+		public void UpdateFromSubqueryRowShouldBeOptimized([IncludeDataSources(TestProvName.AllSQLite, TestProvName.AllOracle, TestProvName.AllPostgreSQL, TestProvName.AllInformix)] string context)
```
MaceWindu, line 619: **SUG001** — The linked issue #5413 is described as a generic regression for providers without native `UPDATE … FROM`. The rewrite in `BasicSqlOptimizer.GetAlternativeUpdate` is shared by DB2, Firebird, Oracle 11/12, SQLite, SqlCe, SapHana, Sybase, Ydb, Informix (plus the PostgreSQL/SQLite fork via `GetAlternativeUpdatePostgreSqlite`). This test only exercises SQLite, Oracle, PostgreSQL, and Informix — missing DB2, Firebird, SqlCe, SapHana, Sybase, Ydb. Given how much of the rewrite logic lives in provider-dialect-sensitive code (row-constructor support branches, subquery validity checks), widening the provider filter catches dialect regressions earlier.

Fix: Switch to `[DataSources(...)]` excluding only the providers known not to support this shape (ClickHouse for correlated update, plus any provider with a documented `ThrowsForProvider` condition), mirroring the companion `UpdateFromSubqueryShouldBeOptimized` test below which uses `[DataSources(TestProvName.AllSqlCe)]`.

Source/LinqToDB/Internal/SqlProvider/BasicSqlOptimizer.cs
```diff
@@ -1080,222 +1060,418 @@ protected static TElement RemapCloned<TElement>(
...
-				if (updateStatement.Output != null)
+			if (dependentArguments.Count > 0)
+			{
+				// generate columns
+				subquery.Select.Columns.Clear();
+				for (var i = 0; i < dependentArguments.Count; i++)
 				{
-					newUpdateStatement.Output = RemapCloned(updateStatement.Output, replaceTree, null);
+					var item = dependentArguments[i];
+
```
MaceWindu: **NIT001** — Blank line contains trailing tabs (indentation preserved on an empty line). Similar trailing-whitespace-only lines exist in `SelectQueryOptimizerVisitor.cs` after `var saveIsExpression = _isExpression;` in `VisitSqlOrderByClause`, `VisitSqlGroupByClause`, `VisitSqlColumnExpression`, and `VisitSqlSearchCondition`.

Fix: Strip trailing whitespace on these empty lines.

— ✓ Fixed in 83d743bb

MaceWindu (changes requested):
## Review notes

- [x] API baselines refreshed — `Source/LinqToDB/CompatibilitySuppressions.xml` was regenerated in this PR to cover the two `LinqToDB.Internal.*` changes (`BasicSqlOptimizer.RemoveUpdateTableIfPossible` 3-arg overload removed; `QueryHelper.SimplifyColumnExpression` removed; `QueryHelper.ExtractSqlSource` added).

## Baselines

413 SQL baselines across 57 providers (246 added, 167 modified), no metrics changes. Cross-provider grouping moves coherently and matches the change intent:

- **New regression tests for #5413 and the new `ProcessUpdateItemsWithRows` / `ProcessUpdateItemsWithoutRows` paths** — `UpdateFromSubqueryShouldBeOptimized`, `UpdateParentTableFromChild`, `UpdateFromWithDuplicateSubqueryColumn_{First,Single}OrDefault`, `UpdateFromSubqueryRow{ShouldBeOptimized,MixedIndependentAndDependent}`. Providers with native `UPDATE … FROM` (PostgreSQL, SQL Server, SQLite MS / Classic MPM / MPU, MariaDB, MySQL 8) emit the native form; providers without (Oracle, DB2, Firebird, Sybase, SqlCe, SapHana, Access, ClickHouse, MySQL 5.7, Informix) emit correlated SET subqueries, or the comma-join multi-table UPDATE idiom on MariaDB / MySQL 5.7 / Access.
- **Alias correction inside correlated SET subqueries** — `UpdateTestWhere[Old]`, `UpdateTestJoin{,Take,SkipTake}`, `Update10[Old]` rewrite the buggy `"Target"."valN"` self-reference to the source-side alias (`c_2` etc.) on DB2, Firebird 2.5–5, Oracle 11–23, SapHana, Sybase. This is exactly the class of bug #5413 reported.
- **Row-SET → scalar SET split** when a single column needs a JOIN (`TestUpdateFromJoin[Old]`, `Update{9,10}[Old]`) on DB2, Firebird 2.5–5, Oracle 11–23, SapHana, Sybase.
- **EXISTS-of-tabless-subquery simplification** drops trailing `EXISTS(SELECT * FROM DUAL/rdb$database)` wrappers on `Test_Feature_Intersect` for Access, Firebird, MySQL 5.7, MySqlConnector 5.7, SqlCe, Sybase.
- **RemoveUpdateTableIfPossible + `allowLeftJoin` + nested SELECT inlining** on `UpdateSetSelect` (DB2, Firebird 2.5–5, Oracle 11–23, SapHana, Sybase), `StringJoinAssociationSubqueryUpdate{1,2}` (Firebird.4/5, SapHana), `Issue4253Test` (Firebird 2.5–4), `UpdateRowSelect` and `UpdateValuesWithConversionWithARowFromOtherTable` (Oracle 11–23).
- **Regression removed on the exact shape #5413 reported** — `Issue5340Tests.UpdateWithSubquery(Oracle.11.Managed)` and `(Informix.DB2)` drop the redundant outer `WHERE EXISTS(...)` wrapper and the leaked `ROW_NUMBER()` projection.

No structural anomalies detected: no missing WHERE, no lost predicates, no duplicate target-table references, no cross-scope alias leaks. `Sybase.Managed` missing from `UpdateFromWithDuplicateSubqueryColumn_FirstOrDefault` is an expected provider-capability skip (no LIMIT/FIRST in SET context), not a regression.

MaceWindu (commented):
_Agentic verification pass by `/verify-review`. Automated diff & baseline cross-check against the commits pushed since the prior review; human review still required._

## Verification update — 2026-04-20, against HEAD `83d743bb`

### Fixed
- [x] MIN001 — `QueryHelper.ExtractSqlSource` XML doc tag nesting
- [x] MIN002 — tabless-EXISTS guard missing HAVING/TAKE/SKIP
- [x] MIN003 — tabless single-column inline missing HAVING/modifier guards
- [x] MIN004 — `ProcessUpdateItemsWithRows` unhandled row-setter fallback (now throws `LinqToDBException`)
- [x] MIN005 — `allowLeftJoin` dead parameter on protected API (3-arg overload restored)
- [x] MIN006 — `INNER`-join subquery-form rewrite comment/TODO
- [x] NIT001 — trailing whitespace on empty lines

### Still actual
- [ ] SUG001 — `UpdateFromSubqueryRowShouldBeOptimized` provider coverage still limited to `[IncludeDataSources(AllSQLite, AllOracle, AllPostgreSQL, AllInformix)]`. Thread was marked resolved by the author; treating as a "won't-expand" decision. (original: review #4138820911)

### Partially / incorrectly fixed
_(none)_

### New findings
_(none)_

## Baselines

No material change vs. the prior review — same 413 SQL baselines (246 added, 167 modified), no metrics. Delta still aligned with the rewrite scope (alias correction inside correlated SET subqueries on DB2/Firebird/Oracle/SapHana/Sybase, row-SET → scalar split, tabless-`EXISTS` simplification, redundant outer `WHERE EXISTS` removal on `Issue5340Tests`). No cross-provider anomalies. See [linq2db.baselines PR #1807](https://github.com/linq2db/linq2db.baselines/pull/1807) · [compare](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5414).

MaceWindu (approved):

PR #5423, opened 2026-03-13
OrderBy removal for providers that do not support it in `UNION ALL`

MaceWindu:


Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
```diff
@@ -163,7 +169,7 @@ bool CorrectOrderByForSelectQuery(SelectQuery selectQuery, SelectQuery? parentSe
 				}
 			}
 
-			if (selectQuery.OrderBy.Items.Count > 1)
+			if (selectQuery.HasOrderBy)
```
viceroypenguin, line 172: Should this be an `else` of the `if()` on line 161?

Issue #5424, opened 2026-03-15
Make it easier/possible to write Clean Paramaterized `SelectQuery`s/`AsQueryable`s.


to11mtm:
### Description

With Linq2Db v6, I notice it is possible to write Parameterized `SelectQuery` entries and even join them with `.UnionAll`/`.Concat`, however when this is done they generate very long (by line count) sql.

#### `.SelctQuery()` + `.Concat()`
```sql
-- Param def elided here

SELECT
	@Id,
	@Data
UNION ALL
SELECT
	@Id_1,
	@Data_1


```

`.AsQueryable` gives us cleaner output, however does not parameterize:

```sql
SELECT
	[t1].[Id],
	[t1].[Data]
FROM
	(
		SELECT NULL [Id], NULL [Data] WHERE 1 = 0
		UNION ALL
		VALUES
			(0,'Data 0'), (1,'Data 1')
		) [t1]
```

### New API

I'm not sure if we need a new API (i.e. way to have `.AsQueryable` parameterized one way or another, TBH would be ideal) or if it would for now be easiest/ok to just handle the DBs that can safely do a `SELECT FROM VALUES` clause on the base end.

### Note:

If there's a way to do what I'm trying to do, Happy to update docs as penance.

to11mtm:
Example Xunit code to provide output for scenarios:

```cs
using System.Diagnostics;
using System.Linq.Expressions;
using LinqToDB;
using LinqToDB.Async;
using LinqToDB.Data;
using LinqToDB.Mapping;
using LinqToDB.Tools;
using Microsoft.Data.Sqlite;
using Xunit.Abstractions;

namespace Linq2Db.Extensions.Tests;

/// <summary>
/// You can probably ignore this, but It was something I was trying to do to see if it would help.
/// </summary>
public static class Example
{
    public static IQueryable<TTable> BuildParameterizedQueryable<TTable>(this DataConnection dataConnection,
        IEnumerable<TTable> table) where TTable : class
    {
        IQueryable<TTable>? query = null;
        foreach (var item in table)
        {
            var props = item.GetType().GetProperties();
            var exprs = new List<MemberAssignment>();
            var newExpr = Expression.New(typeof(TTable));

            var itemConstant = Expression.Constant(item, typeof(TTable));

            foreach (var propertyInfo in props)
            {
                var propertyAccess = Expression.PropertyOrField(itemConstant, propertyInfo.Name);
                var memberInit = Expression.Bind(propertyInfo,
[...]

PR #5431, opened 2026-03-18
Possible Implementation for #5424

to11mtm:
I'm still testing some of this locally (And need to port my tests to NUnit for consideration) however this seems to be able to be able to handle the cases I'm worried about in #5424.

Will admit I need to check against other DBs than SQLite still, which is why I am not saying this is a full solution _yet_.

Disclaimer: Yes I used some AI tools here, yes I tried to make sure it was clean enough to put out.

Am still trying to set up local CI (i.e. want to test more than just SQLite) but at least hope this will help on some level.


sdanyliv:
Superseded by https://github.com/linq2db/linq2db/pull/5495

Issue #5299, opened 2026-01-08
Review concat arguments casting


MaceWindu:
This is not and issue, just inconsistent behavior, discovered in https://github.com/linq2db/linq2db/pull/5291

We need:
- review type conversion generation for string concat operators/functions (`+`, `||` or `concat(...)`) to generate only needed conversions
- update `TestCase2` test asserts to ensure there is no `WHERE` clause

Issue #5441, opened 2026-03-23
Query with ntext workin in V5.x does not work in V6.2


OstryEU:
### Description

Query combining nvarchar and ntext columns works with linq2db 5.4.1, but with linq2db 6.2.1 it generates SQL Error:
The data types nvarchar and ntext are incompatible in the add operator.

### Reproduction Steps

My query with ntext type column [feature].[Caption]. 
var query = from featureData in DBConnection.FeatureData
                   join feature in DBConnection.Features on new { featureData.FeatureId, featureData.ElementId } equals new { FeatureId = feature.Id, feature.ElementId }
                   join element in DBConnection.Elements on feature.ElementId equals element.Id
                   select new { Id = featureData.DataId, Path = $"Element '{element.Caption}', Merkmal '{feature.Caption}'" };

linq2db 6.2.1 has generated SQL script:
SELECT
	[featureData].[DataId],
	(N'Element ''' + Coalesce(CAST([element_1].[Caption] AS NVarChar(4000)), N'') + N''', Merkmal ''' + Coalesce([feature].[Caption], N'')) + N''''
FROM
	[TFeatureData] [featureData]
		INNER JOIN [TFeature] [feature] ON [featureData].[FeatureId] = [feature].[Id] AND [featureData].[ElementId] = [feature].[ElementId]
		INNER JOIN [TElement] [element_1] ON [feature].[ElementId] = [element_1].[Id]

The CAST operartor included is not needed, the column [element_1].[Caption] type is a nvarchar. The second column [feature].[Caption] is ntext type and needs CAST. If I move the CAST operator to the second column the query works.

### Regression?

linq2db 5.4.1 has generated SQL script which worked:
--  SqlServer.2008
SELECT
	[featureData].[DataId],
	[element_1].[Caption],
	[feature].[Caption]
FROM
	[TFeatureData] [featureData]
		INNER JOIN [TFeature] [feature] ON [featureData].[FeatureId] = [feature].[Id] AND [featureData].[ElementId] = [feature].[ElementId]
		INNER JOIN [TElement] [element_1] ON [feature].[ElementId] = [element_1].[Id]

### Linq To DB version

6.2.1

### .NET version

10.0

### Database

SQL Server 2008 R2, SQL Server 2019

### Database provider

Microsoft.Data.SqlClient 6.1.4

### Operating system

Windows

### Configuration

Visual Studio 2022

MaceWindu:
[...]

PR #5442, opened 2026-03-23
Fix string concat operator support

MaceWindu:
On pause. Will be handled in #5299

### DIFF
--- a/Tests/Linq/Linq/StringJoinTests.cs
+++ b/Tests/Linq/Linq/StringJoinTests.cs
@@ -486,5 +486,65 @@ namespace Tests.Linq
 					NVarcharValue = t.Result
 				});
 		}
+
+		sealed class ConcatTable
+		{
+			public int Id { get; set; }
+			[Column(DataType = DataType.VarChar, Length = 10, CanBeNull = true)]
+			public string? VarCharTextNullable { get; set; }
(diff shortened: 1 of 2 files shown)


MaceWindu:
included to #5504

Tests/Linq/Linq/StringJoinTests.cs
```diff
@@ -486,5 +486,65 @@ from t in table
...
+			[Column(DataType = DataType.VarChar, Length = 10, CanBeNull = false)]
+			public string VarCharText { get; set; } = default!;
+			[Column(DataType = DataType.NVarChar, Length = 10, CanBeNull = true)]
+			public string? NVarCharTextNullable { get; set; }
+			[Column(DataType = DataType.NVarChar, Length = 10, CanBeNull = false)]
+			public string NVarCharText { get; set; } = default!;
+		}
+
+		[Test]
+		public void TestStringConcatenation([DataSources] string context)
```
sdanyliv, line 504: What this test is testing? How mapper works? It can be evaluated on the client side. Why AssertQuery is not acceptable here?

MaceWindu: This is followup for #5441, where I discovered that interpolation and concatenation work differently and postoned till unification. I will check if it is client-side later and fix. AssertQuery will not work with sybase...

Issue #1774, opened 2019-06-28
ArgumentOutOfRangeException after using UNION with new SELECT


millworm:
I wanted to make a new SELECT from the result of the UNION, but I got an error. A small test that reproduces it is attached below.

```
Message: System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index

Stack trace:
at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at LinqToDB.SqlQuery.SelectQueryOptimizer.<>c__DisplayClass12_0.<OptimizeUnions>b__2(IQueryElement e)
   at LinqToDB.SqlQuery.QueryVisitor.Visit2(IQueryElement element)
   at LinqToDB.SqlQuery.QueryVisitor.Visit(IQueryElement element, Action`1 action)
   at LinqToDB.SqlQuery.SelectQueryOptimizer.OptimizeUnions()
   at LinqToDB.SqlQuery.SelectQueryOptimizer.FinalizeAndValidate(Boolean isApplySupported, Boolean optimizeColumns)
   at LinqToDB.SqlProvider.BasicSqlOptimizer.<>c__DisplayClass4_0.<Finalize>b__0(SelectQuery selectQuery)
   at LinqToDB.SqlQuery.SqlStatementWithQueryBase.WalkQueries(Func`2 func)
   at LinqToDB.SqlProvider.BasicSqlOptimizer.Finalize(SqlStatement statement)
```

### Steps to reproduce
```
[Test]
public void AfterUnionError([DataSources] string context)
{
	using (var db = GetDataContext(context))
	{
		var q1 = from p in db.Person where p.ID == 1 select new { p.ID};
		var q2 = from p in db.Person where p.ID != 1 select new { p.ID };
		var q = q1.Concat(q2);
		var f = q.Select(t => new {t.ID, rn = Sql.Ext.DenseRank().Over().OrderBy(t.ID).ToValue()}).ToList();				
	}
}
```

### Environment details
linq2db version: 2.7.1.0
.NET Framework: 4.7.2

MaceWindu:
As temporary workaround you can put `t.ID` column to second position in `Select()`

millworm:
Yeap. That's exactly what I did

PR #5443, opened 2026-03-24
Oracle: use NVL for 2 arguments COALESCE

jods4:
Main change in this PR is using proprietary `NVL` in Oracle when `COALESCE` only has 2 arguments.
It's more concise but also it does not fail because of inconsistent charset, which I encountered as a regression when upgrading to v6.

For 3+ arguments I stick to `COALESCE` because it's just cleaner than `NVL(a, NVL(b, NVL(c)))` especially when involved expressions are large. If linq2db does not detect the charset inconsistency then unfortunately it means user has to handle it.

I took the opportunity to refactor the code around charset mismatch detection and handling, with no intended behavior change.

### DIFF
--- a/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
+++ b/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
@@ -203,5 +203,38 @@ namespace LinqToDB.Internal.Common
 				}
 			}
 		}
+	
+		/// <summary>
+		/// Applies <paramref name="map"/> to each element in a readonly <paramref name="source"/>.
+		/// If all elements are the same reference, the source collection is returned unchanged.
+		/// If at least one element changed, a new collection is built and returned.
+		/// </summary>
+		internal static IReadOnlyList<T> MapList<T>(this IReadOnlyList<T> source, Func<T, T> map)
+		{
+			T[]? result = null;
+
+			for (int i = 0; i < source.Count; i++)
+			{
+				var srcItem = source[i];
+				var newItem = map(srcItem);
+
+				if (!ReferenceEquals(srcItem, newItem))
+				{
+					if (result == null)
+					{
+						result = new T[source.Count];						
+						for (int j = 0; j < i; j++) result[j] = source[j];
+					}
+
+					result[i] = newItem;
+				}
+				else if (result != null)
+				{
+					result[i] = srcItem;
+				}
+			}
+
+			return result ?? source;
+		}
 	}
 }
--- a/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
+++ b/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
@@ -209,7 +209,7 @@ namespace LinqToDB.Internal.Common
 		/// If all elements are the same reference, the source collection is returned unchanged.
 		/// If at least one element changed, a new collection is built and returned.
 		/// </summary>
-		internal static IReadOnlyList<T> MapList<T>(this IReadOnlyList<T> source, Func<T, T> map)
+		internal static IReadOnlyList<T> MapList<T>(this IReadOnlyList<T> source, Func<T, T> map) where T: class
 		{
 			T[]? result = null;
 
--- a/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
+++ b/Source/LinqToDB/Internal/Common/EnumerableHelper.cs
@@ -203,7 +203,6 @@ namespace LinqToDB.Internal.Common
 				}
 			}
 		}
-	
 		/// <summary>
 		/// Applies <paramref name="map"/> to each element in a readonly <paramref name="source"/>.
 		/// If all elements are the same reference, the source collection is returned unchanged.
@@ -222,7 +221,7 @@ namespace LinqToDB.Internal.Common
 				{
 					if (result == null)
 					{
-						result = new T[source.Count];						
+						result = new T[source.Count];
 						for (int j = 0; j < i; j++) result[j] = source[j];
 					}
 
--- a/Source/LinqToDB/Internal/Common/Polyfills.cs
+++ b/Source/LinqToDB/Internal/Common/Polyfills.cs
@@ -7,7 +7,7 @@ namespace System.Linq
 {
 #if NET462
 	internal static class EnumerablePolyfills
-	{		
+	{
 		public static IEnumerable<T> Append<T>(this IEnumerable<T> source, T element)
 			=> source.Concat([element]);
 	}
--- a/Source/LinqToDB/Internal/Common/Polyfills.cs
+++ b/Source/LinqToDB/Internal/Common/Polyfills.cs
@@ -1,7 +1,6 @@
 ﻿// Useful methods that are available in current .net releases
 
 using System.Collections.Generic;
-using System.Collections.Generic;
 
 namespace System.Linq
 {
--- a/Source/LinqToDB/Internal/Common/Polyfills.cs
+++ b/Source/LinqToDB/Internal/Common/Polyfills.cs
@@ -1,7 +1,7 @@
 ﻿// Useful methods that are available in current .net releases
 
 using System.Collections.Generic;
-using System.Linq;
+using System.Collections.Generic;
 
 namespace System.Linq
 {
--- a/Source/LinqToDB/Internal/Common/Polyfills.cs
+++ b/Source/LinqToDB/Internal/Common/Polyfills.cs
@@ -1,0 +1,15 @@
+﻿// Useful methods that are available in current .net releases
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace System.Linq
+{
+#if NET462
+	internal static class EnumerablePolyfills
+	{		
+		public static IEnumerable<T> Append<T>(this IEnumerable<T> source, T element)
+			=> source.Concat([element]);
+	}
+#endif
+}
--- a/Source/LinqToDB/Internal/DataProvider/Oracle/Oracle11SqlOptimizer.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Oracle/Oracle11SqlOptimizer.cs
@@ -175,24 +175,14 @@ namespace LinqToDB.Internal.DataProvider.Oracle
 			{
 				if (e is SelectQuery { HasSetOperators: true } query)
 				{
-					IEnumerable<SelectQuery> queries = [query, .. query.SetOperators.Select(o => o.SelectQuery)];
+					IEnumerable<List<SqlColumn>> setsColumns = [query.Select.Columns, .. query.SetOperators.Select(o => o.SelectQuery.Select.Columns)];
 
 					for (var i = 0; i < query.Select.Columns.Count; i++)
 					{
-						if (OracleSqlExpressionConvertVisitor.NeedsCharTypeCorrection(mappingSchema, queries.Select(q => q.Select.Columns[i].Expression)))
+						if (mappingSchema.HasInconsistentCharset(setsColumns.Select(cols => cols[i].Expression)))
 						{
-							foreach (var qry in queries)
-							{
-								var type = QueryHelper.GetDbDataType(qry.Select.Columns[i].Expression, mappingSchema);
-								if (type.DataType is DataType.Char or DataType.VarChar)
-								{
-									qry.Select.Columns[i].Expression = new SqlCastExpression(
-										qry.Select.Columns[i].Expression,
-										type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-										null,
-										isMandatory: true);
-								}
-							}
+							foreach (var cols in setsColumns)
+								cols[i].Expression = mappingSchema.FixCharset(cols[i].Expression);
 						}
 					}
 				}
--- a/Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
@@ -1,0 +1,46 @@
+﻿using System.Collections.Generic;
+
+using LinqToDB.Internal.SqlQuery;
+using LinqToDB.Mapping;
+
+namespace LinqToDB.Internal.DataProvider.Oracle
+{
+	// Extensions methods/helpers specific to Oracle DataProvider
+	internal static class OracleExtensions
+	{
+		// Check if all expressions have a consistent charset (CHAR vs NCHAR, or VARCHAR vs NVARCHAR).
+		public static bool HasInconsistentCharset(this MappingSchema mappingSchema, IEnumerable<ISqlExpression> expressions)
+		{
+			var hasChar = false;
+			var hasNChar = false;
+
+			foreach (var expr in expressions)
+			{
+				var type = QueryHelper.GetDbDataType(expr, mappingSchema);
+
+				hasChar  |= type.DataType is DataType.Char or DataType.VarChar;
+				hasNChar |= type.DataType is DataType.NChar or DataType.NVarChar;
+
+				if (hasChar & hasNChar)
+					return true;
+			}
+
+			return false;
+		}
+
+		// If expr is of type CHAR (resp. VARCHAR), cast it to NCHAR (resp. NVARCHAR).
+		// This is used in conjunction with `HasInconsistentCharset` to unify charset of incompatible expressions.
+		public static ISqlExpression FixCharset(this MappingSchema mappingSchema, ISqlExpression expr)
+		{
+			var type = QueryHelper.GetDbDataType(expr, mappingSchema);
+			return type.DataType switch
+			{
+				// Note that TO_NCHAR(x) works on CHAR type too, but it always returns NVARCHAR2, not NCHAR.
+				DataType.VarChar => new SqlFunction(type.WithDataType(DataType.NVarChar), "To_NChar", expr),
+				DataType.Char => new SqlCastExpression(expr, type.WithDataType(DataType.NChar), null, isMandatory: true),
+				_ => expr,
+			};
+		}
+
+	}
+}
--- a/Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
+using LinqToDB.Internal.Common;
 using LinqToDB.Internal.Extensions;
 using LinqToDB.Internal.SqlProvider;
 using LinqToDB.Internal.SqlQuery;
@@ -19,7 +20,7 @@ namespace LinqToDB.Internal.DataProvider.Oracle
 
 		#region LIKE
 
-		protected static string[] OracleLikeCharactersToEscape = {"%", "_"};
+		protected static readonly string[] OracleLikeCharactersToEscape = ["%", "_"];
 
 		public override string[] LikeCharactersToEscape => OracleLikeCharactersToEscape;
 
@@ -169,59 +170,40 @@ namespace LinqToDB.Internal.DataProvider.Oracle
 			};
 		}
 
-		protected internal override IQueryElement VisitSqlCoalesceExpression(SqlCoalesceExpression element)
+		public override ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, element.Expressions))
+			// If there are exactly two arguments, we prefer proprietary NVL over COALESCE because:
+			// 1. It's more concise SQL (but NVL doesn't support 3+ arguments).
+			// 2. It reduces the risk of regression from v5.x to v6.0, which replaced NVL with COALESCE in Oracle DataProvider.
+			//    One notable difference is that NVL supports different charsets NVL('a', N'a'),
+			//    whereas COALESCE fails with "ORA-12704: character set mismatch".
+			//    We attempt to unify charsets in COALESCE below but it does not always work, 
+			//    for if model doesn't accurately indicate charset (which still worked with NVL in 5.x).
+			if (element.Expressions is [var first, var second])
 			{
-				for (var i = 0; i < element.Expressions.Length; i++)
-				{
-					var type = QueryHelper.GetDbDataType(element.Expressions[i], MappingSchema);
-
-					if (type.DataType is DataType.Char or DataType.VarChar)
-					{
-						element.Expressions[i] = new SqlCastExpression(
-							element.Expressions[i],
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-					}
-				}
+				if (first is SqlValue { Value: null }) return second;
+				if (second is SqlValue { Value: null }) return first;
+				var type = QueryHelper.GetDbDataType(first, MappingSchema);
+				return new SqlFunction(type, "Nvl", parametersNullability: ParametersNullabilityType.IfAllParametersNullable, element.Expressions);
 			}
 
-			return base.VisitSqlCoalesceExpression(element);
+			if (MappingSchema.HasInconsistentCharset(element.Expressions))
+			{				
+				for (var i = 0; i < element.Expressions.Length; i++)
+					element.Expressions[i] = MappingSchema.FixCharset(element.Expressions[i]);
+			}
+
+			return base.ConvertCoalesce(element);
 		}
 
 		protected override ISqlExpression ConvertSqlCondition(SqlConditionExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, [element.TrueValue, element.FalseValue]))
+			if (MappingSchema.HasInconsistentCharset([element.TrueValue, element.FalseValue]))
 			{
-				var type = QueryHelper.GetDbDataType(element.TrueValue, MappingSchema);
-
-				if (type.DataType is DataType.Char or DataType.VarChar)
-				{
-					var trueValue = new SqlCastExpression(
-						element.TrueValue,
-						type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-						null,
-						isMandatory: true);
-
-					return new SqlConditionExpression(element.Condition, trueValue, element.FalseValue);
-				}
-				else
-				{
-					type = QueryHelper.GetDbDataType(element.FalseValue, MappingSchema);
-
-					if (type.DataType is DataType.Char or DataType.VarChar)
-					{
-						var falseValue = new SqlCastExpression(
-							element.FalseValue,
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-
-						return new SqlConditionExpression(element.Condition, element.TrueValue, falseValue);
-					}
-				}
+				return new SqlConditionExpression(
+					element.Condition, 
+					MappingSchema.FixCharset(element.TrueValue), 
+					MappingSchema.FixCharset(element.FalseValue));
 			}
 
 			return base.ConvertSqlCondition(element);
@@ -233,20 +215,10 @@ namespace LinqToDB.Internal.DataProvider.Oracle
 			{
 				for (var i = 0; i < element.Rows[0].Count; i++)
 				{
-					if (NeedsCharTypeCorrection(MappingSchema, element.Rows.Select(r => r[i])))
+					if (MappingSchema.HasInconsistentCharset(element.Rows.Select(r => r[i])))
 					{
 						foreach (var row in element.Rows)
-						{
-							var type = QueryHelper.GetDbDataType(row[i], MappingSchema);
-							if (type.DataType is DataType.Char or DataType.VarChar)
-							{
-								row[i] = new SqlCastExpression(
-									row[i],
-									type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-									null,
-									isMandatory: true);
-							}
-						}
+							row[i] = MappingSchema.FixCharset(row[i]);
 					}
 				}
 			}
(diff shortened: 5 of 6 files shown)


Source/LinqToDB/Internal/Common/EnumerableHelper.cs
```diff
@@ -203,5 +203,38 @@ private async IAsyncEnumerator<T> GetNewEnumerable()
 				}
 			}
 		}
+	
+		/// <summary>
+		/// Applies <paramref name="map"/> to each element in a readonly <paramref name="source"/>.
+		/// If all elements are the same reference, the source collection is returned unchanged.
+		/// If at least one element changed, a new collection is built and returned.
+		/// </summary>
```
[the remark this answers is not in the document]
jods4: adding where `T: class`

Source/LinqToDB/Internal/DataProvider/Oracle/Oracle11SqlOptimizer.cs
```diff
@@ -175,24 +175,14 @@ protected override SqlStatement FixSetOperationValues(MappingSchema mappingSchem
...
-									qry.Select.Columns[i].Expression = new SqlCastExpression(
-										qry.Select.Columns[i].Expression,
-										type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-										null,
-										isMandatory: true);
-								}
-							}
+							foreach (var cols in setsColumns)
+								cols[i].Expression = mappingSchema.FixCharset(cols[i].Expression);
 						}
```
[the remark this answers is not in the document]
jods4: True but that's unchanged from before refactoring.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Collections.Generic;
+
+using LinqToDB.Internal.SqlQuery;
+using LinqToDB.Mapping;
+
+namespace LinqToDB.Internal.DataProvider.Oracle
+{
+	// Extensions methods/helpers specific to Oracle DataProvider
```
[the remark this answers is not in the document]
jods4: This class could hold more than one.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+	internal static class OracleExtensions
+	{
+		// Check if all expressions have a consistent charset (CHAR vs NCHAR, or VARCHAR vs NVARCHAR).
```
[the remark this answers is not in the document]
jods4: Doesn't matter

Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+			var hasNChar = false;
+
+			foreach (var expr in expressions)
+			{
+				var type = QueryHelper.GetDbDataType(expr, mappingSchema);
+
+				hasChar  |= type.DataType is DataType.Char or DataType.VarChar;
+				hasNChar |= type.DataType is DataType.NChar or DataType.NVarChar;
+
+				if (hasChar & hasNChar)
```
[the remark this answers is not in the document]
jods4: Intended. It works and it's faster (single conditional jump)

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
+		public override ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, element.Expressions))
+			// If there are exactly two arguments, we prefer proprietary NVL over COALESCE because:
+			// 1. It's more concise SQL (but NVL doesn't support 3+ arguments).
+			// 2. It reduces the risk of regression from v5.x to v6.0, which replaced NVL with COALESCE in Oracle DataProvider.
+			//    One notable difference is that NVL supports different charsets NVL('a', N'a'),
+			//    whereas COALESCE fails with "ORA-12704: character set mismatch".
+			//    We attempt to unify charsets in COALESCE below but it does not always work, 
+			//    for if model doesn't accurately indicate charset (which still worked with NVL in 5.x).
```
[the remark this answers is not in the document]
jods4: Fixed, it was supposed to say "for example if".

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
+			if (element.Expressions is [var first, var second])
 			{
-				for (var i = 0; i < element.Expressions.Length; i++)
-				{
-					var type = QueryHelper.GetDbDataType(element.Expressions[i], MappingSchema);
+				if (first is SqlValue { Value: null }) return second;
+				if (second is SqlValue { Value: null }) return first;
+				var type = QueryHelper.GetDbDataType(first, MappingSchema);
+				return new SqlFunction(type, "Nvl", parametersNullability: ParametersNullabilityType.IfAllParametersNullable, element.Expressions);
+			}
```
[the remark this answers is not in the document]
jods4: If we want a test case for this, `select coalesce('a', N'a') from dual` is enough to trigger ORA-12704

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
-							element.Expressions[i],
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-					}
-				}
+			if (MappingSchema.HasInconsistentCharset(element.Expressions))
+			{				
+				for (var i = 0; i < element.Expressions.Length; i++)
+					element.Expressions[i] = MappingSchema.FixCharset(element.Expressions[i]);
```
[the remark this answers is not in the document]
jods4: True but so did the code before the refactoring.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -233,20 +215,10 @@ protected internal override IQueryElement VisitSqlValuesTable(SqlValuesTable ele
...
-							{
-								row[i] = new SqlCastExpression(
-									row[i],
-									type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-									null,
-									isMandatory: true);
-							}
-						}
+							row[i] = MappingSchema.FixCharset(row[i]);
 					}
```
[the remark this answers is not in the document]
jods4: True but so did the code before refactoring

jods4:
@sdanyliv One idea I have not implemented because I'm not sure if we can do it easily:
In `FixCharset`, if the expression is a string `SqlValue`, instead of `TO_NCHAR('some string')` we could optimize to the cleaner literal `N'some string'` but I don't know if that has a simple representation in our SQL AST.
Would it be serialized like that if the DbType was changed to `NVARCHAR`?

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-					}
-				}
+			if (MappingSchema.HasInconsistentCharset(element.Expressions))
+			{				
+				for (var i = 0; i < element.Expressions.Length; i++)
+					element.Expressions[i] = MappingSchema.FixCharset(element.Expressions[i]);
 			}
```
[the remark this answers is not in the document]
jods4: True, but so did the code before the refactoring

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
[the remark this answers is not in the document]
jods4: True, but so did the code before refactoring

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
+			//    for example if model doesn't accurately indicate charset (which still worked with NVL in 5.x).
+			if (element.Expressions is [var first, var second])
 			{
-				for (var i = 0; i < element.Expressions.Length; i++)
-				{
-					var type = QueryHelper.GetDbDataType(element.Expressions[i], MappingSchema);
+				if (first is SqlValue { Value: null }) return second;
+				if (second is SqlValue { Value: null }) return first;
+				var type = QueryHelper.GetDbDataType(first, MappingSchema);
+				return new SqlFunction(type, "Nvl", parametersNullability: ParametersNullabilityType.IfAllParametersNullable, element.Expressions);
```
MaceWindu, line 187: **Minor · MIN001** — Emits `NVL(a, b)` in place of `COALESCE(a, b)` for 2-arg calls, but Oracle documents a real semantic difference. The [COALESCE docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/COALESCE.html) state "Oracle Database uses short-circuit evaluation. The database evaluates each expr value and determines whether it is NULL, rather than evaluating all of the expr values before determining whether any of them is NULL" and show the CASE-equivalent form that guarantees `expr2` is not evaluated when `expr1` is non-null. The [NVL docs](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/NVL.html) make no such guarantee. Behavior differs when the 2nd argument is an expensive subquery/function, or may raise: `NVL(a, 1/0)` errors even when `a IS NOT NULL`; `COALESCE(a, 1/0)` does not. The common cases linq2db emits (two column refs, or a column + literal) are unaffected, but user-authored `Sql.Coalesce(col, expensiveScalarSubquery)` or ternaries flattened by the converter will change behavior. The PR body motivates the switch as a v5-compat charset regression fix but doesn''t call out the short-circuit asymmetry, so future readers of this code won''t see the trade-off acknowledged.

Fix: Extend the comment block above this branch (lines 175-181) with an explicit bullet about NVL not short-circuiting. Keeping NVL is defensible — the charset regression is reported by users in practice; the short-circuit divergence is theoretical — but documenting the decision at the emission site costs nothing and saves a future debugging session when someone hits it. The suggestion adds the note as a trailing comment after the return.

```suggestion
				return new SqlFunction(type, "Nvl", parametersNullability: ParametersNullabilityType.IfAllParametersNullable, element.Expressions);
				// Note: per Oracle docs, COALESCE short-circuits but NVL evaluates both arguments. This matters only when
				// the 2nd argument has side effects or can raise (e.g. 1/0). Queries linq2db generates from LINQ do not rely
				// on short-circuit COALESCE; the charset-consistency gain in the NVL path outweighs the asymmetry here.
```

Source/LinqToDB/Internal/Common/Polyfills.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Useful methods that are available in current .net releases
+
+using System.Collections.Generic;
+using System.Linq;
```
MaceWindu: **Nit · NIT001** — `using System.Linq;` is redundant in a file that declares its only type directly inside `namespace System.Linq`. The namespace is already in scope for the body, so the using adds nothing; worse, on non-net462 TFMs where the class is `#if`''d out, the file compiles to an empty namespace with a stray using.

Fix: Drop the unused using line.

```suggestion
using System.Collections.Generic;
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
-		protected internal override IQueryElement VisitSqlCoalesceExpression(SqlCoalesceExpression element)
+		public override ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, element.Expressions))
+			// If there are exactly two arguments, we prefer proprietary NVL over COALESCE because:
+			// 1. It's more concise SQL (but NVL doesn't support 3+ arguments).
+			// 2. It reduces the risk of regression from v5.x to v6.0, which replaced NVL with COALESCE in Oracle DataProvider.
+			//    One notable difference is that NVL supports different charsets NVL('a', N'a'),
+			//    whereas COALESCE fails with "ORA-12704: character set mismatch".
+			//    We attempt to unify charsets in COALESCE below but it does not always work,
```
MaceWindu: **Nit · NIT002** — Trailing whitespace on a new comment line.

Fix: Strip the trailing space.

```suggestion
			//    We attempt to unify charsets in COALESCE below but it does not always work,
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
-					{
-						element.Expressions[i] = new SqlCastExpression(
-							element.Expressions[i],
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-					}
-				}
+			if (MappingSchema.HasInconsistentCharset(element.Expressions))
+			{
```
MaceWindu: **Nit · NIT003** — Trailing tabs after the opening brace on a new line.

Fix: Strip trailing tabs.

```suggestion
			{
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
-							type.WithDataType(type.DataType is DataType.Char ? DataType.NChar : DataType.NVarChar),
-							null,
-							isMandatory: true);
-
-						return new SqlConditionExpression(element.Condition, element.TrueValue, falseValue);
-					}
-				}
+				return new SqlConditionExpression(
+					element.Condition, 
+					MappingSchema.FixCharset(element.TrueValue),
```
MaceWindu: **Nit · NIT004** — Trailing whitespace on two new argument lines of the `SqlConditionExpression` constructor call.

Fix: Strip trailing spaces on both lines.

```suggestion
					element.Condition,
					MappingSchema.FixCharset(element.TrueValue),
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -256,80 +228,27 @@ protected internal override IQueryElement VisitSqlValuesTable(SqlValuesTable ele
 
 		protected override ISqlExpression ConvertSqlCaseExpression(SqlCaseExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, element.Cases.Select(c => c.ResultExpression).Concat(element.ElseExpression == null ? [] : [element.ElseExpression])))
-			{
-				ISqlExpression? elseExpr = null;
-				List<SqlCaseExpression.CaseItem>? cases = null;
+			var expressions = element.Cases.Select(c => c.ResultExpression);
+			if (element.ElseExpression is {} elseCase)
```
MaceWindu: **Nit · NIT005** — Trailing whitespace on a new `if` header line.

Fix: Strip the trailing space.

```suggestion
			if (element.ElseExpression is {} elseCase)
```

Source/LinqToDB/Internal/Common/EnumerableHelper.cs
```diff
@@ -203,5 +203,38 @@ private async IAsyncEnumerator<T> GetNewEnumerable()
 				}
 			}
 		}
+
```
MaceWindu: **Nit · NIT006** — Trailing tab on an otherwise-blank separator line between methods.

Fix: Remove the tab so the line is truly blank.

```suggestion

```

Source/LinqToDB/Internal/Common/EnumerableHelper.cs
```diff
@@ -203,5 +203,38 @@ private async IAsyncEnumerator<T> GetNewEnumerable()
...
+			for (int i = 0; i < source.Count; i++)
+			{
+				var srcItem = source[i];
+				var newItem = map(srcItem);
+
+				if (!ReferenceEquals(srcItem, newItem))
+				{
+					if (result == null)
+					{
+						result = new T[source.Count];
```
MaceWindu: **Nit · NIT007** — Trailing tabs after the assignment.

Fix: Strip the trailing tabs.

```suggestion
						result = new T[source.Count];
```

Source/LinqToDB/Internal/Common/Polyfills.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace System.Linq
+{
+#if NET462
+	internal static class EnumerablePolyfills
+	{
```
MaceWindu: **Nit · NIT008** — Trailing tabs after the class opening brace on a new line.

Fix: Strip trailing tabs.

```suggestion
	{
```

MaceWindu (changes requested):
> [!IMPORTANT]

## Review notes

- **Scope (confirmed with user):** Oracle 2-arg `COALESCE` → `NVL` for v6 charset-regression dodge + refactor charset-inconsistency detection/fix into `OracleExtensions` helpers with no intended behavior change + small `MapList` / `Append` polyfill utilities.
- **User raised three additional scope items** (Meziantou polyfill duplication, NVL vs COALESCE semantics, and 3+-arg chaining alternative) — each was investigated and resolved as noted in findings or out-of-scope observations below.
- Line-level findings (MIN001, NIT001–NIT008) are posted as inline review comments on their respective lines — see the file threads.

## Minor

- [x] **MIN002** — SQLite `NorthwindTest` view-enumeration reordering in the baselines
  - **Why:** The baselines diff shows 2 SQLite baseline files (`Northwind.SQLite` and `Northwind.SQLite.MS`) for `Tests.SchemaProvider.SchemaProviderTests.NorthwindTest` with views returned in a different order (e.g., `[Products by Category]` moved after several other views, `[Alphabetical list of products]` also shifted). This PR touches only Oracle code paths (`OracleSqlExpressionConvertVisitor`, `Oracle11SqlOptimizer`, `OracleExtensions`) and generic helpers (`EnumerableHelper`, `Polyfills`) — nothing that would change the order of SQLite schema views.
  - **Fix:** Please confirm whether this is pre-existing non-determinism in the SQLite schema-provider test (view ordering depends on insertion order of the system catalog), a side effect of rebasing on master, or a real regression. If it's non-determinism, the baselines should be reverted to master's order (or the test should order views by name). If it's from master, the PR description should acknowledge it since the baselines PR rides along on merge.

## Out-of-scope observations

- **Meziantou polyfill for `Enumerable.Append` does not actually exist upstream** — The additional scope brief suggested the custom `Polyfills.cs` duplicates a Meziantou-provided polyfill. Verified against the upstream `Meziantou.Polyfill` README: only `System.Linq.AsyncEnumerable.Append<TSource>` is listed; the synchronous `System.Linq.Enumerable.Append<T>(IEnumerable<T>, T)` is **not** on the supported list. The PR's tiny file-local polyfill is therefore not redundant with Meziantou, and the chosen approach (a single `#if NET462` extension in `namespace System.Linq`) is a valid alternative. If you'd prefer the Meziantou path long-term, that needs an upstream PR to Meziantou first.
- **In-place element mutation inside convert visitors in `allowModify: false` paths** — Preexisting on master, unchanged by this PR. `ConvertCoalesce` (line 192, 3+-arg charset fallback) mutates `element.Expressions[i]` in place, and `VisitSqlValuesTable` (line 221) mutates `row[i]` in place. These overrides bypass the `VisitMode` check that the base class uses; the remoting path (`SqlOptimizerExtensions.PrepareStatementForRemoting`) creates the visitor with `allowModify: false` and drives Transform mode through it. Pre-refactor code at these exact lines already had the same mutation pattern (`element.Expressions[i] = new SqlCastExpression(...)` / `row[i] = new SqlCastExpression(...)`), so this PR neither introduces nor worsens the concern. Two Copilot threads on the PR point at this; they're pointing at real behavior but it's preexisting. Worth a dedicated follow-up issue — fix is structural (build new arrays + new `SqlCoalesceExpression` / `SqlValuesTable` in Transform mode) and touches visitor invariants out of scope here.
- **3+-arg `COALESCE` stays on per-arg `FixCharset` rather than chained `NVL`** — Intentional, covered by the in-code comment on lines 175-181. `FixCharset` rewrites `VarChar`→`TO_NCHAR(…)` and `Char`→NCHAR cast (matching pre-refactor scope exactly); after unification, `base.ConvertCoalesce` receives N-typed arguments so ORA-12704 doesn't re-surface at 3+ args. Chained NVL was considered and rejected on readability grounds — defensible.
- **`setsColumns` double-enumeration in `Oracle11SqlOptimizer`** — Copilot thread kept open, author replied "unchanged from before refactoring." Verified: same `IEnumerable<T> = [a, ..b]` pattern existed pre-PR, and in C# 12 these lower to a materialized array, so re-enumeration is cheap O(n) traversal. Author's reply is correct.

## Baselines

Delta: [linq2db.baselines PR #1774](https://github.com/linq2db/linq2db.baselines/pull/1774) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5443)

### Oracle: COALESCE → NVL rewrite for 2-argument calls

64 Oracle baseline entries (all `Oracle.11.Managed`) show `Coalesce(a, b)` → `Nvl(a, b)`. Expected.

- **2-arg COALESCE rewritten to NVL** — core behavior change; cross-provider structure identical, only Oracle syntax differs.
  - [Tests.Linq.AggregationTests.SumByAssociationSubquery](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5443/Oracle.11.Managed/Tests/Linq/AggregationTests/Tests.Linq.AggregationTests.SumByAssociationSubquery%28Oracle.11.Managed%29.sql) — modified (1 provider: Oracle.11.Managed) — representative sample; 64 patterns total in `Oracle.11.Managed`

    ```diff
    @@ -2,7 +2,7 @@
     SELECT
     	g_1."Id",
    -	SUM(Coalesce(CAST(a_Values."Value" AS Int), 0))
    +	SUM(Nvl(CAST(a_Values."Value" AS Int), 0))
     FROM
     	"Item" g_1
     		LEFT JOIN "ItemValue" a_Values ON g_1."Id" = a_Values."ItemId" AND a_Values."ValueName" = 'Value1'
    ```

    </details>

### Oracle: charset-fix refactor (CAST → TO_NCHAR)

2 Oracle baselines show `CAST(… AS NVarChar2(…))` → `To_NChar(…)` in UNION/CASE-WHEN expressions. Caused by the refactor. Functionally equivalent but the emitted SQL is different — the PR description says "no intended behavior change" for the refactor, so this is worth acknowledging in the PR body.

- **CAST → TO_NChar charset-fix in UNION/CASE expressions**
  - [Tests.Linq.ValueConversionTests.ConditionUnionTest](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5443/Oracle.11.Managed/Tests/Linq/ValueConversionTests/Tests.Linq.ValueConversionTests.ConditionUnionTest%28Oracle.11.Managed%29.sql) — modified (1 provider: Oracle.11.Managed)

    ```diff
    @@ -2,7 +2,7 @@
     SELECT
     	CASE
    -		WHEN t1."EnumNullable" IS NOT NULL THEN CAST(t1."EnumNullable" AS NVarChar2(50))
    +		WHEN t1."EnumNullable" IS NOT NULL THEN To_NChar(t1."EnumNullable")
     		ELSE t1."Enum"
     	END
    ```

    </details>
  - [Tests.Mapping.MapValueTests.Test2](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5443/Oracle.11.Managed/Tests/Mapping/MapValueTests/Tests.Mapping.MapValueTests.Test2%28Oracle.11.Managed%2CFirst%29.sql) — modified (1 provider: Oracle.11.Managed)

    ```diff
    @@ -2,9 +2,9 @@
     SELECT
     	CASE
    -		WHEN t."EnumValue" = 'F' THEN CAST('First' AS NVarChar2(255))
    -		WHEN t."EnumValue" = 'S' THEN CAST('Second' AS NVarChar2(255))
    -		WHEN t."EnumValue" = 'T' THEN CAST('Third' AS NVarChar2(255))
    +		WHEN t."EnumValue" = 'F' THEN To_NChar('First')
    +		WHEN t."EnumValue" = 'S' THEN To_NChar('Second')
    +		WHEN t."EnumValue" = 'T' THEN To_NChar('Third')
     		ELSE t."EnumValue"
     	END
    ```

    </details>

### SQLite: unexpected schema reordering

2 SQLite baselines for `NorthwindTest` show view enumeration reordering. This PR doesn't touch SQLite or the schema provider — see **MIN002** in the findings above.

- **View enumeration order changed in schema provider test**
  - [Tests.SchemaProvider.SchemaProviderTests.NorthwindTest](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5443/Northwind.SQLite.MS/Tests/SchemaProvider/SchemaProviderTests/Tests.SchemaProvider.SchemaProviderTests.NorthwindTest%28Northwind.SQLite.MS%29.sql) — modified (2 providers: Northwind.SQLite, Northwind.SQLite.MS) — same reordering pattern in both SQLite variants

    ```diff
    @@ -68,7 +68,7 @@
     -- Northwind.SQLite.MS SQLite.MS SQLite

    -SELECT * FROM [Products by Category]
    +SELECT * FROM [Summary of Sales by Quarter]

     -- Northwind.SQLite.MS SQLite.MS SQLite

    @@ -76,18 +76,10 @@ SELECT * FROM [Summary of Sales by Year]

     -- Northwind.SQLite.MS SQLite.MS SQLite

    -SELECT * FROM [Summary of Sales by Quarter]
    -
    --- Northwind.SQLite.MS SQLite.MS SQLite
    -
     SELECT * FROM [Order Subtotals]
     ...
    ```

    </details>

### Cross-provider anomalies

- **Tests.SchemaProvider.SchemaProviderTests.NorthwindTest** — Northwind SQLite providers (2 variants) show view enumeration reordering; PR changes only Oracle provider behavior (COALESCE→NVL and charset-fix refactoring). No schema-provider logic changes are stated in the PR summary, so this reordering is either an unintended side effect or test instability.

MaceWindu (commented):
**Retraction of MIN002** (from [review #4143942201](https://github.com/linq2db/linq2db/pull/5443#pullrequestreview-4143942201)).

MIN002 flagged SQLite `NorthwindTest` view-enumeration reordering in the baselines. This is pre-existing non-determinism in the SQLite schema-provider test (documented as a known flaky baseline in `.claude/docs/testing.md` → **Known flaky baselines**), not caused by this PR. Treat MIN002 as resolved / not actionable.

The original review body is unchanged on purpose — it's public history once submitted.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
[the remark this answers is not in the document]
MaceWindu: @sdanyliv ?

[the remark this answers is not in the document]
MaceWindu: @sdanyliv  and here

MaceWindu:
> @sdanyliv One idea I have not implemented because I'm not sure if we can do it easily:
> In `FixCharset`, if the expression is a string `SqlValue`, instead of `TO_NCHAR('some string')` we could optimize to the cleaner literal `N'some string'` but I don't know if that has a simple representation in our SQL AST.
> Would it be serialized like that if the DbType was changed to `NVARCHAR`?

For N'' literal generation you need to add handling to OracleMappingSchema. I think it make sense to add it if we doesn't generate such literals for oracle properly

Source/LinqToDB/Internal/Common/Polyfills.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Useful methods that are available in current .net releases
+
+using System.Collections.Generic;
+using System.Collections.Generic;
```
MaceWindu:
```suggestion
using System.Collections.Generic;
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+		// If expr is of type CHAR (resp. VARCHAR), cast it to NCHAR (resp. NVARCHAR).
+		// This is used in conjunction with `HasInconsistentCharset` to unify charset of incompatible expressions.
+		public static ISqlExpression FixCharset(this MappingSchema mappingSchema, ISqlExpression expr)
+		{
+			var type = QueryHelper.GetDbDataType(expr, mappingSchema);
+			return type.DataType switch
+			{
+				// Note that TO_NCHAR(x) works on CHAR type too, but it always returns NVARCHAR2, not NCHAR.
+				DataType.VarChar => new SqlFunction(type.WithDataType(DataType.NVarChar), "To_NChar", expr),
+				DataType.Char => new SqlCastExpression(expr, type.WithDataType(DataType.NChar), null, isMandatory: true),
```
MaceWindu, line 40: why we use different approach: function vs cast?

Source/LinqToDB/Internal/DataProvider/Oracle/OracleSqlExpressionConvertVisitor.cs
```diff
@@ -169,59 +170,40 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
...
+		public override ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 		{
-			if (NeedsCharTypeCorrection(MappingSchema, element.Expressions))
+			// If there are exactly two arguments, we prefer proprietary NVL over COALESCE because:
+			// 1. It's more concise SQL (but NVL doesn't support 3+ arguments).
+			// 2. It reduces the risk of regression from v5.x to v6.0, which replaced NVL with COALESCE in Oracle DataProvider.
+			//    One notable difference is that NVL supports different charsets NVL('a', N'a'),
+			//    whereas COALESCE fails with "ORA-12704: character set mismatch".
+			//    We attempt to unify charsets in COALESCE below but it does not always work,
+			//    for example if model doesn't accurately indicate charset (which still worked with NVL in 5.x).
```
MaceWindu, line 181: well, incorrect model is an issue on user's side which could backfire in multiple ways

Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
```diff
@@ -1216,7 +1216,7 @@ public virtual ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 				if (element.Expressions[i] is SqlValue { Value: null })
 				{
 					if (element.Expressions.Length == 2)
-						return element.Expressions[(i + 1) % 2];
+						return element.Expressions[i ^ 1];
```
MaceWindu, line 1219: is it correct? Old code returned expression paired with SqlValue, new one - always return second expression (in which case it should be just `[1]` )

MaceWindu (commented):
What I don't like here is that we have both NVL and COALESCE generated based just on parameter count. Taking into account different behavior for parameter evaluation for them I would prefer if we generate only one function. I would prefer to have COALESCE. Does type hinting implemented here handle your case so we can preserve COALESCE?

Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
jods4: `i` is zero or one here.
`i ^ 1` just flips the 1 bit with a `xor`, so 0 -> 1, 1 -> 0.
It's less operations than `(i + 1) % 2`, which is `add` followed by `and` with a data dependency.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleExtensions.cs
jods4: `to_nchar(x)` is the short and direct way in Oracle to convert a character type to `nvarchar2`.

As the comment notes, it always returns `nvarchar2`, so if we want to convert `char` to `nchar` (no var), then we can't use it (and so we fall back to `cast(x as nchar)`.

I'm trying to pick the most idiomatic way, `to_nchar` is what I'd use when writing queries manually.

jods4:
> For N'' literal generation you need to add handling to OracleMappingSchema. I think it make sense to add it if we doesn't generate such literals for oracle properly

Let's leave this outside of this PR. I'm not sure if such handling exists or not, the key change of this PR is COALESCE -> NVL.

jods4:
> What I don't like here is that we have both NVL and COALESCE generated based just on parameter count. Taking into account different behavior for parameter evaluation for them I would prefer if we generate only one function. I would prefer to have COALESCE. 

I see the point... not sure what I like best.

It is true that `coalesce(null, 3, 1/0)` returns 3, whereas `nvl(3, 1/0)` raises an error.
That said, depending on side-effects in SQL is really edgy. linq2db generated `nvl` until v6 and I don't think anybody opened an issue about it.
But in the absolute, I agree the coalesce behavior is better here.

On the other hand, as I said `nvl` was generated until v6 and the switch to `coalesce` has other side effects, such as being more stringent on type compatibility. This creates regressions for users upgrading (and errors when writing queries with incompatible arguments that could work).
A limitation of this argument is that only 2-argument case is solved, 3+ coalesce looks nicer in query than nested `nvl` and is still "broken" in that regard.

> Does type hinting implemented here handle your case so we can preserve COALESCE?

Columns with charsets information are cast automatically, yes.
More complex expressions are not always correctly hinted by linq2db and a cast must be added manually, which works around the issue.
The SQL ends up looking like `coalesce(col1, to_nchar(col2))` instead of concise `nvl(col1, col2)` but that's minor.
Another work-around is to use `Sql.Nvl(..)` instead of `??`.

So.... I think I'm fine with using `nvl` for 2 arguments but I see your point and could understand if you prefer not to.
Curious what others think about it.

MaceWindu:
I personally prefer to use `COALESCE` always and fix any type inference issues we hit intead, but I'm not insisting on it for now. Oracle SQL generation anyway needs a lot of work to support newer dialects in future.

Let's then get answer to my two questions:
- https://github.com/linq2db/linq2db/pull/5443#discussion_r3119270921
- https://github.com/linq2db/linq2db/pull/5443#discussion_r3119246876
and move on with this PR

jods4:
@MaceWindu sorry... somehow I still don't know how to use github 😆 
I had answered a long time ago but my "review" was "pending". It took me a Google search to figure out that the button to submit it is on another tab (Files changed).
Please check out my answers 😁

Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
MaceWindu: my bad. I see what I want to see. confused `i ^1` with `^1`

MaceWindu (approved):
Let's rerun tests and merge.

MaceWindu:
@jods4 , we discussed it again internally. For now we will merge it but probably at some point it will fall victim of refactoring again :)

jods4:
@MaceWindu honestly, if you think this is gonna go, then let's remove it now. We can keep the refactorings from this PR and drop the 2-argument-to-nvl Oracle translation.

The reason is this: 
v6 is a big release with lots of breaking changes. It's expected, we're doing lots of testing and SQL reviews to ensure everything is correct (one system done, three more to go...)
If we have to add a cast or annotate some models to get impacted queries to work with COALESCE, it's not the end of the world, we have identified the problematic ones and we can make changes.

The last thing I want is another breaking change in 6.5, because we decide to eventually go back from NVL to COALESCE, and not know which queries must be fixed again.

MaceWindu:
Good. Please send pr to revert part about NVL

PR #5451, opened 2026-04-04
Add DuckDB data provider

stdray:
## Summary

Add DuckDB data provider support using [DuckDB.NET](https://github.com/Giorgi/DuckDB.NET) (`DuckDB.NET.Data.Full` 1.5.0) as the ADO.NET driver.

[DuckDB](https://duckdb.org/) is an in-process analytical database with PostgreSQL-compatible SQL syntax. This PR adds a full-featured provider, including LINQ translation, schema introspection, bulk copy, and type mapping.

## Provider details

**Provider name:** `DuckDB`

**New files (Source/LinqToDB):**

| File | Description |
|------|-------------|
| `DataProvider/DuckDB/DuckDBTools.cs` | Public API: provider detection, `CreateDataConnection`, `GetDataProvider` |
| `DataProvider/DuckDB/DuckDBOptions.cs` | Provider options (BulkCopyType) |
| `DataProvider/DuckDB/DuckDBFactory.cs` | `DataProviderFactoryBase` implementation |
| `DataProvider/DuckDB/DuckDBSpecificExtensions.cs` | `.AsDuckDB()` extension |
| `DataProvider/DuckDB/IDuckDBSpecificTable.cs` | Marker interface |
| `DataProvider/DuckDB/IDuckDBSpecificQueryable.cs` | Marker interface |
| `Internal/DataProvider/DuckDB/DuckDBDataProvider.cs` | Core provider: flags, parameter handling, reader expressions, bulk copy |
| `Internal/DataProvider/DuckDB/DuckDBProviderAdapter.cs` | ADO.NET adapter via reflection/TypeMapper, DuckDB Appender integration |
| `Internal/DataProvider/DuckDB/DuckDBMappingSchema.cs` | Type mapping and value-to-SQL literal converters |
| `Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs` | SQL generation: quoting, LIMIT/OFFSET, RETURNING, MERGE, ON CONFLICT, LATERAL, parameter CAST wrapping |
| `Internal/DataProvider/DuckDB/DuckDBSqlOptimizer.cs` | UPDATE/DELETE transformation (PostgreSQL-style) |
| `Internal/DataProvider/DuckDB/DuckDBSqlExpressionConvertVisitor.cs` | Expression conversions: `^`→`xor()`, `+`→`\|\|`, `//` integer division, DateTime±TimeSpan CAST to INTERVAL |
| `Internal/DataProvider/DuckDB/DuckDBSchemaProvider.cs` | Schema introspection via `information_schema` + `duckdb_constraints()` |
| `Internal/DataProvider/DuckDB/DuckDBBulkCopy.cs` | Multi-row INSERT bulk copy + native DuckDB Appender bulk copy |
| `Internal/DataProvider/DuckDB/DuckDBSpecificTable.cs` | Internal specific table implementation |
| `Internal/DataProvider/DuckDB/DuckDBSpecificQueryable.cs` | Internal specific queryable implementation |
| `Internal/DataProvider/DuckDB/Translation/DuckDBMemberTranslator.cs` | LINQ member translation: DateTime, Math, String, StringAggregate (STRING_AGG) |

~~**NuGet package (NuGet/DuckDB):** Removed — DuckDB.NET dropped netstandard2.0 in 1.5.0, T4 tooling requires net462. Users should use CLI scaffold tool instead.~~

**CI/CD (Build/Azure):**

| File | Description |
|------|-------------|
| `net80/duckdb.json` | Test provider config for net8.0 |
| `net90/duckdb.json` | Test provider config for net9.0 |
| `net100/duckdb.json` | Test provider config for net10.0 |
| `pipelines/templates/test-matrix.yml` | DuckDB entry: all OS, net8.0+, filter `[duckdb.all]` |

## Supported features

- Full CRUD (SELECT, INSERT, UPDATE, DELETE)
- LINQ translation with all standard operators
- SQL MERGE (WHEN MATCHED / WHEN NOT MATCHED)
- CTE (Common Table Expressions)
- INSERT ... ON CONFLICT DO UPDATE/NOTHING (InsertOrUpdate)
- LATERAL JOIN (CROSS/OUTER APPLY)
- IS DISTINCT FROM
- Window/analytic functions
- Temporary tables (CREATE IF NOT EXISTS, DROP IF EXISTS)
- Schema introspection (tables, columns, primary keys)
- Multi-row INSERT bulk copy
- Native DuckDB Appender bulk copy (ProviderSpecific) with automatic fallback
- STRING_AGG translation for `string.Join` / `StringAggregate`
- RETURNING for identity columns
- Row constructor comparisons (CompareToSelect, Update, UpdateLiteral)

## DuckDB-specific behavior handled

**Parameter handling:** DuckDB.NET treats all parameters as `STRING_LITERAL` regardless of `DbType`. The provider wraps parameters in `CAST($param AS TYPE)` during SQL generation to provide explicit type information. Numeric, DateTime, and DateTimeOffset values are formatted with `InvariantCulture` to avoid locale issues.

**Integer division:** DuckDB performs float division by default (`5/2 = 2.5`). The provider rewrites `/` to `//` (integer division operator) for integer types.

**TIMESTAMPTZ:** DuckDB normalizes all `DateTimeOffset` values to UTC — the original offset is lost on read. The reader converts `DateTime(Kind=Utc)` back to `DateTimeOffset` with zero offset.

**TIME/INTERVAL duality:** `TimeSpan` maps to `DataType.Time` for column storage but is explicitly CAST to `INTERVAL` in DateTime arithmetic expressions.

**Bulk copy (Appender):** DuckDB.NET provides a native `DuckDBAppender` API for high-performance row insertion. Unlike other providers (SqlServer, PostgreSQL, MySQL), the DuckDB Appender requires values for **all** table columns in table order — it does not support partial column lists. The provider queries `information_schema.columns` to get the full table column list, maps entity columns by name, and falls back to `MultipleRows` INSERT when:
- The table column list cannot be retrieved
- The entity has unmapped columns (e.g. partial entity mapped to a wide table) — because `AppendDefault` does not work with `nextval()` identity defaults and causes native Access Violation
- The appender is not supported by the DuckDB.NET version

When the appender is used, values are appended per-row using typed `AppendValue<T>()` calls with type conversions for `char`→`string`, enums→underlying type, and `DateOnly`→`DateTime`.

**BLOB as Stream:** DuckDB.NET returns BLOB columns as `Stream` (`UnmanagedMemoryStream`). Reader expressions convert to `byte[]`.

**TIME as TimeOnly:** DuckDB.NET returns TIME columns as `TimeOnly` (on .NET 6+). Reader expression converts to `TimeSpan`.

**Microsecond precision:** DuckDB has microsecond precision for TIME and TIMESTAMP types. Sub-microsecond ticks are truncated.

**DECIMAL precision:** Parameter CAST includes explicit precision/scale (`DECIMAL(p,s)`) to avoid default truncation.

**Case insensitivity:** `ColumnNameComparer = StringComparer.OrdinalIgnoreCase` — DuckDB identifiers are case-insensitive.

**No FQN:** DuckDB does not support fully qualified names with server/database/schema prefix.

## Not supported

- MERGE NOT MATCHED BY SOURCE (DuckDB limitation — standard MERGE WHEN MATCHED/NOT MATCHED works)
- Associations in MERGE ON clause (DuckDB limitation with subqueries in MERGE ON)
- Foreign key enforcement (DuckDB does not enforce FKs)
- Stored procedures
- True async operations (DuckDB.NET async methods are synchronous, same as SQLite)
- Native appender bulk copy for partial entity mappings or tables with identity columns (DuckDB Appender requires all table columns in order; `AppendDefault` doesn't work with `nextval()` identity defaults, causing native Access Violation)
- Nested types: LIST, STRUCT, MAP (possible future enhancement)

## Type mapping

| DuckDB | .NET | DataType |
|--------|------|----------|
| BOOLEAN | bool | Boolean |
| TINYINT / UTINYINT | sbyte / byte | SByte / Byte |
| SMALLINT / USMALLINT | short / ushort | Int16 / UInt16 |
| INTEGER / UINTEGER | int / uint | Int32 / UInt32 |
| BIGINT / UBIGINT | long / ulong | Int64 / UInt64 |
| HUGEINT | BigInteger | VarNumeric |
| FLOAT | float | Single |
| DOUBLE | double | Double |
| DECIMAL(p,s) | decimal | Decimal |
| VARCHAR | string | NVarChar |
| BLOB | byte[] | VarBinary |
| DATE | DateTime / DateOnly | Date |
| TIME | TimeSpan / TimeOnly | Time |
| TIMESTAMP | DateTime | DateTime2 |
| TIMESTAMPTZ | DateTimeOffset | DateTimeOffset |
| INTERVAL | TimeSpan | Interval |
| UUID | Guid | Guid |

## Test results

1442 passed, 0 failed, 1 skipped (DuckDB provider, DuckDB.NET 1.5.0, net10.0). Test exclusions:
- MERGE NOT MATCHED BY SOURCE tests (unsupported by DuckDB)
- MERGE association in ON clause tests (DuckDB limitation with subqueries)
- AllTypes nchar/nvarchar column tests (columns not present in DuckDB schema)
- AllTypes BLOB literal test (empty blob encoding)
- Sorted merge results ordering (non-deterministic order, same as DB2/Oracle/PostgreSQL)
- `ConvertExpressionTests.LetTest1` (correlated subquery behavior difference in DuckDB 1.5.0)
- FQN tests (unsupported by DuckDB)
- A few edge cases: `Math.Log` precision, null-char string enum mapping, raw SQL tests with `@p` parameter syntax

## Test plan

- [x] Full DuckDB test suite: 1442 passed, 0 failed, 1 skipped
- [x] Verify no regressions on other providers: SQLite.Classic + SQLite.MS + PostgreSQL.15 + DuckDB = 6267 passed, 0 failed
- [x] Review SQL generation for edge cases (fixed unescaped sequence name in NEXTVAL)
- [x] Verify in-memory and file-based DuckDB connections (file-based tested: 95 CreateTable/TempTable tests pass)


sdanyliv (dismissed):

sdanyliv:
@stdray, check why duckdb is not in list above.

stdray:
@sdanyliv Build\Azure\pipelines\testing.yml updated , but need to verify that the test-duckdb pipeline definition exists in Azure DevOps

stdray:
@sdanyliv I'm not  familiar with Azure DevOps, but I think the new pipeline needs to be added manually here
  https://dev.azure.com/linq2db/linq2db/_build?definitionScope=%5Clinq2db

NuGet/DuckDB/LinqToDB.DuckDB.Tools.ttinclude
MaceWindu, line 1: this file could be removed, it's migration file for old nugets

NuGet/DuckDB/linq2db.DuckDB.csproj
```diff
@@ -0,0 +1,44 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFrameworks>net462</TargetFrameworks>
+		<IsPackable>false</IsPackable><!-- DuckDB.NET supports net8.0+ only, NuGet.csproj tools build targets net462 -->
```
MaceWindu: That's a bit complicated... while 3rd party IDEs like Reader use .NET for T4 scaffold, visual studio run T4 host in using .NET Framework. And we test only VS compatibility for T4. 

I see that provider dropped NS2.0 in latest release (1.5) and probably using old 1.4.4 provider for T4 tooling will not work for databases, created with newer version.

My opinion - we shouldn't add T4 support for this provider and tell users to use cli scaffold tool instead.

MaceWindu (changes requested):
I will perform full review later. Need to check SQL reference on DuckDB.

One thing I would like to see is to type mappings tests. For example see tests for other databases here https://github.com/linq2db/linq2db/tree/master/Tests/Linq/DataProvider/Types  . I would recommend to look at YdbTypeTests.cs  as most recent one

sdanyliv:
@stdray, when you push while tests are running - they have stopped automatically. Try do not push in two hours.

NuGet/DuckDB/linq2db.DuckDB.csproj
stdray: linq2db.DuckDB removed

stdray:
```
Test summary: total: 7974; failed: 0; succeeded: 7643; skipped: 133; duration: 692,0s
Build succeeded with 5 warning(s) in 705,4s
```

looks better

MaceWindu (changes requested):
It looks really good. There are some missing features/issues I've listed below.

I've made some changes, as it is easier for me to address them for such big PR and  enabled additional feature tests.

My changes:
- fixed some spacing issues
- ~removed AllDuckDB test provider property. We can use ProviderName.DuckDB directly as we don't have ivariants~ actually we decided to keep it
- enabled different feature tests (probably more could be enabled, but it is about tests refactoring and out of scope of this PR)
  - join types testing
  - cte and recursive cte testing
  - identity insert tests
  - merge tests
  - window function tests (not all)
  - insertorupdate API tests
  - multiple row bulk copy with ignore flag tests
  - lateral join tests

What I found missing/doesn't work up to now:
- [ ] as I mentioned before: DuckDBTypeTests needed to ensure we don't have type mapping issues on our or provider side
- [ ] generation of RECURSIVE clause for CTE missing (e.g. Issue3359_MultipleSets test)
- [ ] some tests fail due to different sql generated for remote context parameters: with and wthout CAST
- [ ] ConflictActionTests - enabled tests for recently added feature, which could be implemented for DuckDB too, as it supports INSERT OR IGNORE - implementation required
- [ ] to TableOptionsTests test class need to add test for DuckDB options
- [ ] looks like `DeleteByTableName[Async]` and `UpdateByTableName[Async]` tests miss sequence reset call and generate different baselines
- [ ] output tests, e.g. `DeleteWithOutputFromQueryAsyncTest` fail with `Binder Error: Unexpected prepared parameter.`. Looks like parameters in RETURNING clause are not supported. If it is the reason - we can fix it by converting parameters to literals in those places
- [ ] use of RAW sql in schema provider is a legacy thing we plan to remove. Please use mappings like it is already done for ClickHouse provider
- [ ] Also for Schema provider add FK metadata support. Even if it is not enforced, it is used for scaffold to generate associations
- [ ] for float/double to string conversion use roundtrip format (e.g. in DuckDBDataProvider.SetParameter)
- [ ] Merge cannot access Source table with current generation (missing AS token?). Also test table
- [ ] Any reason to not use native bulk copy mode as default?
- [ ] check which tests could be enabled in AnalyticTests
- [ ] Looks like IF NOT EXISTS option not implemented for CreateTable: `CreateIfNotExistsTest`
- [ ] add DuckDB support to linq2db.cli

MaceWindu:
Just in case also some claude review comments that could be useful maybe (I didn't yet reviewed implementation in depth, waiting for final version)
```
Code Review: DuckDB Provider (stdray/duckdb branch)                                                                   
  Overview                                                                                                              
  This branch adds DuckDB as a new data provider to linq2db across 87 files (+2813/-78 lines). It follows the
  established provider pattern well and includes provider implementation, SQL generation, bulk copy, schema provider,
  member translation, CI pipeline config, test database script, and test exclusions/adaptations.

  Architecture & Structure

  Positive: The provider follows the established two-layer pattern correctly:
  - Public API in Source/LinqToDB/DataProvider/DuckDB/ (Factory, Options, Tools, Extensions, Specific interfaces)
  - Internal implementation in Source/LinqToDB/Internal/DataProvider/DuckDB/

  Positive: All expected provider components are present: DataProvider, SqlBuilder, SqlOptimizer,
  ExpressionConvertVisitor, MemberTranslator, MappingSchema, ProviderAdapter, SchemaProvider, BulkCopy.

  ---
  Issues Found

  Critical / High Severity

  1. Column alignment in DataConnection.Configuration.cs:211

  AddProviderDetector(LinqToDB.DataProvider.DuckDB   .DuckDBTools    .ProviderDetector);
  Misaligned — the existing entries use column-aligned formatting. DuckDB has 6 chars but only 3 spaces padding, where
  Ydb (3 chars) has 7 spaces. This should be aligned to match the column pattern:
  AddProviderDetector(LinqToDB.DataProvider.DuckDB    .DuckDBTools    .ProviderDetector);
  (needs one more space after DuckDB)

  2. DuckDBTools.ProviderDetector is a method, not a delegate-returning property

  Other providers use patterns like ClickHouseTools.ProviderDetector.DetectProvider (a detector object) or
  YdbTools.ProviderDetector (a static field). DuckDBTools.ProviderDetector is an internal static method (IDataProvider?
  ProviderDetector(ConnectionOptions options)), which matches the Func<ConnectionOptions, IDataProvider?> signature
  directly. This works but is inconsistent — Ydb also does this, so it's acceptable, but the naming convention differs
  from the more structured ClickHouse approach.

  3. SQL Injection risk in DuckDBBulkCopy.GetTableColumns()

  cmd.CommandText = string.Create(CultureInfo.InvariantCulture,
      $"SELECT column_name FROM information_schema.columns WHERE table_name = '{tableName.Replace("'", "''")}' ...");
  While the basic ' → '' escaping is present, this constructs SQL via string interpolation instead of using parameters.
  Since tableName comes from user-configurable BulkCopyOptions.TableName, this is a potential SQL injection vector.
  Should use parameterized queries instead.

  4. SetParameter converts numeric types to strings

  if (value is decimal dec)
      value = dec.ToString(CultureInfo.InvariantCulture);
  else if (value is float flt)
      value = flt.ToString(System.Globalization.CultureInfo.InvariantCulture);
  else if (value is double dbl)
      value = dbl.ToString(System.Globalization.CultureInfo.InvariantCulture);
  This converts decimal/float/double values to strings before passing to the driver, then relies on BuildParameter() to
  add CAST(... AS DECIMAL/FLOAT/DOUBLE). This double conversion (value→string, then SQL CAST) is fragile and could lose
  precision for edge cases. Combined with SetParameterType also setting DbType, there's a conflict — is the value a
  string or a typed DbParameter?

  Question for the author: Is this a DuckDB.NET driver limitation? If so, it should be documented with a comment. If the
   driver supports typed parameters via DbType, the string conversion should be unnecessary for numerics.

  5. DateTimeToDateTimeOffset always forces UTC

  static DateTimeOffset DateTimeToDateTimeOffset(DbDataReader reader, int index)
  {
      var dt = reader.GetDateTime(index);
      return dt.Kind == DateTimeKind.Utc
          ? new DateTimeOffset(dt, TimeSpan.Zero)
          : new DateTimeOffset(DateTime.SpecifyKind(dt, DateTimeKind.Utc), TimeSpan.Zero);
  }
  The else branch forcefully treats non-UTC DateTime as UTC, which could silently corrupt timezone data. If DuckDB
  guarantees TIMESTAMPTZ always returns UTC Kind, the else branch should be a Debug.Assert or throw. If not, this needs
  more careful handling.

  Medium Severity

  6. GetDecimalCastType default precision is very wide

  return "DECIMAL(38, 18)";
  DECIMAL(38,18) allows only 20 integer digits and 18 fractional. This is an unusual default — most ORMs use
  DECIMAL(38,6) or similar. The 18 decimal places could cause issues with large integer values that need >20 integer
  digits. Should match what DuckDB typically uses or align with the mapping schema.

  7. Missing DateTimeOffset literal support in DuckDBMappingSchema

  The mapping schema handles DateTime and DateOnly literals but has no SetValueToSqlConverter for DateTimeOffset. When
  DuckDB encounters a DateTimeOffset constant in a query (not a parameter), it will fall through to the default
  converter, which may not produce valid DuckDB TIMESTAMPTZ syntax.

  8. Missing TimeSpan literal support in DuckDBMappingSchema

  Similarly, TimeSpan is registered as DataType.Time via AddScalarType but has no SetValueToSqlConverter. DuckDB
  TIME/INTERVAL literals have specific syntax requirements.

  9. DuckDBSqlBuilder.BuildParameter CAST wrapping could break NULLs

  When a parameter is NULL, wrapping it in CAST($p AS INTEGER) produces CAST(NULL AS INTEGER) which is valid SQL. But if
   the driver sends NULL as a string literal empty '', the CAST would fail. Need to verify NULL handling through this
  path.

  10. DuckDBBulkCopy swallows exceptions silently

  catch
  {
      return null;
  }
  In GetTableColumns(), any exception (network error, permission denied, invalid SQL) is silently caught and returns
  null, causing fallback to MultipleRowsCopy. This masks real errors. At minimum, it should log or propagate
  non-expected exceptions.

  11. DuckDBBulkCopy async path blocks synchronously

  protected override async Task<BulkCopyRowsCopied> ProviderSpecificCopyAsync<T>(
      ITable<T> table, DataOptions options, IAsyncEnumerable<T> source, CancellationToken cancellationToken)
  {
      // DuckDB Appender is synchronous; convert IAsyncEnumerable to IEnumerable
      var result = ProviderSpecificCopyImpl(
          connections.Value, table, options,
          EnumerableHelper.AsyncToSyncEnumerable(source.GetAsyncEnumerator(cancellationToken)));
  Blocking on async source enumeration in an async method is problematic. Consider adding a comment about why this is
  acceptable (DuckDB is in-process so blocking is fast), or consider materializing to a list first.

  12. DuckDBSchemaProvider.GetForeignKeys returns empty

  protected override IReadOnlyCollection<ForeignKeyInfo> GetForeignKeys(...) => [];
  The comment says "DuckDB does not enforce foreign key constraints" but DuckDB does parse and store FK definitions. The
   schema provider should still return FK metadata even if DuckDB doesn't enforce them, as the scaffolding tool uses
  this information.

  13. BuildCreateTableFieldType casts to SqlCreateTableStatement unsafely

  var tableName = ((SqlCreateTableStatement)Statement).Table.TableName.Name;
  This cast will throw if BuildCreateTableFieldType is ever called from a non-create-table context. Should use as with
  null check or ensure it's documented that this override is only valid in create-table context.

  Low Severity / Nits

  14. ProviderName.DuckDB alignment

  public const string DuckDB     = "DuckDB";
  Two extra spaces compared to adjacent entries — should match the column alignment of the file.

  15. Expressions.cs alignment

  [Sql.Extension(ProviderName.DuckDB       , "RTRIM({0}, {1})"            , ...)]
  Extra spaces after DuckDB — 7 spaces where Sybase uses 8. Minor but the project values column alignment.

  16. DuckDBSqlExpressionConvertVisitor.IsDateType should handle DateOnly

  static bool IsDateType(Type? type) =>
      type == typeof(DateTime) || type == typeof(DateTimeOffset) ||
      type == typeof(DateTime?) || type == typeof(DateTimeOffset?);
  Missing DateOnly/DateOnly? which is also a date type. If someone does dateOnlyField + timeSpan, the INTERVAL CAST
  won't be applied.

  17. DuckDBSqlBuilder.Convert uses LINQ Skip(1).Any()

  || value.Skip(1).Any(c => !char.IsLetter(c) && !c.IsAsciiDigit() && c is not '_');
  This allocates an enumerator. Other providers likely use a loop or span. For a hot path like identifier quoting,
  consider a loop.

  18. Test exclusion pattern for @p parameter syntax

  Multiple test exclusions with comment // DuckDB uses $p parameter syntax, not @p:
  Assume.That(conn.DataProvider.Name, Is.Not.EqualTo(ProviderName.DuckDB));
  This appears 5 times. Consider adding TestProvName.AllDuckDB to the DataSources(exclude) attribute instead of runtime
  Assume.That, which is less discoverable.

  19. DuckDBOptions default BulkCopyType.MultipleRows

  Most providers default to BulkCopyType.MultipleRows. Since DuckDB has a native Appender API, consider defaulting to
  BulkCopyType.ProviderSpecific to give users the best performance by default.

  20. Missing DuckDB in TestProvName feature groups

  AllDuckDB is defined but not added to relevant feature groups like WithApplyJoin, WithWindowFunctions, etc. in
  TestProvName.cs. DuckDB supports LATERAL joins, CTEs, window functions, and many other features. Without being in
  these groups, DuckDB won't be automatically included in feature-gated tests.

  21. DuckDBSqlBuilder.BuildTruncateTableStatement doesn't reset sequences

  DuckDB's TRUNCATE doesn't automatically reset sequences. When truncateTable.ResetIdentity is true, the builder should
  also emit ALTER SEQUENCE ... RESTART statements (similar to how identity reset is done in TestBase.Identity.cs).

  22. Missing net462 support explanation

  The CI config sets enable_fw_netfx: false and commits mention "net462 incompatible". This should be documented — is it
   because DuckDB.NET.Data.Full doesn't support .NET Framework? If so, the NuGet package reference may need a condition
  in the .csproj.

  23. DuckDBMemberTranslator.DateFunctionsTranslator.TranslateDateTimeDateAdd — Quarter uses multiplication instead of
  native

  case Sql.DateParts.Quarter:
      intervalExpr = factory.Multiply(intervalType, ToInterval(increment, "1 Month"), 3); break;
  DuckDB natively supports INTERVAL '1 Quarter' — no need to multiply months by 3.

  24. DuckDBMappingSchema.TIMESTAMP3_FORMAT — millisecond precision only

  DuckDB TIMESTAMP supports microsecond precision (6 digits), but the format only uses fff (3 digits). This loses
  precision for DateTime values with sub-millisecond precision. Consider using ffffff format.

  ---
  What's Done Well

  1. Comprehensive DuckDB.NET adapter: The DuckDBProviderAdapter properly uses expression-based reflection to avoid hard
   dependencies on DuckDB.NET assemblies, following the established pattern.
  2. Native Appender bulk copy: Supporting the DuckDB Appender API for ProviderSpecific bulk copy is an excellent
  performance addition — this is DuckDB's fastest ingestion path.
  3. Proper parameter handling: The $-prefix parameter syntax is correctly handled throughout, including
  PrintParameterName, Convert(SprocParameterToName), and stripping in SetParameter.
  4. DuckDB-specific SQL generation: Good handling of DuckDB quirks like // for integer division, ILIKE for
  case-insensitive LIKE, LATERAL joins for APPLY, and identity via sequences with NEXTVAL defaults.
  5. Thorough test adaptations: ~40 test files adapted with appropriate exclusions, [ThrowsForProvider] attributes, and
  DuckDB-specific test logic.
  6. Complete CI pipeline: Azure pipeline configs for all three .NET TFMs (8.0, 9.0, 10.0) with proper filter support.
  7. Schema provider: Clean implementation using information_schema and duckdb_constraints(), appropriate for DuckDB's
  PostgreSQL-compatible catalog.

  ---
  Summary

  The implementation is solid and follows existing patterns well. The main concerns are:
  - SQL injection in GetTableColumns() (should use parameters)
  - String conversion of numeric parameters (needs clarification/documentation)
  - Missing literal converters for DateTimeOffset/TimeSpan in mapping schema
  - Timestamp precision loss (milliseconds vs microseconds)
  - Silent exception swallowing in bulk copy

  The rest are alignment nits and minor improvements. Overall this is a well-structured contribution that adds
  comprehensive DuckDB support.
```

stdray:
@MaceWindu @igor-tkachev 

# Review Response: PR #5451

## MaceWindu Review (4059802793) — CHANGES_REQUESTED

His changes (already applied):
- Fixed spacing issues
- Enabled feature tests (joins, CTE, identity insert, merge, window functions, insertorupdate, bulk copy ignore, lateral joins)
- Kept `AllDuckDB` test provider property

### Action Items

| # | Done | Item | Notes |
|---|------|------|-------|
| 1 | [x] | **DuckDBTypeTests** — type mapping tests needed | **Fixed** — created `Tests/Linq/DataProvider/Types/DuckDBTypeTests.cs` with 22 tests covering all DuckDB types. Also fixed BLOB literal encoding, float precision (G9/G17), byte[] scalar type mapping, and Appender TimeSpan→TimeOnly/DateOnly conversions |
| 2 | [x] | **RECURSIVE clause for CTE** missing (e.g. `Issue3359_MultipleSets`) | **Already implemented** — `IsRecursiveCteKeywordRequired => true` in DuckDBSqlBuilder |
| 3 | [x] | **Remote context parameters**: different SQL generated (with and without CAST) | **Not a bug** — CAST wrapping is applied to query parameters only; inline mode has no parameters so no CAST. This is expected behavior, same pattern as other providers. |
| 4 | [x] | **ConflictActionTests** — INSERT OR IGNORE not implemented | **Fixed** — added `GetMultipleRowsSuffix` returning `ON CONFLICT DO NOTHING` for `ConflictAction.Ignore` |
| 5 | [x] | **TableOptionsTests** — add test for DuckDB options | **Already done** — MaceWindu added DuckDB to feature source attributes; tests already pass |
| 6 | [x] | **DeleteByTableName/UpdateByTableName** — missing sequence reset, different baselines | **Fixed** — `BuildTruncateTableStatement` now emits `ALTER SEQUENCE ... RESTART` for identity fields via `GetIdentitySequenceName()` helper |
| 7 | [x] | **Output tests** (e.g. `DeleteWithOutputFromQueryAsyncTest`) — `Binder Error: Unexpected prepared parameter` in RETURNING | **Fixed** — added `InlineOutputParametersVisitor` in `DuckDBSqlOptimizer` that sets `IsQueryParameter=false` for all parameters in OUTPUT clauses. 28 tests fixed. |
| 8 | [ ] | **Schema provider**: remove RAW SQL, use mappings like ClickHouse | Architectural refactor — rewrite `DuckDBSchemaProvider` to use LINQ mappings instead of raw SQL queries. Deferred. |
| 9 | [ ] | **Schema provider**: add FK metadata support | DuckDB `duckdb_constraints()` exposes FK constraint_type but doesn't provide referenced table/columns as separate fields — requires parsing `constraint_text`. Deferred. |
| 10 | [x] | **float/double to string**: use roundtrip format in `SetParameter` | **Fixed** — float/double now use `G9`/`G17` format in `SetParameter` to match MappingSchema literal format and ensure DuckDB parses correctly |
| 11 | [ ] | **Merge**: cannot access Source table (missing AS token?) | **⚠️ Need help** — `BasicSqlBuilder.GetTableAlias()` returns null for `SqlTableLikeSource` (method is not virtual, can't override in DuckDB provider). 14 MergeWithOutput tests temporarily disabled for DuckDB. **I might be missing something obvious here — would appreciate guidance on the correct approach to resolve Source alias in RETURNING clause for MERGE.** |
| 12 | [x] | **Native bulk copy as default** — why not ProviderSpecific by default? | **Fixed** — changed default `BulkCopyType` to `ProviderSpecific` in `DuckDBOptions`. Appender with auto-fallback to MultipleRows. |
| 13 | [x] | **AnalyticTests** — check which tests can be enabled | **Fixed** — enabled DuckDB in 21 analytic tests (66 test cases with LinqService). Added `STDEV→STDDEV` mapping in `AnalyticFunctions.cs`. Not added: TestCorrOracle/TestRegrOracle (Decimal overflow from DOUBLE), TestRowNumberOracleSorting (missing assertion), Oracle-only tests (KEEP, LISTAGG, MEDIAN, etc.) |
| 14 | [x] | **CreateIfNotExistsTest** — IF NOT EXISTS not implemented for CreateTable | **Fixed** — added `HasCreateIfNotExists()` check in `BuildCreateTableCommand` |
| 15 | [x] | **linq2db.cli** — add DuckDB support | **Fixed** — added DuckDB.NET.Data.Full package, DatabaseType.DuckDB enum, provider mapping in scaffold command |

---

## MaceWindu Comment (4189467533) — Claude/Copilot review

### Critical / High

| # | Done | Item | Notes |
|---|------|------|-------|
| C1 | [x] | **Column alignment** in `DataConnection.Configuration.cs:211` | **Already correct** — alignment matches Ydb pattern |
| C2 | [x] | **SQL Injection in `GetTableColumns()`** — string interpolation instead of parameters | **Fixed** — now uses parameterized query with `$table_name` and `$table_schema` |
| C3 | [x] | **SetParameter: numeric→string** — fragile, precision loss | **Fixed** — float/double now use `G9`/`G17` format (matching MappingSchema literals). String conversion is required due to DuckDB.NET STRING_LITERAL limitation (documented) |
| C4 | [x] | **DateTimeToDateTimeOffset: UTC forcing** — `else` branch is dangerous | **Fixed** — expanded comment documenting DuckDB TIMESTAMPTZ UTC normalization and that else branch is defensive |

### Medium

| # | Done | Item | Notes |
|---|------|------|-------|
| C5 | [x] | **GetDecimalCastType default DECIMAL(38,18)** — too wide | **Acceptable** — provides max range for unknown precision. DuckDB default DECIMAL(18,3) is too narrow for general use |
| C6 | [x] | **Missing DateTimeOffset literal** in MappingSchema | **Fixed** — added `BuildDateTimeOffset` with microsecond precision, UTC normalization |
| C7 | [x] | **Missing TimeSpan literal** in MappingSchema | **Fixed** — added `BuildTimeSpan` with TIME/INTERVAL handling based on value range |
| C8 | [x] | **BuildParameter CAST with NULL** — verify behavior | **Not an issue** — DuckDB.NET sends NULL as DbNull, `CAST(NULL AS TYPE)` is valid DuckDB SQL |
| C9 | [x] | **GetTableColumns swallows exceptions** | **Fixed** — added comment explaining intentional fallback to MultipleRows |
| C10 | [x] | **Async path blocks synchronously** in BulkCopy | **Already documented** — comment `// DuckDB Appender is synchronous; convert IAsyncEnumerable to IEnumerable` present |
| C11 | [x] | **SchemaProvider.GetForeignKeys returns empty** | Same as item 9 — DuckDB `duckdb_constraints()` doesn't expose FK details as separate fields. Deferred, documented. |
| C12 | [x] | **BuildCreateTableFieldType unsafe cast** to `SqlCreateTableStatement` | **Fixed** — added comment documenting context, added null-forgiving on `.Table!` |

### Low / Nits

| # | Done | Item | Notes |
|---|------|------|-------|
| C13 | [x] | **ProviderName.DuckDB alignment** — extra spaces | **Already correct** — alignment consistent with surrounding entries |
| C14 | [x] | **Expressions.cs alignment** | **Already correct** — DuckDB entries properly column-aligned |
| C15 | [x] | **IsDateType missing DateOnly** | **Fixed** — added `DateOnly`/`DateOnly?` under `#if NET6_0_OR_GREATER` |
| C16 | [x] | **Convert uses LINQ Skip(1).Any()** — allocation | **Acceptable** — same pattern used by PostgreSQL provider |
| C17 | [x] | **Test exclusions @p** — 5 occurrences of `Assume.That` instead of `DataSources(exclude)` | **Out of scope** — test refactoring, deferred per MaceWindu |
| C18 | [x] | **Default BulkCopyType** → ProviderSpecific | **Fixed** — same as item 12, changed default to `ProviderSpecific` in `DuckDBOptions` |
| C19 | [x] | **TestProvName feature groups** — DuckDB not added | **Already done** — MaceWindu added DuckDB to all relevant feature source attributes |
| C20 | [x] | **TruncateTable doesn't reset sequences** | **Fixed** — added RESTART IDENTITY / CONTINUE IDENTITY (PostgreSQL pattern) |
| C21 | [x] | **net462 incompatibility** — needs documentation | **Already documented** — CI sets `enable_fw_netfx: false`, DuckDB.NET targets net8.0+ |
| C22 | [x] | **Quarter DateAdd** — DuckDB natively supports `INTERVAL '1 Quarter'` | **Not a bug** — DuckDB does NOT support `INTERVAL '1 Quarter'`, multiplication by 3 months is correct |
| C23 | [x] | **TIMESTAMP precision** — `fff` (ms) instead of `ffffff` (us) | **Fixed** — `TIMESTAMP3_FORMAT` → `TIMESTAMP6_FORMAT` with `ffffff` microsecond precision |

---

## Copilot Review (4060312725) — inline comments

| # | Done | Item | Notes |
|---|------|------|-------|
| CP1 | [x] | **Unused `using LinqToDB`** in `SupportsAnalyticFunctionsContextAttribute.cs` | **False positive** — file doesn't have `using LinqToDB;`, only `using System;` and `using System.Linq;` |
| CP2 | [x] | **Unused `using LinqToDB`** in `IdentityInsertMergeDataContextSourceAttribute.cs` | **False positive** — file doesn't have `using LinqToDB;` |
| CP3 | [x] | **Unused `using LinqToDB`** in `AllJoinsSourceAttribute.cs` | **False positive** — file doesn't have `using LinqToDB;` |
| CP4 | [x] | **Expressions.cs TrimRight** — removed `RTrimCharactersBuilderNoTrimCharacters` logic | **Already fixed** — Sybase TrimRight now uses `Sql.Extension` with proper `RTrimCharactersBuilderNoTrimCharacters` builder |
| CP5 | [x] | **Expressions.cs TrimLeft** for Sybase — same issue | **Already fixed** — Sybase TrimLeft now uses `Sql.Extension` with `LTrimCharactersBuilder` |
| CP6 | [x] | **HUGEINT inconsistency** — `GetDataType` returns `VarNumeric`, but `GetSystemType` returns `decimal` | **By design** — HUGEINT→`decimal` (CLR type) + `DataType.VarNumeric` (linq2db type) is consistent and intentional |
| CP7 | [x] | **GetDuckDBCastType doesn't handle BigInteger** for HUGEINT/VarNumeric | **Fixed** — added `BigInteger` → `HUGEINT` in SystemType inference and `DataType.VarNumeric` → `HUGEINT` in DataType switch |

---

## Verdict

### Fixed (test failures resolved: 46 → 0)

| What | How |
|------|-----|
| **RETURNING with parameters** (28 tests) | `InlineOutputParametersVisitor` in `DuckDBSqlOptimizer` — marks all parameters in OUTPUT clauses as inline literals |
| **TestMappingWithIdentity** table not found (4 tests) | Added `[Table("TestMergeIdentity", Configuration = ProviderName.DuckDB)]` mapping |
| **HUGEINT/BigInteger CAST** | `GetDuckDBCastType()` now handles `BigInteger` → `HUGEINT` and `DataType.VarNumeric` → `HUGEINT` |
| **SQL injection in GetTableColumns** | Replaced string interpolation with parameterized query |
| **float/double precision loss** | `G9`/`G17` format in `SetParameter`, `'value'::FLOAT` cast in MappingSchema |
| **Missing DateTimeOffset/TimeSpan literals** | Added `BuildDateTimeOffset` and `BuildTimeSpan` in `DuckDBMappingSchema` |
| **TIMESTAMP precision** | Upgraded from millisecond (`fff`) to microsecond (`ffffff`) |
| **TruncateTable identity reset** | `ALTER SEQUENCE ... RESTART` with `GetIdentitySequenceName()` helper |
| **IsDateType missing DateOnly** | Added `DateOnly`/`DateOnly?` under `#if NET6_0_OR_GREATER` |
| **CreateIfNotExists** | Added `IF NOT EXISTS` in `BuildCreateTableCommand` |
| **ConflictAction.Ignore** | Added `GetMultipleRowsSuffix` → `ON CONFLICT DO NOTHING` in `DuckDBBulkCopy` |
| **Default BulkCopyType** | Changed to `ProviderSpecific` (native Appender with auto-fallback) |
| **RECURSIVE CTE** | Already implemented (`IsRecursiveCteKeywordRequired => true`) |
| **DuckDBTypeTests** (22 tests) | Created `DuckDBTypeTests.cs` covering all DuckDB types: bool, integers, float, double, decimal, varchar, blob, uuid, date, time, timestamp, timestamptz, interval, json, hugeint |
| **BLOB inline literal corruption** | DuckDB `\x` reads exactly 2 hex chars — emit `\xHH` per byte instead of single `\x` prefix |
| **Float inline literal parsed as DOUBLE** | Added explicit `'value'::FLOAT` cast in MappingSchema |
| **byte[] not mapped as BLOB** | Added `AddScalarType(typeof(byte[]), DataType.VarBinary)` |
| **Appender TimeSpan→TIME failure** | Map DuckDB-specific types (DuckDBTimeOnly, DuckDBDateOnly) to System types via implicit operators in `BuildAppendValueDelegates`; early TimeSpan→TimeOnly conversion in `AppendValue` |
| **AnalyticTests** (21 tests, 66 cases) | Enabled DuckDB in 21 analytic tests. Added `STDEV→STDDEV` provider mapping in `AnalyticFunctions.cs` |
| **linq2db.cli scaffold** | Added DuckDB provider to CLI scaffold tool (package ref, enum, provider mapping) |
| **InsertWithOutputWithSchema** (4 tests) | Added `CREATE SCHEMA "TestSchema"` to DuckDB.sql init script |
| **MergeWithOutput*** (14 tests) | Disabled for DuckDB — `GetTableAlias()` returns null for `SqlTableLikeSource` (non-virtual, needs shared code change) |
| **Issue3359_MultipleSets** (1 test) | Excluded DuckDB — multiple recursive CTE references in different UNION ALL branches not supported |

### Remaining failures (0 tests)

All test failures resolved. 14 MergeWithOutput tests disabled with comment explaining the shared code blocker.

### Not addressed (deferred / out of scope)

| # | Item | Reason |
|---|------|--------|
| 8 | Schema provider: remove RAW SQL | Architectural refactor (rewrite to LINQ mappings like ClickHouse) |
| 9/C11 | Schema provider: FK metadata | `duckdb_constraints()` doesn't expose referenced table/columns as separate fields — needs `constraint_text` parsing |

---

## Final test run (net10.0)

```
Test Run Successful.
Total tests: 7809
     Passed: 7690
     Failed: 0
    Skipped: 119
 Total time: 10.36 Minutes
```

Source/LinqToDB/DataProvider/DuckDB/IDuckDBSpecificQueryable.cs
```diff
@@ -0,0 +1,8 @@
+using System.Linq;
+
+namespace LinqToDB.DataProvider.DuckDB
+{
+	public interface IDuckDBSpecificQueryable<out TSource> : IQueryable<TSource>
+	{
+	}
```
viceroypenguin:
```suggestion
	public interface IDuckDBSpecificQueryable<out TSource> : IQueryable<TSource>;
```

Source/LinqToDB/DataProvider/DuckDB/IDuckDBSpecificTable.cs
```diff
@@ -0,0 +1,7 @@
+namespace LinqToDB.DataProvider.DuckDB
+{
+	public interface IDuckDBSpecificTable<out TSource> : ITable<TSource>
+		where TSource : notnull
+	{
+	}
```
viceroypenguin:
```suggestion
	public interface IDuckDBSpecificTable<out TSource> : ITable<TSource>
		where TSource : notnull;
```

Source/LinqToDB/Internal/DataProvider/DuckDB/Translation/DuckDBMemberTranslator.cs
```diff
@@ -0,0 +1,240 @@
...
+using LinqToDB.Linq.Translation;
+using LinqToDB.SqlQuery;
+
+namespace LinqToDB.Internal.DataProvider.DuckDB.Translation
+{
+	public class DuckDBMemberTranslator : ProviderMemberTranslatorDefault
+	{
+		protected override IMemberTranslator CreateSqlTypesTranslator()    => new SqlTypesTranslation();
+		protected override IMemberTranslator CreateDateMemberTranslator()  => new DateFunctionsTranslator();
+		protected override IMemberTranslator CreateStringMemberTranslator()=> new StringMemberTranslator();
```
viceroypenguin:
```suggestion
		protected override IMemberTranslator CreateSqlTypesTranslator()     => new SqlTypesTranslation();
		protected override IMemberTranslator CreateDateMemberTranslator()   => new DateFunctionsTranslator();
		protected override IMemberTranslator CreateStringMemberTranslator() => new StringMemberTranslator();
```

Source/LinqToDB/Internal/DataProvider/DuckDB/Translation/DuckDBMemberTranslator.cs
```diff
@@ -0,0 +1,240 @@
...
+				}
+
+				ISqlExpression resultExpression = new SqlExpression(intDataType, $"EXTRACT({partStr} FROM {{0}})", Precedence.Primary, dateTimeExpression);
+
+				if (datepart == Sql.DateParts.WeekDay)
+				{
+					resultExpression = factory.Increment(resultExpression);
+				}
+
+				return resultExpression;
```
viceroypenguin:
```suggestion
				var resultExpression = new SqlExpression(intDataType, $"EXTRACT({partStr} FROM {{0}})", Precedence.Primary, dateTimeExpression);

				return datepart switch
                {
                    Sql.DateParts.WeekDay => factory.Increment(resultExpression),
                    _                     => resultExpression,
				};
```

Source/LinqToDB/Internal/DataProvider/DuckDB/Translation/DuckDBMemberTranslator.cs
```diff
@@ -0,0 +1,240 @@
...
+					case Sql.DateParts.Month:        intervalExpr = ToInterval(increment, "1 Month");       break;
+					case Sql.DateParts.Week:         intervalExpr = factory.Multiply(intervalType, ToInterval(increment, "1 Day"), 7); break;
+					case Sql.DateParts.Day:          intervalExpr = ToInterval(increment, "1 Day");         break;
+					case Sql.DateParts.Hour:         intervalExpr = ToInterval(increment, "1 Hour");        break;
+					case Sql.DateParts.Minute:       intervalExpr = ToInterval(increment, "1 Minute");      break;
+					case Sql.DateParts.Second:       intervalExpr = ToInterval(increment, "1 Second");      break;
+					case Sql.DateParts.Millisecond:  intervalExpr = ToInterval(increment, "1 Millisecond"); break;
+					default:
+						return null;
+				}
```
viceroypenguin:
```suggestion
				var intervalExpr = datepart switch
				{
					Sql.DateParts.Year        => ToInterval(increment, "1 Year"),
					Sql.DateParts.Quarter     => factory.Multiply(intervalType, ToInterval(increment, "1 Month"), 3),
					Sql.DateParts.Month       => ToInterval(increment, "1 Month"),
					Sql.DateParts.Week        => factory.Multiply(intervalType, ToInterval(increment, "1 Day"), 7),
					Sql.DateParts.Day         => ToInterval(increment, "1 Day"),
					Sql.DateParts.Hour        => ToInterval(increment, "1 Hour"),       
					Sql.DateParts.Minute      => ToInterval(increment, "1 Minute"),
					Sql.DateParts.Second      => ToInterval(increment, "1 Second"),
					Sql.DateParts.Millisecond => ToInterval(increment, "1 Millisecond"),
                    _                         => null,
				};
```

Source/LinqToDB/Internal/DataProvider/DuckDB/Translation/DuckDBMemberTranslator.cs
```diff
@@ -0,0 +1,240 @@
...
+								filterCondition = info.FilterCondition;
+
+							var aggregateModifier = info.IsDistinct ? Sql.AggregateModifier.Distinct : Sql.AggregateModifier.None;
+
+							var fn = factory.Function(valueType, "STRING_AGG",
+								[new SqlFunctionArgument(value, modifier : aggregateModifier), new SqlFunctionArgument(separator, suffix : suffix)],
+								[true, true],
+								filter: filterCondition,
+								isAggregate : true,
+								canBeAffectedByOrderBy : true
```
viceroypenguin:
```suggestion
								isAggregate: true,
								canBeAffectedByOrderBy: true
```

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBMappingSchema.cs
```diff
@@ -0,0 +1,167 @@
...
+			}
+		}
+
+		static void ConvertBinaryToSql(StringBuilder stringBuilder, byte[] value)
+		{
+			// DuckDB \x escape reads exactly 2 hex chars per escape.
+			// Must emit \xHH for EACH byte: '\x00\x01\x02'::BLOB
+			stringBuilder.Append('\'');
+			foreach (var b in value)
+				stringBuilder.Append($"\\x{b:X2}");
```
viceroypenguin: Fix `CA1305` on this line.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSchemaProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+				new() { TypeName = "INTERVAL",   DataType = typeof(TimeSpan).       AssemblyQualifiedName! },
+				new() { TypeName = "UUID",       DataType = typeof(Guid).           AssemblyQualifiedName! },
+				new() { TypeName = "JSON",       DataType = typeof(string).         AssemblyQualifiedName! },
+				new() { TypeName = "HUGEINT",    DataType = typeof(decimal).        AssemblyQualifiedName! },
+			};
+		}
+
+		protected override List<TableInfo> GetTables(DataConnection dataConnection, GetSchemaOptions options)
+		{
+			return dataConnection.Query(
```
viceroypenguin: Please review `ClickHouseSchemaProvider` and use similar strong-type initial mapping instead of magic sql strings. (NB: For this and other schema methods in this file.)

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSpecificTable.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace LinqToDB.Internal.DataProvider.DuckDB
+{
+	sealed class DuckDBSpecificTable<TSource>
+		: DatabaseSpecificTable<TSource>,
+			IDuckDBSpecificTable<TSource>
+		where TSource : notnull
+	{
+		public DuckDBSpecificTable(ITable<TSource> table) : base(table) { }
+	}
```
viceroypenguin:
```suggestion
	sealed class DuckDBSpecificTable<TSource>(ITable<TSource> table)
		: DatabaseSpecificTable<TSource>(table),
			IDuckDBSpecificTable<TSource>
		where TSource : notnull;
```

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs
```diff
@@ -0,0 +1,440 @@
...
+				DataType.SByte          => "TINYINT",
+				DataType.Byte           => "UTINYINT",
+				DataType.UInt16         => "USMALLINT",
+				DataType.UInt32         => "UINTEGER",
+				DataType.UInt64         => "UBIGINT",
+				DataType.Single         => "FLOAT",
+				DataType.Double         => "DOUBLE",
+				DataType.Decimal
+				or DataType.Money
+				or DataType.SmallMoney  => GetDecimalCastType(parameter),
```
viceroypenguin: Here and others
```suggestion
				DataType.Decimal     or
				DataType.Money       or
				DataType.SmallMoney  => GetDecimalCastType(parameter),
```

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlExpressionConvertVisitor.cs
```diff
@@ -0,0 +1,129 @@
...
+		{
+		}
+
+		protected override bool SupportsNullInColumn => false;
+
+		public override ISqlPredicate ConvertSearchStringPredicate(SqlPredicate.SearchString predicate)
+		{
+			var searchPredicate = ConvertSearchStringPredicateViaLike(predicate);
+
+			if (false == predicate.CaseSensitive.EvaluateBoolExpression(EvaluationContext) && searchPredicate is SqlPredicate.Like likePredicate)
```
viceroypenguin:
```suggestion
			if (predicate.CaseSensitive.EvaluateBoolExpression(EvaluationContext) == false
                && searchPredicate is SqlPredicate.Like likePredicate)
```

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlExpressionConvertVisitor.cs
```diff
@@ -0,0 +1,129 @@
...
+						)
+					),
+					Sub(p2, 1)
+				),
+
+				// DuckDB lpad/rpad require VARCHAR first argument
+				{
+						Name      : "Lpad" or "Rpad",
+						Parameters: [var str, ..],
+					} when str.SystemType != typeof(string) =>
```
viceroypenguin:
```suggestion
				{
					Name      : "Lpad" or "Rpad",
					Parameters: [var str, ..],
				} when str.SystemType != typeof(string) =>
```

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlOptimizer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+			var visitor = new InlineOutputParametersVisitor();
+			visitor.Visit(output);
+		}
+
+		sealed class InlineOutputParametersVisitor : SqlQueryVisitor
+		{
+			public InlineOutputParametersVisitor() : base(VisitMode.ReadOnly, null)
+			{
+			}
```
viceroypenguin:
```suggestion
		sealed class InlineOutputParametersVisitor() : SqlQueryVisitor(VisitMode.ReadOnly, null)
		{
```

Tests/Base/Attributes/FeatureSources/AllJoinsSourceAttribute.cs
```diff
@@ -13,6 +13,7 @@ public sealed class AllJoinsSourceAttribute : IncludeDataSourcesAttribute
 				TestProvName.AllFirebird,
 				TestProvName.AllPostgreSQL,
 				TestProvName.AllClickHouse,
+				TestProvName.AllDuckDB
```
viceroypenguin:
```suggestion
				TestProvName.AllDuckDB,
```

Tests/Base/TestProvName.cs
```diff
@@ -258,6 +258,10 @@ public static class TestProvName
 		public const string AllInformix   = $"{ProviderName.Informix},{ProviderName.InformixDB2}";
 		public const string AllClickHouse = $"{ProviderName.ClickHouseDriver},{ProviderName.ClickHouseOctonica},{ProviderName.ClickHouseMySql}";
 
+		#region DuckDB
```
viceroypenguin: This secttion should probably go above `NoopProvidier`, not below

viceroypenguin (changes requested):

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSchemaProvider.cs
stdray: Any type mismatch in DuckDB's native Appender API is guaranteed to corrupt the process state — it results in a native Access Violation at the C++ level, not a manageable .NET exception. This is fundamentally different from most other providers where a wrong type just throws a catchable error.

Because of this, I intentionally prefer pulling type information from the database itself (via `information_schema` / SQL queries) rather than maintaining a hardcoded strong-typed mapping that could silently drift out of sync with the actual DuckDB version. A stale mapping means a native crash with no recovery.

Once the DuckDB provider stabilizes and we have confidence in the type surface, this can be revisited and tightened up. For now, using the most up-to-date source of truth (the database's own metadata) is the safer choice, even if it means working with strings.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSpecificTable.cs
stdray: Formatting fixes are better done after tests pass — let's get the functional part green first.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSchemaProvider.cs
viceroypenguin: I think you may have missed the point. Is it likely that the columns here:
```sql
SELECT table_catalog, table_schema, table_name, table_type
FROM information_schema.tables
```
will change types?

Similar to CH provider, create a private DTO for representing this dataset, then filter and transform into `TableInfo` from there. If you want to do `.ToList()` and _then_ transform to avoid the typing issues mentioned here, then that would be viable.

stdray:
# CI Fixes 

## 1. Baseline mismatch — parameter CAST inconsistency (32 CI tests)

**Root cause:** `DuckDBSqlBuilder.BuildParameter()` used `parameter.Type` for CAST type resolution. This metadata differs between Direct and LinqService modes because LinqService serialization uses `GetParameterValue().DbDataType` (which can resolve differently from `parameter.Type`). Result: same parameter gets different CAST types in Direct vs LinqService SQL.

**Fix:** Changed `BuildParameter()` to use `GetParameterValue().DbDataType` + `MappingSchema.GetDataType()` for type resolution — the same path used during LinqService serialization. Now both modes produce identical CAST output. Follows the Firebird provider pattern.

Added VARCHAR fallback: if no type can be determined, CAST to VARCHAR (matches DuckDB.NET STRING_LITERAL default behavior). This handles enum parameters and other unmapped types consistently.

**Files changed:** `Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs`

## 2. Identity sequence leak in DROP TABLE

**Root cause:** `BuildDropTableStatement` only dropped the table, not the associated identity sequence (`{table}_{field}_seq`). When `CreateLocalTable`/`DropTable` cycle ran twice (Direct + LinqService), the sequence persisted and continued incrementing, causing baseline value differences (e.g., `$id = 119` vs `$id = 120`).

**Fix:** `BuildDropTableStatement` now emits `DROP SEQUENCE IF EXISTS` for each identity field after `DROP TABLE`, separated by semicolons.

**Files changed:** `Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs`

## 3. Client-side GUID generation — DistinctInsert baseline mismatch

**Root cause:** `DistinctInsert1`/`DistinctInsert2` tests use `Sql.NewGuid()` which generates random GUIDs — different values between Direct and LinqService runs cause baseline mismatch.

**Fix:** Added `TestProvName.AllDuckDB` to `DisableBaseline` check (same pattern as SapHana).

**Files changed:** `Tests/Linq/Update/InsertTests.cs`

## 4. Issue4702Test — sequence not advanced after KeepIdentity bulk copy (2 CI tests)

**Root cause:** `BulkCopy` with `KeepIdentity = true` inserts explicit Ids (1, 2), but the DuckDB sequence stays at 1. Next `INSERT` via sequence returns 1 → duplicate key. Same issue as PostgreSQL, Firebird, DB2, and other sequence-based providers.

**Fix:** Added `TestProvName.AllDuckDB` to `[ActiveIssue]` list (matching existing PostgreSQL entry).

**Files changed:** `Tests/Linq/Update/InsertTests.cs`

stdray:
###  Refactor DuckDB SchemaProvider to use DTO+LINQ instead of raw SQL** (`925079d6`)

- `GetTables` and `GetColumns` rewritten from raw SQL to LINQ queries via strongly-typed DTOs (`InformationSchemaTable`, `InformationSchemaColumn`), following the ClickHouse provider pattern.
- `GetPrimaryKeys` kept as raw SQL — `duckdb_constraints()` is a table function with `unnest()` that cannot be expressed via DTO mapping.
- `BulkCopy.GetTableColumns` kept as raw SQL — LINQ queries via `DataConnection` add extra open/close cycles that break `CloseAfterUse` mode and connection interceptor counts.

**Tests:** 7688 passed, 0 failed, 119 skipped, 0 baseline mismatches.

stdray:
pls "/azp run test-duckdb"

Tests/Linq/UserTests/Issue681Tests.cs
```diff
@@ -281,6 +281,12 @@ private async Task TestTableFQN<TTable>(
 				throws = true;
 			}
 
+			if ((withServer || withDatabase || withSchema) && context.IsAnyOf(TestProvName.AllDuckDB))
+			{
+				// DuckDB doesn't support fully-qualified table names with server/database/schema
```
MaceWindu: That's not true. For table, view and macro DuckDB supports database(catalog) and schema.

database name missing in documentation for macro, but works. probably documentation bug

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs
```diff
@@ -0,0 +1,484 @@
...
+			SqlObjectName name,
+			ConvertType objectType = ConvertType.NameToQueryTable,
+			bool escape = true,
+			TableOptions tableOptions = TableOptions.NotSet,
+			bool withoutSuffix = false
+		)
+		{
+			var schemaName = tableOptions.HasIsTemporary() ? null : name.Schema;
+
+			if (schemaName != null)
```
MaceWindu, line 254: add database name support. Looks like generation template is
`[dbname.][schema_name.]object_name` (yep, schema name could be skipped)

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs
```diff
@@ -0,0 +1,484 @@
...
+		public override StringBuilder BuildObjectName(
+			StringBuilder sb,
+			SqlObjectName name,
+			ConvertType objectType = ConvertType.NameToQueryTable,
+			bool escape = true,
+			TableOptions tableOptions = TableOptions.NotSet,
+			bool withoutSuffix = false
+		)
+		{
+			var schemaName = tableOptions.HasIsTemporary() ? null : name.Schema;
```
MaceWindu: ensure that database name also not allowed for temp table

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs
```diff
@@ -0,0 +1,484 @@
...
+
+			// DuckDB uses sequences for identity (CREATE SEQUENCE + DEFAULT nextval()).
+			// TRUNCATE does not automatically reset sequences, so emit ALTER SEQUENCE ... RESTART.
+			if (truncateTable.ResetIdentity)
+			{
+				foreach (var field in table.IdentityFields)
+				{
+					var seqName = GetIdentitySequenceName(table.TableName.Name, field.PhysicalName);
+					AppendIndent();
+					StringBuilder.Append("ALTER SEQUENCE ");
```
MaceWindu: it doesn't work: `DuckDBException : Not implemented Error: ALTER SEQUENCE option not supported yet!`

MaceWindu (changes requested):
Missing/incomplete functionality:
- [ ] LINQPad support
- [ ] FQN support (see comments)
- [x] add NewGuid mapping to [gen_random_uuid](https://duckdb.org/docs/current/sql/functions/utility#gen_random_uuid)
- [x] `UseDuckDB` DataOptions extensions missing
- [x] implement test helpers GetSchemaName and GetDatabaseName in TestUtils

stdray:
## DuckDB full test run

`net10.0`, in-memory DuckDB, single worker:

**22 failed / 8325 passed / 7 skipped / 8354 total** (≈26 min)

The 22 failures fall into three buckets, all rooted in DuckDB / DuckDB.NET behaviour rather than linq2db logic.

### 1. Per-statement FK checking — 16 MERGE tests

Tests in `Tests/Linq/Update/MergeTests.Operations.Associations.cs` and `Tests/Linq/Update/MergeTests.Operations.IdentityInsert.cs` whose setup helper `PrepareIdentityData` runs `Patient.Delete → Doctor.Delete → Person.Delete` inside a transaction.

DuckDB checks foreign keys per-statement against committed state and does not see deletes performed earlier in the same transaction. The third `DELETE` fails with `Constraint Error: Violates foreign key constraint because key 'PersonID: 1' is still referenced` because DuckDB still sees the (already-deleted-in-tx) `Doctor`/`Patient` rows referencing `Person`.

Upstream tracker (still open): [duckdb/duckdb#13819](https://github.com/duckdb/duckdb/issues/13819) — *Transactions not working as expected with DELETE/TRUNCATE*.

### 2. Decimal parameter facets not in trace — 1 test

`ExposeSqlDecimalParameter` in `Tests/Linq/Linq/ParameterTests.cs`. The test asserts the diagnostic trace contains the column's `(6, 3)` precision/scale. Because the decimal-binding workaround in `DuckDBSqlBuilder.BuildParameter` sends decimals as invariant-culture strings wrapped in `CAST(... AS DECIMAL(p,s))`, the parameter that reaches the trace is a string and the facets are gone — so the assertion fails. PostgreSQL / Oracle / DB2 / Firebird / Informix / ClickHouse / Sybase / SAP HANA / SqlCe are excluded from this test for analogous reasons.

### 3. `@p` parameter prefix in user-supplied SQL — 5 tests

`TestObject3`, `TestObject4`, `TestObject5`, `TestObject6`, `TestDataParameterMapping1` in `Tests/Linq/Data/DataExtensionsTests.cs`. These tests are not parameterised over `[DataSources]` — they use `new DataConnection()` and hardcode `"SELECT @p"`, so they run only against whatever provider is configured as default.

DuckDB.NET only recognises `$name` / `$1` parameter placeholders. The user-supplied command text is forwarded to DuckDB unchanged, where DuckDB parses `@p` as an identifier reference and fails with `Binder Error: Referenced column "p" was not found because the FROM clause is missing`. linq2db's diagnostic trace shows the parameters were bound correctly (`DECLARE $p ... SET $p = '123'`); only the literal `@p` in the command text is the problem.

MaceWindu:
@stdray will get back to review later next week

MaceWindu:
ok, last bit (linqpad) is working.
image

Tests/Tests.T4/Compat/Stubs.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace System.Net
+{
+	public readonly struct IPNetwork;
+}
+
+namespace System
+{
+	public readonly struct DateOnly;
+	public readonly struct TimeOnly;
```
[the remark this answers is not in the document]
MaceWindu: lie

Tests/Base/Attributes/FeatureSources/SupportsAnalyticFunctionsContextAttribute.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Linq;
 
+using LinqToDB;
+
```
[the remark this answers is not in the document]
MaceWindu: Fixed at HEAD - `using LinqToDB;` removed.

Tests/Base/Attributes/FeatureSources/IdentityInsertMergeDataContextSourceAttribute.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Linq;
 
+using LinqToDB;
+
```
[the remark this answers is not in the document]
MaceWindu: Fixed at HEAD - `using LinqToDB;` removed.

Tests/Base/Attributes/FeatureSources/AllJoinsSourceAttribute.cs
```diff
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Linq;
 
+using LinqToDB;
+
```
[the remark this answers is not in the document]
MaceWindu: Fixed at HEAD - `using LinqToDB;` removed.

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -1066,8 +1066,8 @@ public static void MapMember(string providerName, Type objectType, MemberInfo me
 		[CLSCompliant(false)]
 		[Sql.Extension(ProviderName.Firebird      , "TRIM(TRAILING {1} FROM {0})", ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(TrailingRTrimCharactersBuilder))]
 		[Sql.Extension(ProviderName.ClickHouse    , "trim(TRAILING {1} FROM {0})", ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(RTrimCharactersBuilder))]
-		[Sql.Extension(ProviderName.SqlServer     , "RTRIM({0})"                 , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(RTrimCharactersBuilderNoTrimCharacters))]
-		[Sql.Extension(ProviderName.SqlCe         , "RTRIM({0})"                 , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(RTrimCharactersBuilderNoTrimCharacters))]
+		[Sql.Expression(ProviderName.SqlServer    , "RTRIM({0})"                 , ServerSideOnly = false, PreferServerSide = false)]
+		[Sql.Expression(ProviderName.SqlCe        , "RTRIM({0})"                 , ServerSideOnly = false, PreferServerSide = false)]
```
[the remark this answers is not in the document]
MaceWindu: Inaccurate at HEAD - SqlServer/SqlCe `TrimRight` mappings are unchanged from master and still use `Sql.Extension` with `RTrimCharactersBuilderNoTrimCharacters`. The diff for `Expressions.cs` in this PR shows only two new DuckDB lines added; nothing else changed.

Source/LinqToDB/Linq/Expressions.cs
```diff
@@ -1104,7 +1105,8 @@ public static void MapMember(string providerName, Type objectType, MemberInfo me
 		[Sql.Extension(ProviderName.Access        , "LTRIM({0}, {1})"           , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(LTrimCharactersBuilder))]
 		[Sql.Extension(ProviderName.SqlServer     , "LTRIM({0}, {1})"           , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(LTrimCharactersBuilder))]
 		[Sql.Extension(ProviderName.SqlCe         , "LTRIM({0}, {1})"           , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(LTrimCharactersBuilder))]
-		[Sql.Extension(ProviderName.Sybase        , "LTRIM({0})"                , ServerSideOnly = false, PreferServerSide = false, BuilderType = typeof(LTrimCharactersBuilder))]
+		[Sql.Expression(ProviderName.Sybase       , "LTRIM({0})"                , ServerSideOnly = false, PreferServerSide = false)]
```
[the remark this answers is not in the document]
MaceWindu: Inaccurate at HEAD - Sybase `TrimLeft` mapping is unchanged from master.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSchemaProvider.cs
```diff
@@ -0,0 +1,198 @@
...
+				new() { TypeName = "VARCHAR",    DataType = typeof(string).         AssemblyQualifiedName!, CreateParameters = "length" },
+				new() { TypeName = "BLOB",       DataType = typeof(byte[]).         AssemblyQualifiedName! },
+				new() { TypeName = "DATE",       DataType = typeof(DateTime).       AssemblyQualifiedName! },
+				new() { TypeName = "TIME",       DataType = typeof(TimeSpan).       AssemblyQualifiedName! },
+				new() { TypeName = "TIMESTAMP",  DataType = typeof(DateTime).       AssemblyQualifiedName! },
+				new() { TypeName = "TIMESTAMPTZ",DataType = typeof(DateTimeOffset). AssemblyQualifiedName! },
+				new() { TypeName = "INTERVAL",   DataType = typeof(TimeSpan).       AssemblyQualifiedName! },
+				new() { TypeName = "UUID",       DataType = typeof(Guid).           AssemblyQualifiedName! },
+				new() { TypeName = "JSON",       DataType = typeof(string).         AssemblyQualifiedName! },
+				new() { TypeName = "HUGEINT",    DataType = typeof(decimal).        AssemblyQualifiedName! },
```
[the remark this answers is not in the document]
MaceWindu: Fixed at HEAD - `GetDataTypes` and `GetSystemType` both return `BigInteger`; `GetDataType` returns `DataType.Int128`.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlBuilder.cs
```diff
@@ -0,0 +1,413 @@
...
+				else if (type == typeof(decimal))          return GetDecimalCastType(parameter);
+				else if (type == typeof(bool))             return "BOOLEAN";
+				else if (type == typeof(Guid))             return "UUID";
+				else if (type == typeof(DateTime))         return "TIMESTAMP";
+				else if (type == typeof(DateTimeOffset))   return "TIMESTAMPTZ";
+				else if (type == typeof(TimeSpan))         return "INTERVAL";
+#if NET6_0_OR_GREATER
+				else if (type == typeof(DateOnly))         return "DATE";
+				else if (type == typeof(TimeOnly))         return "TIME";
+#endif
```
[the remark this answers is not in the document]
MaceWindu: Fixed at HEAD - `GetDuckDBCastType` handles `DataType.Int128 -> HUGEINT`, `DataType.UInt128 -> UHUGEINT`, `DataType.VarNumeric -> BIGNUM`.

Tests/Linq/Mapping/ConversionTypeTests.cs
```diff
@@ -10,9 +10,10 @@ namespace Tests.Mapping
 {
 	public class ConversionTypeTests : TestBase
 	{
+		// this is a garbage test. it mess with mapped schema queries
 		static readonly MappingSchema _trimMappingSchema = new MappingSchema("Trim *")
 			.SetConvertExpression<string,string>(s => s.Trim('*'),  conversionType: ConversionType.FromDatabase)
-			.SetConvertExpression<string,string>(s => $"***{s}***", conversionType: ConversionType.ToDatabase)
+			.SetConvertExpression<string,string>(s => s.ToLower() == nameof(TrimTestTable).ToLower() ? s : $"***{s}***", conversionType: ConversionType.ToDatabase)
```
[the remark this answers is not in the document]
MaceWindu: Comparison is between user input `s` and the constant `nameof(TrimTestTable)`, which is pure ASCII - no locale-sensitivity risk in practice.

Source/LinqToDB/Internal/DataProvider/DuckDB/DuckDBSqlExpressionConvertVisitor.cs
```diff
@@ -0,0 +1,108 @@
...
+				),
+
+				// DuckDB lpad/rpad require VARCHAR first argument
+				{
+					Name      : "Lpad" or "Rpad",
+					Parameters: [var str, ..],
+				} when str.SystemType != typeof(string) =>
+					base.ConvertSqlFunction(new SqlFunction(func.Type, func.Name, false, true,
+						new SqlCastExpression(str, new DbDataType(typeof(string), DataType.VarChar), null, false), func.Parameters[1], func.Parameters[2])),
+
```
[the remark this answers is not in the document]
MaceWindu: Fixed in 64cce910f - the branch was unreachable since the only `LPAD` call site with a non-string first argument already wraps its own CAST. Dead code removed.

Source/LinqToDB/ProviderName.cs
```diff
@@ -335,5 +335,9 @@ public static class ProviderName
 		/// YDB provider.
 		/// </summary>
 		public const string Ydb        = "YDB";
+		/// <summary>
+		/// DuckDB data provider.
+		/// </summary>
+		public const string DuckDB     = "DuckDB";
```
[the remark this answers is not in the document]
MaceWindu: `PublicAPI/*.txt` files are updated by maintainers before each release; they intentionally lag source changes within a release cycle.

Directory.Packages.props
```diff
@@ -117,6 +117,7 @@
 		<PackageVersion Include="System.Data.SQLite"                                    Version="2.0.2"                        />
 		<PackageVersion Include="SourceGear.sqlite3"                                    Version="3.50.4.5"                     />
 		<PackageVersion Include="Ydb.Sdk"                                               Version="0.30.0"                       />
+		<PackageVersion Include="DuckDB.NET.Data.Full"                                  Version="1.5.2"                        />
 		<PackageVersion Include="linq2db4iSeries"                                       Version="6.1.0"                        />
```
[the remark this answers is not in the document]
MaceWindu: Confirmed - PR description will be updated to 1.5.2.

sdanyliv (approved):

MaceWindu (approved):

PR #5450, opened 2026-04-01
Introduced two eager loading strategies for fastest data retrieval

sdanyliv:
## Summary

Introduces two new eager loading execution strategies — **KeyedQuery** and **CteUnion** — alongside the existing `Default` strategy. Works with both `LoadWith`/`ThenLoad` and `select new { ... }` projections with inline child sub-queries. Controllable per-query or globally via `LinqOptions`.

### Problem

The existing `Default` eager loading strategy issues one `SELECT DISTINCT` pre-query per child association, projecting the **full parent entity** each time. For parents with many columns or queries with many child associations, this produces excess data transfer and multiple round-trips.

### Solution

Three strategies are now available:

| Strategy | How it works | Best for |
|---|---|---|
| `Default` | `SELECT DISTINCT` full parent + child join per association | Existing behavior, always available |
| `KeyedQuery` | Buffers main query results; extracts parent keys client-side; loads children with `WHERE key IN (...)` | Wide parent entities (many columns) |
| `CteUnion` | Combines multiple same-level child associations into a single `UNION ALL` CTE query | 2+ child collections at the same level |

### Eager loading — two styles

**LoadWith/ThenLoad:**
```csharp
db.Companies
    .LoadWith(c => c.Departments)
    .ThenLoad(d => d.Employees)
    .ToList();
```

**Inline Select projection:**
```csharp
(from c in db.Companies
 orderby c.Id
 select new
 {
     c.Id,
     c.Name,
     Departments = db.Departments
         .Where(d => d.CompanyId == c.Id)
         .OrderBy(d => d.Id)
         .Select(d => new
         {
             d.Id,
             Employees = db.Employees
                 .Where(e => e.DepartmentId == d.Id)
                 .ToList(),
         })
         .ToList(),
 }).ToList();
```

### New API

**Per-query strategy markers** (on the root query):
```csharp
// CteUnion: combine child pre-queries into a single UNION ALL CTE
query.WithUnionLoadStrategy().ToList();

// KeyedQuery: buffer main query, load children with WHERE key IN (...)
query.WithKeyedLoadStrategy().ToList();

// Default (explicit)
query.WithSeparateLoadStrategy().ToList();
```

**Global default** via `LinqOptions`:
```csharp
db.UseLinqOptions(o => o with { DefaultEagerLoadingStrategy = EagerLoadingStrategy.KeyedQuery });
```

[...]

PR #5452, opened 2026-04-05
Initial setup for Claude code

MaceWindu:
- adds CLAUDE.md (cherry-picked from #5450)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ packages/
 .vscode/
 .build/
 **/launchSettings.json
+/.claude/settings.local.json
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,11 +96,11 @@ The main flow for translating LINQ to SQL:
 
 1. **LINQ Expressions** (`Source/LinqToDB/Linq/`) — Entry point. LINQ method calls are captured as expression trees. `Translation/` contains the member translator interfaces (`IMemberTranslator`, `ISqlExpressionTranslator`) that convert .NET method calls to SQL expressions.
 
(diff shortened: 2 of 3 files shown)


igor-tkachev:
Then let us also add `AGENTS.md` and `.github/copilot-instructions.md` right away. Different agents use different files, so adding only `CLAUDE.md` only helps one specific tool.

To avoid duplicating the same information, we can either keep the main content in `AGENTS.md` and have the other files point to it, or keep everything in something like `docs/ai-guidelines.md` and make all three files reference that document.

igor-tkachev:
I think this document is still missing the "do not do this" instructions for agents. Without them, they will start "improving" the architecture, introducing unnecessary abstractions (to the delight of certain unfortunate programmers), changing formatting, and making far-reaching decisions based on a local task.

So I think we should add something like this:

```md id="d0hb1q"
## Agent Priorities

1. Preserve existing architecture, public behavior, and public API
2. Prefer minimal, local changes
3. Do not modify public APIs unless there is a clear and explicit reason
4. Do not introduce new abstractions unless clearly justified
5. If multiple solutions are possible, describe trade-offs instead of choosing arbitrarily
6. Do not reformat, rename, or clean up unrelated code
7. Do not make global architectural changes for local convenience
8. If a change requires making assumptions or introducing arbitrary values, document those assumptions explicitly with a comment or TODO
9. In ambiguous situations, prefer asking questions over making a potentially suboptimal decision
```

sdanyliv:
@MaceWindu, let's commit this. Other agents support should go to other PR.

PR #5463, opened 2026-04-11
Correct DropTable throwExceptionIfNotExists behavior

### LINKED ISSUE #798: Drop operation incorrectly handles throwExceptionIfNotExists parameter false value

xamtam54:
Fixes incorrect handling of throwExceptionIfNotExists behavior in DropTable operations.

Ensures that only "table not found" exceptions are suppressed when throwExceptionIfNotExists is false, instead of swallowing all exceptions.

This aligns behavior with expected semantics and fixes #798.

---

## Maintainer follow-ups

- `fd579a347` — Escape apostrophes in identifiers inside `EXECUTE IMMEDIATE` / `EXECUTE STATEMENT` wrappers (DB2, SAP HANA, Oracle, Firebird). Routes through `BuildValue`. Regression test `DropTable_IfExists_IdentifierWithApostrophe`.
- `a64463fc2` — Oracle: `TRUNCATE TABLE` before `DROP TABLE` for GTTs to avoid ORA-14452 ("temporary table already in use"). Previously masked by the removed client-side catch-all; fixes 5 tests (`Issue4671Test`, `NullableCoalesceJoinTest`, `InsertIntoTempTableWithPrimaryKey`, `CreateTableEnumerableWithNameAndDescriptionTest` + async variant).


viceroypenguin (changes requested):
* Would be nice to see some unit tests
* Please research and add, as much as possible, the other providers currently supported by linq2db

Source/LinqToDB/DataExtensions.cs
```diff
@@ -1356,14 +1365,159 @@ await QueryRunner.DropTable<T>
...
+			return false;
+		}
+
+		/// <summary>
+		/// Detects provider-specific "table not found" exceptions.
+		/// </summary>
+		private static bool IsProviderTableNotFoundException(Exception ex)
+		{
+			var typeName = ex.GetType().FullName ?? string.Empty;
+			var message  = ex.Message ?? string.Empty;
```
viceroypenguin: Check for `null`/whitespace on `typeName` and `message` here; and return false if any are null.

Source/LinqToDB/DataExtensions.cs
```diff
@@ -1356,14 +1365,159 @@ await QueryRunner.DropTable<T>
...
+
+			return false;
+		}
+
+		/// <summary>
+		/// Checks Access-specific "table not found" messages.
+		/// </summary>
+		private static bool IsAccessTableNotExists(string? message)
+		{
+			if (string.IsNullOrWhiteSpace(message))
```
viceroypenguin: This can be elided and moved back into parent method since we are checking for this at start of method.

viceroypenguin (commented):
Looking good! Make these two quick changes and then I'll run the full test suite on them. Thank you!

Source/LinqToDB/DataExtensions.cs
```diff
@@ -1356,14 +1365,152 @@ await QueryRunner.DropTable<T>
...
+			{
+				return message.Contains("60");
+			}
+
+			// OleDb / Access
+			if (typeName.Contains("OleDbException", StringComparison.Ordinal))
+			{
+				if (string.IsNullOrWhiteSpace(message))
+					return false;
+
```
viceroypenguin: Duplicate check to L1427
```suggestion
```

sdanyliv (dismissed):

viceroypenguin (dismissed):

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaDMLService.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace LinqToDB.Internal.DataProvider.SapHana
+{
+	public class SapHanaDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "HanaException"))
+				return false;
+
+			return exception.Message.Contains("259", StringComparison.Ordinal);
```
MaceWindu: **Minor · MIN001** — The substring `"259"` is extremely permissive — it matches any HANA error message that happens to contain those three digits anywhere (line numbers in stack-like detail, other numeric error codes such as 2590/1259/12590, timestamps, identifiers, etc.). Because this swallows exceptions only when `throwExceptionIfNotExists: false` was requested, a matched false positive still hides a real error (permission, syntax, deadlock, …) and masks it as a successful no-op drop. Every other provider in this PR uses a more specific anchor (e.g. `ORA-00942`, `42P01`, `SQLSTATE 42S02`, `Code: 60`/`UNKNOWN_TABLE`, `1051`/`Unknown table`, `-204`, `no such table`); HANA is an outlier. SAP HANA error 259 is documented as `invalid table name`, and its canonical message is `invalid table name: Could not find table/view XXX in schema YYY: line N col M (at pos P)` — a better anchor is `"invalid table name"` (or `"feature not supported: 259"` if the intent is also to catch the numeric form produced by some drivers).

Fix: Tighten the marker to a string that only appears in the HANA "table not found" message. Match on `"invalid table name"` (case-insensitive) instead of the bare `"259"`; optionally keep a secondary `"Could not find table"` check to cover locale/variant wording.

```suggestion
			return exception.Message.Contains("invalid table name", StringComparison.OrdinalIgnoreCase)
				|| exception.Message.Contains("Could not find table", StringComparison.OrdinalIgnoreCase);
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDMLService.cs
```diff
@@ -0,0 +1,20 @@
...
+		{
+			if (!TypeOrMessageContains(exception, "OracleException"))
+				return false;
+
+			var message = exception.Message;
+
+			// Only ORA-00942 alone — NOT a compound error that includes other codes.
+			return  message.Contains("ORA-00942", StringComparison.Ordinal)
+				&& !message.Contains("ORA-14452", StringComparison.Ordinal)
+				&& !message.Contains("ORA-06512", StringComparison.Ordinal);
```
MaceWindu: **Minor · MIN002** — The inline comment says "Only ORA-00942 alone — NOT a compound error that includes other codes", but the implementation only excludes ORA-14452 (attempt to create index on temp table in use) and ORA-06512 (PL/SQL stack backtrace). Other common wrappers that legitimately embed ORA-00942 in a larger failure are not filtered: ORA-06550 (PL/SQL compilation error) and ORA-24344 (success with compilation error) both surface with an embedded ORA-00942 but represent a *compilation* failure rather than a genuine missing-table DROP path; ORA-04045 (error occurred while recompiling/revalidating) can do the same. With `throwExceptionIfNotExists: false`, these compound errors are silently swallowed as "table not found" and the caller never learns that recompilation failed. Either widen the exclusion list or invert the check so that we accept ORA-00942 only when it is the single ORA- code present in the message.

Fix: Either (a) extend the exclusion list to ORA-06550, ORA-24344, ORA-04045 alongside ORA-14452/ORA-06512, or (b) stop chasing the blacklist and accept ORA-00942 only when it is the single `ORA-` token in the message — count `ORA-` occurrences and reject when > 1. Option (b) is more robust against future wrappers. The suggestion below takes option (b) and removes the code-design-comment drift.

```suggestion
			// Accept ORA-00942 only when it is the sole ORA- code in the message; compound
			// errors (ORA-06512/ORA-06550/ORA-14452/ORA-24344/…) embed the same code inside
			// an unrelated failure that must not be swallowed.
			if (!message.Contains("ORA-00942", StringComparison.Ordinal))
				return false;

			var firstOra = message.IndexOf("ORA-", StringComparison.Ordinal);
			var nextOra  = firstOra < 0 ? -1 : message.IndexOf("ORA-", firstOra + 4, StringComparison.Ordinal);

			return nextOra < 0;
```

Tests/Linq/Update/DropTableTests.cs
```diff
@@ -204,13 +202,37 @@ public void Drop_Fail_NotFromExistCheck_EarlyError([DataSources] string context,
...
+		public void ShouldNotSuppress_NonDropTableErrors(
+			[DataSources(TestProvName.AllSQLite)] string context)
+		{
+			using var db = GetDataContext(context, o =>
+				o.UseOnEntityDescriptorCreated((_, _) => throw new CustomException()));
+
+			Assert.That(() =>
+				db.DropTable<NotTable>(throwExceptionIfNotExists: false),
+				Throws.InstanceOf<CustomException>());
+		}
```
MaceWindu: **Minor · MIN003** — This test duplicates `DropTable_Fail_NotFromExistCheck_EarlyError` at L159-165 (the pre-existing #798 test that had `[ActiveIssue]` removed in this PR). Both tests use the same `UseOnEntityDescriptorCreated((_, _) => throw new CustomException())` trick, so the thrown exception occurs **before** the new code's `try { ... } catch (Exception ex) when (suppress && IsTableNotFound(...))` block is even entered — `SqlTable.Create<T>` in `QueryRunner.DropTable<T>.PrepareQuery` is what invokes the entity-descriptor factory. The test therefore does not exercise the new `IsTableNotFoundException` path at all; it validates the same early-path propagation that the existing test already covers. It also excludes SQLite via `[DataSources(TestProvName.AllSQLite)]` with no comment explaining why a pre-query exception would behave differently on SQLite. What''s missing from this PR''s test coverage is a positive test of the *late* suppression path: with a real data context (not a throwing factory), provoke a non-missing-table provider error during DROP (e.g. `DropTable(tableName: "table-with-special-chars")` to trigger a syntax error, or a deliberately invalid schema name) and assert `throwExceptionIfNotExists: false` still surfaces that error — the actual regression #798 describes.

Fix: Either remove this duplicate, or replace it with a test that exercises the new code path: use a real connected context, cause the server to emit a non-table-not-found exception during DROP TABLE (invalid identifier, bad schema, insufficient permissions), and assert the error propagates even with `throwExceptionIfNotExists: false`. If the duplicate is kept intentionally, add a comment explaining why SQLite is excluded.

Tests/Linq/Update/DropTableTests.cs
```diff
@@ -204,13 +202,37 @@ public void Drop_Fail_NotFromExistCheck_EarlyError([DataSources] string context,
...
+		[Test]
+		public void DropTable_ShouldIgnore_Only_TableNotExists(
+			[DataSources(TestProvName.AllSQLite)] string context)
+		{
+
+			using var db = GetDataContext(context);
+
+			Assert.DoesNotThrow(() =>
+				db.DropTable<Table>(throwExceptionIfNotExists: false));
+		}
```
MaceWindu: **Minor · MIN004** — This test duplicates `DropTable_Missing_Ignore` at L145-150 (identical behavior: construct a real context, drop an unmapped table with `throwExceptionIfNotExists: false`, expect no throw). The only differences are the SQLite exclusion and the stray blank line at L230. The SQLite exclusion has no accompanying comment — SQLite supports `DROP TABLE IF EXISTS` natively, so the test would trivially pass there as it does everywhere, making the exclusion puzzling. As written the test also does not verify the stated claim in its name (`ShouldIgnore_Only_TableNotExists`): it never provokes a non-table-not-found error to assert it is *not* ignored; it only exercises the happy path of a missing table being ignored, which the sibling test already covers.

Fix: Either delete this test as a duplicate, or rewrite it to match its name: split into two assertions — (a) a real missing table drops without throwing (already covered by `DropTable_Missing_Ignore`, so drop this half), and (b) a DROP that fails for a reason *other* than missing table (bad schema, invalid identifier, permission) still throws. If a non-trivial exclusion reason for SQLite exists, annotate it; otherwise remove the exclusion. Also drop the blank line at L230 for consistency with the rest of the file.

Source/LinqToDB/Internal/DataProvider/DMLServiceBase.cs
```diff
@@ -0,0 +1,45 @@
...
+	{
+		public virtual bool IsTableNotFoundException(Exception exception)
+		{
+			ArgumentNullException.ThrowIfNull(exception);
+
+			for (var current = exception; current != null; current = current.InnerException)
+			{
+				if (IsTableNotFoundExceptionCore(current))
+					return true;
+			}
```
MaceWindu, line 27: **Suggestion · SUG001** — The chain walker traverses `InnerException` links only. `AggregateException.InnerExceptions` (multiple inners) is skipped except for its first-inner shortcut via `AggregateException.InnerException`. This matters for async DROP paths where multiple async faults can be surfaced as an AggregateException, and for some drivers that wrap transport errors into an AggregateException when retrying. Because `DropTable.QueryAsync` awaits the element task, .NET normally unwraps to the first inner for await propagation — but calling `Task.Result`/`.Wait()` or accessing the exception from a continuation can surface the full AggregateException. If a genuine `UNKNOWN_TABLE` is hidden under a second inner, the current walker misses it and a non-suppression is thrown even though the user asked for `throwExceptionIfNotExists: false`.

Fix: Teach the walker to recurse into every `AggregateException.InnerExceptions` branch in addition to following `InnerException`. This is a conservative extension — we widen matching, not narrowing — and it aligns with how the BCL''s `Flatten()` treats async exception trees. A small helper that yields every exception reachable via both `InnerException` and `InnerExceptions` keeps the public override surface unchanged.

```suggestion
			for (var current = exception; current != null; current = current.InnerException)
			{
				if (IsTableNotFoundExceptionCore(current))
					return true;

				if (current is AggregateException agg)
				{
					foreach (var inner in agg.Flatten().InnerExceptions)
						if (IsTableNotFoundException(inner))
							return true;
				}
			}
```

MaceWindu (changes requested):
> [!IMPORTANT]

## Review notes
- [x] API baselines refreshed — `Source/LinqToDB/CompatibilitySuppressions.xml` updated in this PR for the new `IDMLService` / `DMLServiceBase` Internal API additions and the `QueryHelper.SimplifyColumnExpression` removal.
- [x] **Scope.** Branch absorbed PR #5414's UPDATE-transformation work (merge commit 3a156a8) on top of the #798 DropTable fix — both maintainers have already approved. Flagging for awareness; no action requested.

## Baselines
Delta: [linq2db.baselines PR #1793](https://github.com/linq2db/linq2db.baselines/pull/1793) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5463)

### New DropTable_ShouldIgnore_Only_TableNotExists test baseline
47 new baseline files added across all providers for a single new test `Tests.xUpdate.DropTableTests.DropTable_ShouldIgnore_Only_TableNotExists`. SQL generation matches each provider's native capabilities.

- **Modern IF-EXISTS-capable** — SQL Server 2016+ and PostgreSQL emit native `DROP TABLE IF EXISTS`.
  - [DropTable_ShouldIgnore_Only_TableNotExists](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5463/SqlServer.2016.MS/Tests/xUpdate/DropTableTests/Tests.xUpdate.DropTableTests.DropTable_ShouldIgnore_Only_TableNotExists%28SqlServer.2016.MS%29.sql) — added (20 providers: SqlServer.2016, SqlServer.2016.MS, SqlServer.2017, SqlServer.2017.MS, SqlServer.2019, … (20 providers total))

    ```diff
    @@ -0,0 +1,4 @@
    +-- SqlServer.2016.MS SqlServer.2016
    +
    +DROP TABLE IF EXISTS [Table]
    +
    ```
    </details>
- **Legacy SQL Server (2005–2014)** — uses `IF (OBJECT_ID(…, ''U'') IS NOT NULL)` guard.
  - [DropTable_ShouldIgnore_Only_TableNotExists](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5463/SqlServer.2008/Tests/xUpdate/DropTableTests/Tests.xUpdate.DropTableTests.DropTable_ShouldIgnore_Only_TableNotExists%28SqlServer.2008%29.sql) — added (8 providers: SqlServer.2005, SqlServer.2005.MS, SqlServer.2008, SqlServer.2008.MS, SqlServer.2012, SqlServer.2012.MS, SqlServer.2014, SqlServer.2014.MS)

    ```diff
    @@ -0,0 +1,5 @@
    +-- SqlServer.2008
    +
    +IF (OBJECT_ID(N'[Table]', N'U') IS NOT NULL)
    +	DROP TABLE [Table]
    +
    ```
    </details>
- **Sybase variant** — `OBJECT_ID` without the `''U''` type specification.
  - [DropTable_ShouldIgnore_Only_TableNotExists](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5463/Sybase.Managed/Tests/xUpdate/DropTableTests/Tests.xUpdate.DropTableTests.DropTable_ShouldIgnore_Only_TableNotExists%28Sybase.Managed%29.sql) — added (1 providers: Sybase.Managed)

    ```diff
    @@ -0,0 +1,5 @@
    +-- Sybase.Managed Sybase
    +
    +IF (OBJECT_ID(N'Table') IS NOT NULL)
    +	DROP TABLE [Table]
    +
    ```
    </details>
- **Plain `DROP TABLE`** — MySQL/MariaDB/ClickHouse/Firebird/DB2/Informix/Access/SqlCe rely on client-side suppression via `IDMLService.IsTableNotFoundException`.
  - [DropTable_ShouldIgnore_Only_TableNotExists](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5463/MySql.5.7/Tests/xUpdate/DropTableTests/Tests.xUpdate.DropTableTests.DropTable_ShouldIgnore_Only_TableNotExists%28MySql.5.7%29.sql) — added (18 providers: MySql.5.7, MySql.8.0, MySqlConnector.5.7, MySqlConnector.8.0, MariaDB.11, … (18 providers total))

    ```diff
    @@ -0,0 +1,4 @@
    +-- MySql.5.7
    +
    +DROP TABLE [Table]
    +
    ```
    </details>

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaDMLService.cs
xamtam54: Fixed - use specific messages ("invalid table name", "could not find table[/view]") and added ODBC handling

Tests/Linq/Update/DropTableTests.cs
xamtam54: Removed as a duplicate of DropTable_Missing_Ignore.

Tests/Linq/Update/DropTableTests.cs
xamtam54: Removed. Also attempted a replacement using SchemaName("nonexistent_schema_xyz") but dropped it too.

Source/LinqToDB/Remote/RemoteDataContextBase.cs
```diff
@@ -193,6 +195,32 @@ public Expression Convert(Expression expression, out bool handled)
...
+		sealed class RemoteDMLService : IDMLService
+		{
+			static readonly MemoryCache<Type, IDMLService> _cache = new (new ());
+
+			public RemoteDMLService(IDMLService providerService)
+			{
+				ProviderService = providerService;
+			}
+
+			public IDMLService ProviderService { get; }
```
MaceWindu: I assume we don't need public on those members

MaceWindu (commented):
Looks really good. My main concern is a proper exception categorization. I will submit second review after I review exception.
Right now at least Oracle implementation needs  fix (see failing tests)

Source/LinqToDB/Internal/DataProvider/Access/AccessDMLService.cs
```diff
@@ -0,0 +1,30 @@
...
+		{
+			var message = exception.Message;
+
+			// Access via OleDb
+			if (TypeOrMessageContains(exception, "OleDbException"))
+			{
+				return message.Contains("could not find table",  StringComparison.OrdinalIgnoreCase)
+					|| message.Contains("cannot find the table", StringComparison.OrdinalIgnoreCase)
+					|| message.Contains("does not exist",        StringComparison.OrdinalIgnoreCase);
+			}
```
MaceWindu, line 17: - observed: `System.Data.OleDb.OleDbException : Table 'DropTableTest' does not exist.`
- HResult  set correctly to 0x80040e37 DB_E_NOTABLE

Can use HResult and `does not exist` token

Source/LinqToDB/Internal/DataProvider/Access/AccessDMLService.cs
```diff
@@ -0,0 +1,30 @@
...
+
+			// Access via ODBC - SQLSTATE 42S02 = base table or view not found
+			if (TypeOrMessageContains(exception, "OdbcException"))
+			{
+				return message.Contains("42S02",          StringComparison.OrdinalIgnoreCase)
+					|| message.Contains("does not exist", StringComparison.OrdinalIgnoreCase)
+					|| message.Contains("could not find", StringComparison.OrdinalIgnoreCase);
+			}
```
MaceWindu, line 24: observed: `System.Data.Odbc.OdbcException : ERROR [42S02] [Microsoft][ODBC Microsoft Access Driver] Table 'DropTableTest' does not exist.`
- 42S02 "Base table or view not found." is ODBC code we can use for all ODBC providers

Can use code lookup in message `42S02` and `does not exist`

Source/LinqToDB/Internal/DataProvider/SqlCe/SqlCeDMLService.cs
```diff
@@ -0,0 +1,18 @@
...
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "SqlCeException"))
+				return false;
+
+			var message = exception.Message;
+
+			return message.Contains("does not exist", StringComparison.OrdinalIgnoreCase)
+				|| message.Contains("not exist",      StringComparison.OrdinalIgnoreCase);
```
MaceWindu: - second string is a sub-string of first one
- observed: `System.Data.SqlServerCe.SqlCeException : The specified table does not exist. [ DropTableTest ]`

SqlCeException.HResult stores proper error code  0x80040e37 `DB_E_NOTABLE`, but it is implemented incorrectly in exception, so we hardly can use.

We can add HResult check (with note that it will not work due to broken implementation of it in `SqlCeException`) and message `specified table does not exist`

Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteDMLService.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace LinqToDB.Internal.DataProvider.SQLite
+{
+	public class SQLiteDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "SqliteException"))
+				return false;
+
+			return exception.Message.Contains("no such table", StringComparison.OrdinalIgnoreCase);
```
MaceWindu: not needed. we use `IF EXISTS` check

Source/LinqToDB/Internal/DataProvider/Firebird/FirebirdDMLService.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.Firebird
+{
+	public class FirebirdDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "FbException"))
```
MaceWindu: never hits as we already generate custom exists table check. Actually there is a functionality gap - for FB 5 dialect we can use `IF EXISTS` check, but it is out of scope of this pr (I've created separate issue #5483 )

Source/LinqToDB/Internal/DataProvider/PostgreSQL/PostgreSQLDMLService.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.PostgreSQL
+{
+	public class PostgreSQLDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "PostgresException"))
```
MaceWindu: not needed. We use IF EXISTS check

Source/LinqToDB/Internal/DataProvider/MySql/MySqlDMLService.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.MySql
+{
+	public class MySqlDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
```
MaceWindu: IF EXISTS used, not needed

Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseDMLService.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.ClickHouse
+{
+	public class ClickHouseDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "ClickHouse"))
```
MaceWindu: not needed, we generate EXISTS check

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDMLService.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.SqlServer
+{
+	public class SqlServerDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			// SQL Server (System.Data.SqlClient / Microsoft.Data.SqlClient)
```
MaceWindu: NOt need. We have Exists checks

Source/LinqToDB/Internal/DataProvider/DMLServiceBase.cs
```diff
@@ -0,0 +1,52 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider
+{
+	/// <summary>
+	/// Default <see cref="IDMLService"/> implementation. Conservative defaults:
+	/// <see cref="IsTableNotFoundException"/> returns <see langword="false"/> so that providers
+	/// without specific knowledge never swallow exceptions.
+	/// </summary>
+	public class DMLServiceBase : IDMLService
```
MaceWindu: make abstract and don't use as default implementation?

Source/LinqToDB/Internal/DataProvider/Informix/InformixDMLService.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.Informix
+{
+	public class InformixDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
```
MaceWindu: not needed, we have exists checks

Source/LinqToDB/Internal/DataProvider/DB2/DB2DMLService.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.DB2
+{
+	public class DB2DMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
```
MaceWindu: not needed, exist check generated

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaDMLService.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace LinqToDB.Internal.DataProvider.SapHana
+{
+	public class SapHanaDMLService : DMLServiceBase
+	{
+        protected override bool IsTableNotFoundExceptionCore(Exception exception)
+        {
+            var message = exception.Message;
+
+            // SAP HANA via native driver (Sap.Data.Hana)
+            if (TypeOrMessageContains(exception, "HanaException"))
```
MaceWindu: observed: `Sap.Data.Hana.HanaException : invalid table name: DropTableTest: line 1 col 12 (at pos 11)`

could use message `invalid table name`

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaDMLService.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                return message.Contains("invalid table name",   StringComparison.OrdinalIgnoreCase)
+                    || message.Contains("Could not find table", StringComparison.OrdinalIgnoreCase);
+            }
+
+            // SAP HANA via ODBC - SQLSTATE 42S02 = base table or view not found
+            if (TypeOrMessageContains(exception, "OdbcException"))
```
MaceWindu: observed: `System.Data.Odbc.OdbcException : ERROR [42S02] [SAP AG][LIBODBCHDB DLL][HDBODBC] Base table or view not found;259 invalid table name: DropTableTest: line 1 col 12 (at pos 11)`

as for access ODBC we can check for standard ODBC code 42S02 and for `Base table or view not found` message

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaDMLService.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace LinqToDB.Internal.DataProvider.SapHana
+{
+	public class SapHanaDMLService : DMLServiceBase
+	{
+        protected override bool IsTableNotFoundExceptionCore(Exception exception)
+        {
+            var message = exception.Message;
```
MaceWindu: Proper fix will be wrapper for Drop Table instead of this service. It almost the same wrapper as we use for DB2.

```sql
DO BEGIN
      DECLARE EXIT HANDLER FOR SQL_ERROR_CODE 259 BEGIN END;
      EXECUTE IMMEDIATE 'DROP TABLE "FOO"';
  END
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDMLService.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+
+namespace LinqToDB.Internal.DataProvider.Oracle
+{
+	public class OracleDMLService : DMLServiceBase
+	{
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
```
MaceWindu: not needed, we have exist check (and it actually creates issues for other tests), starting from v23 we can use `IF EXITSTS` when we add this dialect

MaceWindu (changes requested):
There are several concerns:
- too wide exception filters not reflecting observed error
- unclear how localization will affect matcher by message text
- for many providers we generate different kinds of EXISTS checks already and added DML service for them will never hit. This feature basically need implementation for several providers without checks.

I propose for providers with check (must be default implementation) either:
- make service optional (null) - preferable?
- use default implementation which return false from check method (as guardrail)

For remaining providers where exception possible I propose following checks when available:
- check HResult code in `ex.HResult`
- check HResult case-insensitive check (as hex, e.g. `0x80040e37`) in `ex.Message` (will work at least for Grpc transport)
- check for error message substring in `ex.Message` - fragile if messages could be localized

I've added notes per provider. basically we need exception inspection only for:
- Access
- SQL CE
- SAP HANA currently in this category, but for it we should instead implement missing exists check as better solution

Not included in PR, but YDB and Sybase (and upcoming DuckDB) also don't need service

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaSqlBuilder.cs
```diff
@@ -61,6 +61,38 @@ protected override string OffsetFormat(SelectQuery selectQuery)
...
+			{
+				// HANA has no native DROP TABLE IF EXISTS; wrap in an anonymous block that swallows
+				// SQL_ERROR_CODE 259 ("invalid table name") so the caller sees a clean no-op.
+				AppendIndent().Append(
+					"""
+					DO BEGIN
+						DECLARE EXIT HANDLER FOR SQL_ERROR_CODE 259 BEGIN END;
+						EXECUTE IMMEDIATE 'DROP TABLE
+					"""
+				);
```
MaceWindu: **Suggestion · SUG002** — Two concerns with the `EXECUTE IMMEDIATE ''DROP TABLE …''` wrapper:

1. **Identifier apostrophe escaping.** `BuildPhysicalTable(table, null)` writes the serialised identifier into the middle of the outer single-quoted SQL literal. HANA''s default identifier quoting uses double quotes (`"Schema"."Table"`), so under normal usage this is safe — but any table / schema name containing a literal apostrophe (legal inside a double-quoted identifier on HANA) would terminate the outer string early and either produce a syntax error or a malformed DROP. DB2''s `BuildDropTableStatement` has the identical gap; the new code is at least consistent across providers, but it could be hardened by escaping `''` → `''''` on the serialised identifier before emit.
2. **Formatting.** The C# raw-string strip puts `EXECUTE IMMEDIATE ''DROP TABLE` on its own line, then `BuildPhysicalTable` writes the identifier at column 0, then the closing `'';\nEND` appears on the next lines. HANA parses that fine (newlines are permitted inside a `''…''` literal), but the generated dynamic SQL reads as:
   ```
   EXECUTE IMMEDIATE ''DROP TABLE
   "SchemaName"."TableName"'';
   END
   ```
   — the identifier lands on a column-zero line, breaking visual alignment vs. DB2''s equivalent (which keeps `DROP TABLE <name>` on one line). Doesn''t affect correctness; does affect log readability.

Fix: capture the identifier into a scratch `StringBuilder` via `WithStringBuilder` (pattern already used in `BuildStartCreateTableStatement` below), apply `.Replace("''", "''''")` to the captured text, then emit the full `DROP TABLE <escaped-name>` on the same line as `EXECUTE IMMEDIATE ''…''`. Single-line dynamic SQL + safe against identifier apostrophes in one change. No concrete `suggestion` block since the fix crosses the method boundary.

Source/LinqToDB/Internal/DataProvider/SqlCe/SqlCeDMLService.cs
```diff
@@ -0,0 +1,20 @@
...
+		// HResult check below is best-effort and will usually miss.
+		const int DB_E_NOTABLE = unchecked((int)0x80040E37);
+
+		protected override bool IsTableNotFoundExceptionCore(Exception exception)
+		{
+			if (!TypeOrMessageContains(exception, "SqlCeException"))
+				return false;
+
+			return HResultMatches(exception, DB_E_NOTABLE)
+				|| exception.Message.Contains("specified table does not exist", StringComparison.OrdinalIgnoreCase);
```
MaceWindu, line 17: **Minor · MIN005** — The comment at L7–8 already admits the `HResultMatches(…)` clause will "usually miss" because `SqlCeException` never exposes the underlying code on `HResult`. In a local (non-remote) context the call is pure overhead per exception: `exception.HResult == DB_E_NOTABLE` never matches, and the secondary `"0x80040E37"` hex-in-`Message` check only helps on the remote path where the transport has serialised `exception.ToString()` into the wrapping exception''s message. Effectively every local DROP-failure dispatch is relying solely on the `"specified table does not exist"` substring — which is the one reliable signal.

Fix: drop the `HResultMatches(…)` clause and rely on the message substring alone (option a below). The comment + `const DB_E_NOTABLE` at L7–9 become dead and can be removed in the same change. Alternative (option b): keep the call but update the comment to clarify it only helps on the remote transport path — defensible but leaves the misleading "best-effort and will usually miss" wording in place.

```suggestion
			return exception.Message.Contains("specified table does not exist", StringComparison.OrdinalIgnoreCase);
```

Source/LinqToDB/Internal/DataProvider/DMLServiceBase.cs
```diff
@@ -0,0 +1,66 @@
...
+		/// provider exception is wrapped — the type name survives only as text inside the
+		/// wrapping exception's <see cref="Exception.Message"/> (populated via <see cref="Exception.ToString"/>).
+		/// </summary>
+		protected static bool TypeOrMessageContains(Exception exception, string marker)
+		{
+			var typeName = exception.GetType().FullName;
+
+			return (typeName != null && typeName.Contains(marker, StringComparison.Ordinal))
+				|| exception.Message.Contains(marker, StringComparison.Ordinal);
+		}
```
MaceWindu, line 50: **Minor · MIN006** — The `|| exception.Message.Contains(marker, …)` branch is explicitly a remote-transport accommodation (the XML doc above says so). On a local (non-remote) context, though, every exception whose `Message` just *mentions* the marker substring passes the gate — e.g. a user-authored wrapper `throw new InvalidOperationException("OleDbException chain while dropping X", inner)` would be accepted as an `OleDbException` candidate by `AccessDMLService.IsTableNotFoundExceptionCore`, which then falls through to the HResult/substring check on an exception whose type is not in fact `OleDbException`.

For Access the false positive still has to line up with *both* the type-or-message gate **and** the HResult / `"does not exist"` clause, so the blast radius is narrow — but it''s defense-in-depth the base class was meant to provide, given up gratuitously on the local path. The remote transport is where the message-text lookup is actually needed; the local path should inspect the real runtime type only.

Fix: split the helper into a local-only `TypeContains` (checks `exception.GetType().FullName` only) and a remote-aware `TypeOrMessageContainsRemote` (current logic). `AccessDMLService` / `SqlCeDMLService` call the local variant; only `RemoteDMLService` (or a remote-specific pre-filter) calls the remote variant. Alternative: thread an `isRemote` flag through `IsTableNotFoundException` — but splitting the method keeps the contract cleaner and leaves the Access / SqlCe call sites unchanged. Structural fix — no `suggestion` block.

Source/LinqToDB/Internal/DataProvider/Access/AccessDMLService.cs
MaceWindu, line 1: **Minor · MIN007** — No test exercises the *negative* case that issue #798 is actually about — an Access / SqlCe DROP that fails with a non-table-not-found error (bad identifier, bad schema, permission denied, lock) must still propagate when `throwExceptionIfNotExists: false`. The new contract on `IDMLService.IsTableNotFoundException` is precisely "suppress only table-not-found, rethrow everything else"; the tests under `Tests/Linq/Update/DropTableTests.cs` only cover the positive path (`DropTable_Missing_Ignore`, the new `DropTable_ShouldIgnore_Only_TableNotExists`) and the `throwExceptionIfNotExists: true` path — never the path where `IsTableNotFoundException` returns `false` and an exception flows out.

Prior review (MIN004) flagged the same gap; commit `4952108` attempted a replacement via `SchemaName("nonexistent_schema_xyz")` but the replacement was dropped as unreliable across providers. The gap therefore still stands on the two providers where the new client-side detection actually runs.

Fix: either (a) a test that provokes a non-missing-table error from DROP on `TestProvName.AllAccess | AllSqlCe` (bad identifier via a reserved word, or bad schema via `SchemaName("...")` scoped to just Access / SqlCe where the trigger is reliable) and asserts the exception propagates with `throwExceptionIfNotExists: false`; or (b) a unit test that stubs `IDMLService` and asserts `IsTableNotFoundException` returns `false` for a known-non-match (e.g. `new InvalidOperationException("permission denied")`) — at least locks in the detector''s contract.

Scoped to the file rather than a line: the fix lives in a new test method, not in `AccessDMLService.cs`.

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review, so this complements — rather than replaces — human judgement. Individual findings can be wrong, overconfident, or miss context a human would catch. Feel free to disagree, dismiss, or ask for clarification on any finding — the agent''s judgement is not authoritative. Findings are starting points for discussion, not verdicts.

## Review notes
- [x] **Prior review continuation.** IDs MIN001–MIN004 and SUG001 were used in the 2026-04-20 review. This review continues from MIN005 / SUG002. MIN001 (HANA "259" substring) and MIN002 (Oracle ORA-00942 compound) were removed in commit `b7a1af8` when those providers dropped client-side detection entirely — no re-flag needed.
- [x] **Public-API surface.** 11 new symbols. 10 live under `LinqToDB.Internal.*` (not part of the stable contract). One addition touches the public surface — `LinqToDB.Remote.LinqServiceInfo.DMLServiceType` (nullable string, `[DataMember(Order = 8)]`) — additive and non-breaking on the serialisation contract.
- [x] **Scope.** Review covers the DropTable exception-suppression refactor and its `IDMLService` plumbing. The #5414 UPDATE-transformation merge (commit `3a156a8`) is absorbed in the branch and out of scope here.

## Baselines

## Out-of-scope observations

Surfaced during review but fall outside this PR''s scope. Not findings on this PR — included as FYI.

- **Access / SqlCe "does not exist" message substring is English-only** — `AccessDMLService` and `SqlCeDMLService` fall back to the English message substring when the HResult / SQLSTATE anchor doesn''t match. OleDb / Odbc / SqlCe driver messages for these stacks are historically English on non-English Windows, so the fallback should stay reliable in practice, but a future MDAC / Access-runtime update that switched to Windows-localized strings would degrade the Odbc path (no HResult anchor there). Pre-existing behaviour on `master` was "swallow everything" — this PR only narrows that, never widens. Noted for future-proofing.
- **Remote-transport `IDMLService` dispatch depends on the server''s exception reaching the client intact** — `RemoteDMLService` resolves the server-side `IDMLService` type name (`Type.GetType(info.DMLServiceType)`) and calls `IsTableNotFoundException` on the **client** side. By the time a remote DROP error surfaces, the original provider exception is usually serialised to a string and re-thrown as a transport exception (`RpcException`, `FaultException`, etc.) — `HResult` is lost; `Message` is whatever the transport preserved. `DMLServiceBase.HResultMatches` accommodates this by also matching the hex form of the HResult inside `Message`, and `TypeOrMessageContains` checks the type name as text. The remote detection path is best-effort: it works when the transport round-trips `exception.ToString()` into the fault body, and misses when the transport strips it. Pre-existing on `master` (remote DROP always swallowed everything); the new code can only narrow suppression here. HANA''s move to a SQL-level `DO BEGIN … END` wrapper means the remote path for HANA no longer matters at all, which is a net improvement.

Source/LinqToDB/Internal/DataProvider/SqlCe/SqlCeDMLService.cs
MaceWindu: Don't drop. It is here at least for documentation reasons and potentially we can implement fix later using adapter mappings

Source/LinqToDB/Internal/DataProvider/DMLServiceBase.cs
MaceWindu: it doesn't worth it

Source/LinqToDB/Internal/DataProvider/Access/AccessDMLService.cs
MaceWindu: not interested

Source/LinqToDB/Internal/DataProvider/SapHana/SapHanaSqlBuilder.cs
MaceWindu: This is fixed. FOr all providers with similar generation

MaceWindu (approved):
Looks ready. Waiting for baselines update

MaceWindu:
Baselines looks great.
@xamtam54 , thanks for working on it! It also allowed us to detect some other issues.

PR #5470, opened 2026-04-16
Fix PostgreSQL array parameter query cache misses

### LINKED ISSUE #5302: Performance regression with query compilation
### Description

After upgrading to 6.1.0 one specific query seems to somewhat hang after connection but before actually sending the query to the database.

Rewriting the same query to Dapper (or downgrading back to 5.4.1) fixes that regression, but I'm unsure about what the actual cause is.

The odd thing is it seems like other queries are just fine, only this one seems to be affected.

In the tracings I see quite a discernable lag between connection open and query execution which is absent on both Dapper variant and 5.4.1. It looks like it spends too much time actually parsing.

Usually the query runs up to about 150 ms in 0.99q and 40 ms in 0.5q, while with 6.1.0 it runs with about 210 ms in 0.99q and 80-100 ms in 0.5q, a quite big increase in latency

### Reproduction Steps

```csharp
var availabilityQuery = PredicateBuilder.New<WarehouseTableDto>(false);

        availabilityQuery.Or(warehouse =>
            Sql.Ext.PostgreSQL()
                .ValueIsEqualToAny(warehouse.ClearingWarehouseId, availableClearingWarehouseIds));

        availabilityQuery.Or(warehouse =>
            Sql.Ext.PostgreSQL()
                .ValueIsNotEqualToAny(
                    warehouse.ClearingWarehouseId,
                    unavailableClearingWarehouseIds)
            && warehouse.CountryCode == defaultCountryCode);

        var typeQuery = PredicateBuilder.New<WarehouseTableDto>(false);

        foreach (var (type, index) in types
                     .Select(
                         (wmsGrp, index) => ((
                             WmsTypeId: wmsGrp.Key,
                             WarehouseTypeIds: wmsGrp.ToArray()), index)))
        {
            typeQuery.Or(warehouse => warehouse.WmsType == type.WmsTypeId
                && Sql.Ext.PostgreSQL().ValueIsEqualToAny(warehouse.Type, type.WarehouseTypeIds));
        }

        var query = connection
            .GetTable<WarehouseTableDto>()
            .Where(availabilityQuery)
            .Where(warehouse => warehouse.IsActive
                && Sql.Ext.PostgreSQL()
                    .Contains(
                        warehouse.Attributes.BitFlipEnum().FlagToFlagidxArray(), // FlagToFlagidxArray see below
                        WarehouseAttributeDto.Sandbox.FlagToFlagidxArray()))
            .Where(typeQuery)
.SelectMany(
                warehouse => connection.GetTable<CountryDto>()
                    .Where(c => c.NumericCode == warehouse.CountryCode),
                (warehouse, country) => new WarehouseDto(
                    /* A lot of unrelevant fields here */))


[Sql.Expression("flag_to_flagidx_array({0})", ServerSideOnly = true)]
    public static int[] FlagToFlagidxArray<T>([ExprParameter] this T flags)
        where T : Enum
    {
        throw new InvalidOperationException("For Linq2Db usage only");
    }
```

This query variant with Dapper (the same) looks like this (and runs fine):
```csharp
var builder = new SqlBuilder();

        var query = builder.AddTemplate(
            $"""
             SELECT
             	/* A lot of unrelevant fields here */
             FROM
             	warehouses warehouse
             		INNER JOIN countries country ON country.numeric_code = warehouse.country_code
             /**where**/
             """);

        builder.Where("""
                      warehouse.is_active = true
                      AND
                          flag_to_flagidx_array(~(warehouse."attribute_flags")) @>
                              flag_to_flagidx_array(@IsSandboxAttribute)
                      """,
                      new
                      {
                          IsSandboxAttribute
                      });

        builder.Where(
            $"""
                 (warehouse.clearing_warehouse_id = ANY(@AvailableClearingWarehouseIds)
                 OR (
                    warehouse.clearing_warehouse_id <> ANY(@UnavailableClearingWarehouseIds)
                    AND warehouse.country_code = @DefaultCountryCode
                    )
                 )
                 """,
            new
            {
                AvailableClearingWarehouseIds = availableClearingWarehouseIds,
                UnavailableClearingWarehouseIds = unavailableClearingWarehouseIds,
                DefaultCountryCode = defaultCountryCode
            });

        if (types.Any())
        {
            var typesQueryParts = new List<string>(types.Count);
            var typesParams = new DynamicParameters();
            foreach (var (type, index) in types
                         .Select(
                             (wmsGrp, index) => ((
                                 WmsTypeId: wmsGrp.Key,
                                 WarehouseTypeIds: wmsGrp.ToImmutableArray()), index)))
            {
                typesQueryParts.Add(
                    $"""
                     (warehouse.wms_type_id = @WmsTypeId{index}
                        AND warehouse.type_id = ANY(@TypeIds{index}))
                     """);

                typesParams.Add($"@WmsTypeId{index}", type.WmsTypeId);
                typesParams.Add($"@TypeIds{index}", type.WarehouseTypeIds);
            }

            builder.Where($"({string.Join(" OR ", typesQueryParts)})", typesParams);
        }
```

Should result in SQL something like this:
```sql
SELECT
/* A lot of unrelevant fields here */
	FROM
		warehouses warehouse
			INNER JOIN countries country ON country.numeric_code = warehouse.country_code
	WHERE warehouse.is_active = true
	AND
	    flag_to_flagidx_array(~(warehouse."attribute_flags")) @>
	        flag_to_flagidx_array($1) AND (warehouse.clearing_warehouse_id = ANY($2)
	OR (
	   warehouse.clearing_warehouse_id <> ANY($3)
	   AND warehouse.country_code = $4
	   )
	) AND ((warehouse.wms_type_id = $5
	   AND warehouse.type_id = ANY($6)) OR (warehouse.wms_type_id = $7
	   AND warehouse.type_id = ANY($8)))
	
```

### Exception details

---

### Regression?

Yes, works ok with 5.4.1

### Linq To DB version

6.1.0

### .NET version

10.0.100

### Database

Postgresql 13

### Database provider

Npgsql 10

### Operating system

macOS, Linux

### Configuration

_No response_

sdanyliv:
Fixes #5302

Array parameters in `Sql.Extension` methods (e.g. `ValueIsEqualToAny`) caused query cache misses on every execution because arrays were compared by reference in the cache key comparison.

Register primitive array types as scalar in `PostgreSQLMappingSchema` so they are treated as normal parameters instead of going through `MarkAsValue` path.

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
+++ b/Source/LinqToDB/Internal/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
@@ -6,17 +6,16 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Net;
-using System.Net.NetworkInformation;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 
-using LinqToDB;
 using LinqToDB.Common;
 using LinqToDB.Data;
 using LinqToDB.Expressions;
 using LinqToDB.Internal.Expressions;
 using LinqToDB.Internal.Expressions.Types;
+using LinqToDB.Internal.Extensions;
 using LinqToDB.Mapping;
 
 namespace LinqToDB.Internal.DataProvider.PostgreSQL
@@ -487,11 +486,15 @@ namespace LinqToDB.Internal.DataProvider.PostgreSQL
 					{
 						mappingSchema.AddScalarType(type, null, true, dataType);
 						mappingSchema.AddScalarType(type.MakeArrayType(), null, true, dataType | DataType.Array);
+						mappingSchema.AddScalarType(type.MakeListType(), null, true, dataType | DataType.Array);
+						mappingSchema.AddScalarType(type.MakeIReadOnlyListType(), null, true, dataType | DataType.Array);
 					}
 					else
 					{
 						mappingSchema.AddScalarType(type, dataType);
 						mappingSchema.AddScalarType(type.MakeArrayType(), dataType | DataType.Array);
+						mappingSchema.AddScalarType(type.MakeListType(), dataType | DataType.Array);
+						mappingSchema.AddScalarType(type.MakeIReadOnlyListType(), dataType | DataType.Array);
 					}
 				}
 			}
--- a/Source/LinqToDB/Internal/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
+++ b/Source/LinqToDB/Internal/DataProvider/PostgreSQL/NpgsqlProviderAdapter.cs
@@ -275,18 +275,18 @@ namespace LinqToDB.Internal.DataProvider.PostgreSQL
 
 				// date/time types
 				if (npgsqlDateType != null)
-					AddUdtType(npgsqlDateType);
+					AddScalarType(npgsqlDateType);
 				if (npgsqlDateTimeType != null)
-					AddUdtType(npgsqlDateTimeType);
+					AddScalarType(npgsqlDateTimeType);
 				if (npgsqlTimeSpanType != null)
 				{
-					mappingSchema.AddScalarType(npgsqlTimeSpanType, DataType.Interval);
+					AddScalarType(npgsqlTimeSpanType, DataType.Interval);
 				}
 
 				Expression? npgsqlIntervalReader = null;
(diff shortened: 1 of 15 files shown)


viceroypenguin (dismissed):
This will fix the immediate issue; we may should have a separate conversation about the applicability of improving caching for all arrays/lists below a certain size. For example: `context.Values.Where(v => v.Col1.In(1, 2, 3))....`.

MaceWindu (dismissed):

MaceWindu (approved):

PR #1808, opened 2019-07-10
DB2 date/time/timestamp support improvements

MaceWindu:
Fixes #1663 

- [refactoring] removed unused `MappingSchema` parameter from `IDataContext.GetReaderExpression`, `IDataProvider.GetReaderExpression`
- `InlineParameters` option applied only to parameters of types, that we can convert to literal
- fixed literal generation for DB2 date/time/timestamp types
- reenable TimeSpan literals for DB2

PR #5478, opened 2026-04-19
Add MATERIALIZED CTE hint via AsCte builder (#5323)

### LINKED ISSUE #5323: PostgreSQL CTE MATERIALIZED
### Description

Starting with PostgreSql 12 it [supports](https://www.postgresql.org/docs/12/queries-with.html) special keywords `MATERIALIZED` and `NOT MATERIALIZED` to rule CTE materialization inside query.

Seems like current Linq2Db API doesn't support this and it is very odd, because PG released this feature **7 years ago** (PG 12 published in 2019)...

I tried to find workaround with current hints implementation, but no success:

```cs
query.SubQueryHint("MATERIALIZED").AsCte(); // Produces MATERIALIZED after CTE body of subquery
query.QueryHint("MATERIALIZED").AsCte(); // Produces MATERIALIZED after CTE body of entire query
```

No one seems really works and I got exception:
```
Npgsql.PostgresException (0x80004005): 42601: syntax error at or near "MATERIALIZED"
```

### New API

I assume it should work like query hints, something like:

```cs
query.CteHint("MATERIALIZED").AsCte();
query.CteHint("NOT MATERIALIZED").AsCte();
```

Or be like AsCte extensions in PostgreSQL specific namespaces:

```cs
query.AsPostgreSQL().AsCte(LinqToDB.(?).Materialized);
query.AsPostgreSQL().AsCte(LinqToDB.(?).NotMaterialized);
```

sdanyliv:
## Summary

Closes #5323.

Adds a provider-neutral fluent way to mark a CTE as `MATERIALIZED` / `NOT MATERIALIZED`:

```csharp
query.AsCte(b => b.IsMaterialized());              // AS MATERIALIZED
query.AsCte(b => b.IsMaterialized(false));         // AS NOT MATERIALIZED
query.AsCte(b => b.HasName("x").IsMaterialized()); // named + materialized
query.AsCte();                                     // unchanged baseline
```

Provider support matrix:

| Provider       | `AS MATERIALIZED` | `AS NOT MATERIALIZED`                      |
|----------------|-------------------|--------------------------------------------|
| PostgreSQL 12+ | ✅ emitted        | ✅ emitted                                 |
| SQLite 3.35+   | ✅ emitted        | ✅ emitted                                 |
| ClickHouse 26.3+ | ✅ emitted      | ⚠️ silently dropped (engine has no keyword)|
| Others         | silently ignored  | silently ignored                           |

## Design notes

- `CteClause.Annotations : Annotatable` — open-ended metadata bag on the SQL AST node, backed by new `Infrastructure` types (`IAnnotation`, `IMutableAnnotatable`, `AnnotatableBase`, …) intended for provider/extension metadata that shouldn't bloat the core AST.
- `ICteBuilder` (public) exposes only `HasName(…)`. Provider-specific fluent extensions reach the annotation bag through `IAnnotatableBuilderInternal` (a narrow internal surface) instead of downcasting to the concrete builder implementation — symmetric with how third-party extensions would be authored.
- `IsMaterialized(bool=true)` stashes `CteAnnotationNames.Materialized` into the builder's annotation bag and that is copied onto `CteClause` during expression translation.
- `BasicSqlBuilder` has two opt-in flags `SupportsMaterializedCteHint` (default false) and `SupportsNotMaterializedCteHint` (defaults to the former). Supporting providers override with a single line each. Emission replaces the default `AS` line with `AS MATERIALIZED` / `AS NOT MATERIALIZED` when the annotation is present.
- `IExpressionCacheKey` marker: `CteAnnotationsContainer` implements it so `ExpressionCacheHelpers.ShouldRemoveConstantFromCache` keeps the value in the cache key (otherwise distinct configurations collapse to the same compiled query).
- `EagerEvaluationAttribute` + hook in `ExposeExpressionVisitor.TryConvertIQueryable`: desugars `AsCte(Action<ICteBuilder>)` inside lambda/`SelectMany` contexts where the compiler captures the whole call as an expression tree rather than executing the public overload.
- `LinqServiceSerializer` round-trips `CteClause.Annotations`; `QueryElementVisitor.VisitCteClause` (Transform mode) and `SqlQueryCloneVisitor` carry annotations forward when cloning.
- CI / Docker setup (`Build/Azure/scripts/clickhouse.sh`, `Data/Setup Scripts/clickhouse.cmd`) enables `enable_analyzer` + `enable_materialized_cte` in the default user profile so the feature is exercisable against the ClickHouse container.

## Out of scope

- Recursive CTEs — they are always materialized by the engine; the hint is redundant (PG/SQLite) or rejected. The fluent builder only feeds the non-recursive `BuildCteContext` path; a comment in `BuildRecursiveCteContextTable` documents that annotations are intentionally not carried there.
- Oracle `/*+ MATERIALIZE */` / `/*+ INLINE */` (different syntax — future work, can override `BuildCteHeaderHint` directly).
- PostgreSQL/SQLite/ClickHouse version gating — the underlying engine already raises a clear error on older versions.

### `AsCte(b => b.IsMaterialized())`

**PostgreSQL.16**
```sql
WITH "CTE_1" ("ParentID")
AS MATERIALIZED
(
	SELECT
		c_1."ParentID"
	FROM
		"Child" c_1
	WHERE
		c_1."ParentID" > 1
)
SELECT
	p."ParentID",
	p."Value1"
FROM
	"Parent" p
		INNER JOIN "CTE_1" c_2 ON p."ParentID" = c_2."ParentID"
```

**SQLite**
```sql
WITH [CTE_1] ([ParentID])
AS MATERIALIZED
(
	SELECT
		[c_1].[ParentID]
	FROM
		[Child] [c_1]
	WHERE
		[c_1].[ParentID] > 1
)
SELECT
	[p].[ParentID],
	[p].[Value1]
FROM
	[Parent] [p]
		INNER JOIN [CTE_1] [c_2] ON [p].[ParentID] = [c_2].[ParentID]
```

**ClickHouse**
```sql
WITH CTE_1 AS MATERIALIZED
(
	SELECT
		c_1.ParentID as ParentID
	FROM
		Child c_1
	WHERE
		c_1.ParentID > 1
)
SELECT
	p.ParentID,
	p.Value1
FROM
	Parent p
		INNER JOIN CTE_1 c_2 ON p.ParentID = c_2.ParentID
```

**SQL Server** (hint silently ignored — plain `AS` emitted.)

---

### `AsCte(b => b.IsMaterialized(false))`

**PostgreSQL.16**
```sql
WITH "CTE_1" ("ParentID")
AS NOT MATERIALIZED
(
	SELECT
		c_1."ParentID"
	FROM
		"Child" c_1
	WHERE
		c_1."ParentID" > 1
)
SELECT
	p."ParentID",
	p."Value1"
FROM
	"Parent" p
		INNER JOIN "CTE_1" c_2 ON p."ParentID" = c_2."ParentID"
```

**SQLite** (same `AS NOT MATERIALIZED` shape with `[brackets]`.)

**ClickHouse** (silently falls back — no `NOT MATERIALIZED` keyword in the engine)
```sql
WITH CTE_1 AS
(
	SELECT
		c_1.ParentID as ParentID
	FROM
		Child c_1
	WHERE
		c_1.ParentID > 1
)
SELECT
	p.ParentID,
	p.Value1
FROM
	Parent p
		INNER JOIN CTE_1 c_2 ON p.ParentID = c_2.ParentID
```

---

### `AsCte(b => b.HasName("active_children").IsMaterialized())`

**PostgreSQL.16**
```sql
WITH active_children ("ParentID")
AS MATERIALIZED
(
	SELECT
		c_1."ParentID"
	FROM
		"Child" c_1
	WHERE
		c_1."ParentID" > 1
)
SELECT
	p."ParentID",
	p."Value1"
FROM
	"Parent" p
		INNER JOIN active_children c_2 ON p."ParentID" = c_2."ParentID"
```

---

### Inside `SelectMany` (query-syntax `from c in …AsCte(…)`)

Verifies the `EagerEvaluation` desugaring in `ExposeExpressionVisitor` — the public overload isn't executed directly here.

**PostgreSQL.16**
```sql
WITH "CTE_1" ("ParentID")
AS MATERIALIZED
(
	SELECT
		c_1."ParentID"
	FROM
		"Child" c_1
	WHERE
		c_1."ParentID" > 1
)
SELECT
	p."ParentID",
	p."Value1"
FROM
	"Parent" p,
	"CTE_1" c_2
WHERE
	p."ParentID" = c_2."ParentID"
```

## Test plan

- [x] `Source/LinqToDB/LinqToDB.csproj` Debug + Release clean (0 warnings, 0 errors on net462, netstandard2.0, net8, net9, net10; analyzers on in Release).
- [x] `Tests/Linq/Linq/CteTests.cs` — existing suite passing on net10.0 (direct + LinqService).
- [x] `Tests/Linq/Linq/CteMaterializedTests.cs` — 9 tests covering positive emission, silent fallbacks, lambda-captured calls, empty-builder baseline, hint overwrite, and cache-key differentiation. Executes with `.ToList()` against the actual DB engine.
- [x] `Tests/Linq/Infrastructure/AnnotatableTests.cs` — 17 tests incl. `OnAnnotationSet` callback and no-op `SetAnnotation` short-circuit.
- [x] Full combined run on net10.0: **646/646 passing**.
- [x] CI ClickHouse container picks up `enable_analyzer` + `enable_materialized_cte` from the updated `clickhouse.sh`.


Source/LinqToDB/Internal/Linq/Builder/CteAnnotationsContainer.cs
```diff
@@ -0,0 +1,57 @@
...
+	/// reference-equal delegate values would otherwise collapse to the same cache entry.
+	/// Wraps an <see cref="Annotatable"/> so callers can reuse the generic annotation API
+	/// for reading and writing provider-specific CTE metadata.
+	/// </summary>
+	public sealed class CteAnnotationsContainer : IExpressionCacheKey, IEquatable<CteAnnotationsContainer>
+	{
+		public static readonly CteAnnotationsContainer Empty = new(null, new Annotatable());
+
+		public string?     Name        { get; }
+		public Annotatable Annotations { get; }
```
MaceWindu: **MAJ001** — The class header describes this container as an "Immutable carrier" and it implements `IExpressionCacheKey` (whose contract is "Implementers must be immutable and provide value-based equality"), but `Annotations` exposes the underlying mutable `Annotatable` by reference, and `_cacheKey` is computed once in the constructor. A caller that captures the builder (`AsCte(b => { _keep = b; b.IsMaterialized(); })`) — or any code holding the same `Annotatable` instance passed into the container — can mutate the bag after construction: `container.Annotations.SetAnnotation(...)` succeeds, `container.GetHashCode()` / `Equals(other)` keep returning the snapshotted values, but `CteClause.Annotations` is later populated from the *live* bag in `TableBuilder.CteTableContext.BuildCteContext`. That's a silent cache-key / emitted-SQL divergence — the query cache can return a plan built from one annotation set for input whose hash still matches but whose live state is different.

Fix: Snapshot the annotations into an internally-owned bag and stop handing out the mutable `Annotatable` by reference. Simplest path: keep the bag `private readonly`, expose only read-only accessors (`GetAnnotations()` / `FindAnnotation(string)`), and have the constructor `AddAnnotations` from the caller's bag into a new local `Annotatable`. Downstream consumers (`TableBuilder.CteTableContext`, tests) already iterate via `GetAnnotations()`, so the change is local.

— ✓ Fixed in 05915240

Source/LinqToDB/LinqExtensions/CteBuilderExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+		/// <param name="cteBuilder">Builder provided by the <c>AsCte</c> callback.</param>
+		/// <param name="materialized">
+		/// <see langword="true"/> to emit <c>AS MATERIALIZED</c>, <see langword="false"/> to emit <c>AS NOT MATERIALIZED</c>.
+		/// </param>
+		/// <returns>The same builder for chaining.</returns>
+		public static ICteBuilder IsMaterialized(this ICteBuilder cteBuilder, bool materialized = true)
+		{
+			ArgumentNullException.ThrowIfNull(cteBuilder);
+
+			((IAnnotatableBuilderInternal)cteBuilder).SetAnnotation(CteAnnotationNames.Materialized, materialized);
```
MaceWindu: **MAJ002** — `ICteBuilder` is `[PublicAPI]` in the `LinqToDB` namespace — an interface users are expected to consume and can trivially re-implement. The in-box implementation (`CteBuilderImpl`) happens to implement `IAnnotatableBuilderInternal`, but nothing on `ICteBuilder` requires or documents that extra interface. The unchecked downcast here throws `InvalidCastException` the first time a caller passes a hand-written `ICteBuilder` (test fake, mock, custom wrapper) to `IsMaterialized`. The exception surfaces from a one-line public extension method with no hint as to why.

Fix: Either make the contract defensive (`if (cteBuilder is not IAnnotatableBuilderInternal impl) throw new NotSupportedException("…");` then `impl.SetAnnotation(...)`) or make it explicit by exposing a `SetAnnotation` method on `ICteBuilder` itself so the interface advertises everything the extensions need. Either way, the `InvalidCastException` should not reach end users.

— ✓ Fixed in 05915240

Source/LinqToDB/Internal/Linq/Builder/CteAnnotationsContainer.cs
```diff
@@ -0,0 +1,57 @@
...
+	/// Immutable carrier for the state produced by the <see cref="ICteBuilder"/> callback.
+	/// Placed into the query expression tree as a single constant so linq2db's query-cache
+	/// equality comparer can discriminate distinct <c>AsCte(b =&gt; ...)</c> configurations:
+	/// reference-equal delegate values would otherwise collapse to the same cache entry.
+	/// Wraps an <see cref="Annotatable"/> so callers can reuse the generic annotation API
+	/// for reading and writing provider-specific CTE metadata.
+	/// </summary>
+	public sealed class CteAnnotationsContainer : IExpressionCacheKey, IEquatable<CteAnnotationsContainer>
+	{
+		public static readonly CteAnnotationsContainer Empty = new(null, new Annotatable());
```
MaceWindu: **MIN001** — `CteAnnotationsContainer.Empty` has no consumers in `Source/*` or `Tests/*` — the `AsCte(Action<ICteBuilder>)` overload always constructs a fresh container even for an empty callback (verified by `AsCte_Builder_Empty_MatchesBaseline`). Shipping unreferenced public statics on a type just enlarges the API surface and commits to maintaining them.

Fix: Drop `Empty` until there's a real consumer — or, if the intent is for third-party CTE extensions to use it as a sentinel, mark it `internal` and add a call site that needs it.

— ✓ Fixed in 05915240

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
```diff
@@ -1044,6 +1044,50 @@ public static IQueryable<TSource> AsCte<TSource>(
...
+			ArgumentNullException.ThrowIfNull(source);
+			ArgumentNullException.ThrowIfNull(cteBuilder);
+
+			// Evaluate the builder up front and reduce its state to a cache-friendly
+			// value-equatable container. Placing the delegate itself in the expression tree
+			// causes the query cache to collapse distinct configurations.
+			var impl = new CteBuilderImpl();
+			cteBuilder(impl);
+
+			var container = new CteAnnotationsContainer(impl.Name, impl.Annotations);
```
MaceWindu: **MIN002** — `CteAnnotationsContainer` receives `impl.Annotations` (the *live* `Annotatable` instance from `CteBuilderImpl`). The common case returns before `impl` is reachable again, but nothing stops the caller from stashing the builder in their lambda (`AsCte(b => { _saved = b; b.IsMaterialized(); })`) and mutating `impl.Annotations` later. Because `container.Annotations` is the same reference, subsequent `CteClause` construction in `BuildCteContext` sees the mutated state while the already-computed `container._cacheKey` still reflects the original. Same underlying issue as **MAJ001** — this is the producer side.

Fix: Once `CteAnnotationsContainer` copies annotations into an internally-owned bag (per **MAJ001**), this site becomes a straight hand-off. If MAJ001 is resolved via a separate snapshot helper instead, copy the annotations here rather than passing `impl.Annotations` through by reference.

— ✓ Fixed in 05915240

Source/LinqToDB/Internal/Infrastructure/AnnotatableBase.cs
```diff
@@ -0,0 +1,283 @@
...
+		/// <summary>
+		///     Adds annotations to this object.
+		/// </summary>
+		/// <param name="annotations">The annotations to be added.</param>
+		public virtual void AddAnnotations(IEnumerable<IAnnotation> annotations)
+		{
+			ArgumentNullException.ThrowIfNull(annotations);
+
+			AddAnnotations(this, annotations);
+		}
```
MaceWindu: **MIN003** — `AnnotatableBase.AddAnnotations(IEnumerable<IAnnotation>)` null-guards and loops through the static helper. `AnnotatableExtensions.AddAnnotations(IMutableAnnotatable, IEnumerable<IAnnotation>)` duplicates the same logic and is selected by overload resolution when the caller's variable is typed as `IMutableAnnotatable` rather than `AnnotatableBase`. Two parallel implementations doing the same `foreach` is a maintenance hazard — future changes (validation, tracing, etc.) have to be mirrored.

Fix: Have one delegate to the other. Since `IMutableAnnotatable.AddAnnotation` is already on the interface, the extension can route through it and the instance method can call the extension — keeping a single loop. If the intent is that the instance method is virtual and overridable, document that and make the extension go through it so subclasses' overrides are respected.

— ✓ Fixed in 05915240

Source/LinqToDB/Internal/Linq/Builder/Visitors/ExposeExpressionVisitor.cs
```diff
@@ -382,6 +383,19 @@ bool TryConvertIQueryable(Expression node, out Expression converted)
...
+					// call patterns like AsCte(Action<ICteBuilder>) even when wrapped inside
+					// a SelectMany lambda.
+					if (mc.Method.HasAttribute<EagerEvaluationAttribute>(inherit: false))
+					{
+						if (IsCompilable(mc))
+						{
+							converted = ConvertIQueryable(node);
+							return !ExpressionEqualityComparer.Instance.Equals(converted, node);
+						}
+					}
```
MaceWindu, line 403: **MIN004** — When `IsCompilable(mc)` is false, the `[EagerEvaluation]` branch silently falls through to the default `mc.IsQueryable` handling below. Because `AsCte(IQueryable<T>, Action<ICteBuilder>)` isn't tagged `[IsQueryableAttribute]`, that path returns `converted = mc; false;` — the call reaches `TableBuilder.CteTableContext.BuildCteContext` unchanged, which only decodes two arg shapes (`string` or `CteAnnotationsContainer`). Neither matches the untransformed `Action<ICteBuilder>` argument, so `HasName` / `IsMaterialized` settings are silently dropped. A non-compilable builder lambda is unusual (it would require a non-client-evaluable closure referent) but the failure is a silent correctness regression.

Fix: When `[EagerEvaluation]` is present and `IsCompilable` is false, throw a `LinqException` / `InvalidOperationException` explaining that the method requires client-side evaluation. Alternatively, explicitly reject any `AsCteInternal` arg shape other than `string` / `CteAnnotationsContainer` in `BuildCteContext`. A short test with a non-evaluable capture in the builder lambda would pin the behaviour.

— ✓ Fixed in 05915240

Source/LinqToDB/Internal/Linq/Builder/CteBuilderImpl.cs
```diff
@@ -0,0 +1,27 @@
...
+	/// Captures configuration chosen by the user so it can be copied onto the emitted <see cref="SqlQuery.CteClause"/>.
+	/// </summary>
+	/// <remarks>
+	/// Extension methods that live under <c>LinqToDB</c> set provider-specific state by casting to
+	/// <see cref="IAnnotatableBuilderInternal"/> and calling its members.
+	/// </remarks>
+	public sealed class CteBuilderImpl : ICteBuilder, IAnnotatableBuilderInternal
+	{
+		public string?     Name        { get; private set; }
+		public Annotatable Annotations { get; } = new();
```
MaceWindu: **SUG001** — `CteBuilderImpl.Annotations` is public and returns the mutable `Annotatable` that `CteAnnotationsContainer` ends up sharing. `CteBuilderImpl` is the only `ICteBuilder` implementation shipped in-box and lives in `Internal.Linq.Builder`, so nothing outside the internal pipeline should reach it — but the type and its members are `public`, making `((CteBuilderImpl)cteBuilder).Annotations` implicitly reachable when consumers downcast.

Fix: Narrow `Annotations` (and the `Name` setter) to `internal`. The `IAnnotatableBuilderInternal.SetAnnotation` method is the documented access channel; the public property is redundant.

— ✓ Fixed in 05915240

Source/LinqToDB/LinqExtensions/CteBuilderExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+	/// <summary>
+	/// Extension methods that expose CTE options on <see cref="ICteBuilder"/>.
+	/// </summary>
+	[PublicAPI]
+	public static class CteBuilderExtensions
+	{
+		/// <summary>
+		/// Sets the <c>MATERIALIZED</c> / <c>NOT MATERIALIZED</c> CTE hint.
+		/// Recognized by PostgreSQL 12+, SQLite 3.35+, and ClickHouse 26.3+. Other providers silently ignore it.
```
MaceWindu: **SUG002** — The PR docs and XML comments say the materialization hint is "silently ignored" on providers that don't support it (covered by `AsCte_Materialized_UnsupportedProvider_IgnoredSilently`). For a targeted performance hint this is debatable UX — a developer who wrote `.IsMaterialized()` for PostgreSQL perf tuning gets no signal when their query later runs against SQL Server. Consider routing the dropped hint through `DataConnection.TraceSwitch` / `OnTraceConnection` at `TraceLevel.Warning` (the linq2db idiom for "engine can't support this" notices). Flagging so the maintainers have an explicit position documented.

Fix: If the "silent" stance is intentional, add a sentence to the `IsMaterialized` / `CteAnnotationNames.Materialized` XML docs making it explicit ("Providers that don't recognise the hint drop it without tracing."). If a trace is acceptable, emit it from `BasicSqlBuilder.BuildCteHeaderHint` when `ShouldEmitMaterializedCteHint` returns false but an annotation is present.

— ✓ Fixed in 05915240

Source/LinqToDB/Internal/Linq/EagerEvaluationAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// <remarks>
+	///     Use this on syntactic-sugar overloads that normalise themselves into a richer internal
+	///     call (for example <c>AsCte(Action&lt;ICteBuilder&gt;)</c> which desugars to
+	///     <c>AsCteInternal(CteAnnotationsContainer)</c>). Without this marker, the exposer only
+	///     evaluates calls whose first argument is a <see cref="System.Linq.Expressions.MemberExpression"/>
+	///     or <see cref="System.Linq.Expressions.ConstantExpression"/>; lambda-captured calls with
+	///     a method-chain source would never get the client-side rewrite.
+	/// </remarks>
+	[AttributeUsage(AttributeTargets.Method, Inherited = false)]
+	public sealed class EagerEvaluationAttribute : Attribute
```
MaceWindu, line 39: **SUG003** — `[EagerEvaluation]` is now a fully general hook — `ExposeExpressionVisitor.TryConvertIQueryable` will client-evaluate *any* method so marked, invoking side-effecting code (like the builder callback that mutates an `Annotatable`) during expression analysis. The attribute lives in `Internal.Linq.*` so isn't strictly public API, but adding a general-purpose "run this method on the client" escape hatch is worth a design note. Two concerns: (1) the invocation happens during expression exposure, which can run more than once per query if an intercepting visitor rewrites the tree and re-exposes, so the builder action should be idempotent; (2) any future method tagged with this attribute inherits the same trust model.

Fix: Add a `<remarks>` paragraph to `EagerEvaluationAttribute` stating that marked methods may be invoked more than once per `IQueryable` materialisation and therefore must be pure / idempotent. Also worth noting the attribute is only observed by the in-box exposer — custom `IQueryProvider`s that don't route through `ExposeExpressionVisitor` won't desugar methods tagged with it.

— ✓ Fixed in 05915240

MaceWindu (changes requested):
## Review notes

- [x] PR has no milestone set. Per the Pull request rules, a milestone must be assigned; this PR expands the public `LinqToDB` namespace, so it needs a major-release milestone (e.g. `6.0.0`) to justify the surface expansion ΓÇö see **BLK001** below.
- [ ] API baselines need a refresh (run the `api-baselines` skill). The PR adds new internal surface under `LinqToDB.Internal.Infrastructure.*`, `LinqToDB.Internal.Linq.*`, `LinqToDB.Internal.Linq.Builder.*`, `LinqToDB.Internal.SqlQuery.*`, and `LinqToDB.Internal.SqlProvider.*`, but no `CompatibilitySuppressions.xml` changes are present.

### Blockers (BLK)

- [ ] **BLK001** ΓÇö Public API expansion without a major-release milestone
  Why: Per `.claude/docs/api-surface-classification.md`, non-`Internal.*` additions are allowed only under a major-release milestone (regex `^\d+\.0\.0(...)?$`). This PR adds five public symbols but has no milestone assigned:
  - `T:LinqToDB.ICteBuilder`
  - `M:LinqToDB.ICteBuilder.HasName(System.String)`
  - `T:LinqToDB.CteBuilderExtensions`
  - `M:LinqToDB.CteBuilderExtensions.IsMaterialized(LinqToDB.ICteBuilder,System.Boolean)`
  - `M:LinqToDB.LinqExtensions.AsCte``1(System.Linq.IQueryable{``0},System.Action{LinqToDB.ICteBuilder})`

  Fix: Assign a major-release milestone to the PR before merge (fetch via `gh api repos/linq2db/linq2db/milestones?state=open`). Once set, this BLK converts to the standard informational note per the classification policy.

### Suggestion (SUG)

- [ ] **SUG004** ΓÇö Test coverage gaps in `CteMaterializedTests`
  Why: The 9 new CTE tests cover happy-path emission, fallback, cache-key differentiation, and the `SelectMany` desugar. A few invariants that the author stated (in code comments and PR description) are not pinned by tests: (1) `AsCte(source, null)` should throw `ArgumentNullException` ΓÇö guard exists but is unverified; (2) `LinqServiceSerializer` round-trip for `CteClause.Annotations` is new in this PR but only exercised via provider-specific emission paths, not by a direct serialize/deserialize round-trip assertion; (3) `ICteBuilder.HasName(null)` / `HasName("")` edge cases aren't covered; (4) there's no negative test asserting recursive CTEs ignore the annotation (the comment in `BuildRecursiveCteContextTable` would benefit from a test anchor).

  Fix: Add short tests for each gap. None of these block merge, but they'd protect invariants the author already stated in comments.

## Baselines

Delta: [linq2db.baselines PR #1808](https://github.com/linq2db/linq2db.baselines/pull/1808) (OPEN) ┬╖ [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5478)

All baseline changes are new `.sql` fixtures for `CteMaterializedTests` and map cleanly onto the documented feature matrix (PG/SQLite/ClickHouse emit the hint; SQL Server variants silently ignore; ClickHouse drops `NOT MATERIALIZED`). No cross-provider anomalies; no drift in unrelated tests.

### CTE Materialization Hints ΓÇö New Tests, All Correct

Seven new test methods validating the `AsCte` builder with `IsMaterialized()` and `HasName()` overloads. All tests emit correct SQL syntax per provider capability.

- **MATERIALIZED hint emission (supporting providers)** ΓÇö PostgreSQL 13ΓÇô18, SQLite 3.35+, and ClickHouse 26.3+ correctly emit `AS MATERIALIZED` in CTE headers when `IsMaterialized(true)` is called.
  - [Tests.Linq.CteMaterializedTests.AsCte_Materialized_EmitsKeyword](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/ClickHouse.Driver/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Materialized_EmitsKeyword%28ClickHouse.Driver%29.sql) ΓÇö added (12 providers: ClickHouse.Driver, ClickHouse.MySql, PostgreSQL.13, PostgreSQL.14, PostgreSQL.15, PostgreSQL.16, PostgreSQL.17, PostgreSQL.18, SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS). Identical SQL structure across all 12 providers; syntactic variation (identifier quoting, parameter prefixes) is routine and expected.
    <details><summary>sample diff (ClickHouse.Driver)</summary>

    ```diff
    +-- ClickHouse.Driver ClickHouse
    +
    +WITH CTE_1 AS MATERIALIZED
    +(
    +	SELECT
    +		c_1.ParentID as ParentID
    +	FROM
    +		Child c_1
    +	WHERE
    +		c_1.ParentID > 1
    +)
    +SELECT
    +	p.ParentID,
    +	p.Value1
    +FROM
    +	Parent p
    +		INNER JOIN CTE_1 c_2 ON p.ParentID = c_2.ParentID
    ```
    </details>

- **NOT MATERIALIZED hint emission (supporting providers)** ΓÇö PostgreSQL 13ΓÇô18 and SQLite 3.35+ correctly emit `AS NOT MATERIALIZED` when `IsMaterialized(false)` is explicitly called.
  - [Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_EmitsKeyword](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/PostgreSQL.13/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_EmitsKeyword%28PostgreSQL.13%29.sql) ΓÇö added (10 providers: PostgreSQL.13, PostgreSQL.14, PostgreSQL.15, PostgreSQL.16, PostgreSQL.17, PostgreSQL.18, SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS).
    <details><summary>sample diff (PostgreSQL.13)</summary>

    ```diff
    +-- PostgreSQL.13 PostgreSQL
    +
    +WITH "CTE_1" ("ParentID")
    +AS NOT MATERIALIZED
    +(
    +	SELECT
    +		c_1."ParentID"
    +	FROM
    +		"Child" c_1
    +	WHERE
    +		c_1."ParentID" > 1
    +)
    +SELECT
    +	p."ParentID",
    +	p."Value1"
    +FROM
    +	"Parent" p
    +		INNER JOIN "CTE_1" c_2 ON p."ParentID" = c_2."ParentID"
    ```
    </details>

- **ClickHouse NOT MATERIALIZED fallback** ΓÇö ClickHouse correctly falls back to plain `AS` syntax (dropping the `NOT MATERIALIZED` keyword) as documented in the PR summary ΓÇö ClickHouse does not support `NOT MATERIALIZED`.
  - [Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_ClickHouse_FallsBackToPlainAs](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/ClickHouse.Driver/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_ClickHouse_FallsBackToPlainAs%28ClickHouse.Driver%29.sql) ΓÇö added (2 providers: ClickHouse.Driver, ClickHouse.MySql). Correctly emits plain `AS` with no `MATERIALIZED` keyword.
    <details><summary>sample diff (ClickHouse.Driver)</summary>

    ```diff
    +-- ClickHouse.Driver ClickHouse
    +
    +WITH CTE_1 AS
    +(
    +	SELECT
    +		c_1.ParentID as ParentID
    +	FROM
    +		Child c_1
    +	WHERE
    +		c_1.ParentID > 1
    +)
    +SELECT
    +	p.ParentID,
    +	p.Value1
    +FROM
    +	Parent p
    +		INNER JOIN CTE_1 c_2 ON p.ParentID = c_2.ParentID
    ```
    </details>

- **Unsupported providers (hint silently ignored)** ΓÇö SQL Server 2005ΓÇô2025 and all other unsupported providers correctly emit plain `AS` with no `MATERIALIZED` hint.
  - [Tests.Linq.CteMaterializedTests.AsCte_Materialized_UnsupportedProvider_IgnoredSilently](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SqlServer.2022.MS/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Materialized_UnsupportedProvider_IgnoredSilently%28SqlServer.2022.MS%29.sql) ΓÇö added (22 providers: SqlServer.2005, SqlServer.2005.MS, SqlServer.2008, SqlServer.2008.MS, SqlServer.2012, ΓÇª (22 providers total)). All SQL Server versions show identical behavior.
    <details><summary>sample diff (SqlServer.2022.MS)</summary>

    ```diff
    +-- SqlServer.2022.MS SqlServer.2022
    +
    +WITH [CTE_1] ([ParentID])
    +AS
    +(
    +	SELECT
    +		[c_1].[ParentID]
    +	FROM
    +		[Child] [c_1]
    +	WHERE
    +		[c_1].[ParentID] > 1
    +)
    +SELECT
    +	[p].[ParentID],
    +	[p].[Value1]
    +FROM
    +	[Parent] [p]
    +		INNER JOIN [CTE_1] [c_2] ON [p].[ParentID] = [c_2].[ParentID]
    ```
    </details>

- **Builder API methods (HasName, IsMaterialized overwrite, SelectMany)** ΓÇö Builder API tests validate fluent method chaining: `HasName()` sets a custom CTE name; `IsMaterialized(false)` after `IsMaterialized(true)` correctly emits `NOT MATERIALIZED`; the builder overload works inside `SelectMany`.
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_HasName_SetsCteName](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SQLite.Classic/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_HasName_SetsCteName%28SQLite.Classic%29.sql) ΓÇö added (4 providers: SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS). CTE name correctly changed from auto-generated `CTE_1` to `CustomName`.
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_IsMaterialized_Overwrite](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/PostgreSQL.13/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_IsMaterialized_Overwrite%28PostgreSQL.13%29.sql) ΓÇö added (10 providers: PostgreSQL.13, PostgreSQL.14, PostgreSQL.15, PostgreSQL.16, PostgreSQL.17, PostgreSQL.18, SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS). Validates that a final `IsMaterialized(false)` overrides earlier `IsMaterialized(true)`.
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_InsideSelectMany_Works](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SQLite.Classic/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_InsideSelectMany_Works%28SQLite.Classic%29.sql) ΓÇö added (4 providers: SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS). Builder-configured `MATERIALIZED` hint works within `SelectMany` join syntax.

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -686,8 +737,14 @@ protected virtual void BuildWithClause(SqlWithClause? with)
 				else
 					StringBuilder.Append(' ');
 
-				AppendIndent();
-				StringBuilder.AppendLine("AS");
+				BuildCteHeaderHint(cte);
+
+				if (!SuppressCteAsKeyword(cte))
```
MaceWindu, line 742: it results in two `ShouldEmitMaterializedCteHint` calls by default

Source/LinqToDB/Internal/Remote/LinqServiceSerializer.cs
```diff
@@ -1397,6 +1397,26 @@ void Visit(IQueryElement e, EvaluationContext evaluationContext)
 							Append(elem.Fields);
 							Append(elem.IsRecursive);
 
+							// Annotations bag: count, then (name, type?, value) triples.
+							// Type is null when value is null. Values are serialized via
+							// SerializationConverter — primitives and simple types survive.
+							var annotations = elem.Annotations.GetAnnotations().ToArray();
```
MaceWindu, line 1403: maybe add `GetUnorderedAnnotations` ?
- no need to sort on call
- no need to materialize to get size

MaceWindu (changes requested):
optional perf. related notes

Tests/Linq/Infrastructure/AnnotatableTests.cs
MaceWindu, line 1: **Suggestion · SUG005** — Residual test-coverage gaps from SUG004

Two of the four SUG004 coverage gaps are still not pinned by tests:

1. **`LinqServiceSerializer` direct round-trip for `CteClause.Annotations`.** The PR adds a bespoke serialization path at `Source/LinqToDB/Internal/Remote/LinqServiceSerializer.cs:1400–1419` (Visit) and `:2519–2527` (Parse): count prefix plus `(name, type?, value)` triples, with `SerializationConverter` for values and a nullable `Type` marker for the null-value case. No test currently round-trips a `CteClause` with annotations through `Serialize` → `Deserialize` to verify names/types/values survive intact — a regression here would silently break remote-context queries that use `IsMaterialized()`.
2. **Recursive-CTE annotation contract pin.** `Source/LinqToDB/Internal/Linq/Builder/TableBuilder.CteTableContext.cs:71–75` documents that `BuildRecursiveCteContextTable` intentionally drops annotations, but nothing pins the contract under test. If a future refactor wires the annotation bag into the recursive path the comment would be stale and no test would flag it.

Why: both gaps were flagged in SUG004 as "won't block merge but protect invariants the author already stated in comments"; the first two gaps (`AsCte(source, null)` + `HasName` edge cases) were addressed, these two remain.

Fix:
- Add a `LinqServiceSerializer` round-trip test under `Tests/Linq/Infrastructure/AnnotatableTests.cs` (or a new `LinqServiceSerializerCteTests.cs`) that builds a `CteClause` with a couple of annotations (e.g. one `bool`, one `string`, one `null`-value), serializes via `LinqServiceSerializer.Serialize`, deserializes with `Deserialize`, and asserts `Annotations.GetAnnotations()` survives with matching name / type / value.
- Add a negative-path recursive-CTE test asserting that the recursive `GetCte` overload does not accept or carry a `CteAnnotationsContainer` — simplest form: a compile-time check (no `AsCte(Action<ICteBuilder>)` overload exists for the recursive shape) plus a runtime assertion that a recursive CTE has no annotations on the emitted `CteClause`.

MaceWindu (commented):
## Verification update — 2026-04-20, against HEAD `05915240`

All 11 ID-tagged findings from the earlier review re-checked against the current PR head. Two informational top-bullet notes and per-line findings updated in-place (checkboxes flipped on the earlier review body; `— ✓ Fixed in 05915240` appended on the resolved line comments).

### Fixed
- [x] BLK001 — Public API expansion under non-major milestone (fixed by policy re-evaluation: current `api-surface-classification.md` → step 3 row 2 allows additive non-`Internal.*` surface under any milestone; all 5 new public symbols pass the step-4 sanity check)
- [x] MAJ001 — `CteAnnotationsContainer` no longer exposes mutable `Annotatable` by reference
- [x] MAJ002 — `IsMaterialized` defensive cast with `NotSupportedException`
- [x] MIN001 — `CteAnnotationsContainer.Empty` removed
- [x] MIN002 — `AsCte` producer side: snapshot hand-off
- [x] MIN003 — `AddAnnotations` overload routing consolidated
- [x] MIN004 — `[EagerEvaluation]` fail-fast when `!IsCompilable`
- [x] SUG001 — `CteBuilderImpl` members narrowed to `internal`
- [x] SUG002 — silent-no-op stance documented in XML
- [x] SUG003 — `EagerEvaluationAttribute` `<remarks>` on idempotence / in-box-exposer-only / fail-fast

### Partially / incorrectly fixed
- [~] SUG004 — Test coverage gaps (original: review #4139528003; follow-up: SUG005 below)
  - Covered now: `AsCte(source, null)` guard, `HasName(null)` / `HasName("")` edge cases.
  - Still missing: direct `LinqServiceSerializer` round-trip for `CteClause.Annotations`; recursive-CTE annotation contract pinning test.

### New findings

See **SUG005** below.

### Unresolved prior comments (no IDs — not in verify protocol)

The two comments from the earlier "optional perf. related notes" review (`#4142442610`) remain open and were not touched:

- `BasicSqlBuilder.cs:742` — *"it results in two `ShouldEmitMaterializedCteHint` calls by default"*
- `LinqServiceSerializer.cs:1403` — *"maybe add `GetUnorderedAnnotations`?"*

These are informal perf-tuning suggestions, not tracked findings. Mentioned here only for visibility.

## API surface (fresh classification at HEAD)

19 additive entries; no modifications or removals.

- **5 non-`Internal.*`** (additive public surface — all pass the step-4 "intentionally public" sanity check):
  - `T:LinqToDB.ICteBuilder`, `M:LinqToDB.ICteBuilder.HasName(System.String)`
  - `T:LinqToDB.CteBuilderExtensions`, `M:LinqToDB.CteBuilderExtensions.IsMaterialized(LinqToDB.ICteBuilder,System.Boolean)`
  - `M:LinqToDB.LinqExtensions.AsCte``1(System.Linq.IQueryable{``0},System.Action{LinqToDB.ICteBuilder})`
- **14 under `LinqToDB.Internal.*`** — annotations infrastructure (7), cache-key marker (1), eager-evaluation attribute (1), CTE builder plumbing (3), `CteAnnotationNames` (1), `CteClause.Annotations` property (1).

- [ ] API baselines need a refresh (run the `api-baselines` skill). No `CompatibilitySuppressions.xml` change in the PR diff; the additive surface still needs a regenerated suppressions file before merge.

## Baselines

Delta: [linq2db.baselines PR #1808](https://github.com/linq2db/linq2db.baselines/pull/1808) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5478)

Regrouped from scratch against `baselines/pr_5478` at HEAD. 64 new `.sql` fixtures across 7 test methods. No cross-provider anomalies, no drift in unrelated tests. Grouping matches the prior assessment cleanly.

### CTE Materialization Hints — All Correct

- **MATERIALIZED hint emission** — PostgreSQL 13–18, SQLite 3.35+, ClickHouse 26.3+ correctly emit `AS MATERIALIZED`.
  - [Tests.Linq.CteMaterializedTests.AsCte_Materialized_EmitsKeyword](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/PostgreSQL.13/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Materialized_EmitsKeyword%28PostgreSQL.13%29.sql) — added (12 providers)

- **NOT MATERIALIZED emission (PG/SQLite only)** — PostgreSQL 13–18 and SQLite correctly emit `AS NOT MATERIALIZED`.
  - [Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_EmitsKeyword](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/PostgreSQL.13/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_EmitsKeyword%28PostgreSQL.13%29.sql) — added (10 providers)

- **ClickHouse NOT MATERIALIZED fallback** — ClickHouse drops the keyword and emits plain `AS`.
  - [Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_ClickHouse_FallsBackToPlainAs](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/ClickHouse.Driver/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_NotMaterialized_ClickHouse_FallsBackToPlainAs%28ClickHouse.Driver%29.sql) — added (2 providers)

- **Silent ignore on unsupported providers** — all SQL Server variants emit plain `AS`.
  - [Tests.Linq.CteMaterializedTests.AsCte_Materialized_UnsupportedProvider_IgnoredSilently](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SqlServer.2022.MS/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Materialized_UnsupportedProvider_IgnoredSilently%28SqlServer.2022.MS%29.sql) — added (22 providers)

- **Builder pattern** — `HasName`, `IsMaterialized` overwrite, and usage inside `SelectMany` all emit the expected SQL.
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_HasName_SetsCteName](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SQLite.Classic/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_HasName_SetsCteName%28SQLite.Classic%29.sql) — added (4 providers)
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_IsMaterialized_Overwrite](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/PostgreSQL.13/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_IsMaterialized_Overwrite%28PostgreSQL.13%29.sql) — added (10 providers)
  - [Tests.Linq.CteMaterializedTests.AsCte_Builder_InsideSelectMany_Works](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5478/SQLite.Classic/Tests/Linq/CteMaterializedTests/Tests.Linq.CteMaterializedTests.AsCte_Builder_InsideSelectMany_Works%28SQLite.Classic%29.sql) — added (4 providers)

---

*Automated follow-up verification. Draft review — please submit manually after preview.*

MaceWindu (approved):
Approve per offline discussion

PR #5468, opened 2026-04-12
Window Functions: new Sql.Window API

sdanyliv:
## Summary

New `Sql.Window.*` API for window functions, replacing the old `Sql.Ext.*().Over().ToValue()` pattern. Old API is transparently converted to the new pipeline via `LegacyMemberConverterBase`.

### API Design

Lambda-based fluent builder with compile-time safety via interface chains:

```csharp
var query =
    from t in db.Employees
    let wnd = Sql.Window.DefineWindow(f => f.PartitionBy(t.Department).OrderBy(t.HireDate))
    select new
    {
        RN         = Sql.Window.RowNumber(f => f.PartitionBy(t.Department).OrderBy(t.HireDate)),
        RunningSum = Sql.Window.Sum(t.Salary, f => f.UseWindow(wnd)),
        NextSalary = Sql.Window.Lead(t.Salary, 1, 0m, f => f.UseWindow(wnd)),
        AvgActive  = Sql.Window.Average(t.Salary, f => f.Filter(t.IsActive).UseWindow(wnd)),
        MinFirst   = Sql.Window.Min(t.Salary, f => f.KeepFirst().OrderBy(t.HireDate).PartitionBy(t.Department)),
    };
```

### Backward Compatibility

Old `Sql.Ext.*` API is automatically converted to `Sql.Window.*` at the expression tree level via `LegacyMemberConverterBase`. Existing user code continues to work without changes, including KEEP chains.

### Functions Implemented

- **Ranking:** ROW_NUMBER, RANK, DENSE_RANK, PERCENT_RANK, CUME_DIST, NTILE
- **Offset:** LEAD, LAG (with offset and default overloads)
- **Value:** FIRST_VALUE, LAST_VALUE, NTH_VALUE
- **Aggregates:** COUNT (and `LongCount` → `long`), SUM, AVG, MIN, MAX (with frame support)
- **Statistical:** STDDEV (STDEV on SQL Server), STDDEV_POP, STDDEV_SAMP, VARIANCE, VAR_POP, VAR_SAMP
- **Regression / covariance:** COVAR_POP, COVAR_SAMP, CORR, REGR_SLOPE, REGR_INTERCEPT, REGR_COUNT, REGR_R2, REGR_AVGX, REGR_AVGY, REGR_SXX, REGR_SYY, REGR_SXY
- **Ordered-set:** PERCENTILE_CONT, PERCENTILE_DISC (both `WITHIN GROUP` aggregate and windowed `OVER` forms)
[...]

PR #5485, opened 2026-04-22
Firebird 6 support

### LINKED ISSUE #5483: Firebird 6: use native `DROP TABLE IF EXISTS` / `CREATE TABLE IF NOT EXISTS` instead of `EXECUTE BLOCK` workaround
### Summary

For the Firebird 6 dialect (`FirebirdVersion.v6`), linq2db should drop the legacy `EXECUTE BLOCK` + `rdb$relations` lookup workaround for `CreateIfNotExists` / `DropIfExists` in favor of the native `IF [NOT] EXISTS` clauses added in Firebird 6.

### Background

Firebird 6 introduces native `IF EXISTS` / `IF NOT EXISTS` support for `DROP` and `CREATE` statements across many DDL object types (`TABLE`, `VIEW`, `INDEX`, `TRIGGER`, `PROCEDURE`, `FUNCTION`, `SEQUENCE`/`GENERATOR`, `EXCEPTION`, `DOMAIN`, `ROLE`, `USER`, `COLLATION`, `PACKAGE`, `FILTER`, external functions). Tracking: [FirebirdSQL/firebird#8062](https://github.com/FirebirdSQL/firebird/issues/8062) (fix-version 6.0 Alpha 1, PR #8072).

### Current behavior

`FirebirdSqlBuilder` applies the same workaround to every dialect up to and including v5:

- `BuildDropTableStatement` — `Source/LinqToDB/Internal/DataProvider/Firebird/FirebirdSqlBuilder.cs:322`
- `BuildStartCreateTableStatement` / `BuildEndCreateTableStatement` — same file, `:474` / `:517`

Both wrap the DDL in `EXECUTE BLOCK AS BEGIN ... IF ((NOT )EXISTS(SELECT 1 FROM rdb$relations WHERE rdb$relation_name = '...')) THEN EXECUTE STATEMENT '...' END`.

### Expected behavior

On v6, emit:

```sql
DROP TABLE IF EXISTS "Foo"
CREATE TABLE IF NOT EXISTS "Foo" (...)
```

— no `EXECUTE BLOCK` / `EXECUTE STATEMENT` indirection, no `rdb$relations` probe.

### Notes

- Scope here is `CREATE/DROP TABLE`, matching the existing `TableOptions.CreateIfNotExists` / `DropIfExists` surface. The same pattern could later be extended to other DDL paths that currently emulate `IF [NOT] EXISTS`.
- Temporary-table handling (`IsTemporary`) shares the same branch in the current builder; care is needed to keep the GTT path working, since `CREATE GLOBAL TEMPORARY TABLE` still benefits from the native clause but the identity-field trigger/generator creation still needs the `EXECUTE BLOCK` wrapper.
- Full v6 support (provider-name constant, data-provider class, mapping schema, factory/detector routing, test groups) lands when Firebird 6 ships. This issue is scoped to the SQL-builder side.

MaceWindu:
## Summary

Full Firebird 6 provider support. The original scope of this PR (native DDL `IF [NOT] EXISTS`) was deliberately limited until Firebird 6 shipped a tagged container image; now that `firebirdsql/firebird:6-snapshot` exists, this branch completes the deferred work.

- **Provider wiring** — `ProviderName.Firebird6`, `FirebirdDataProvider6`, `FirebirdMappingSchema.Firebird6MappingSchema`, provider-detector routing (name/version dispatch, config-string heuristic, server-version auto-detect `major >= 6 → v6`), `FirebirdFactory` `"6"`, and the LINQPad provider UI entry.
- **Native `IF [NOT] EXISTS`** — `Firebird6SqlBuilder` flips a `SupportsNativeIfExists` toggle so CREATE/DROP TABLE (and the inner generator/trigger DDL for identity tables) emit `IF [NOT] EXISTS` instead of the legacy `EXECUTE BLOCK` + `rdb$` probe. Re-implemented onto master's refactored create-table builder; pre-FB6 paths unchanged.
- **SQL-standard schemas** — `Firebird6SqlBuilder.BuildObjectName` emits schema-qualified object names (suppressed for temporary tables), mirroring the Oracle/PostgreSQL builders.
- **Schema provider** — rewritten from ADO `GetSchema(...)` to LINQ over mapped `rdb$` system-table entity classes (the ClickHouse/DuckDB pattern), using INNER JOINs (no APPLY, so FB2.5/3 stay supported). FB6 reads `RDB$SCHEMA_NAME` to populate `SchemaName`/`IsDefaultSchema` (default `PUBLIC`); pre-FB6 keeps owner-as-schema for parity. Procedure metadata (UNION-ALL SQL) is unchanged.
- **SQL feature translation** — `Firebird6MemberTranslator` adds FB6 `GREATEST`/`LEAST` (math translator, replacing the `CASE WHEN` fallback used by earlier versions) and ordered string aggregation via `LIST(...) WITHIN GROUP (ORDER BY ...)` (`LIST` is an alias of the SQL-standard `LISTAGG` in FB6).
- **Docker** — `Data/Setup Scripts/firebird60.cmd` (local) and `Build/Azure/scripts/firebird6.sh` (CI) on `firebirdsql/firebird:6-snapshot` (jacobalberty has no FB6 image), using the firebirdsql env convention.
- **CI** — `Firebird6` test-matrix job (mirrors Firebird5, gated by `[firebird.all]`/`[all]`) + `Build/Azure/net{80,90,100}/firebird6.json`; connection strings in `DataProviders.json` + `UserDataProviders.json.template`.
- **Tests** — `TestProvName` v6 version ladder (`AllFirebird6Plus`/`AllFirebirdLess6`); `TestFb6NativeIfExists`, `TestFb6GreatestLeast`, `TestFb6ListAggWithinGroup`. Existing Math/StringJoin suites also exercise GREATEST/LEAST and ordered aggregation on FB6 via the ladder.

Fixes #5483.

## Verification

Verified against the live `firebirdsql/firebird:6-snapshot` container (Firebird 6.0):
- Connect + CREATE/DROP DDL smoke test ✔
- Schema-provider load on **FB5** (flat namespace, `SYSDBA`) and **FB6** (`PUBLIC` schema) — tables, columns + types, primary keys, and foreign keys all correct ✔
- `TestFb6NativeIfExists`, `TestFb6GreatestLeast`, `TestFb6ListAggWithinGroup` green ✔ (`GREATEST(...)`/`LEAST(...)`, `LIST(...) WITHIN GROUP (ORDER BY ...)`)
- Clean builds: Debug/Testing, Release (analyzers), netstandard2.0 (TFM availability) ✔

## FB6 features evaluated but intentionally not included here

Assessed from a `doc/` diff of FB `v5.0-release`→`master` plus empirical probing on the snapshot:

- **UUIDv7** (`GEN_UUID(7)`) → split into a cross-provider feature request: #5646 (`Sql.NewGuid7` + `Guid.CreateVersion7()`).
- **Server-side `DateTime.ToString(format)`** via `CAST … FORMAT` — not implemented: no existing provider translates arbitrary date format strings server-side, so FB6 wouldn't pioneer it.
- **Length-less `VARCHAR` default** — no functional gain (FB6's implicit default is also 255) and the length originates in the mapping schema (parameter-sizing impact); not worth the churn.
- **Aggregate-function schema detection** — FB6 supports `CREATE AGGREGATE FUNCTION`, but `RDB$FUNCTIONS` exposes no queryable aggregate indicator (an aggregate's row is identical to a scalar's), so the schema provider can't distinguish them.
- **`GROUPS` window frame / frame `EXCLUDE`, `PERCENTILE_CONT/DISC`** — deferred to the `Sql.Window` API work (#5468); FB6 supports `GROUPS BETWEEN`/`EXCLUDE` at the SQL level.

## Follow-ups (not in this PR)

- `PublicAPI.Unshipped` reconciliation via `/release-publicapi` — the PublicApiAnalyzers gate is opt-in and the branch base is currently out of sync with it (unrelated DuckDB / Firebird-translator entries), so it is reconciled at release time across the whole backlog rather than partially here.
- Full cross-version (FB 2.5 / 3 / 4) schema-provider regression + baselines run via CI.

## Follow-up commit — window functions + UUIDv7 + charset-independent Guid reads

Now that #5468 (`Sql.Window`) and #5648 (`Sql.NewGuid7`/`Guid.CreateVersion7()`) are merged, the deferred FB6 items from *"features evaluated but intentionally not included"* are completed on top of them:

- **Window functions** — `Firebird6WindowFunctionsMemberTranslator` enables the `GROUPS` frame unit, frame `EXCLUDE`, and `PERCENTILE_CONT/DISC` (`WITHIN GROUP` group form). The windowed `OVER` percentile form stays unsupported (Firebird has no such form). Percentile group-form window tests un-gated for FB6; `WindowFunctions.FeatureMatrix.md` documents the FB6 deltas. Verified 28+ window tests green on FB6.
- **UUIDv7** — `Firebird6MemberTranslator.TranslateNewGuid7Method` emits `GEN_UUID(7)`, so `Sql.NewGuid7()` / `Guid.CreateVersion7()` produce a server-side UUIDv7 on FB6 (the FB6 half of #5646, deferred until this provider landed). `NewGuid7`/`CreateVersion7` verified on FB6.
- **Charset-independent binary Guid reads** — binary `CHAR(16) OCTETS` Guid values (columns, `GEN_UUID`, literals) are read via `UUID_TO_CHAR` in the SELECT projection, returning canonical text instead of raw OCTETS bytes. Reading the raw bytes transliterates them to the connection/database charset and raises *"Malformed string"* on any non-ASCII byte over a non-`NONE` charset (e.g. FB6's UTF8-default database). This is **DB-encoding-driven, not version-specific**, so it applies to all Firebird versions; textual `CHAR(38)` Guids and predicate/write positions are untouched.
- **Consolidated Guid test set** (`FirebirdTests`) — binary vs textual × literal / parameter / generated / column round-trip.

**Known limitation (below linq2db):** binding a binary Guid/`byte[]` **parameter** to an OCTETS column fails *"Malformed string"* on a non-`NONE`-charset database (`Version12.GdsStatement.Execute`, FbClient 10.3.4; FbClient 11 fixes it but drops net462, which linq2db still targets). The affected round-trip **write** tests are gated for FB6 via `[ActiveIssue]`. The limitation is DB-encoding-driven — a UTF8-charset FB 3/4/5 database would hit it too — so it is a client/server matter to resolve upstream, not in linq2db.

## Follow-up commit - MERGE/INSERT Guid writes, FB6 schemas, prerelease gates

Builds on the charset-independent Guid read-wrap above. That wrap (`UUID_TO_CHAR`) is only correct for values fetched to the client; when a Guid column read feeds a server-side write target it must stay binary.

- **Guid read-wrap no longer corrupts writes** - the `UUID_TO_CHAR` wrap is now suppressed inside modification statements (`Firebird6SqlExpressionConvertVisitor`). Previously a Guid column read inside a `MERGE` source, `INSERT ... SELECT`, or `UPDATE ... SET` produced a `CHAR(36)` value written back into a `BINARY(16)` column -> "string right truncation, expected length 16, actual 36". Fixes the FB6 `MERGE` / `INSERT` Guid round-trips.
- **Schema-qualified names quoted correctly** - `ConvertType.NameToSchema` was missing from `FirebirdSqlBuilder.Convert`'s quoting group, so a mixed-case schema was emitted unquoted and folded to upper case (`TESTSCHEMA`) while the table stayed quoted -> "Schema not found". Now quoted consistently (FB6-only; earlier versions never emit a schema qualifier). The `"TestSchema"` schema is created for FB6 in the Firebird create script (mirrors SqlServer / DuckDB / PostgreSQL).
- **Prerelease-blocked tests gated with a version note** - remaining FB6 failures that are FbClient 10.3.4 / Firebird 6.0.0.2068 prerelease limitations (not linq2db bugs) are `[ActiveIssue]`-gated with a note anchored to that build, to be revisited when a newer FB6 / client ships: native `VARBINARY` read truncation (`TestBinary`), OCTETS Guid "Malformed string" (`ClientConversion_*`), server "internal error" reading `AllTypes` (`TestDataTypes`), a boolean IN/Contains subquery wrong-result (`Test_FieldInSubquery`), and the INSERT-setter-subquery Guid write (`TestInsert`, where the read-wrap is applied to a subquery converted out of statement context so the modification-statement guard cannot reach it).

**Not releasable as-is** - the prerelease limitations above block a clean FB6 run, so this PR is moved to the **6.5.0** milestone pending newer Firebird 6 builds / FbClient fixes.


Source/LinqToDB/Internal/DataProvider/Firebird/FirebirdDataProvider.cs
```diff
@@ -101,7 +101,9 @@ public override ISqlBuilder CreateSqlBuilder(MappingSchema mappingSchema, DataOp
 			return Version switch
 			{
 				FirebirdVersion.v3    => new Firebird3SqlBuilder(this, mappingSchema, dataOptions, GetSqlOptimizer(dataOptions), SqlProviderFlags),
-				>= FirebirdVersion.v4 => new Firebird4SqlBuilder(this, mappingSchema, dataOptions, GetSqlOptimizer(dataOptions), SqlProviderFlags),
+				FirebirdVersion.v4    => new Firebird4SqlBuilder(this, mappingSchema, dataOptions, GetSqlOptimizer(dataOptions), SqlProviderFlags),
+				FirebirdVersion.v5    => new Firebird4SqlBuilder(this, mappingSchema, dataOptions, GetSqlOptimizer(dataOptions), SqlProviderFlags),
```
MaceWindu, line 130:
```suggestion
				FirebirdVersion.v4 or FirebirdVersion.v5 => new Firebird4SqlBuilder(this, mappingSchema, dataOptions, GetSqlOptimizer(dataOptions), SqlProviderFlags),
```

PR #5493, opened 2026-04-27
Duplicate parameter optimization

igor-tkachev:
## Summary

This PR introduces the opt-in `OptimizeDuplicateParameters` option for reusing duplicate SQL parameters generated from repeated LINQ parameter access expressions.

`OptimizeDuplicateParameters` defaults to `false`, so existing behavior is preserved unless the option is explicitly enabled.

A secondary `OptimizeDuplicatePropertyParameters` option was added to control property-access reuse within `OptimizeDuplicateParameters`. It defaults to `true`, preserving the practical default behavior for captured object properties while allowing users to disable property-based reuse if they need stricter handling of property getters.

## Changes

- Added `OptimizeDuplicateParameters` to LINQ options and data options APIs.
- Added `OptimizeDuplicatePropertyParameters` to LINQ options, data options APIs, global configuration, and configuration identity.
- Added duplicate SQL parameter reuse when the same stable LINQ parameter access expression is used multiple times.
- Restricted optimized reuse to stable member-access chains rooted in captured constants/closure objects.
- Made property member access configurable through `OptimizeDuplicatePropertyParameters`; when disabled, optimized reuse still allows field-based captured parameter access but rejects property access.
- Prevented optimized reuse for non-stable expressions such as method calls, constructor calls, invocations, array/index access, and other non-member-chain shapes.
- Passed `LinqOptions` into parameter registration instead of threading individual optimization flags through the call chain.
- Kept the duplicate parameter scan to a single pass over registered parameter entries.
- Added a set of `OptimizeDuplicateParameters_XXX` tests covering enabled/disabled behavior, safe reuse cases, configurable property reuse, incompatible parameter cases, and non-pure expression cases.


sdanyliv:
Actually we do all possible deduplications allready. New optimizations may introduce "parameter sniffing".
Activate your option by default and check for regressions.

sdanyliv (dismissed):
Actually we do all possible deduplications allready. New optimizations may introduce "parameter sniffing".
Activate your option by default and check for regressions.

igor-tkachev:
> Actually we do all possible deduplications allready. New optimizations may introduce "parameter sniffing". Activate your option by default and check for regressions.

No, we do not. Just look at the tests. Also I have queries that uses a C# DateTime variable several times and I see same number of duplicates in SQL. The option by default disables this optimization.

Parameter sniffing can happen with duplicate parameters too (especially on pre-2019 SQL Server), so duplicates are not a reliable safeguard by themselves.

Also our priority is predictable C# -> SQL correspondence: what is written in C# should be reflected in SQL, without artificial duplicate parameters.

When sniffing does become an issue, it is usually easier to fix it directly with `OPTION(RECOMPILE)` or parameter inlining.

Tests/Linq/Linq/ParameterTests.cs
```diff
@@ -601,6 +602,92 @@ public class ParameterDeduplication
...
+			using (var db = GetDataContext(context, o => o.UseOptimizeDuplicateParameters(true)))
+			{
+				var query = db.GetTable<Table404Two>()
+					.Where(v => allUsages || v.Usage == usage.GetValueOrDefault());
+
+				sqlWithOptimization = query.ToSqlQuery();
+				Debug.WriteLine(query.ToSqlQuery().Sql);
+			}
+
+			sqlWithoutOptimization.Parameters.Count.ShouldBe(1);
```
jods4: I am not sure what this test is supposed to demonstrate: both queries have a single parameter named `@usage`, so what's the difference highlighted here?

Why check for `ShouldNotContain("@p")` when you check that count is one and `@usage` is present?

Maybe add a comment explaining the purpose of the test?

Tests/Linq/Linq/ParameterTests.cs
```diff
@@ -601,6 +602,92 @@ public class ParameterDeduplication
 			};
 		}
 
+		[Test]
+		public void OptimizeDuplicateParameters_ReusesSameLinqParameter([IncludeDataSources(TestProvName.AllSqlServer)] string context)
```
viceroypenguin: I would like to see this test with a `string` and possibly a small array as well.

Tests/Linq/Linq/ParameterTests.cs
igor-tkachev: Here we are not testing duplicate count behavior,  we are testing parameter name consistency across modes (`UseOptimizeDuplicateParameters(false/true)`) for the same query shape.
The `ShouldNotContain("@p")` assertion is an explicit guard against fallback-name regression, not a duplicate-count check.
I’ll add a comment in the test to make this intent clear.

Source/LinqToDB/Internal/Linq/ExpressionCacheManager.cs
```diff
@@ -232,16 +248,31 @@ void EnsureEvaluated(ParameterCacheEntry localEntry, Expression expr)
...
 			finalParameterId = paramEntry.ParameterId;
 
+			static bool CanReuseDuplicateParameter(ParameterCacheEntry paramEntry, Expression paramExpression, Expression testedExprExpression, ParameterCacheEntry testedEntry)
+			{
+				return
+					paramEntry.ItemAccessor  == null                                                                                        &&
+					testedEntry.ItemAccessor == null                                                                                        &&
+					paramExpression.Type.UnwrapNullableType() == testedExprExpression.Type.UnwrapNullableType()                             &&
+					testedEntry.DbDataType.EqualsDbOnly(paramEntry.DbDataType)                                                              &&
+					ExpressionEqualityComparer.Instance.Equals(testedEntry.ClientValueGetter,         paramEntry.ClientValueGetter)         &&
```
jods4: Some expressions can be not pure and return different values for the same invocation.

It'd be cool to remove duplicates when the same local variable is referenced multiple times, but we should be careful to not remove multiple calls to `Guid.NewGuid()` or `Random.Next()` for example, even though the expression tree is identical across repetitions.

Likewise some properties or method calls may have side-effects that user expects during evaluation, even something as simple as like `i++`.

It would be good to have some tests cases that demonstrate these cases are not de-duplicated.

igor-tkachev: done

igor-tkachev:
@codex review

Use the strongest reasoning model available.

Deep review PR https://github.com/linq2db/linq2db/pull/5493 code changes. Focus on:

correctness risks
DRY violations / duplicated logic
weak parts of the implementation
detailed analysis of the newly added strategies
what additional strategies could be useful
possible improvements
potential problems, including performance regressions
Please review the whole PR change set, not only the latest commit.
Analyze changes against the PR base branch.

Add inline comments on suspicious, risky, weak, or problematic code locations with concrete explanations and suggested improvements.

Source/LinqToDB/Internal/Linq/ExpressionCacheManager.cs
```diff
@@ -232,16 +235,40 @@ void EnsureEvaluated(ParameterCacheEntry localEntry, Expression expr)
...
+
+				while (expression is MemberExpression next)
+				{
+					if (next.Expression == null)
+						return false;
+
+					expression = next.Expression.UnwrapConvert();
+				}
+
+				return expression.NodeType == ExpressionType.Constant;
```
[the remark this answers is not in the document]
jods4: It is very unlikely but it's true that usage of an hypothetical `scope.NextUid` value in query should not be deduped.

If we only look at **one** level of member access `Constant.MemberAccess` then it's certain that this is either a captured variable, or maybe the member of a primitive type like `"abc".Length`, which should be safe.

With this, users that want to dedupe larger expression could do so by putting them in local variables before query.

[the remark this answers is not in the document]
igor-tkachev: I will add another Option to control it. I can just remove properties from here, but I use this case a lot.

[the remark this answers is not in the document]
igor-tkachev: Also we can make empty method `Sql.AsParameter()` which converts it to separate parameter.

viceroypenguin (dismissed):

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -445,6 +445,35 @@ public static bool ParameterizeTakeSkip
...
+						Options = Options with { OptimizeDuplicateParameters = value };
+				}
+			}
+
+			/// <summary>
+			/// Allows <see cref="OptimizeDuplicateParameters"/> to reuse duplicate SQL parameters for repeated captured object property access expressions.
+			/// This option has effect only when <see cref="OptimizeDuplicateParameters"/> is enabled.
+			/// Default value: <see langword="true"/>.
+			/// </summary>
+			public static bool OptimizeDuplicatePropertyParameters
```
MaceWindu, line 467: What the reasoning behind addding new option?
If optimization works - it should be just added ungated. If it doesn't work - option will not fix it.

Tests/Linq/Linq/ParameterTests.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
MaceWindu: not needed

MaceWindu (commented):
_

Source/LinqToDB/Common/Configuration.cs
MaceWindu: On the option's purpose: a per-member purity attribute would be a more precise alternative to a global `OptimizeDuplicatePropertyParameters` switch — a linq2db-owned `[Pure]` / `[Impure]` on a property / method, defaulting properties to pure (reusable) and methods and other members to impure, overridable per member. That tunes reuse without a context-wide option and could supersede `OptimizeDuplicatePropertyParameters`. Existing attributes do not fit: `JetBrains.Annotations.[Pure]` is not a linq2db runtime dependency, and `Sql.*.IsPure` (`SqlFlags.IsPure`) is a server-side-function optimizer flag, not a client-member hint.

MaceWindu (commented):
🤖 Agentic review — findings are suggestions; verify before acting.

**Summary.** Opt-in `OptimizeDuplicateParameters` (default off) reuses one SQL parameter for repeated *stable* captured-value accesses; `OptimizeDuplicatePropertyParameters` (default true) sub-gates property getters. Implemented in `ExpressionCacheManager.RegisterParameterEntry` behind an `IsStableParameterAccess` gate.

## What the option deduplicates (when enabled)

| Case | Result |
|---|---|
| Same captured local, same DB type (`t.Int1 == v \|\| t.Int2 == v`) | 1 param |
| Same captured field (`h.ValueField` twice) | 1 (always) |
| Same captured property (`h.ValueProperty` twice) | 1 (only while `OptimizeDuplicatePropertyParameters` = true) |
| Nullable / non-nullable of same underlying type | 1 param |

Not deduplicated (safeguards): method calls (`Guid.NewGuid()`, `Random.Next()`, mutating getters) — rejected by the stability gate (must be a member chain rooted in a `ConstantExpression`); array / indexer access; different variables even at equal values; different runtime values; different DB types (full `DbDataType.Equals`); parameter-order-dependent providers (Access, SAP HANA, native Informix) re-expand to N params downstream (binding preserved). ClickHouse inlines values, so the option is a no-op there.

## Risks (with maintainer assessment)

- **Property-getter side effects.** With property reuse on by default, a non-idempotent getter used twice is evaluated once. *Acceptable* — linq2db never guaranteed getter invocation count; an unstable getter is the caller's contract. Preferred long-term direction discussed on the option thread (per-member purity attribute).
- **Query plan / parameter sniffing.** *Not a real risk* — sniffing is value / plan-cache behavior, unchanged by whether a value arrives as one parameter or two; collapsing `col1 = @a OR col2 = @b` (equal values) to `col1 = @p OR col2 = @p` does not change per-branch selectivity. Parameter count ≠ plan caching.
- **Default-behavior correctness.** No regression — the option-on match set is a strict subset of the default match set plus a stability gate, so enabling it cannot merge a pair the engine would not already merge by value.

## Test coverage note

No test isolates the flag on a single query (same query, only the option flipped, 2 → 1); the one A/B test asserts 1 == 1. Consider one A/B case on a same-typed-column query to lock the headline behavior.

## Applied fixes (follow-up commit a3471161)

- Removed unused `using System.Diagnostics;` in `ParameterTests.cs`.
- `OptimizeDuplicateParameters_DefaultKeepsCurrentBehavior` no longer overrides the option (`GetDataContext(context)`), so it exercises the shipped default.

Source/LinqToDB/Common/Configuration.cs
igor-tkachev: The main reason for the option is compatibility policy for features that affect generated SQL or observable query behavior.

Even if optimization is logically correct, it still changes SQL shape: parameter count, parameter names, and sometimes provider-specific execution/binding behavior. We cannot guarantee that every user scenario, provider combination, SQL snapshot test, query-plan workaround, or diagnostic workflow will accept this change without problems.

For such kind of feature, I think staged rollout should be used:

1. Add the feature behind opt-in option, disabled by default.
2. Let users enable it explicitly and report real-world issues.
3. After few versions, if it proves stable, consider enabling it by default.
4. Only after it was validated in default-on mode for some time, consider removing the option as proven implementation detail.

Without an option, if we introduce regression, users have no practical way to disable it until we ship a fix. With an option, feature can be tested in real production scenarios without forcing new behavior on everyone immediately.

I also think this approach should become mandatory for all features that change behavior or generated SQL in any observable way. If we want linq2db to be reliable professional tool, compatibility cannot be treated as afterthought. SQL-shaping changes should be introduced in a way that gives users control, time to validate, and safe fallback path.

igor-tkachev:
@MaceWindu 

The primary use case for this PR is simpler and more practical: avoid generating many duplicate SQL parameters when the same captured value is reused across a complex query.

For example, in our code we can use the same COB date value in multiple `WHERE` predicates across complex JOIN queries, but generated SQL ends up with many parameters like `@COBDate1`, `@COBDate2`, `@COBDate3`, `@COBDate4`, etc. They all represent the same logical input value, but they make generated SQL noisier and harder to read/analyze.

Per-member purity metadata could be a more precise model for property/method reuse, and I think it is worth discussing. But it is a separate feature with its own API design, naming, documentation, and tests. I would not mix it into this PR.

This PR should stay focused on duplicate parameter reuse and compatibility-safe rollout through an opt-in option.

igor-tkachev:
I checked the requested A/B test, but the simple suggested shape is not valid for current behavior.

For:

```csharp
.Where(t => t.String2 == value || t.String3 == value)
```

SQL Server already produces one parameter even with `UseOptimizeDuplicateParameters(false)`. The same happened with a close `Int1 / IntN1` shape. So the expected `disabled: 2, enabled: 1` assertion would be incorrect for the current baseline.

The reason is that existing exact-expression deduplication in `ExpressionCacheManager.RegisterParameterEntry` already handles this simple case independently of the new optimized path.

igor-tkachev (commented):
@sdanyliv, I think the concerns from the original review are addressed now.

The optimization is still disabled by default intentionally. This is not because it is known to be broken, but because it changes generated SQL shape and observable query behavior. For this kind of change I prefer staged rollout: opt-in first, real-world validation, then possibly enable by default in a later version.

The technical risks were also narrowed since the original review:

- reuse is restricted to supported captured member access patterns;
- method calls and non-stable expressions like `Guid.NewGuid()` / `Random.Next()` are not reused;
- property reuse is controlled separately by `OptimizeDuplicatePropertyParameters`;
- incompatible parameter DB types are not reused;
- provider-specific parameter behavior remains covered by tests.

Regarding parameter sniffing: this optimization does not change the parameter value passed to the server. It can reduce duplicate parameters that represent the same logical input value, but sniffing is still based on parameter values and plan-cache behavior, not on whether the same value is represented by one parameter or several duplicate parameters.

The main target case is complex queries where the same input value, e.g. COB date, is used across multiple predicates/JOINs and currently can appear as many generated parameters like `@COBDate1`, `@COBDate2`, `@COBDate3`, etc. This PR keeps that behavior opt-in for compatibility-safe rollout.

MaceWindu (changes requested):
on hold; there is a no go from some mainteners for this feature

Issue #5125, opened 2025-09-10
v6 - error with interpolated string expression


detoxhby:
### Description

Using interpolated string causes invalid table reference in generated SQL

'42P01: missing FROM-clause entry for table "x_1"

### Reproduction Steps

```csharp
[Table("Test", IsColumnAttributeRequired = false)]
public class Test
{
    public int Id { get; set; }

    public string? StrVal { get; set; }

    public int IntVal { get; set; }

    [AllowNull]
    [ExpressionMethod(nameof(Expr1Impl))]
    public string Expr1 { get; set; }

    private static Expression<Func<Test, IDataContext, string>> Expr1Impl() =>
        (e, ctx) => !String.IsNullOrEmpty(e.StrVal) ? e.StrVal : e.IntVal.ToString();

    [AllowNull]
    [ExpressionMethod(nameof(Expr2Impl))]
    public string Expr2 { get; set; }

    private static Expression<Func<Test, IDataContext, string>> Expr2Impl() =>
        (e, ctx) => 
            // WITHOUT INTERPOLATION WORKS: !String.IsNullOrEmpty(e.StrVal) ? e.StrVal : e.IntVal.ToString();
            $"{(!String.IsNullOrEmpty(e.StrVal) ? e.StrVal : e.IntVal.ToString())}";
}

...

var table = connection.GetTable<Test>();
var query = table
    .OrderBy(x => x.Expr1)
    .Select(x => new Test()
    {
        Id = x.Id,
        Expr1 = x.Expr1,
        Expr2 = x.Expr2
    });
var list = await query.ToListAsync();
```

which generates the following SQL:
```sql
SELECT
	x_1."Id",
	CASE
		WHEN NOT (x_1."StrVal" IS NULL OR x_1."Length_1" = 0 AND x_1."Length_1" IS NOT NULL)
			THEN x_1."StrVal"
		ELSE x_1."IntVal"::text
	END,
	Coalesce(CASE
		WHEN NOT (x_1."StrVal" IS NULL OR x_1."Length_1" = 0 AND x_1."Length_1" IS NOT NULL)
			THEN x_1."StrVal"
		ELSE x_1."IntVal"::text
	END, '')
FROM
	(
		SELECT
			x."Id",
			x."StrVal",
			Length(x_1."StrVal") as "Length_1",
			x."IntVal"
		FROM
			"Test" x
	) x_1
ORDER BY
	CASE
[...]

PR #5508, opened 2026-05-05
Add regression test for #5125

sdanyliv:
Regression test for the original repro from #5125: non-nullable `string` + `[AllowNull]` + `[ExpressionMethod]` projected via `new T { ... }` after `OrderBy`. Reproduces on v6.0.0-rc.3 with `42P01: missing FROM-clause entry for table "x_1"` on PostgreSQL; passes on master, locking in the fix from #5126.

The v6.2.x regression mentioned in a later comment (`END NULLS FIRST as c1` syntax error 42601) was NOT reproducible — tried 7 query shape variants (`Take` / `Skip` / `Distinct` / `AsSubQuery` / eager-load via inner `Select` / `GroupBy` with inner `OrderBy` / `OrderBy.Select.OrderBy`) on v6.2.1 and v6.0.0-rc.3 against the same model. None triggered that SQL shape. Source-grep confirms `END NULLS FIRST` is only emitted from `Sql.AnalyticFunctions.OrderItemBuilder` (window-function builder), not from regular `ORDER BY`, so the commenter's repro likely involves explicit window/analytic functions or another path my variants don't reach. A separate test is needed once that shape is identified.

Refs #5125.

### DIFF
--- a/Tests/Linq/UserTests/Issue5125Tests.cs
+++ b/Tests/Linq/UserTests/Issue5125Tests.cs
@@ -30,7 +30,7 @@ namespace Tests.UserTests
 			public string Expr1 { get; set; } = null!;
 
 			private static Expression<Func<InterpolatedTest5125, IDataContext, string>> Expr1Impl() =>
-				(e, ctx) => !string.IsNullOrEmpty(e.StrVal) ? e.StrVal : e.IntVal.ToString();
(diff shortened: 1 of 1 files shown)


MaceWindu:
This PR doesn't add anything new - we need a real test case which demonstrates issue from issue reporter.

PR #5510, opened 2026-05-05
Fix interface member access

igor-tkachev:
## Summary

Fixes translation of interface-declared member access when it appears inside grouped aggregate projections.

In the failing cases, expressions such as:

```csharp
group => group.Max(i => ((IGenericParentable)i).Value1)
```

or generic-constrained equivalents were normalized to interface member access. The builder then tried to translate the interface member directly and failed before resolving it back to the mapped entity member.

The fix rewrites interface-declared member access to the concrete entity member when the current context can be traced back to a table/CTE context implementing that interface.

## Changes

- Update `ExpressionBuildVisitor` to resolve interface member access through the concrete mapped entity type before it can become a `SqlErrorExpression`.
- Keep the rewrite constrained to interface-declared members and table/CTE-backed contexts implementing the interface.
- Add regression coverage for:
  - interface-typed `IGrouping` projection with `Max`
  - generic-constrained `IGrouping` projection with `Max`

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
+++ b/Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
@@ -509,7 +509,7 @@ namespace LinqToDB.Internal.DataProvider.ClickHouse
 		{
 			if (table.SqlQueryExtensions is not null)
 			{
-				BuildTableExtensions(StringBuilder, table, alias, null, ", ", null,
+				BuildTableExtensions(StringBuilder, table, alias, " ", ", ", null,
 					ext => ext.Scope is Sql.QueryExtensionScope.TableHint or Sql.QueryExtensionScope.TablesInScopeHint);
 			}
 		}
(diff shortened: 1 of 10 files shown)


igor-tkachev:
@sdanyliv, need your expert opinion.

igor-tkachev:
@code review

sdanyliv:
Fix moved into `ExpressionBuilder.Project`. The `Convert` handler used to recurse with the operand even when it was the sequence root (`ContextRef`), which dropped member info and returned `body` unchanged — that was the parent-projection result the previous downstream guard had to filter.

Now `Project` recurses only when the operand is a member chain; otherwise it makes the path projection. Member resolution then rewrites the interface member to the concrete-entity member directly. As a result, the `ExpressionBuildVisitor` and `ReflectionExtensions` changes from earlier commits were reverted; only `ExpressionBuilder.SqlBuilder.cs` and the regression tests remain vs `master`.

MaceWindu (approved):

MaceWindu (commented):
we probably should remove test default provider feature - it produce no baselines. or add baselines support to it

Issue #1831, opened 2019-07-30
IDataConnection Interface 


tb-mtg:
Would it be possible to add an interface called IDataConnection (shown below) that LinqToDB.Data.DataConnection could implement.

It would be helpful when using custom data providers.

```
  public interface IDataConnection : ICloneable, IDataContext, IEntityServices, IDisposable {
    event Action<DataConnection, IDbConnection> OnBeforeConnectionOpen;
    event Func<DataConnection, IDbConnection, CancellationToken, Task> OnBeforeConnectionOpenAsync;
    event EventHandler OnClosed;
    event Action<DataConnection, IDbConnection> OnConnectionOpened;
    event Func<DataConnection, IDbConnection, CancellationToken, Task> OnConnectionOpenedAsync;

    IDbCommand Command { get; set; }
    int CommandTimeout { get; set; }
    string ConfigurationString { get; }
    IDbConnection Connection { get; }
    string ConnectionString { get; }
    IDataProvider DataProvider { get; }
    bool Disposed { get; }
    int ID { get; }
    bool IsMarsEnabled { get; set; }
    Action<TraceInfo> OnTraceConnection { get; set; }
    IRetryPolicy RetryPolicy { get; set; }
    bool? ThrowOnDisposed { get; set; }
    IDbTransaction Transaction { get; }
    DataConnection AddMappingSchema(MappingSchema mappingSchema);
    DataConnectionTransaction BeginTransaction(IsolationLevel isolationLevel);
    DataConnectionTransaction BeginTransaction();
    Task<DataConnectionTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default);
    Task<DataConnectionTransaction> BeginTransactionAsync(IsolationLevel isolationLevel, CancellationToken cancellationToken = default);
    void CheckAndThrowOnDisposed();
    Task CloseAsync(CancellationToken cancellationToken = default);
    void CommitTransaction();
    Task CommitTransactionAsync(CancellationToken cancellationToken = default);
    IDbCommand CreateCommand();
    Task DisposeAsync(CancellationToken cancellationToken = default);
    void DisposeCommand();
    Task EnsureConnectionAsync(CancellationToken cancellationToken = default);
    ITable<T> GetTable<T>(object instance, MethodInfo methodInfo, params object[] parameters) where T : class;
    ITable<T> GetTable<T>() where T : class;
    SqlStatement ProcessQuery(SqlStatement statement);
    void RollbackTransaction();
[...]

PR #5511, opened 2026-05-05
Fix #5355: Resolve association via realized concrete type for interface/abstract-base member access

### LINKED ISSUE #5355: Collection Contains() cannot be translated
### Description

Linq2Db can translate neither `string[].Contains()` nor `IEnumerable<T>.Contains()`. 

Specifying the ForeignKey by doing
```cs
public interface IProfile
{
    [ForeignKey(nameof(ProfileId))]
    LicenseProfile Profile { get; set; }
    int ProfileId { get; set; }
}
```
does not work or help. And interfaces cannot have a fluent API definition. 

I suspect, that whatever is analyzing the generic, may need to also check its recursive parent class to ensure the Profile is a Navigation Property / Foreign Key?

Notes:
- FTS Contains() (which is needed for our use-case) is necessary to replicate, as Linq2DB seems to leave the translation to EFCore otherwise 
- IProfile and FilterLicense is necessary to replicate, as the root cause of the issue. (Profiles are our Tenant)
- The inheritance is necessary to replicate. We have two versions of the same class, but many many columns are the same. For replication purpose only the one implementation is necessary to show

### Reproduction Steps

```xml
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net10.0</TargetFramework>
        <RootNamespace>l2db_borked</RootNamespace>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>
    </PropertyGroup>

    <ItemGroup>
      <PackageReference Include="linq2db.EntityFrameworkCore" Version="10.1.0" />
      <PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="10.0.3">
        <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
        <PrivateAssets>all</PrivateAssets>
      </PackageReference>
      <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="10.0.3" />
    </ItemGroup>

</Project>
```

```cs
using LinqToDB;
using LinqToDB.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore;

LinqToDBForEFTools.Initialize();

await using var db = new AppDbContext();
await db.Database.EnsureDeletedAsync();
await db.Database.EnsureCreatedAsync();
db.Database.ExecuteSqlRaw("""
    CREATE FULLTEXT CATALOG ftCatalog AS DEFAULT;
    CREATE FULLTEXT INDEX ON Customers(Name) KEY INDEX PK_Customers;
    """);

var profile = new LicenseProfile { License = "12345" };
db.LicenseProfiles.Add(profile);
await db.SaveChangesAsync();

db.Customers.Add(new Customer { Name = "Alice Smith", Profile = profile });
await db.SaveChangesAsync();

string[] licenseFilter = ["12345"];
var query = db.Customers
    .ToLinqToDBTable()
    .FilterLicense(licenseFilter)
    .Where(x => Sql.Ext.SqlServer().Contains("Alice", x.Name));
_ = await query.ToListAsyncLinqToDB();







public class AppDbContext : DbContext
{
    public DbSet<LicenseProfile> LicenseProfiles => Set<LicenseProfile>();
    public DbSet<Customer> Customers => Set<Customer>();

    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
    {
        optionsBuilder.UseSqlServer(
            "Server=localhost;Database=L2dbBorked;Trusted_Connection=True;TrustServerCertificate=True");
    }
}

public interface IProfile
{
    LicenseProfile Profile { get; set; }
    int ProfileId { get; set; }
}

public class LicenseProfile
{
    public int Id { get; set; }
    public required string License { get; set; }
}

public abstract class CustomerBase : IProfile
{
    public int Id { get; set; }
    public int ProfileId { get; set; }
    public LicenseProfile Profile { get; set; } = null!;
}

public sealed class Customer : CustomerBase
{
    public string Name { get; set; } = null!;
}

internal static class LicenseFilterExtensions
{
    public static IQueryable<T> FilterLicense<T>(this IQueryable<T> source, IEnumerable<string>? licenseFilter) where T : class, IProfile
    {
        if (licenseFilter != null)
        {
            return source.Where(x => licenseFilter.Contains(x.Profile.License));
        }

        return source;
    }
}
```

### Exception details

```
LinqToDB.LinqToDBException: The LINQ expression could not be converted to SQL.
Expression:
<>c__DisplayClass0_0<Customer>.licenseFilter
	.Contains(x.Profile.License)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 379
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.Build[T](IQueryExpressions& expressions) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 128
   at LinqToDB.Internal.Linq.Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, IQueryExpressions& expressions) in /_/Source/LinqToDB/Internal/Linq/Query{T}.cs:line 403
   at LinqToDB.Internal.Linq.Query`1.GetQuery(IDataContext dataContext, IQueryExpressions& expressions, Boolean& dependsOnParameters) in /_/Source/LinqToDB/Internal/Linq/Query{T}.cs:line 380
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetQuery(IQueryExpressions& expression, Boolean cache, Boolean& dependsOnParameters) in /_/Source/LinqToDB/Internal/Linq/ExpressionQuery.cs:line 91
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetForEachAsync(Action`1 action, CancellationToken cancellationToken) in /_/Source/LinqToDB/Internal/Linq/ExpressionQuery.cs:line 201
   at LinqToDB.Async.AsyncExtensions.ToListAsync[TSource](IQueryable`1 source, CancellationToken token) in /_/Source/LinqToDB/Async/AsyncExtensions.cs:line 281
```

### Regression?

_No response_

### Linq To DB for Entity Framework Core version

10.1.0

### Entity Framework Core version

10.0.3

### .NET version

.NET 10

### Database

SQL Server 2026

### Entity Framework database provider

Microsoft.EntityFrameworkCore.SqlServer 10.0.3

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
## Summary

Fix [#5355](https://github.com/linq2db/linq2db/issues/5355) — `Contains(navigation.column)` over a query produced by `ToLinqToDBTable()` failed with *"The LINQ expression could not be converted to SQL"* when the navigation member was declared on an interface or abstract base.

## Cause

`IsAssociationInRealization` step 3 realized the member to `expression.Type.GetMemberEx(member)` but then recursed with `expression=null`, so the recursive step 1 keyed the association lookup on `newMember.DeclaringType` — which for an EFCore-mapped model is an interface or abstract base that EF doesn't register as an entity. `EFCoreMetadataReader.GetAttributes` returned nothing → `IsAssociation` returned false → navigation never expanded → `ConvertInPredicate` failed.

Pure linq2db with `[Association]` was unaffected because reflection finds the attribute regardless of which type does the lookup.

## Fix

Before falling back to the recursion, try `MappingSchema.GetAttribute<AssociationAttribute>(expression.Type, newMember)` — keyed on the realized concrete type. `EFCoreMetadataReader.CompareProperty` already accepts that combination.

## Test plan

- [x] New EFCore-side regression test in `IssueTests.Issue5355_ContainsViaIEnumerableInGenericMethod` / `…ContainsViaArrayInGenericMethod`
- [x] Full `Tests.EntityFrameworkCore.EF10` (401 / 401)
- [x] `AssociationTests` on `Tests.csproj` (1259 / 1259)
- [x] `Issue5069Tests` + `GenericsFilterTests` (16 / 16)
- [x] CI matrix (will trigger after PR creation)

Fixes #5355

## Drive-by

Also makes `FluentMappingBuildTests` use unique temp-table names per test (was `"FluentTemp"` shared across 5 tests, surfaced as `"object TABLE FluentTemp is in use"` on Firebird 3 when the previous test's connection still held a metadata reference at CREATE time).


viceroypenguin (approved):

Tests/EntityFrameworkCore/Tests/IssueTests.cs
```diff
@@ -1042,6 +1042,42 @@ public void AttachToExistingTransaction([EFDataSources] string provider)
...
+			var result = ctx.Issue5355Customers
+				.ToLinqToDBTable()
+				.FilterIssue5355License(licenseFilter)
+				.OrderBy(x => x.Id)
+				.ToList();
+
+			result.Count.ShouldBe(2);
+			result[0].Id.ShouldBe(1);
+			result[1].Id.ShouldBe(2);
+		}
```
MaceWindu, line 1079: **Suggestion · SUG001** — Redundant with `Issue5355_ContainsViaIEnumerableInGenericMethod` directly above.

`FilterIssue5355License<T>` declares its parameter as `IEnumerable<string>?`, so the captured lambda body resolves to `Enumerable.Contains<string>(IEnumerable<string>, string)` regardless of whether the caller passes `string[]` or `IEnumerable<string>`. The expression tree handed to LINQ-to-DB is identical between the two tests — independently confirmed by the per-provider `.sql` baselines for `ContainsViaArray*` and `ContainsViaIEnumerable*` being byte-identical blobs in the baselines PR. The test name `ContainsViaArray` is also slightly misleading: at the point the regression is triggered, the filter is statically `IEnumerable<string>`, not `string[]`.

Fix: Drop the redundant test (auto-applied by the suggestion below). If you want distinct coverage of an array call site, the helper would need a `string[]?`-typed overload, or a separate helper whose lambda binds to a different `Contains` member (e.g. `Array.IndexOf`) — without that, both tests compile to the same expression tree.

```suggestion
```

MaceWindu (commented):
> [!IMPORTANT]

## Review notes

- [ ] **Linked issue.** Closes #5355 (area: efcore, milestone 6.3.0). The fix lands exactly where the issue's repro fails — `Contains(navigation.column)` on a `ToLinqToDBTable()`-typed source where the navigation member is declared on an interface or abstract base.
- [ ] **Cause / fix narrative cross-checked.** `IsAssociationInRealization` step 3 realized the member to `expression.Type.GetMemberEx(member)` but recursed with `expression = null`, so the recursive step 1 keyed `IsAssociation` on `newMember.DeclaringType` — for an EF-mapped interface or abstract base this is not a registered entity and `EFCoreMetadataReader.GetAttributes` returns nothing. Probing `MappingSchema.GetAttribute<AssociationAttribute>(expression.Type, newMember)` *before* the recursion catches this case at the concrete type. Strict superset of the prior behavior (recursion is preserved as fallback); `EFCoreMetadataReader.CompareProperty` already supports the (concrete-type, base-member) combination.
- [ ] **Drive-by scope.** The Firebird-driven `FluentMappingBuildTests` temp-table rename (`FluentTemp` → `FluentTemp_<TestSuffix>`) is unrelated to #5355 and called out separately in the PR body. The cross-provider baseline rename pattern is consistent across all 5 affected tests; no shape changes accompany it.
- [ ] **Public-API surface.** No public-API changes on this PR.
- [ ] **One inline suggestion** on `Tests/EntityFrameworkCore/Tests/IssueTests.cs` — see the comment thread.

## Baselines

Delta: [linq2db.baselines PR #1831](https://github.com/linq2db/linq2db.baselines/pull/1831) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5511)

366 files changed: 112 added + 254 modified, in two cleanly separated buckets, both fully explained by the confirmed scope. No cross-provider anomalies.

### New EF regression baselines for Issue5355 (2 tests × 56 providers = 112 files)

Both `Issue5355_ContainsViaArrayInGenericMethod` and `Issue5355_ContainsViaIEnumerableInGenericMethod` emit the same SQL: `SELECT … FROM Issue5355Customers x INNER JOIN Issue5355LicenseProfiles a_Profile ON x.ProfileId = a_Profile.Id WHERE a_Profile.License IN ('12345') ORDER BY x.Id`. Across all providers only string-literal prefix (`N'…'` for SqlServer vs `'…'` for PostgreSQL/SQLite) and identifier quoting style differ — both expected. The two tests' per-provider `.sql` blobs are byte-identical.

- **Issue5355_ContainsViaArrayInGenericMethod — SqlServer EF variants** (12 providers: SqlServer.2019.MS.EF8/9/10/31, SqlServer.Contained.MS.EF8/9/10/31, SqlServer.SA.MS.EF8/9/10/31; same SQL shape repeated for SqlServer.2016/2017/2022/2025 EF variants — total 5 SqlServer cohorts).
  - [sample baseline](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2019.MS.EF10/LinqToDB/EntityFrameworkCore/Tests/IssueTests/LinqToDB.EntityFrameworkCore.Tests.IssueTests.Issue5355_ContainsViaArrayInGenericMethod%28SqlServer.2019.MS%29.sql) — added
- **Issue5355_ContainsViaArrayInGenericMethod — PostgreSQL EF + SQLite.MS EF** (19 providers: PostgreSQL.13–17.EF8/9/10, SQLite.MS.EF8/9/10/31).
  - [sample baseline](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/PostgreSQL.15.EF10/LinqToDB/EntityFrameworkCore/Tests/IssueTests/LinqToDB.EntityFrameworkCore.Tests.IssueTests.Issue5355_ContainsViaArrayInGenericMethod%28PostgreSQL.15%29.sql) — added
- **Issue5355_ContainsViaIEnumerableInGenericMethod — all EF providers** (byte-identical to the ContainsViaArray baseline blobs per provider).
  - [sample baseline](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/PostgreSQL.15.EF10/LinqToDB/EntityFrameworkCore/Tests/IssueTests/LinqToDB.EntityFrameworkCore.Tests.IssueTests.Issue5355_ContainsViaIEnumerableInGenericMethod%28PostgreSQL.15%29.sql) — added

### FluentMappingBuildTests temp-table rename (5 tests × ~48–56 providers = 254 files)

Pure name substitution: `FluentTemp` → `FluentTemp_<TestSuffix>` (Cache / Update / UpdateAsync / Merge / InsertOrUpdate), including the matching `PK_FluentTemp*` constraint name. SQL operation shape (CREATE TABLE, INSERT, SELECT, UPDATE, MERGE, DROP TABLE) is unchanged across every provider. Provider-specific DDL variants (DB2 `EXECUTE IMMEDIATE`, MySQL backtick quoting, Oracle/PostgreSQL ANSI quoting, SqlServer `OBJECT_ID` guards) are all expected.

- [CacheTest sample (SqlServer.2016.MS)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2016.MS/Tests/Mapping/FluentMappingBuildTests/Tests.Mapping.FluentMappingBuildTests.CacheTest%28SqlServer.2016.MS%29.sql) — modified (all 56 providers)

  ```diff
  @@ -2,8 +2,8 @@
  -IF (OBJECT_ID(N'[FluentTemp]', N'U') IS NULL)
  -	CREATE TABLE [FluentTemp]
  +IF (OBJECT_ID(N'[FluentTemp_Cache]', N'U') IS NULL)
  +	CREATE TABLE [FluentTemp_Cache]
   ...
  -DROP TABLE IF EXISTS [FluentTemp]
  +DROP TABLE IF EXISTS [FluentTemp_Cache]
  ```
  </details>
- [UpdateTest sample (SqlServer.2016.MS)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2016.MS/Tests/Mapping/FluentMappingBuildTests/Tests.Mapping.FluentMappingBuildTests.UpdateTest%28SqlServer.2016.MS%29.sql) — modified (all 56 providers; pattern same as CacheTest with suffix `_Update`)
- [UpdateTestAsync sample (SqlServer.2016.MS)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2016.MS/Tests/Mapping/FluentMappingBuildTests/Tests.Mapping.FluentMappingBuildTests.UpdateTestAsync%28SqlServer.2016.MS%29.sql) — modified (all 56 providers; suffix `_UpdateAsync`)
- [MergeTestAsync sample (SqlServer.2016.MS)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2016.MS/Tests/Mapping/FluentMappingBuildTests/Tests.Mapping.FluentMappingBuildTests.MergeTestAsync%28SqlServer.2016.MS%29.sql) — modified (38 providers — SqlServer, DB2, Oracle, MariaDB/MySQL; absent on PostgreSQL/SQLite/Access by design)
- [InsertOrUpdatePrimaryKeyTest sample (SqlServer.2016.MS)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5511/SqlServer.2016.MS/Tests/Mapping/FluentMappingBuildTests/Tests.Mapping.FluentMappingBuildTests.InsertOrUpdatePrimaryKeyTest%28SqlServer.2016.MS%29.sql) — modified (48 providers; uses provider-appropriate upsert syntax — MERGE on SqlServer, ON CONFLICT on PostgreSQL, ON DUPLICATE KEY UPDATE on MySQL/MariaDB, INSERT OR REPLACE on SQLite)

MaceWindu (approved):

PR #5512, opened 2026-05-06
Fix C# string concatenation null semantics

igor-tkachev:
**Summary**

Fixes C# string concatenation null semantics for translated LINQ expressions.

`string + string` now treats nullable operands like C# does: `null` is concatenated as an empty string instead of letting SQL null-propagation turn the whole result into `NULL`. The translation only adds `COALESCE(..., '')` for operands that can actually be nullable, avoiding unnecessary SQL for non-null operands.

**Changes**

- Added nullable-aware string concatenation handling in `ExpressionBuildVisitor`.
- Added Sybase-specific handling to avoid generating broken empty-string `COALESCE` patterns where Sybase already handles one-null string concatenation cases.
- Extended `CreateLocalTable` to support optional FluentMapping configuration, matching `CreateTempTable` usage.
- Added independent test data for string concatenation null semantics.
- Added parameterized tests for:
  - `string + string`
  - `string + int`
  - nullable/non-null operand combinations
  - local and remote/LinqService providers through `[DataSources]`
- Documented preferred test comparison helpers:
  - use `AssertQuery` when the expected result can be expressed as the same LINQ query;
  - use `AreEqual` when expected and actual need different query shapes.

**Sybase Note**

Sybase cannot represent the C# empty string result for `null + null` string concatenation: `''` materializes as a single space, while zero-length string expressions such as `RTRIM('')`, `Space(0)`, and similar expressions return `NULL`. Because of that, only the impossible `both nullable` Sybase case is skipped; all other Sybase combinations remain covered, including LinqService.

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/Sybase/SybaseSqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Sybase/SybaseSqlExpressionConvertVisitor.cs
@@ -88,6 +88,37 @@ namespace LinqToDB.Internal.DataProvider.Sybase
 			};
 		}
 
+		public override IQueryElement ConvertSqlBinaryExpression(SqlBinaryExpression element)
+		{
+			if (element is { Operation: "+", SystemType: var type } && type.IsStringType)
+			{
+				var expr1 = UnwrapEmptyStringCoalesce(element.Expr1);
+				var expr2 = UnwrapEmptyStringCoalesce(element.Expr2);
+
+				if (!ReferenceEquals(expr1, element.Expr1) || !ReferenceEquals(expr2, element.Expr2))
+					return new SqlBinaryExpression(element.Type, expr1, element.Operation, expr2, element.Precedence);
+			}
+
+			return base.ConvertSqlBinaryExpression(element);
+
+			static ISqlExpression UnwrapEmptyStringCoalesce(ISqlExpression expression)
+			{
+				return expression switch
+				{
+					SqlFunction
+					{
+						Name      : "Coalesce",
+						Parameters: [var value, SqlValue { Value: "" }]
+					} => value,
+					SqlCoalesceExpression
+					{
+						Expressions: [var value, SqlValue { Value: "" }]
+					} => value,
+					_ => expression
+				};
+			}
+		}
+
 		protected override ISqlExpression WrapColumnExpression(ISqlExpression expr)
 		{
 			if (expr is SqlValue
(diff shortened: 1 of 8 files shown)


sdanyliv:
Looks like conflict with https://github.com/linq2db/linq2db/pull/5504
Do not merge before that

Source/LinqToDB/Internal/DataProvider/Sybase/SybaseSqlExpressionConvertVisitor.cs
```diff
@@ -88,6 +88,37 @@ public override ISqlExpression ConvertSqlFunction(SqlFunction func)
 			};
 		}
 
+		public override IQueryElement ConvertSqlBinaryExpression(SqlBinaryExpression element)
+		{
+			if (element is { Operation: "+", SystemType: var type } && type.IsStringType)
```
viceroypenguin:
```suggestion
			if (element is { Operation: "+", SystemType: { IsStringType: true } type })
```

viceroypenguin (dismissed):

sdanyliv (changes requested):
Should be totally fixed by proper branch https://github.com/linq2db/linq2db/pull/5504
Better move infrastructure changes in separate branch. String concatenation is fully rewritten.

jods4 (changes requested):
I would really like this to be conditioned by option `CompareNulls` (even though it's a slightly bad name, it's really the same intention).

Note that this is a **breaking change**: for example in the project I work on we sometimes concatenate a nullable column with a non-null column with the expectation that if the second is null, the result is null.

So at least when we compare nulls as SQL, I would like string concatenation (and everything else) to work as SQL.

When people opt into C#/.NET semantics, why not.

Also note that Oracle has no "empty string", it's represented by NULL. So contrary the SQL standard, `'a' || null` is not null, because it also represents `'a' || ''`.

igor-tkachev:
> I would really like this to be conditioned by option `CompareNulls` (even though it's a slightly bad name, it's really the same intention).

Sure, just forgot about it.

igor-tkachev:
> Should be totally fixed by proper branch #5504 Better move infrastructure changes in separate branch. String concatenation is fully rewritten.

When do you think we finish it. It is urgent for me. If it takes time we can consider this patch as quick fix and remove it later. I will add TODOs.

sdanyliv:
> When do you think we finish it. It is urgent for me. If it takes time we can consider this patch as quick fix and remove it later. I will add TODOs.
  
 Tomorrow, if test pipelines will be not be overloaded.

sdanyliv:
Superseded by https://github.com/linq2db/linq2db/pull/5504

PR #5514, opened 2026-05-08
Fix TrimStart/TrimEnd methods

igor-tkachev:
Fix TrimStart/TrimEnd(trimChars) translation fallback to preserve client-side semantics in final projection.


sdanyliv:
Superseded by https://github.com/linq2db/linq2db/pull/5515

PR #5522, opened 2026-05-12
Fix IN-to-EXISTS conversion for set-based subqueries

igor-tkachev:
## What was fixed

This PR fixes the `Contains`/`In` case when the subquery contains a `Union`.

The issue was in the `IN -> EXISTS` conversion: the correlation condition could be applied at the wrong level for set-based queries, which could produce incorrect results.

### `SqlExpressionConvertVisitor`
`Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs`

- Updated `WrapIfNeeded(...)` to also wrap queries with set operators (`HasSetOperators`).
- As a result, `ConvertToExists(...)` now applies correlation at the correct query level (to the full set result).

## Tests

`Tests/Linq/Linq/InSubqueryTests.cs`

Added regression tests:

- `ContainsWithUnionKeepsProjectionTest`
- `ContainsWithGroupByKeepsProjectionTest`

Both tests use `AreEqual(expected, actual)` with explicitly computed expected data.

## Follow-up commits

- `ab82cc4` — Switched the two new tests' `expected` computation from `db.Parent`/`db.Child` re-queries to the cached `Parent`/`Child` sequences exposed by `TestBase.Tables.cs`, matching the fixture neighborhood's convention and avoiding extra per-test round-trips.
- `a984a64` — Reworded the new `.github/copilot-instructions.md` indentation rule to defer to `.editorconfig` rather than asserting "never spaces" universally, which would have misfired on `.yml` / `.sh` / `.md` / `.fs` files.
- `d47bbd6` — Moved the two new tests from `SelectQueryTests.cs` to `InSubqueryTests.cs`, co-locating them with the other `UsePreferExistsForScalar`-gated tests (`ContainsTest`, `ContainsExprTest`, `ContainsNullTest`, `In*Test`). CI will regenerate baselines at the new test paths on the next run.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -21,4 +21,3 @@ When performing a code review, **comment on formatting only when it is clearly p
 ## Testing Guidelines
 
 - Use Shouldly for assertions in tests instead of NUnit Assert.
-- For test methods in this repository, always suffix method names with 'Test'.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -15,6 +15,10 @@ When performing a code review, **comment on formatting only when it is clearly p
 - indentation that is clearly broken (e.g., half-indented blocks or accidental deep indentation),
 - mixed tabs and spaces *when it creates visibly misaligned code*.
 
+## Indentation Rules
+- Always use tabs for leading indentation and never spaces in this repository.
+
 ## Testing Guidelines
 
 - Use Shouldly for assertions in tests instead of NUnit Assert.
+- For test methods in this repository, always suffix method names with 'Test'.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -16,7 +16,7 @@ When performing a code review, **comment on formatting only when it is clearly p
 - mixed tabs and spaces *when it creates visibly misaligned code*.
 
 ## Indentation Rules
-- Always use tabs for leading indentation and never spaces in this repository.
+- Respect the repository's `.editorconfig` for indentation rules.
 
 ## Testing Guidelines
 
--- a/Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
@@ -368,8 +368,8 @@ namespace LinqToDB.Internal.SqlProvider
 		public IQueryElement ConvertIsDistinctPredicate(SqlPredicate.IsDistinct predicate)
 		{
 			/*
-				(value1 IS NULL AND value2 IS NOT NULL) OR 
-				(value1 IS NOT NULL AND value2 IS NULL) OR 
+				(value1 IS NULL AND value2 IS NOT NULL) OR
+				(value1 IS NOT NULL AND value2 IS NULL) OR
 				(value1 <> value2)
 			 */
 
@@ -1333,7 +1333,7 @@ namespace LinqToDB.Internal.SqlProvider
 
 		static SelectQuery WrapIfNeeded(SelectQuery selectQuery)
 		{
-			if (selectQuery.Select.HasModifier || !selectQuery.GroupBy.IsEmpty || QueryHelper.IsAggregationQuery(selectQuery))
+			if (selectQuery.Select.HasModifier || selectQuery.HasGroupBy || selectQuery.HasSetOperators || QueryHelper.IsAggregationQuery(selectQuery))
 			{
 				var newQuery = new SelectQuery();
 				newQuery.From.Tables.Add(new SqlTableSource(selectQuery, null));
@@ -1383,7 +1383,7 @@ namespace LinqToDB.Internal.SqlProvider
 				var testValue = testExpressions[i];
 				var expr      = subQuery.Select.Columns[i].Expression;
 
-				predicates.Add(new SqlPredicate.ExprExpr(testValue, SqlPredicate.Operator.Equal, expr, DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeClr ? true : null));
+                predicates.Add(new SqlPredicate.ExprExpr(testValue, SqlPredicate.Operator.Equal, expr, DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeClr ? true : null));
 			}
 
 			subQuery.Select.Columns.Clear();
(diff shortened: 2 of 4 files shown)


Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
```diff
@@ -1383,7 +1383,7 @@ ISqlPredicate ConvertToExists(SqlPredicate.InSubQuery inPredicate)
 				var testValue = testExpressions[i];
 				var expr      = subQuery.Select.Columns[i].Expression;
 
-				predicates.Add(new SqlPredicate.ExprExpr(testValue, SqlPredicate.Operator.Equal, expr, DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeClr ? true : null));
+                predicates.Add(new SqlPredicate.ExprExpr(testValue, SqlPredicate.Operator.Equal, expr, DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeClr ? true : null));
```
viceroypenguin:
```suggestion
			predicates.Add(
				new SqlPredicate.ExprExpr(
					testValue,
					SqlPredicate.Operator.Equal,
					expr,
					DataOptions.LinqOptions.CompareNulls == CompareNulls.LikeClr ? true : null
				)
			);
```

viceroypenguin (dismissed):

.github/copilot-instructions.md
```diff
@@ -15,6 +15,10 @@ When performing a code review, **comment on formatting only when it is clearly p
 - indentation that is clearly broken (e.g., half-indented blocks or accidental deep indentation),
 - mixed tabs and spaces *when it creates visibly misaligned code*.
 
+## Indentation Rules
+- Always use tabs for leading indentation and never spaces in this repository.
+
 ## Testing Guidelines
 
 - Use Shouldly for assertions in tests instead of NUnit Assert.
+- For test methods in this repository, always suffix method names with 'Test'.
```
sdanyliv: I do not agree with this rule. Everything is test in test project. Why we need superfluous word?

viceroypenguin (dismissed):

.github/copilot-instructions.md
igor-tkachev: The issue is not whether code is located in a test project. The issue is everyday navigation and search.

When I search for `SelectQuery`, use Search everywhere, read stack traces, browse test runner output, or grep through repository, having `SelectQuery` and `SelectQueryTest` immediately tells me what is production code and what is test code. No ambiguity, no accidental jumps into wrong method.

If both methods are named `SelectQuery`, then navigation becomes unnecessarily annoying for everyone else. Adding `Test` costs almost nothing, but removes constant friction during everyday work.

This is also a very common convention in .NET projects exactly for this reason. Personally I would make it not only an agent rule, but also a review guideline and general project-wide coding convention.

And after tenth accidental jump into wrong `SelectQuery`, people will eventually start blaming person who decided that four extra characters were too expensive :) I am trying to protect your karma here.

.github/copilot-instructions.md
```diff
@@ -15,6 +15,10 @@ When performing a code review, **comment on formatting only when it is clearly p
 - indentation that is clearly broken (e.g., half-indented blocks or accidental deep indentation),
 - mixed tabs and spaces *when it creates visibly misaligned code*.
 
+## Indentation Rules
+- Always use tabs for leading indentation and never spaces in this repository.
```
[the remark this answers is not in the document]
MaceWindu: Fixed in `a984a64` — rule reworded to defer to `.editorconfig`, which already encodes the per-extension rules.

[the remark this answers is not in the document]
MaceWindu: Fixed in `a984a64` — same change as the duplicate finding on the prior commit.

Tests/Linq/Linq/SelectQueryTests.cs
```diff
@@ -47,6 +47,61 @@ public void UnionTest([DataSources(TestProvName.AllAccess)] string context)
...
+
+			var query = db.Parent.Where(p => union.Contains(p.ParentID));
+
+			var actual = query.OrderBy(p => p.ParentID).ToArray();
+
+			var expected = Parent
+				.Where(p =>
+					Parent.Where(p1 => p1.ParentID <= 2).Select(p1 => p1.ParentID)
+						.Union(Child.Where(c => c.ParentID >= 3).Select(c => c.ParentID))
+						.Contains(p.ParentID))
```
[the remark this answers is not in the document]
MaceWindu: Inaccurate at HEAD. The comparison to `groupedKeys` in `ContainsWithGroupByKeepsProjectionTest` doesn't hold — `groupedKeys` is also a deferred `IEnumerable<int>` (`.Select(g => g.Key)`, no `.ToList()`/`.ToHashSet()`), so `.Contains(p.ParentID)` re-iterates the source per Parent row in both tests. Materialising the Union expression into a variable would not change re-enumeration behaviour. Test data is tiny and correctness is unchanged.

.github/copilot-instructions.md
sdanyliv: Never had such issues. Why? Probably because methods have name with common sense, `SlectQuery_With_Take_ShouldThrow` ?

.github/copilot-instructions.md
```diff
@@ -15,6 +15,10 @@ When performing a code review, **comment on formatting only when it is clearly p
 - indentation that is clearly broken (e.g., half-indented blocks or accidental deep indentation),
 - mixed tabs and spaces *when it creates visibly misaligned code*.
 
+## Indentation Rules
+- Respect the repository's `.editorconfig` for indentation rules.
+
 ## Testing Guidelines
 
 - Use Shouldly for assertions in tests instead of NUnit Assert.
+- For test methods in this repository, always suffix method names with 'Test'.
```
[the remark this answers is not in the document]
MaceWindu: Duplicate of [#discussion_r3237187800](https://github.com/linq2db/linq2db/pull/5522#discussion_r3237187800) — same content, still under maintainer discussion on the separate thread started by @sdanyliv.

MaceWindu (dismissed):
waiting for baselines

.github/copilot-instructions.md
MaceWindu:
```suggestion
```
removing unrelated opionated change till consensus

PR #5524, opened 2026-05-12
Add ActiveIssue regression test for #5441 (NText concat)

MaceWindu:
## Summary

- Adds `TestNTextConcat` covering `string + r.NText` and interpolated `$"... {r.NText} ..."` projections on SQL Server.
- Marked `[ActiveIssue(5441)]` — currently throws `The data types nvarchar and ntext are incompatible in the add operator` because the translator doesn't emit `CAST(NText AS NVARCHAR(MAX))` on the `string + NText` add path.
- Original report: #5441 (reopened). Originally attempted in #5442 (closed, "will be handled in #5299") and again in #5504; removed from #5504 since the underlying translator fix isn't in scope.

Refs #5441

## Test plan

- [ ] Test is skipped on all configurations (`[ActiveIssue]` gating).

### DIFF
--- a/Tests/Linq/DataProvider/SqlServerTests.cs
+++ b/Tests/Linq/DataProvider/SqlServerTests.cs
@@ -2368,5 +2368,50 @@ DROP TABLE IF EXISTS TemporalTable3History
 
 			Assert.That(tb.Where(s => s.LastModified.Date == TestData.DateTime.Date).Count(), Is.EqualTo(1));
 		}
+
+		sealed class NTextTable
+		{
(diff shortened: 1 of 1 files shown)


MaceWindu:
Superseded by #5504 — the underlying NText/Text concat fix turned out to depend on `SqlConcatExpression` / `ConvertConcat` infrastructure introduced in #5504, so both fix and regression test (without `[ActiveIssue]`) have been folded into that PR. #5504 now lists `Fixes #5441` and will close the issue on merge.

Issue #5513, opened 2026-05-06
Tests: remote-mode SQL trace drops scalar IQueryable<T> aggregate queries


MaceWindu:
### Summary

Remote-mode SQL trace drops the SQL of scalar `IQueryable<T>` aggregate queries (e.g. `db.Child.CovarSamp(...)`) before it reaches the test framework's baseline writer, producing spurious `.sql.other` mismatch failures on Oracle. Pre-existing — 4 entries already in `master`, plus a 5th surfaced by #5467 via test ordering changes.

### Background

`Tests/Base/BaselinesWriter.cs:62-79` writes `.sql.other` and throws `"Baselines for remote context doesn't match direct access baselines"` whenever the same test produces different captured SQL between **direct** (`DataConnection`) and **remote** (`LinqService` → `DataConnection`) execution.

The trace pipeline at `Tests/Base/TestBase.cs:38-75` registers a global `DataConnection.WriteTraceLine` handler that forwards every `BeforeExecute` event to `BaselinesManager.LogQuery` (`Tests/Base/BaselinesManager.cs:7-22`), which appends to the `BASELINE` `StringBuilder` on `CustomTestContext`. End of test, the `StringBuilder` is dumped to `.sql` (or `.sql.other` on direct-vs-remote mismatch).

### Current behavior

**Direct mode** — every query through the test's `DataConnection` fires `BeforeExecute` → trace handler captures → SQL appears in the baseline.

**Remote mode** — queries through `LinqService` use the service's internal `DataConnection` to actually run the SQL; for most query shapes the global trace handler does fire as expected. But for the `IQueryable<TEntity>` aggregate path — `CovarSamp`, `CovarPop`, and ~14 other extensions in `Source/LinqToDB/AnalyticFunctions.cs` (lines 408, 444, 476, 492, 532, 572, 604, 710, 742, 770, 896, 929, 956, 991, 1018) — the SQL never reaches the `BASELINE` `StringBuilder`.

The shape those methods share:

```csharp
public static decimal? CovarSamp<TEntity>(this IQueryable<TEntity> source, …)
{
    var currentSource = source.GetLinqToDBSource();
    return currentSource.Execute<decimal>(Expression.Call(...));
}
```

[...]

PR #5525, opened 2026-05-13
Add named (multiple) query filters mirroring EF Core 10 API

sdanyliv:
## Summary

Adds named-filter support to `EntityMappingBuilder<T>.HasQueryFilter`,
mirroring EF Core 10's keyed-filter shape:

- Five new keyed `HasQueryFilter(string filterKey, ...)` overloads (lambda + Func, with and without `TDataContext`). Passing `null` removes the entry under that key. Multiple keys per entity are AND-combined at query time.
- `IQueryable<T>.IgnoreFilters` gains `(string[] filterKeys)` and `(string[] filterKeys, Type[] entityTypes)` overloads to selectively disable filters by key or by key×type intersection.
- New `EntityQueryFilter` public type plus `EntityDescriptor.QueryFilters` collection; the legacy `QueryFilterLambda` / `QueryFilterFunc` accessors remain as back-compat shims for the default (anonymous) entry.
- `TranslationModifier` switches to a `FilterIgnoreScope` list to express the keys×types disable matrix uniformly; existing `Push/Pop` behavior is preserved.
- **EF Core 10 integration**: `EFCoreMetadataReader` forwards `IQueryFilter.Key` into `QueryFilterAttribute.FilterKey`; `TransformExpressionVisitor` remaps EF's `IgnoreQueryFilters(IReadOnlyCollection<string>)` to the new keyed `IgnoreFilters` overload (gated `#if EF10`).

## Test plan

- [x] 9 new `QueryFilterTests` cases cover AND-combine, null-removal, ignore-by-key, ignore-by-type, ignore-by-key+type intersection, anonymous+named coexistence, scope accumulation, and func-overload routing — passing on SQLite.Classic and SQLite.MS.
- [x] 3 new EF10-gated `ToolsTests` cases cover EF → linq2db named-filter remapping plus back-compat for the no-arg `IgnoreQueryFilters()`.
- [x] Existing `QueryFilterTests` + `TestGlobalQueryFilters` continue to pass (back-compat through the descriptor / TableBuilder / TranslationModifier refactor).
- [x] Full `linq2db.slnx -c Release` build clean across `net462`, `netstandard2.0`, `net8.0`, `net9.0`, `net10.0` — analyzers and banned-API checks pass.
- [x] API baselines refresh (`/api-baselines`) — no new suppressions needed (additive public surface; legacy accessors retained).

## Follow-up (interface/base-typed filter lambdas)

Merged latest `master`. Addresses the review feedback on interface/base-typed filter lambdas:

- `TableBuilder.ApplyQueryFilters`: both the lambda-only fast path and the func dynamic-accessor path now route through a shared `BuildFilterLambda` helper that adapts a filter lambda typed against an interface/base type (e.g. `Expression<Func<ISoftDelete, IDataContext, bool>>` applied to a concrete entity) to the concrete entity type. Previously this threw `ArgumentException` at query build. (MAJ001 / MAJ002 + the maintainer change-request.)
- Corrected the pre-EF10 query-filter comment in `NorthwindContext` (MIN002).
- New tests: interface-typed lambda via fluent `HasAttribute` (fast + dynamic paths); declarative `QueryFilterAttribute` subclass applied via `[Attribute]` (concrete + interface-typed); 3 direct `EntityDescriptor.QueryFilters` introspection tests (MIN001).

Verified: full `QueryFilterTests` fixture green on SQLite.MS; the interface-typed regression is confirmed red without the fix (the exact `ArgumentException`) and green with it.


igor-tkachev (dismissed):
There is one more query-filter shape that currently doesn't work: a `QueryFilterAttribute` subclass that defines the filter against an interface/base type, while the attribute is applied to a concrete entity.

This matters for reusable attribute-based filters. For example, a soft-delete/effective-period filter naturally wants to be declared once against an interface and then reused by many entities implementing that interface.

I verified this locally with the following regression test. It fails before the fix when `TableBuilder` tries to pass `Expression<Func<ISoftDelete,bool>>` to `Queryable.Where<AttributeFilteredEntity>`.

```csharp
interface ISoftDelete
{
	bool IsDeleted { get; set; }
}

sealed class InterfaceSoftDeleteQueryFilterAttribute : QueryFilterAttribute
{
	public InterfaceSoftDeleteQueryFilterAttribute()
	{
		FilterKey = SoftDeleteKey;

		FilterLambda =
			(Expression<Func<ISoftDelete, MyDataContext, bool>>)
			((e, dc) => !dc.IsSoftDeleteFilterEnabled || !e.IsDeleted);
	}
}

[Table]
[InterfaceSoftDeleteQueryFilter]
sealed class AttributeFilteredEntity : ISoftDelete
{
	[Column] public int  Id        { get; set; }
	[Column] public bool IsDeleted { get; set; }
}

[Test]
public void QueryFilterAttribute_WithInterfaceTypedLambda(
	[IncludeDataSources(false, TestProvName.AllSQLite, TestProvName.AllClickHouse)] string context)
{
	var data = Enumerable.Range(1, 10)
		.Select(i => new AttributeFilteredEntity { Id = i, IsDeleted = i % 2 == 0 })
		.ToArray();

	using var db = new MyDataContext(context, new MappingSchema());
	using (db.CreateLocalTable(data))
	{
		db.IsSoftDeleteFilterEnabled = true;

		var result = db.GetTable<AttributeFilteredEntity>().ToList();

		result.ShouldAllBe(e => !e.IsDeleted);
		result.Count.ShouldBe(5);
	}
}
```

Failure:

```text
Expression<Func<ISoftDelete,bool>> cannot be used for parameter of type
Expression<Func<AttributeFilteredEntity,bool>>
```

The issue is in `TableBuilder.ApplyQueryFilters`: it reuses `filterLambdaExpr.Parameters[0]` as the lambda parameter for `Where<TEntity>`. That works only when the filter parameter type is exactly the concrete entity type. For interface/base-class filters, the lambda should be adapted to the concrete `entityType`.

Suggested fix:

```csharp
static LambdaExpression BuildFilterLambda(
	MappingSchema mappingSchema,
	Type entityType,
	LambdaExpression filterLambdaExpr)
{
	var entityParam = filterLambdaExpr.Parameters[0];
	var dcParam     = filterLambdaExpr.Parameters[1];
	var dcExpr      = SqlQueryRootExpression.Create(mappingSchema, dcParam.Type);

	Expression body        = filterLambdaExpr.Body.Replace(dcParam, dcExpr);
	var        lambdaParam = entityParam;

	if (entityParam.Type != entityType)
	{
		if (!entityParam.Type.IsAssignableFrom(entityType))
			throw new InvalidOperationException(
				$"Query filter parameter type '{entityParam.Type}' is not assignable from entity type '{entityType}'.");

		lambdaParam = Expression.Parameter(entityType, entityParam.Name);
		body        = body.Replace(entityParam, Expression.Convert(lambdaParam, entityParam.Type));
	}

	return Expression.Lambda(body, lambdaParam);
}
```

Then use it in both lambda-filter paths in `ApplyQueryFilters` instead of rebuilding the lambda directly from `filterLambdaExpr.Parameters[0]`.

With this change, the regression test above passes locally, and the existing `QueryFilterTests` suite still passes.

igor-tkachev:
One design concern with parameterized filters: using `DataContext` members as filter parameters can silently produce wrong queries when the caller forgets to initialize them.

For example, a reusable effective-period filter might look like this:

```csharp
(e, db) =>
	e.EffectiveDate <= db.AsOfDate &&
	(e.ExpirationDate == null || e.ExpirationDate > db.AsOfDate)
```

This is convenient, but `AsOfDate` is just a normal member on the data context. There is no explicit contract that it must be set before querying. If the caller forgets to initialize it, the filter will use `default(DateTime)` and return incorrect data instead of failing fast.

This is especially risky for filters that encode business/security visibility rules, such as tenant, effective-date, organization scope, permissions, etc. In those cases, “missing parameter” should probably be an error, not an implicit default.

I don’t have a concrete API proposal yet, so this is not a blocker by itself. But I think the PR should at least acknowledge this limitation, because named filters make reusable/attribute-based filters more attractive, and those filters often need required runtime parameters.

Source/LinqToDB/Internal/Linq/Builder/TableBuilder.cs
```diff
@@ -98,58 +99,95 @@ static Expression ApplyQueryFilters(ExpressionBuilder builder, MappingSchema map
...
-				filteredExpression = Expression.Call(Methods.Queryable.Where.MakeGenericMethod(entityType), filteredExpression, Expression.Quote(filterLambda));
+				foreach (var entry in activeFilters)
+				{
+					var filterLambdaExpr = entry.FilterLambda!;
+					var dcParam          = filterLambdaExpr.Parameters[1];
+					var dcExpr           = SqlQueryRootExpression.Create(mappingSchema, dcParam.Type);
+					var filterLambda     = Expression.Lambda(filterLambdaExpr.Body.Replace(dcParam, dcExpr), filterLambdaExpr.Parameters[0]);
+
+					filteredExpression = Expression.Call(Methods.Queryable.Where.MakeGenericMethod(entityType), filteredExpression, Expression.Quote(filterLambda));
+				}
```
MaceWindu, line 139: **Major · MAJ001** — The lambda-only fast path reuses `filterLambdaExpr.Parameters[0]` verbatim as the parameter of the `Where<TEntity>` lambda, where `TEntity == entityType` (the concrete entity being queried). That holds only when the filter predicate's first parameter type is exactly `entityType`. For the publicly reachable shape where a `QueryFilterAttribute` subclass declares `FilterLambda` against an interface or base type — e.g. `Expression<Func<ISoftDelete, IDataContext, bool>>` applied to `class AttributeFilteredEntity : ISoftDelete` — `Methods.Queryable.Where.MakeGenericMethod(entityType)` is handed a `Func<ISoftDelete, bool>` and throws `ArgumentException: Expression of type 'Func<ISoftDelete,bool>' cannot be used for parameter of type 'Expression<Func<AttributeFilteredEntity,bool>>'` at query build. `QueryFilterAttribute.FilterLambda` is public and settable, and `EntityDescriptor` stores it as-is, so this is reachable user API. No test covers an interface/base-typed attribute filter (the fluent `HasQueryFilter` tests all build the lambda with the entity type as the first parameter). Same core issue as the maintainer's change-request; this is the lambda-only path.

Fix: Adapt the lambda to the concrete `entityType` before building the `Where` call — when `filterLambdaExpr.Parameters[0].Type != entityType` and is assignable from it, rebuild the lambda with a fresh `entityType` parameter and replace the original parameter in the body with `Expression.Convert(newParam, originalParamType)`. The change-request review includes a ready `BuildFilterLambda(mappingSchema, entityType, filterLambdaExpr)` helper; extract it once and call it from both this site and the dynamic-accessor path (MAJ002) rather than inlining the adaptation twice.

Source/LinqToDB/Internal/Linq/Builder/TableBuilder.cs
```diff
@@ -98,58 +99,95 @@ static Expression ApplyQueryFilters(ExpressionBuilder builder, MappingSchema map
...
+							var dcParam      = filterLambdaExpr.Parameters[1];
+							var dcExpr       = SqlQueryRootExpression.Create(ms, dcParam.Type);
+							var filterLambda = Expression.Lambda(filterLambdaExpr.Body.Replace(dcParam, dcExpr), filterLambdaExpr.Parameters[0]);
 
-					if (filterFunc != null)
-					{
-						var query    = ExpressionQueryImpl.CreateQuery(entityType, dc, sequenceExpr);
-						var filtered = filterFunc.DynamicInvokeExt<IQueryable>(query, dc);
+							sequenceExpr = Expression.Call(Methods.Queryable.Where.MakeGenericMethod(entityType), sequenceExpr, Expression.Quote(filterLambda));
+						}
```
MaceWindu, line 177: **Major · MAJ002** — Same defect as MAJ001 in the dynamic-accessor path, taken when at least one active filter on the entity is a `FilterFunc`. `filterLambdaExpr.Parameters[0]` is reused as the `Where<entityType>` lambda parameter, so an interface/base-typed lambda filter on an entity that also carries a func filter throws the same `ArgumentException` at query build. Both code paths must be fixed; an entity mixing a func-form filter with an interface-typed lambda filter hits this branch rather than the fast path.

Fix: Apply the same `entityType`-adaptation as MAJ001. Routing both sites through a single shared `BuildFilterLambda` helper fixes this path and MAJ001 together and avoids duplicating the adaptation logic.

Tests/EntityFrameworkCore/Models/Northwind/NorthwindContext.cs
```diff
@@ -57,10 +65,20 @@ public void ConfigureEntityFilter<TEntity>(ModelBuilder builder)
...
 
+#if EF10
+			// EF10: use named filter for soft-delete constraint to coexist with other named filters
+#if !NETFRAMEWORK
+			builder.Entity<TEntity>().HasQueryFilter("SoftDeleteFilter", e => !obj!.IsSoftDeleteFilterEnabled || !e.IsDeleted || !EF.Property<bool>(e, "IsDeleted"));
+#else
+			builder.Entity<TEntity>().HasQueryFilter("SoftDeleteFilter", e => !obj!.IsSoftDeleteFilterEnabled || !e.IsDeleted);
+#endif
+#else
+			// Earlier EF versions: anonymous filter (can coexist with at most one other anonymous filter)
```
MaceWindu: **Minor · MIN002** — This comment overstates pre-EF Core 10 behavior. Before EF Core 10 an entity supports exactly one query filter — a second `HasQueryFilter(...)` call replaces the first, it does not coexist with it. "can coexist with at most one other anonymous filter" implies two anonymous filters can be active at once on one entity, which is not the case (and the sibling comment a few lines up already states the single-filter reality correctly).

Fix: state the single-filter reality.

```suggestion
			// Earlier EF versions: only one query filter per entity (a later HasQueryFilter call replaces an earlier one)
```

Tests/Linq/Linq/QueryFilterTests.cs
MaceWindu, line 1: **Minor · MIN001** — The new public introspection surface — the `EntityQueryFilter` type and the `EntityDescriptor.QueryFilters` collection — has no test that reads it directly. The new cases all assert query results, which exercises filter construction indirectly, but nothing asserts the shape a user observes via `mappingSchema.GetEntityDescriptor(typeof(T)).QueryFilters`: entry count, per-entry `FilterKey`, which entries carry `FilterLambda` vs `FilterFunc`, ordering, and the default (anonymous) slot. Since these are documented user-facing API, each new public member warrants at least one direct test.

Fix: Add a unit test (no DB roundtrip needed) that builds a mapping schema with several named filters, fetches the `EntityDescriptor`, and asserts `QueryFilters` count / `FilterKey` set / `FilterLambda`-vs-`FilterFunc` population, plus the derived-overrides-base and null-removal outcomes at the collection level.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] **New public API surface (additive).** `6.4.0` is a minor release, so the surface expansion is allowed; purely additive surface needs no `Source/**/CompatibilitySuppressions.xml` changes, consistent with the PR's stated baseline refresh. New public symbols:
  - `LinqToDB.Mapping`: `EntityQueryFilter` (type — `FilterKey` / `FilterLambda` / `FilterFunc` + ctor); `EntityDescriptor.QueryFilters`; `QueryFilterAttribute.FilterKey`; `EntityMappingBuilder<T>.HasQueryFilter(string, …)` (5 keyed overloads).
  - `LinqToDB`: `LinqExtensions.IgnoreFilters<T>(string[])` and `IgnoreFilters<T>(string[], Type[])`.
- [x] **Interface/base-typed filter-lambda gap matches the existing change-request.** The maintainer's change-request flagged that a `QueryFilterAttribute` whose lambda is typed against an interface/base type throws at query build. This review confirms it (MAJ001 / MAJ002) and adds that **both** filter-application paths in `TableBuilder.ApplyQueryFilters` are affected, not just one.
- [x] **Prior automated-review threads re-checked at HEAD.** The seven earlier bot threads (filter-key null-validation ordering, `GetObjectID` delimiter collision, `FilterIgnoreScope` set-equality + empty/null normalization, unused `using`, base→derived attribute precedence, test-constant rename) are all addressed by the review-response commits — no premature closures, nothing still actionable.

## Out-of-scope observations

- **Parameterized filters that read uninitialized `DataContext` members silently produce wrong queries.** A filter like `(e, db) => e.EffectiveDate <= ((MyContext)db).AsOfDate` emits `default(DateTime)` when the caller forgets to set the member, quietly returning wrong rows instead of failing fast. This is pre-existing behavior of parameterized filters (reproducible on the base branch), not introduced here — but named filters make reusable, configuration-driven filters more attractive, so the limitation is worth acknowledging in the feature docs. (Raised separately in the maintainer's design comment.)
- **`FilterIgnoreScope.Normalize` returns single-element arrays without copying.** For a one-element keys/types array, `Normalize` returns the caller-supplied array instance directly rather than a defensive copy, so the stored `Keys` / `Types` can alias the array evaluated from the `IgnoreFilters` argument. Equality and hashing are unaffected (both iterate elements), and the source arrays come from query-dependent constants that aren't realistically mutated after construction, so this is not a correctness defect — noted only because the type frames these arrays as a canonical, immutable representation.
- **Pre-existing XML-doc typo in `QueryFilterAttribute.FilterLambda`.** `Source/LinqToDB/Mapping/QueryFilterAttribute.cs:35` ends the example with a stray `"/>` (`For example (e, db) => e.IsDeleted == false "/>`), a leftover from a malformed `<see>` tag. It pre-dates this PR and sits outside the changed hunks, but it's a one-character cleanup worth making while the file is open.

## Baselines

linq2dbot reported baseline changes on this PR, but the tracking baselines branch (`baselines/pr_5525`) no longer exists on the baselines repo — its PR ([linq2db.baselines#1847](https://github.com/linq2db/linq2db.baselines/pull/1847)) is closed and the branch was deleted — so the SQL/metrics delta can't be inspected here. Re-running the test pipeline to regenerate baselines before merge is recommended. Two shape changes to verify on the regenerated baselines:

- **Combined `WHERE` predicates.** With multiple named filters active on one entity, `TableBuilder.ApplyQueryFilters` chains them as separate `Queryable.Where` calls — correct AND composition, but the emitted `WHERE` now carries N AND-ed fragments where a single-filter case carried one. Fragment order follows `EntityDescriptor.QueryFilters` declaration order (deterministic), so baselines shouldn't churn run-to-run.
- **INNER → LEFT JOIN promotion.** `AssociationHelper` now promotes inline associations to LEFT JOIN + `DefaultIfEmpty` whenever `QueryFilters.Count > 0` (previously only for a default-slot lambda). This newly covers named-only and func-style filtered entities — the intended fix — and can't over-fire for tombstone-only descriptors (those are removed from the collection). Expect INNER→LEFT JOIN movement on tests exercising inline associations on filtered entities.

Source/LinqToDB/Internal/Linq/Builder/TableBuilder.cs
sdanyliv: Fixed in 7be369d383: both lambda-application paths in `ApplyQueryFilters` now route through a shared `BuildFilterLambda` helper that adapts an interface/base-typed filter lambda to the concrete `entityType` (fresh parameter + `Expression.Convert`). Covered by new `QueryFilterTests` regression tests (fluent fast/dynamic paths and declarative `[QueryFilter]` attribute, concrete + interface-typed).

Source/LinqToDB/Internal/Linq/Builder/TableBuilder.cs
sdanyliv: Fixed in 7be369d383 together with MAJ001 - the dynamic-accessor path calls the same `BuildFilterLambda` helper, so the interface/base-typed adaptation applies there too. The `NamedFilter_InterfaceTypedLambda_WithFunc_DynamicPath` regression test exercises this branch.

Tests/EntityFrameworkCore/Models/Northwind/NorthwindContext.cs
sdanyliv: Fixed in 7be369d383: the comment now states the single-filter reality for pre-EF10 (a later `HasQueryFilter` call replaces the earlier one).

Tests/Linq/Linq/QueryFilterTests.cs
sdanyliv: Fixed in b5e6832179: added direct `EntityDescriptor.QueryFilters` introspection tests - entry count/order and per-entry `FilterKey`, `FilterLambda`-vs-`FilterFunc` population, the legacy `QueryFilterLambda`/`QueryFilterFunc` default-slot mirror, null-removal, and derived-overrides-base.

sdanyliv:
@igor-tkachev Good catch, thanks - and your `BuildFilterLambda` suggestion was exactly right, so I went with it.

Both filter paths in `TableBuilder.ApplyQueryFilters` (the lambda-only fast path and the func/dynamic one) now go through that shared helper, so an interface- or base-typed filter lambda gets adapted to the concrete entity type before it reaches `Where<T>`. That landed in 7be369d383.

I added tests too, including one that's essentially your repro: a `QueryFilterAttribute` subclass with an `Expression<Func<ISoftDelete, ...>>` lambda applied to a concrete entity and picked up from a plain `MappingSchema`. There are also variants for a concrete-typed lambda and for registering the same filter through fluent `HasAttribute`, so both code paths are covered.

And I made sure it really catches the regression: without the fix the test fails with the same `ArgumentException` you hit (the `Func<ISoftDelete, bool>` vs `Where<TEntity>` mismatch), and passes with it. The whole `QueryFilterTests` suite is green on SQLite.

Source/LinqToDB/Internal/Linq/Builder/FilterIgnoreScope.cs
```diff
@@ -0,0 +1,117 @@
...
+
+		public bool MatchesType(Type entityType)
+		{
+			if (Types == null || Types.Length == 0)
+				return true;
+
+			return Array.IndexOf(Types, entityType) >= 0;
+		}
+
+		public bool Equals(FilterIgnoreScope? other)
```
MaceWindu: **Blocker · BLK001** — Release build fails here with `MA0186`. Both `Equals` overloads (this `IEquatable<FilterIgnoreScope>.Equals` on line 78 and the `object?` override on line 89) need `[NotNullWhen(true)]` on their nullable parameter; with `TreatWarningsAsErrors` that turns the analyzer warning into a build error. The dedicated `build` CI check is red on HEAD for exactly these two diagnostics (`test-all` stays green because it builds a non-analyzer config — Meziantou rules are Release-only). This is the only build blocker; nothing else in the log fails.

Fix: add `using System.Diagnostics.CodeAnalysis;` to the file, annotate this parameter (suggestion below), and apply the same `[NotNullWhen(true)]` to `object? obj` on line 89.

```suggestion
		public bool Equals([NotNullWhen(true)] FilterIgnoreScope? other)
```

Source/LinqToDB.EntityFrameworkCore/Internal/TransformExpressionVisitor.cs
```diff
@@ -182,6 +182,21 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
 						return newMethod;
 					}
 
+#if EF10
+					if (generic == ReflectionMethods.IgnoreQueryFiltersByKeyMethodInfo)
+					{
+						// EF: IgnoreQueryFilters(IReadOnlyCollection<string>) → linq2db: IgnoreFilters(string[])
+						var keysExpr = Expression.Call(
+							typeof(Enumerable), nameof(Enumerable.ToArray), [typeof(string)],
+							node.Arguments[1]);
```
MaceWindu: **Minor · MIN003** — The EF10 bridge maps `IgnoreQueryFilters(IReadOnlyCollection<string>)` → linq2db `IgnoreFilters(string[])` unconditionally, but the two APIs differ on the **empty-collection** input. EF Core 10 treats an empty collection as a no-op — its `QueryableMethodNormalizingExpressionVisitor` only records keys under `if (filterKeys?.Count > 0)`, so all filters stay applied. linq2db's `IgnoreFilters(new string[0])` normalizes empty→null in `FilterIgnoreScope` (`MatchesAnyKey()`), which **disables every filter**. So an EF query using `IgnoreQueryFilters([])` returns *more* rows through linq2db's translation than under native EF — a silent results divergence on an edge input. (The non-empty and parameterless cases map faithfully.)

Fix: mirror EF's `Count > 0` guard — skip the call when no keys are supplied. Please confirm `EvaluateExpression<IReadOnlyCollection<string>>()` resolves across the EF TFMs; the keys arg is `[NotParameterized]` so it's a constant here. A red EF10 regression (`IgnoreQueryFilters([])` keeps filters applied) would pin it.

```suggestion
						// EF: IgnoreQueryFilters(IReadOnlyCollection<string>) → linq2db: IgnoreFilters(string[]).
						// EF treats an empty collection as a no-op (filterKeys?.Count > 0 guard in EF's
						// QueryableMethodNormalizingExpressionVisitor); linq2db's IgnoreFilters(empty) disables every
						// filter, so mirror EF and skip the call when no keys are supplied.
						var filterKeys = node.Arguments[1].EvaluateExpression<IReadOnlyCollection<string>>();
						if (filterKeys is not { Count: > 0 })
							return Visit(node.Arguments[0]);

						var keysExpr = Expression.Call(
							typeof(Enumerable), nameof(Enumerable.ToArray), [typeof(string)],
							node.Arguments[1]);
```

Source/LinqToDB/Mapping/QueryFilterAttribute.cs
MaceWindu, line 1: **Nit · NIT001** — Pre-existing XML-doc typo in the `FilterLambda` `<summary>` (line 35): the example ends with a stray `"/>`, a leftover of a malformed tag. It's outside this PR's changed hunks (so it can't be anchored as a line suggestion), but the file is open here, so a one-character cleanup can ride along.

```diff
-		/// For example (e, db) => e.IsDeleted == false "/>
+		/// For example (e, db) => e.IsDeleted == false
```

MaceWindu (commented):
> [!IMPORTANT]

Re-review at HEAD `75e8ab0d`. The named-filter implementation is solid; three independent review passes (correctness / SQL-provider / API-test) found no defects in the feature logic, and the interface/base-typed-lambda fix is correct. Two actionable items below (one is a CI-red build blocker), plus the prior-review audit and baselines.

## Review notes

- [ ] **Release build is RED on `MA0186` — see BLK001 inline.** The dedicated `build` check fails on HEAD; `test-all` is green only because it builds a non-analyzer config (Roslyn/Meziantou analyzers run in Release). The build log's *only* two errors are the `FilterIgnoreScope.Equals` overloads missing `[NotNullWhen(true)]`; fixing those (BLK001) clears the build.
- [ ] **New public surface is additive (6.4.0 minor) but not recorded in `PublicAPI.Unshipped.txt`.** `EntityQueryFilter`, `EntityDescriptor.QueryFilters`, `QueryFilterAttribute.FilterKey`, the 5 keyed `HasQueryFilter` overloads and the 2 keyed `IgnoreFilters` overloads are absent from `Source/LinqToDB/PublicAPI/**/PublicAPI.Unshipped.txt` on the branch. RS0016 isn't gating CI here, so it's not a blocker — but confirm whether the new surface needs Unshipped entries before release (the release PublicAPI step promotes Unshipped→Shipped; an empty Unshipped won't capture it). No `CompatibilitySuppressions.xml` changes needed (purely additive).

## Prior-review audit

All prior findings verified **Fixed** at HEAD; threads already resolved by the author. Nothing still actionable, no reopens.

- [x] **Interface/base-typed filter-lambda gap** (maintainer change-request + MAJ001 lambda fast path + MAJ002 func dynamic path) — fixed in `7be369d3` via the shared `BuildFilterLambda` helper. Both application paths route through it; the `IsAssignableFrom` guard + `Expression.Convert` to the declared interface/base type are correct, and the new regression tests are genuinely red-without-fix.
- [x] **MIN001** (no direct introspection test for `EntityQueryFilter` / `EntityDescriptor.QueryFilters`) — added in `b5e68321`: count/order, per-entry `FilterKey`, `FilterLambda`-vs-`FilterFunc`, legacy default-slot mirror, null-removal, derived-overrides-base.
- [x] **MIN002** (NorthwindContext pre-EF10 comment overstated multi-filter coexistence) — corrected to state the single-filter reality.
- [x] **Copilot threads** (null-key validation ordering, `GetObjectID` delimiter, `FilterIgnoreScope` set-equality + empty/null normalization, unused `using`, base→derived precedence, `HasMasterKey`→`IdRangeKey` rename) — all addressed.

## Out-of-scope observations

Surfaced during review but outside this PR's scope. FYI only — not findings on this PR.

- **`InitQueryFilters` base→derived dedup relies on attribute-instance reference identity.** The hierarchy walk subtracts inherited attributes via a `HashSet<QueryFilterAttribute>` reference-equality check (`inherited.Contains(qf)`), correct only while `MappingSchema.GetAttributes` returns the same instances when queried via the base/interface vs the derived type (cache reuse, which the code comments acknowledge). Correct at HEAD; the coupling is to cache behavior rather than a structural key. There's no clean fix through the current MappingSchema API (no `inherit:false` equivalent), and the observable precedence is already covered by `NamedFilter_DerivedOverridesBase` — noted only for awareness.
- **Interface-vs-base-class filter precedence is undocumented and untested.** For a shared `FilterKey`, an interface-declared filter overwrites a base-class-declared one (derived class wins over both). Defensible and consistent, but not pinned by a test or stated in the XML docs.

## Baselines

Delta: [linq2db.baselines#1973](https://github.com/linq2db/linq2db.baselines/pull/1973) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5525)

Independently confirmed headline (three-dot vs merge-base): **102 added / 308 modified / 0 deleted**, across EF10 (7× SqlServer, PostgreSQL 13–18, SQLite.MS) and SQLite-classic providers only — scope-appropriate. No cross-provider anomalies.

- **Modified (308)** — all EF10 `ToolsTests` / `InterceptorTests` querying `Products`, which gains two extra named filters. Each file shows the same expected shape: one single-predicate bypass guard → three AND-ed bypass guards (`ProductID > 2`, `NOT Discontinued`, `NOT IsDeleted`), each with its own `@ef_filter__pN` param — the combined-WHERE change from the PR description. Structurally uniform; no outliers. (The `IsDeleted = 0 OR IsDeleted = 0` double-predicate is a pre-existing EF10 artifact, not introduced here.)
- **Added (102)** — 3 new EF10 `ToolsTests` (`TestIgnoreQueryFilters_All_StillWorks`, `TestIgnoreQueryFilters_ByKey`, `TestNamedQueryFilter_AppliesAll`) across 14 EF10 providers + 15 new SQLite `QueryFilterTests`. `NamedFilter_InlineAssociationKeepsParentRow` confirms the INNER→LEFT-JOIN promotion (filter pushed into the `ON` clause); the `IgnoreFilters_*` baselines show correct selective suppression.

MaceWindu:
Pushed five follow-up commits to `feature/filter-category` addressing the open review findings (HEAD was unchanged since the prior review, so nothing had been pushed against them yet). Each was validated locally; red→green where applicable.

- **`97d8d72`** — `FilterIgnoreScope.Equals` (both the `IEquatable<>` and `object?` overloads) gain `[NotNullWhen(true)]`. This was the **sole** cause of the red `build` CI leg (`MA0186`, error-severity for source code; `test-all` stayed green because it builds a non-analyzer config). Release build of `LinqToDB` is clean (0/0) with the annotation.
- **`1f4d6cc`** — the EF10 bridge for `IgnoreQueryFilters(IReadOnlyCollection<string>)` now treats an **empty** collection as a no-op, matching EF Core (all filters stay applied) instead of normalizing empty→wildcard and disabling every filter. New EF10 regression test `TestIgnoreQueryFilters_Empty_IsNoOp` is red without the guard (EF returns 67 rows, linq2db 77) and green with it.
- **`d605317`** — `QueryFilterAttribute.GetObjectID` length-prefixes the user-controlled `FilterKey` so a key containing `.` can no longer collide on the id-segment boundaries (the id feeds query-cache identity).
- **`f306ac7`** — `FilterIgnoreScope.Normalize` single-element fast path now returns a fresh array (the multi-element branch already allocated one), keeping `Keys`/`Types` an immutable canonical representation.
- **`351a936`** — dropped the stray `"/>` in the `FilterLambda` XML-doc example.

On the new-public-surface point from the prior review: it does **not** need `PublicAPI.Unshipped.txt` entries — `RunApiAnalyzersDuringBuild=true` reports ~1368 pre-existing shipped symbols as unrecorded (the `PublicAPI.*.txt` files aren't the maintained API-surface mechanism here; ApiCompat / `CompatibilitySuppressions.xml` is), so the additive surface is fine as-is, consistent with the original baseline-refresh note.

The new test adds one baseline; CI will regenerate the baselines branch on this push.

Source/LinqToDB/Internal/Linq/Builder/FilterIgnoreScope.cs
MaceWindu: Fixed in `97d8d72`: `[NotNullWhen(true)]` added to both `FilterIgnoreScope.Equals` overloads (the `IEquatable<>` and `object?` forms). A Release build of `LinqToDB` is clean (0/0), clearing the red `build` leg.

Source/LinqToDB.EntityFrameworkCore/Internal/TransformExpressionVisitor.cs
MaceWindu: Fixed in `1f4d6cc`: the EF10 bridge now skips the `IgnoreQueryFilters` remap when the key collection is empty, mirroring EF's no-op semantics instead of disabling every filter. New EF10 regression test `TestIgnoreQueryFilters_Empty_IsNoOp` is red without the guard (EF returns 67 rows, linq2db 77) and green with it.

Source/LinqToDB/Mapping/QueryFilterAttribute.cs
MaceWindu: Fixed in `351a936`: the stray `"/>` was removed from the `FilterLambda` XML-doc example.

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
```diff
@@ -1581,6 +1581,56 @@ public static IQueryable<TSource> IgnoreFilters<TSource>(this IQueryable<TSource
 			return currentSource.Provider.CreateQuery<TSource>(expr);
 		}
 
+		/// <summary>
+		/// Disables query filters in current query, identified by filter key.
+		/// </summary>
+		/// <typeparam name="TSource">Source query record type.</typeparam>
+		/// <param name="source">Source query.</param>
+		/// <param name="filterKeys">Filter keys to disable. An empty array means "all keys" — disables every filter (named and anonymous) on every entity in the query.</param>
```
MaceWindu: I'm concerned about different API contract in ef.core and linq2db. EFCore treat empty/null filter list as no-op, and we as 'disable all'.
As it is new API, makes sense to unify it and introduce parameter-less IgnoreFilters() oveload to disable filters?

Source/LinqToDB/Mapping/QueryFilterAttribute.cs
```diff
@@ -9,14 +9,30 @@ namespace LinqToDB.Mapping
...
+		/// non-keyed <c>HasQueryFilter</c> overloads.
+		/// </para>
+		/// </summary>
+		public string?           FilterKey    { get; set; }
+
 		/// <summary>
 		/// Filter LambdaExpression. <code>Expression&lt;Func&lt;TEntity, IDataContext, bool&gt;&gt;</code>
 		/// <para>
-		/// For example (e, db) => e.IsDeleted == false "/>
+		/// For example (e, db) => e.IsDeleted == false
```
MaceWindu: use code block

MaceWindu (changes requested):

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
MaceWindu: https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.entityframeworkqueryableextensions.ignorequeryfilters?view=efcore-10.0

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
```diff
@@ -1581,6 +1581,39 @@ public static IQueryable<TSource> IgnoreFilters<TSource>(this IQueryable<TSource
...
+		/// </para>
+		/// </summary>
+		/// <typeparam name="TSource">Source query record type.</typeparam>
+		/// <param name="source">Source query.</param>
+		/// <param name="filterKeys">Filter keys to disable. <see langword="null"/> or empty disables nothing.</param>
+		/// <param name="entityTypes">Optional entity types the disable is scoped to. Empty means "any entity type".</param>
+		/// <returns>Query with the specified named filters disabled.</returns>
+		[LinqTunnel]
+		[Pure]
+		public static IQueryable<TSource> IgnoreFilters<TSource>(this IQueryable<TSource> source, [SqlQueryDependent] IEnumerable<string> filterKeys, [SqlQueryDependent] params Type[] entityTypes)
```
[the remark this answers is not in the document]
MaceWindu: Resolved in `cc579bf`. Rather than widen the parameter to nullable, the null-accepting wording was removed from the XML docs (the empty-list no-op stays documented). `null` remains defensively handled but isn't an encouraged input.

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
```diff
@@ -1581,6 +1581,39 @@ public static IQueryable<TSource> IgnoreFilters<TSource>(this IQueryable<TSource
...
+		{
+			ArgumentNullException.ThrowIfNull(source);
+
+			var keys  = filterKeys?.ToArray() ?? [];
+			var types = entityTypes           ?? [];
+
```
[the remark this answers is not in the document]
MaceWindu: Fixed in `cc579bf`. The overload now short-circuits and returns the source unchanged for an empty key list, so no `IgnoreFilters` call node / translation-modifier entry is injected for a documented no-op. Behavior-preserving — `QueryFilterTests` green on SQLite.Classic + SQLite.MS.

Source/LinqToDB/LinqExtensions/LinqExtensions.cs
MaceWindu: Addressed. The keyed API was reshaped in `79658cad` so empty/null keys no-op (mirroring EF Core) and disable-all is the parameterless / type-based overload; follow-up `cc579bf` removed the residual null-doc wording.

Source/LinqToDB/Mapping/QueryFilterAttribute.cs
MaceWindu: Done in `cc579bf` — the `FilterLambda` doc example is now wrapped in `<c>`.

MaceWindu (approved):
> [!IMPORTANT]

Re-review at HEAD `79658cad`, with follow-up commit `cc579bf` pushed alongside this review. The prior `CHANGES_REQUESTED` is addressed: the keyed `IgnoreFilters` API was reshaped so an empty/null key list is a no-op (mirroring EF Core), and disable-all remains the parameterless / type-based path. Three independent passes (correctness / SQL-provider / API-test) found the feature logic sound and every earlier finding fixed at HEAD. The three small items that remained are handled in the follow-up commit rather than left as review comments; details below.

## Review notes

- [x] **New public surface is additive (6.4.0 minor).** All new symbols are additive — no `CompatibilitySuppressions.xml` changes needed. The keyed `IgnoreFilters` overload was reshaped since the last review (two keyed overloads → a single `IgnoreFilters(IEnumerable<string> filterKeys, params Type[] entityTypes)`), but both prior forms were `Unshipped`-only, so there is no shipped-compat concern. New surface is recorded in `PublicAPI.Unshipped.txt`.

## Prior-review audit

Every prior finding verified **Fixed** at HEAD; the three residual items are handled in follow-up commit `cc579bf`.

- [x] **CHANGES_REQUESTED — EF/linq2db empty-list contract.** Addressed: an empty/null keyed `IgnoreFilters` now disables nothing (EF-aligned); disable-all is the parameterless / type-based overload. The distinction is load-bearing on `FilterIgnoreScope.Normalize` (Keys kept empty vs Types collapsed empty→null) — correct end-to-end and test-pinned.
- [x] **Interface/base-typed filter lambda** (change-request + prior MAJ001/MAJ002) — fixed via the shared `BuildFilterLambda` helper on both application paths; the regression tests are genuinely red-without-fix.
- [x] **Release-build MA0186, MIN001/MIN002/MIN003, and all earlier bot rounds** — fixed at HEAD.
- [x] **`filterKeys` null-doc vs non-nullable annotation.** The docs advertised a valid `null` input the signature doesn't invite. Resolved in the follow-up by removing the null wording from the docs (null stays defensively handled; the empty-list no-op stays documented).
- [x] **Empty-key no-op still injected a call node.** The follow-up short-circuits and returns the source unchanged for an empty key list, avoiding a redundant translation-modifier / query-cache entry. Behavior-preserving (empty keys already matched nothing); full `QueryFilterTests` green on SQLite.Classic + SQLite.MS (74/74).
- [x] **"use code block" (FilterLambda doc).** The follow-up wraps the example in `<c>`.

## Out-of-scope observations

Surfaced during review but outside this PR's scope. FYI only — not findings on this PR.

- **`TranslationModifier` treats the ignore-scope list as an ordered sequence for cache keying.** `.IgnoreFilters(["A"]).IgnoreFilters(["B"])` and the reverse produce two distinct compiled-query cache entries despite identical filtering. Harmless (results identical; push order is deterministic per query shape) and matches the pre-existing order-sensitive comparison. `FilterIgnoreScope` itself is order-independent internally.
- **`FilterIgnoreScope.MatchesType` uses exact-type equality, not `IsAssignableFrom`.** `IgnoreFilters(keys, typeof(BaseEntity))` scopes the disable to queries whose entity type is exactly `BaseEntity`, not derived types — preserving the pre-existing exact-match semantics. Noted because EF Core migrators (whose `IgnoreQueryFilters` is not type-scoped) may expect hierarchy matching.
- **`AssociationHelper` INNER→LEFT promotion broadened to `QueryFilters.Count > 0`.** An inline association to a func-only-filtered entity now correctly promotes to LEFT JOIN + `DefaultIfEmpty` (was INNER). Two by-design consequences: baselines move for func-only-filtered inline associations, and the promotion is decided from the cached, scope-independent association lambda, so it still emits LEFT JOIN under `IgnoreFilters` (result-benign for required FK associations). Tombstone-only descriptors (Count==0) correctly do not over-fire.

## Baselines

Delta: [linq2db.baselines#1973](https://github.com/linq2db/linq2db.baselines/pull/1973) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5525)

Independently confirmed headline (three-dot vs merge-base, after the branch was regenerated on the 2026-07-01 push): **137 added / 308 modified / 0 deleted**, across EF10 (SqlServer, PostgreSQL 13–18, SQLite.MS) and SQLite-classic providers. No cross-provider anomalies.

- **Modified (308)** — EF10 `ToolsTests` / `InterceptorTests` querying `Products`, which gains extra named filters: the single anonymous-filter predicate decomposes into three AND-ed bypass-guarded fragments, each with its own `@ef_filter__pN` param. Structurally uniform across providers.
- **Added (137)** — new EF10 `ToolsTests` (`TestIgnoreQueryFilters_*`, `TestNamedQueryFilter_AppliesAll`) + new SQLite `QueryFilterTests`; `NamedFilter_InlineAssociationKeepsParentRow` confirms the INNER→LEFT-JOIN promotion (filter pushed into the `ON` clause).
- **Not caused by this PR:** 8 of the added SQLite baselines are `MappingTests.Issue4437Test2/Test4` — those tests were added to master by #5659 and the baselines repo hasn't been refreshed for them yet (baselines-repo lag), so this PR's CI regeneration produced them fresh. One `.sql.other` (Oracle `TestVarPopOracle`) is the pre-existing #5513 carve-out.

> A preliminary automated baseline scan flagged a 57-provider `FSharpTests.Issue4132Test2` self-assignment change. On verification against the actual baseline files it is **not** present in the current merge-base diff — a stale artifact from before the branch was force-recreated on the 2026-07-01 push. No such regression exists.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Query filters can now be **named**, mirroring EF Core 10's keyed filters. New `HasQueryFilter(string filterKey, …)` overloads on the fluent `EntityMappingBuilder<T>` register multiple independent filters per entity (AND-combined at query time); passing `null` for a filter lambda (with a key) removes that filter. Named filters can also be declared **by attribute** via `QueryFilterAttribute.FilterKey`, alongside the fluent builder. The matching `IgnoreFilters(IEnumerable<string> filterKeys, params Type[] entityTypes)` overload disables filters selectively — by key, or by key×entity-type intersection — instead of all-or-nothing. Existing single (anonymous) filters and the parameterless `IgnoreFilters()` keep working.

**[LinqToDB for EntityFramework]** EF Core 10 keyed query filters and `IgnoreQueryFilters(IReadOnlyCollection<string>)` now flow through to linq2db's named query filters.


Named (keyed) query filters — `HasQueryFilter(key, …)` / `QueryFilterAttribute.FilterKey` plus selective `IgnoreFilters(keys)` — mirroring EF Core 10 (with EF Core keyed-filter flow-through).


---
<sub>Generated from commit `cc579bf`.</sub>

PR #5539, opened 2026-05-18
linq2db.cli: per-RID tool packages + nupkg size verification

MaceWindu:
## Summary

Two related changes addressing the 6.3.0 `linq2db.cli` publish failure (HTTP 413 at ~416 MB single nupkg, well above nuget.org's 250 MB upload ceiling):

1. **`Add nupkg size verification`** — pre-publish CI gate (`Build/Azure/scripts/verify-nuget-sizes.ps1` + `nuget-job.yml`). Warns at ≥180 MB, fails at ≥200 MB. Catches size growth before the push round-trips to nuget.org.

2. **`linq2db.cli: per-RID tool packages`** — replaces the custom `MultiArchPublish` + nuspec packaging with .NET SDK 10's `ToolPackageRuntimeIdentifiers`. `dotnet pack` now emits a thin pointer package + one sub-package per RID; `dotnet tool install -g linq2db.cli` auto-selects the matching variant for the user's SDK architecture.

## RID matrix shipped

`win-x64`, `win-x86`, `win-arm64`, `linux-x64`, `linux-arm64`, `osx-arm64`, `osx-x64`.

## Local pack sizes (6.3.0 dependency set)

```
   55 MB  linq2db.cli.win-x86       ✓
   95 MB  linq2db.cli.win-x64       ✓
   96 MB  linq2db.cli.win-arm64     ✓
  115 MB  linq2db.cli.linux-arm64   ✓
  122 MB  linq2db.cli.linux-x64     ✓
  156 MB  linq2db.cli.osx-arm64     ✓
  156 MB  linq2db.cli.osx-x64       ✓
    0 MB  linq2db.cli               ✓ pointer
```

All under the new 200 MB ceiling — vs the single 416 MB omnibus today.

## UX changes

- `--architecture x86 | x64 | arm64` scaffold option **removed**. The in-process restart it triggered (`RestartIfNeeded` + bundled `.win-x86.exe` / `.win-x64.exe` shims) is obsolete: users now pick architecture at install time via `dotnet tool install -g linq2db.cli --arch x86` (etc.).
- `SqlCe` and `SAP HANA` error messages updated: the "add `--architecture x86`" guidance is replaced with "install the matching x86 or x64 variant of linq2db.cli (see `dotnet linq2db help`)".
- New `Microsoft.Jet.OLEDB` detection in the Access case: when the connection string requests Jet AND the current process isn't 32-bit, the tool errors out with a clear message pointing at the x86 variant (Jet is 32-bit-only).
- Help text and `readme.md` now include a "Choosing 32-bit vs 64-bit (Windows)" section near the top covering the `--arch x86` install path and the `--tool-path` recipe for keeping x86 + x64 installs side-by-side.

## Test plan

- [x] `dotnet build` clean (0 warnings, 0 errors).
- [x] `dotnet pack` produces 7 RID sub-packages + 1 pointer package; all sizes captured above.
- [ ] Local install of the win-x64 variant: scaffold SQLite, Access, DuckDB end-to-end (this PR's reviewer flow).
- [ ] Deploy CI green.

## Post-merge follow-ups (not in this PR)

- Wiki release notes for next version mentioning the install UX change (`--arch x86` for Access x86 etc.).
- DuckDB audit (osx/linux variants are 115-156 MB — well under the 200 MB ceiling but still the largest contributors; future review of `DuckDB.NET.Data.Full` vs `DuckDB.NET.Data` + selective native asset declarations could shrink further).


viceroypenguin:
@MaceWindu Remember that the reason we chose not to go this route originally is that the default tool will install win-x64, and `dotnet tool` _does not_ allow multiple versions of the same tool to be installed at one time. For a user to install win-x86, they would have to **uninstall** win-x64 and explicitly override the default.

MaceWindu:
yes, instructions how to do it added
there is no other option as nuget.org will not accept single-nuget due to size

PR #5541, opened 2026-05-18
Respect ExpressionMethodAttribute.IsColumn=false at entity materialization

### LINKED ISSUE #5540: [v6 regression] [ExpressionMethod] substitutions fire at entity materialization regardless of IsColumn
Trying to upgrade from v5 to v6 I ran into various issues. Here's the first one.

Since v6, `[ExpressionMethod]` substitutions are applied during entity materialization regardless of `IsColumn`. [The XML doc on `ExpressionMethodAttribute.IsColumn`](https://github.com/linq2db/linq2db/blob/4c2d02ef5d994260cd377b721b7af829a5418075/Source/LinqToDB/ExpressionMethodAttribute.cs#L81-L85) says:

> *"When applied to property and set to true, Linq To DB will load data into property using expression during entity materialization."*

The implied complement — `IsColumn=false` (the default) means "don't load with the expression at materialization" — held by construction in v5: the materialization path didn't route through `ExposeExpressionVisitor`, so substitutions never fired there for either `IsColumn` value. The refactor in #3401 introduced a materialization-time call through the visitor at `TableBuilder.TableContext.MakeExpression` (`ConvertExpressionTree(fullEntity)`), and that new call expands every `[ExpressionMethod]` substitution regardless of `IsColumn`.

`IsColumn=false` is what makes `[ExpressionMethod]` safe for *query-only* rewrites: substitution bodies whose sole job is to drive server-side SQL (json_each-style table forms, navigation chains, custom `Sql.*` helpers) and aren't meant to ever execute on a materialized client value. With the v6 call site in place these bodies now fire during every entity load:

- `[Sql.TableFunction]` / `[Sql.Extension]` marker bodies throw "server-side only" — every `SELECT` against the affected entity fails.
- Non-marker bodies may run but assign incompatible values (e.g. an `IQueryable<T>` substituted into an `IList<T>` property → `InvalidCastException`).

This breaks real-world jsonb / `json_each` patterns on collection properties. The XML doc states the intended contract; the new materialization call path doesn't observe it.

myieye:
Fixes #5540

Thread a `forEntityMaterialization` flag through `ConvertExpressionTree` / `ExposeExpression` (default `false`; only `TableBuilder.TableContext` sets `true`). `ConvertExpressionMethodAttribute` returns `null` for `IsColumn=false` substitutions when the flag is set; the caller falls back to bare member access.

Four test properties needed `IsColumn = true` added — they're testing materialization-time substitution but were written without the explicit opt-in. Each is consistent with the original `CalculatedColumnTests` shape from the 2018 calculated-columns commit (495299474), where every property sets `IsColumn = true`. Affected: `QueryableAssociationTests.Issue4723Test_ExpressionMethod`, `CalculatedColumnTests.InterpolatedStringTable` (×2), and `IssueTests.Issue155Test` (fluent attribute, EFCore suite).

Regression tests in `MappingTests.cs` (`Issue5540Test1`/`2`) cover both directions: materialization skips `IsColumn=false` substitution, and the substitution still fires in `Where`/`Any`.


myieye:
@sdanyliv 

I think I fixed the failing tests by propagating my behaviour change to them - at least the pass more me locally against sqlite. My impression is that those tests really should have `IsColumn=true` assuming my understanding of the intention of that property is correct.

sdanyliv:
`ExpressionMethodAttribute.IsColumn = true` means that the property must be included in the full entity projection because it is not part of the data model and is not mapped to a database column.

myieye:
> `ExpressionMethodAttribute.IsColumn = true` means that the property must be included in the full entity projection because it is not part of the data model and is not mapped to a database column.

@sdanyliv Right, so `ExpressionMethodAttribute.IsColumn = false` means it should *not* be included in the entity projection, which is the behavior I've tried to implement.

I'm currently working around this in my project by using a duplicate property for the expression method, so I can choose to use either the property with (for JSON querying) or without (for simply selecting and deserializing the JSON).

myieye:
@sdanyliv perhaps you can retrigger tests on this branch as a baby step to maybe getting this merged in?

MaceWindu (commented):
1. Looks good, but I don't like `_calculatedColumnsOnly`/`calculatedColumnsOnly` naming - it is hard to understand intent from it. Something like `forEntityMaterialization` will be better.

~2. There is mismatch between "v6 regression" claim and test changes where we add `IsColumn = true` for old mappings. If behavior we fix here existed before v6 - merging this PR now will introduce big breaking change for users.~ verified - v6 regression

myieye:
@MaceWindu Thanks for the review! Good call, I renamed `calculatedColumnsOnly` to `forEntityMaterialization`.

One alternative worth a look: `respectIsColumn`. It names what the flag actually gates (the `!attr.IsColumn` skip) rather than the single caller that sets it — so it stays accurate if anything else ever needs the same behavior on this shared visitor, and it matches the issue/branch wording ("respect IsColumn"). I went with your suggestion since the flag only serves the entity-materialization path today, but happy to switch to the contract-named version if you'd rather.

MaceWindu (approved):

MaceWindu (changes requested):
on hold, we need to re-verify fix direction internally

MaceWindu:
superseeded by #5578

PR #1867, opened 2019-08-17
[Azure DevOps] Continue work on azure integration

MaceWindu:
- update npgsql provider dependency
- update firebird provider dependency
- fill test providers table with remaining targets
- disable automatic azure build on prs

Planned:
- try and implement support for azp commands to trigger builds manually from PRs
- add more test providers. important missing targets
  - DB2
  - Informix
  - SAP HANA2
  - Sybase ASE
  - Firebird
  - Oracle

PR #5542, opened 2026-05-18
AsQueryable: UseTempTable threshold + IDisposableTracker infrastructure

sdanyliv:
## The problem

`AsQueryable(items, db, ...)` always rendered the source as an inline `VALUES (…)` clause, and `localCollection.Contains(column)` always expanded into a flat `IN (?, ?, ?, …)` list. For collections of more than a few hundred rows both shapes produced oversized SQL, hit per-provider parameter-count limits (Oracle 1000, SQL Server ~2100 per batch, Firebird's stricter limits), blew out parse time, and tended to ruin query plans. The workaround was hand-rolling `using var tt = db.CreateTempTable(items); tt.Join(…).ToList();` boilerplate at every call site.

### Opt-in temp tables on `AsQueryable`

```csharp
var rows = LoadHotRowsFromSomewhere();   // 5_000 entries

var query =
    from row in rows.AsQueryable(db, b => b.Parameterize().UseTempTable(threshold: 100))
    join c   in db.Customers on row.CustomerId equals c.Id
    select new { c.Name, row.Amount };

var result = query.ToList();
// CREATE TEMPORARY TABLE → BulkCopy → JOIN → DROP, all behind the scenes.
```

Below the threshold the inline `VALUES` path is unchanged — small collections pay nothing. Above the threshold the collection is materialised into a real temp table via BulkCopy at execute time, joined into the main query, and dropped right after. The same `IQueryable<T>` referenced multiple times in one query (self-joins, cross-products) shares one `CREATE` / `BulkCopy` / `DROP` cycle. Scalar element types (`int[]`, `string[]`, …) work the same way.

When you plan to reuse the `IQueryable<T>` across multiple executes on the same context, tie the lifetime to the data context:

```csharp
var q = rows.AsQueryable(db, b => b.Parameterize().UseTempTable(20).DisposeWithConnection());
var a = q.Where(x => x.Active).Count();      // first execute creates and populates.
var b = q.Join(db.Customers, …).ToList();    // subsequent executes reuse the same temp table.
// Dropped when the data context is disposed.
```

`DisposeWithConnection` captures the source on first execute and reuses that snapshot on subsequent executes; it doesn't refresh from a different captured collection on the next call.

### Opt-in temp tables for `Contains`

The most common case where you'd reach for this — a large local list compared against a column via `Contains` — gets its own configuration flag so you don't need to wrap every call site in `AsQueryable`:

```csharp
    .UseSQLite(connStr)
    .UseTempTablesForContains(b => b.Threshold(100));

using var db = new DataConnection(options);

var lookup = LoadLargeIdList();   // 5,000 entries
var query  = db.Person.Where(p => lookup.Contains(p.ID));
// Above threshold: query reads from a temp table populated before the main query and dropped after.
// Below threshold: the regular flat IN list, unchanged.
```

The opt-in covers scalar element types (`int`, `long`, `Guid`, `string`, and their nullable variants) and entity lookups including composite primary keys:

```csharp
var people = LoadLargePersonList();
var query  = db.Person.Where(p => people.Contains(p));
// Above threshold: temp table populated with TempTable<Person>; predicate rewritten to EXISTS over the PK columns.
```

For entity lookups the temp table inherits the entity's `EntityDescriptor`, so any `[Column]` overrides you wrote on the entity (custom `DataType`, `ValueConverter`, etc.) propagate to both the bulk copy and the DDL — strict-typing providers like PostgreSQL see matching column types on both sides of the comparison. There's no fixed primary-key arity cap. `LikeClr` null semantics behave the same way regardless of which path fires, and `!Contains` (NOT IN / NOT EXISTS) handles correctly. The same captured list referenced multiple times in one query shares one `CREATE` / `BULK-INSERT` / `DROP` cycle.

### Compiled queries

Compiled queries play nicely with `UseTempTable`. The temp table is created on each invocation of the compiled delegate and dropped at the end, just like in a regular (non-compiled) query:

```csharp
static readonly Func<IDataContext, IEnumerable<int>, int> CountAbove =
    CompiledQuery.Compile<IDataContext, IEnumerable<int>, int>(
        (db, ids) => ids.AsQueryable(db, b => b.Parameterize().UseTempTable(threshold: 10)).Count());

var n = CountAbove(myContext, ids30);   // 30 items → temp table created, populated, used, dropped.
var m = CountAbove(myContext, ids5);    // 5 items → inline VALUES, no temp table.
```

The compiled delegate is reusable across calls with different lists — that's the whole point of `CompiledQuery`, and the temp-table machinery follows the parameter values on each invocation.

## Where it works and where it doesn't

The temp-table path is supported on **SQLite, SQL Server, PostgreSQL, and MySQL**. On providers without session-scoped runtime temp tables (Oracle, Firebird, Sybase, SAP HANA, Informix, DB2, SQL CE, DuckDB, ClickHouse), `UseTempTable(...)` is silently dropped and the chain falls through to the inline form — no hard failure, just no optimisation.

Access can't render an inline-rows source at all (every `SELECT` requires a `FROM` and `FakeTable` doesn't help), so `AsQueryable(IDataContext, …)` itself is now rejected at build time with a clear `LinqToDBException` rather than letting the provider surface a cryptic ODBC error at execute time.

On ClickHouse, inline-rows sources now render via the native `VALUES('col Type, …', (row), …)` table function instead of the verbose `SELECT col UNION ALL …` fallback — same rows, far more compact SQL, types conveyed explicitly to the planner. (Heads up: `b.Parameterize()` still hard-fails on ClickHouse — the provider doesn't accept parameters inside `VALUES` row tuples by spec.)

## What stays on the inline path

Compile-time literal arrays (`new[] { 1, 2, 3 }.Contains(col)`) stay inline by design — the rewrite only targets the parameter-backed form. Anonymous-composite shapes (`list.Contains(new { p.Id, p.Tag })`) stay inline because there's no entity descriptor to drive the temp-table machinery for an anonymous shape. Remote LinqService contexts stay inline too — temp-table creation requires a local `DataConnection` / `DataContext`.

For compiled queries specifically, `db.Table.Where(r => ids.Contains(r.Id))` with `ids` declared as a compiled-query parameter (rather than wrapped in `AsQueryable`) doesn't translate — the LINQ-to-SQL translator rejects the free parameter at build time. The `AsQueryable.UseTempTable` shape shown above is the supported path for parameterized collections inside compiled queries.

## Caveats

- Scalar string method calls (`s.Length`, `s.Substring`, …) on `AsQueryable` sources don't translate — a pre-existing limitation that affects the inline-VALUES path too; the temp-table path inherits it unchanged. Equality on the scalar value works.

## Tested

Unit tests on SQLite cover below- / above-threshold paths, filter composition, `DisposeWithConnection` reuse, self-join dedup, drop-after-query on both single-usage and self-join, scalar element types (`int[]` / `string[]`), self-join with scalars, leaked-`TempTable<T>` safety net, eager-loading + temp-table interaction, entity / composite-PK Contains with EXISTS rewrite, anonymous-composite fallback, `LikeClr` nulls on `!Contains`, cache-key isolation across `DataOptions` changes, and compiled-query interop. Cross-provider execute tests run on SQLite + SQL Server + PostgreSQL + MySQL; the full non-Oracle `AsQueryable*` and `TempTable*` suites are green. Release build is clean across `net462`, `netstandard2.0`, `net8.0`, `net9.0`, `net10.0`.

## Follow-ups

Unify the bulk-copy options API so options can be supplied here and at the `DataConnection`-options level uniformly.


MaceWindu:
make sure temp tables doesn't conflict with each other, databases have diffectent concepts about temp tables (if any)

jods4:
The idea is nice because bulk-inserting 100 items into a temp table is much better than inlining that many rows, even when `values` is supported (or worst, it falls back to `union all`).

MaceWindu is right, the execution requires caution depending on the target DB.
For example in our favorite RDBMS (Oracle of course):

- Temp tables concept exist but they must be created once upfront for all sessions, like a normal table. This is very different from SQL Server where you can `create table #temp` at any point in in any session. 
Basically Oracle is a normal table with a hidden Session ID column and rows are garbage-collected when transactions (or sessions, depending on table config.) end.
Whereas SQL Server really create a unique temporary table in tempdb storage.
- As such, they require `CREATE TABLE` DDL permission (in Oracle), which a typical application session should _not_ have.
- Like all Oracle DDL, it's not transactional. Creating that temp table would commit any existing transaction, and it would not be rolled back if the connection to DB is lost and you can't execute the `DROP TABLE` statement.

So creating on-the-fly tables is a tricky topic. I strongly suggests this feature is limited to RDBMS with "real" temporary table.

igor-tkachev:
Is this still draft or ready to review?

sdanyliv:
@igor-tkachev, missed to change PR status. It is ready for review.

igor-tkachev (changes requested):
First of all, I like the idea itself. Having automatic strategy for local data sources, where small collections use inline `VALUES` and large collections use temp tables, could be very useful.

But I think current API shape has a kind of abstraction inversion. Temp-table functionality is pushed into existing `AsQueryable` configuration, while I think it should be the opposite: temp-table API should be extended with optional inlining strategy.

`AsQueryable(db)` should stay about representing local collection as query source. If we introduce temp-table materialization, it should look like temp-table API, because now we create DB resource, use bulk copy, manage lifetime and cleanup.

Current shape makes this look like just another query rendering option:

```csharp
rows.AsQueryable(db, b => b.Parameterize().UseTempTable(threshold: 100))
```

but `UseTempTable()` is not just rendering option. It changes execution model.

I also think current builder style is too heavy for this API. For simple collection rendering options, something like this would be easier to understand:

```csharp
rows.AsQueryable(db, parameterize: true)
```

or maybe enum/options object. Fluent builder could make sense only if we really need complex per-column configuration, but adding temp-table lifetime into same builder makes API harder to reason about.

If we want this as temp-table feature, API should be closer to existing `CreateTempTable`: normal parameters/options for table name, create options, mapping, bulk copy options, etc. Collection options should stay with collection definition. Temp-table options should stay with temp-table API.

Something like this direction looks more consistent to me:

```csharp
var data = rows.AsQueryable(db, parameterize: true);

using var tmp = data.ToTempTable(
	tableName: "#MyTemp",
	threshold: 100,
	createOptions: ...,
	bulkCopyOptions: ...);
```

Exact naming and signature are discussable. Important part is separation of concepts and explicit lifetime.

Because of that, I am also concerned about `DisposeWithConnection()`. It looks like nice option to use everywhere, but it changes behavior: data is captured on first execution and reused later. This can easily produce stale results and is not obvious from the name.

Separate concern: `DisposeWithConnection()` requires new lifetime tracking infrastructure in `DataConnection` / `DataContext`. PR adds `IDisposableTracker`, makes both contexts expose it, changes dispose pipeline and makes `TempTable` auto-register itself there.

I think this is too big conceptual change for this feature. If we can solve the problem without adding new ownership model to `DataConnection` / `DataContext`, we should prefer that. Those types are already central and complex enough.

With explicit `ToTempTable(...)` / `CreateTempTable`-like API we could stay within existing concepts: temp table is a resource, user controls lifetime with `using`, and this feature just extends temp-table capabilities instead of changing general resource ownership model.

sdanyliv:
Thanks for the deep look — replying to each point.

## Three separate features, not one chain extension

The PR adds three distinct things that all happen to involve temp tables, but they serve different use cases. I think conflating them is the root of the "abstraction inversion" reading.

| Feature | User intent | Who owns the temp table |
|---|---|---|
| `CreateTempTable<T>` *(existing, unchanged)* | "Give me a temp table I manage explicitly with `using`." | User |
| `AsQueryable(db, b => b.UseTempTable(N))` *(new)* | "Use VALUES below N rows, temp-table above. Don't make me restructure." | Runtime |
| `UseTempTablesForContains` *(new DataOption)* | "Rewrite `Contains(largeList)` predicates above threshold." User never sees the temp table. | Runtime |

The proposed alternative `using var tmp = data.ToTempTable(threshold: 100, …)` is essentially today's `CreateTempTable<T>` plus a row-count gate. That's a fine feature too, but it's not what this PR is offering — the value here is the *transparent* opt-in. The user keeps writing `rows.AsQueryable(db, ...)` or `Where(p => list.Contains(p.Id))` and the runtime decides whether to inline or materialise based on the runtime row count. Restructuring the call site defeats the point.

This is especially true for the `Contains` rewrite: there's no point in the query lifecycle where the user could hold a `using` over the temp table — it's created and dropped inside one query execution. Runtime-managed cleanup is mandatory there, not optional.

## Interplay with `AsKeyedQuery` (#5450)

This is the concrete example of why the transparent shape matters. `AsKeyedQuery` materialises parent keys as a `VALUES (...)` table and joins child rows against it — exactly the shape that benefits from this PR's `AsQueryable.UseTempTable` machinery. Configure a `LocalCollections` threshold once at the `DataOptions` level, and eager-loading via `AsKeyedQuery` automatically flips its parent-key `VALUES` to a temp table above threshold. No changes inside `AsKeyedQuery` itself, no new ownership for the caller. Eager-loading a few thousand parents stops producing oversized SQL and stops hitting per-provider parameter-count limits.

Under the proposed alternative shape, `AsKeyedQuery` would have to either (a) build the temp table itself via `ToTempTable(...)` and wire its own lifetime, or (b) leave large parent-key sets on the broken inline `VALUES` path. Neither is good. The transparent threshold is what makes one config flag propagate the optimisation across every consumer that emits a `VALUES`-table source — whether the user wrote `AsQueryable(...)` directly, `AsKeyedQuery` wrote it during eager loading, or a future feature emits one.

## On the chain-builder vs. flag-args shape

The chain already carries `Parameterize() / Inline() / Except(…) / UseTempTable(…) / ConfigureBulkCopy(…)`. Flag args (`parameterize: true`) work for two-flag cases but break down past that, and `ConfigureBulkCopy` is itself a nested builder for bulk-copy options. The fluent shape is consistent with what `AsQueryable` already does today; switching now would split the API style.

## On `DisposeWithConnection()` — point taken

The first-execute-snapshot-then-reuse behaviour is genuinely a sharp edge and the method name doesn't telegraph it. Concrete options:

1. Rename to `ReuseAcrossExecutes()` or `KeepForContextLifetime()` — drops the "dispose" framing, leads with the actual lifetime behaviour.
2. Or split into two: `KeepForContextLifetime()` for the long-lived case and leave the default per-execute behaviour as the only shape on `UseTempTable`. The "dispose" naming was inherited from `TempTable<T>.DisposeWithConnection` — that's where the name came from, but you're right that the implied semantic is different here.

Happy to go with whichever you prefer.

## On `IDisposableTracker`

The tracker is only exercised by (a) the `DisposeWithConnection()` lifetime mode, and (b) safety-net cleanup of forgotten `TempTable<T>` instances. It doesn't introduce ownership semantics for general user resources; user code never registers with it.

That said — your worry that adding any public surface to `DataConnection` / `DataContext` raises the conceptual weight of those types is fair. Concrete options:

1. Move `IDisposableTracker` and the `IInfrastructure<IDisposableTracker>` exposure to `LinqToDB.Internal.Common` (internal surface). The runtime still uses it, but it's no longer part of the public model. The only public effect is `DisposeWithConnection()` working — which the user reaches via the chain method, not via direct tracker access.
2. Expose the tracker only on `DataConnection` (skip `DataContext`). `DataContext` already recycles its underlying connection per operation — `DisposeWithConnection()` semantics there are murky anyway (per a comment in `DataContext.cs:542`) and we could just reject the chain method on `DataContext` at build time.

I'd lean toward (1) — keeps the safety net for forgotten `TempTable<T>` while dropping the public surface.

## Summary of what I'd commit to if these land

- Rename `DisposeWithConnection()` → `KeepForContextLifetime()` (or your preferred name) with XmlDoc that leads with "first-execute snapshot, reused across executes".
- Move `IDisposableTracker` to internal namespace. Public users continue to use the chain method; the tracker is just plumbing.

That removes the two real concerns. The three-feature framing is the harder design call — let me know whether the table above lands or whether you want to dig in further.

igor-tkachev (changes requested):
Thanks, but I think this is the core disagreement.

As I said before, I like the idea itself. Using a better strategy for large local data sets is useful. I disagree with the direction where local collection APIs are promoted to temp tables as hidden implementation detail.

> the value here is the *transparent* opt-in

For linq2db this is not a value. linq2db is about explicit control. **We should not make a simple local-collection API secretly depend on a more complex database-resource model.** Opt-in doesn't make hidden ownership model obvious.

Local array/list and temp table are not just two SQL renderings. Temp table is a database resource. It has DDL, bulk copy, connection/session lifetime, cleanup, permissions, transaction/provider-specific behavior, different failure modes and configuration that cannot be fully hidden: column types, indexes, table options, bulk copy options, etc.

**Global DataOptions switch is the most dangerous part for me. It opens Pandora's box.** It can change behavior of many queries in the whole application without changing those query call sites. One option can make unrelated `AsQueryable(db)` / `Contains` usages start creating temp tables in unpredictable places.

`Contains` is the worst example of this. `list.Contains(x)` is a simple local-collection predicate. With `UseTempTablesForContains`, this predicate can become temp-table workflow based on runtime size and global options. `Contains` can use temp tables, this is not the question. The question is that this behavior is not visible at the call site.

And those temp tables will be created with default configuration in many cases: default column types, for example string columns can become provider defaults like `nvarchar(4000)`, no indexes, default bulk copy options, etc. So instead of performance improvement we can easily get worse plans, worse joins, more IO, collation/type issues, or permission/lifetime failures. This is too much power for a global option, because the new behavior is not just SQL formatting. It creates database resources.

It also creates action-at-a-distance maintenance risk. I can write and test many `Contains` queries, deploy them, and they work. Later another developer enables this global option to optimize his own query. Test suite can still be green because thresholds/data/providers in tests don't hit problematic path. Then my unrelated query starts failing in production because it now creates temp tables. The change is not visible in my code, but my code pays the cost. Then we get questions like: why did this suddenly stop working in production?

**This is not transparency. It is visible syntax with hidden semantics.**

**The direction should be opposite. Temp table should be explicit higher-level abstraction.** If user selected temp-table workflow, then small data can be optimized to `IN` / `VALUES`. This is safe direction: replace complex resource with simpler representation. The reverse direction is abstraction inversion: ordinary local collection becomes database resource with side effects.

About `AsKeyedQuery`: I think this mixes internal implementation strategy with public API shape. If eager loading internally needs temp tables or another scalable key-set representation, linq2db can choose it internally because library owns the whole flow. Good internal implementation detail doesn't automatically become good public API, especially when public ownership becomes unclear.

About `DisposeWithConnection()` / `KeepForContextLifetime()`: rename doesn't solve the problem. It is still the same model: first hide temp table behind collection syntax, then add indirect switches to control hidden resource.

This is the wrong level of control. If user wants reuse, lifetime, and cleanup control, the temp table should be explicit. Then it is clear when data is loaded, when table is reused, and when it is dropped. With hidden temp table we need extra non-obvious switches and documentation to explain behavior that would be obvious with explicit resource ownership.

`IDisposableTracker` looks like infrastructure added to compensate this design. Once temp table is hidden, we need hidden ownership rules: per-query vs context lifetime, `Close()` vs `Dispose()`, `DataConnection` vs `DataContext`, forgotten table cleanup, fallback when tracking is unavailable. If temp table is explicit, most of this is already controlled by existing temp-table lifetime model.

And the minimal alternative is almost the same code.

Current proposal:

```csharp
var q = rows.AsQueryable(db, b => b.Parameterize().UseTempTable(threshold: 100));
```

One possible explicit resource shape:

```csharp
using var t = rows.AsQueryable(db, parameterize: true).ToTempTable(threshold: 100);
```

All this discussion is mostly about this `using`.

**With `using`, syntax and semantics match: the code shows that this is a resource with explicit lifetime.**

`AsQueryable` stays responsible for representing local collection as query source and configuring local value rendering. `ToTempTable` is the explicit boundary where database-resource semantics start. From that point user can see ownership and lifetime.

This API can still optimize small inputs to `IN` / `VALUES` internally. But it starts from explicit temp-table workflow and optimizes down to simpler representation. It doesn't start from ordinary local collection usage and promote it up to hidden database resource.

**So we are adding hidden ownership model, DataOptions switches and cleanup infrastructure mostly to avoid one `using`. But this `using` is exactly what makes the complex resource explicit and solves the main design problem.**

igor-tkachev:
Separate API concern: builder pattern here is too much ceremony.

Builder is not a default API shape. It is a heavy tool. It is justified only when the alternative is even heavier: many options, complex staged DSL, or validation of allowed call sequences.

We have places where it makes sense: FluentMapping, Merge API, Window API. But even there it has a cost. Usually you need documentation to write correct chain. IntelliSense doesn't save the situation, because it still shows many available methods and doesn't explain the intended sequence.

Here we use builder for a small set of regular options. This feels like using builder because builder is available, not because this API really needs it. For this kind of API normal parameters or options records are easier to read, easier to discover, and harder to misuse.

With the current shape, once we add bulk-copy options, the API naturally moves in this direction:

```csharp
rows.AsQueryable(
    db,
    b => b.Parameterize().UseTempTable(
        tt => tt.Threshold(100).ConfigureBulkCopy(
            bc => bc.UseMultiRows(
                mr => mr.WithMaxBatchSize(5000).WithBulkCopyTimeout(60)))))
```

This is not a good direction. It is very easy for useful fluent API to turn into an absurd show of nested lambdas and configuration ceremony. We should stop before that.

Also, using expression tree for this configuration makes it even heavier. Configuration is not a query expression. It is just options. Passing it as `Expression<Func<...>>`, parsing method-call chain manually, and compiling nested configuration lambda back into delegate is too much machinery for regular options.

We already have option types in this area: `CreateTempTableOptions`, `BulkCopyOptions`, `TableOptions`. Builders eventually produce the same kind of data anyway. If the final result is `BulkCopyOptions`, accepting `BulkCopyOptions` directly is simpler and more consistent.

Builder pattern can be useful, but it is hard to use and should be justified by real complexity. Here it adds custom DSL around simple options without enough benefit.

viceroypenguin:
I'm with @igor-tkachev on this one. I see the value in "auto-temp-table" for a single query, since there's no lifetime concern. For multiple queries, I like igor's proposal of `using var t = rows.AsQueryable(db, parameterize: true).ToTempTable(threshold: 100);`, where there's explicit resource management, and the temp table options can be isolated to a dedicated function for it, and the table can be explicitly disposed, as it can be today. The hidden ownership using the DC adds too much mental overhead for me, if I leave a function, do I need to remember if that temp table got dropped or not, etc.

igor-tkachev:
I thought more about `Contains` and local collections, and I think we are mixing two different features here.

The first feature is query-scoped local input. It works well for `Contains`: we mark local collection as input for current query execution, and linq2db can choose `IN`, `VALUES` or temp table. With `let` we can also reuse same input several times in one query. Lifetime is owned by linq2db and bound to one query execution.

The second feature is explicit temp table resource. User creates it, uses it in one or more queries, and controls lifetime with `using`. This feature can also optimize small data to `VALUES` instead of creating real temp table.

I think this second part is important by itself. Current temp table API is old and already somewhat overloaded after 10 years of evolution. This could be a good chance to introduce cleaner new temp table API without breaking compatibility. Many users could prefer it over old API, especially if it also gives additional `VALUES` fallback for small data.

Those features can probably share internal implementation, but they have different lifetime and ownership semantics, so I think we should not expose them as one public API. If we split them, design should be simpler and more predictable.

sdanyliv:
Thanks for the detailed round, and I do hear the disagreement. Let me try to narrow it.

## What the PR is actually proposing

I don't think the PR is promoting a local collection to a hidden database resource. It's letting the runtime pick the most efficient *representation* for the same LINQ — `VALUES (…)` below threshold, temp table above — with two explicit knobs that the user controls: the threshold (per-call or global) and the lifetime (default per-execute, or `DisposeWithConnection()` for context-scoped).

The collection in the user's hand stays a plain `IEnumerable<T>`. The user never receives a temp-table handle, never has to call `using`, never has to thread anything across methods. That is the goal — not to hide a resource, but to keep the call site looking like LINQ.

## Reuse cases — what "transparent" actually means

**Within one query: same source, one temp table, deterministic lifetime.**

```csharp
var hotIds = LoadHotIds();   // 5,000 entries

// AsQueryable as a reusable IQueryable<int> source — referenced twice in one query
var hotQry = hotIds.AsQueryable(db, b => b.Parameterize().UseTempTable(threshold: 100));

var pairs = from a in hotQry
            from b in hotQry
            where a < b
            select new { a, b };

var rows = pairs.ToList();
```

Emitted SQL: **one** `CREATE TEMPORARY TABLE [T_xxxxxxxx]` → one `BulkCopy` of `hotIds` → main query joins against `T_xxxxxxxx` twice (as `t1` and `t2`) → **one** `DROP` after `ToList()` returns. The lifetime is bounded by exactly one query execution. This is covered by `AsQueryable_UseTempTable_SelfJoin_SharesSingleTempTable` in the test suite; the assertion is exactly "one distinct `[T_xxxxxxxx]` name in the emitted SQL".

The dedup happens at `ExpressionBuilder.GetOrAssignTempTableName` — keyed by source-expression equality, so any number of `hotQry` references inside one query collapse to one temp-table name and one Setup/Teardown cycle.

**Across methods: same source, untouched method signatures.**

```csharp
// Service methods — plain IEnumerable<int> signatures
public IQueryable<Person> ActiveCustomersFromList(IEnumerable<int> ids) =>
    db.Person.Where(p => ids.Contains(p.ID) && p.IsActive);

public IQueryable<Order> OrdersFromList(IEnumerable<int> ids) =>
    db.Order.Where(o => ids.Contains(o.PersonID));

// Caller — plain List<int>
var lookup    = LoadLargeIdList();   // 5,000 entries
var customers = ActiveCustomersFromList(lookup).ToList();
var orders    = OrdersFromList(lookup).ToList();
```

With `db.Options.UseTempTablesForContains(b => b.Threshold(100))` set once at startup, both queries automatically use the temp-table path above threshold. **Service-method signatures stay `IEnumerable<int>`** — they compose like normal LINQ.

Under the proposed `using var t = lookup.ToTempTable(threshold: 100)` shape, the caller has to construct + own the temp table, and *every method that touches the collection* changes its signature from `IEnumerable<int>` to `ITable<int>` (or wrapper). Helpers, third-party libs, dapper-style loaders — all become incompatible without changes. That's an intrusive cross-cutting refactor for an optimisation that should be invisible to the service layer.

`AsQueryable.UseTempTable(b => b.DisposeWithConnection())` covers the case where the same captured collection is referenced across multiple query executions in one data context — the temp table is created on first execute and reused. Without a runtime-managed cleanup mechanism, expressing that lifetime is impossible — the caller would have to thread `using var t = …` across every method that touches the collection.

## DataOptions opt-ins like this exist already

This is not a new pattern in linq2db. There are already at least five `DataOptions` / `LinqOptions` flags where one global setting changes the SQL emitted for the same LINQ across the whole context:

| Option | Same LINQ, different SQL |
|---|---|
| `CompareNulls` (`LikeClr` vs `LikeSql`) | equality against nullable columns becomes `IS NULL OR col = …` vs `col = …` |
| `ConcatenateOrderBy` | second `OrderBy` appends vs replaces |
| `GuardGrouping` | `GroupBy` throws vs executes with eager loading |
| `PreferExistsForScalar` | `Contains` becomes `EXISTS (SELECT…)` instead of `IN (…)` |
| `OptimizeJoins` | redundant JOINs removed vs preserved |

`PreferExistsForScalar` is the closest analogue — a global flag that rewrites `Contains` predicates app-wide. It ships today. `UseTempTablesForContains` is the same design pattern, not a new "Pandora's box". If the precedent is acceptable, so is the addition.

All of these flags are part of `ConfigurationID` — flipping any of them invalidates the LINQ-query cache and re-translates fresh. The "test passes locally, breaks in production after a colleague flips a flag" scenario applies identically to all of them. It's a real concern for any global SQL-shaping option, but it's not specific to this feature — and the existing design handles it the same way the others do.

## On default temp-table column types and indexes

For entity Contains (`personList.Contains(p)` where `Person` is a registered entity), the temp table is `TempTable<Person>` and inherits the entity's `EntityDescriptor`. Every `[Column(DataType=…)]`, `ValueConverter`, `[PrimaryKey]` etc. propagates to the DDL and BulkCopy — the temp table's column types match the live table's column types. No `nvarchar(4000)` defaults; the user's mapping is honored.

For scalar Contains, the column type comes from the element type's `MappingSchema` lookup — same path used for parameter binding. For composite-PK Contains, same as entity Contains (the entity's PK descriptors drive the temp-table shape).

The only path where "provider defaults" enter is the anonymous-composite shape (`list.Contains(new { x.Id, x.Tag })`), which intentionally falls back to OR-AND inline today — no temp table is created.

Per-call `ConfigureBulkCopy(…)` is available where the user wants to tune bulk-copy options. So the configuration surface for the cases that matter is fully under user control.

## On `AsKeyedQuery`

`AsKeyedQuery` is part of the **public** API in PR #5450 (`AsKeyedQuery()` / `AsSeparateQuery()` extension methods on `IQueryable<T>` plus `LinqOptions.DefaultEagerLoadingStrategy`). The user invokes it explicitly, and its emitted SQL contains parent-key `VALUES` sources that the `UseTempTable` machinery automatically optimises above threshold.

If `AsKeyedQuery` had to hard-code a temp-table threshold internally — without going through the general `UseTempTable` path — we'd have two places deciding when to materialise, no consistent default, and no user override. The transparent optimisation is what makes one mechanism cover every consumer that emits a `VALUES`-shaped source.

The reverse also holds: `AsKeyedQuery` can adopt the same configure-builder shape this PR introduces. Extending its signature to `query.AsKeyedQuery(c => c.Parameterize().Except(p => p.SomeCol).UseTempTable(threshold: 200))` is a small step — `IAsQueryableBuilder<T>` / `IAsQueryableExceptBuilder<T>` aren't `AsQueryable`-specific, they're general configure interfaces. A user who wants to direct parent-key handling for one specific eager-load chain can do so explicitly on that call site, reusing the exact builders this PR defines, without touching global `DataOptions`. The chain isn't an `AsQueryable` accessory — it's a reusable contract for any feature that emits a parameterised IEnumerable source.

## On parser / compile cost

The `Expression<Func<…>>` configure chain is parsed once per query translation in `EnumerableBuilder.TryParseConfigure` (~25 lines, walks the method-call chain). The inner `UseTempTable(b => …)` lambda is compiled once at translation, invoked once against the builder impl to snapshot the `TempTableSpec`, then discarded. The translated query is cached by `ExpressionQuery.GetQuery` keyed on `MainExpression` (which includes the quoted `configure`) — repeated calls with the same lambda hit the cache and skip parsing entirely. Per-execute cost is effectively zero.

I can produce a microbenchmark later if you want concrete numbers. The reason the outer shape is `Expression<Func<…>>` is `Except(p => p.Id, p => p.Data)` — that genuinely needs member-access reflection. The other chain methods inherit the shape; pulling them out to plain method overloads would fragment the API for no measurable per-execute win.

## On builder ceremony

The current chain is two methods (`b.Parameterize().UseTempTable(100)`), three with `Except`, four if the user explicitly tunes bulk copy. The worst-case nested-lambda example you sketched isn't what the PR ships and isn't the direction it's growing. `ConfigureBulkCopy(…)` is a single nested builder, used only by callers who actually need to tune bulk copy — most won't.

If the bulk-copy nesting becomes a real concern, we can also accept a `BulkCopyOptions` instance directly as an alternative overload, sitting alongside the fluent form. That's a small additive change inside the existing shape, not a redesign.

The flag-args alternative isn't just a style preference — it's strictly less expressive than the current chain. The existing API lets you mix-and-match: `b.Parameterize().Except(p => p.Id, p => p.CreatedDate)` says "parameterize everything **except** these two columns; those go inline as literals." A `parameterize: true` flag can only flip the global default; there's no per-field override. To get the same per-field exclusion you'd need a second parameter — `parameterize: true, except: ...` — and the flag-args shape ends up with the same surface area as the chain, just less discoverable and harder to grow.

The `.ToTempTable(threshold: 100)` half of the proposed shape has a separate soundness problem. `ToTempTable` would have to be an extension on `IQueryable<T>` generally — so it compiles against any `IQueryable<T>`, including a real LINQ query like `db.Person.Where(p => p.Active).ToTempTable(threshold: 100)` where "threshold" is meaningless and behaviour at execute time is undefined. The current `b.UseTempTable(...)` lives inside the `AsQueryable` configure chain precisely because it's only valid in that context — the type system rejects misuse. Moving it to a free extension method loses that guarantee.

## On the existing concessions

The two concessions I offered in round 1 still stand:

- Rename `DisposeWithConnection()` → `KeepForContextLifetime()` (or your preferred name), with XmlDoc that leads with "first-execute snapshot, reused across executes".
- Move `IDisposableTracker` to internal namespace. Public users continue to use the chain method; the tracker becomes plumbing.

These remove the public-surface exposure that motivated most of the IDisposableTracker concern and the "rename doesn't capture the semantics" objection on `DisposeWithConnection()`.

## What would make this acceptable?

The PR follows the same design pattern as `PreferExistsForScalar` and four other existing `LinqOptions` flags — a global SQL-shaping opt-in with per-query override where applicable. If the principle is acceptable for those, I'd like to understand what specific scenarios would convince you this addition is OK:

- Specific documentation (e.g. a clear "when to enable" / "when not to" note in the XmlDoc and release notes)?
- A test that verifies the cache invalidates on flag flip (already true via `ConfigurationID`, easy to add)?
- A per-query opt-out marker for `Contains` so a specific call site can disable the global default?
- Something else?

A concrete list of deliverables would be more actionable than the philosophical framing — the design itself is consistent with linq2db's existing pattern for this kind of optimisation.

igor-tkachev:
Thanks, I think we are much closer now.

Just to clarify my final position: I like the feature itself. Large local collections are a real problem, and having a better strategy than huge `IN` lists or inline `VALUES` is useful.

But I strongly prefer explicit lifetime control for anything that can create temporary tables.

I see three separate modes here.

First, `AsQueryable(...).UseTempTable(...)`. If this creates a temp table, it should be query-scoped only: create, use, drop within one query execution. This model is clear and predictable.

Second, explicit temp-table-like resource controlled by user code, through `using`. This can support reuse across several queries, and for small data it can optimize down to `VALUES` instead of creating a physical temp table. This could be a new API, or maybe an extension of existing temp table API if it fits cleanly.

Third, `Contains`. Instead of global flags, I think there is a safer and cleaner alternative,  **please read this part to the end** before assuming I mean explicit `ToTempTable` everywhere. I mean a separate disposable controller/scope that explicitly enables this behavior and owns all temp tables created for local collections inside that scope.

```csharp
using var localCollections = db.CreateLocalCollectionScope(options);

var q1 = db.Person.Where(p => ids.Contains(p.ID));
var q2 = db.Order.Where(o => ids.Contains(o.PersonID));
```

The important part is that the scope, not base `DataContext` / `DataConnection`, controls lifetime and ownership of temp tables created for `Contains`. When the scope is disposed, all temp tables registered in it are dropped.

If some brave soul wants behavior similar to global flags, they can create this controller in their custom `DataContext` constructor and dispose it together with the context. But I don't think base `DataContext` / `DataConnection` should silently become temp-table lifetime controllers for everyone.

This is especially risky because context/connection lifetime is not always short or obvious. In some data-processing scenarios a context can live for hours. With DI, contexts can be created, cached, reused or wrapped by different components in ways that are not visible at the query call site. If temp tables are implicitly attached to that lifetime, we can get hidden state, stale snapshots, name conflicts, cleanup issues and production-only bugs.

Those types are already central and overloaded. Adding hidden temp-table ownership to them makes them even more complex and weakens SRP. If an application needs this model, it can compose it explicitly.

So my preference is:

* query-scoped temp tables for `AsQueryable(...).UseTempTable(...)`;
* explicit `using`-controlled temp-table-like resource for reuse across queries, possibly with `VALUES` fallback;
* explicit disposable controller/scope for `Contains`, instead of global flags;
* no hidden lifetime semantics in base context/connection.

The feature is good, but this area needs full control and visible ownership.

igor-tkachev:
And, yes, I still hate builders :)

Source/LinqToDB/TempTableSpec.cs
```diff
@@ -0,0 +1,30 @@
...
+	/// the single query execution.
+	/// </param>
+	/// <param name="BulkCopyOptions">
+	/// Options forwarded to the <c>TempTable&lt;T&gt;</c> bulk-copy call when the temp table is
+	/// materialised; <see langword="null"/> means "use provider default".
+	/// </param>
+	public sealed record TempTableSpec(
+		int?             Threshold,
+		bool             DisposeWithConnection,
+		BulkCopyOptions? BulkCopyOptions);
```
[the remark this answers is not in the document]
MaceWindu: Inaccurate at HEAD per this repo's release workflow: `PublicAPI.Shipped.txt` / `PublicAPI.Unshipped.txt` are allowed to lag source within a release cycle and are reconciled by maintainers at the release boundary, so the RS0016/RS0017 lag is accepted until then. The new public types do need to land in `PublicAPI.Unshipped.txt` before 6.4.0 ships, but that isn't a change this PR's diff has to carry.

Source/LinqToDB/Internal/Linq/ExpressionQuery.cs
```diff
@@ -228,7 +255,7 @@ public async Task GetForEachUntilAsync(Func<T,bool> func, CancellationToken canc
 			if (!dependsOnParameters)
 				Expression = expressions.MainExpression;
 
-			var enumerable = (IAsyncEnumerable<T>)query.GetResultEnumerable(DataContext, expressions, Parameters, Preambles);
+			var enumerable = (IAsyncEnumerable<T>)query.GetResultEnumerable(DataContext, expressions, Parameters, Preambles, null);
```
MaceWindu, line 258: **Blocker · BLK001** — `GetForEachUntilAsync` is the one execute path that never wires in the temp-table plumbing. Unlike `GetForEachAsync` (this file, ~219) it doesn't create a `QueryExecutionContext`, never calls `InitQueriesAsync`, passes `null` as `execContext`, and reuses the stale `Preambles` field. `ForEachUntilAsync` is a public async terminator (it also backs `SelectAsync`). When a query that opts into the feature — `AsQueryable(...).UseTempTable(n)`, or a `UseTempTablesForContains` `Contains` over a large local collection — is consumed through this path above the threshold, the temp-table Setup run-step never fires, so no temp table is created; yet with `execContext == null` the count-vs-threshold fallback in `BasicSqlBuilder.BuildSqlValuesTable` still emits a reference to the (non-existent) temp table, giving a runtime "no such table" SQL error at execute time. That's exactly the cryptic execute-time failure the PR added build-time rejection for on Access — but shipped enabled here. The comment at `BasicSqlBuilder.cs:1947-1949` already names `ForEachUntilAsync` as a path that "produce[s] SQL referencing a non-existent table".

Fix: Mirror `GetForEachAsync` — create a `QueryExecutionContext`, `await query.InitQueriesAsync(...)` before `StartLoadTransactionAsync`, re-init `Preambles` via `InitPreamblesAsync`, thread the context into `GetResultEnumerable`, and dispose/teardown it (or wrap the enumerator in `QueryRunStepTeardownAsyncEnumerator`). If covering this path is out of scope for now, make `UseTempTable`-configured queries fail loudly at build time on this path rather than emitting broken SQL.

Source/LinqToDB/Internal/Linq/QueryExecutionContext.cs
```diff
@@ -0,0 +1,178 @@
...
+				catch (Exception ex)
+				{
+					(errors ??= new()).Add(ex);
+				}
+			}
+
+			_activeSteps = null;
+
+			if (errors is { Count: > 0 })
+				throw new AggregateException(errors);
```
MaceWindu, line 146: **Major · MAJ001** — `QueryExecutionContext.Dispose` runs as teardown in the `finally` / enumerator-`Dispose` path after the query body. When the query itself threw (SQL error mid-read, cancellation, mapping failure) and a teardown step also fails (e.g. `DROP TABLE` errors), this `throw new AggregateException(errors)` replaces the in-flight primary exception — the caller sees the teardown failure, not the real error. Even on a fully successful query, a `DROP`-time failure surfaces as an `AggregateException` from enumerator disposal that callers never expect. The sibling `DisposableTracker.DisposeAll`/`DisposeAllAsync` deliberately swallows per-resource teardown exceptions for exactly this reason; the two teardown paths are inconsistent.

Fix: Match the `DisposableTracker` policy — don't throw from teardown; swallow (optionally record via the existing trace/diagnostics channel) so the primary query exception propagates intact. The suggestion deletes the rethrow; the `errors` accumulation above can be dropped or repurposed to a trace call.

```suggestion
```

Source/LinqToDB/Internal/Linq/QueryExecutionContext.cs
```diff
@@ -0,0 +1,178 @@
...
```
MaceWindu, line 175: **Major · MAJ002** — Async counterpart of MAJ001. `DisposeAsync` is the teardown that runs from `QueryRunStepTeardownAsyncEnumerator.DisposeAsync` (and the `await using` execute paths). A teardown `DROP` failure here throws `AggregateException` from the enumerator's `DisposeAsync`, masking any primary query exception that's unwinding and surprising callers of an otherwise-successful async enumeration. Inconsistent with `DisposableTracker.DisposeAllAsync`, which swallows per-resource teardown failures.

Fix: Same as MAJ001 — don't throw from async teardown; swallow (or record). The suggestion deletes the rethrow.

```suggestion
```

Source/LinqToDB/IDisposableTracker.cs
```diff
@@ -0,0 +1,43 @@
...
+	/// Tracker for lifetime-bound disposable resources owned by a data context — typically temp
+	/// tables created by <see cref="LinqExtensions.AsQueryable{TElement}(System.Collections.Generic.IEnumerable{TElement},IDataContext,System.Linq.Expressions.Expression{System.Func{Linq.IAsQueryableBuilder{TElement},Linq.IAsQueryableExceptBuilder{TElement}}})"/>
+	/// when the configure chain includes <c>UseTempTable(b =&gt; b.Threshold(N).DisposeWithConnection())</c>.
+	/// A data context exposes its tracker via <see cref="IInfrastructure{T}"/>:
+	/// <c>((IInfrastructure&lt;IDisposableTracker&gt;)dc).Instance</c>. Registered resources are
+	/// released by the context's Dispose pipeline; soft <c>Close()</c> intentionally does not
+	/// drain (a close-then-reuse cycle on the data context preserves user-owned temp tables).
+	/// One bad resource does not block disposal of the others.
+	/// </summary>
+	public interface IDisposableTracker
```
MaceWindu, line 18: **Major · MAJ003** — `IDisposableTracker` is new public surface in the root `LinqToDB` namespace, but it's pure lifetime-tracking infrastructure: its only consumers are inside the library (`TempTable<T>` auto-registers; `DataConnection`/`DataContext` drain it on dispose), and the sole user-reachable seam is the explicit `IInfrastructure<IDisposableTracker>.Instance` implementation — there's no user-facing factory or `new`. Public API is a permanent stability contract, so an infrastructure interface with only internal callers should live in `LinqToDB.Internal.*`. The author has already agreed to move it internal, and it's part of the maintainer's CHANGES_REQUESTED objection.

Fix: Move `IDisposableTracker` to `LinqToDB.Internal.*` (keep the `IInfrastructure<IDisposableTracker>` seam if the runtime needs it). Public users continue to reach the behavior via the `UseTempTable(b => b.DisposeWithConnection())` chain method; the tracker interface itself never needs to be public.

Source/LinqToDB/ITempTableConfigBuilder.cs
```diff
@@ -0,0 +1,48 @@
...
+		/// refreshed from a possibly-different captured collection). The table is dropped when
+		/// the data context is disposed — soft <c>Close()</c> intentionally does not drain.
+		/// <para>
+		/// When the context does not expose <see cref="IDisposableTracker"/> via
+		/// <see cref="LinqToDB.Internal.Infrastructure.IInfrastructure{T}"/>, the option silently
+		/// degrades to per-query disposal: the table is dropped after the query completes as if
+		/// this chain method had not been called.
+		/// </para>
+		/// </summary>
+		ITempTableConfigBuilder DisposeWithConnection();
```
MaceWindu, line 36: **Suggestion · SUG001** — `DisposeWithConnection()` reads as "control when the table is dropped", but its actual semantics are "capture the source data on first execute and reuse that snapshot across all later executes of the same `IQueryable`" — a first-execute snapshot that can silently return stale results. The author has already agreed to rename it to `KeepForContextLifetime()` and lead the XML doc with the snapshot-reuse behavior.

Fix: Rename `DisposeWithConnection()` → `KeepForContextLifetime()` (interface + `TempTableConfigBuilderImpl` + the `TempTableSpec` field + the `EnumerableBuilder` merge + tests) and rewrite the XML doc to lead with "first-execute snapshot, reused across executes" so the staleness implication is obvious. Spans several files, so it isn't a one-click suggestion.

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -1933,7 +1934,65 @@ protected virtual void BuildFromExtensions(SelectQuery selectQuery)
...
+			// registered the init-query at build time; the init-query ran in
+			// ExpressionQuery.InitQueries before StartLoadTransaction and either created the
+			// temp table (recording UseTempTable) or stashed the materialized rows for inline
+			// VALUES emission (recording UseInlineValues with the rows). When the execContext is
+			// absent (e.g. ToSqlQuery for diagnostics) the threshold check falls back to the
+			// just-materialized BuildRows count — this shows the temp-table reference that the
+			// query WOULD emit at execute time, not what the SQL builder could actually run
+			// without Setup. Execute paths that miss execContext entirely (CompiledQuery's
+			// generated delegate, ForEachUntilAsync) produce SQL referencing a non-existent
+			// table; those are tracked as separate limitations rather than papered over here.
```
MaceWindu, line 2003: **Nit · NIT001** — This comment lists "CompiledQuery's generated delegate" as a path that misses `execContext`, but the PR wired `CompiledTable<T>.Execute`/`ExecuteAsync` to create a `QueryExecutionContext` and run init-queries before `GetElement`, so the compiled path now does carry it. `ForEachUntilAsync` remains genuinely uncovered (see BLK001), so keep that half.

Fix: Drop the stale CompiledQuery reference.

```suggestion
			// without Setup. Execute paths that miss execContext entirely (e.g.
			// ForEachUntilAsync) produce SQL referencing a non-existent table; those
			// are tracked as separate limitations rather than papered over here.
```

Source/LinqToDB/Data/DataConnection.cs
```diff
@@ -6,6 +6,7 @@
 using System.Diagnostics;
 using System.Globalization;
 using System.Threading;
+using System.Threading.Tasks;
```
MaceWindu, line 9: **Nit · NIT002** — The PR adds `using System.Threading.Tasks;` here, but there are no `Task` / `ValueTask` / async references in `DataConnection.cs` — it's a partial class whose async members live in `DataConnection.Async.cs`. The directive is unused.

Fix: Remove it.

```suggestion
```

Source/LinqToDB/Linq/IAsQueryableExceptBuilder.cs
```diff
@@ -1,12 +1,13 @@
-﻿using System;
+using System;
```
MaceWindu, line 1: **Nit · NIT003** — This PR removed the file's UTF-8 BOM (the diff shows the leading byte-order mark on `using System;` deleted). `.editorconfig` sets `charset = utf-8-bom` for `*.cs`, so the BOM should be retained for consistency with the rest of the C# sources.

Fix: Restore the UTF-8 BOM at the start of the file. (Not expressible as a text suggestion since the visible content is unchanged — only the leading BOM byte differs.)

Source/LinqToDB/DataOptionsExtensions.TempTable.cs
MaceWindu, line 1: **Major · MAJ004** — The global `UseTempTablesForContains(...)` `DataOptions` switch changes the execution model of every `localCollection.Contains(value)` predicate in the application once set, with no signal at any call site. Unlike the per-call `AsQueryable(...).UseTempTable(...)` surface — where the opt-in is visible in the query — a global Contains switch silently turns ordinary `IN`-predicate queries into CREATE-temp-table + bulk-insert + EXISTS/IN-subquery executions for any query whose runtime collection happens to exceed the threshold. The temp tables get default column types (e.g. `nvarchar(4000)` for strings), no indexes, and default bulk-copy options; an unrelated query elsewhere can start failing in production — the test file itself documents real per-provider breakage that only surfaces on the temp-table path (SQL Server tempdb collation conflicts; MySQL `ER_CANT_REOPEN_TABLE` on self-joins). This is the maintainer's primary CHANGES_REQUESTED objection ("action-at-a-distance… visible syntax with hidden semantics"). The per-call `AsQueryable` opt-in is not contested — only the global Contains switch.

Fix: Either drop the global `UseTempTablesForContains` surface in favor of an explicit per-call / per-scope opt-in (e.g. the maintainer's suggested disposable `CreateLocalCollectionScope(options)` that owns the temp tables, or a query-local marker), or — if the global default is kept — make the default conservative, document the action-at-a-distance behavior prominently on the option, and surface the per-provider hazards (tempdb collation, single-temp-table-reference limits) as part of the option's contract rather than only in test comments.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes

- [ ] **API baselines need a refresh.** This PR adds new public surface. Regenerate `Source/**/CompatibilitySuppressions.xml` by deleting the existing files and running `dotnet pack -p:ApiCompatGenerateSuppressionFile=true` on every affected project under `Source/`. The new public types also aren't yet in `Source/LinqToDB/PublicAPI/**/PublicAPI.Unshipped.txt` — those are reconciled at the release boundary, so the analyzer lag is expected within the 6.4.0 cycle.
  - **Added** under `LinqToDB`: `TempTableSpec`, `TempTableOptions` (records); `DataOptions.TempTableOptions` (property); `ITempTableConfigBuilder`, `ITempTableBulkCopyOptionsBuilder`, `ITempTableBulkCopyTypeConfigBuilder` (fluent builder interfaces); `IDisposableTracker` (interface — but see **MAJ003**: likely belongs in `LinqToDB.Internal.*`); `DataOptionsExtensions.UseTempTablesForLocalCollections` / `UseTempTablesForContains` (extension methods).
  - **Added** under `LinqToDB.Linq`: `IAsQueryableExceptBuilder<T>.UseTempTable(int)` / `UseTempTable(Func<…>)` (overloads).
  - **Modified** under `LinqToDB` / `LinqToDB.Data` — purely additive (new interface implementation, not a breaking change): `DataConnection` and `DataContext` now implement `IInfrastructure<IDisposableTracker>`.
- [ ] **Core design is under unresolved maintainer dispute.** Two CHANGES_REQUESTED reviews object to the global `UseTempTablesForContains` switch, hiding temp-table lifetime behind `AsQueryable`/`Contains`, and adding `IDisposableTracker` ownership to the data context. Findings **MAJ003**, **MAJ004**, and **SUG001** overlap with that discussion and partly track concessions the author has already offered (move the tracker internal; rename `DisposeWithConnection`). They're surfaced here for completeness — the design call is the maintainers'.
- [ ] **9 findings posted as inline file threads** — 1 blocker, 4 major, 1 suggestion, 3 nits. They carry no body-section duplicate; read them on the file threads.

## Out-of-scope observations

- **API shape (abstraction inversion) is an open design debate.** The maintainer argues `AsQueryable(db)` should stay about representing a local collection as a query source, with temp-table materialization (a DB resource with DDL, bulk copy, lifetime, cleanup) modelled as an explicit temp-table-style API (`using var t = data.ToTempTable(...)`) rather than another option on the rendering chain — and that adding a new lifetime-tracking concern to the already-central `DataConnection`/`DataContext` weakens SRP. Both shapes are defensible; the per-call `AsQueryable(...).UseTempTable(...)` machinery itself is functionally correct and well-tested. Worth settling the abstraction (chain option vs. explicit resource) before this public surface locks in. The narrower, graded slices (MAJ003 placement, MAJ004 global switch, SUG001 naming) are filed as findings.
- **`TempTable<T>` now auto-registers every instance with the context's disposable tracker.** `RegisterWithDisposableTracker(db)` is called unconditionally from all four `TempTable<T>` constructors, so every existing `using var tt = new TempTable<T>(db, …)` / `CreateTempTable` caller now registers with the context tracker until `Dispose()` (which unregisters) or context close — a broader behavior change than the opt-in feature implies. The mechanics look sound (drain snapshots-and-clears before the unregister no-ops; drain runs before `Disposed = true`), but whether non-opt-in `TempTable` usage should participate at all is worth a deliberate decision.
- **Diagnostic / non-execute paths can render a temp-table reference for a table that was never created.** When the per-execute context is absent — the `ToSqlQuery` diagnostic path (and the `CompiledQuery`/`ForEachUntilAsync` execute paths) — `BasicSqlBuilder.BuildSqlValuesTable` falls back to a count-vs-threshold check and, above threshold, emits a bare temp-table name. For `ToSqlQuery` this only affects diagnostics (no execution), but the emitted SQL doesn't reflect what would actually run. The execute-path instance of this (`ForEachUntilAsync`) is filed as BLK001.
- **Confirm parameter-dependent caching across above/below-threshold executes.** The temp-table-vs-inline decision is read from the per-execute context during conversion/emission, so it's only correct if these statements are always treated as parameter-dependent (SQL rebuilt every `GetCommand`, never served from `CachedPreparedQuery`). Worth an explicit test: execute the same cached `Query<T>` first with a collection above the threshold then below it (and vice versa) and assert the second execute emits the other form.

## Baselines

Delta: [linq2db.baselines PR #1867](https://github.com/linq2db/linq2db.baselines/pull/1867) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5542)

1,105 SQL baselines changed (984 added, 121 modified, 0 deleted); no metrics baselines moved.

### ClickHouse: inline-rows `SELECT … UNION ALL` → native `VALUES(…)` table function (121 modified)

All 121 modified files are ClickHouse.Driver / ClickHouse.MySql. Every change is the same structural replacement of `( SELECT v1 AS col UNION ALL SELECT v2 … )` with `VALUES('col Nullable(Type)', (v1), (v2), …)` — the type-annotation row is new, data values unchanged. Uniform across EnumerableSourceTests, CteTests, AllAnyTests, SqlRowTests, StringFunctionsTests, ValueConversionTests, MappingTests, EnumerableInQuery. No structural difference between the two ClickHouse providers.

<details><summary>sample diff — InnerJoinArray(ClickHouse.Driver,1)</summary>

```diff
-		INNER JOIN (
-			SELECT 'Janet' AS item
-			UNION ALL
-			SELECT 'Doe'
-			UNION ALL
-			SELECT 'John'
-			UNION ALL
-			SELECT 'Doe') n ON p.LastName = n.item
+		INNER JOIN VALUES(
+			'item Nullable(String)',
+			('Janet'), ('Doe'), ('John'), ('Doe')
+		) n ON p.LastName = n.item
```
</details>

### New fixture `ContainsUseTempTableTests` (added, 37-provider temp-table cohort)

New baselines on all temp-table-capable providers (SqlServer 2005–2025 + SA/Contained, PostgreSQL 13–18, MySQL/MariaDB/MySqlConnector, SQLite Classic/MS/MPM/MPU). Per-family shape is correct: `CREATE` temp table → bulk/`VALUES` insert → `SELECT … WHERE id IN (SELECT … FROM temp)` → `DROP`. SqlServer 2008+ uses `INSERT BULK` + `DROP TABLE IF EXISTS`; SqlServer 2005 uses the `SELECT … UNION ALL` insert + `IF OBJECT_ID(...) DROP` idiom; PostgreSQL uses `ON COMMIT PRESERVE ROWS`; SQLite uses `temp.[T_…]`. All consistent.

### New fixture `EnumerableSourceTests.AsQueryable_UseTempTable_*` (added, same cohort)

Above-threshold path emits the full CREATE/INSERT/SELECT/DROP lifecycle; below-threshold emits plain inline `VALUES` with no DDL (threshold gate confirmed). `SqlServer.Contained` correctly emits parametrized inline `(VALUES (@Id,@Data), …)` below threshold.

### Cross-provider notes

- **`AsQueryable_*_OnInlineRowsUnsupportedProvider_Throws` produce SQL on non-Access providers — expected, not an anomaly.** These tests run across all `[DataSources]` with `[ThrowsForProvider(typeof(LinqToDBException), TestProvName.AllAccess)]`, so only Access asserts a build-time throw; every other provider runs and captures inline-`VALUES` SQL. DuckDB appearing with `(VALUES …)` output is correct — DuckDB supports inline-rows sources (`IsInlineRowsSourceSupported` is on by default); only `IsRuntimeTempTableCreationSupported` is off for it.
- **PostgreSQL.17/18 SchemaProvider function-order swap (4 tests, M).** `TestTableFunction` / `TestTableFunction1` swap positions in `SchemaProviderTests.PrimaryForeignKeyTest`, `.Test`, `Issue792Tests.TestWithoutTransaction`/`.TestWithTransaction` — PG.17/18 only, not PG.13–16. This PR touches no SchemaProvider code; looks like a `pg_catalog` ordering difference between server versions or CI environment drift. Worth confirming it's benign before merging the baselines PR.
- **DB2 `Issue2763Test`: `'SESSION'` schema removed from 6 filter lists (M).** Single-provider, single-test; this PR has no DB2 changes. Likely CI environment drift (or an unrelated fix that rode the same baselines run). Confirm before merging the baselines PR.

Issue #1856, opened 2019-08-09
Combining FromSql and AsSubquery/AsCte


SaltyDH:
I'm trying to generate a complex query using a combination of FromSql and either AsCte or AsSubQuery (whichever works). The core of the query is an IQueryable that's produced based on user provided filters but this problem is reproducible off a standard GetTable call.

```
from q in Connection.GetTable<T>().AsSubQuery()
from t in Connection.FromSql<Test2>($"SELECT {q.Id}, {q.Deleted} FROM {q} GROUP BY {q.Id}")
select new
	   {
		   t.Id,
		   t.Deleted
	   };
```

Produced sql (invalid):
```
SELECT
	[t].[Id],
	[t].[Deleted]
FROM
	(
		SELECT
			[t1].[Id],
			[t1].[Deleted],
			[t1].*
		FROM
			[table1] [t1]
	) [q]
		CROSS APPLY (
			SELECT [q].[Id], [q].[Deleted] FROM [q].[*] GROUP BY [q].[Id]
		) [t]
```

When I project the table to specify the fields explicitly:
```
from q in Connection.GetTable<T>().Select(x => new {x.Id, x.Deleted}).AsSubQuery()
```

I get the following exception. Which seems to be related to more than 1 column coming from the subquery.

```
System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at ISqlExpression LinqToDB.Linq.Builder.ExpressionBuilder.ConvertToSql(IBuildContext context, Expression expression, bool unwrap) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\ExpressionBuilder.SqlBuilder.cs:line 991
   at TResult[] System.Linq.Enumerable+SelectIListIterator<TSource, TResult>.ToArray()
   at TSource[] System.Linq.Enumerable.ToArray<TSource>(IEnumerable<TSource> source)
   at IBuildContext LinqToDB.Linq.Builder.TableBuilder.BuildRawSqlTable(ExpressionBuilder builder, BuildInfo buildInfo) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\TableBuilder.RawSqlContext.cs:line 29
   at IBuildContext LinqToDB.Linq.Builder.TableBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo) in C:\projects\linq2db\Source\LinqToDB\Linq\Builder\TableBuilder.cs:line 142
[...]

PR #5545, opened 2026-05-19
Fix #5543: nested fluent column mapping in MERGE implicit setter

### LINKED ISSUE #5543: UpdateWhenMatched in Merge API fails with 'Member not found' exception if nested fields exists in object mapping
### Description

Using update when matched in `Merge` API with nested objects (fields) mapping causes `Member not found` exception, which completely block bulk upsert operations by not pkey fields.

### Reproduction Steps

```csharp
// NuGet packages:
// linq2db@6.3.0 (latest, also reproduced on linq2db@6.1.0)
// Npgsql@10.0.2 (latest, also reproduced on Npgsql@6.0.13)

// Database arrange:
/*

create table "primary" (
	id uuid primary key,
	code varchar,
    field boolean
);

*/

using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;

namespace TestApp;

internal static class Program
{
    private record Primary
    {
        public required Guid Id { get; init; }
        public required string Code { get; init; }
        public required Secondary Secondary { get; init; }
    }

    private record Secondary
    {
        public required bool Field { get; init; }
    }

    private static MappingSchema Schema()
    {
        var builder = new FluentMappingBuilder();
        
        builder.Entity<Primary>()
            .HasTableName("primary")
            .Property(o => o.Id).HasColumnName("id").IsPrimaryKey()
            .Property(o => o.Code).HasColumnName("code")
            .Property(o => o.Secondary.Field).HasColumnName("field");
        
        builder.Build();
        
        return builder.MappingSchema;
    }
    
    public static async Task Main(string[] args)
    {
        var opts = new DataOptions()
            .UseMappingSchema(Schema())
            .UsePostgreSQL("Host=localhost;Port=5432;Username=postgres;Password=postgres;Pooling=True;Database=postgres;Command Timeout=300;Include Error Detail=true");

        var db = new DataConnection(opts);

        var newRecords = new[]
        {
            new Primary { Id = Guid.CreateVersion7(), Code = "first", Secondary = new Secondary { Field = false } },
            new Primary { Id = Guid.CreateVersion7(), Code = "second", Secondary = new Secondary { Field = false } },
            new Primary { Id = Guid.CreateVersion7(), Code = "third", Secondary = new Secondary { Field = false } },
        };

        // OK
        await db.GetTable<Primary>()
            .Merge()
            .Using(newRecords)
            .On(o => o.Code, o => o.Code)
            .InsertWhenNotMatched()
            .MergeAsync();

        var updateRecords = new[]
        {
            new Primary { Id = Guid.CreateVersion7(), Code = "first", Secondary = new Secondary { Field = true } },
        };
        
        // EXCEPTION
        await db.GetTable<Primary>()
            .Merge()
            .Using(updateRecords)
            .On(o => o.Code, o => o.Code)
            .UpdateWhenMatched()
            .MergeAsync();
    }
}
```

### Exception details

```
Unhandled exception. System.InvalidOperationException: Member 'Boolean Field' not found in type 'TestApp.Program+Primary'.
   at LinqToDB.Internal.Linq.Builder.MergeBuilder.UpdateWhenMatched.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/MergeBuilder.UpdateWhenMatched.cs:line 62
   at LinqToDB.Internal.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/MethodCallBuilder.cs:line 10
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.TryBuildSequence(BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 358
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 373
   at LinqToDB.Internal.Linq.Builder.MergeBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/MergeBuilder.cs:line 43
   at LinqToDB.Internal.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/MethodCallBuilder.cs:line 10
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.TryBuildSequence(BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 358
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 373
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.Build[T](IQueryExpressions& expressions) in /_/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:line 128
   at LinqToDB.Internal.Linq.Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, IQueryExpressions& expressions) in /_/Source/LinqToDB/Internal/Linq/Query{T}.cs:line 403
   at LinqToDB.Internal.Linq.Query`1.GetQuery(IDataContext dataContext, IQueryExpressions& expressions, Boolean& dependsOnParameters) in /_/Source/LinqToDB/Internal/Linq/Query{T}.cs:line 380
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetQuery(IQueryExpressions& expression, Boolean cache, Boolean& dependsOnParameters) in /_/Source/LinqToDB/Internal/Linq/ExpressionQuery.cs:line 91
   at LinqToDB.Internal.Linq.ExpressionQuery`1.LinqToDB.Internal.Async.IQueryProviderAsync.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken) in /_/Source/LinqToDB/Internal/Linq/ExpressionQuery.cs:line 104
   at TestApp.Program.Main(String[] args) in C:\Development\Sources\TestApp\TestApp\Program.cs:line 80
   at TestApp.Program.<Main>(String[] args)

```

### Regression?

No info

### Linq To DB version

6.1.0

### .NET version

10.0.203

### Database

PostgreSQL 18.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 15.2.0) 15.2.0, 64-bit

### Database provider

Npgsql@6.0.13

### Operating system

Any (reproduced on WIndows)

### Configuration

Tested only on PostgreSQL, but I assume not only one provider is affected.

sdanyliv:
## Summary

- `MergeBuilder.UpdateWhenMatched` and `UpdateWhenMatchedThenDelete` (implicit-setter branch) resolved source/target column expressions by the leaf `MemberInfo` name on the entity root, which fails when `FluentMappingBuilder` maps the column via a *nested* member path (e.g. `Property(o => o.Sub.Field).HasColumnName(...)`). Reproduced as `InvalidOperationException: Member 'X' not found in type 'Y'`.
- New internal extension `ColumnDescriptor.GetMemberAccessExpression(...)` walks the full dot-path stored in `ColumnDescriptor.MemberName` when the instance type matches the entity root, falling back to the leaf-name lookup for cross-type `Merge().Using<TSource>()` and dynamic columns.
- Both MERGE sites plus the two inheritance-discriminator emission sites (`ExpressionBuilder.SqlBuilder.cs:790`, `ExpressionBuildVisitor.cs:4919`) now route through the same helper.

Fixes #5543

## Test plan

- [x] New `MergeTests.ComplexProperty_UpdateWhenMatched` / `_InsertUpdate` / `_UpdateWithDelete` pass on PostgreSQL.16 + Oracle.23.Managed (+ LinqService variants)
- [x] Full `MergeTests` suite passes (631/631) on enabled providers
- [x] `UpdateTests` / `InsertTests` / `InsertOrUpdate` suites pass
- [x] CI matrix on all merge-capable providers (Firebird, SAP HANA, SQL Server 2008+, Sybase, DB2)

## Follow-up commit (ed88e907e8)

Addresses review feedback on the implicit-setter helper:

- **Explicit-interface guard.** `GetMemberAccessExpression` now gates the dot-path walk on `MemberAccessor.IsComplex && !MemberInfo.IsDynamicColumnProperty` (the `TableBuilder.TableContext.cs` idiom) instead of `MemberName.Contains('.')` - an explicit interface implementation's CLR member name itself contains dots, so the old guard split it and threw at MERGE query build / discriminator emission. Explicit-interface and dynamic columns now route to the leaf-member fallback.
- **Tests.** `ExplicitInterfaceProperty_UpdateWhenMatched` (fails on the old guard, passes on the fix) and `ComplexProperty_NestedDiscriminator` (covers the inheritance-discriminator sites with a nested-mapped discriminator). Verified on DuckDB + Oracle.23.Managed (direct + LinqService).


Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+		/// suitable for SQL conversion. When <paramref name="instance"/>'s CLR type
+		/// doesn't match the column's entity root (e.g. <c>Merge().Using&lt;TSource&gt;()</c>
+		/// with TSource ≠ TTarget), the column's leaf member is looked up on the
+		/// instance type by name.
+		/// </remarks>
+		/// <exception cref="InvalidOperationException">
+		/// Thrown when no matching member exists on <paramref name="instance"/>'s type.
+		/// </exception>
+		public static Expression GetMemberAccessExpression(this ColumnDescriptor descriptor, Expression instance)
+		{
```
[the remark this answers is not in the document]
sdanyliv: Verified — `Release` build passes against `Source/LinqToDB/PublicAPI/*.txt` and `Source/LinqToDB/CompatibilitySuppressions.xml` as committed. Re-running `/api-baselines` (`dotnet pack -p:ApiCompatGenerateSuppressionFile=true`) regenerates the suppression files byte-identical to `HEAD` — the addition is purely additive, no suppressions needed.

`public` matches sibling types under `Source/LinqToDB/Internal/Mapping/` (`DynamicColumnInfo`, `MappingSchemaInfo`, `LockedMappingSchema`, etc., all `public sealed class`). The `LinqToDB.Internal.*` namespace is the convention for "public-but-internal-use APIs" consumed by sister linq2db libraries.

Tests/Linq/Update/MergeTests.ComplexProperty.cs
```diff
@@ -0,0 +1,152 @@
+using System;
```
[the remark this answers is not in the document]
sdanyliv: Removed in 01607556f6.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+		/// (e.g. <c>o =&gt; o.Sub.Field</c>).
+		/// </summary>
+		/// <remarks>
+		/// Unlike <see cref="MemberAccessor.GetGetterExpression"/>, the result has no
+		/// null-check block wrapping intermediate reference-type members, so it is
+		/// suitable for SQL conversion. When <paramref name="instance"/>'s CLR type
+		/// doesn't match the column's entity root (e.g. <c>Merge().Using&lt;TSource&gt;()</c>
+		/// with TSource ≠ TTarget), the column's leaf member is looked up on the
+		/// instance type by name.
+		/// </remarks>
```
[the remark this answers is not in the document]
sdanyliv: Reworded in 01607556f6 — "member access expression" instead of `MemberExpression`, and added a note that the `DynamicColumnInfo` fallback path forwards through `ExpressionExtensions.GetMemberGetter` and may return a `MethodCallExpression`.

Tests/Linq/Update/MergeTests.ComplexProperty.cs
```diff
@@ -0,0 +1,152 @@
...
+			{
+				AssertRowCount(1, rows, context);
+				Assert.That(result,                 Has.Count.EqualTo(2));
+				Assert.That(result[0].Id,           Is.EqualTo(1));
+				Assert.That(result[0].Code,         Is.EqualTo("first-updated"));
+				Assert.That(result[0].Nested.Field, Is.True);
+				Assert.That(result[1].Id,           Is.EqualTo(2));
+				Assert.That(result[1].Code,         Is.EqualTo("skipped"));
+				Assert.That(result[1].Nested.Field, Is.False);
+			}
```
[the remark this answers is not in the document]
sdanyliv: Keeping NUnit `Assert.That(...)` here for consistency with the rest of the `MergeTests` partial class. All sibling files (`MergeTests.cs`, `MergeTests.Issues.cs`, `MergeTests.DynamicColumns.cs`, `MergeTests.Operations.*.cs`, `MergeTests.WithOutput.cs`, etc.) use NUnit asserts; switching one file to Shouldly would make this the odd one out.

igor-tkachev (dismissed):
The new tests cover the intended nested-member case (`Nested.Field`), but I think we also need a regression case for dotted member names that are **not** nested paths.

The new helper treats any `ColumnDescriptor.MemberName` containing `.` as a path to split. That also matches explicit interface implementation member names, where dots are part of the CLR member name rather than a property path. I verified this locally against `MergeDataContextSource`: `UpdateWhenMatched()` reaches the provider, creates the table, inserts the row, and then fails during MERGE query build with:

```text
System.InvalidOperationException:
Instance property or field with name Tests not found on type
Tests.xUpdate.MergeTests+ExplicitComplexPropertyTarget

at ColumnDescriptorExtensions.GetMemberAccessExpression(...) line 42
at MergeBuilder.UpdateWhenMatched.BuildMethodCall(...) line 59
```

Suggested regression test:

```csharp
public interface IExplicitComplexProperty
{
	bool Field { get; set; }
}

public sealed class ExplicitComplexPropertyTarget : IExplicitComplexProperty
{
	public int Id { get; set; }

	[Column("Field")]
	bool IExplicitComplexProperty.Field { get; set; }
}

[Test]
public void ExplicitInterfaceProperty_UpdateWhenMatched([MergeDataContextSource] string context)
{
	var ms = new MappingSchema();

	new FluentMappingBuilder(ms)
		.Entity<ExplicitComplexPropertyTarget>()
			.HasTableName("ExplicitComplexPropertyTarget")
			.Property(o => o.Id).IsPrimaryKey().HasColumnName("Id")
		.Build();

	var target = new ExplicitComplexPropertyTarget { Id = 1 };
	((IExplicitComplexProperty)target).Field = false;

	var source = new ExplicitComplexPropertyTarget { Id = 1 };
	((IExplicitComplexProperty)source).Field = true;

	using var db    = GetDataContext(context, ms);
	using var table = db.CreateLocalTable<ExplicitComplexPropertyTarget>();

	db.Insert(target);

	table.Merge()
		.Using(new[] { source })
		.OnTargetKey()
		.UpdateWhenMatched()
		.Merge();

	var result = table.Single();

	Assert.That(((IExplicitComplexProperty)result).Field, Is.True);
}
```

The production-side guard probably needs to distinguish real nested paths from explicit interface member names, e.g.:

```csharp
if (memberName.Contains('.', StringComparison.Ordinal)
	&& !descriptor.MemberInfo.Name.Contains('.', StringComparison.Ordinal)
	&& instance.Type == descriptor.MemberAccessor.TypeAccessor.Type)
```

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+		/// <exception cref="InvalidOperationException">
+		/// Thrown when no matching member exists on <paramref name="instance"/>'s type.
+		/// </exception>
+		public static Expression GetMemberAccessExpression(this ColumnDescriptor descriptor, Expression instance)
+		{
+			var memberName = descriptor.MemberName;
+
+			if (memberName.Contains('.', StringComparison.Ordinal)
+				&& instance.Type == descriptor.MemberAccessor.TypeAccessor.Type)
```
MaceWindu: **Blocker · BLK001** — The dot-path guard fires for *any* `MemberName` containing `.`. But `MemberName` falls back to `MemberInfo.Name`, and the CLR name of an explicit-interface-implemented member is itself dotted (e.g. `Namespace.IFoo.Bar`). For such a column reached through any of the four call sites this PR converts (`UpdateWhenMatched` / `UpdateWhenMatchedThenDelete` implicit setters and both `MakeIsPredicateExpression` discriminator sites), `memberName.Contains('.')` is true **and** `instance.Type == descriptor.MemberAccessor.TypeAccessor.Type` is true (the instance *is* the entity root), so the walk runs: `memberName.Split('.')` yields `["Namespace", "IFoo", "Bar"]` and `ExpressionHelper.PropertyOrField(instance, "Namespace")` throws `InvalidOperationException: Instance property or field with name … not found on type …`. This is a query-build crash, not a fallback, and it's a regression — the previous code (`GetMemberEx(MemberInfo)` + `GetMemberGetter`) resolved explicit-interface members correctly. The project lead reproduced this exact failure locally, and it's the open review thread on this file.

The repo's canonical dot-walk at `Source/LinqToDB/Internal/Linq/Builder/EntityConstructorBase.cs:71` (and again at `:107`, commented `// explicit interface implementation`) guards precisely this case with `&& !MemberInfo.Name.Contains('.')`; this helper dropped that guard.

Fix: add the explicit-interface guard so these columns fall through to the leaf-name branch (which resolves them correctly). The suggestion below is the minimal, maintainer-verified form. Cleaner still — see MIN002 — is to gate on the structured `descriptor.MemberAccessor.IsComplex && !descriptor.MemberInfo.IsDynamicColumnProperty` (the form already used at `Source/LinqToDB/Internal/Linq/Builder/TableBuilder.TableContext.cs:439`) instead of string-sniffing the name; that also routes nested dynamic columns to the correct branch.

```suggestion
			if (memberName.Contains('.', StringComparison.Ordinal)
				&& !descriptor.MemberInfo.Name.Contains('.', StringComparison.Ordinal)
				&& instance.Type == descriptor.MemberAccessor.TypeAccessor.Type)
```

Tests/Linq/Update/MergeTests.ComplexProperty.cs
MaceWindu, line 1: **Minor · MIN001** — The two converted inheritance-discriminator sites have no new test coverage. This PR also rewrites the discriminator `is`-predicate member access in `ExpressionBuildVisitor.cs` and `ExpressionBuilder.SqlBuilder.cs` (replacing `ttype.GetMemberEx(field.ColumnDescriptor.MemberInfo)` + `Expression.MakeMemberAccess`), but no new test exercises an inheritance hierarchy whose discriminator column is nested-mapped — so the behavioral change at those two sites is unverified by the diff, and a nested-mapped discriminator would hit the same dot-walk path as BLK001.

Fix: add a test that maps a discriminator to a nested member (e.g. `o => o.Disc.Value`) on an inheritance-mapped entity and runs a query whose predicate forces `MakeIsPredicateExpression` (an `.OfType<>()` / `is`-typed filter). If covering the discriminator sites is considered out of scope for this fix, say so explicitly — but the PR does change them.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
MaceWindu, line 1: **Minor · MIN002** — The "is this column a nested member path?" test is now duplicated across four sites with divergent forms, and this helper is the one that diverged. `EntityConstructorBase.cs:71` and `:102`+`:107` guard the dot-walk with `MemberName.Contains('.') && !MemberInfo.Name.Contains('.')`; this helper (`:37`) and `ParametersContext.cs:121` use only `MemberName.Contains('.')`. Dropping the `!MemberInfo.Name.Contains('.')` half is exactly what produced BLK001 — adding it back inline here would create a fifth divergent copy.

The codebase already has a structured answer: `MemberAccessor.IsComplex` is `true` for nested member paths and dynamic columns, `false` for both simple and explicit-interface members, and `TableBuilder.TableContext.cs:439` already uses the canonical form `IsComplex && !MemberInfo.IsDynamicColumnProperty`.

Fix: gate the dot-walk on `descriptor.MemberAccessor.IsComplex && !descriptor.MemberInfo.IsDynamicColumnProperty` (keeping the `instance.Type == …TypeAccessor.Type` cross-type guard). That fixes BLK001 by construction, routes nested dynamic columns to the correct branch, and removes the string-sniff. Consolidating the predicate into one `ColumnDescriptor` helper that all four sites call would prevent this class of divergence recurring; `ParametersContext.cs:121` having the same missing guard suggests it's worth doing.

Tests/Linq/Update/MergeTests.ComplexProperty.cs
MaceWindu, line 1: **Major · MAJ001** — Missing regression test for the explicit-interface-member case the fix doesn't handle. The three new tests cover only the nested-path happy case (`o => o.Nested.Field`). The project lead verified (CHANGES_REQUESTED) that the helper *fails* for explicit-interface-implementation columns whose CLR member name contains dots (see BLK001) and supplied a ready-to-drop regression test (`ExplicitInterfaceProperty_UpdateWhenMatched` over `[MergeDataContextSource]`, with an `ExplicitComplexPropertyTarget : IExplicitComplexProperty` carrying a `[Column("Field")]` explicit-interface member). That test is not present at HEAD. Without it, the BLK001 fix can be dropped or silently re-regress later.

Fix: add the maintainer-supplied `ExplicitInterfaceProperty_UpdateWhenMatched` test, paired with the BLK001 guard fix — it fails at MERGE query build today, so it doubles as the guard's verification.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] **Baseline anomaly to confirm.** This PR's baselines added a `.sql.other` file for `CharTypesTests.StringTrimming(Oracle.12.Managed)` — that file is written only when a test's direct-access SQL and its remote (LinqService) SQL diverge, i.e. the test is currently failing in remote mode on Oracle.12. `StringTrimming` is unrelated to this MERGE fix, so it is most likely pre-existing / collateral CI noise rather than something this PR caused — but it shouldn't ship silently. Please confirm it isn't a regression before merge. (It is not covered by the #5513 Oracle remote-trace carve-out, which is scoped to scalar aggregate queries.)
- [x] **New public API is in `LinqToDB.Internal.*` (additive — fine).** `ColumnDescriptorExtensions` / `GetMemberAccessExpression` land in `LinqToDB.Internal.Mapping`, matching sibling public-but-internal types there; the change is additive only. The earlier public-API-analyzer concern was correct to dismiss — a Release build passes and the API suppression files regenerate unchanged. `PublicAPI.Unshipped.txt` will pick this up at the release refresh; no action now.

## Out-of-scope observations

- **`OnTargetKey()` uses the same unconverted leaf-name lookup over primary-key columns.** `Source/LinqToDB/Internal/Linq/Builder/MergeBuilder.On.cs:77-83` builds the key-match condition via `targetLambdaType.GetMemberEx(column.MemberInfo)` + `Expression.MakeMemberAccess` over the entity's primary-key columns, and was not routed through the new helper. A primary key mapped through a nested member path (`Property(o => o.Sub.Key).IsPrimaryKey()`) would throw the same `Member '…' is not defined` there. This is pre-existing on `master` — neither introduced nor worsened by this PR, and the new tests use a flat `Id` PK so they don't exercise it. Flagged only so a follow-up can decide whether nested-mapped MERGE keys are worth supporting.
- **Nested *dynamic* columns can't be resolved by the dot-walk.** The helper's nested branch walks the path with `ExpressionHelper.PropertyOrField`, which resolves only real CLR properties/fields and throws for a dynamic column. A nested dynamic-column mapping in a MERGE implicit setter or discriminator would therefore fail. Not a regression: on `master` these sites used `GetMemberEx` on the entity root, which already failed for nested mappings (the #5543 bug). Worth noting for a follow-up if nested dynamic columns are ever expected — keying the branch off `MemberAccessor.IsComplex` (see MIN002) and emitting `Sql.Property` for a dynamic leaf hop would cover it.

## Baselines
Delta: [linq2db.baselines PR #1856](https://github.com/linq2db/linq2db.baselines/pull/1856) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5545)

85 files added, 9 modified, 0 deleted across 39 providers. Virtually all additions are new SQL baselines for the 3 new `MergeTests.ComplexProperty_*` tests; the new SET-clause shape `[Field] = [Source].[Nested_Field]` confirms the nested-property column name resolves correctly end-to-end on every merge-capable provider. No regression archetypes fired.

### New tests: `MergeTests.ComplexProperty_*` (3 tests, 85 new files)
- **`ComplexProperty_UpdateWhenMatched`** — added (39 providers: DB2, DuckDB, Firebird.2.5, Firebird.3, Firebird.4, … (39 providers total)). Single `WHEN MATCHED THEN UPDATE SET [Code]=…, [Field]=[Source].[Nested_Field]`.
- **`ComplexProperty_InsertUpdate`** — added (39 providers). `WHEN MATCHED … UPDATE` + `WHEN NOT MATCHED … INSERT`. Representative SqlServer sample:

  ```diff
  @@ -0,0 +1,87 @@
  +MERGE INTO [ComplexPropertyTarget] [Target]
  +USING (VALUES
  +	(1,N'alpha-new',1), (3,N'gamma',1)
  +) [Source]
  +(
  +	[Id],
  +	[Code],
  +	[Nested_Field]
  +)
  +ON ([Target].[Id] = [Source].[Id])
  +WHEN MATCHED THEN
  +UPDATE
  +SET
  +	[Code]  = [Source].[Code],
  +	[Field] = [Source].[Nested_Field]
  +WHEN NOT MATCHED THEN
  +INSERT ([Id],[Code],[Field])
  +VALUES ([Source].[Id],[Source].[Code],[Source].[Nested_Field])
  +;
  ```
  </details>
- **`ComplexProperty_UpdateWithDelete`** — added (6 providers: Oracle.11.Managed … Oracle.23.Managed). Correctly scoped to Oracle only (proprietary `UPDATE … DELETE WHERE` MERGE extension); emits `UPDATE SET … WHERE 1 = 1 DELETE WHERE "Source"."Code" = 'remove-updated'`.

### Modified pre-existing baselines (9 files)
- **PostgreSQL 17/18 — function-listing order swap** (4 tests × 2 versions): `SchemaProviderTests.PrimaryForeignKeyTest`, `SchemaProviderTests.Test`, `Issue792Tests.TestWithoutTransaction`, `Issue792Tests.TestWithTransaction`. The `TestTableFunction(…)` and `TestTableFunction1(…)` entries swap positions; net delta 0 bytes. Looks like non-deterministic schema-reflection ordering, unrelated to the MERGE fix.
- **DB2 — schema-list trim** (`DB2Tests.Issue2763Test`): `'SESSION'` removed from the schema IN-lists across 7 introspection queries. DB2-scoped, unrelated to this fix — looks like collateral from a concurrent schema-provider change on `master`.

### Cross-provider anomalies
- **Suspect — new `.sql.other` on Oracle.12.Managed (`CharTypesTests.StringTrimming`).** See the first review note. A direct-vs-remote SQL mismatch surfaced as a new `.sql.other`; the test is unrelated to this PR. Needs maintainer confirmation it's pre-existing rather than a regression riding along in this PR's baselines.

sdanyliv:
Addressed in ed88e907e8.

`GetMemberAccessExpression` now gates the dot-path walk on `MemberAccessor.IsComplex && !MemberInfo.IsDynamicColumnProperty` instead of `MemberName.Contains('.')`, so explicit-interface members (dotted CLR names) fall through to the leaf-member lookup. Added your `ExplicitInterfaceProperty_UpdateWhenMatched` test — verified it throws at MERGE query build on the old guard and passes on the fix (DuckDB, direct + LinqService).

I used the structured `IsComplex` form (matching `TableBuilder.TableContext.cs`) rather than the `!MemberInfo.Name.Contains('.')` string form you sketched — it also routes nested dynamic columns to the leaf branch and avoids adding a fifth divergent string check. Glad to switch to the string form if you prefer it.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+			{
+				var expr = instance;
+				foreach (var part in memberName.Split('.'))
+					expr = ExpressionHelper.PropertyOrField(expr, part);
```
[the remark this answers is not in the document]
sdanyliv: Fixed in ed88e907e8 — the guard now uses `MemberAccessor.IsComplex && !MemberInfo.IsDynamicColumnProperty`, so explicit-interface members fall through to the leaf-member lookup. Regression test `ExplicitInterfaceProperty_UpdateWhenMatched` added.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
sdanyliv: Fixed in ed88e907e8. Used the structured `MemberAccessor.IsComplex && !MemberInfo.IsDynamicColumnProperty` form (per MIN002, matching `TableBuilder.TableContext.cs`) rather than the `!MemberInfo.Name.Contains('.')` string guard — fixes the crash and also routes nested dynamic columns to the leaf branch. Verified the explicit-interface test fails on the old guard and passes on the fix.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
sdanyliv: Done in ed88e907e8 — gated on `MemberAccessor.IsComplex && !MemberInfo.IsDynamicColumnProperty` (the `TableBuilder.TableContext.cs` form) instead of string-sniffing `MemberName`. Scoped to this helper; the `ParametersContext.cs` consolidation you flagged is pre-existing and left as a follow-up.

Tests/Linq/Update/MergeTests.ComplexProperty.cs
sdanyliv: Added `ExplicitInterfaceProperty_UpdateWhenMatched` (the maintainer-supplied repro) in ed88e907e8. Confirmed it throws at MERGE query build on the pre-fix guard and passes with the fix (DuckDB, direct + LinqService).

sdanyliv: Added `ComplexProperty_NestedDiscriminator` in ed88e907e8 — maps the inheritance discriminator through a nested member (`o => o.Meta.Kind`) and forces `MakeIsPredicateExpression` via `OfType<>()`. Passes on the fix.

sdanyliv:
Re: the `CharTypesTests.StringTrimming(Oracle.12.Managed)` `.sql.other` baseline note — confirmed not a regression from this PR.

This PR changes only the MERGE implicit-setter / inheritance-discriminator member-access path; it touches no string-translation code, so it cannot alter StringTrimming's generated SQL. Ran `StringTrimming` on Oracle.23.Managed (direct + LinqService) at this PR's HEAD — both pass, no direct-vs-remote divergence. The Oracle.12-specific `.sql.other` is pre-existing remote-trace behavior on that version, surfaced as CI collateral.

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: individual findings can be wrong, overconfident, or miss context. Feel free to disagree, dismiss, or ask for clarification — the findings are starting points for discussion, not verdicts.

## Review notes
- [x] **All prior review findings are resolved at HEAD.** BLK001 (explicit-interface dotted-name crash) is fixed by the new `IsComplex && !IsDynamicColumnProperty` guard — it matches the canonical form at `TableBuilder.TableContext.cs`; MAJ001 (`ExplicitInterfaceProperty_UpdateWhenMatched`) and MIN001 (`ComplexProperty_NestedDiscriminator`) tests are present; MIN002 (gate on `IsComplex`, not string-sniffing `MemberName`) is done. The broader `ParametersContext.cs` consolidation MIN002 also mentioned is explicitly deferred — fine.
- [x] **`CharTypesTests.StringTrimming(Oracle.12.Managed)` `.sql.other` — confirmed not a regression.** Re-verified against the actual baseline file: it's a real direct-vs-remote SQL divergence on Oracle.12 (the same flaky Oracle remote-trace artifact class), unrelated to this PR's member-access change, and you confirmed it independently.
- [x] **Copilot's two BOM threads are valid — please address.** Both new files (`ColumnDescriptorExtensions.cs`, `MergeTests.ComplexProperty.cs`) are missing the UTF-8 BOM that `.editorconfig` mandates for `[*.{cs,csx,vb,vbx}]` (`charset = utf-8-bom`). Verified against the git blobs — they start at `using` (`75 73 69…`), not `EF BB BF`; every sibling `.cs` file in the repo carries the BOM. Line endings are fine (`.gitattributes` normalizes to CRLF on checkout). Re-save both files with a BOM.
- [x] **Copilot's Oracle-vs-PostgreSQL test-plan thread is valid but harmless.** `ComplexProperty_UpdateWithDelete` is correctly `[IncludeDataSources(true, TestProvName.AllOracle)]` — `… THEN DELETE` is an Oracle-proprietary MERGE extension, matching every sibling in `MergeTests.Operations.UpdateWithDelete.cs`. Only the PR body's "verified on PostgreSQL.16" claim for *that* test is inaccurate; no code change needed, just a stale test-plan line.

## Suggestions

- [ ] **SUG001** — The helper's cross-type `Merge().Using<TSource>()` leaf-lookup branch is untested *with a nested column*.

  Why: all five tests use the same CLR type for source and target, so the source side always takes the dot-walk branch — never the `GetMemberEx` leaf-lookup the helper's own XML-doc calls out as the `TSource ≠ TTarget` fallback. That's the helper's most fragile branch and it's unexercised for a nested mapping.

  Fix: add a MERGE test mapping `Property(o => o.Nested.Field)` on the target while feeding a *different* source type (projection / separate POCO), so `sourceRef.Type != descriptor.MemberAccessor.TypeAccessor.Type`.

- [ ] **SUG002** — The dynamic-column path through the new guard is untested.

  Why: the guard deliberately excludes dynamic columns (`!IsDynamicColumnProperty`) and routes them to `GetMemberGetter`, which can return a `MethodCallExpression`. No new test exercises a dynamic column in a MERGE implicit setter, so the exclusion branch and that result shape are unverified.

  Fix: add a dynamic-column MERGE setter test (the fixture shape exists in `MergeTests.DynamicColumns.cs`), or add a one-line comment in the helper if dynamic + nested is explicitly unsupported.

## Out-of-scope observations

- **`OnTargetKey()` still uses the unconverted leaf-lookup over PK columns.** `Source/LinqToDB/Internal/Linq/Builder/MergeBuilder.On.cs:77-83` builds the key-match condition via `targetLambdaType.GetMemberEx(column.MemberInfo)` + `MakeMemberAccess` and was not routed through the new helper, so a primary key mapped via a nested member path (`Property(o => o.Sub.Key).IsPrimaryKey()`) would throw the same `Member not found` there. Pre-existing on `master`, not introduced or worsened here (the new tests use a flat `Id` PK), and already raised in the earlier review — repeated only so a follow-up can decide whether nested-mapped MERGE keys are worth supporting.

## Baselines
Delta: [linq2db.baselines PR #1856](https://github.com/linq2db/linq2db.baselines/pull/1856) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5545)

177 files added, 0 modified, 0 deleted (cleaner than the earlier run — the PostgreSQL schema-order / DB2 schema-trim collateral from a stale base is gone). All additions are new SQL baselines for the 5 `MergeTests` tests across merge-capable providers. The SET clause emits `[Field] = [Source].[Nested_Field]` everywhere — confirming nested-property column resolution works end-to-end; the explicit-interface test emits the mangled `…IExplicitComplexPropertyField` source-column name (expected); `ComplexProperty_NestedDiscriminator` is an INSERT+SELECT with `WHERE "Kind" = 'Dog'` across 55 providers. No regression archetypes fired.

One suspect (see review note 2): the `StringTrimming(Oracle.12.Managed)` `.sql.other` — verified collateral, not this PR.

Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs
```diff
@@ -0,0 +1,57 @@
+using System;
```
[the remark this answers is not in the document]
MaceWindu: Fixed in b53825ec7 — re-saved with the UTF-8 BOM (`EF BB BF`) per `.editorconfig` `charset = utf-8-bom`.

Tests/Linq/Update/MergeTests.ComplexProperty.cs
```diff
@@ -0,0 +1,274 @@
+using System.Linq;
```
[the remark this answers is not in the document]
MaceWindu: Fixed in b53825ec7 — re-saved with the UTF-8 BOM (`EF BB BF`) per `.editorconfig` `charset = utf-8-bom`.

MaceWindu (approved):
## Verification update — 2026-06-12, against HEAD b53825ec7

Follow-up commit `b53825ec7` addresses the open items from the prior review.

### Fixed
- [x] **UTF-8 BOM** on `ColumnDescriptorExtensions.cs` + `MergeTests.ComplexProperty.cs` — both re-saved with the BOM (`EF BB BF`) per `.editorconfig` `charset = utf-8-bom`. Copilot's two BOM threads resolved.

### Retracted
- ~~**SUG001**~~ — withdrawn. The cross-type leaf-lookup branch I flagged as untested is not reachable through the public API: the no-arg implicit `UpdateWhenMatched()` is constrained to `IMergeableSource<TTarget, TTarget>`, so a nested target column always resolves the source side via the same-type dot-walk. The branch is defensive-only; no test can exercise it.

### Already covered — SUG002
- The dynamic-column path through the `!IsDynamicColumnProperty` guard is already exercised by the existing `MergeTests.DynamicColumns` suite (implicit `UpdateWhenMatched` over `Sql.Property` columns routes through `GetMemberAccessExpression` → leaf → `GetMemberGetter`). Nested+dynamic remains unsupported by design.

### New follow-up fix (this commit)
- **`OnTargetKey()` now routes its PK-match through `GetMemberAccessExpression`** (`MergeBuilder.On.cs`) — the same helper the four other sites use, removing the last inline `GetMemberEx` + `MakeMemberAccess` duplication. Closes the out-of-scope `MergeBuilder.On.cs` observation from the prior reviews: a primary key mapped via a nested path (`[Column(MemberName = "Sub.Key", IsPrimaryKey = true)]`) threw `InvalidOperationException "Member '…' is not defined"`.
  - `ComplexProperty_NestedPrimaryKey_OnTargetKey` added — **red→green verified**: pre-fix throws `Member 'Value' is not defined in 't'`; post-fix passes on DuckDB + DuckDB.LinqService.
  - **Zero baseline impact:** the full `MergeTests` suite (394 tests, all exercising `OnTargetKey` on flat PKs) passes with no baseline changes — flat-PK SQL is byte-identical.

> [!NOTE]
> Agentic review — input for discussion, not a verdict.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed `InvalidOperationException` ("Member '...' not found") when a column is mapped through a nested member path (e.g. `Property(o => o.Sub.Field)` or `[Column(MemberName = "Sub.Field")]`). Such mappings now work in MERGE implicit setters (`UpdateWhenMatched` / `UpdateWhenMatchedAndThenDelete`), in `OnTargetKey()` when the primary key is nested-mapped, and when an inheritance discriminator is mapped via a nested path.


Nested member-path column mappings (e.g. `o => o.Sub.Field`) now work in MERGE setters, `OnTargetKey()`, and inheritance discriminators instead of throwing.


---
<sub>Generated from commit `b53825e`.</sub>

Issue #1972, opened 2019-11-08
Define LinqToDB mapping attributes in another partial class


Franki1986:
Hi,

till now I am using table and column attributes to generate the mappings.
I would like to split the specific LinqToDb mapping to another partial class, so that I have one clean class in a file and another with the specific attributes.
My idea was to split it like this:

File Data.Person.cs:

```
public partial class Person{
    public string Name {get; set;}
    public int Age {get; set;}
}
```
 
File Data.LinqToDB.Person.cs:

```
using Metadata = System.ComponentModel.DataAnnotations;

[Table("PERSONS")]
[Metadata.MetadataType(typeof(Person_DB))]
public partial class Person{
}
public class Person_DB{
    [Column("NAME")]
    public string Name {get; set;}
    [Column("AGE")]
    public int Age {get; set;}
}

```
In this way I could cleanly split mapping and my data class. Is it somehow possible that this could work? First try din't work.

MaceWindu:
Don't think we can redirect mappings to another type, but you can apply column mappings to class directly and specify MemberName:
```cs
// main file
public partial class Person{
    public string Name {get; set;}
    public int Age {get; set;}
}

// mappings file
[Table("PERSONS")]
[Column("NAME", nameof(Name))]
[Column("AGE", nameof(Age))]
public partial class Person{
}
```

PR #5555, opened 2026-05-26
ClickHouse: fix and expand join hints

igor-tkachev:
ClickHouse supports a two-axis join hint syntax: a **strictness** modifier (ALL, ANY, SEMI, ANTI, ASOF) and a **distribution** modifier (GLOBAL). Previously the API modeled this incorrectly — ALL was treated as a compound prefix (producing AllOuter, AllSemi, etc.) rather than as a standalone strictness modifier, which did not reflect actual ClickHouse SQL syntax.

This PR corrects the model:

* ALL, ANY, SEMI, ANTI, ASOF, OUTER are now first-class standalone hints.
* **GLOBAL** and **GLOBAL strictness** combinations are available as explicit hints, including the previously missing GLOBAL ALL
* The old All* aliases (AllOuter, AllSemi, etc.) are deprecated with clear migration guidance but kept for backward compatibility.
* A SQL builder bug is fixed where a standalone ALL hint was incorrectly matched as a prefix, producing malformed SQL.

Compatibility note for 6.4.0 release notes:

* Deprecated ClickHouse `All*` join hint aliases now emit the corrected standalone strictness hints when consumers recompile (`AllOuter` -> `OUTER`, `AllSemi` -> `SEMI`, etc.). Existing already-compiled callers keep their inlined legacy string values; the SQL builder keeps compatibility handling for explicit/legacy `ALL <strictness>` hint strings.

Tests for join hints are migrated to T4 code generation, consistent with how other provider hint tests are structured in the repository.


Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
```diff
@@ -27,19 +27,27 @@ public static class Join
...
-			public const string AllSemi     = All + " SEMI";
-			public const string AllAnti     = All + " ANTI";
-			public const string AllAny      = All + " ANY";
-			public const string AllAsOf     = All + " ASOF";
+			public const string GlobalOuter = $"{Global} {Outer}";
+			public const string GlobalSemi  = $"{Global} {Semi}";
+			public const string GlobalAnti  = $"{Global} {Anti}";
+			public const string GlobalAny   = $"{Global} {Any}";
+			public const string GlobalAll   = $"{Global} {All}";
+			public const string GlobalAsOf  = $"{Global} {AsOf}";
```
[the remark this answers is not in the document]
igor-tkachev: C# 14 supports string interpolation constants.

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.cs
```diff
@@ -27,19 +27,27 @@ public static class Join
...
+			[Obsolete("Use Outer instead.")]
+			public const string AllOuter    = Outer;
+			[Obsolete("Use Semi instead.")]
+			public const string AllSemi     = Semi;
+			[Obsolete("Use Anti instead.")]
+			public const string AllAnti     = Anti;
+			[Obsolete("Use Any instead.")]
+			public const string AllAny      = Any;
+			[Obsolete("Use AsOf instead.")]
+			public const string AllAsOf     = AsOf;
```
MaceWindu: **Blocker · BLK001** — These five `Join.AllOuter` / `AllSemi` / `AllAnti` / `AllAny` / `AllAsOf` constants are part of the **shipped** public API, and this PR changes their compile-time values (`"ALL OUTER"` → `"OUTER"`, etc.). Because `const` values are inlined at each consumer's compile site, this is a breaking change for any code referencing them once recompiled — and it ships on a **minor** milestone (6.4.0), which the project's API-stability contract reserves for non-breaking changes. The change is deliberate and a genuine bug fix (the old `"ALL OUTER"` value produced malformed `ALL LEFT OUTER JOIN` SQL), and the constants are deprecated in the same PR, so this is a milestone/communication confirmation — not a request to revert.

Fix: confirm 6.4.0 is the intended milestone for this break (vs deferring to a major release), and call out the deprecated-alias SQL behavior change in the 6.4.0 release notes.

Source/LinqToDB/DataProvider/ClickHouse/ClickHouseHints.tt
```diff
@@ -17,54 +17,83 @@ namespace LinqToDB.DataProvider.ClickHouse
...
+		/// <summary>
+<#+ if (useInstead != null) { #>
+		/// Adds the same ClickHouse join hint as <c>Join<#= useInstead #>Hint</c>. This method is deprecated.
+<#+ } else { #>
+		/// Adds a ClickHouse <c><#= sqlHint ?? hint #></c> join hint.
+<#+ } #>
+		/// </summary>
+		/// <remarks>
+		/// AI-Tags: Group=Hints; HintType=Join; Execution=Deferred; Composability=Composable; Affects=SqlSemantics; Pipeline=ExpressionTree,SqlAST,SqlText; Provider=ProviderDefined;
+		/// </remarks>
```
MaceWindu: **Blocker · BLK002** — This introduces a new `AI-Tags` documentation convention — a `<remarks>` block containing `AI-Tags: Group=Hints; HintType=Join; Execution=Deferred; Composability=Composable; Affects=SqlSemantics; Pipeline=ExpressionTree,SqlAST,SqlText; Provider=ProviderDefined;` — emitted into every generated ClickHouse join-hint method. `AI-Tags` appears nowhere else in the repository, so this PR establishes the convention. Two concerns: (1) it is a cross-cutting documentation convention being established through a provider-scoped PR, before the feature has been approved; (2) `<remarks>` is human-facing — these machine tags ship in the public XML documentation (IntelliSense tooltips, docfx / API-site output, the NuGet `.xml` doc file), exposing internal metadata to every API consumer.

Fix: revert the `AI-Tags` `<remarks>` blocks from this template — both occurrences (here at lines 54-56 and the identical block at 80-82) — and regenerate `ClickHouseHints.generated.cs`, keeping the legitimate `<summary>` docs. Alternatively, put this PR on hold until the AI-doc-hints feature is designed and approved as its own change.

Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
```diff
@@ -536,15 +536,13 @@ protected override bool BuildJoinType(SqlJoinedTable join, SqlSearchCondition co
 
 					h = h[(ClickHouseHints.Join.Global.Length + 1)..];
 				}
-				else if (h.StartsWith(ClickHouseHints.Join.All, StringComparison.Ordinal))
+				else if (h.StartsWith(ClickHouseHints.Join.All, StringComparison.Ordinal) && !string.Equals(h, ClickHouseHints.Join.All, StringComparison.Ordinal))
 				{
+					// TODO: remove this branch after obsolete All* join hint APIs are removed.
```
MaceWindu: **Minor · MIN001** — This `else if (h.StartsWith(All) && !string.Equals(h, All))` branch is already unreachable, so the TODO's stated removal condition is wrong. After this PR no hint value begins with `"ALL "`: the standalone strictness constants are bare (`OUTER`/`ALL`/…), every `Global*` constant starts with `"GLOBAL"` (handled by the preceding branch, including `GlobalAll`), and the obsolete `All*` aliases were re-pointed to their bare values. Removing the obsolete APIs will not change that — the branch is dead now, not "after obsolete All* removed."

Fix: correct the comment (suggested), or delete the dead branch entirely.

```suggestion
					// Currently unreachable: no join hint value begins with "ALL " (the obsolete All* aliases now map to their bare strictness values). Kept defensively for a future "ALL <strictness>" hint string.
```

Tests/Linq/Data/TransactionTests.cs
```diff
@@ -310,8 +310,8 @@ public async Task AttachToExistingTransaction([DataSources(false, ProviderName.S
 		{
 			var connectionString = GetConnectionString(context);
 
-			using var db = GetDataConnection(context);
-			using var tr = db.BeginTransaction();
+			await using var db = GetDataConnection(context);
+			await using var tr = await db.BeginTransactionAsync();
```
MaceWindu: **Minor · MIN002** — Scope creep: this `using` → `await using` + `BeginTransaction()` → `await db.BeginTransactionAsync()` conversion is unrelated to the ClickHouse join-hint scope, and it moved the `AttachToExistingTransaction` baseline for **every provider in the suite (56)** — see the Baselines section of the review summary. A reasonable change on its own, but bundling it here widens the blast radius if the hints change later needs reverting.

Fix: move this to a separate PR focused on async-transaction test hygiene.

Source/LinqToDB/Async/AsyncExtensions.cs
```diff
@@ -183,9 +183,9 @@ public AsyncEnumerableAdapter(IQueryable<T> query)
 				_query = query ?? throw new ArgumentNullException(nameof(query));
 			}
 
-#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
+#pragma warning disable CS1998, IDE0391 // Async method lacks 'await' operators and will run synchronously
```
MaceWindu: **Minor · MIN003** — Scope creep: adding `IDE0391` here (and `IDE0390` in `EnumerableHelper.cs` 22/24 & 138/140, `EnumerableHelperTest.cs` 86/88, `BulkCopyTests.cs` 232/234 & 443/445 & 451/453) to the `#pragma warning disable/restore CS1998` lines is unrelated to the join-hint scope.

Fix: split the IDE0390/IDE0391 pragma additions into a separate analyzer-hygiene PR. If they are required to keep this branch's build green (e.g. a newer analyzer started flagging async-without-await iterators), note that justification in the PR body.

Tests/Linq/Extensions/ClickHouseTests.tt
```diff
@@ -0,0 +1,60 @@
...
+	GenerateJoinHint("Anti",        "LEFT ANTI");
+	GenerateJoinHint("Any",         "LEFT ANY");
+	GenerateJoinHint("All",         "LEFT ALL");
+
+	GenerateJoinHint("Global",      "GLOBAL LEFT");
+	GenerateJoinHint("GlobalOuter", "GLOBAL LEFT OUTER");
+	GenerateJoinHint("GlobalSemi",  "GLOBAL LEFT SEMI");
+	GenerateJoinHint("GlobalAnti",  "GLOBAL LEFT ANTI");
+	GenerateJoinHint("GlobalAny",   "GLOBAL LEFT ANY");
+	GenerateJoinHint("GlobalAll",   "GLOBAL LEFT ALL");
```
MaceWindu: **Minor · MIN005** — The PR removed the inline test that pinned the deprecated `JoinAllSemiHint` output, and the new template does not regenerate any `All*` alias test — so the re-pointed behavior of these still-public (now `[Obsolete]`) methods ships untested. Risk is low (they delegate to the tested non-deprecated methods), but the specific output is now unverified.

Fix: add the five `All*` aliases to the template's test list with their new expected SQL (suppressing CS0618 in the generated test), or deliberately note in the PR body that deprecated-alias coverage is intentionally dropped because they delegate to tested methods.

Tests/Linq/Update/BulkCopyTests.cs
MaceWindu, line 1: **Minor · MIN004** — Scope creep: trailing-whitespace cleanups in `BulkCopyDateOnly` / `BulkCopyDateOnlyArrayBound` (around lines 415, 417-418, 420, 433, 435-436, 438) and in `EnumerableHelperTest.cs` line 22 touch lines the join-hint change does not otherwise modify.

Fix: drop the unrelated whitespace edits, or fold them into the same separate hygiene PR as the pragma additions (MIN003).

Source/LinqToDB/Internal/DataProvider/ClickHouse/ClickHouseSqlBuilder.cs
```diff
@@ -536,15 +536,13 @@ protected override bool BuildJoinType(SqlJoinedTable join, SqlSearchCondition co
...
 				{
+					// Keep legacy "ALL <strictness>" values working for already compiled callers and explicit hint strings.
 					StringBuilder
 						.Append(ClickHouseHints.Join.All)
 						.Append(' ');
 
-					if (string.Equals(h, ClickHouseHints.Join.All, StringComparison.Ordinal))
-						return base.BuildJoinType(join, condition);
-
 					h = h[(ClickHouseHints.Join.All.Length + 1)..];
```
MaceWindu: **Minor · MIN006** — A legacy or explicit `"ALL <strictness>"` join-hint string (an already-compiled caller's inlined value, or an explicit `JoinHint("ALL OUTER")`) emits `ALL LEFT OUTER JOIN` — strictness *before* the join type. Per the ClickHouse docs the grammar is `[GLOBAL] [INNER|LEFT|…] [OUTER|SEMI|ANTI|ANY|ALL|ASOF] JOIN` (one modifier, placed after the join type), so `ALL LEFT OUTER JOIN` is non-canonical (the docs list `ALL LEFT JOIN` as invalid; current ClickHouse executes it leniently) and it diverges from the recompiled path, which emits `LEFT OUTER JOIN`. The branch comment says it keeps legacy values "working," but it emits a form the docs reject. Verified with a red→green regression test (`LegacyAllOuterJoinHintStringTest`); all 58 ClickHouseTests pass with the fix.

Fix: strip the `ALL ` prefix instead of prepending it (a matching regression test asserting `LEFT OUTER JOIN` locks it in).

```suggestion
					// Map legacy/explicit "ALL <strictness>" values (e.g. "ALL OUTER") to the bare <strictness> so they emit
					// canonical "LEFT OUTER JOIN" rather than the invalid "ALL LEFT OUTER JOIN", matching the re-pointed All* aliases.
					h = h[(ClickHouseHints.Join.All.Length + 1)..];
```

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review. Individual findings can be wrong, overconfident, or miss context a human would catch — treat them as starting points for discussion, not verdicts.

## Review notes

Build CI green (`build` + `test-all`). Re-review; my earlier findings are all resolved.

**Prior-review status (2026-05-30):**
- [x] **BLK001** (shipped `const` value change `"ALL OUTER"`→`"OUTER"` etc. on minor 6.4.0) — addressed: the new `CHANGELOG.md` entry documents the deprecated-alias SQL behavior change.
- [x] **BLK002** (`AI-Tags` `<remarks>` convention) — addressed: removed entirely.
- [x] **MIN001** (dead-branch comment) — addressed: reworded; the branch is reachable for legacy callers (see MIN006).
- [x] **MIN002 / MIN003 / MIN004** (scope-creep edits to TransactionTests / AsyncExtensions / EnumerableHelper / BulkCopyTests) — addressed: those files are no longer in the diff.
- [x] **MIN005** (dropped deprecated-alias tests) — addressed: regenerated under `#pragma CS0618`; the new `InnerJoinAllHintTest` pins the standalone-ALL fix.
- [x] Copilot's earlier comments (compile errors, missing BOM, partial-class accessibility, interpolated `const`) — all resolved.

**Public API:** the new surface (`Join.GlobalAll`, `JoinGlobalAllHint` ×2) is present in `PublicAPI.Unshipped.txt`; additive on a minor milestone.

## Minor

- [ ] **MIN007** — the ASOF join-hint surface is untested. `JoinAsOfHint` / `JoinGlobalAsOfHint` / `JoinAllAsOfHint` (and `Join.GlobalAsOf`) have no generated test. The previously-removed ASOF test used an equi-join `ON` clause, which ClickHouse rejects at execution (`Code: 403 … ASOF join requires one inequality predicate in JOIN ON expression`), so removing it was correct — but a theta-join with an inequality predicate produces a valid, executable `LEFT ASOF JOIN` and closes the gap. Verified against ClickHouse at this HEAD:

    ```csharp
    [Test]
    public void AsOfJoinHintTest([IncludeDataSources(true, TestProvName.AllClickHouse)] string context)
    {

        var q = db.Child.Join(
            db.Parent.AsClickHouse().JoinAsOfHint(),
            SqlJoinType.Left,
            (c, p) => c.ParentID == p.ParentID && c.ChildID >= p.ParentID,
            (c, p) => p);

        _ = q.ToList();

        Assert.That(LastQuery, Contains.Substring("ASOF JOIN"));
    }
    ```

  Fix: add an ASOF test along these lines, or note in the PR body that ASOF coverage is intentionally deferred.

- [ ] **MIN008** — the branch is 3 commits behind `master` (#5604 `PreferClientCalculation`, #5645, #5656). The baselines PR (linq2db.baselines #1972) currently carries 108 *unrelated* modified files plus an `Oracle.23 TestVarPopOracle.sql.other`; these are stale-branch artifacts (projection / client-calc baselines that #5604 changed on master), not changes this PR makes — verified against baselines `master`. Fix: merge `master` and let CI regenerate so the baselines PR carries only the 60 ClickHouse files.

## Out-of-scope observations

Surfaced during review — included as FYI.

- **CROSS-join strictness — could not reproduce.** I attempted to route a strictness hint into the `CROSS {h} JOIN` arm of `BuildJoinType`; the keyless `from…from…JoinAllHint()` shape throws `InvalidOperationException` from the `default:` arm and never reaches the CROSS path, so an invalid `CROSS ALL JOIN` is not reachable via the API shapes I tried. The bare `InvalidOperationException` on an unsupported hinted join type is a pre-existing rough edge (unchanged here) — FYI only.
- **`CHANGELOG.md` + review-tooling instruction.** This PR adds a root `CHANGELOG.md` release-notes staging file and a matching agent instruction — a cross-cutting convention established via a provider-scoped PR. Owner-authored; noted for awareness.

## Baselines

Delta: [linq2db.baselines #1972](https://github.com/linq2db/linq2db.baselines/pull/1972) (OPEN) — 61 A / 108 M / 0 D (verified three-dot).
- **60 added** ClickHouse join-hint baselines — correct ClickHouse syntax for every variant. The `All*`/bare pairs (`AllAnti`/`Anti`, etc.) emit byte-identical SQL, as expected (the deprecated aliases now re-point to bare values).
- **108 modified + 1 added Oracle `.sql.other`** — stale-branch artifacts, not caused by this PR (see MIN008).

.claude/skills/review-pr/SKILL.md
MaceWindu, line 1: 1. we already have release notes skill
2. all agents changes should go to curation branch, which accumulate changes for periodical merge

CHANGELOG.md
MaceWindu, line 1: this conflicts with approach, used by existing release notes generation skill

MaceWindu (changes requested):

Tests/Linq/Extensions/ClickHouseTests.tt
```diff
@@ -0,0 +1,98 @@
...
+namespace Tests.Extensions
+{
+	public partial class ClickHouseTests
+	{
+<#
+	GenerateLeftJoinHint("Outer",       "LEFT OUTER");
+	GenerateLeftJoinHint("Semi",        "LEFT SEMI");
+	GenerateLeftJoinHint("Anti",        "LEFT ANTI");
+	GenerateLeftJoinHint("Any",         "LEFT ANY");
+	GenerateLeftJoinHint("All",         "LEFT ALL");
```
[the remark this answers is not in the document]
MaceWindu: Still actual at HEAD (`bb7891f`): no ASOF test exists in `ClickHouseTests.tt` / `.generated.cs`. `JoinAsOfHint` / `JoinGlobalAsOfHint` / `JoinAllAsOfHint` (whose value changes `"ALL ASOF"`→`"ASOF"`) are unpinned. An equi-join `ON` is rejected by ClickHouse ASOF, but a theta-join with an inequality predicate (e.g. `c.ChildID >= p.ParentID`) produces a valid, executable `LEFT ASOF JOIN` and closes the gap. Reopening.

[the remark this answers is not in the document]
igor-tkachev: I would remove this API right away, but we do not want to do it for minor releases. This code worked incorrect anyway. And tests will just prove this incorrectness.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

**Added:** new ClickHouse `GLOBAL ALL` join hint (`JoinGlobalAllHint()` / `ClickHouseHints.Join.GlobalAll`), completing the GLOBAL + strictness join-hint combinations.

**Fixed:** a standalone `ALL` ClickHouse join hint produced malformed SQL (it was matched as a compound prefix rather than a standalone strictness modifier); it now emits correctly.

**Deprecated:** the compound `All*` join-hint aliases (`AllOuter`, `AllSemi`, `AllAnti`, …) are deprecated in favor of the standalone strictness hints (`OUTER`, `SEMI`, `ANTI`, …), matching ClickHouse's actual two-axis strictness (ALL/ANY/SEMI/ANTI/ASOF) + distribution (GLOBAL) syntax. The aliases still work for backward compatibility; recompiling maps them to the corrected standalone hints.


_(none)_


---
<sub>Generated from commit `8f9e911`.</sub>

Issue #1876, opened 2019-09-03
Possibility to map bitmask to integer list and filter it


frankiDotNet:
Hi,

is it possible to map a bitmask to an integer list and make filter operations to it?

```
    [PERSON]
    public class Person{
        [ExpressionMethod(nameof(MaskToList), IsColumn = true)]
        public List<int> PermissionList{ get; set;}
    
        public static Expression<Func<Person, List<int>>> MaskToList(){
              return p => new List<int>(){p.Permissions & 1, p.Permissions & 2, p.Permissions & 4};
        }
        [PERMISSIONS]
        public int Permissions{ get; set;}
    }


using(var db = new TestDb())
{
    var result = db.GetTable<Person>().PermissionList.Contains(2).ToList();
}

```

This does not work with the given code!

DB-Provider: Firebird 2.1
LingToDb: 2.7.4
```

MaceWindu:
It is definitely possible to map field to custom type and define custom operations, but I'm afraid with list and contains it will conflict with already existing default mappings. Still you can try.

To define mapping for well-known members, like Contains or for custom operators you can use Expression.MapMember/MapBinary: https://github.com/linq2db/linq2db/issues/1723

But as it probably will conflict with default mapping for Contains, I recommed to define your own extension method, e.g. `ContainsFlags` and map it to SQL.

To define custom type mapping check my answer here https://stackoverflow.com/a/53961787

PS: Don't close this issue if it is answered, as I will use it to track that we have this stuff documented finally.

Franki1986:
Thanks for your help! Unfortunately I am using [DevExtreme.AspNet](https://github.com/DevExpress/DevExtreme.AspNet.Data) framework that only uses the default conatains method, so this helps only partially. But good point.

sdanyliv:
I can not understand how do you plan to use it with DeveExpress. I saw their DataAccess library and i know what to expect from it. How do you plan to feed grid with this data?

Franki1986:
Yes, sadly I followed the code on github and saw that the contains method is only used for string operations, unfortunately the filter operations are not customizable as desired. The idea is nice that the filter is directly translated into a linq expression, but it is restricted to the most use cases.
In fact a long time ago I did start an issue for binary operation filters, but it seems there is no need for this... [](url)..

sdanyliv:
You can create boolean properties for each permission.  In this case DevExpress can easily filter such recordset.
```cs
[Sql.Function(ServerSideOnly = true)]
[...]

PR #5558, opened 2026-05-28
Fix InvalidCastException on correlated Contains subquery (APPLY-join optimization)

igor-tkachev:
While writing the test I ran into an `InvalidCastException` ("Unable to cast SqlColumn to SelectQuery") whenever a `Contains` sits inside a correlated subquery whose source is filtered by an outer column.

It happens during the APPLY→JOIN conversion. `MovingOuterPredicateVisitor` walks the join predicate and pulls inner-only expressions out into columns. Since `SelectQuery` is itself an `ISqlExpression`, the nested IN-subquery got caught by that branch and turned into a `SqlColumn` — which then blew up on the `(SelectQuery)` cast in `VisitInSubQueryPredicate`.

The fix just skips `SelectQuery` there, the same way we already skip `SqlSearchCondition`, so we recurse into it and only rewrite the leaves that actually depend on the inner sources.

The test now goes through `AssertQuery`. Access and ClickHouse are excluded — neither can run a correlated subquery as an IN argument, so that's a backend limitation rather than something we can fix here.


Tests/Linq/Linq/ContainsTests.cs
```diff
@@ -429,5 +429,24 @@ public void Issue4317Test4([IncludeDataSources(true, TestProvName.AllSQLite)] st
 
 			Assert.That(res, Is.EqualTo(expected));
 		}
+
+		[Test]
+		public void ContainsSubqueryTest([DataSources(TestProvName.AllAccess, TestProvName.AllClickHouse)] string context)
```
MaceWindu, line 435: **Major · MAJ001** — `ContainsSubqueryTest` excludes Access and ClickHouse but not YDB, which has the same level-0 correlated-subquery limitation as ClickHouse (`SupportedCorrelatedSubqueriesLevel = 0`). I built this branch and ran the test against a YDB instance: it **fails** (both direct and remote). linq2db's validator lets the simple correlated subquery through and emits

    SELECT a.ParentID FROM Parent a
      CROSS JOIN Child t
    WHERE a.ParentID IN (SELECT c_1.ParentID FROM GrandChild c_1 WHERE c_1.ChildID = t.ChildID)

and the YDB server rejects it with `YdbException … Member not found: t` — it cannot resolve the outer `t` correlation inside the IN-subquery.

This is a *server-side* failure, not linq2db's `Error_Correlated_Subqueries` validation exception, so `[ThrowsRequiresCorrelatedSubquery]` would **not** cover it (that attribute expects a `LinqToDBException`). The correct fix mirrors the existing ClickHouse handling — exclude YDB from `[DataSources]` — matching `SubQueryTests.Test3`/`Test4`, which exclude only ClickHouse for runnable-but-unsupported correlated subqueries.

YDB isn't in the Azure CI baseline matrix, so this doesn't fail the PR's CI run — but it fails for anyone running the suite with YDB enabled (it is in the default `UserDataProviders.json.template`).

Fix: add `ProviderName.Ydb` to the `[DataSources]` exclusion.

```suggestion
		public void ContainsSubqueryTest([DataSources(TestProvName.AllAccess, TestProvName.AllClickHouse, ProviderName.Ydb)] string context)
```

.editorconfig
```diff
@@ -595,6 +595,8 @@ dotnet_diagnostic.IDE0300.severity = none # IDE0300: Collection initialization c
 dotnet_diagnostic.IDE0301.severity = none # IDE0301: Collection initialization can be simplified
 dotnet_diagnostic.IDE0305.severity = none # IDE0305: Collection initialization can be simplified
 dotnet_diagnostic.IDE0320.severity = none # IDE0320: Make anonymous function static
+dotnet_diagnostic.IDE0390.severity = none # IDE0390: Method can be made synchronous
+dotnet_diagnostic.IDE0391.severity = none # IDE0391: Method can be made synchronous
```
MaceWindu, line 599: **Minor · MIN001** — These IDE0390 / IDE0391 suppressions are unrelated to the `InvalidCastException` fix and aren't mentioned in the PR description — scope creep. Bundling unrelated config changes into a bug-fix PR makes the change harder to bisect later.

Fix: drop these two lines from this PR (land them separately if the suppressions are wanted), or note in the PR description why they belong here.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] **Core fix verified correct and complete.** Excluding `SelectQuery` from the column-hoisting branch in `MovingOuterPredicateVisitor` is the right fix: `SelectQuery` is the only `ISqlExpression` subtype that reaches the failing `(SelectQuery)` cast (`VisitInSubQueryPredicate` / `VisitExistsPredicate`, where the operand is statically typed `SelectQuery`), and the change mirrors the existing `and not SqlSearchCondition` exclusion. The inline comment is accurate. No public-API impact (`Internal.*` only).
- [ ] **Prior automated-review threads re-verified at HEAD.** The test rename (→ `ContainsSubqueryTest`) and the switch to `AssertQuery` are addressed. The correlated-subquery provider-exclusion concern is **not** fully addressed — the test fails on YDB (see MAJ001); that thread has been reopened.

## Baselines
Delta: [linq2db.baselines PR #1876](https://github.com/linq2db/linq2db.baselines/pull/1876) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5558)

### `ContainsSubqueryTest` — 51 new SQL baselines (expected)
All 51 are new files for the new regression test. Two shapes, both correct:
- **IN-subquery** (most providers) — `INNER JOIN Child t ON a.ParentID IN (SELECT c_1.ParentID FROM GrandChild c_1 WHERE c_1.ChildID = t.ChildID)`. [sample (DuckDB)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5558/DuckDB/Tests/Linq/ContainsTests/Tests.Linq.ContainsTests.ContainsSubqueryTest%28DuckDB%29.sql)
- **EXISTS** (Firebird 2.5 / 3 / 4 / 5) — `INNER JOIN Child t ON EXISTS(SELECT * FROM GrandChild c_1 WHERE c_1.ChildID = t.ChildID AND a.ParentID = c_1.ParentID)`; the standard Firebird IN→EXISTS rewrite. [sample (Firebird.2.5)](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5558/Firebird.2.5/Tests/Linq/ContainsTests/Tests.Linq.ContainsTests.ContainsSubqueryTest%28Firebird.2.5%29.sql)

Access and ClickHouse are absent as expected (excluded by the test). YDB is also absent — consistent with it not being in the baseline CI matrix and with MAJ001 (it cannot execute this query).

### New `.sql.other` on the branch — unrelated test, worth a look
`Oracle.18.Managed/…/CharTypesTests.StringTrimming(Oracle.18.Managed).sql.other` is **new** on this branch (master carries only the two `Oracle.21` `AnalyticTests` ones). A `.sql.other` file is a **failure indicator** — it is written when a test's direct vs. remote (`LinqService`) SQL diverge, and the test fails with *"Baselines for remote context doesn't match direct access baselines"*. It is for `CharTypesTests.StringTrimming`, which is unrelated to this PR's change, so it is almost certainly the known remote-mode trace limitation (tracked as #5513) surfacing rather than something this PR caused — but it should not ship silently. Worth confirming `CharTypesTests.StringTrimming` on `Oracle.18.Managed` before merging baselines PR #1876.

Tests/Linq/Linq/ContainsTests.cs
```diff
@@ -429,5 +429,25 @@ public void Issue4317Test4([IncludeDataSources(true, TestProvName.AllSQLite)] st
...
+				from t in
+					from t in db.Child
+					where db.GrandChild
+						.Where(c => c.ChildID == t.ChildID)
+						.Select(c => c.ParentID)
+						.Contains(a.ParentID)
+					select t
+				select a.ParentID
+			)
+			.ToList();
```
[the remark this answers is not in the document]
MaceWindu: Reopening — still actual at HEAD. The `AssertQuery` switch and the test rename are addressed, but the provider exclusion is incomplete: the test excludes ClickHouse but not YDB. Running `ContainsSubqueryTest` against YDB on this branch fails — linq2db lets the simple correlated subquery through and emits `… CROSS JOIN Child t … WHERE a.ParentID IN (SELECT … WHERE c_1.ChildID = t.ChildID)`, and the YDB server rejects it (`YdbException: … Member not found: t`). Because that is a server-side error rather than linq2db's `Error_Correlated_Subqueries`, `[ThrowsRequiresCorrelatedSubquery]` does not apply — YDB should be excluded from `[DataSources]` the same way ClickHouse is.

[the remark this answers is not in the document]
sdanyliv: Both points addressed. `AssertQuery` is in place, and the level-0 correlated-subquery case is now asserted with `[ThrowsRequiresCorrelatedSubquery(simple: true)]` (137285de). After merging master, #5574 rejects this correlated IN-subquery on YDB with the clean `LinqToDBException` (`Error_Correlated_Subqueries`) before it reaches the server, so the attribute covers it; ClickHouse stays excluded via `[DataSources]` since its server rejects correlated-IN. Same pattern as `SubQueryTests.Contains1`/`Contains2`.

Tests/Linq/Linq/ContainsTests.cs
sdanyliv: Handled with `[ThrowsRequiresCorrelatedSubquery(simple: true)]` (137285de) instead of a `[DataSources]` exclusion. After merging current master, #5574 sets YDB `IsSupportedSimpleCorrelatedSubqueries = false`, so `SqlQueryValidatorVisitor` now rejects this correlated IN-subquery with the clean `LinqToDBException` (`Error_Correlated_Subqueries`) before it reaches the server — the `YdbException` server path from the original report no longer occurs, and the attribute's expected exception matches. Mirrors `SubQueryTests.Contains1`/`Contains2`, which sit in the `CorrelatedSubquery` category #5574 verified green on YDB.

.editorconfig
sdanyliv: Resolved by the merge with master: #5580 added the same `IDE0390`/`IDE0391` suppressions, so the branch now carries no `.editorconfig` delta (`git diff master...HEAD -- .editorconfig` is empty). Nothing left to drop.

sdanyliv:
Merged current `master` and addressed the open review threads (commit 137285de):

- **Correlated subquery on YDB** (Copilot / MAJ001): now asserted with `[ThrowsRequiresCorrelatedSubquery(simple: true)]` rather than a bare `[DataSources]` exclusion. After the merge, #5574 sets YDB `IsSupportedSimpleCorrelatedSubqueries = false`, so `SqlQueryValidatorVisitor` rejects the correlated IN-subquery with the clean `LinqToDBException` (`Error_Correlated_Subqueries`) before it reaches the server. ClickHouse and Access stay excluded via `[DataSources]`. Mirrors `SubQueryTests.Contains1`/`Contains2`.
- **`.editorconfig` IDE0390/IDE0391** (MIN001): mooted by the merge — #5580 added the same suppressions on master, so the branch no longer carries any `.editorconfig` delta.

Baselines [#1876](https://github.com/linq2db/linq2db.baselines/pull/1876) will need regeneration against the merged tree before this lands.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed an `InvalidCastException` thrown for a query containing a `Contains` (IN) call inside a correlated subquery whose source is filtered by an outer column.


Fixed InvalidCastException on a Contains inside a correlated subquery.


---
<sub>Generated from commit `137285d`.</sub>

PR #5561, opened 2026-05-29
Add NULLS FIRST/LAST support to OrderBy/ThenBy

### LINKED ISSUE #2068: Order by nulls last
Hi,

how can I use order by with nulls last on Postgres?

Thank you for response.

Pavel

sdanyliv:
Fixes #2068

Adds built-in `NULLS FIRST` / `NULLS LAST` control to LINQ ordering, so `NULL` values can be placed first or last without raw-SQL workarounds.

## Per query

New `OrderBy` / `OrderByDescending` / `ThenBy` / `ThenByDescending` overloads taking the existing `Sql.NullsPosition` enum:

```cs
db.Table.OrderBy(x => x.Value, Sql.NullsPosition.Last).ThenBy(x => x.Id);
```

## Default for all queries

A `DataOptions` setting applies a default placement to every ordering key that doesn't specify one:

```cs
var options = new DataOptions().UseDefaultNullsPosition(Sql.NullsPosition.Last);
```

Also available via `Configuration.Sql.DefaultNullsPosition`.

## Providers

Rendered natively where the database supports it (PostgreSQL, Oracle, DB2, Firebird, SQLite, DuckDB, ClickHouse, SAP HANA) and emulated automatically for the rest (SQL Server, MySQL, MariaDB, Access, Sybase, …), so results are consistent across all providers.

## Also covered

The requested position (and the `DataOptions` default) flows through ordering that the provider lowers internally — `DISTINCT` / set operations, `DistinctBy` / `UnionBy` / `ExceptBy` / `IntersectBy`, `MinBy` / `MaxBy`, indexed `Select((x, i) => …)` and other `ROW_NUMBER()` paths, and aggregate ordering (`String.Join` / `StringAggregate`). It is a no-op for ordering keys that can't be null (skipped, so no `CASE`/token is emitted).

## Redundant ordering elided

Each provider's natural `NULL` placement is captured via `SqlProviderFlags.DefaultNullsOrdering` (new `NullsDefaultOrdering` enum: `Smallest` / `Largest` / `AlwaysFirst` / `AlwaysLast`). When a requested position already matches the provider's natural order for the sort direction, no emulation key or native `NULLS` token is emitted — so index-backed ordering isn't defeated by a redundant computed key. An explicit `Sql.NullsPosition.None` opts a key out of the configured `DataOptions` default.

## Comparer overloads

The BCL `OrderBy`/`ThenBy(keySelector, IComparer<TKey>)` overloads have no SQL equivalent and are now rejected with a translation error, rather than being silently translated as if the comparer were absent.


Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -2432,6 +2441,47 @@ protected virtual void BuildExpressionForOrderBy(ISqlExpression expr)
...
+			if (item.NullsPosition == Sql.NullsPosition.None || IsNullsOrderingSupported)
+				return null;
+
+			var nullsLast = item.NullsPosition == Sql.NullsPosition.Last;
+
+			return new SqlConditionExpression(
+				new SqlPredicate.IsNull(item.Expression, false),
+				new SqlValue(nullsLast ? 1 : 0),
+				new SqlValue(nullsLast ? 0 : 1));
+		}
```
MaceWindu: **Major · MAJ002** — The `CASE WHEN … IS NULL` NULLS-emulation key is synthesized only here, at render time, and never enters the SQL AST. The optimizer passes that make `ORDER BY` valid for `DISTINCT` / set-operation queries — which require every ordering expression to appear in the projected select list — therefore never see it. On an emulated provider (SQL Server, MySQL, MariaDB, Access, Sybase), `Select(…).Distinct().OrderBy(x => x.Nullable, Sql.NullsPosition.Last)` renders `ORDER BY CASE WHEN col IS NULL THEN … END, col` where the CASE is not in the `DISTINCT` select list — SQL Server rejects this at execution (for `SELECT DISTINCT`, every `ORDER BY` item must appear in the select list). Same failure mode for `UNION` / `EXCEPT` / `INTERSECT` combined with a nullable NULLS key. No test covers `Distinct` / set-operations combined with a `NullsPosition`.

Fix: represent the null-grouping key in the SQL AST (e.g. inject it as an order-by / projection element during optimization on emulated providers) before the ORDER-BY-correction and DISTINCT/set-op column-promotion passes run, so those passes project it into the select list when required — or extend those passes to account for the synthetic key. Add regression coverage for `Distinct().OrderBy(x => x.Nullable, Sql.NullsPosition.Last)` and a `Union(...).OrderBy(... NullsPosition ...)` on SQL Server.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Both open Copilot review threads are confirmed still-actual at HEAD — the ROW_NUMBER pagination null-order drop (**MAJ001** below) and the DISTINCT/set-operation emulation-key projection gap (**MAJ002**, posted inline on `BasicSqlBuilder.cs`). Left open so the discussion stays anchored.
- [x] Public API: 8 new public members under `LinqToDB` / `LinqToDB.Common` (4 `OrderBy`/`ThenBy` overloads, `SqlOptions.DefaultNullsPosition`, `WithDefaultNullsPosition`, `UseDefaultNullsPosition`, `Configuration.Sql.DefaultNullsPosition`) — all carry XML docs and are recorded in `PublicAPI.Unshipped.txt`. Appropriate additive surface for the 6.4.0 milestone.
- [ ] No baseline branch exists yet — CI has not generated baselines for this PR. These `ORDER BY` changes will move baselines on the next CI run; re-check before merge.

### Major (MAJ)

- [x] **MAJ001** — ROW_NUMBER rewrites drop the requested NULLS position (2 sites)
  Why: the PR threads `NullsPosition` through the order-by rewrites it touched (`BasicSqlOptimizer` / `Oracle11SqlOptimizer` skip-take, `QueryHelper.MoveOrderByUp`, etc.), but two `SqlOrderByItem` → `SqlWindowOrderItem` conversions still hardcode `Sql.NullsPosition.None`:
  - `Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs:1259` — Skip/Take pagination on ROW_NUMBER providers (e.g. SQL Server 2008).
  - `Source/LinqToDB/Internal/Linq/Builder/SelectBuilder.cs:101` — the `Select((x, i) => …)` row-index rewrite.
  So `OrderBy(x => x.Value, Sql.NullsPosition.Last)` combined with `Skip`/`Take` (on ROW_NUMBER providers) or indexed `Select` numbers/pages rows using the provider default null order instead of NULLS LAST. This is the internal pagination rewrite of a plain `OrderBy`, distinct from window-function ordering. `SqlWindowOrderItem` already accepts a position (its 3rd constructor parameter), so the value is silently discarded.
  Fix: pass `o.NullsPosition` instead of `Sql.NullsPosition.None` at both sites. Add regression coverage for `Skip`/`Take` and indexed-`Select` with an explicit position on a nullable key.

### Minor (MIN)

- [x] **MIN001** — Emulation key emitted even when the requested position already matches the provider default
  Why: `GetOrderByNullsEmulationKey` in `Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs` emits the `CASE WHEN <expr> IS NULL THEN … END` key for any explicit position on an emulated provider, even when that position already matches the provider's natural NULL collation (SQL Server / MySQL / SQLite sort `NULL` low ⇒ default `NULLS FIRST` for ASC, `NULLS LAST` for DESC). Correctness-neutral, but it adds a redundant sort key and moves more baselines than necessary on the common case.
  Fix: add a provider-overridable "NULLs sort low/high" hook and skip the emulation key when the requested position equals the provider default for the item's sort direction. (Always-emitting is defensible for simplicity — flagging the redundancy so it's a conscious choice.)

- [x] **MIN002** — Two of the three default-position surfaces have no direct test
  Why: only the per-query overloads and `UseDefaultNullsPosition` (DataOptions) are exercised. `DataOptionsExtensions.WithDefaultNullsPosition(SqlOptions, …)` and `Configuration.Sql.DefaultNullsPosition` (process-global) have no test of their own. The shared implementation is covered via the DataOptions path, so Minor — but neither entry point is pinned against a future refactor.
  Fix: add a unit assertion for `WithDefaultNullsPosition`, and a `try`/`finally`-guarded test for the global that restores the previous value so the static mutation doesn't leak into sibling tests.

## Baselines
No baseline branch exists yet — CI has not run for this PR. The `ORDER BY` SQL changes here will move baselines on the next CI run.

Source/LinqToDB/Internal/Linq/Builder/OrderByBuilder.cs
```diff
@@ -132,15 +141,22 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
...
 					}
 				}
+				else if (itemNulls == Sql.NullsPosition.None && !byIndex)
+				{
+					// No explicit position on a normal ordering key — apply the configured default.
+					itemNulls = defaultNulls;
+				}
 
-				sequence.SelectQuery.OrderBy.Expr(orderSql, methodCall.Method.Name.EndsWith("Descending", StringComparison.Ordinal), isPositioned);
+				sequence.SelectQuery.OrderBy.Expr(orderSql, methodCall.Method.Name.EndsWith("Descending", StringComparison.Ordinal), isPositioned, itemNulls);
```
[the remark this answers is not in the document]
sdanyliv: Resolved in b9dc5ffbc4. `NullsPosition` is now propagated into both `SqlOrderByItem` → `SqlWindowOrderItem` conversions (`SelectBuilder.CounterContext` and the skip/take path in `SelectQueryOptimizerVisitor`), and `OVER(ORDER BY …)` rendering emulates `NULLS` via a `CASE WHEN … IS NULL` key on providers without native support. Coverage: `SelectTests.SelectWithIndexerOrderByNulls` (First/Last × asc/desc).

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -2413,11 +2413,20 @@ protected virtual void BuildOrderByClause(SelectQuery selectQuery)
 					}
 				}
 
+				var nullsEmulationKey = GetOrderByNullsEmulationKey(item);
+				if (nullsEmulationKey != null)
+				{
+					BuildExpressionForOrderBy(nullsEmulationKey);
+					StringBuilder.Append(InlineComma);
```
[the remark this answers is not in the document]
sdanyliv: Resolved in b9dc5ffbc4. The NULLS emulation key is now lowered into the SQL AST before optimization (`SqlNullsOrderingLoweringVisitor`, run in `BasicSqlOptimizer.Finalize` before `OptimizeQueries`, gated on `!SqlProviderFlags.IsNullsOrderingSupported`) as a `CASE WHEN … IS NULL` order expression, so the DISTINCT / set-operation column-promotion and sub-query wrapping passes treat it as an ordinary derived order expression and project it where required. Coverage: `DistinctTests.DistinctOrderByNulls`.

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
sdanyliv: Resolved in b9dc5ffbc4 (+ test in ed4fa592c9). The null-grouping key is now represented in the SQL AST before the order-by-correction and DISTINCT/set-op column-promotion passes: `SqlNullsOrderingLoweringVisitor` (run in `BasicSqlOptimizer.Finalize` before `OptimizeQueries`, only when `!SqlProviderFlags.IsNullsOrderingSupported`) rewrites a NULLS-positioned `ORDER BY` item into a leading `CASE WHEN … IS NULL` derived order item. The existing DISTINCT/set-op sub-query wrapping then projects it where required (verified: ordering by a derived expression under DISTINCT already wraps into a sub-query). Regression coverage added — `DistinctTests.DistinctOrderByNulls` and `ConcatUnionTests.UnionOrderByNulls`, both First/Last × asc/desc. SQL Server and the other emulated providers run via the CI provider matrix.

Source/LinqToDB/Internal/Linq/Builder/OrderByBuilder.cs
```diff
@@ -114,13 +114,22 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
 				return BuildSequenceResult.Error(methodCall);
 			}
 
+			// New OrderBy/ThenBy overloads accept an explicit Sql.NullsPosition as the third argument.
+			var nullsPosition = methodCall.Arguments.Count > 2
+				? (Sql.NullsPosition)builder.EvaluateExpression(methodCall.Arguments[2])!
+				: Sql.NullsPosition.None;
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 55930b104a. The third argument is now read as `Sql.NullsPosition` only when the method's third parameter type is `Sql.NullsPosition`; otherwise it falls back to `None`. So the BCL `OrderBy`/`ThenBy(keySelector, IComparer<TKey>)` overloads keep their previous translation behavior instead of casting the comparer.

Source/LinqToDB/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,13 @@
...
+static LinqToDB.LinqExtensions.ThenByDescending<TSource, TKey>(this System.Linq.IOrderedQueryable<TSource>! source, System.Linq.Expressions.Expression<System.Func<TSource, TKey>!>! keySelector, LinqToDB.Sql.NullsPosition nulls) -> System.Linq.IOrderedQueryable<TSource>!
+LinqToDB.SqlOptions.DefaultNullsPosition.get -> LinqToDB.Sql.NullsPosition
+LinqToDB.SqlOptions.DefaultNullsPosition.init -> void
+static LinqToDB.Common.Configuration.Sql.DefaultNullsPosition.get -> LinqToDB.Sql.NullsPosition
+static LinqToDB.Common.Configuration.Sql.DefaultNullsPosition.set -> void
+static LinqToDB.DataOptionsExtensions.WithDefaultNullsPosition(this LinqToDB.SqlOptions! options, LinqToDB.Sql.NullsPosition defaultNullsPosition) -> LinqToDB.SqlOptions!
+static LinqToDB.DataOptionsExtensions.UseDefaultNullsPosition(this LinqToDB.DataOptions! options, LinqToDB.Sql.NullsPosition defaultNullsPosition) -> LinqToDB.DataOptions!
+LinqToDB.Internal.SqlProvider.SqlProviderFlags.IsNullsOrderingSupported.get -> bool
+LinqToDB.Internal.SqlProvider.SqlProviderFlags.IsNullsOrderingSupported.set -> void
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 55930b104a. The remaining new/changed public surface is now recorded in `PublicAPI.Unshipped.txt`: the `SqlOrderByItem` constructor + `NullsPosition` property, the new `SqlOrderByClause.Expr` overload, the `SqlExpressionOptimizerVisitor.VisitSqlOrderByItem` override, and the `WindowFunctionHelpers` tuple signature changes (new signatures added, old ones marked `*REMOVED*`). Verified with `-p:RunApiAnalyzersDuringBuild=true` — no RS0016/RS0017 remain for these symbols.

Source/LinqToDB/Internal/Linq/Builder/DistinctByBuilder.cs
```diff
@@ -19,13 +19,13 @@ public static bool CanBuildMethod(MethodCallExpression call)
...
 
-		static Expression BuildDistinctByViaRowNumber<T>(ExpressionBuilder builder, IBuildContext sequence, Expression[] partitionPart, (LambdaExpression lambda, bool isDescending)[] orderByPart)
+		static Expression BuildDistinctByViaRowNumber<T>(ExpressionBuilder builder, IBuildContext sequence, Expression[] partitionPart, (LambdaExpression lambda, bool isDescending, Sql.NullsPosition nulls)[] orderByPart)
 		{
 			var           contextRef = new ContextRefExpression(typeof(IQueryable<T>), sequence);
 			IQueryable<T> query      = new ExpressionQueryImpl<T>(builder.DataContext, contextRef);
 
 			var orderByPrepared = orderByPart
-				.Select(o => (SequenceHelper.PrepareBody(o.lambda, sequence), o.isDescending))
+				.Select(o => (SequenceHelper.PrepareBody(o.lambda, sequence), o.isDescending, o.nulls))
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 3c7f9e16c0. After `ExtractOrderByPart`, keys with `Sql.NullsPosition.None` are now mapped to `builder.DataOptions.SqlOptions.DefaultNullsPosition` before building the ROW_NUMBER, so `UseDefaultNullsPosition(...)` is applied to the DistinctBy window ordering. Regression test: `DistinctByMethodTests.DistinctByDefaultNullsPosition` (plain OrderBy under the default matches the explicit overload).

Source/LinqToDB/Internal/Linq/Builder/SetOperationByBuilder.cs
```diff
@@ -268,14 +268,14 @@ static Expression BuildUnionBy(Expression source, Expression second, LambdaExpre
...
-		static (Expression expr, bool isDescending)[] BuildOrderByPart(Expression source, Expression parameter)
+		static (Expression expr, bool isDescending, Sql.NullsPosition nulls)[] BuildOrderByPart(Expression source, Expression parameter)
 		{
 			var orderByPart = WindowFunctionHelpers.ExtractOrderByPart(source, out _);
 			if (orderByPart.Length == 0)
 				return [];
 
 			return orderByPart
-				.Select(o => (o.lambda.GetBody(parameter), o.isDescending))
+				.Select(o => (o.lambda.GetBody(parameter), o.isDescending, o.nulls))
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 3c7f9e16c0. `BuildOrderByPart` now maps `Sql.NullsPosition.None` to `DataOptions.SqlOptions.DefaultNullsPosition` (threaded in from `BuildMethodCall`) before `BuildRowNumber`, so `ExceptBy`/`IntersectBy`/`UnionBy` apply the configured default to the dedup window ordering. The same fix was applied to `MinBy`/`MaxBy` tiebreaker extraction. Regression test: `UnionByMethodTests.UnionByDefaultNullsPosition`.

Source/LinqToDB/Internal/SqlQuery/Visitors/SqlNullsOrderingLoweringVisitor.cs
```diff
@@ -0,0 +1,55 @@
+namespace LinqToDB.Internal.SqlQuery.Visitors
```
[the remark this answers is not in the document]
sdanyliv: Fixed in c4419dfa0c — re-saved `SqlNullsOrderingLoweringVisitor.cs` with the UTF-8 BOM to match the `.editorconfig` `utf-8-bom` convention.

Tests/Linq/Linq/UnionByMethodTests.cs
```diff
@@ -16,9 +16,45 @@ public class UnionByMethodTests : TestBase
...
+				.OrderBy(x => x.Priority)
+				.UnionBy(rightTable, x => x.Key)
+				.OrderBy(x => x.Key).Select(x => x.Id).ToList();
+
+			var byExplicit = leftTable
+				.OrderBy(x => x.Priority, Sql.NullsPosition.Last)
+				.UnionBy(rightTable, x => x.Key)
+				.OrderBy(x => x.Key).Select(x => x.Id).ToList();
+
+			Assert.That(byDefault, Is.EqualTo(byExplicit));
```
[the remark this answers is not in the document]
sdanyliv: Fixed in c4419dfa0c — switched to `byDefault.ShouldBe(byExplicit)` (added `using Shouldly;`).

Tests/Linq/Linq/DistinctByMethodTests.cs
```diff
@@ -56,6 +57,53 @@ public void DistinctBy([DataSources] string context)
...
+				.OrderBy(t => t.Priority).ThenBy(t => t.Id)
+				.DistinctBy(x => x.Group)
+				.OrderBy(x => x.Group).Select(x => x.Id).ToList();
+
+			var byExplicit = table
+				.OrderBy(t => t.Priority, Sql.NullsPosition.Last).ThenBy(t => t.Id)
+				.DistinctBy(x => x.Group)
+				.OrderBy(x => x.Group).Select(x => x.Id).ToList();
+
+			Assert.That(byDefault, Is.EqualTo(byExplicit));
```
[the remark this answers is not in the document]
sdanyliv: Fixed in c4419dfa0c — switched to `byDefault.ShouldBe(byExplicit)` (added `using Shouldly;`).

Source/LinqToDB/Internal/Linq/Builder/OrderByBuilder.cs
```diff
@@ -114,13 +114,27 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
...
+			// TODO: a custom IComparer<TKey> cannot be translated to SQL, yet CanBuildMethod still accepts
+			// those overloads and we silently ignore the comparer (ordering by the key only). Should they be
+			// rejected in CanBuildMethod instead of producing a translation that disregards the comparer?
+			var nullsPosition = methodCall.Arguments.Count > 2
+				&& methodCall.Method.GetParameters()[2].ParameterType == typeof(Sql.NullsPosition)
+				? (Sql.NullsPosition)builder.EvaluateExpression(methodCall.Arguments[2])!
+				: Sql.NullsPosition.None;
+
+			// When no position is specified, fall back to the configured default.
+			var defaultNulls = builder.DataOptions.SqlOptions.DefaultNullsPosition;
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 50a415d9c0. `OrderByBuilder` now tracks whether the NULLS argument was supplied (`nullsSpecified`) and applies `DefaultNullsPosition` only when it wasn't — an explicit `Sql.NullsPosition.None` is honored as a provider-default opt-out. Regression test `OrderByTests.ExplicitNullsNoneOptsOutOfDefault` asserts the explicit-None result is invariant to the configured default.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
```diff
@@ -60,51 +67,70 @@ public static Expression BuildWindowDefinition(Expression[] partitionBy, (Expres
...
+							supported = false;
+							break;
+					}
+				}
+				// linq2db OrderBy/ThenBy overloads that carry an explicit Sql.NullsPosition.
+				else if (typeof(LinqExtensions) == mc.Method.DeclaringType
+					&& mc.Arguments.Count == 3
+					&& mc.Method.GetParameters()[2].ParameterType == typeof(Sql.NullsPosition))
+				{
+					var nulls = (Sql.NullsPosition)((ConstantExpression)mc.Arguments[2]).Value!;
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 50a415d9c0. `ExtractOrderByPart` now returns `Sql.NullsPosition?` (null = unspecified plain `OrderBy`, a concrete value including `None` = explicit). `DistinctByBuilder`/`SetOperationByBuilder`/`MinByMaxByBuilder` resolve `?? DefaultNullsPosition`, and re-emitted orderings carry the resolved position explicitly, so an explicit `None` is preserved through the ROW_NUMBER and tiebreaker rewrites instead of being replaced by the default.

Source/LinqToDB/Internal/SqlQuery/Visitors/SqlNullsOrderingLoweringVisitor.cs
```diff
@@ -0,0 +1,55 @@
...
+					new SqlPredicate.IsNull(item.Expression, false),
+					new SqlValue(nullsLast ? 1 : 0),
+					new SqlValue(nullsLast ? 0 : 1));
+
+				// Replace the NULLS-positioned item with a leading ascending CASE key + the plain item.
+				element.Items[i] = new SqlOrderByItem(nullsKey, false, false, Sql.NullsPosition.None);
+				element.Items.Insert(i + 1, new SqlOrderByItem(item.Expression, item.IsDescending, item.IsPositioned, Sql.NullsPosition.None));
+
+				i++; // skip the inserted plain item
+			}
```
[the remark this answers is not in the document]
sdanyliv: Fixed in dd7a580e35. The lowering visitor now resolves the clause's `NullabilityContext` and, for a non-nullable ordering key, drops the position to `None` instead of injecting the CASE key — matching `SqlExpressionOptimizerVisitor.VisitSqlOrderByItem`, which already does this for providers with native NULLS support. Verified on MySQL/ClickHouse: ordering a non-nullable key with an explicit `Sql.NullsPosition` now emits a plain `ORDER BY` with no CASE.

Tests/Linq/Linq/StringFunctionsTests.cs
```diff
@@ -253,6 +253,72 @@ public void FinalAggregationTest([StringTestSources] string context)
...
+
+			var actual = ordered.Select(x => x.Value2).StringAggregate(" -> ").ToValue();
+
+			var grouped = nulls == Sql.NullsPosition.First
+				? data.OrderBy(d => d.Value1 != null)   // null group first
+				: data.OrderBy(d => d.Value1 == null);  // null group last
+			var expectedSeq = descending ? grouped.ThenByDescending(d => d.Value1) : grouped.ThenBy(d => d.Value1);
+			var expected    = Sql.ConcatStringsNullable(" -> ", expectedSeq.Select(d => d.Value2));
+
+			Assert.That(actual, Is.EqualTo(expected));
```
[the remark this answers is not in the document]
sdanyliv: Fixed in 7eec6d33ff. Converted the new aggregate NULLS assertions to Shouldly (`actual.ShouldBe(expected)`) and added `using Shouldly;`.

Source/LinqToDB/Internal/Linq/Builder/OrderByBuilder.cs
```diff
@@ -114,13 +114,29 @@ protected override BuildSequenceResult BuildMethodCall(ExpressionBuilder builder
...
+			var nullsSpecified = methodCall.Arguments.Count > 2
+				&& methodCall.Method.GetParameters()[2].ParameterType == typeof(Sql.NullsPosition);
+			var nullsPosition = nullsSpecified
+				? (Sql.NullsPosition)builder.EvaluateExpression(methodCall.Arguments[2])!
+				: Sql.NullsPosition.None;
+
+			// When no position is specified, fall back to the configured default. An explicit Sql.NullsPosition.None
+			// is honored as-is (opt out of the default) and must not be treated as "unspecified".
+			var defaultNulls = builder.DataOptions.SqlOptions.DefaultNullsPosition;
+
```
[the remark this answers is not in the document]
sdanyliv: Done in 98ae177fb4. `OrderByBuilder.CanBuildMethod` now accepts only the 2-arg key-selector and the 3-arg `Sql.NullsPosition` overloads; any other extra-argument overload (the BCL `IComparer<TKey>` ones) is declined, so it surfaces as a normal "cannot be converted to SQL" `LinqToDBException` instead of silently ordering by the key. Regression test: `OrderByTests.OrderByWithComparerIsNotTranslated`.

Source/LinqToDB/SqlOptions.cs
```diff
@@ -54,8 +54,18 @@ public SqlOptions() : this(false)
...
 			EnableConstantExpressionInOrderBy = original.EnableConstantExpressionInOrderBy;
 			GenerateFinalAliases              = original.GenerateFinalAliases;
+			DefaultNullsPosition              = original.DefaultNullsPosition;
 		}
 
+		/// <summary>
+		/// Default position of <c>NULL</c> values in an <c>ORDER BY</c> clause for ordering keys that do not
+		/// specify a <see cref="Sql.NullsPosition"/> explicitly. When set to <see cref="Sql.NullsPosition.First"/>
+		/// or <see cref="Sql.NullsPosition.Last"/>, it is applied to every <c>OrderBy</c>/<c>ThenBy</c> key (and
+		/// emulated for providers without native <c>NULLS FIRST</c>/<c>NULLS LAST</c> support).
```
MaceWindu: **Suggestion · SUG001** — The XML doc is internally inconsistent on precedence. The opening clause correctly scopes the default to keys "that do not specify a `Sql.NullsPosition` explicitly", but this sentence then says it "is applied to every `OrderBy`/`ThenBy` key", which reads as overriding even keys that pass an explicit position. The actual behavior (`OrderByBuilder` short-circuits the default when the position was supplied; an explicit `Sql.NullsPosition.None` opts out) is that an explicit per-call position always wins.

Fix: scope the sentence to keys that don't pass an explicit position, and state the precedence.

```suggestion
		/// specify a <see cref="Sql.NullsPosition"/> explicitly. When set to <see cref="Sql.NullsPosition.First"/>
		/// or <see cref="Sql.NullsPosition.Last"/>, it is applied to every <c>OrderBy</c>/<c>ThenBy</c> key that does
		/// not pass an explicit position (a per-call <see cref="Sql.NullsPosition"/> always wins; an explicit
		/// <see cref="Sql.NullsPosition.None"/> opts out of the default), emulated for providers without native
		/// <c>NULLS FIRST</c>/<c>NULLS LAST</c> support.
```

Source/LinqToDB/Internal/SqlQuery/Visitors/SqlNullsOrderingLoweringVisitor.cs
MaceWindu, line 1: **Minor · MIN004** — Emulation key emitted even when the requested position already matches the provider's natural NULL collation. On emulated providers (SQL Server, MySQL, SQLite, ...) `NULL` sorts low by default => the natural order is already `NULLS FIRST` for ASC and `NULLS LAST` for DESC. The lowering visitor (and the `OVER` builder / MySQL / DB2 emulation paths) emit the leading `CASE`/`IIF`/boolean key whenever `NullsPosition != None` and the key is nullable — with no check that the requested position equals the provider default for that direction. So `OrderBy(x => x.Value, Sql.NullsPosition.First)` (ASC) on SQL Server adds a redundant `IIF(... IS NULL, 0, 1)` key even though plain ASC already places `NULL`s first. (Raised previously as MIN001 and unaddressed; the non-nullable-key no-op the PR added is a different optimization.)

Beyond baseline noise, this is a possible **performance** regression: the synthetic `CASE`/`IIF` ordering key is a computed expression, so a query whose `ORDER BY` could otherwise be satisfied by an index on the column now has to sort — even when the requested position matched the index's natural null ordering. Worth confirming on the index-backed ordering / pagination paths.

Fix: add a provider-overridable "NULLs sort low/high" capability and skip the emulation key when the requested position already equals the provider default for the item's sort direction. (Always-emitting is defensible for simplicity — flagging the redundancy + the index-usage concern so it's a conscious choice.)

Source/LinqToDB/Linq/Translation/AggregateFunctionBuilder.cs
MaceWindu, line 1: **Minor · MIN005** — The non-nullable NULLS no-op the PR adds here (drop the position when `obSql!.CanBeNullable(orderNullability)` is false, where `orderNullability = NullabilityContext.GetContext(raw.SelectQuery)`) does not fire for the aggregate `ORDER BY` on native providers. Baselines for `StringFunctionsTests.AggregationNonNullableNullsNoop` show PostgreSQL (13–18) and SapHana.Odbc emitting `STRING_AGG(... ORDER BY "Id" NULLS FIRST)` on the non-nullable `Id` key, while Oracle / SQL Server omit the token (their `LISTAGG`/`STRING_AGG ... WITHIN GROUP` syntax has no NULLS slot, so they avoid it for a different reason). Since the `nulls` value is computed provider-independently, the fact that PG/SapHana still render it means `CanBeNullable(orderNullability)` returned true for a non-nullable key in the aggregate `SelectQuery` context — i.e. the nullability context isn't resolving the key as non-nullable there, unlike the top-level `ORDER BY` path (where `OrderByNullsNonNullableColumn` is a clean no-op across all providers).

Correctness-neutral (`NULLS FIRST` on a non-nullable column changes nothing), but it contradicts the PR's stated invariant and the companion test's name.

Fix: confirm whether `NullabilityContext.GetContext(raw.SelectQuery)` is the right context for the aggregate ordering key (it appears not to see the base-table column's non-nullability), or accept the redundant token as intentional and clarify the test.

Source/LinqToDB/DataOptionsExtensions.cs
MaceWindu, line 1: **Suggestion · SUG002** — `WithDefaultNullsPosition(SqlOptions, ...)` has no direct test. It is exercised transitively via the remote round-trip (`LinqServiceSerializer.DeserializeOptions` calls it), so any remote-context run of the default tests covers it — hence SUG, not MIN. But the `SqlOptions`-level fluent extension is a documented public surface; if its `options with { DefaultNullsPosition = ... }` wiring regressed, only the remote path would catch it.

Fix (optional): add a small unit assertion — `new SqlOptions().WithDefaultNullsPosition(Sql.NullsPosition.Last).DefaultNullsPosition.ShouldBe(Sql.NullsPosition.Last)` (and that the original instance is unchanged) — or accept the transitive remote coverage as sufficient.

Source/LinqToDB/Common/Configuration.cs
MaceWindu, line 1: **Minor · MIN006** — Of the three default-NULLS surfaces this PR exposes, `Configuration.Sql.DefaultNullsPosition` (the process-global static) has no test at all — a search of the test tree finds zero references to it. It is a distinct public surface with process-wide semantics: its setter does `Options = Options with { ... }`, so it affects every `DataOptions` built afterward. The DataOptions surface (`UseDefaultNullsPosition`) is well covered, but a regression in the global getter/setter wiring would be caught by nothing. (Raised previously as part of MIN002; still open.)

Fix: add a test that sets `Configuration.Sql.DefaultNullsPosition = Sql.NullsPosition.Last` inside a `try`/`finally` that restores the previous value, builds a fresh context, runs a plain `OrderBy` over a nullable column, and asserts the result matches the explicit `Sql.NullsPosition.Last` overload. The `try`/`finally` is mandatory — the global leaks into every sibling test otherwise.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
MaceWindu, line 1: **Minor · MIN003** — `ExtractOrderByPart` treats any `Queryable`/`Enumerable` `OrderBy`/`ThenBy`/`OrderByDescending`/`ThenByDescending` as supported without checking the argument count (the `typeof(Queryable) == ... || typeof(Enumerable) == ...` branch reads `mc.Arguments[1]` as the key selector and adds the item unconditionally). The 3-arg BCL comparer overload (`OrderBy(keySelector, IComparer<TKey>)`) therefore matches, and its comparer at `mc.Arguments[2]` is silently dropped when the preceding ordering is extracted for a `DistinctBy` / `UnionBy` / `ExceptBy` / `IntersectBy` / `MinBy` / `MaxBy` `ROW_NUMBER` rewrite.

This PR added a comparer rejection to `OrderByBuilder.CanBuildMethod` (so a standalone comparer-`OrderBy` now surfaces a translation error), but the same fix was not extended here — leaving an inconsistency: the main `OrderBy` path rejects the comparer overload, while the same overload feeding `DistinctBy`/`UnionBy`/`MinBy` is silently mis-translated (ordered by the key only). This is the open review thread on `ExtractOrderByPart`.

Fix: gate the `Queryable`/`Enumerable` branch on `mc.Arguments.Count == 2` so the comparer overloads fall through to `supported = false` (rejected as not convertible), matching `CanBuildMethod`. Add a regression test for `.OrderBy(x => x.Key, comparer).DistinctBy(...)` asserting it is not silently translated.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
```diff
@@ -60,51 +67,73 @@ public static Expression BuildWindowDefinition(Expression[] partitionBy, (Expres
...
 				var mc = (MethodCallExpression)current;
+
+				var supported = true;
+
 				if (typeof(Queryable) == mc.Method.DeclaringType || typeof(Enumerable) == mc.Method.DeclaringType)
 				{
-					var supported = true;
 					switch (mc.Method.Name)
 					{
 						case nameof(Enumerable.OrderBy):
```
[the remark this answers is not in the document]
MaceWindu: Confirmed still-actual at HEAD (bd55fc6). The comparer rejection added in 98ae177fb4 covers `OrderByBuilder.CanBuildMethod` but not `WindowFunctionHelpers.ExtractOrderByPart` — its `Queryable`/`Enumerable` branch still reads `mc.Arguments[1]` without checking `mc.Arguments.Count`, so a 3-arg comparer `OrderBy` feeding `DistinctBy`/`UnionBy`/`MinBy` has its comparer silently dropped. Filed as **MIN003** with a suggested `mc.Arguments.Count == 2` gate.

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review, so this complements — rather than replaces — human judgement. Individual findings can be wrong, overconfident, or miss context a human would catch. Feel free to disagree, dismiss, or ask for clarification on any finding — the agent's judgement is not authoritative.

## Review notes
- [x] **Prior MAJ001 (ROW_NUMBER rewrites dropped the requested NULLS position) — fixed.** Both `SqlOrderByItem` -> `SqlWindowOrderItem` conversions (`SelectBuilder`, `SelectQueryOptimizerVisitor`) now carry the position; baselines confirm `Skip`/`Take` and indexed-`Select` emit the key.
- [x] **Prior MAJ002 (DISTINCT/set-op emulation key invisible to the AST passes) — fixed.** `SqlNullsOrderingLoweringVisitor` lowers the `CASE` key into the AST before optimization; SQL Server `Distinct`/`Union` baselines show it correctly projected into the select list.
- [x] **Prior MIN001 (emulation key emitted even when the requested position matches the provider default) — not addressed.** Still emits the redundant key on the common matching case — see **MIN004** (now also a possible perf concern).
- [x] **Prior MIN002 (untested default surfaces) — partially addressed.** The DataOptions path is now well covered; `Configuration.Sql.DefaultNullsPosition` still has zero tests (**MIN006**) and `WithDefaultNullsPosition(SqlOptions, ...)` only transitive coverage (**SUG002**).
- [x] **Open `ExtractOrderByPart` thread — still actual.** The comparer-rejection added to `OrderByBuilder.CanBuildMethod` was not extended to `WindowFunctionHelpers.ExtractOrderByPart`, so a comparer-based `OrderBy` feeding `DistinctBy`/`UnionBy`/`MinBy` silently drops the comparer — see **MIN003**.
- [x] **Public API:** 8 additive public members under `LinqToDB` / `LinqToDB.Common` (4 `OrderBy`/`ThenBy` `Sql.NullsPosition` overloads, `SqlOptions.DefaultNullsPosition`, `WithDefaultNullsPosition`, `UseDefaultNullsPosition`, `Configuration.Sql.DefaultNullsPosition`) — all carry XML docs and are recorded in `PublicAPI.Unshipped.txt`. Appropriate additive surface for 6.4.0. The `WindowFunctionHelpers` tuple-signature changes are `Internal.*` (breaking but policy-exempt).

## Out-of-scope observations

Surfaced during review but fall outside this PR's scope. Not findings — included as FYI.

- **YDB native-NULLS render branch is dead code, and there is no YDB NULLS test.** The PR adds a native `NULLS FIRST/LAST` branch to `YdbSqlBuilder.BuildOrderByClause`, but YDB does not set `IsNullsOrderingSupported`, so the lowering visitor strips the position to `None` before SQL build — the branch never fires. YQL does support native `NULLS FIRST/LAST`, so YDB could instead set the flag and render natively. Either way no YDB NULLS-ordering test is added (YDB is experimental, and the PR touches `YdbSqlBuilder`).
- **`WindowFunctionHelpers` tuple-signature change is breaking for `Internal.*` consumers.** `ApplyOrderBy`/`BuildRowNumber`/`BuildWindowDefinition`/`ExtractOrderByPart` changed their order-tuple element type to add a `Sql.NullsPosition` member (old overloads marked `*REMOVED*`). Acceptable under the `Internal.*` namespace policy — noted only so the `modified` API entries are not mistaken for additive overloads.

## Baselines
Delta: [linq2db.baselines#1887](https://github.com/linq2db/linq2db.baselines/pull/1887) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5561) — ~1,787 added / 203 modified across 57 providers, overwhelmingly new NULLS-feature coverage.

- **Native providers** (PostgreSQL, Oracle, DB2, Informix, Firebird, SQLite, DuckDB, ClickHouse, SapHana) emit literal `NULLS FIRST`/`NULLS LAST`. OK.
- **Emulated providers** emit the leading sort key — SQL Server 2012+ `IIF(col IS NULL, ...)`, 2005/2008 `CASE WHEN`, Access `IIF`, MySQL boolean — both in the outer `ORDER BY` and threaded into `ROW_NUMBER() OVER (...)` rewrites (DISTINCT, DistinctBy, indexed-`Select`, UnionBy). OK.
- **`OrderByNullsNonNullableColumn`** — all 57 providers emit a plain `ORDER BY` (no key, no token): the top-level non-nullable no-op works. OK.
- **Suspect — `AggregationNonNullableNullsNoop`:** PostgreSQL (13–18) and SapHana.Odbc emit `STRING_AGG(... ORDER BY "Id" NULLS FIRST)` on a non-nullable key, diverging from Oracle / SQL Server (no token). See **MIN005**.
- **Note — `JoinWithGroupingVarious` (ClickHouse):** the `arraySort` null flag flips on both the nullable and the non-nullable path; the non-nullable flip is redundant (harmless).
- **Expected:** 203 `Priority`-column fixture additions across the DistinctBy/UnionBy families; one ClickHouse.MySql duplicate-block cleanup.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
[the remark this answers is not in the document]
sdanyliv: Fixed in a1e2754c86. `ExtractOrderByPart`'s `Queryable`/`Enumerable` branch now requires `mc.Arguments.Count == 2`, so a 3-arg BCL comparer overload is no longer extracted as a plain 2-arg ordering. It falls through to `OrderByBuilder` (which rejects comparer overloads — no SQL form) instead of silently dropping the comparer into a ROW_NUMBER rewrite. Regression test: `DistinctByMethodTests.DistinctByWithComparerOrderShouldFail`.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
sdanyliv: Fixed in a1e2754c86. `ExtractOrderByPart`'s `Queryable`/`Enumerable` branch now requires `mc.Arguments.Count == 2`, so a 3-arg BCL comparer overload is no longer extracted as a plain 2-arg ordering. It falls through to `OrderByBuilder` (which rejects comparer overloads — no SQL form) instead of silently dropping the comparer into a ROW_NUMBER rewrite. Regression test: `DistinctByMethodTests.DistinctByWithComparerOrderShouldFail`.

Source/LinqToDB/SqlOptions.cs
sdanyliv: Fixed in 9655b27352. Reworded the `DefaultNullsPosition` XML doc to scope the default to keys that don't specify a position, and to state that an explicit per-call position always takes precedence — including an explicit `Sql.NullsPosition.None`, which opts the key out of the default and uses the provider's natural null ordering.

Source/LinqToDB/Linq/Translation/AggregateFunctionBuilder.cs
sdanyliv: Investigated — the gate isn't the cause. `AggregateFunctionBuilder`'s `CanBeNullable(orderNullability)` correctly resolves `Id` as non-nullable: the MySQL and DB2 baselines for this test are clean (`ORDER BY Id`, no token), so `nulls` is `None` on every provider.

The `NULLS FIRST` on PostgreSQL/SapHana/DuckDB comes from those translators rendering `nulls == None` as `FIRST` (`... is First or None ? "FIRST" : "LAST"`, emitted whenever `!IsNullFiltered`). That `None -> FIRST` mapping is intentional and load-bearing: those providers' natural ASC null ordering is NULLS LAST, and pinning nulls-first keeps aggregate ordering consistent with .NET in-memory semantics — `FinalAggregationOrderedTest` orders by a nullable `Value1` (with a null row) against in-memory `OrderBy` (nulls first); dropping the token would flip them to NULLS LAST and fail it.

So on a non-nullable key the token is redundant but correctness-neutral (no NULLs to position), and removing it for `None` would regress the nullable default. Accepting as-is; clarified the test comment in c0a3a9e884 to note the result-level (not SQL-level) no-op on these providers.

Source/LinqToDB/Internal/SqlQuery/Visitors/SqlNullsOrderingLoweringVisitor.cs
sdanyliv: Implemented across `a994e35063` (a `NullsDefaultOrdering` provider capability + `QueryHelper.MatchesNaturalNullsPosition`; the skip is wired into regular `ORDER BY`, window `OVER`/`WITHIN GROUP`, and the `BuildOrderBy`-routed aggregates), `0a4145e259` (natural-ordering values for the remaining providers), and `b9eae2a3e1` (PG/SAP HANA/DuckDB aggregate `STRING_AGG` consolidation + skip).

A requested `Sql.NullsPosition` that already equals the provider's natural placement for the item's direction is now elided — both the emulation key (`CASE`/`IIF`/boolean) and the native `NULLS` token. The capability distinguishes direction-dependent providers (`Smallest`: SQL Server/MySQL/SQLite/SAP HANA/Firebird/Informix/Sybase/Access/SQL CE; `Largest`: PostgreSQL/Oracle/DB2) from direction-independent ones (`AlwaysLast`: ClickHouse/DuckDB). Values were verified against live containers for all 13 testable providers (Access/SQL CE confirmed via vendor docs).

Source/LinqToDB/Common/Configuration.cs
sdanyliv: Added `DataOptionsTests.ConfigurationSqlDefaultNullsPositionTest` in `8876441c6f`: it sets the process-global `Configuration.Sql.DefaultNullsPosition`, asserts both the getter and its propagation to a freshly-built `DataOptions.SqlOptions.DefaultNullsPosition`, and restores the prior value in a `finally`. Marked `[NonParallelizable]` since it mutates a global.

Source/LinqToDB/DataOptionsExtensions.cs
sdanyliv: Added a direct test in `8876441c6f`: `DataOptionsTests.WithDefaultNullsPositionTest` asserts `new SqlOptions().WithDefaultNullsPosition(Last).DefaultNullsPosition` and the `UseDefaultNullsPosition(DataOptions, …)` flow through to `SqlOptions.DefaultNullsPosition` — no longer only transitively covered via the remote round-trip.

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
```diff
@@ -60,51 +67,78 @@ public static Expression BuildWindowDefinition(Expression[] partitionBy, (Expres
...
+							supported = false;
+							break;
+					}
+				}
+				// linq2db OrderBy/ThenBy overloads that carry an explicit Sql.NullsPosition.
+				else if (typeof(LinqExtensions) == mc.Method.DeclaringType
+					&& mc.Arguments.Count == 3
+					&& mc.Method.GetParameters()[2].ParameterType == typeof(Sql.NullsPosition))
+				{
+					var nulls = (Sql.NullsPosition)((ConstantExpression)mc.Arguments[2]).Value!;
```
MaceWindu: **Minor · MIN007** — `ExtractOrderByPart` casts the NULLS-position argument straight to `ConstantExpression`. This path runs on the user's own query expression (`DistinctBy` / `MinBy`-`MaxBy` / `UnionBy` all route their preceding `OrderBy` through here), so a caller passing a non-literal position — e.g. `query.OrderBy(x => x.Value, somePositionVariable).DistinctBy(...)`, where the argument is a captured-variable `MemberExpression` rather than a `ConstantExpression` — throws `InvalidCastException` at build time. `OrderByBuilder` resolves the identical argument via `EvaluateExpression`, so the two NULLS entry points disagree on which argument shapes they accept.

Fix: guard the cast and fall through to the unsupported path (which routes to `OrderByBuilder`'s evaluator), or evaluate the argument here the way `OrderByBuilder` does. The guard form:

```suggestion
					if (mc.Arguments[2] is not ConstantExpression nullsConst)
					{
						supported = false;
						break;
					}

					var nulls = (Sql.NullsPosition)nullsConst.Value!;
```

Source/LinqToDB/Internal/DataProvider/Ydb/YdbSqlBuilder.cs
```diff
@@ -540,6 +540,14 @@ protected override void BuildOrderByClause(SelectQuery selectQuery)
 				if (item.IsDescending)
 					StringBuilder.Append(" DESC");
 
+				// NULLS positioning is lowered to a CASE key in the AST for YDB (no native support), so any
+				// position remaining here would only be present on a native provider.
+				if (item.NullsPosition != Sql.NullsPosition.None)
+				{
+					StringBuilder.Append(" NULLS ");
+					StringBuilder.Append(item.NullsPosition == Sql.NullsPosition.First ? "FIRST" : "LAST");
+				}
```
MaceWindu, line 549: **Minor · MIN008** — Dead `NULLS`-token block. YDB doesn't set `IsNullsOrderingSupported`, so `SqlNullsOrderingLoweringVisitor` always rewrites a requested position into a `CASE` key before SQL build — `item.NullsPosition` is always `None` by the time it reaches here. If it were ever reached it would emit `NULLS FIRST`/`NULLS LAST`, which YDB can't parse, producing an invalid statement rather than a clean fallback. The emulation already happens in the AST, so this block is unreachable. (Pairs with MIN009 — set YDB's `DefaultNullsOrdering`.) There is also no YDB test covering the emulated NULLS path.

Fix: delete the dead block (suggestion). Add a YDB `OrderBy(..., Sql.NullsPosition.First/Last)` test over a nullable key.

```suggestion
```

MaceWindu (commented):
_Automated review (Claude). Findings are suggestions for the author/maintainer to weigh, not blocking gates. Line-anchored findings are inline on the diff; broader notes are below._

## Review notes

Verified against the actual baseline `.sql` files and PR HEAD. AST propagation of the new `SqlOrderByItem.NullsPosition` is complete — threaded through clone/`Modify`, `QueryElementVisitor`, `Equals`/`GetElementHashCode`, `ToString`, `LinqServiceSerializer`, and both `SqlWindowOrderItem` conversion sites. Redundant-ordering elision is confirmed working from the baselines (natural-match positions emit no key/token; non-redundant cases do). Per-provider `DefaultNullsOrdering` flags match documented DB defaults. The three default surfaces (per-call overload, `DataOptions.UseDefaultNullsPosition`, `Configuration.Sql.DefaultNullsPosition`) are wired to one implementation with documented precedence and each tested; the `IComparer` overloads are rejected with a translation error. All prior review threads are addressed at HEAD.

Two line-level findings (inline) plus one below.

## Minor

- [ ] **MIN009** — YDB leaves `SqlProviderFlags.DefaultNullsOrdering` unset (defaults to `Unknown`). `YdbDataProvider` is the one emulated provider this PR doesn't give a `DefaultNullsOrdering`, so `QueryHelper.MatchesNaturalNullsPosition` always returns `false` for YDB and `SqlNullsOrderingLoweringVisitor` emits a `CASE` sort key even when the requested position already matches YDB's natural `NULL` placement — defeating index-backed ordering for that case, unlike every other emulated provider.

  Fix: set `SqlProviderFlags.DefaultNullsOrdering = NullsDefaultOrdering.Smallest;` in `YdbDataProvider`'s flag initialization — YDB sorts `NULL` as the smallest value (ascending => first, descending => last), matching SQL Server / MySQL / SQLite / Firebird / SapHana / Sybase / Access / SqlCe / Informix. Add a YDB NULLS-ordering test alongside it. Pairs with the dead-token cleanup inline (MIN008).

## Out-of-scope observations

- **SQLite assumes a runtime SQLite >= 3.30.0** — `SQLiteDataProvider` sets `IsNullsOrderingSupported = true` unconditionally. SQLite added `NULLS FIRST`/`NULLS LAST` in 3.30.0 (2019); a user supplying an older native sqlite3 library would get a syntax error rather than the `CASE` emulation. True for the bundled providers, and the flag can't be version-gated at builder-construction time, so this is acceptable — worth a provider-docs note.
- **`Sql.Ordinal` vs constant-index positioned ordering** — an explicit NULLS position is retained on the `Sql.Ordinal` `byIndex` path but reset to `None` on the constant-integer positioned path. Both emit valid SQL; a small asymmetry only.
- **3-level default precedence** (global -> `DataOptions` -> per-call) is covered pairwise / by composition (all paths funnel through `SqlOptions.DefaultNullsPosition`) rather than by one dedicated end-to-end test. Adequate.

## Baselines

Delta: [linq2db.baselines#1887](https://github.com/linq2db/linq2db.baselines/pull/1887) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5561)

The new-test baselines look correct — native `NULLS FIRST`/`NULLS LAST` tokens for PostgreSQL/Oracle/DB2/Firebird/SQLite/DuckDB/ClickHouse/SAP HANA, and `CASE`/`IIF` emulation keys for SQL Server/MySQL/MariaDB/Sybase/Access/SqlCe/Informix. Redundant-position elision is verified directly against the files: `OrderByNulls(SqlServer,First,ASC)` and `OrderByNulls(PostgreSQL,Last,ASC)` emit a plain `ORDER BY` (no key/token), while the non-redundant `(First,DESC)` case emits the `IIF` key. Two things to confirm before merging the baselines PR:

- **~30 unrelated modified baselines.** A **Sybase.Managed** cluster (DateTime/DateOnly/GroupBy/StringConcat/StringFunctions/Mapping null-guard simplifications) plus **Informix** (`Issue4717Test`, `Issue3226Test4`). These are not produced by this PR's code; they look like drift between the baselines-branch base and baselines master. Confirm they are already on linq2db master (expected drift) rather than a side effect of this branch.
- **ClickHouse `StringJoinTests.JoinWithGroupingVarious`.** The `arraySort` NULL-comparator key flips `if(isNull(t.1), 1, 0)` -> `if(isNull(t.1), 0, 1)`, changing the `NULL` placement of this pre-existing aggregation. This is an intentional, in-code-documented fix — `MakeNullsKey` now inverts the key when the whole array is reversed (single string-DESC key), correcting the old mis-placement. Confirm the new placement is the intended one.

Source/LinqToDB/Internal/DataProvider/SapHana/Translation/SapHanaMemberTranslator.cs
```diff
@@ -359,29 +359,7 @@ protected class StringMemberTranslator : StringMemberTranslatorBase
...
-										if (!info.IsNullFiltered)
-										{
-											sb.Value.Append(" NULLS ");
-											sb.Value.Append(info.OrderBySql[i].nulls is Sql.NullsPosition.First or Sql.NullsPosition.None ? "FIRST" : "LAST");
-										}
-									}
-
-									suffix = factory.Fragment(sb.Value.ToString(), args);
-								}
+								var suffix = BuildAggregateNullsOrderBy(factory, info.OrderBySql, info.IsNullFiltered, NullsDefaultOrdering.Smallest);
```
MaceWindu, line 362: code smell - we should read provider flags instead of hardcoding NullsDefaultOrdering values

Source/LinqToDB/Internal/Expressions/WindowFunctionHelpers.cs
sdanyliv: The position can't arrive as a non-constant here. All four `LinqExtensions` NULLS overloads build the call node with `Expression.Constant(nulls)` (e.g. `LinqExtensions.cs:741`), so by the time `ExtractOrderByPart` runs `mc.Arguments[2]` is always a `ConstantExpression` -- a caller's captured variable is already boxed into the constant at the API boundary. `DistinctByMethodTests.DistinctByOrderByNulls` exercises exactly that (a captured `nulls` parameter flowing through `OrderBy(.., nulls)` -> `DistinctBy`) and is green, so the cast never throws.

Switched it to `mc.Arguments[2].EvaluateExpression()` in e5ec2d08a2 anyway, to match how `OrderByBuilder` resolves the same argument and stay robust against a hand-built expression tree. Consistency change, not a bug fix.

Source/LinqToDB/Internal/DataProvider/Ydb/YdbSqlBuilder.cs
sdanyliv: Keeping the branch as an intentional placeholder. YDB's ORDER BY docs don't document `NULLS FIRST` / `NULLS LAST`, and I couldn't confirm YDB's natural null collation -- enabling native now (`IsNullsOrderingSupported = true` + a `DefaultNullsOrdering`) would risk silently mis-placing NULLs on the elision path if the assumed natural order is wrong. YDB stays emulated (positions lowered to a `CASE` key, same as the other non-native providers), so NULLS ordering is already correct for YDB through that path. The render branch stays in place so enabling it later is a one-line flag flip once native support + natural order are confirmed against a live YDB.

MaceWindu (dismissed):

Source/LinqToDB/Internal/DataProvider/Firebird/FirebirdDataProvider.cs
```diff
@@ -32,6 +33,8 @@ protected FirebirdDataProvider(string name, FirebirdVersion version)
 			SqlProviderFlags.IsIdentityParameterRequired       = true;
 			SqlProviderFlags.IsCommonTableExpressionsSupported = true;
 			SqlProviderFlags.IsSubQueryOrderBySupported        = true;
+			SqlProviderFlags.IsNullsOrderingSupported          = true;
+			SqlProviderFlags.DefaultNullsOrdering              = NullsDefaultOrdering.Smallest; // Firebird 2.0+ sorts NULL as the smallest value
 			SqlProviderFlags.IsUnionAllOrderBySupported        = true;
```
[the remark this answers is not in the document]
MaceWindu: Fixed in 8a3d379e4 - verified against the Firebird Null Guide: Firebird 2.0+ sorts NULL as the *smallest* value (ascending NULLS FIRST, descending NULLS LAST), so the `DefaultNullsOrdering.Smallest` flag set in `FirebirdDataProvider` is correct. The inconsistency was in the enum XML docs - Firebird was mistakenly listed under `Largest`; moved it to the `Smallest` example list. Ref: https://www.firebirdsql.org/manual/nullguide-sorts.html

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Added `NULLS FIRST` / `NULLS LAST` control to LINQ ordering. New `OrderBy` / `OrderByDescending` / `ThenBy` / `ThenByDescending` overloads accept a `Sql.NullsPosition` (`First` / `Last` / `None`), and a default placement can be set via `new DataOptions().UseDefaultNullsPosition(...)` or `Configuration.Sql.DefaultNullsPosition`. Rendered natively where supported (PostgreSQL, Oracle, DB2, Firebird, SQLite, DuckDB, ClickHouse, SAP HANA) and emulated automatically elsewhere (SQL Server, MySQL, MariaDB, Access, Sybase, etc.), with consistent results; it also flows through `DISTINCT` / set operations, `DistinctBy` / `UnionBy` / `MinBy` / `MaxBy`, indexed `Select`, and aggregate ordering. The BCL `OrderBy` / `ThenBy(keySelector, IComparer<TKey>)` overloads - which have no SQL equivalent - are now rejected with a translation error instead of being silently ignored.


Added `NULLS FIRST` / `NULLS LAST` control to `OrderBy` / `ThenBy`, plus a `DataOptions` default placement, rendered natively or emulated across all providers.


---
<sub>Generated from commit `d73b7bc`.</sub>

PR #5567, opened 2026-05-31
Fix #5529: restore IS NULL pushdown through SqlConcatExpression

### LINKED ISSUE #5529: WhereTests.LengthTest1 — SqlServer no longer simplifies LEN(...) - 1 <> 0 or (col + literal) IS NULL
### Summary

`WhereTests.LengthTest1` no longer simplifies `LEN(... + '.') - 1 <> 0` to `LEN(... + '.') <> 1` on SQL Server, and no longer simplifies `(col + literal) IS NULL` to `col IS NULL`. Both simplifications fired on master; PR #5504's baselines lose them.

### Current behavior

Test source: `Tests/Linq/Linq/WhereTests.cs:1515-1525` (`!(nm?.Length == 0)`).

**SqlServer.2022.MS, master:**

```sql
WHERE LEN([nm].[MiddleName] + N'.') <> 1 OR [nm].[MiddleName] IS NULL
```

**SqlServer.2022.MS, PR #5504 baselines branch:**

```sql
WHERE LEN([nm].[MiddleName] + N'.') - 1 <> 0 OR ([nm].[MiddleName] + N'.') IS NULL
```

Two regressions in one query:

1. **Subtraction-against-zero not folded.** The optimizer no longer rewrites `expr - n <> 0` to `expr <> n`. `LEN(... + '.')` is the SqlServer-specific shape used because `LEN` strips trailing spaces from CHAR/NCHAR-padded strings; the trailing `.` is added so the count is faithful, then `-1` removes the dot.
2. **`IS NULL` over a concat not pushed down.** `(col + literal) IS NULL` is equivalent to `col IS NULL` (string `+` is null-propagating on SQL Server). The optimizer no longer pushes the `IS NULL` through the concat to the base column.

Scope: SqlServer-specific — all 20 SqlServer variants in `Tests.Linq.WhereTests.LengthTest1` baselines diff identically. PostgreSQL / SQLite / SAP HANA / DB2 / Access emit `Length(col) <> 0 OR col IS NULL` and are unaffected because their `Length`-equivalent doesn't need the `+ '.'` trick.

### Expected behavior

Restore both optimizations in the SQL expression optimizer (`Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs` is the likely site):

1. `binaryExpr <op> 0` where one side is `x - const` → `x <op> const`. Same for `<>`, `=`, `<`, `>`, `<=`, `>=`.
2. `(col + literal) IS NULL` → `col IS NULL` when the concat operand is a non-null literal and the concat is null-propagating (SqlServer `+`, not pipe-concat providers where `||` may be null-safe).

The first is a general algebraic rewrite. The second needs care around provider semantics — e.g. on PostgreSQL `'a' || NULL` returns NULL too, but on Oracle and DB2 `||` treats NULL as empty string, so pushing `IS NULL` through `||` would be wrong there. Gate on provider null-concat behavior, or only apply when the operator is SqlServer `+`.

### Test coverage

- Per-provider generated-SQL tests for both simplifications, on SqlServer (and any other provider that uses the `LEN(... + literal) - n` shape).
- Execution tests asserting the result matches before/after the simplification — important because pushing `IS NULL` through a concat is sound only on null-propagating concat dialects.
- Specific regression coverage for `WhereTests.LengthTest1` so the unsimplified shape can't reappear silently.

### Notes

- Originated from baselines review of PR #5504 — see https://github.com/linq2db/linq2db/pull/5504#issuecomment-4451221303.
- Master baseline last touched 2024-09 via PR #5426 (`96c108b020` in linq2db.baselines), so the simplification was working at least through that point.
- Sibling follow-up tasks from the same review: #5527 (provider precedence), #5528 (nested `LCase`/`Lower`).

MaceWindu:
## Problem

#5504 replaced the string-concat AST node (`SqlBinaryExpression('+')` → `SqlConcatExpression`). `ReduceIsNullExpressionVisitor` only recurses into a type whitelist that excluded `SqlConcat`, so two SqlServer simplifications regressed (surfaced in #5504's baselines review):

1. `(col + literal) IS NULL` no longer pushed down to `col IS NULL`.
2. `LEN(... + '.') - 1 <> 0` no longer folded to `LEN(... + '.') <> 1`.

Regression 2 is **downstream** of regression 1: once the `IS NULL` over the concat reduces to the base column, the comparison's nullability factors out and the existing subtract fold re-fires — so only the reducer needed fixing.

## Fix

`ReduceIsNullExpressionVisitor`: add `SqlConcat` to the visit whitelist + a `VisitSqlConcatExpression` override that reduces each operand's `IS NULL` (mirroring the `+` case), gated on `PreserveNull`. Null-as-empty `||` providers (Oracle/DB2) build `PreserveNull:false` concats and are correctly left untouched.

## Verification

Baselines regenerated regressed → fixed against live DBs:

| Test / provider | After fix |
|---|---|
| `WhereTests.LengthTest1` SqlServer 2022 (`+`) | `LEN([nm].[MiddleName] + N'.') <> 1 OR [nm].[MiddleName] IS NULL` |
| `WhereTests.LengthTest1` SqlServer 2025 (`\|\|`) | `LEN([nm].[MiddleName] \|\| N'.') <> 1 OR [nm].[MiddleName] IS NULL` |
| `StringFunctionsTests.ConcatStringsTest` Access (`+`) | `IIF([t].[Value3] IS NULL, '', ' -> ' + [t].[Value3])` |

Change is global (any null-propagating-concat-under-`IS NULL` query); full cross-provider baselines regenerate via CI.

Fixes #5529.

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/ReduceIsNullExpressionVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/ReduceIsNullExpressionVisitor.cs
@@ -32,10 +32,11 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			return element?.ElementType switch
 			{
 				QueryElementType.SqlNullabilityExpression or
-				QueryElementType.SqlBinaryExpression or 
-				QueryElementType.SqlCondition or 
-				QueryElementType.SqlCast or 
-				QueryElementType.SqlFunction or 
+				QueryElementType.SqlBinaryExpression or
+				QueryElementType.SqlConcat or
+				QueryElementType.SqlCondition or
+				QueryElementType.SqlCast or
+				QueryElementType.SqlFunction or
 				QueryElementType.SqlExpression =>
 					base.Visit(element),
 
@@ -183,6 +184,22 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			{
 				ReduceOrAdd(element.Expr1);
 				ReduceOrAdd(element.Expr2);
+			}
+
+			return element;
+		}
+
+		protected internal override IQueryElement VisitSqlConcatExpression(SqlConcatExpression element)
+		{
+			// Null-propagating concat (e.g. SqlServer / Access `+`): the result is NULL iff any operand
+			// is NULL, so `concat IS NULL` reduces to the OR of each operand's IS NULL — mirroring the
+			// `+` case in VisitSqlBinaryExpression. Strict-null=false concats (`||` treating NULL as the
+			// empty string on Oracle / DB2) never yield NULL from a NULL operand, so they are not
+			// nullable and never reach this reducer (folded in VisitIsNullPredicate).
+			if (element.PreserveNull)
+			{
+				foreach (var expr in element.Expressions)
+					ReduceOrAdd(expr);
 			}
 
 			return element;


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Restored `IS NULL` simplification over string concatenation: a `WHERE (column + value) IS NULL` filter again generates the simpler `column IS NULL` on null-propagating-concat providers (e.g. SQL Server, Access), instead of testing the whole concatenation for null.


_(none)_


---
<sub>Generated from commit `9b299a5`.</sub>

PR #5568, opened 2026-05-31
Fix #5531: Informix SUM null-guard emits Nvl(x, NULL) instead of the intended default

### LINKED ISSUE #5531: Informix: SUM null-guard emits Nvl(..., NULL) instead of Nvl(..., 0)
### Summary

Follow-up to PR #5502. On Informix, the new SUM null-guard wrapping emits `Nvl(x, NULL)` or `Nvl(x, Nvl(NULL, 0))` instead of the intended `Nvl(x, 0)`. The SQL diverges from every other provider and, in one of the two cases, from what master already produced.

### Background

PR #5502 ("Fix #5404: wrap non-nullable subquery aggregates with COALESCE") wraps correlated scalar `(SELECT SUM(...))` subqueries with a null guard so the result coerces NULL to 0. All providers receive `Coalesce((SELECT SUM(...)), 0)` (standard), `Nvl(..., 0)` (Oracle), `Nvl(..., 0)` (Informix — intended), or `IIF(... IS NULL, 0, ...)` (Access). Informix is the only provider where two tests get the wrong default literal.

### Current behavior (PR #5502 HEAD)

1. **`Tests.Linq.CteTests.Issue4717Test`** — emits `Nvl(..., NULL)`.

   Baseline `Informix.DB2/Tests/Linq/CteTests/Tests.Linq.CteTests.Issue4717Test(Informix.DB2).sql`:
   ```sql
   Nvl((
       SELECT SUM(wp_1.StockOnHand)
       FROM Issue4717WarehouseProductMapping wp_1
       WHERE wp_1.WarehouseId = source.WarehouseId
   ), NULL)
   ```

   `Nvl(x, NULL)` is a no-op — the projected column is NULL when the inner SUM has no rows, defeating the null-guard intent.

2. **`Tests.Linq.EagerLoadingTests.Issue3226Test4`** — `master` emitted `Nvl(x_1.Sum_1, 0)`; PR #5502 changes it to `Nvl(x_1.Sum_1, Nvl(NULL, 0))`.

   Baseline `Informix.DB2/Tests/Linq/EagerLoadingTests/Tests.Linq.EagerLoadingTests.Issue3226Test4(Informix.DB2).sql`:
   ```sql
   ORDER BY
       Nvl(x_1.Sum_1, Nvl(NULL, 0))
   ```

Both shapes are consistent with a `Coalesce(x, NULL, 0)` expression being constructed upstream and then folded to nested `Nvl` by `Source/LinqToDB/Internal/DataProvider/Informix/InformixSqlExpressionConvertVisitor.cs:64-71` (`ConvertCoalesceToBinaryFunc(element, "Nvl", supportsParameters: false)`). The fold is correct; the upstream Coalesce being built with `NULL` as one of its arguments is the bug.

### Expected behavior

Informix should emit `Nvl((SELECT SUM(...)), 0)` for both tests, matching the Coalesce shape every other provider produces.

### Notes

- Discovered while reviewing baselines for PR #5502 (linq2db/linq2db.baselines#1852).
- `Issue3226Test4` is a generated-SQL cleanliness regression — `Nvl(NULL, 0)` folds to `0` at runtime.
- `Issue4717Test` is potentially a wrong-result regression: `Nvl(x, NULL)` returns NULL when the inner SUM matches no rows. Actual impact at runtime depends on whether the test data exercises an empty inner subquery.
- Likely caused by the new null-guard wrapping logic in PR #5502 constructing a Coalesce with `NULL` as one of its arguments on the Informix code path; trace upstream from the Informix Coalesce conversion.

MaceWindu:
## Summary

Follow-up to #5502. Fixes #5531 — on Informix the new non-nullable-`SUM` null-guard emitted `Nvl(x, NULL)` / `Nvl(x, Nvl(NULL, 0))` instead of the intended default.

## Root cause

PR #5502's null-guard (`AggregateFunctionsMemberTranslatorBase`) gated on the aggregate **argument**'s nullability, not the method's **result** type. A `(decimal?)`-cast `.Sum()` over a non-nullable column has a non-nullable argument but a nullable result, so the guard fired — and `DefaultValue.GetValue(decimal?)` returns `null`, so it built `Coalesce(<ref>, NULL)`, a no-op default.

Every provider builds this. All strip the NULL operand in `ConvertCoalesce` **except** Informix (folds straight to `Nvl`) and Access (folds to `IIF`), which leaked the NULL into generated SQL.

> The issue's "expected" for the correlated-subquery case (`Issue4717Test`) assumed `Nvl(SUM, 0)`, but every other provider (DB2 / SQLite / Oracle / SqlServer) actually emits **bare `SUM`** there — the no-op guard vanishes entirely. This fix makes Informix consistent: bare `SUM` for that case, `Nvl(x, 0)` for the lifted-`ORDER BY` case (`Issue3226Test4`).

## Fix

- **A** — `AggregateFunctionsMemberTranslatorBase`: the SUM null-guard now also requires a non-nullable `resultType`, so a nullable `.Sum()` no longer builds the no-op wrap (matches #5502's "non-nullable subquery aggregates" intent).
- **B** — extract `RemoveNullValues` in the base `SqlExpressionConvertVisitor` (base `ConvertCoalesce` refactored to use it — behavior-preserving) and call it from the Informix / Access `ConvertCoalesce` overrides **before** they fold to `Nvl` / `IIF`, so they apply the same NULL-stripping normalization the base does. Closes the structural gap for any future NULL-bearing coalesce, not just #5531's source.

## Generated SQL (verified in-process)

`Issue4717Test` (correlated subquery), Informix:
```diff
- Nvl((SELECT SUM(wp_1.StockOnHand) ...), NULL)
+ (SELECT SUM(wp_1.StockOnHand) ...)            -- bare SUM, matches DB2/SQLite/Oracle
```

`Issue3226Test4` (lifted `ORDER BY`), Informix:
```diff
- ORDER BY Nvl(x_1.Sum_1, Nvl(NULL, 0))
+ ORDER BY Nvl(x_1.Sum_1, 0)
```

SQLite unchanged (bare `SUM` / `Coalesce(SUM, 0)`).

## Test plan

- [x] Core builds clean — Testing + Release (net10.0); analyzers, banned-API, PublicAPI all pass.
- [x] In-process SQL-shape check on Informix + SQLite confirms the shapes above; SQLite output is byte-identical to current baselines (proves the `RemoveNullValues` base refactor is behavior-preserving).
- [ ] CI full provider matrix — regenerates the `Informix.DB2` baselines for `Issue4717Test` / `Issue3226Test4` (which currently show `Nvl(..., NULL)`); the resulting baselines PR should be reviewed.

Fixes #5531

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/Access/AccessSqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Access/AccessSqlExpressionConvertVisitor.cs
@@ -134,6 +134,15 @@ namespace LinqToDB.Internal.DataProvider.Access
 			if (element.SystemType == null)
 				return element;
 
+			// Strip NULL-literal operands before folding to IIF, matching base ConvertCoalesce —
+			// otherwise a no-op guard like Coalesce(x, NULL) folds to IIF(x IS NULL, NULL, x)
+			// (issue #5531).
+			var reduced = RemoveNullValues(element);
+			if (reduced is not SqlCoalesceExpression coalesce)
+				return reduced;
+
+			element = coalesce;
+
 			if (element.Expressions.Length == 2)
 			{
 				return new SqlConditionExpression(new SqlPredicate.IsNull(element.Expressions[0], false), element.Expressions[1], element.Expressions[0]);
--- a/Source/LinqToDB/Internal/DataProvider/Informix/InformixSqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Informix/InformixSqlExpressionConvertVisitor.cs
@@ -63,11 +63,16 @@ namespace LinqToDB.Internal.DataProvider.Informix
 
 		public override ISqlExpression ConvertCoalesce(SqlCoalesceExpression element)
 		{
-			return element.SystemType switch
-			{
-				null => element,
-				_ => ConvertCoalesceToBinaryFunc(element, "Nvl", supportsParameters: false),
-			};
+			if (element.SystemType == null)
+				return element;
+
+			// Strip NULL-literal operands before folding to Nvl, matching base ConvertCoalesce —
+			// otherwise a no-op guard like Coalesce(x, NULL) folds to Nvl(x, NULL) (issue #5531).
+			var reduced = RemoveNullValues(element);
+			if (reduced is not SqlCoalesceExpression coalesce)
+				return reduced;
+
+			return ConvertCoalesceToBinaryFunc(coalesce, "Nvl", supportsParameters: false);
 		}
 
 		//TODO: Move everything to SQLBuilder
--- a/Source/LinqToDB/Internal/DataProvider/Translation/AggregateFunctionsMemberTranslatorBase.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Translation/AggregateFunctionsMemberTranslatorBase.cs
@@ -329,7 +329,16 @@ namespace LinqToDB.Internal.DataProvider.Translation
 						// /optimization (so ClickHouse can still reject correlated subqueries) and
 						// DefaultIfEmpty's CASE-injection on the SUM argument is unaffected. The lifted
 						// reference is wrapped with COALESCE(<ref>, default).
-						if (!info.IsGroupBy && argumentValue.SystemType?.IsNullableOrReferenceType == false && functionName is "SUM")
+						// Gate on the method *result* type, not just the argument: a nullable-projected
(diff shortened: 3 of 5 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Informix: fixed a nullable-projected `Sum` (e.g. `Sum(x => (decimal?)x)`) over a non-nullable column where the generated null-guard was a no-op (`Nvl(x, NULL)`) that didn't apply the intended default; the aggregate now behaves as intended. The same no-op null-guard was also fixed in the Access `IIF` fold.


_(none)_


---
<sub>Generated from commit `7b50cf6`.</sub>

PR #5570, opened 2026-05-31
Fix #5528: collapse redundant nested case-conversion wraps in Guid→string translation

### LINKED ISSUE #5528: Nested LCase/Lower in GuidToString translation — collapse redundant case-conversion wraps
### Summary

`Guid → string` conversion combined with `.ToLower()` produces nested `LCase(LCase(...))` on Access and nested `Lower(...Lower(...))` (with an intermediate `CAST`) on Firebird. Both arise because the provider's cast-from-Guid translator already wraps the result in a lower-case call to normalise output, and a user-supplied `.ToLower()` on top adds a second redundant wrap.

### Current behavior

Test `Tests.Linq.ConvertTests.GuidToString` (`Tests/Linq/Linq/ConvertTests.cs:578-585`) issues `Sql.ConvertTo<string>.From(t.GuidValue).ToLower() == guid`. Under PR #5504's baselines branch the generated SQL is:

- **Access** (`Access.Ace.Odbc`, `Access.Ace.OleDb`, `Access.Jet.Odbc`, `Access.Jet.OleDb`):
  ```sql
  LCase(LCase(Mid(CStr([t].[GuidValue]), 2, 36))) = ?
  ```
- **Firebird** (`Firebird.2.5`, `Firebird.3`, `Firebird.4`, `Firebird.5`):
  ```sql
  Lower(CAST(Lower(UUID_TO_CHAR("t"."GuidValue")) AS VarChar(36) CHARACTER SET UNICODE_FSS)) = @guid
  ```

The inner `LCase` / `Lower` comes from the cast-from-Guid translator:

- `Source/LinqToDB/Internal/DataProvider/Access/Translation/AccessMemberTranslator.cs:393-402` — `TranslateGuildToString` wraps `Mid(CStr(...), 2, 36)` in `LCase(...)`.
- `Source/LinqToDB/Internal/DataProvider/Firebird/Translation/FirebirdMemberTranslator.cs:355-365` — `TranslateGuildToString` wraps `UUID_TO_CHAR(...)` in `Lower(...)`.
- `Source/LinqToDB/Internal/DataProvider/Firebird/FirebirdSqlExpressionConvertVisitor.cs:164-168` — same lower-case wrap for the `cast(guid as string)` path; the file already carries `// TODO: think how to use FirebirdMemberTranslator.GuidMemberTranslator.TranslateGuildToString instead of code duplication here`.

The outer `LCase` / `Lower` comes from the user's `.ToLower()`. No optimization step collapses the pair, so the redundant wrap survives into the emitted SQL.

Note: master baselines for `GuidToString` show only one `LCase` / `Lower` on these providers, so the duplication is exposed (or introduced) by PR #5504 and worth treating as a regression-risk surface even though the result is still correct.

### Expected behavior

Either of:

1. **Idempotent collapse at the optimizer.** Add a rule that rewrites `Lower(Lower(x))` → `Lower(x)` (and `Upper(Upper(x))` → `Upper(x)`, and the provider-specific `LCase`/`UCase` equivalents). This catches the general case and benefits other tests, not just `GuidToString`.
2. **Drop the inner wrap when the outer is present.** Have the cast-from-Guid translator emit the un-cased shape (`Mid(CStr(...), 2, 36)` on Access, `UUID_TO_CHAR(...)` on Firebird) when the surrounding context already forces lower-case. This is provider-specific and brittle.

Option 1 is the right scope — it's a single optimizer pass and doesn't require coordination between translation phases.

While here, fix the duplication called out by the existing TODO at `FirebirdSqlExpressionConvertVisitor.cs:166` so the cast path and the `.ToString()` path share one translator.

### Test coverage

Per-provider tests asserting:

- **Generated SQL** for `Lower(Lower(x))` / `Upper(Upper(x))` collapses to a single call (baseline-driven, both case-conversion directions).
- **Execution result** matches between the collapsed and non-collapsed form against a real provider, to catch any dialect where `Lower(x)` is not idempotent in practice (collations, locale-sensitive characters).
- Specific regression coverage for `ConvertTests.GuidToString` on Access and Firebird so the nested-call shape can't reappear silently in a future baseline update.

### Notes

- Originated from baselines review of PR #5504 — see https://github.com/linq2db/linq2db/pull/5504#issuecomment-4451221303.
- No duplicate issues found via `gh search issues "nested Lower LCase" | "double ToLower duplicate" | "GuidToString Access"`.
- Related to but separate from #5527 (provider-specific SQL precedence) — both surface generated-SQL noise, but the optimizer pass needed here is independent of precedence.

MaceWindu:
## Summary

The SQL optimizer collapsed adjacent `TO_LOWER(TO_LOWER(x))`, but not when a provider's
Guid→string translator inserts a case-preserving layer between the two calls. #5504 added
such layers — a `CAST` on Firebird, a nullability annotation on Access — so a user
`.ToLower()` over an already-lower-cased Guid→string survived as a redundant outer wrap:

- Access: `LCase(LCase(Mid(CStr([t].[GuidValue]), 2, 36)))`
- Firebird: `Lower(CAST(Lower(UUID_TO_CHAR("t"."GuidValue")) AS VarChar(36) CHARACTER SET UNICODE_FSS))`

## Changes

- `SqlExpressionOptimizerVisitor.OptimizeFunction`: new `IsSameCaseConversion` peels
  case-preserving wrappers (string casts, nullability annotations) and drops the redundant
  outer `TO_LOWER`/`TO_UPPER`. Works on the pseudo-functions, so every provider benefits
  regardless of final `Lower`/`LCase` naming.
- `FirebirdMemberTranslator.TranslateGuidToString` extracted and shared by the
  `Guid.ToString()` member translator and the `cast(guid as string)` path in
  `FirebirdSqlExpressionConvertVisitor` — resolves the existing TODO; both now emit the
  identical `CAST(Lower(UUID_TO_CHAR(...)) AS VARCHAR(36))`.
- `GuidTests.GuidToStringNoRedundantCaseWrap`: inline `LastQuery` guard against the nested
  shape, independent of baselines.

## Verification

- Executed against real Firebird 5.0 and offline Access — both collapse to a single case call.
- `GuidTests.*` + `ConvertTests.GuidToString` pass on SQLite, Firebird.5, SqlServer.2022, YDB, DuckDB.

Baselines for `{ConvertTests,GuidTests}.GuidToString` × {Access×4, Firebird×4} will change
nested → single (needs `/azp run`).

Fixes #5528


Source/LinqToDB/Internal/DataProvider/Firebird/Translation/FirebirdMemberTranslator.cs
```diff
@@ -370,32 +371,39 @@ [new SqlFunctionArgument(valuesExpr, suffix: factory.Fragment("FROM {0}", factor
...
-				// SqlCe / SapHana Guid translators which all converge on a VARCHAR(36) result.
+			var stringDataType = mappingSchema.GetDbDataType(typeof(string));
+			var charType       = stringDataType.WithDataType(DataType.Char).WithLength(36);
+			var varCharType    = stringDataType.WithDataType(DataType.VarChar).WithLength(36);
 
-				var factory        = translationContext.ExpressionFactory;
-				var stringDataType = factory.GetDbDataType(typeof(string));
-				var charType       = stringDataType.WithDataType(DataType.Char).WithLength(36);
-				var varCharType    = stringDataType.WithDataType(DataType.VarChar).WithLength(36);
+			var toChar  = new SqlFunction(charType, "UUID_TO_CHAR", guidExpr);
```
sdanyliv: Use Factory

sdanyliv (changes requested):

Source/LinqToDB/Internal/DataProvider/Firebird/Translation/FirebirdMemberTranslator.cs
MaceWindu: Done in 2dcbb6e40 — `TranslateGuidToString` now takes an `ISqlExpressionFactory` and builds via `factory.Function`/`factory.ToLower`/`factory.Cast`. Both callers (member translator, convert visitor) pass their existing factory.

Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs
```diff
@@ -1391,15 +1391,52 @@ protected virtual IQueryElement OptimizeFunction(SqlFunction function)
...
 			}
 
 			return function;
 		}
 
+		// Returns true when <paramref name="expr"/> is already the result of the case conversion named by
+		// <paramref name="caseFunctionName"/> (TO_LOWER / TO_UPPER), looking through case-preserving wrappers:
+		// nullability annotations and string casts (casting an already-cased string to another string type
+		// preserves the case, so the surrounding conversion is idempotent).
+		static bool IsSameCaseConversion(ISqlExpression expr, string caseFunctionName)
```
sdanyliv: It is better to move such function to QueryHelper and split to `IsUpperString` and `IsLowerString` or something like that.

Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs
```diff
@@ -1391,15 +1391,52 @@ protected virtual IQueryElement OptimizeFunction(SqlFunction function)
 		{
 			if (function.Parameters.Length == 1 && function.Name is PseudoFunctions.TO_LOWER or PseudoFunctions.TO_UPPER)
 			{
-				if (function.Parameters[0] is SqlFunction { Parameters.Length: 1, Name: PseudoFunctions.TO_LOWER or PseudoFunctions.TO_UPPER } func)
+				var argument = function.Parameters[0];
+
+				// Adjacent nested case conversion (Lower(Lower(x)), Lower(Upper(x)), …): the outer call wins.
+				if (argument is SqlFunction { Parameters.Length: 1, Name: PseudoFunctions.TO_LOWER or PseudoFunctions.TO_UPPER } func)
 				{
 					return new SqlFunction(function.Type, function.Name, func.Parameters[0]);
```
sdanyliv: SqlFunction has `WithXXX` methods to simplify such modifications.

sdanyliv (changes requested):

Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs
MaceWindu: Done in 12f4257bb — moved to `QueryHelper.IsLowerString`/`IsUpperString` (both delegating to a private `IsCaseConversion` core).

Source/LinqToDB/Internal/SqlProvider/SqlExpressionOptimizerVisitor.cs
MaceWindu: Done in 12f4257bb — now `func.WithName(function.Name)` instead of constructing a new `SqlFunction`.

sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed redundant nested case-conversion in generated SQL when calling `.ToLower()` or `.ToUpper()` on a Guid converted to string. The provider's own Guid-to-string conversion already lower-cases its result, so a user-supplied `.ToLower()` on top produced a doubled wrap (for example `LCase(LCase(...))` on Access or `Lower(Cast(Lower(...)))` on Firebird); the optimizer now collapses these to a single case call.


_(none)_


---
<sub>Generated from commit `12a5b20`.</sub>

PR #5577, opened 2026-06-02
Builder: expand member/method mappings during the initial expose

MaceWindu:
## Problem 

 `MapMember` / `ConvertMember` mappings were only expanded when the expose pass ran with `includeConvert = true`, which was set solely for the build-time per-expression expose (`ConvertExpression`). A mapped method therefore survived unexpanded through the **initial** query expose and was expanded too late — during build, after the projection context had already routed the inner association to an **eager-load**.  For a mapped correlated aggregate — e.g. `Expressions.MapMember<Parent,int>(p => Count1(p), p => p.Children.Count(...))` projected as `Select(p => Count1(p))` — this produced a client-side count over an eager-loaded collection that left an unresolved `ContextRefExpression` in the result mapper (`System.ArgumentException : must be reducible node`), instead of the SQL scalar subquery that a directly-written `p.Children.Count()` yields.  ## Fix  Remove the `includeConvert` flag entirely so expose always applies member / method / binary / constructor mappings. The mapped expression now flows through the normal build as an aggregate, matching a directly-written association aggregate: a correlated subquery the provider can reject cleanly (`Error_Correlated_Subqueries`) or run where supported.  ## Impact  Provider-agnostic change to expose timing — mappings now expand in the initial pass for every query. Surfaced via the experimental YDB provider, where `MapMember`-mapped correlated aggregates misfired into `must be reducible node`.  🤖 Generated with [Claude Code](https://claude.com/claude-code)  

 ## Follow-up commit

Always-expanding member/method mappings during the initial expose surfaced a
latent defect in the `string.Compare` / `CompareTo` mapping. The mapping routed
through `Expressions.ConvertToCaseCompareTo`, whose managed body returns `null`
when either operand is null. That body is only the **local-evaluation** fallback —
the `[Sql.Extension]` builder emits `SqlCompareToExpression` for SQL, which lowers
to a `CompareNulls.LikeClr` predicate that already matches BCL semantics. While the
mapping expanded only for SQL, the null-returning body never executed.

With the mapping now expanded in the initial expose, the test harness `AssertQuery`
began evaluating it in-memory, where `ConvertToCaseCompareTo(null, x)!.Value`
collapses to `0` via null-propagation — flipping comparisons (`0 <= compare(...)`)
and breaking `WhereTests.Issue2424*` on every provider, while the actual SQL stayed
correct.

Fix: added a private `ConvertToCaseCompareToImpl` with a faithful body
(`StringComparer.Ordinal.Compare` — null sorts first, always `-1/0/1`) and repointed
the member mappings at it. SQL generation is unchanged (same `[Sql.Extension]`
builder); the obsolete public `ConvertToCaseCompareTo` keeps its frozen behavior.

## Follow-up commit (2)

Always-expanding binary/member/method mappings in the initial expose also moved
`ConvertBinary`'s `ChangeTypeExpression` wrapper ahead of the `ParametersContext` /
`GetExpressionAccessors` parameter-accessor pass. `PathVisitor` walks that tree but
handles linq2db custom nodes only via the `ExpressionType.Extension` branch;
`ChangeTypeExpression` is the lone node using a bespoke numeric `NodeType`
(`(ExpressionType)1000`, non-reducible), so it fell through to
`default: throw NotSupportedException("Unhandled expression type: 1000")`.

This broke `ExpressionsTests.MapOperator` (`int >> int` mapped to `Shr(long,int)`,
whose `long` result is wrapped `(int)(…)`) on every provider — the throw fires before
any SQL is generated.

Fix: add a `ChangeTypeExpression` case to `PathVisitor` that recurses into the inner
`Expression` (matching the existing single-child-wrapper pattern, `TypeIs` /
`MemberAccess`), plus the `ReflectionHelper.ChangeType.Expression` accessor it needs.
Verified green on SQLite + ClickHouse.

## Baseline note

Beyond the literal→parameter reshaping in `SelectScalarTests.Add` (a minor improvement —
plan-cache-friendlier), one baseline shifts semantically: `EvaluationTests.Evaluate_RemappedTimeSpan`.

`MapMember<TimeSpan>(ts => ts.Ticks, ts => ToTicks(ts))` maps `TimeSpan.Ticks` to a
`[Sql.Expression(ServerSideOnly = true)]` method whose managed body throws. Previously the
mapping was not expanded in the initial expose, so `new TimeSpan(88888888L).Ticks` stayed the
real CLR property and the whole `(long)(5.988M * …)` subexpression constant-folded to `532266661`
— silently bypassing the server-side-only `ToTicks`. With mappings now expanded in the initial
expose, `.Ticks` becomes `ToTicks(ts)`, whose throwing body precludes client folding, so the
arithmetic is emitted server-side as `CAST(Floor(5.988 * CAST(88888888 AS BIGINT)) AS BigInt)`.
Result is identical (`0 > 532266661` → false), and the new SQL is the more correct outcome:
the `ServerSideOnly` mapping is now actually honored instead of folded away.


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed two problems with custom expression mappings registered through `Expressions.MapMember`. A mapped member used to project a correlated aggregate directly - for example a helper method mapped to `p.Children.Count(...)` - previously threw an error during query translation instead of producing the expected subquery; it now translates the same as the equivalent hand-written association aggregate. Separately, a mapped `string.CompareTo` / `string.Compare` with a null operand could return a wrong result when evaluated in memory (the generated SQL was already correct); that comparison is now correct.


_(none)_


---
<sub>Generated from commit `e6b347f`.</sub>

PR #5584, opened 2026-06-04
Fix #5413: Oracle UPDATE ROW correlated subquery

### LINKED ISSUE #5413: Subquery in update issues (v6 regression)

jods4:
Fixes #5413.

Repro test cases for various Oracle UPDATE ROW shapes (by @jods4), plus the fix
for the remaining failing shape - `UpdateFromSubqueryRowShouldRemainSimple2`, a
non-PK join under `.Single()`:

- `SelectQueryOptimizerVisitor.JoinMoveSubQueryUp`: flatten a subquery-expression
  apply only when its source is a single table whose joins are each
  `IsLimitedToOneRecord`.
- `BasicSqlOptimizer.FoldJoinConditionAndDetach`: rebase the lifted apply's `ON`
  onto the subquery's source columns so the correlation key isn't a dangling
  self-reference.

Result: the apply lifts to a clean correlated subquery - no `EXISTS`, no dangling
column. `UpdateFromTests` green on Oracle / SQLite / SQL Server / PostgreSQL.


MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] No findings. The two engine changes are narrowly gated, well-commented, and sound; the regression test matches the linked issue. Both helpers (`JoinMoveSubQueryUp`, `FoldJoinConditionAndDetach`) are private static — no public-API surface change.
- [x] This PR has no milestone set; linked issue #5413 is on `6.3.0`. Worth setting one before merge.

## Out-of-scope observations

- **`JoinMoveSubQueryUp` guard is cross-cutting, but only the Oracle path is explicitly asserted** — The new early-return in `SelectQueryOptimizerVisitor.JoinMoveSubQueryUp` blocks flattening of any `IsSubqueryExpression` apply whose single source table has a join that is not one-record-limited. That branch runs for every provider, yet both regression tests are Oracle12Plus-only. The change is safe by construction (returning `false` leaves the apply intact — always valid SQL), and non-Oracle coverage is implicit via the existing `UpdateFromTests.Row` suite. Confirmed empirically: the baselines diff shows **zero** non-Oracle baseline movement, so no other provider's SQL shifted.

## Baselines
Delta: [linq2db.baselines PR #1893](https://github.com/linq2db/linq2db.baselines/pull/1893) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5584)

10 baselines added (2 new Oracle-only tests × 5 Oracle versions 12–23), 0 modified, 0 deleted. Independently verified against the files via three-dot diff.

### New Oracle-only tests for row-setter correlated-subquery UPDATE (#5413)
- **Complex case (chained `OUTER APPLY` + `ROWNUM` guards + DUAL anchor)** — `UpdateFromSubqueryRowShouldRemainSimple1`, identical across Oracle 12/18/19/21/23. Correct row-setter syntax, correct correlation, no spurious `EXISTS`.
- **Simple case (flat `LEFT JOIN`, no APPLY/ROWNUM)** — `UpdateFromSubqueryRowShouldRemainSimple2`, identical across the five versions. Optimizer correctly kept it flat (neither join is `Take(1)`-bound). Oracle 11 absent — correct, it lacks `(col1,col2) = (subquery)` row-setter syntax.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed an Oracle `UPDATE` with a correlated row-subquery setter (e.g. a non-PK join narrowed by `.Single()`) generating invalid SQL. The correlated subquery now lifts cleanly instead of leaving a dangling self-reference. Thanks to [jods](https://github.com/jods4). ([#5584](https://github.com/linq2db/linq2db/pull/5584))


_(none)_


---
<sub>Generated from commit `f17d81a`.</sub>

Issue #1923, opened 2019-10-09
Best way to populate master detail


ferbenor:
Good afternoon, excuse the question, which is the most optimal way to fill out a master detail scheme?, for example, a department with its employees, considering that each department has several employees.

sdanyliv:
This feature is still in development #1756. 
So best way is to map it yourself. Otherwise `LoadWith` will introduce N+1 queries.

MaceWindu:
#1756

PR #5586, opened 2026-06-06
Fix "Called when root is not initialized" on Nullable<T>.HasValue over an unbound member

### LINKED ISSUE #5575: InvalidOperationException: Called when root is not initialized when projecting a conditional with Nullable<T>.HasValue/.Value
### Description

Hi,

While migrating from 5.4 to 6.x I hit a build-time crash on a query that worked fine on 5.4. It throws while *building* the query (before any SQL runs):

```
System.InvalidOperationException : Called when root is not initialized.
   at LinqToDB.Internal.Linq.Builder.ExpressionBuildVisitor.GetSqlCacheKey(Expression path)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuildVisitor.VisitMember(MemberExpression node)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuildVisitor.VisitConditional(ConditionalExpression node)
   ...
```

It seems to be triggered by a `? :` in the projection that does a `Nullable<T>.HasValue` / `.Value` access on a member. In my case the member comes from a LEFT JOINed in-memory `AsQueryable()` source, and the projection computes a rate column like `x.Count.HasValue ? x.Count.Value / (decimal)x.Total * 100 : (decimal?)null`.

**Versions:** reproduced on 6.3.0 and on current master (6.4.0, around commit `2e67bafc9`). Provider is SQL Server, but the failure is in the provider-agnostic expression builder.

### Reproduction Steps

```csharp
sealed class Counts { public int Id { get; set; } public int Count { get; set; } }

var counts = new[] { new Counts { Id = 1, Count = 5 } };

var q =
    from t in db.GetTable<SomeTable>()                 // SomeTable.Value is int
    from c in counts.AsQueryable()
                    .Where(c => c.Id == t.Id)
                    .DefaultIfEmpty()                  // LEFT JOIN -> c.Count nullable
    select new
    {
        t.Id,
        Rate = ((int?)c.Count).HasValue
                   ? (decimal?)(((int?)c.Count).Value / (decimal)t.Value * 100)
                   : null
    };

_ = q.ToList();   // throws "Called when root is not initialized" during build
```

**What I found (in case it helps):** in `ExpressionBuildVisitor.VisitMember`, the method captures `var context = FoundRoot;`, then calls `HandleNullableT(node)`. `HandleNullableT` does an inner `Visit(node.Expression)` which resets `FoundRoot` to null. The captured `context` is still non-null though, so the `if (context != null)` block below runs and calls `GetSqlCacheKey(node)`, which reads `FoundRoot` (now null) and throws. Restoring `FoundRoot = context;` right after the `HandleNullableT` block fixes it for me and the full test suite still passes — but you'll know better whether that's the right place.

Happy to refine the repro if it doesn't reproduce on your end.

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.3.0, 6.4.x

### .NET version

9.0.312

### Database

SQL Server 2022

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5575.

`Nullable<T>.HasValue` on a member that was left **unbound** in an earlier projection (never assigned, so it folds to `NULL`) threw at build time, before any SQL ran:

```
System.InvalidOperationException : Called when root is not initialized.
   at ExpressionBuildVisitor.GetSqlCacheKey(Expression path)
   at ExpressionBuildVisitor.VisitMember(MemberExpression node)
```

The original minimal repro pointed at a LEFT JOIN, but the join actually *bound* the member, so it never reproduced. The real trigger is a two-stage projection where the nullable is never assigned:

```csharp
db.GetTable<T>()
    .Select(c => new StatEntity { Id = c.Id })                                     // LeadCount left unbound
    .Select(s => new { s.Id, Rate = s.LeadCount.HasValue ? (decimal?)1m : null }); // threw at build
```

## Cause

`ExpressionBuildVisitor.HandleNullableT` visits the inner expression to test for a SQL placeholder. For an unbound member the inner resolves to a member access on a `SqlGenericConstructorExpression` (not a placeholder), so it falls through unhandled, but that inner `Visit` has already cleared `FoundRoot`. The caller's `GetSqlCacheKey` then reads a null `FoundRoot` and throws. Projecting the member directly already folds it to `null`; only the `HasValue` / `Value` path hit this.

## Fix

Restore `FoundRoot` after the inner visit in the `HasValue` fall-through, mirroring the save/restore that `MakeWithCache` already does around its sub-translation. `HasValue` then resolves to `false` for the null member.

## Tests

Replaced the original (non-reproducing) LEFT JOIN test with the unbound-member shape: the reported conditional form and the minimal bare-`HasValue` form, both via `AssertQuery`.


Tests/Linq/UserTests/Issue5575Tests.cs
```diff
@@ -0,0 +1,61 @@
...
+		{
+			using var db = GetDataContext(context);
+			using var tb = db.CreateLocalTable(new[] { new SomeTable { Id = 1, Value = 4 } });
+
+			var query = tb
+				.Select(c => new StatEntity { Id = c.Id }) // LeadCount left unbound
+				.Select(s => new
+				{
+					s.Id,
+					Rate = s.LeadCount.HasValue ? (decimal?)1m : null
```
MaceWindu: **Minor · MIN001** — The issue title and the reporter's repro both feature `Nullable<T>.Value` over the unbound member (`((int?)c.Count).Value` in the conditional true-branch), but neither test exercises a `.Value` access on the unbound member: `ConditionalNullableHasValue` uses a constant `(decimal?)1m` true-branch and `UnboundNullableHasValue` only reads `.HasValue`. The fix is correctly scoped to the `.HasValue` fall-through (the `.Value` branch returns early and never reaches the crash site), so this is a coverage gap rather than a fix defect — a future regression that re-broke the reported `.Value`-in-true-branch shape would still pass these tests.

Fix: consume the unbound member's `.Value` in the conditional true-branch so both `.HasValue` and `.Value` over the unbound member are exercised in one projection. This changes the generated SQL, so this test's baselines regenerate — confirm the suggested form actually passes before relying on it.

```suggestion
					Rate = s.LeadCount.HasValue ? (decimal?)(s.LeadCount.Value / 2m) : null
```

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Fix is correct and minimal. The `FoundRoot = saveRoot;` restore added in `HandleNullableT`'s `HasValue` fall-through mirrors the existing `MakeWithCache` save/restore, and `HasValue` is the only path that reaches the `GetSqlCacheKey` crash site (the `.Value` branch returns early). The placeholder case is still returned before the restore, so legitimate SQL resolution is unaffected.
- [x] No public API surface changes — the fix is in `LinqToDB.Internal.*`; the remainder is test-only.
- [x] Baselines clean — see below.

One line-level Minor finding (MIN001) is posted inline on the test file.

## Out-of-scope observations
- **`FoundRoot` restore is not exception-safe, matching the existing precedent.** The new `FoundRoot = saveRoot;` is a plain assignment after the inner `Visit(node.Expression)`; if that inner visit threw, `FoundRoot` would not be restored. This is not a defect: it mirrors the existing `MakeWithCache` save/restore (also a plain assignment without try/finally), and on a propagating build exception the whole build fails so the stale value is unobservable. Noted only so the intentional symmetry is visible — converting either site to try/finally would be a separate, broader change.

## Baselines
Delta: [linq2db.baselines PR #1923](https://github.com/linq2db/linq2db.baselines/pull/1923) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5586)

114 new SQL files (2 new tests × 57 providers) plus 2 unrelated modifications. No deletions, no cross-provider anomalies (independently confirmed against the files).

### New regression tests — `Issue5575Tests`
Both `ConditionalNullableHasValue` and `UnboundNullableHasValue` emit the same two-query shape across all providers: `SELECT [s].[Id] FROM [SomeTable] [s]`, then `SELECT [t1].[Id], [t1].[Value] FROM [SomeTable] [t1]` — no WHERE, consistent with `HasValue` over the unbound (null-folding) member resolving at build time. SqlCe differs only in rendering the column alias as `[Value] as [Value_1]` (pre-existing provider quirk, not a structural difference).

### Unrelated modifications — `DontCloseAfterUse` (SqlServer.Contained)
`DontCloseAfterUse_DataConnection` and `DontCloseAfterUse_DataContext` each gained an 11-line tail append for the SqlServer.2019 config. These baselines are cumulative CI-trace accumulators; the append is a normal CI artefact unrelated to this PR (0 deletions).

Tests/Linq/UserTests/Issue5575Tests.cs
MaceWindu: Addressed in f0c8e466d: the conditional true-branch now consumes the unbound member's .Value (s.LeadCount.Value / 2m), so both .HasValue and .Value over the unbound member are exercised. Verified locally on SQLite (8/8); baselines regenerate as expected.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed an `InvalidOperationException` ("Called when root is not initialized") thrown at query-build time when `Nullable<T>.HasValue` (or `.Value`) was used on a member left unbound by an earlier projection — for example a two-stage `Select` where the nullable member is never assigned and is then tested with `.HasValue` in the next projection. `HasValue` now correctly resolves to `false` for such an unassigned (null) member. ([#5586](https://github.com/linq2db/linq2db/pull/5586))


_(none)_


---
<sub>Generated from commit `d6b74cc`.</sub>

Issue #1912, opened 2019-10-01
Guidance on integrating linq2db with other code to manage connections


davidglassborow:
I am looking for any thoughts on how best to integrate / extend l2db, much like the [linq2db.EntityFrameworkCore](https://github.com/linq2db/linq2db.EntityFrameworkCore) project.

Currently we have a home grown ORM that handles plain textual SQL, and I am looking to add Linq into it, ideally using l2db.

Our current system provides some features that we need to keep like caching, throttling, remote data over http, etc.  It's a huge system with thousands of queries, so I'm looking to support Linq alongside the current plain SQL. We are running against SQL Server.

I can probably hack something that works, but if possible wanted some guidance on how best to work with the l2db framework.

Our core challenges integrating l2db:

1. Generating the metadata without using T4 templates

2. Overriding the Command object that linq2db uses for retrieving data
    - Ideally I'd also provide the connection
    - I want to use the built in l2db mapping, so don't just need the SQL text

## 1. Metadata & T4

The T4 issue is that we already have a migration system via scripts, that runs on CI, handles multiple replicating databases, etc, so T4 doesn't work for our use case.

My current plan is to covert the T4 templates back into C# and host that logic inside our existing tool, unless there is an easier way you can think of ?

## 2. Overide connection / command

Investigations so far:

1. Override `ExecuteReader` in my `LinqToDB.Data.DataConnection` subclass to return a `IDataReader` from my own command subclass
    - The returned IDataReader is downcast internally e.g. if I create my own MiniProfiler command, I get `Unable to cast object of type 'StackExchange.Profiling.Data.ProfiledDbDataReader' to type 'System.Data.SqlClient.SqlDataReader`. This is occurring inside LinqToDB.Linq.IQueryRunner somewhere
    - This is also wasteful as l2db still has a connection, command, etc.
2. Looked at [LinqExtensions.ProcessSourceQueryable](https://github.com/linq2db/linq2db/pull/1116), as it's used [here](https://github.com/linq2db/linq2db.EntityFrameworkCore/blob/master/Source/LinqToDB.EntityFrameworkCore/LinqToDBForEFTools.cs#L54)
    - I can see it uses reflection to extract the Db Context out of the EF.Core IQueryable
    - It then extracts the connection info from the Db Context, creating a DataConnection with the details
[...]

PR #5587, opened 2026-06-06
Fix #5576: spurious [item] column on local-collection LEFT JOIN with decimal projection

### LINKED ISSUE #5576: In-memory AsQueryable() LEFT JOIN emits a spurious whole-object [item] column -> "Failed to convert parameter value from a <Type> to a Decimal"
### Description

Hi,

Another one from the 5.4 → 6.x migration. When I LEFT JOIN a db query to an in-memory collection via .AsQueryable(), and the element is a small class (multiple members) where one member flows into decimal arithmetic in the projection,  the generated VALUES clause gets an extra bogus column holding the whole element object, typed Decimal:

```
  LEFT JOIN (VALUES
      ('7a41…', 5, @value),          -- @value = the whole element object!
      ('c44d…', 10, @value_1)
  ) [lc]([CampaignGuid], [Count], [item]) ON [c].[CampaignGUID] = [lc].[CampaignGuid]
  -- DECLARE @value Decimal  /  SET @value = <Namespace>.Counts
```
  At execution ADO.NET throws:
```
  System.InvalidCastException : Failed to convert parameter value from a Counts to a Decimal.
   ---> System.InvalidCastException : Object must implement IConvertible.
     at Microsoft.Data.SqlClient.SqlParameter.CoerceValue(...)
```

The first two columns are correct — it's the third [item] column (the element bound as one parameter) that's wrong. Interestingly, a COUNT(*) variant of the same query (no decimal arithmetic in scope) emits the correct 2-column VALUES  with no [item], and a scalar source like Guid[].AsQueryable() correctly emits a single legitimate [item]. So it only goes wrong for a multi-member element while a Decimal descriptor is in scope.

### Reproduction Steps

```csharp
sealed class Counts { public Guid Key { get; set; } public int Count { get; set; } }   // multi-member class

 var counts = new[] { new Counts { Key = someGuid, Count = 5 } };

 var q = db.GetTable<Campaign>()                       // Campaign.Sold is int
     .LeftJoin(
         counts.AsQueryable(),
         (c, lc) => c.Guid == lc.Key,
         (c, lc) => new
          {
             c.Guid,
             Count = lc.Count,                          // member projected …
             Rate  = lc.Count / (decimal)c.Sold * 100   // … and used in decimal arithmetic
          });

 _ = q.ToList();   // VALUES gets a spurious whole-object [item] column -> InvalidCastException
```

the element must be a class (a struct is treated as scalar, separate path); and removing the decimal arithmetic makes it build correctly.

in EnumerableContext.MakeExpression the "render element as a single item column" vs "decompose into columns" decision looks at Builder.CurrentDescriptor — when that's a Decimal descriptor leaking in  from the surrounding arithmetic, dbType comes from the descriptor (Decimal, != Undefined) instead of the element type (which is Undefined for the unmapped class), so a complex entity wrongly takes the scalar [item] path and gets bound  as one parameter. Basing the decision on the element type only (ignoring CurrentDescriptor) fixes it for me — it then takes the same BuildFullEntityExpression path the COUNT(*) query already takes — and the full test suite still  passes.

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.3.0, 6.4.x

### .NET version

9.0.312

### Database

SQL Server 2022

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes the spurious whole-object `[item]` column emitted when a DB table is LEFT JOINed
to a **local (in-memory) collection** (rendered as a `VALUES` clause) where the element
is a multi-member class and one member flows into decimal arithmetic in a later projection.
The element was bound as a single parameter typed by a leaked `Decimal` descriptor, giving
`InvalidCastException: Failed to convert parameter value from a <Type> to a Decimal` at execution.

## Repro
The three-layer shape triggers it: LEFT JOIN -> a `.Select` doing the decimal arithmetic ->
a re-mapping `.Select`. The re-mapping keeps the decimal projection as an inner query; a
single collapsed projection does not reproduce.

## Root cause
When the LEFT JOIN's `DefaultIfEmpty` re-expands the local-collection source, the enclosing
projection member's `Decimal` descriptor is still the ambient `CurrentDescriptor`, so
`EnumerableContext` takes the scalar `[item]` path and binds the whole multi-member element
as one `Decimal` parameter.

## Fix
Reset the ambient column descriptor at the subquery boundary (`ExpressionBuildVisitor.GetSubQuery`),
next to the existing build-flags reset, so current-level state can't leak into subquery building.

## Test
Regression test for the three-layer scenario (fails before the fix, passes after). Verified
locally on `#5576`, `#5057`, and `AllAnyTests`; full provider matrix via CI.

Fixes #5576


Tests/Linq/UserTests/Issue5576Tests.cs
```diff
@@ -0,0 +1,136 @@
...
+		// Stage 3: a re-mapping projection (data-model -> domain-model).
+		sealed class Result
+		{
+			public Guid     CampaignGuid { get; set; }
+			public int?     Leads        { get; set; }
+			public decimal? Rate         { get; set; }
+		}
+
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/5576")]
+		public void LeftJoinLocalClassWithDecimalArithmetic([IncludeDataSources(TestProvName.AllSqlServer2008Plus)] string context)
```
MaceWindu: **Minor · MIN001** — The bug lives in shared query-builder code (`ExpressionBuildVisitor.GetSubQuery`) — the leaked `CurrentDescriptor` is resolved during expression building, before any provider-specific SQL emission — so the regression is provider-agnostic. But `[IncludeDataSources(TestProvName.AllSqlServer2008Plus)]` runs it on SQL Server only, so a future change that re-introduces the descriptor leak on a path another provider exercises would go uncaught. The sibling local-collection tests (`EnumerableSourceTests.InnerJoinArray` / `InnerJoinArray2`) use exclude-style gating to run on every capable provider.

Fix: Switch to an exclude-style gate so the regression runs wherever a local-collection LEFT JOIN can render, excluding only providers that can't. The replacement mirrors the sibling tests' exclusion list — widen/narrow it per the CI matrix result if an excluded provider still can't render the `VALUES`/derived-table source.

```suggestion
		public void LeftJoinLocalClassWithDecimalArithmetic([DataSources(TestProvName.AllAccess, ProviderName.DB2, TestProvName.AllInformix)] string context)
```

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] The fix is correct and minimal — confirmed against the regenerated baselines: the local-collection `VALUES` join now emits a clean 2-column `[lc]([Key], [Count])`, and the spurious whole-object `[item]` column is gone.
- [ ] Two unrelated baselines on `SqlServer.Contained` (`DontCloseAfterUse_DataConnection` / `_DataContext`) gained a duplicate trailing `SELECT [Person]` block — only on that one variant, no other SqlServer provider. These are query-accumulation baselines and a subquery descriptor reset has no path to emitting an extra whole-table query, so this looks like a trace-ordering/accumulation flake rather than an effect of the fix. Worth regenerating those two from a clean run (or confirming the diff is benign) before merging the baselines PR.

## Out-of-scope observations

- **Subquery-boundary reset vs the issue's narrower fix — the author's choice is the more robust one.** The issue proposed fixing this in `EnumerableContext.MakeExpression` by basing the scalar-vs-decompose decision on the element type alone (ignoring `CurrentDescriptor`). The PR instead resets the ambient descriptor at the subquery boundary in `GetSubQuery`, next to the pre-existing flags reset. The boundary reset is the more general choice: resetting the current-level descriptor is the established idiom across this file, the descriptor is current-level projection state by the same rationale the flags already are, and any subquery that legitimately needs a descriptor re-establishes its own. The narrower fix would have left the leak live for other subquery paths that read `CurrentDescriptor`. Backed by two full `test-all` CI runs.

## Baselines
Delta: [linq2db.baselines PR #1912](https://github.com/linq2db/linq2db.baselines/pull/1912) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5587)

26 files — 20 added (new regression test), 6 modified.

### New test: `Issue5576Tests.LeftJoinLocalClassWithDecimalArithmetic` (20 providers)
Verified output is clean — 2-column `VALUES [lc]([Key], [Count])`, no spurious `[item]` column.
- **SqlServer 2012+ / SA / Contained — IIF form** (18 providers): `IIF([lc].[Count] IS NOT NULL, (CAST([lc].[Count] AS Decimal(18, 10)) / [r].[Sold]) * 100, NULL)`
- **SqlServer 2008 — CASE WHEN form** (2 providers): same arithmetic body, `CASE WHEN … ELSE NULL END` (IIF is 2012+). Expected dialect split, not an anomaly.

### `Issue3260Test` — `ELSE 0` → `ELSE CAST(0 AS Decimal)` (4 SQLite providers)
Side-effect of the descriptor reset: the `SUM(CASE … ELSE 0 END)` constant now derives its type from the expression and is explicitly cast. Semantically equivalent for `SUM`. SQLite-only test (no SqlServer baselines exist on master), so there's no missing cross-provider regeneration.

### Cross-provider anomalies
- **`DontCloseAfterUse_DataConnection` / `_DataContext`** (SqlServer.Contained only): appended duplicate `SELECT [Person]` block; no other SqlServer provider affected. Likely a trace-accumulation flake — see review note above.

MaceWindu (approved):
waiting for CI

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed an `InvalidCastException` ("Failed to convert parameter value ... to a Decimal") thrown when a table is LEFT JOINed to a local in-memory collection (e.g. via `.AsQueryable()`) whose element is a multi-member class and one of its members feeds decimal arithmetic in a later projection. A spurious whole-object column was emitted in the generated `VALUES` clause; the join now builds correctly.


Fixed InvalidCastException when LEFT JOINing a local collection whose member flows into decimal arithmetic in a later projection.


---
<sub>Generated from commit `9da1c3b`.</sub>

PR #5583, opened 2026-06-04
Tests: add primary keys to keyless local tables

MaceWindu:
## What

Adds a `[PrimaryKey]` to test models that are materialized as **keyless** local tables (`CreateLocalTable`) by a number of fixtures. The key is a plain `[PrimaryKey]` (all providers) on a column that is already unique in the seeded data, so behavior is unchanged for existing providers.

## Why

Some providers reject tables without a primary key (e.g. YDB requires a PK on every table). These fixtures created keyless tables and so couldn't run there. Adding a PK on an existing-unique column makes the tables creatable everywhere without changing query semantics.

Split out of the experimental YDB provider work (#5564) so the all-provider `CREATE TABLE` baseline changes are reviewed/CI'd on their own, independent of the YDB PR.

## Scope

Test-only — `[PrimaryKey]` attributes on test model classes across:
`CteTests`, `ExpressionsTests`, `CountByMethodTests`, `DistinctByMethodTests`, `ExceptByMethodTests`, `IndexMethodTests`, `IntersectByMethodTests`, `MinByMaxByMethodTests`, `UnionByMethodTests`, `ConvertTests`, `StringConcatTests`, `StringFunctionsTests`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5589, opened 2026-06-07
Tests: keyless-table corrections + deterministic Issue5549 baseline (follow-up to #5583)

MaceWindu:
Follow-up to #5583. Extracts the provider-agnostic test corrections that are currently bundled into the YDB provider PR (#5564) into a standalone PR, so that #5564's baseline diff stays YDB-only and is easier to review. These changes move every provider's baselines (they are deliberately provider-agnostic), which is exactly what we don't want mixed into the YDB review.

## Test corrections

- **CharTypesTests** — convert `StringTestTable` / `CharTestTable` from the shared keyless `AllTypes` table (with per-provider column maps) to self-contained `[Table]` + `[PrimaryKey]` types created via `CreateLocalTable<>`, and reorganize the per-provider trailing-space/`\0` assertions. Carries YDB-aware assertion branches as dead-but-compiling code (they only execute once the YDB provider lands).
- **EnumMappingTests** — insert a distinct row (`Id + 1`) rather than re-inserting the source PK (so it doesn't collide with an enforced primary key); convert `TestTable5` to a local table with a `[PrimaryKey]`; `Cleaner(db, 2)`.
- **BulkCopyTests** — `DateOnlyTable` uses the `DateOnly` value as a natural `[PrimaryKey]` instead of an identity `int Id`.
- **TestProvName** — add the `AllYdb` constant (referenced by the CharTypes assertions).

## Determinism fix

- **PostgreSQLTests.Issue5549** — `Issue5549Populate` used `NodaTime.SystemClock.Instance.GetCurrentInstant()`, embedding wall-clock time into the captured baseline SQL (non-deterministic — the `.sql` drifts on every regeneration). Replaced with a fixed `NodaTime.Instant`; the test only relies on relative arithmetic and row counts, so behavior is unchanged. Pre-existing on master, surfaced while reviewing #5564's baselines.

## Sequencing

Land this first; it regenerates the affected providers' baselines on master. #5564 then rebases on master and these corrections de-duplicate out of its diff, leaving its baseline movement YDB-only.

### DIFF
--- a/Tests/Base/TestProvName.cs
+++ b/Tests/Base/TestProvName.cs
@@ -252,6 +252,10 @@ namespace Tests
 		public const string AllDuckDB = ProviderName.DuckDB;
 		#endregion
 
+		#region YDB
+		public const string AllYdb = ProviderName.Ydb;
+		#endregion
+
 		/// <summary>
 		/// Fake provider, which doesn't execute any real queries. Could be used for tests, that shouldn't be affected
 		/// by real database access.
--- a/Tests/Linq/DataProvider/PostgreSQLTests.cs
+++ b/Tests/Linq/DataProvider/PostgreSQLTests.cs
@@ -3096,7 +3096,9 @@ $function$
 
 		static NodaTime.Instant Issue5549Populate(IDataContext db)
 		{
-			var now = NodaTime.SystemClock.Instance.GetCurrentInstant();
+			// Fixed instant (not SystemClock) so the captured baseline SQL is deterministic; the test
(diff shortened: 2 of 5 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Test-only: convert several keyless shared-table tests (CharTypes, EnumMapping, BulkCopy) to self-contained local tables with primary keys, and make the PostgreSQL Issue5549 baseline deterministic by using a fixed instant instead of the system clock.


_(none)_


---
<sub>Generated from commit `a397688`.</sub>

PR #2013, opened 2019-12-26
Add notes about error when wrong t4 tool selected

MaceWindu:
Fix #2002 
well, kinda fix - we cannot fix it, only provide some limited guideance to users

PR #5600, opened 2026-06-08
Move parameter creation into providers

igor-tkachev:
## Summary

This PR fixes a regression in Oracle parameter binding where large XML/string values could be bound as regular string parameters instead of NCLOB-compatible parameters, causing Oracle failures for `OracleXmlTable` and other long string parameter scenarios.

The fix moves `DbParameter` creation into the data provider so provider-specific parameter logic can be applied consistently for both LINQ-generated parameters and raw SQL `DataParameter` values.

## Changes

- Added `IDataProvider.CreateParameter(...)` and `DataProviderParameterContext`.
- Updated LINQ query execution parameter creation to call the provider.
- Updated raw SQL `DataParameter` binding to call the provider.
- Kept default parameter type inference in `DataProviderBase`.
- Added Oracle-specific inference for long string parameters:
  - undefined string parameters with length >= `OracleOptions.MaxStringParameterLength` are bound as `NText`
  - default threshold is `4000`
  - `null` disables automatic NCLOB inference
- Added `OracleOptions.MaxStringParameterLength` with binary-compatibility constructor/deconstruct overloads.
- Added Oracle regression tests for long string and `OracleXmlTable` parameter handling.


.codex/skills/compatibility/SKILL.md
MaceWindu, line 1: all agents changes should go to curation branch

MaceWindu (changes requested):

Source/TODO.md
MaceWindu, line 1: better to fill task or on curation we have knowledge-base infrastructure that also tracks such issues

MaceWindu (changes requested):

Source/LinqToDB/DataProvider/DataProviderParameterContext.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Data;
+
```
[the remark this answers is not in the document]
MaceWindu: **Correction:** Verified at HEAD (`31e7608`) — this compiles. `DataProviderParameterContext.cs` is declared in `namespace LinqToDB.DataProvider`, which nests inside `LinqToDB`, so `DbDataType` (defined in `LinqToDB`) resolves via enclosing-namespace lookup without an explicit `using LinqToDB;`; `ParameterDirection` is covered by the file's `using System.Data;`. The full build + test-all matrix is green at this commit, so this is not actionable.

Source/LinqToDB/DataProvider/DataProviderParameterContext.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace LinqToDB.DataProvider
+{
+	public readonly struct DataProviderParameterContext
```
MaceWindu, line 10: **Major · MAJ001** — New public type `DataProviderParameterContext` carries no XML documentation — not on the struct, the constructor, or any of its five public properties (`Name`, `DbDataType`, `Value`, `Direction`, `IsDbDataTypeExplicit`). This is the public payload of the PR's headline new API (`IDataProvider.CreateParameter`), and the repo convention requires XML docs on new public types/members.

Fix: Add a `<summary>` to the struct and a one-line `<summary>` to the constructor and to each of the five properties. The suggestion documents the type declaration; the constructor and properties still need their own `///` summaries.

```suggestion
	/// <summary>
	/// Provides parameter metadata to <see cref="IDataProvider.CreateParameter"/> so a provider can
	/// materialize a <see cref="System.Data.Common.DbParameter"/> for a query or command.
	/// </summary>
	public readonly struct DataProviderParameterContext
```

Source/LinqToDB/DataProvider/IDataProvider.cs
```diff
@@ -48,7 +48,9 @@ public interface IDataProvider
 		object?            GetConnectionInfo     (DataConnection dataConnection, string parameterName);
 		Expression         GetReaderExpression   (DbDataReader reader, int idx, Expression readerExpression, Type toType);
 		bool?              IsDBNullAllowed       (DataOptions options, DbDataReader reader, int idx);
+		DbParameter        CreateParameter       (DataConnection dataConnection, DbCommand command, DataProviderParameterContext context);
```
MaceWindu, line 58: **Major · MAJ002** — The new public interface member `CreateParameter` carries no XML documentation, though it is the PR's primary new public entry point and the repo convention requires docs on new public members. (Some sibling members here — `SetParameter`, `GetReaderExpression`, `IsDBNullAllowed` — are also undocumented, so this partly follows local style; but new surface should be documented.)

Fix: Add a `<summary>` plus `<param>`/`<returns>` above the declaration.

```suggestion
		/// <summary>
		/// Creates a <see cref="DbParameter"/> for the given <paramref name="context"/> on the specified command.
		/// </summary>
		/// <param name="dataConnection">Data connection instance.</param>
		/// <param name="command">Command the parameter will be attached to.</param>
		/// <param name="context">Parameter metadata (name, type, value, direction).</param>
		/// <returns>Configured parameter instance.</returns>
		DbParameter        CreateParameter       (DataConnection dataConnection, DbCommand command, DataProviderParameterContext context);
```

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDataProvider.cs
```diff
@@ -357,6 +357,19 @@ protected override void SetParameterType(DataConnection dataConnection, DbParame
 			}
 		}
 
+		protected override DbDataType InferParameterDataType(DataConnection dataConnection, DbDataType dbDataType, object? paramValue)
+		{
+			if (dbDataType.DataType == DataType.Undefined &&
+			    paramValue is string value                &&
+			    dataConnection.Options.FindOrDefault(OracleOptions.Default).MaxStringParameterLength is { } maxStringParameterLength &&
+			    value.Length >= maxStringParameterLength)
```
MaceWindu, line 362: **Suggestion · SUG001** — The threshold test compares `value.Length` (UTF-16 char count) against a limit whose XML doc (`OracleOptions.cs:27`) describes it as "aligned with Oracle MAX_STRING_SIZE semantics" with a 4000 default. Oracle's `VARCHAR2`/`NVARCHAR2` limits gated by `MAX_STRING_SIZE` are **byte**-based (4000 bytes for `STANDARD`), not char-based. For multibyte content a string can exceed the byte limit at fewer than `MaxStringParameterLength` characters, stay under the char threshold, and *not* be promoted to NCLOB — the exact overflow this fix targets. For ASCII the 4000-char default lines up closely with the 4000-byte limit (the common case, handled well), so the "aligned with MAX_STRING_SIZE" claim is a char-vs-byte approximation.

Fix: Given the threshold is user-tunable (a user hitting multibyte overflow can lower it) and `null` disables it, the low-risk resolution is to soften the doc wording — drop the literal "aligned with Oracle MAX_STRING_SIZE semantics" and state it is a character-count heuristic with the 4000 default chosen to match the common single-byte `VARCHAR2` limit. Alternatively, compare an encoded byte length when precise alignment is required.

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -435,6 +435,40 @@ public void TestString([IncludeDataSources(TestProvName.AllOracle)] string conte
 			}
 		}
 
+		[Test]
+		public void LongStringParameterTest([IncludeDataSources(TestProvName.AllOracle)] string context)
```
MaceWindu, line 439: **Minor · MIN001** — `MaxStringParameterLength` exposes two documented non-default behaviors — `null` disables NCLOB inference, and a custom value changes the `>=` promotion threshold — but only the default (4000) path is exercised here. Neither the `null`-disables path nor a custom-threshold path is tested, so a regression that (e.g.) ignored a `null` setting and always inferred NCLOB would not be caught.

Fix: Add a test that sets `MaxStringParameterLength = null` (via `DataOptions`/`OracleOptions`) and asserts a long undefined-type string is *not* bound as NCLOB, and optionally one with a small custom threshold asserting a shorter string *is* promoted. (See also MAJ003 — the bulk-copy path needs its own coverage.)

Source/LinqToDB/DataProvider/DataProviderParameterContext.cs
igor-tkachev: fixed

Source/LinqToDB/DataProvider/IDataProvider.cs
igor-tkachev: fixed

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDataProvider.cs
igor-tkachev: I agree that Oracle limits are byte-based, while the current check uses .NET string length. I don't think we should switch this logic to byte-length calculation in this PR.

`MaxStringParameterLength` is intentionally a user-configurable threshold. Users who work with multibyte data or non-default Oracle settings can lower or adjust it for their environment. Trying to calculate the exact byte length here would require assumptions about database/session character set and provider behavior, and could easily introduce false precision.

The current logic is a conservative, simple heuristic for the common case and keeps the regression fix focused. I can clarify the XML-doc to avoid implying byte-exact `MAX_STRING_SIZE` semantics.

Tests/Linq/DataProvider/OracleTests.cs
igor-tkachev: I would keep the `MaxStringParameterLength` name.

This option is intentionally modeled as an Oracle-specific string size threshold, not as an implementation-specific `StringToNClobThreshold`. The naming is close to Oracle terminology around `MAX_STRING_SIZE`, which makes it easier to understand for Oracle users.

The value is also user-configurable, so we shouldn't try to encode all Oracle byte/charset semantics into the option name. The implementation uses .NET string length as a simple threshold for deciding when an undefined string parameter should be promoted to NCLOB-compatible binding. Users who need a different cutoff for their database charset or provider behavior can adjust the value.

I agree that the XML documentation should avoid implying byte-exact Oracle `MAX_STRING_SIZE` behavior. We can clarify that the value is a character-count threshold used by linq2db for parameter binding, with the default chosen to match the common Oracle string limit.

.codex/skills/compatibility/SKILL.md
MaceWindu: Addressed at HEAD (a7a53d3): the agent-config (`.codex`/`.claude`) skill changes are no longer part of this PR. Resolving.

Source/TODO.md
MaceWindu: Addressed at HEAD (a7a53d3): `Source/TODO.md` and its `linq2db.slnx` reference were removed. Resolving.

Tests/Linq/DataProvider/OracleTests.cs
```diff
@@ -435,6 +435,97 @@ public void TestString([IncludeDataSources(TestProvName.AllOracle)] string conte
...
+			var value = "LongStringValue"; // length >= 10
+
+			conn.Execute(
+				$"INSERT INTO {tableName} (\"Id\", \"NClob\") VALUES (1, :p)",
+				new DataParameter { Name = "p", Value = value });
+
+			parameterInterceptor.Parameters.Length.ShouldBe(1);
+
+			if (parameterInterceptor.Parameters[0] is OracleParameter p)
+				p.OracleDbType.ShouldBe(OracleDbType.NClob);
```
MaceWindu, line 478: **Minor · MIN002** — The `OracleDbType` assertion sits inside `if (... is OracleParameter p)`, so it is silently **skipped** (not failed) when the captured parameter is not a managed `Oracle.ManagedDataAccess.Client.OracleParameter`. For `LongStringParameterNClobInferenceDisabledTest` and `LongStringParameterExplicitTypeNotPromotedTest`, which skip execution via `SaveAndSkipCommandInterceptor`, this check is the *only* meaningful assertion — so on a non-managed Oracle context the test passes having verified nothing.

Fix: assert the type instead of guarding on it, so a type mismatch fails the test. Applies at all three sites (477-478, 502-503, 525-526); the suggestion covers the first.

```suggestion
			parameterInterceptor.Parameters[0].ShouldBeOfType<OracleParameter>().OracleDbType.ShouldBe(OracleDbType.NClob);
```

MaceWindu (commented):
🤖 Agent-assisted review. Verified against PR HEAD (`5c2366b`).

## Prior-review audit

All 13 review threads are resolved; re-verified each at HEAD:

- **Copilot** (Length/Precision/Scale regression, NUnit `Assert.That` vs Shouldly, `[NotNull]` on nullable `ID2`, missing BOM, "`DbDataType` won't compile") → all fixed or were inaccurate. `DataProviderBase.CreateParameter` applies Size/Precision/Scale only for an explicit `DbDataType`; the raw-SQL path passes explicit, the LINQ path does not — matching pre-refactor behavior.
- **Earlier maintainer review** (XML docs on `DataProviderParameterContext` + `IDataProvider.CreateParameter`, byte-vs-char doc wording, null / custom-threshold test coverage) → all addressed at HEAD.
- **Copilot** `IDataProvider.CreateParameter` binary break → real but handled: `virtual` on `DataProviderBase` (derived providers unaffected) + CP0006 suppressions across all five TFMs.

## Findings

One minor test-robustness issue, posted inline (MIN002). No blockers.

## Public API (milestone 6.4.0)

Adds `DataProviderParameterContext` (struct + ctor + 5 properties), `IDataProvider.CreateParameter`, and `OracleOptions.MaxStringParameterLength` (+ binary-compat ctor / `Deconstruct`). The `IDataProvider` member is a documented binary break, covered by CP0006 baseline suppressions across net462 / netstandard2.0 / net8.0 / net9.0 / net10.0.

## Baselines

Clean — 30 added, 0 modified, 0 deleted; Oracle-only across 6 versions (11/12/18/19/21/23 Managed) for the 5 new tests. No cross-provider drift: the `IsDbDataTypeExplicit` / `CreateParameter` threading did not shift default parameter typing outside Oracle. [Baselines PR #2013](https://github.com/linq2db/linq2db.baselines/pull/2013).

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed long string and XML values being bound as ordinary string parameters, which failed for `OracleXmlTable` and other long-string parameter scenarios. A string parameter with no explicit data type whose length reaches `OracleOptions.MaxStringParameterLength` (new, default `4000`) is now bound as `NText`/NCLOB; set it to `null` to disable the inference. This applies to raw SQL and `DataParameter` values — a LINQ-generated parameter is typed from its column or CLR type, so it never reaches the inference and is unaffected. [component: Oracle]


_(none)_


---
<sub>Generated from commit `5c2366b`.</sub>

Issue #1935, opened 2019-10-17
nested transactions


lobster2012-user:
Hi!

I want to use nested transactions, 
how do i create them without ```transactionscope```?
When creating a new transaction, I see the destruction of the old

lobster2012-user:
I can not use TransactionScope
due to the inability to use distributed transactions.(sql server + oracle)
That is, ideally, I need something like this.

```c#
using(var point = connection.SavePointAsync())
{
      await point.ReleaseAsync()
      await point.RollbackAsync()
}
```

sdanyliv:
If you can do that by arbitrary DbConnection commands, why not? Linq2db do not have some special transactions processing, it is a thin layer between DbConnection and LINQ queries.
So just write needed extensions to do that. We do not implement everything because we are trying to support as many databases as we can and CheckPoints are too tied t SQL Server. We have thoughts how to do it and separate between different providers but it is still on design stage.

Shane32:
@sdanyliv is there any update/plan on supporting savepoints within transactions?  I believe they are supported by SQL Server, Oracle, Postgres, MySQL, SQLite, and others.  Many of them are implemented within the .NET provider; others would have to be implemented as SQL statements.

Shane32:
See:
- https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqltransaction.save?view=sqlclient-dotnet-standard-5.1
- https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqltransaction.save?view=dotnet-plat-ext-7.0
- https://www.npgsql.org/doc/api/Npgsql.NpgsqlTransaction.html
- https://dev.mysql.com/doc/dev/connector-net/latest/api/data_api/MySql.Data.MySqlClient.MySqlTransaction.html
- https://docs.oracle.com/en/database/oracle/oracle-database/23/odpnt/TransactionSave.html#GUID-85B498C1-87F9-4A69-BD8E-2DE468400E3B
- https://dev.mysql.com/doc/dev/connector-net/latest/api/data_api/MySql.Data.MySqlClient.MySqlTransaction.html - no Save support
- https://dev.mysql.com/doc/refman/8.0/en/savepoint.html

timbze:
In case anyone finds this helpful, I implemented the feature in my Db Context like this:

```c#
public async Task Save(string savepointName, CancellationToken cancellationToken = default)
    {
[...]

PR #5604, opened 2026-06-10
Add PreferClientCalculation option to keep projection computations client-side

sdanyliv:
Adds opt-in `LinqOptions.PreferClientCalculation`. When enabled, computed
expressions in the final projection (arithmetic, conditionals, unary ops, and
mapped members/methods that don't prefer server-side) are evaluated client-side
during materialization instead of being translated into extra SQL columns — the
v5 projection behavior that v6 changed by forcing everything through
`ExpressionBuildVisitor.TryConvertToSql`.

Expressions that prefer/require server-side evaluation (gated via
`Builder.PreferServerSide` — `PreferServerSide` / `ServerSideOnly` functions) and
set projections (`ForSetProjection`) still translate to SQL.

Default `false`, so existing behavior and SQL baselines are unchanged.

## Changes
- `LinqOptions.PreferClientCalculation` + `With`/`UsePreferClientCalculation` extensions.
- `ExpressionBuildVisitor`: `PreferClientCalculation(node)` predicate gated by
  `Builder.PreferServerSide`, applied at `TryConvertToSql`, `VisitUnary`,
  `VisitMethodCall`, `VisitMember`, `VisitBinary`.
- `PreferClientCalculationTests`: asserts on the SQL AST
  (`GetSelectQuery().Select.Columns` / `.Find`) plus `AssertQuery` correctness,
  across both option states.

## Notes
- Remote serializer unchanged: the option is consumed client-side during query
  building (the `SqlStatement` is built before serialization).

## Follow-up commit

`baf79f0541` — fixes a NULL regression the option exposed:

- `Sql.ToNullable` carried an `[Expression]` attribute, so the gate pulled it
  client-side and a SQL NULL (e.g. a missed `LEFT JOIN`) collapsed to
  `default(T)` at the client read. It now translates server-side via a
  `SqlFunctionsMemberTranslatorBase` registration (falling back to client-side
  when its argument can't be turned into SQL), preserving the NULL.
- The `VisitMethodCall` gate is keyed on attribute presence — only
  attribute-mapped functions move client-side, so attribute-free `ToNullable`
  is excluded; the predicate also skips `IsServerSideOnly` expressions. The
  earlier `VisitMember` guard is reverted (members always translate to SQL).
- Adds `ToNullable` regression tests (missing `LEFT JOIN` → NULL; non-SQL
  argument → client-side fall-through) to `PreferClientCalculationTests`.

## Follow-up: YDB fix, global toggle, tests, master merge

- `a99ed59c79`: added `[PrimaryKey]` to the `ClientCalcEntity` test model. YDB requires a primary key, so the cross-provider test failed at `CreateLocalTable` without it.
- `bf567d88c3`: added `Configuration.Linq.PreferClientCalculation`, a process-global static toggle that proxies to the default `LinqOptions` (parity with the other `Configuration.Linq.*` toggles). New public API.
- `8785e8d33e`: added regression tests in `PreferClientCalculationTests`. Independent of the option, they assert that SQL-independent client expressions are pre-evaluated into SQL parameters/constants across `WHERE`, `JOIN`, `GROUP BY`, `ORDER BY` and `HAVING`; row-dependent client-only expressions with no SQL mapping throw; server-preferred functions stay server-side; and mixed expressions split. This establishes that the option's scope is the final projection only.
- Merged latest `master` (incl. #5630, #5639). `CompatibilitySuppressions.xml` regenerated and confirmed clean against current master.

## Follow-up: AsNullable handled by the member translator

`73709d2054` — applies the `ToNullable` treatment to `Sql.AsNullable`:

- Drops `AsNullable`'s `[Expression("{0}", 0, CanBeNull = true)]` attribute and the dedicated `ExpressionBuildVisitor` special-case (plus its `Methods` reflection entry). It's now handled in `SqlFunctionsMemberTranslatorBase` as a pass-through that returns the argument's SQL placeholder unchanged.
- `AsNullable<T>` returns `T` (annotates nullability), not `T?` like `ToNullable`, so the handler must not re-type the placeholder via `MakeNullable` — that would invalidate a wrapping `(T?)` cast. The two now use separate handlers; the emitted SQL is unchanged (the bare column).


MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope: opt-in `LinqOptions.PreferClientCalculation` (default `false`) keeps computed final-projection expressions client-side during materialization; server-side-preferred/required expressions (`Builder.PreferServerSide`) and set projections (`ForSetProjection`) still translate to SQL. Projection-gating logic, `ConfigurationID` plumbing, and test coverage all check out.
- [x] The `build` / `Pack Solution for Nuget` checks are **red** on the head commit — see BLK001 for the cause.

### Blockers (BLK)
- [x] **BLK001** — `build` check fails: `LinqOptions` record signature change is an unsuppressed ApiCompat breaking change
  Why: Appending `PreferClientCalculation` to the `LinqOptions` positional record changes the generated primary constructor and `Deconstruct` signatures. ApiCompat compares the built package against the shipped `linq2db 6.0.0` baseline and reports `CP0002` on all five TFMs (`net462`, `netstandard2.0`, `net8.0`, `net9.0`, `net10.0`): `Member 'LinqToDB.LinqOptions.LinqOptions(bool, …, bool)' exists on [Baseline] … but not on …` (same for `Deconstruct(...)`). This fails `Pack Solution for Nuget` and the whole `build` check; the PR cannot merge while it is red.
  Fix: Regenerate `Source/LinqToDB/CompatibilitySuppressions.xml` by rebuilding the package with `dotnet pack -p:ApiCompatGenerateSuppressionFile=true` under `Source/`, then commit the new suppression entries (the same intentional-breaking-change suppression pattern already used for prior record-parameter additions). Don't hand-edit the file. The build log confirms: *"API breaking changes found. If those are intentional, the APICompat suppression file can be updated by rebuilding with '/p:ApiCompatGenerateSuppressionFile=true'."*

### Minor (MIN)
- [x] **MIN001** — New public API not registered in `PublicAPI.Unshipped.txt`
  Why: The PR adds `LinqOptions.PreferClientCalculation` (get/init), `DataOptionsExtensions.WithPreferClientCalculation`, and `DataOptionsExtensions.UsePreferClientCalculation`, and changes the `LinqOptions` constructor / `Deconstruct` signatures — but none of the per-TFM `Source/LinqToDB/PublicAPI/<tfm>/PublicAPI.Unshipped.txt` files are updated. Not a CI blocker today (`RunApiAnalyzersDuringBuild` defaults to `false`, so RS0016/RS0017 don't run in the normal build), but the release-time public-API reconciliation will surface these as pending diagnostics. The sibling `PreferExistsForScalar` shows the expected entries (`LinqToDB.LinqOptions.PreferExistsForScalar.get -> bool`, etc.).
  Fix: Add the new members and the updated `LinqOptions(...)` / `Deconstruct(...)` signatures to the `PublicAPI.Unshipped.txt` files (IDE quick-fix "Add to public API", or a Release build with the public-API analyzer enabled).

## Baselines
Delta: [linq2db.baselines PR #1913](https://github.com/linq2db/linq2db.baselines/pull/1913) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5604)

172 baseline files changed: 170 new + 2 modified, 0 deleted.

### `PreferClientCalculationTests` — new suite (170 files, expected)
All 170 additions are baselines for the new fixture, parameterized `(provider, preferClientCalculation: bool)`. Verified against the files: the `False` variant emits the computed expression in SQL; the `True` variant strips it and selects only the underlying columns. Confirmed across the SQLite providers and consistent for the 57-provider `ProjectionResultsMatchAcrossProviders` sweep.
- `BinaryArithmeticProjection`: `False` → `[e].[Value1] + 12345`; `True` → `[e].[Value1]` only.
- `ClientSideAllowedFunctionFollowsOption`: `False` → `ABS([e].[Value1])`; `True` → `[e].[Value1]` only.
- `MixedServerAndClientExpression`: `True` keeps the server-forced `ABS([e].[Value2])` column and splits only the arithmetic — pins the `Builder.PreferServerSide` gate.
- `ServerSidePreferredFunctionStaysInSql` / `SetProjectionStaysInSql` / `UnaryNegationProjection`: identical SQL for both option values — confirms server-annotated expressions are never pushed client-side.

### Cross-provider anomaly — verify before merging the baselines PR
- `DataContextTests.DontCloseAfterUse_DataConnection` and `DontCloseAfterUse_DataContext` each gained one extra `SELECT [t1].[FirstName], … FROM [Person] [t1]` block (verified in the actual `.sql` files at line 1087), **only** on `SqlServer.Contained` — no other SqlServer variant changed. These tests don't use computed projections and the option defaults `false`, so this is unrelated to the feature — it looks like an accumulation/ordering artifact of these stateful "don't close after use" baselines under `SqlServer.Contained`. Worth confirming it's known-flaky before merging baselines PR #1913 rather than a real side effect.

sdanyliv:
**BLK001**: fixed in 0dd105e090. Regenerated `Source/LinqToDB/CompatibilitySuppressions.xml` via `dotnet pack -p:ApiCompatGenerateSuppressionFile=true`, adding the `CP0002` suppressions for the `LinqOptions` primary ctor + `Deconstruct` signature change (all 5 TFMs), following the existing record-parameter suppression pattern. The `build` check is re-running.

**MIN001**: addressed in the same commit. Added `PreferClientCalculation` (get/init), `With`/`UsePreferClientCalculation`, and the updated ctor/`Deconstruct` signatures (with `*REMOVED*` for the prior ones) to `PublicAPI.Unshipped.txt`.

For context, the PR also gained `baf79f0541` after this review: `Sql.ToNullable` was being pulled client-side under the option and collapsing a SQL NULL to `default(T)`. It now translates server-side via `SqlFunctionsMemberTranslatorBase`, with regression tests. See the Follow-up commit section in the PR description.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
```diff
@@ -867,14 +867,20 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
 					return translatedExposed;
 				}
 
-				if (TranslateMember(BuildContext, node, out var translatedMember))
+				// Honor PreferClientCalculation only for a mapped function whose attribute does not declare a NULL-propagating
+				// result (e.g. Sql.ToNullable is IfAnyParameterNullable) — pulling such a function client-side would turn a
+				// SQL NULL into a .NET default. Structural LINQ methods (aggregates) carry no attribute and translate as usual.
```
MaceWindu: **Minor · MIN002** — Stale comment. It still describes `Sql.ToNullable` as protected via an `IfAnyParameterNullable` attribute, but this PR removed that attribute from `ToNullable` (`Sql.cs`). With no attribute, `GetExpressionAttribute` returns null, so `MappedFunctionAllowsClientCalculation(ToNullable)` returns **false** — `ToNullable` is kept server-side precisely because it carries *no* attribute, the opposite of what the example says. A future reader tracing why `ToNullable` stays server-side will look for an `IfAnyParameterNullable` attribute that no longer exists.

Fix: describe the actual mechanism — a mapped function is allowed client-side only when it carries an `[Expression]` attribute; `Sql.ToNullable` deliberately carries none (translated server-side via `SqlFunctionsMemberTranslatorBase` instead), so it fails the check and keeps translating.

```suggestion
				// Honor PreferClientCalculation only for a mapped function that carries an [Expression] attribute
				// (MappedFunctionAllowsClientCalculation). Functions like Sql.ToNullable deliberately carry no attribute and
				// are translated server-side by SqlFunctionsMemberTranslatorBase, so they fail the check and keep translating;
				// structural LINQ methods (aggregates) likewise have no attribute and translate as usual.
```

— ✓ Fixed in 8944966b

Tests/Linq/Linq/PreferClientCalculationTests.cs
```diff
@@ -0,0 +1,255 @@
...
+			public static readonly ClientCalcEntity[] Seed =
+			[
+				new() { Id = 1, Value1 = 10, Value2 = 100, Name = "Alpha" },
+				new() { Id = 2, Value1 = 20, Value2 = 200, Name = "Beta"  },
+				new() { Id = 3, Value1 = 30, Value2 = 300, Name = "Gamma" },
+			];
+		}
+
+		// Mapped SQL functions (ABS). Under the arithmetic-core scope a method call always translates to SQL,
+		// even when client calculation is preferred — regardless of the PreferServerSide flag.
```
MaceWindu: **Nit · NIT001** — This comment claims a mapped method call "always translates to SQL … regardless of the PreferServerSide flag", but `NonServerPreferredFunctionMovesClientSide` below demonstrates the opposite: `PreferClient` — a mapped ABS function with `PreferServerSide = false` — moves client-side when `PreferClientCalculation` is on. The comment contradicts the semantics the tests beneath it exercise.

Fix: state that `PreferServerSide` is what governs whether the function stays server-side.

```suggestion
		// Mapped SQL functions (ABS). PreferServerSide controls whether the function stays server-side:
		// PreferServerSide = true keeps it in SQL even when client calculation is preferred; PreferServerSide = false
		// lets it move client-side when client calculation is preferred.
```

— ✓ Fixed in 8944966b

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope: opt-in `LinqOptions.PreferClientCalculation` (default `false`) keeps computed final-projection expressions (arithmetic / conditional / unary / mapped members & methods that don't prefer server-side) client-side during materialization instead of emitting extra SQL columns; server-side-preferred/required (`Builder.PreferServerSide`) and set projections still translate. The gate is conservative (`PreferServerSide` recurses, so any server-preferred leaf keeps the whole node in SQL) — confirmed by `MixedServerAndClientExpression`.
- [x] Prior **BLK001** (ApiCompat `CP0002` on the `LinqOptions` record signature) and **MIN001** (`PublicAPI.Unshipped.txt`) are resolved in `0dd105e0` — `CompatibilitySuppressions.xml` (+70) and `PublicAPI.Unshipped.txt` (+8) are both in the diff. The currently-red `build` check is a transient build-server file lock (`CS2012: Cannot open 'linq2db.dll' … being used by another process … VBCSCompiler`), not a code error; `test-all` passes. A re-run should go green.
- [x] API baselines refreshed in this PR. New public surface under `LinqToDB`: `LinqOptions.PreferClientCalculation` (record property), `DataOptionsExtensions.WithPreferClientCalculation` / `UsePreferClientCalculation`. The `LinqOptions` positional record's primary constructor + `Deconstruct` change is the intentional record-parameter append — suppressed via `CP0002` across all 5 TFMs, matching the existing `PreferExistsForScalar` pattern. No major-milestone gate needed for an additive, suppressed parameter.
- [ ] The default-`false` option doesn't change SQL, but the follow-up `Sql.ToNullable` server-side move (`baf79f0541`) and the projection-visitor restructuring are **always active** and shift 46 pre-existing baselines (see **Baselines**). The PR body's "existing behavior and SQL baselines are unchanged" is true for the *option* but stale w.r.t. the follow-up commit. Two of those changes warrant confirmation (`CoalesceColumnSelection_AsFieldInit` extra column; `DontCloseAfterUse` extra SELECT).

## Out-of-scope observations

Surfaced during review but fall outside this PR's scope. Not findings — included as FYI.

- **`Sql.AsNullable<T>` may have the same client-side NULL-collapse exposure this PR fixes for `Sql.ToNullable`.** `AsNullable` keeps its `[Expression("{0}", 0, CanBeNull = true)]` attribute, so `MappedFunctionAllowsClientCalculation` returns true for it and, under `PreferClientCalculation=true`, `AsNullable(<value-type column>)` is eligible to be pulled client-side via the same gate. If the inner placeholder materializes as non-nullable in that path, `AsNullable` over an unmatched LEFT JOIN row would collapse to `default(T)` — the exact regression `ToNullableOverMissingLeftJoinReturnsNull` pins for `ToNullable`. Couldn't reproduce within budget, and the whole `As*/To*` family was clearly considered (`AsNotNull`/`AsNotNullable`/`ToNotNull`/`ToNotNullable` left with attributes intentionally) — so this is a verification request, not an asserted defect. Worth either an `AsNullable`-over-unmatched-LEFT-JOIN case under the option, or a one-line note on why `AsNullable` is safe where `ToNullable` was not (its result type is already `T?`, so the placeholder reads nullable).
- **`Sql.ToNullable` NULL-preservation is verified only on SQLite.** `ToNullableOverMissingLeftJoinReturnsNull` / `ToNullableOverNonSqlArgumentEvaluatesClientSide` are `[IncludeDataSources(AllSQLite)]`; the cross-provider sweep has no `ToNullable` case. The new translator is inherited uniformly (no provider overrides `CreateSqlFunctionsMemberTranslator`), so a per-provider gap is unlikely, but one `ToNullable`-over-LEFT-JOIN assertion in the `[DataSources]` sweep would close it cheaply.
- **`WithPreferClientCalculation(this LinqOptions)` has no direct test.** Every test drives the option through `UsePreferClientCalculation(this DataOptions)`. This matches house convention (none of the 15 sibling `WithXxx(this LinqOptions)` extensions are individually tested), so it's not a finding against this PR — noted only if the team wants per-surface coverage for the whole family.

## Baselines
Delta: [linq2db.baselines PR #1913](https://github.com/linq2db/linq2db.baselines/pull/1913) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5604)

**210 added, 46 modified, 0 deleted** (verified against the branch base — comparing against current `linq2db.baselines` master shows extra drift from PRs merged since the branch was cut, which is not part of this PR).

### New `PreferClientCalculationTests` suite — 210 files, expected
Parameterized `(provider, preferClientCalculation: bool)`. The `False` variant emits the computed expression in SQL; the `True` variant strips client-computable arithmetic/unary and keeps server-preferred functions. Verified across providers: `BinaryArithmeticProjection` False → `[e].[Value1] + 12345`, True → `[e].[Value1]`; `MixedServerAndClientExpression` True keeps `ABS([e].[Value2])` and splits the arithmetic; `GroupByAggregateStaysServerSide` keeps `SUM(...)` in both variants. Consistent.

### Modified pre-existing baselines (46 files) — mostly expected
- **`ProjectionTests.GroupByWithToNullable`** (7: Oracle 11–23 Managed, SqlCe) — GroupBy key column alias dropped (`"PersonID" as "id"` → `"PersonID"`). Direct consequence of routing `ToNullable` through the translator; no semantic change.
- **`OrderByTests.OrderByUnionImmutable`** (6: 4×SQLite, 2×ClickHouse) — constant order element changes from a parameter / `toInt32(1)` to `Abs(1)` (one fewer parameter). No regression.
- **`SelectTests.Issue1788Test1` / `Issue1788Test2` / `SequentialAccessTest`** (3, SqlCe) — redundant generated column aliases removed.

### Cross-provider anomalies — confirm before merging baselines PR #1913
- **`EnumerableInQuery.CoalesceColumnSelection_AsFieldInit`** (28: all SqlServer + PostgreSQL versions) — gains an 8th projected column `ColorId0` = `Coalesce(CASE WHEN [a_Color_1].[Id] IS NULL THEN NULL ELSE [t2].[ColorId] END, 0)` in the VALUES/UNION rowset (verified in `…(PostgreSQL.13).sql`; SqlServer 2012+ uses the `IIF` form, older + PostgreSQL use `CASE WHEN`). This test is unrelated to the option, so the extra projected column is a real SQL change introduced by the projection restructuring — confirm it's intentional and not an unintended duplicate-column emission.
- **`DataContextTests.DontCloseAfterUse_DataConnection` / `DontCloseAfterUse_DataContext`** (SqlServer.Contained only) — each gains a trailing extra `SELECT [t1].[FirstName], … FROM [Person] [t1]` block (verified at line 1087). Only on `SqlServer.Contained`, not `Contained.MS` — looks like a stateful accumulation/ordering artifact of these "don't close after use" baselines rather than a deterministic feature effect (the prior review flagged the same). Confirm known-flaky vs. a real extra query execution.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
sdanyliv: Fixed in 52dbf28794. The comment no longer describes ToNullable as protected by IfAnyParameterNullable; it now states ToNullable carries no [Expression] attribute, so MappedFunctionAllowsClientCalculation returns false and it stays server-side.

Tests/Linq/Linq/PreferClientCalculationTests.cs
sdanyliv: Fixed in 52dbf28794. The comment now states PreferServerSide controls whether the function stays server-side (PreferServerSide=false moves it client-side under PreferClientCalculation).

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope: opt-in `LinqOptions.PreferClientCalculation` (default `false`) keeps computed final-projection expressions client-side during materialization; server-side-preferred (`Builder.PreferServerSide`) and set projections still translate. Follow-up commits move `Sql.ToNullable` server-side (NULL-preservation) and dedup VALUES columns in `EnumerableContextDynamic`. Gate logic, translator wiring, and test coverage check out.
- [x] Prior **BLK001** (ApiCompat `CP0002` on the `LinqOptions` record signature) and **MIN001** (`PublicAPI.Unshipped.txt`) are resolved in `0dd105e0`; the stale-comment items on `ExpressionBuildVisitor.cs` and the test file are resolved in `52dbf28794`. `CompatibilitySuppressions.xml` (+70) and `PublicAPI.Unshipped.txt` (+8) are present.
- [x] No new findings across correctness / SQL-provider / API-test review. The `PreferClientCalculation` gate correctly recurses `PreferServerSide` / `ServerSideOnly`, so a single server-side leaf pins the enclosing op in SQL. The `Sql.ToNullable` translator is registered on the shared `SqlFunctionsMemberTranslatorBase` (inherited by every provider, no override) and re-types the placeholder to `T?` while preserving the underlying SQL NULL — verified against the regression tests.
- [ ] The PR is **`mergeable: CONFLICTING`** — it needs a merge/rebase with `master` before it can merge. Not a code issue.

## Out-of-scope observations

Surfaced during review but fall outside this PR's scope. Not findings — included as FYI.

- **VALUES-column type/key asymmetry in `EnumerableContextDynamic`.** The dedup key is now the member's own column descriptor (`keyDescriptor`), but the generated field's `dbDataType` is still derived from the ambient `currentDescriptor`. Two same-column references reached under differing ambient descriptors now collapse to one column whose type is fixed by whichever reference built first. Because both target the same physical column the type is expected to be consistent, so this is not a defect — noted only as a latent surprise if a future caller deduplicates same-column references whose ambient descriptors imply different SQL types.
- The prior review's verification requests still stand and are unaddressed: the `Sql.ToNullable` NULL-preservation fix is verified only on SQLite (the cross-provider sweep has no `ToNullable` case), and `Sql.AsNullable` retains its `[Expression]` attribute (potential same-family client-side exposure under the option). Both are cheap to close with one non-SQLite `ToNullable`-over-LEFT-JOIN assertion / an `AsNullable` case, or a one-line note on why `AsNullable` is safe.

## Baselines
Delta: [linq2db.baselines PR #1935](https://github.com/linq2db/linq2db.baselines/pull/1935) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5604)

**204 added / 106 modified / 0 deleted** (verified against the merge-base).

### New `PreferClientCalculationTests` suite — 204 files, expected
Parameterized `(provider, preferClientCalculation: bool)`. `False` emits the computed expression in SQL; `True` strips client-computable arithmetic/unary and keeps server-preferred functions. Consistent across providers.

### Modified pre-existing baselines (106) — all explained by the follow-up commits
- **`SelectTests.SequentialAccessTest`** (58 providers) — a duplicate `PersonID` projection column is removed. Direct effect of the VALUES-column dedup fix; beneficial.
- **`EnumerableInQuery.CoalesceColumnSelection_AsFieldInit`** (28: all SqlServer + PostgreSQL) — the inline VALUES rowset drops 3 redundant duplicate columns (`ColorName0`, `StyleName0`, `StyleName1`), going from 7 projected columns to 4. The dedup fix collapsing repeated references to the same `[a_Color].[Name]` / `[a_Style].[Name]`; beneficial, no semantic change.
- **`ProjectionTests.GroupByWithToNullable`** (7: Oracle 11–23, SqlCe) — GroupBy key column alias dropped. Consequence of routing `ToNullable` through the translator.
- **`OrderByTests.OrderByUnionImmutable`** (7: 4×SQLite, 3×ClickHouse) — constant order element becomes `Abs(1)` (one fewer parameter on SQLite). No regression.
- **`AssociationTests.Issue3260Test`** (4: SQLite) — `SUM(CASE … ELSE 0 END)` → `ELSE CAST(0 AS Decimal)`. The ELSE literal now matches the SUM's decimal type; semantically equivalent.
- **`SelectTests.Issue1788Test1` / `Issue1788Test2`** (2: SqlCe) — redundant column aliases removed.

### Cross-provider anomalies
None. The `DontCloseAfterUse_*` extra-SELECT change that earlier reviews flagged on `SqlServer.Contained` no longer appears in the regenerated baselines.

MaceWindu:
stale api baselines

Source/LinqToDB/Sql/Sql.cs
```diff
@@ -89,8 +89,10 @@ public static T AsNotNullable<T>(T value)
 			return value;
 		}
 
+		// Pure nullability widener — handled by SqlFunctionsMemberTranslatorBase so it always stays server-side.
+		// Deliberately carries no [Expression] attribute: that would let projection client-calculation
+		// (LinqOptions.PreferClientCalculation) pull it client-side, collapsing a SQL NULL to default(T) before the widen.
```
MaceWindu, line 96: **Major · MAJ001** — `Sql.AsNullable<T>` has the same client-side NULL-collapse exposure this PR just fixed for `Sql.ToNullable`. `AsNullable` (`Sql.cs:72-73`) keeps `[Expression("{0}", 0, CanBeNull = true)]` with no `PreferServerSide`/`ServerSideOnly` marker, so `MappedFunctionAllowsClientCalculation` returns true and `PreferClientCalculation(node)` pulls it client-side under the new option. `AsNullable<T>` returns `T` (not `T?`), so for a value-type column over an absent LEFT JOIN row the client materializes `default(T)` instead of NULL — the exact regression `ToNullableOverMissingLeftJoinReturnsNull` pins for `ToNullable`. Reachable only when `PreferClientCalculation = true` (introduced by this PR); not exercised by the new fixture.

Fix: give `AsNullable<T>` the same treatment — register a server-side translator in `SqlFunctionsMemberTranslatorBase` and drop the `[Expression]` attribute, or mark it `PreferServerSide = true`. Add an `AsNullable`-over-missing-LEFT-JOIN regression asserting null under both option states. If judged out of scope, document why `AsNullable` is safe.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope: opt-in `LinqOptions.PreferClientCalculation` (default `false`) keeps computed final-projection expressions client-side during materialization; server-side-preferred (`Builder.PreferServerSide`) and set projections still translate. Follow-ups route `Sql.ToNullable` server-side (NULL preservation) and dedup VALUES columns in `EnumerableContextDynamic`; latest commit adds a `[PrimaryKey]` to the test entity for YDB. Gate logic, translator wiring, dedup re-keying, and test coverage all check out.
- [x] No new correctness or SQL/provider findings. The `PreferClientCalculation` predicate recurses `PreferServerSide`/`ServerSideOnly`, so a single server-side leaf pins the enclosing op in SQL; the `ToNullable` translator (`SqlFunctionsMemberTranslatorBase`, inherited by every provider, no override) re-types the placeholder to `T?` while preserving the SQL NULL.
- [x] Public API surface registered: `LinqOptions.PreferClientCalculation`, `DataOptionsExtensions.WithPreferClientCalculation`/`UsePreferClientCalculation`, and the positional-record primary-ctor + `Deconstruct` growth are all in `PublicAPI.Unshipped.txt` with XML docs. Additive on a minor (6.4.0) — no major-gate.
- [x] **Stale `CompatibilitySuppressions.xml`.** The committed file adds 10 `CP0002` suppressions (5 TFMs × ctor + `Deconstruct`) targeting the **unchanged** 15-bool compat-shim `LinqOptions.#ctor(...)`/`Deconstruct(...)` signature — not the synthesized primary ctor this PR actually grows. A clean regen against current `master` produces **zero** `LinqOptions` suppressions (the optional-param append is no longer flagged), so the committed entries are spurious/stale and should be regenerated to match `master`. Not build-breaking, but it's the divergence behind the "stale api baselines" note.

## Baselines
Delta: [linq2db.baselines PR #1944](https://github.com/linq2db/linq2db.baselines/pull/1944) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5604)

**210 added / 108 modified / 0 deleted** (verified independently against the merge-base, three-dot).

### New `PreferClientCalculationTests` suite — 210 files, expected
Parameterized `(provider, preferClientCalculation: bool)`. `False` emits the computed expression in SQL; `True` strips client-computable arithmetic/unary and keeps server-preferred functions (`ABS`, aggregates, set ops, LEFT JOIN all stay regardless of the flag). Consistent across providers (IIF vs CASE WHEN, `negate()` on ClickHouse, etc. are routine dialect differences).

### Modified pre-existing baselines (108) — all explained by the always-active follow-ups
- **`SelectTests.SequentialAccessTest`** (61 providers) — duplicate `[p].[PersonID]` column removed (VALUES-column dedup). Beneficial.
- **`EnumerableInQuery.CoalesceColumnSelection_AsFieldInit`** (28: all SqlServer + PostgreSQL) — inline VALUES rowset deduped from 7 → 4 columns. Beneficial, no semantic change.
- **`ProjectionTests.GroupByWithToNullable`** (8: Oracle 11–23, SqlCe, YDB) — GroupBy key alias dropped; consequence of routing `ToNullable` through the translator.
- **`OrderByTests.OrderByUnionImmutable`** (7: 4×SQLite, 3×ClickHouse) — constant order element becomes `Abs(1)` (one fewer parameter). No regression.
- **`SelectTests.Issue1788Test1`/`Issue1788Test2`** (SqlCe, YDB) — redundant `ToNullable` projection aliases removed.

### Cross-provider anomalies
None. The `DontCloseAfterUse_*` extra-SELECT on `SqlServer.Contained` that earlier reviews flagged no longer appears in the regenerated baselines (verified: no `DontCloseAfterUse` files in the delta).

igor-tkachev (changes requested):
Thanks for the work on this. The direction is useful, but I don't think the current implementation matches the intended semantics of `PreferClientCalculation` yet.

`PreferClientCalculation` should be a global LINQ translation mode: when an expression can be evaluated on the client without depending on server-side row data, linq2db should pre-evaluate it and pass the result to SQL as a parameter/constant. This should apply consistently across the query pipeline: `WHERE`, `JOIN`, `GROUP BY`, `ORDER BY`, `HAVING`, projections, etc.

The current implementation is still projection/materialization-oriented. It keeps computed final-projection expressions client-side, but it doesn't establish the global rule "SQL-independent client expression -> pre-evaluated SQL parameter" for non-projection clauses.

The important semantic distinction is:

```csharp
where p.ID == ClientOnlyDouble(10)
```

should be translated as a SQL predicate with a pre-evaluated parameter value.

But:

```csharp
where ClientOnlyDouble(p.ID) == 20
```

must not become client-side post-filtering. Since it depends on row data, it must either translate to SQL if a mapping exists, or fail as before.

Please add implementation and tests for this global behavior, at least for:

- SQL-independent client expression in `WHERE` -> parameter/constant;
- same class of expression in `JOIN`, `GROUP BY`, `ORDER BY`, and `HAVING`;
- row-dependent client-only expression -> still fails unless it has SQL translation;
- server-side-only / server-preferred expressions -> still stay SQL-side;
- mixed expressions such as `p.ID == ClientOnly(10) + 1` -> client part pre-evaluated, row-dependent SQL part preserved.

Also, the existing `Sql.AsNullable<T>` review comment still looks valid and should be addressed. It appears to have the same NULL-collapse exposure that was fixed for `Sql.ToNullable<T>` under this option.

Finally, please regenerate `CompatibilitySuppressions.xml` against current `master`; the committed suppressions appear stale.

igor-tkachev:
One important clarification to my review: this global client pre-evaluation behavior is not a new semantic direction. It matches the default behavior users had in previous linq2db versions. The option should preserve/restore that behavior in an explicit, configurable form.

Source/LinqToDB/Sql/Sql.cs
MaceWindu: **Fixed in 874703830.**

`Sql.AsNullable<T>(T)` only annotates nullability and returns its argument unchanged. Under `PreferClientCalculation` it was pulled client-side as a method call whose non-nullable argument read collapsed a SQL `NULL` (e.g. from a missed `LEFT JOIN`) to `default(T)` before the surrounding `(int?)` cast — so `(int?)Sql.AsNullable(col)` materialized `0` instead of `null`. (Plain `(int?)col` was unaffected.)

Fix: unwrap `Sql.AsNullable` to its argument when the option pulls it client-side, so the surrounding read keeps the column's own nullability. The server-side path (its `[Expression]` attribute, option off) is unchanged, so existing SQL and baselines are unaffected.

Added `AsNullableOverMissingLeftJoinReturnsNull` (red→green); the full `PreferClientCalculationTests` fixture passes on SQLite (Classic + MS).

MaceWindu:
Follow-up commit `874703830` addresses the `Sql.AsNullable` review comment (MAJ001).

**Root cause:** `Sql.AsNullable<T>(T)` only annotates nullability and returns its argument unchanged. Under `PreferClientCalculation` it was pulled client-side as a method call whose non-nullable argument read collapsed a SQL `NULL` (e.g. from a missed `LEFT JOIN`) to `default(T)` before the surrounding `(int?)` cast — `(int?)Sql.AsNullable(col)` yielded `0` instead of `null`. Plain `(int?)col` was unaffected, which is why the nullable-inside form `Sql.AsNullable((int?)col)` never reproduced it.

**Fix:** unwrap `Sql.AsNullable` to its argument when the option pulls it client-side (it is a pure passthrough). The server-side path (its `[Expression]` attribute, option off) is unchanged — existing SQL and baselines are unaffected.

**Verified** on SQLite (Classic + MS): the new `AsNullableOverMissingLeftJoinReturnsNull` test goes red→green, and the full `PreferClientCalculationTests` fixture passes (56/56).

sdanyliv:
Thanks for the review. Addressed below.

**On `PreferClientCalculation` as a global rule.** The "SQL-independent client expression → pre-evaluated parameter/constant" behavior already happens by default across non-projection clauses, independent of this option — `HandleValue` folds any row-data-free expression via `CanBeEvaluatedOnClient` → `ParametersContext.BuildParameter`. Verified on SQLite, and **the option on/off produces identical SQL**:

| LINQ (in `WHERE`) | SQL |
|---|---|
| `p.ID == ClientOnlyDouble(10)` | `WHERE ID = @p` (pre-evaluated) |
| `ClientOnlyDouble(p.ID) == 20` | throws `… could not be converted to SQL` |
| `p.ID == ClientOnlyDouble(10) + p.Value` | `WHERE ID = @p + Value` (client part folded, column kept) |
| `PreferServer(p.Value) == 20` | `WHERE ABS(Value) = 20` (server-preferred stays SQL) |

So `p.ID == ClientOnlyDouble(10)` → parameter, `ClientOnlyDouble(p.ID) == 20` → fails, and the mixed case splits — exactly the semantics requested. `PreferClientCalculation` does not drive this; its scope is the final **projection** (whether a row-dependent computed column is materialized client-side vs emitted as a SQL column).

**Tests added** in `PreferClientCalculationTests` locking in the rule across `WHERE / JOIN / GROUP BY / ORDER BY / HAVING`: SQL-independent → pre-evaluated; row-dependent client-only → throws; server-preferred (`PreferServerSide`) → stays SQL; mixed → split. A client-preferred (`PreferServerSide = false`) function with a row argument also stays SQL inside a predicate — confirming the option is projection-only. Each is parameterized over the option to assert it doesn't change the behavior.

**`Sql.AsNullable<T>`** — fixed in `874703830f` (same server-side nullability-widener handling as `Sql.ToNullable<T>`) with a regression test.

**`CompatibilitySuppressions.xml`** — regenerated (`8944966b8b`) and reconciled in the latest `master` merge; I'll confirm it's clean against current master via the ApiCompat pass.

Source/LinqToDB/Internal/Linq/Builder/EnumerableContextDynamic.cs
```diff
@@ -315,7 +322,7 @@ bool TryGetExpandedRows([NotNullWhen(true)] out Expression[]? expandedRows, [Not
 			var field = new SqlField(dbDataType, fieldName, true);
 			var fieldPlaceholder = ExpressionBuilder.CreatePlaceholder(this, field, currentPath);
 
-			_fieldsMap.Add((currentPath, currentDescriptor, fieldPlaceholder));
+			_fieldsMap.Add((currentPath, keyDescriptor, fieldPlaceholder));
```
MaceWindu, line 325: **Suggestion · SUG001** — Possible descriptor inconsistency in the VALUES-column dedup, flagged for your awareness: the field's declared DB type (line 320, `currentDescriptor?.GetDbDataType(true)`) derives from the ambient `currentDescriptor`, while the values themselves are built with the column's own descriptor (`ownColumnDescriptor ?? currentDescriptor`, line 271). After the dedup keys two references to the same column on `ownColumnDescriptor` (this line), the surviving column's declared type is fixed by whichever ambient descriptor created it first, which can in principle differ from the column's own type.

I could **not** reproduce an actual defect: projecting the same column both bare (int) and inside a string expression on PostgreSQL yields a single, correctly-typed `int` VALUES column — the string use (`ToString()`) is applied client-side, so the placeholder's ambient stays the column's own type and the divergence never surfaces. I added `Tests.Linq.EnumerableInQuery.CoalesceColumnSelection_DedupKeepsColumnOwnType` to pin that current behavior. Raising it only in case a *server-side* string/format scenario over a deduped column could hit the split; if you're confident it can't, this is safe to dismiss. Otherwise deriving `dbDataType` from `keyDescriptor ?? currentDescriptor` at line 320 would make it consistent with line 271.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope: opt-in `LinqOptions.PreferClientCalculation` (default `false`) keeps computed final-projection expressions client-side during materialization, with a `Configuration.Linq.PreferClientCalculation` global toggle proxying the default options. The gate fires only for `BuildPurpose.Expression`; `WHERE` / `JOIN` / `ORDER BY` / `GROUP BY` keys build with `BuildPurpose.Sql`, so the option does not pull filtering/ordering client-side. Server-side-preferred (`PreferServerSide` / `ServerSideOnly`) and set projections still translate. Gate logic, the `ToNullable` / `AsNullable` server-side NULL-preservation, and the VALUES-column dedup all check out — no correctness or SQL/provider defects found.
- [x] `CompatibilitySuppressions.xml` is clean against current `master` (not present in the diff at HEAD).
- [x] New public surface (`LinqOptions.PreferClientCalculation`, `DataOptionsExtensions.WithPreferClientCalculation` / `UsePreferClientCalculation`, `Configuration.Linq.PreferClientCalculation`) carries XML docs; additive on a minor (6.4.0).
- [x] Added a regression test (`Tests.Linq.EnumerableInQuery.CoalesceColumnSelection_DedupKeepsColumnOwnType`) pinning the VALUES-column dedup typing on PostgreSQL/SqlServer — see SUG001 below.

## Prior-review audit
- **Fixed** · igor-tkachev — "add global SQL-independent pre-evaluation + tests for `WHERE` / `JOIN` / `GROUP BY` / `ORDER BY` / `HAVING`": regression tests covering each clause plus the row-dependent-throws, server-preferred-stays-in-SQL, and mixed-expression cases were added in `8785e8d3` and pass; they show the pre-evaluation holds independent of the option. Whether that meets the intent of making `PreferClientCalculation` a global translation *mode* (vs. demonstrating the already-present behavior) is a design call left to you.
- **Fixed** · igor-tkachev — "`Sql.AsNullable` has the same NULL-collapse exposure as `Sql.ToNullable`": addressed in `874703830` — `AsNullable` is unwrapped to its argument when the option pulls it client-side, preserving the column's own nullability; `AsNullableOverMissingLeftJoinReturnsNull` pins it. (resolved thread)
- **Fixed** · igor-tkachev — "regenerate stale `CompatibilitySuppressions.xml`": regenerated against `master`; no `LinqOptions` suppression diff remains at HEAD.

## Baselines
Delta: [linq2db.baselines PR #1944](https://github.com/linq2db/linq2db.baselines/pull/1944) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5604)

**218 added / 108 modified / 0 deleted** (verified independently against the merge-base, three-dot).

### New `PreferClientCalculationTests` suite — 218 added
The parameterized `ProjectionResultsMatchAcrossProviders` covers 61 providers × `{False, True}` = 122 files; ten named sub-tests run on SQLite × 4 × `{False, True}` for the remainder. `False` emits the computed expression in SQL; `True` strips client-computable arithmetic/unary/conditional and keeps server-preferred functions (`ABS`, aggregates, set ops, `LEFT JOIN` nullability) regardless of the flag. Cross-provider shape differences (IIF vs CASE WHEN, ClickHouse `negate()`) are routine dialect variation.

### Modified pre-existing baselines (108) — all explained by the two always-active follow-ups
- **`SelectTests.SequentialAccessTest`** (61 providers) — duplicate `[p].[PersonID]` column removed (VALUES-column dedup).
- **`EnumerableInQuery.CoalesceColumnSelection_AsFieldInit`** (28: SqlServer + PostgreSQL) — inline VALUES rowset deduped 7 → 4 columns; WHERE / outer SELECT unchanged. Same reduction across the IIF (2012+), CASE WHEN (2008), UNION ALL (2005), and PostgreSQL LATERAL forms.
- **`ProjectionTests.GroupByWithToNullable`** (8: Oracle 11–23, SqlCe, YDB) — spurious `as "id"` alias dropped (consequence of routing `ToNullable` server-side).
- **`OrderByTests.OrderByUnionImmutable`** (7: SQLite × 4, ClickHouse × 3) — constant ORDER BY element inlined as `Abs(1)` instead of a bound `@p` parameter.
- **`SelectTests.Issue1788Test1`/`Issue1788Test2`** (SqlCe, YDB) — redundant `ToNullable`-routing aliases removed.

No cross-provider anomalies.

igor-tkachev (dismissed):

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

New opt-in `LinqOptions.PreferClientCalculation` (default `false`; also `DataOptions.UsePreferClientCalculation(...)`, `LinqOptions.WithPreferClientCalculation(...)`, and the global `Configuration.Linq.PreferClientCalculation`). When enabled, computed expressions in the final projection — arithmetic, conditionals, unary operations, and mapped members/methods that don't require server-side evaluation — are evaluated client-side during materialization instead of being emitted as extra SQL columns, restoring the v5 projection behavior. Expressions that require server-side evaluation still translate to SQL.


New opt-in `PreferClientCalculation` option evaluates non-server-required projection expressions client-side (restoring v5 projection behavior).


---
<sub>Generated from commit `8cdf0df`.</sub>

PR #5605, opened 2026-06-10
Add SQL Server GetSqlDecimal overflow protection

igor-tkachev:
## Summary
- Added SQL Server-specific `[GetSqlDecimal]` attribute for opt-in materialization of high-precision SQL Server decimal values through `SqlDecimal`.
- Added shared `SqlServerDecimalUtils.ConvertSqlDecimal` helper for converting overflow-prone `SqlDecimal` values to CLR `decimal`.
- Added scaffold/CLI/T4 option to generate `[GetSqlDecimal]` for SQL Server `decimal` columns whose precision or scale can exceed CLR `decimal` limits.
- Kept protection scoped to explicitly attributed/generated columns; ordinary SQL Server decimal reads retain existing behavior.

## Testing
- `dotnet test --project Tests/Linq/Tests.csproj -c Testing -f net10.0 --settings .runsettings --filter "FullyQualifiedName~Tests.Scaffold.SqlServerDecimalOverflowProtectionTests" --test-progress`
- `dotnet test --project Tests/Linq/Tests.csproj -c Testing -f net10.0 --settings .runsettings --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~Tests.DataProvider.SqlServerTests.GetSqlDecimalAttributeTest|FullyQualifiedName~Tests.DataProvider.SqlServerTests.SqlDecimalValueOverflowTest" --provider SqlServer.2019 SqlServer.2019.MS --test-progress`
- `dotnet build Source/LinqToDB.Scaffold/LinqToDB.Scaffold.csproj -c Release -f net10.0`
- `dotnet build Tests/Base/Tests.Base.csproj -c Testing -f net10.0`
- `dotnet build Tests/FSharp/Tests.FSharp.fsproj -c Testing -f net10.0`
- `git diff --check`


.github/copilot-instructions.md
```diff
@@ -1,5 +1,9 @@
 # Copilot Pull Request Review Rules
 
+## General Guidelines
+- Preserve the user's manual code changes and do not overwrite or revert them.
+- When optimizing reader logic, avoid per-read dictionary lookups in hot paths. For hot-path reader logic in this repository, avoid dictionaries entirely; prefer a self-rewriting lambda/delegate where `ReadSqlServerDecimal` calls the current implementation and the implementation can replace itself after the first overflow.
```
MaceWindu: **Nit · NIT001** — This guidance references `ReadSqlServerDecimal`, which does not exist anywhere in the codebase. The actual reader is the nested `AdaptiveDecimalReader` (entry point `GetDecimal`; the self-rewrite happens in `DefaultReader`). The instruction is inaccurate the moment it lands and would misdirect future edits searching for that method.

Fix: Reference `AdaptiveDecimalReader.GetDecimal` (or drop the example).

```suggestion
- When optimizing reader logic, avoid per-read dictionary lookups in hot paths. For hot-path reader logic in this repository, avoid dictionaries entirely; prefer a self-rewriting lambda/delegate where `AdaptiveDecimalReader.GetDecimal` calls the current implementation and the implementation can replace itself after the first overflow.
```

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDataProvider.cs
```diff
@@ -225,6 +229,116 @@ protected SqlServerDataProvider(string name, SqlServerVersion version, SqlServer
...
+		/// with <c>GetSqlDecimal(i)</c>. Sequential-access users who need this conversion must configure a custom reader
+		/// expression that reads <see cref="SqlDecimal"/> first, so the column is consumed only once.
+		/// </summary>
+		sealed class AdaptiveDecimalReader
+		{
+			const int ClrPrecision = 29;
+
+			readonly Type _dataReaderType;
+
+			Func<DbDataReader, int, decimal> _reader;
```
MaceWindu: **Minor · MIN001** — The `AdaptiveDecimalReader` instance is captured as a closure constant in the reader lambda, so a single instance is shared across every concurrent execution of the cached column reader. `_reader` is a plain (non-volatile) field swapped from `DefaultReader` under no lock. The race is benign — reference writes are atomic and `DefaultReader` is idempotent under overflow (a stale read just re-catches the `OverflowException` and re-swaps) — but the cross-thread sharing and self-rewrite contract is undocumented, so a future reader could wrongly assume single-threaded use.

Fix: Document the contract at the field (optionally mark `_reader` `volatile` to make the publish explicit). No functional change required.

```suggestion
			// Shared across concurrent executions of the cached column reader; the unsynchronized swap is
			// safe because DefaultReader is idempotent under overflow (a stale read just re-catches and re-swaps).
			Func<DbDataReader, int, decimal> _reader;
```

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDataProvider.cs
```diff
@@ -225,6 +229,116 @@ protected SqlServerDataProvider(string name, SqlServerVersion version, SqlServer
...
+						Expression.Call(
+							Expression.Convert(reader, dataReaderType),
+							SqlTypes.GetSqlDecimalReaderMethod,
+							Type.EmptyTypes,
+							index)),
+					reader,
+					index).CompileExpression();
+			}
+
+			static decimal ConvertSqlDecimal(SqlDecimal value)
```
MaceWindu: **Suggestion · SUG001** — When the SQL value's precision exceeds CLR decimal precision, `ConvertSqlDecimal` reduces scale via `ConvertToPrecScale` and returns the first scale that fits — i.e. the materialized `decimal` is a *rounded* approximation of the stored value, silently. The class doc mentions values "representable after reducing scale" but does not state the conversion is lossy.

Fix: Note in the XML doc (or a comment here) that values whose SQL precision exceeds CLR decimal precision are rounded, so the materialized `decimal` can differ from the stored value in its least-significant digits. Documentation only — the rounding is the intended trade-off for the feature.

Tests/Linq/DataProvider/SqlServerTests.cs
```diff
@@ -1420,12 +1420,126 @@ sealed class DecimalOverflow2
...
+			secondRead[0].Value.ShouldBe(expectedDec);
+		}
+
+		[Test]
+		public void OverflowAutoFallbackSequentialTest([IncludeDataSources(TestProvName.SqlServerSequentialAccessMS)] string context)
+		{
+			using var db = GetDataContext(context);
+			using var table = db.CreateLocalTable<SqlServerDecimalOverflowAutoRead>("SqlServerDecimalOverflowAutoRead");
+
+			const string expectedValue1 = "1.2345678901234567890123456789";
```
MaceWindu: **Minor · MIN002** — This test runs against `SqlServerSequentialAccessMS` (sequential access *not* suppressed) and asserts both reads succeed, but the `AdaptiveDecimalReader` XML doc states the catch-and-reread fallback is **not** compatible with `CommandBehavior.SequentialAccess` (after `GetDecimal(i)` reads column `i`, the same column can't be re-read with `GetSqlDecimal(i)`). `Value1` (`1.2345678901234567890123456789` into a `decimal(38,35)` column) should overflow `GetDecimal` and trigger the fallback — yet `test-all` is green on the SqlServer2019/2025 jobs, so the test demonstrably passes under sequential access. That makes the doc's blanket incompatibility claim suspect: either the value does not actually overflow under sequential `Microsoft.Data.SqlClient` (then the test name is misleading and the fallback path isn't exercised here), or a same-ordinal re-read is in fact permitted (the SequentialAccess restriction is about not reading *earlier* columns) and the doc is too strong.

Fix: Reconcile the doc with the test. If the fallback genuinely fires under sequential access, correct the XML-doc paragraph. If it does not (the values fit `GetDecimal` directly), rename the test / assert that the fallback path is not taken, and consider a separate test that forces the fallback under sequential access to pin the real behavior.

Tests/Linq/Mapping/SetUseMappingSchemaTests.cs
MaceWindu, line 1: **Minor · MIN004** — Removing `DecimalNegativeTest` drops coverage this PR doesn't replace. The deleted test exercised the `System.Data.SqlClient` reader path and the fluent `HasPrecision(38).HasScale(37)` mapping route; the new SqlServer tests cover the success path via attribute mapping only. The deletion is mechanically safe (no other references), but converting the test to assert the new successful materialization (instead of the old `OverflowException`) would preserve the `System.Data.SqlClient` + fluent-mapping coverage.

Fix: Convert `DecimalNegativeTest` to assert the value now reads back correctly rather than deleting it, keeping the non-MS-driver and fluent-mapping paths covered.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] Core change is well-scoped: the `ReferenceEquals(expr, _defaultDecimalReaderExpression)` guard limits the adaptive reader to the *default* decimal reader, so user `SetFieldReaderExpression` customizations are respected, and money/smallmoney correctly fall through (their provider-specific type is `SqlMoney`, not `SqlDecimal`). CI `build` and `test-all` are green at HEAD.

## Minor (MIN)
- [ ] **MIN003** — Unrelated changes bundled into a decimal-reader PR.
  Why: The PR adds `Source/TODO.md` (a generic cleanup note about a parameterless `ColumnReaderAttribute` constructor), edits `.github/copilot-instructions.md` with general AI-agent guidance, and adds a `/.claude/skills/` solution folder plus `Source/CodeMetricsConfig.txt` and `Source/TODO.md` to `linq2db.slnx` — none related to the SqlServer decimal-read fallback. Bundling unrelated changes makes the PR harder to review and to revert cleanly.
  Fix: Split the TODO / solution-file / instructions changes into a separate PR (or drop them), keeping this PR scoped to the decimal-reader change and its tests.

## Baselines
Delta: [linq2db.baselines PR #1971](https://github.com/linq2db/linq2db.baselines/pull/1971) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5605) — **78 added / 108 modified / 0 deleted** (counted independently from the baseline files).

- **78 added** — the new SqlServer decimal-fallback test baselines (`SqlDecimalValueOverflowTest`, `OverflowAutoFallbackTest`, `OverflowAutoFallbackSequentialTest` [SQL Server SA provider only, matching the test's `[IncludeDataSources]`], `MoneyTypesReadAfterDecimalFallbackTest`, `OverflowMicrosoftDataSqlClientReadTest` [`.MS` providers only]). Expected and correct — plain INSERT/SELECT shapes against purpose-built tables.
- **108 modified** — six test groups (`SequentialAccessTest` ×61, `CoalesceColumnSelection_AsFieldInit` ×28, `GroupByWithToNullable` ×8, `OrderByUnionImmutable` ×7, `Issue1788Test1` / `Issue1788Test2` ×4). **None of these tests is touched by this PR.** Checked file-by-file: these are SQL-generation differences between this branch and the current `master` baselines (e.g. `CoalesceColumnSelection_AsFieldInit` differs only in projection-column deduplication) — the branch's baselines lag `master`, not changes this PR introduces. Heads-up for whoever merges the baselines: PR #1971 therefore also carries these 108 unrelated modifications; they are not a regression in this PR and warrant care before that baselines PR is merged.

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDataProvider.cs
```diff
@@ -228,6 +232,92 @@ protected SqlServerDataProvider(string name, SqlServerVersion version, SqlServer
 
 		#region Overrides
 
+		public override Expression GetReaderExpression(DbDataReader reader, int idx, Expression readerExpression, Type? toType)
```
MaceWindu: **Blocker · BLK001** — The automatic decimal-read fallback installed here for **every** SqlServer decimal column should be removed. The opt-in `GetSqlDecimalAttribute` this PR also adds is the correct mechanism, and the scaffold detection emits it automatically for overflow-prone SqlServer columns.

Three problems, all imposed on every SqlServer decimal read (not just overflow-prone ones):

1. **`CommandBehavior.SequentialAccess`** — the catch-then-reread-the-same-ordinal approach cannot work under sequential access; this reader's own XML doc admits it.
2. **Wrapped readers (MiniProfiler etc.)** — `BuildFallbackReader` hard-casts the live reader to the concrete `Adapter.DataReaderType` (`Expression.Convert(reader, dataReaderType)`), bypassing linq2db's profiler-safe reader unwrapping. A `ProfiledDbDataReader` (or any `DbDataReader` wrapper) throws `InvalidCastException` the moment the fallback fires; the ordinary `GetDecimal` path worked because it's virtual.
3. **Cost + silent behavior change for everyone** — per-row delegate indirection plus a try frame on the materialization hot path, and overflow changes from `OverflowException` to a silently-rounded value, for all users — none of which they opted into.

`GetSqlDecimalAttribute` is a `ValueConverterAttribute` that reads `SqlDecimal` directly (no reread), so it is SequentialAccess-safe, routes through the standard profiler-safe column-reader machinery, is scoped to the columns that need it, and reuses `SqlServerDecimalUtils.ConvertSqlDecimal`.

Fix: remove the `GetReaderExpression` override, the `AdaptiveDecimalReader` nested class, and the `_defaultDecimalReaderExpression` field/init. Keep `SqlServerDecimalUtils.ConvertSqlDecimal`, `GetSqlDecimalAttribute`, and the scaffold detection. The auto-fallback tests (`OverflowAutoFallbackTest`, `OverflowAutoFallbackSequentialTest`, `MoneyTypesReadAfterDecimalFallbackTest`, `OverflowMicrosoftDataSqlClientReadTest`) and their new baselines go with it; keep a test that reads an overflow value through the `[GetSqlDecimal]` attribute.

Tests/Base/Tests.Base.csproj
```diff
@@ -3,6 +3,12 @@
 	<PropertyGroup>
 		<AssemblyName>linq2db.Tests.Base</AssemblyName>
 		<RootNamespace>Tests</RootNamespace>
+
+		<TestProject>false</TestProject>
```
MaceWindu: **Minor · MIN005** — These Microsoft.Testing.Platform properties (`TestProject` / `IsTestProject` / `IsTestable` / `IsTestingPlatformApplication` / `DisableTestingPlatformServerCapability` — the same block is also added to `Tests/FSharp/Tests.FSharp.fsproj`) are unrelated to the SqlServer decimal fallback, and the build succeeds without them.

Fix: remove them, or add a rationale for why `Tests.Base` / `Tests.FSharp` specifically need them (ideally on their own commit/PR).

Source/LinqToDB.Scaffold/ModelGeneration/ModelGenerator.LinqToDB.cs
```diff
@@ -555,6 +557,19 @@ orderby gr.Key
 
 					c.Attributes.Insert(0, ca);
 
+					if (GenerateSqlServerDecimalOverflowProtection && ShouldUseGetSqlDecimal(c))
```
MaceWindu: **Minor · MIN006** — This T4 emission has no provider guard, unlike the scaffold path: `DataModelLoader.Entities.cs` `ShouldUseGetSqlDecimal` returns false unless `_schemaProvider.DatabaseOptions is SqlServerDatabaseOptions`. `GenerateSqlServerDecimalOverflowProtection` is exposed globally in the shared `LinqToDB.ttinclude` (used by every provider's T4 template), so enabling it from a non-SqlServer T4 template emits an invalid SqlServer-only `[GetSqlDecimal]` on that provider's decimal columns. The two surfaces are also asymmetric in coverage — the scaffold path is tested, the T4 emission path is not. (Low real-world risk: opt-in, default off.)

Fix: gate the T4 emission on a SqlServer signal (matching the scaffold `DatabaseOptions` guard), or document `GenerateSqlServerDecimalOverflowProtection` as valid only in the SqlServer T4 template; add a T4-emission test if the codegen-test infrastructure supports it.

Source/LinqToDB.Scaffold/ModelGeneration/ModelGenerator.LinqToDB.cs
```diff
@@ -26,26 +27,27 @@ public partial class ModelGenerator
...
-		public string SchemaDataContextTypeName           { get; set; } = "DataContext";
-		public bool   GenerateNameOf                      { get; set; } = true;
-		public bool   GenerateTableRegion                 { get; set; } = true;
+		public bool   GenerateDataOptionsConstructors            { get; set; } = true;
+		public bool   GenerateObsoleteAttributeForAliases        { get; set; }
+		public bool   GenerateFindExtensions                     { get; set; } = true;
+		public bool   IsCompactColumns                           { get; set; } = true;
+		public bool   IsCompactColumnAliases                     { get; set; } = true;
+		public bool   GenerateDataTypes                          { get; set; }
+		public bool   GenerateSqlServerDecimalOverflowProtection { get; set; }
```
MaceWindu, line 41: **Suggestion · SUG002** — This new public member is listed in `PublicAPI.Unshipped.txt` but has no XML doc. Note: the `ModelGenerator` class is marked `/// For internal use.` and its entire options block is undocumented, so this matches house style and is fine to dismiss — but the sibling `DataModelOptions.GenerateSqlServerDecimalOverflowProtection` and `ColumnMetadata.UseGetSqlDecimal` both got docs.

Fix: add a summary matching the `DataModelOptions` member.

```suggestion
		/// <summary>
		/// Enables generation of the SQL Server-specific <c>GetSqlDecimalAttribute</c> on <see cref="decimal"/> columns whose precision or scale can exceed CLR <see cref="decimal"/> limits.
		/// </summary>
		public bool   GenerateSqlServerDecimalOverflowProtection { get; set; }
```

Source/LinqToDB.Scaffold/Scaffold/Options/DataModelOptions.cs
```diff
@@ -159,6 +159,16 @@ internal DataModelOptions() { }
 		/// </summary>
 		public bool GenerateScale { get; set; }
 
+		/// <summary>
+		/// Enables generation of SQL Server-specific <c>GetSqlDecimalAttribute</c> for <see cref="decimal"/> entity columns
+		/// with precision or scale outside of CLR <see cref="decimal"/> limits.
+		/// <list type="bullet">
+		/// <item>Default: <see langword="false"/></item>
+		/// <item>In T4 compability mode: <see langword="false"/></item>
```
MaceWindu: **Nit · NIT003** — Typo in new public XML doc: "compability" → "compatibility". Ships into IntelliSense / docfx output.

```suggestion
		/// <item>In T4 compatibility mode: <see langword="false"/></item>
```

Tests/Linq/Scaffold/SqlServerDecimalOverflowProtectionTests.cs
```diff
@@ -0,0 +1,151 @@
...
+		public void EmitsGetSqlDecimalAttributeInGeneratedCode([Values(MetadataSource.Attributes, MetadataSource.FluentMapping)] MetadataSource metadata)
+		{
+			var options = ScaffoldOptions.Default();
+			options.DataModel.GenerateSqlServerDecimalOverflowProtection = true;
+			options.DataModel.Metadata                                   = metadata;
+
+			var source = GenerateSource(
+				options,
+				SqlServerDatabaseOptions.Instance,
+				new Column("SafeDecimal"   , null, new DatabaseType("decimal", null, 29,  0), true, true, true, 0),
```
MaceWindu: **Nit · NIT004** — This column is named `SafeDecimal` but `decimal(29,0)` is not safe — `ExceedsClrLimits(29,0)` is true, so it is marked (contradicting the sibling `MarksSqlServerDecimalColumnsOutsideClrLimits` test, where `SafeDecimal` is correctly `decimal(28,0)`). The mislabel re-encodes the off-by-one the `>= 29` fix corrected, and it is masked here because the assertion only checks `Does.Contain("GetSqlDecimal")` — already satisfied by the `LargePrecision(38,35)` column. Dropping it to `decimal(28,0)` makes the name accurate and makes the assertion actually discriminate (driven solely by `LargePrecision`).

```suggestion
				new Column("SafeDecimal"   , null, new DatabaseType("decimal", null, 28,  0), true, true, true, 0),
```

MaceWindu (changes requested):
> [!IMPORTANT]

## Review notes
- [ ] Headline (BLK001): the runtime auto-reader installed for every SqlServer decimal read should be removed. The opt-in `[GetSqlDecimal]` value-converter attribute this PR also adds — plus the scaffold detection that emits it — is the correct, scoped, SequentialAccess- and profiler-safe mechanism and is retained.
- [ ] Boundary logic verified correct at HEAD: `precision >= 29` in both `SqlServerDecimalUtils.ConvertSqlDecimal` and the scaffold `ExceedsClrLimits` (`precision >= 29 || scale > 28`); the primary discrimination test asserts `decimal(28,0)`→unmarked / `decimal(29,0)`→marked. Money/smallmoney fall-through (they report `SqlMoney`), both SqlClient adapters, and nullable `decimal?` columns all confirmed handled.

## Baselines
Delta: [linq2db.baselines PR #1971](https://github.com/linq2db/linq2db.baselines/pull/1971) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5605) — **78 added / 108 modified / 0 deleted** (counted independently from the baseline files).

- **78 added** — the new SqlServer decimal-fallback test baselines (`SqlDecimalValueOverflowTest`, `OverflowAutoFallbackTest`, `OverflowAutoFallbackSequentialTest`, `MoneyTypesReadAfterDecimalFallbackTest`, `OverflowMicrosoftDataSqlClientReadTest`). Correct as generated — but most back the runtime auto-reader tests and should be removed together with the auto-reader per BLK001; a single attribute-based overflow-read baseline is what should remain.
- **108 modified** — six test groups (`SequentialAccessTest`, `CoalesceColumnSelection_AsFieldInit`, `GroupByWithToNullable`, `OrderByUnionImmutable`, `Issue1788Test1`/`Issue1788Test2`) that this PR does not touch. This is **stale-branch churn**: `baselines/pr_5605` is frozen ~12 days behind `master` and lacks unrelated SQL-generation changes merged since (e.g. projection-column dedup). Not a regression in this PR — but PR #1971 therefore carries these unrelated modifications; regenerate the baselines branch before merging it.

Tests/Base/Tests.Base.csproj
igor-tkachev: > Fix: remove them, or add a rationale for why Tests.Base / Tests.FSharp specifically need them (ideally on their own commit/PR).

.. and wait another 3 months for approval.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- New `[GetSqlDecimal]` attribute for SQL Server `decimal` columns whose precision or scale falls outside the CLR `decimal` range. Reading such a column previously threw `OverflowException`; with the attribute the value is read through SQL Server's own `SqlDecimal` and its scale reduced to fit, which can round away least-significant digits. A value whose magnitude still exceeds `decimal.MaxValue` continues to throw. Opt-in per column, and applied only when the provider is SQL Server. [component: SQL Server]
- Scaffolding can emit `[GetSqlDecimal]` for SQL Server `decimal` columns with precision or scale outside CLR `decimal` limits — `--sqlserver-decimal-overflow-protection` (off by default), or `DataModelOptions.GenerateSqlServerDecimalOverflowProtection` when driving the scaffolder in code. [component: LinqToDB CLI]


New SQL Server `[GetSqlDecimal]` attribute to read `decimal` columns whose precision exceeds the CLR `decimal` range instead of throwing.


---
<sub>Generated from commit `9e3ab11`.</sub>

PR #5610, opened 2026-06-12
Live test-progress heartbeat for long test runs

sdanyliv:
Adds an opt-in live progress heartbeat for long NUnit runs so progress can be
read mid-run (current test, completed/total, pass/fail tally, recent failures)
without scraping console output.

## Test harness
- `Tests/Base/TestProgressReporter.cs` — assembly-level NUnit `ITestAction` writes
  a JSON heartbeat to `.build/.claude/test-progress.<tfm>.<pid>.json` while a run
  is in flight (throttled ~1/s, immediate on failure). `[assembly: TestProgressReporter]`
  wired into Linq, Playground, and EF Core (EF3/8/9/10).
- Opt-in via the `LINQ2DB_TEST_PROGRESS` env var; a no-op when unset, so default
  runs are byte-identical.

## Agent tooling (.claude)
- `/test-progress` skill toggles the trace via `settings.local.json` → `env`.
- `.claude/scripts/test-status.ps1` prints a one-line summary.
- `/test` auto-enables + background-monitors long runs; documented in `testing.md`.

Caveat: under a `--filter`, `total`/`etaSec` reflect the discovered (pre-filter)
count; `completed`/`currentTest` are always accurate.

### DIFF
--- a/.claude/agents/test-runner.md
+++ b/.claude/agents/test-runner.md
@@ -72,6 +72,7 @@ dotnet test <project> --filter "<testPattern>" -c <config>
 ```
 
 Notes:
+- **Live progress heartbeat.** When the `LINQ2DB_TEST_PROGRESS` environment variable is set, the test assembly writes a live JSON heartbeat to `.build/.claude/test-progress.<tfm>.<pid>.json` while the run is in flight (see [Tests/Base/TestProgressReporter.cs](../../Tests/Base/TestProgressReporter.cs)). Don't set it inline on the command — it's toggled session-wide via the `/test-progress` skill (which injects it through `.claude/settings.local.json` → `env`), so the plain `dotnet test` form above is correct whether the trace is on or off. When it's on, a watcher can poll `.claude/scripts/test-status.ps1` without waiting for this agent to return.
 - The four EFCore projects each have a single TFM, so `-f <tfm>` is redundant for them. Include `-f <tfm>` only for `Tests/Linq/Tests.csproj` (multi-TFM).
 - `--logger "console;verbosity=detailed"` when `verbosity: "detailed"`.
 - Don't pipe output to `head`/`tail` — read the whole log. Per `testing.md`: NUnit and `dotnet test` interleave relevant info across the log; setup exceptions can come well before the assertion, and stack traces may be truncated if you skim.
--- a/.claude/agents/test-runner.md
+++ b/.claude/agents/test-runner.md
@@ -72,7 +72,7 @@ dotnet test <project> --filter "<testPattern>" -c <config>
 ```
 
 Notes:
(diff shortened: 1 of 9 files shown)


MaceWindu:
## Follow-up commit

Pushed f4991a47e5f7fa13bdd1081adfa853e92a42cb71 to this branch: removed a dead `if ($null -eq $seconds)` guard in `.claude/scripts/test-status.ps1`'s `Format-Duration` — `$seconds` is typed `[double]`, so a `$null` argument is coerced to `0.0` and the branch can never execute (both call sites also pass `[double]`-coerced values). Pure dead-code removal, no behavior change.

MaceWindu (commented):
> [!IMPORTANT]

No blocking, major, or correctness/API findings. The opt-in heartbeat is a no-op when `LINQ2DB_TEST_PROGRESS` is unset; the `ITestAction`'s shared state is consistently guarded by a single lock; the NUnit lifecycle and `--filter` caveat claims check out against the docs; and the env-var / file-path / throttle semantics agree across the reporter, `test-status.ps1`, both skills, and the docs.

A single dead null-guard in `.claude/scripts/test-status.ps1` (`if ($null -eq $seconds)` on a `[double]` param, which can never be null) was fixed directly in a follow-up commit on this branch rather than posted as a finding.

## Out-of-scope observations

- **Heartbeat fallback write is non-atomic, but the reader compensates** — `TestProgressReporter.WriteFile` publishes atomically via a temp file + `File.Replace`/`Move`, but its `catch` fallback writes directly to the target (`File.WriteAllText`), which a concurrent reader could observe truncated. This is acceptable by design: the fallback only fires on IO contention, the write must never fail the run, and `test-status.ps1` (lines 47–57) already retries parse failures 5×100ms, so a torn read just resolves on the next poll. No change needed.
- **EF Core assembly-info is compiled into all four EF test assemblies** — `Tests/EntityFrameworkCore/AssemblyInfo.TestProgress.cs` is picked up by default SDK globbing in EF3/EF8/EF9/EF10, so each gets its own `[assembly: TestProgressReporter]`. This is correct and collision-free: each runs as a distinct process and the heartbeat filename embeds `<tfm>.<pid>`. Matches the PR description's EF3/8/9/10 claim.
- **`testing.md` field list is a representative subset, not exhaustive** — the documented "Fields:" list omits `tfm`, `pid`, `startedUtc`, `updatedUtc`, and `started`, which the JSON also emits and `test-status.ps1` reads. Not a contradiction (the doc doesn't claim completeness) and no consumer breaks; noted only in case a future doc edit wants the list exhaustive.

## Baselines

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Added an opt-in live test-progress heartbeat for long NUnit test runs (test-harness/agent-tooling only; no effect on the library or its consumers).


_(none)_


---
<sub>Generated from commit `f4991a4`.</sub>

Issue #2424, opened 2020-08-20
Wrong query generation for 0 <= i.ToReceipt.CompareTo(receiptTo)


Firebie:
```c#
  public class ReceiptingBatchDc
  {
   public string   FromReceipt { get; set; }
  }

      dc.GetTable<ReceiptingBatchDc>().FirstOrDefault(
          i => 0 <= i.ToReceipt.CompareTo(receiptTo));
```
```sql
--  Oracle.Managed Oracle12
DECLARE @receiptTo Varchar2(8) -- String
SET     @receiptTo = '06600000'
DECLARE @take Int32
SET     @take = 1

SELECT
	i.ToReceipt
FROM
	ReceiptingBatchDc i
WHERE
	i.ToReceipt > :receiptTo
FETCH NEXT :take ROWS ONLY
```
while expected to be "i.ToReceipt >= :receiptTo"

### Environment details
linq2db version: 3.0.0.0+bf9c05b57af061b07f8b1e05d7eed63ee2630b87
Database Server: Oracle 12
Database Provider: ODP.NET
Operating system: Windows 10
.NET Framework: 4.6.2

Firebie:
Proposed fix:
```diff
diff --git i/Tools/LinqToDB/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs w/Tools/LinqToDB/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
index ecc8ab869d..83928325a7 100644
--- i/Tools/LinqToDB/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
+++ w/Tools/LinqToDB/Source/LinqToDB/SqlProvider/BasicSqlOptimizer.cs
@@ -1123,10 +1123,10 @@ namespace LinqToDB.SqlProvider
 				case SqlPredicate.Operator.NotEqual       : return skipEqual ? op : SqlPredicate.Operator.Equal;
 				case SqlPredicate.Operator.Greater        : return SqlPredicate.Operator.LessOrEqual;
 				case SqlPredicate.Operator.NotLess        :
-				case SqlPredicate.Operator.GreaterOrEqual : return SqlPredicate.Operator.Less;
+				case SqlPredicate.Operator.GreaterOrEqual : return skipEqual ? SqlPredicate.Operator.LessOrEqual : SqlPredicate.Operator.Less;
[...]

PR #5613, opened 2026-06-13
Migrate string.CompareTo to a member translator

MaceWindu:
## Summary

Migrates `string.CompareTo(string)` and `string.CompareTo(object)` translation from the legacy `Expressions._commonMembers` + `[Sql.Extension]` path into the member-translator pipeline (`StringMemberTranslatorBase`), continuing the move off the expose-mapping system (after `Math.Max/Min`, `string.Replace`, `Sql.NullIf`, …).

## Changes

- **`StringMemberTranslatorBase`** — new `TranslateCompareTo` registered for both overloads, emitting `SqlCompareToExpression` directly (the same node the legacy `[Sql.Extension]` builder produced, so the SQL `CASE` lowering is unchanged). The `object` overload converts its argument to a string (`arg.ToString()`) before comparison, matching the legacy mapping body.
- **`Expressions.cs`** — removed both `_commonMembers` `CompareTo` mappings and the now-unused private `ConvertToCaseCompareToImpl`. Kept the frozen public obsolete `ConvertToCaseCompareTo` and `ConvertToCaseCompareToBuilder` (the builder is still referenced by the public method).

## Behavior change

In-memory / client-side evaluation of `string.CompareTo` now uses **BCL semantics** instead of the removed ordinal local-eval body (`ConvertToCaseCompareToImpl`, added in #5577):

- string overload → culture-sensitive (BCL `string.CompareTo`);
- object overload → BCL `string.CompareTo(object)` throws for a non-string argument.

**SQL translation is unchanged** (ordinal `CASE`). Because `AssertQuery` compares the SQL result against an in-memory expected set, the #2424 null-comparison regression tests (`Issue2424Nullable`, `Issue2424Fields`) are reworked to assert against a real-BCL expected set via a new `AssertCompareQuery` helper rather than `AssertQuery`, whose expected side re-expanded the removed mapping and diverged on null operands.

## Verification

- Release build clean on `net8.0` and `netstandard2.0` (analyzers + banned-API).
- SQLite: all `StringFunctionTests` `CompareTo`/`Compare`/`CompareOrdinal` and all `Issue2424*` tests pass.
- Baselines: only the two reworked tests' baselines change; every other comparison baseline is byte-identical, confirming SQL equivalence.

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/Translation/StringMemberTranslatorBase.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Translation/StringMemberTranslatorBase.cs
@@ -32,6 +32,9 @@ namespace LinqToDB.Internal.DataProvider.Translation
 			Registration.RegisterMethod(() => Sql.Replace("", (char?)null, null), TranslateSqlReplace);
 			Registration.RegisterMember(() => "".Length, TranslateLength);
 
+			Registration.RegisterMethod(() => "".CompareTo(""), TranslateCompareTo);
+			Registration.RegisterMethod(() => "".CompareTo(1),  TranslateCompareTo);
+
 			// ReSharper disable ReturnValueOfPureMethodIsNotUsed
 			Registration.RegisterMethod(() => "".Replace("", ""), TranslateStringReplace);
 			Registration.RegisterMethod(() => "".Replace(' ', ' '), TranslateStringReplace);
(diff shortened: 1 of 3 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `03f5db9`.</sub>

PR #5614, opened 2026-06-14
Parallel test execution across database providers

MaceWindu:
Needs  #5711

## What

Run NUnit tests in parallel **across database providers** while serializing tests
that target the same database, via a custom `IWorkItemDispatcher`. NUnit has no
built-in "parallelize by resource" capability (nunit/nunit#165; see nunit/nunit#3122).

## How

- **Per-test `CustomTestContext`** — `AsyncLocal` storage + a serialized `_remote`
  register for the shared LinqService server thread (unreachable by the test's AsyncLocal).
  `LastQuery` is now per-test too.
- **Thread-safe `BaselinesWriter`**.
- **`ParallelDatabaseWorkItemDispatcher`** — leaf provider cases → per-provider
  single-thread lanes (a provider's direct + LinqService variants share one lane);
  composites → original dispatcher; `[NonParallelizable]` work → a globally-exclusive
  lane; a remote gate serializes all LinqService tests; every leaf body runs through a
  read/write gate (with a `[ThreadStatic]` reentrancy guard).
- **`CreateDatabase` readiness latch** — schema creation runs off-lane; other provider
  tests wait on it.
- **`[assembly: Parallelizable]`** + dispatcher swap in `TestsInitialization` (guarded,
  so serial runs are unaffected). Global-state tests marked `[NonParallelizable]`.

## Status — DRAFT / WIP

Mechanism complete and validated on the no-docker SQLite testbed (SQLite.Classic +
SQLite.MS, direct + LinqService). **Full parallel suite green: 18,177 passed /
0 failed / 76 skipped, across 2 consecutive clean runs.** Branch synced with
master; CI (`/azp run test-all`) running.

Done:
- [x] Per-test `LastQuery`; `MiniProfilerTests` / `StackUseTests` global state.
- [x] `Configuration string is not provided` collateral (DefaultConfiguration/Settings).
- [x] Cache-counter tests — `GetCacheMissCount` (~25 files) + bespoke eval-counter
      tests (ParameterTests, CommonTests.Issue288Test, QueryFilterTests.Issue4508Test).
- [x] Expressions registry races — body-level `MapMember`/`MapBinary` writers marked,
      and `ExpressionsTests`/`Issue2468Tests` static-ctor registrations pre-warmed in OneTimeSetUp.
- [x] `MappingSchema.Default` mutators + mapping-schema-cache identity asserts.
- [x] Cross-provider shared-file races (`TestDatabase.sqlite`/`TestDatabase` in SQLite/Access CreateDatabase).
- [x] Tests mutating shared per-fixture instance fields across parallel cases (Issue822Tests).
- [x] Shared in-memory model — re-entrant lock on the publish-before-populate
      `Parent`/`Child` getters in `TestBase.Tables.cs`.
- [x] Full SQLite suite green (2 consecutive clean runs).

Remaining:
- [ ] Docker-provider matrix validation (real SqlServer/PostgreSQL/etc.).
- [ ] CI job consolidation (merge same-engine / same-family / single-runner jobs) —
      deferred until the docker matrix is proven.

## Follow-up commits (review pass)

All four self-labelled temporary items are gone, so `Source/` is now identical to `master` and the
PR's footprint is `Tests/` + `Build/` + root only. CI job consolidation above is done.

- **Readiness latch** - providers that get no `CreateDatabase` case (the Northwind contexts,
  `TestNoopProvider`) are pre-marked ready. Each previously stalled its lane for the full 2-minute
  timeout on every run while holding the dispatcher's read lock.
- **Ungated work** runs on its own per-key lane instead of the dispatching NUnit worker, so provider
  creations overlap and the waiters' timeout is per-provider rather than cumulative across the run.
- **Lane robustness** - item failures are contained rather than killing the lane thread, and lanes are
  ended and reported at teardown instead of living to process exit.
- **State isolation** - the `Configuration.Sql.DefaultNullsPosition` test is serialized;
  `LowerCaseMappingTest` is unmarked (it no longer touches a global); the fallback test context no
  longer retains trace/baseline buffers; the metrics baseline name no longer depends on which test
  wrote a baseline last; the query-cache tests no longer clear the process-wide cache.
- **Tests** - lane routing and readiness-latch invariants are pinned (`ParallelExecutionTests`).
- **CI** - Firebird readiness is probed with `isql` rather than a fixed sleep; SQL Server 2005 and
  EXTRAS setup steps now fail when the container never starts; YDB in-memory pdisks are actually
  enabled (`=1` was inert - the image compares against `"true"`); orphaned single-version 2022
  scripts removed; `README.md`, `CONTRIBUTING.md` and `test-matrix.yml` docs corrected.
- **Env vars** settled on the `L2DB_` prefix (`L2DB_ASSERT_STATE`, new `L2DB_PARALLEL_DIAG`).

Verified: full DuckDB suite with `L2DB_ASSERT_STATE=1`, which compares the shared tables after every
test - 9712 tests, 0 failures.


Tests/Base/BaselinesWriter.cs
```diff
@@ -23,17 +23,20 @@ enum BaselineType
 
 		static string? _context;
 
+		// guards _baselines, _context and the baseline file writes against concurrent
+		// access when tests run in parallel across providers
+		static readonly object _sync = new object();
```
MaceWindu: we use Lock type for sync

Tests/Base/BaselinesWriter.cs
```diff
@@ -23,17 +23,20 @@ enum BaselineType
 
 		static string? _context;
```
MaceWindu: we mustn't have static non-shared state in baselines writer. It's responsibility of CustomTestContext to store such data

Tests/Base/CustomTestContext.cs
```diff
@@ -9,26 +12,72 @@ public sealed class CustomTestContext
...
 		public static string TRACE_DISABLED    = "key-trace-disabled";
+		public static string LASTQUERY         = "key-lastquery";
 
-		// because we don't use parallel test run, we just use single global context instance
-		// it allows us to access context from other threads, started by tests and from linqservice server
-		private static readonly CustomTestContext _context = new CustomTestContext();
+		// Per-test context for the local/direct execution path. Established in [SetUp]
+		// (TestBase.OnBeforeTest); flows to the test body, [TearDown] and any threads/tasks
+		// the test spawns, so parallel tests on different providers stay isolated.
+		private static readonly AsyncLocal<CustomTestContext?> _current = new();
```
MaceWindu: AsyncLocal is unreliable in non-continuous async execution flows, which nunit implement. What will work for us is context-keyed storage, e.g. ConcurrentDictionary<string, CustomTestContext> we already have below.

Tests/Base/CustomTestContext.cs
```diff
@@ -9,26 +12,72 @@ public sealed class CustomTestContext
...
+		// test's AsyncLocal does not flow. Each remote (LinqService) test publishes its context
+		// here keyed by provider; the server resolves the right one via _serverProvider. Keyed by
+		// provider so a lagging or cross-provider server write can't land in another test's capture.
+		private static readonly ConcurrentDictionary<string, CustomTestContext> _remoteByProvider =
+			new ConcurrentDictionary<string, CustomTestContext>(StringComparer.OrdinalIgnoreCase);
+
+		// Set by the test's LinqService server (TestLinqService.CreateDataContext) to the provider
+		// the current request executes for; AsyncLocal so it flows through the server's async query
+		// execution and is visible to the trace sink (Get) on the executing thread.
+		private static readonly AsyncLocal<string?> _serverProvider = new();
```
MaceWindu, line 38: same here, better to store it in CustomTestContext instance directly, AsyncLocal must be avoided when possible

Tests/Base/CustomTestContext.cs
```diff
@@ -9,26 +12,72 @@ public sealed class CustomTestContext
...
 		{
-			_state.Clear();
+			var ctx = _current.Value;
+			_current.Value = null;
+
+			// only the remote test that published the register clears it (atomic key+value remove,
+			// so a newer test's entry for the same provider is never dropped)
+			if (ctx?._remoteProvider is { } provider)
+				((ICollection<KeyValuePair<string, CustomTestContext>>)_remoteByProvider)
+					.Remove(new KeyValuePair<string, CustomTestContext>(provider, ctx));
```
MaceWindu: any reason to overcomplicate simple remove operation like that?

Tests/Base/TestBase.Tables.cs
```diff
@@ -14,6 +14,14 @@ public partial class TestBase
 	{
 		internal const int MaxPersonID = 4;
 
+		// Guards lazy init of the shared in-memory Parent/Child model below. Those getters publish
+		// their backing field before fully populating element associations (.Children/.GrandChildren)
+		// to break a circular Parent<->Child init dependency; under [Parallelizable] one fixture instance
+		// is shared across concurrent cases, so a second case could otherwise observe a half-built model
+		// (null .Children -> NRE). A plain Monitor lock is re-entrant (unlike System.Threading.Lock), so
+		// same-thread cyclic calls re-enter freely while a concurrent case blocks until init completes.
+		readonly object _modelSync = new object();
```
MaceWindu: same - use Lock

Tests/Base/TestBase.Tables.cs
MaceWindu, line 1: candidate for separate pre-flight PR

Tests/Base/TestBase.cs
```diff
@@ -19,7 +20,30 @@ public abstract partial class TestBase
...
+		}
+
+		// Set when the parallel dispatcher is installed (TestsInitialization). Gates the
+		// per-provider database-readiness wait below, so serial / filtered runs are unaffected.
+		public static bool ParallelExecutionEnabled;
+
+		// Under parallel execution a provider's tests wait until its CreateDatabase has populated
+		// the schema. CreateDatabase runs off the provider lane (see ParallelDatabaseWorkItemDispatcher),
+		// so this never deadlocks the lane. Keyed by the bare provider context (remote suffix stripped).
+		static readonly ConcurrentDictionary<string, ManualResetEventSlim> _databaseReady =
```
MaceWindu: don't spread context-specific data all over codebase. move to CustomTestContext instance

Tests/Linq/DataProvider/PostgreSQLArrayTests.cs
```diff
@@ -40,7 +40,7 @@ public static AllTypesValueTable[] Seed()
 			}
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/DataProvider/PostgreSQLExtensionsTests.cs
```diff
@@ -52,7 +52,7 @@ public static SampleClass[] Seed()
 			}
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/DataProvider/SQLiteTests.cs
```diff
@@ -492,7 +492,8 @@ public class CreateTableTest
 			public int Id;
 		}
 
-		[Test]
+		// NonParallelizable: creates/drops a hardcoded shared "TestDatabase.sqlite" used by both SQLite providers; their lanes run concurrently and collide on the file. Future: per-provider filename.
```
MaceWindu: use context-specific file name instead

Tests/Linq/Linq/CachingTests.cs
```diff
@@ -234,7 +234,7 @@ from c2 in db.Child.OrderBy(r => r.ParentID).Take(10, takeHint)
 		}
 
 		[ActiveIssue]
-		[Test(Description = "https://github.com/linq2db/linq2db/issues/4266")]
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/4266"), NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/ConstantTests.cs
```diff
@@ -140,7 +140,7 @@ from e in table
 			AssertQuery(query);
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/DistinctTests.cs
```diff
@@ -51,7 +51,7 @@ public void Distinct4([DataSources] string context)
 				(from p in db.Parent select new Parent { ParentID = p.Value1 ?? p.ParentID % 2, Value1 = p.Value1 }).Distinct());
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/EnumerableSourceTests.AsQueryable.cs
```diff
@@ -116,7 +116,7 @@ public void AsQueryable_Inline_ExceptData_ParameterisesData(
 			sql.ShouldNotContain("'Data 7778'");
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/EnumerableSourceTests.cs
```diff
@@ -16,7 +16,7 @@ namespace Tests.Linq
 	[TestFixture]
 	public partial class EnumerableSourceTests : TestBase
 	{
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/EvaluationTests.cs
```diff
@@ -12,7 +12,8 @@ namespace Tests.Linq
 	[TestFixture]
 	public class EvaluationTests : TestBase
 	{
-		[Test]
+		// NonParallelizable: mutates the process-global Expressions registry from the test body (init-only/not thread-safe) — would corrupt concurrent readers. Future: register once at init.
```
MaceWindu: propose library fix on separate PR

Tests/Linq/Linq/ExpressionsTests.cs
```diff
@@ -1118,7 +1122,7 @@ public void TestNullCheckInExpressionUsingFieldRight([IncludeDataSources(TestPro
 		#endregion
 
 		#region Regression: query comparison
-		[Test(Description = "Tests regression introduced in 3.5.2")]
+		[Test(Description = "Tests regression introduced in 3.5.2"), NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing

Tests/Linq/Linq/FromSqlTests.cs
```diff
@@ -227,7 +227,7 @@ from s in db.FromSql<SampleClass>($"SELECT * FROM {GetName(table)} where {GetCol
 			Assert.That(projection, Is.EqualTo(expected));
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/FullTextTests.SqlServer.cs
```diff
@@ -856,7 +856,7 @@ orderby t.Rank descending
 			db.LastQuery!.ShouldContain("CONTAINSTABLE([Categories], ([Description]), @search, @top)");
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/MathFunctionTests.cs
```diff
@@ -340,7 +340,7 @@ from t in from p in Types select Math.Round((double)p.MoneyValue, 1, MidpointRou
 		}
 
 		// TODO: implement other MidpointRounding values (and remove NUnit4001 suppress)
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/ParameterTests.cs
```diff
@@ -530,7 +530,7 @@ IQueryable<Person> GetPersons2(ITestDataContext db, int? personId)
 			return db.Person.Where(p => p.ID == personId!.Value);
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/QueryFilterTests.cs
```diff
@@ -126,7 +126,7 @@ sealed class DcParams
 			public bool IsSoftDeleteFilterEnabled { get; set; } = true;
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/SelectTests.cs
```diff
@@ -1088,7 +1088,7 @@ sealed class IntermediateChildResult
 			public Child? Child    { get; set; }
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/SpecialFunctionsTest.cs
```diff
@@ -26,7 +26,7 @@ orderby Sql.Ordinal(q.ParentID), q.Value1
 			Assert.That(orderBy.Items[0].IsPositioned, Is.True);
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/StringTrimTests.cs
```diff
@@ -369,7 +369,7 @@ public void TrimEnd2Test([DataSources] string context)
 		// different chars value is safe — asserting a miss there would lock in
 		// avoidable cache churn.
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/TakeSkipTests.cs
```diff
@@ -44,7 +44,7 @@ static void CheckTakeSkipParameterized(IDataContext dc, int additional = 0)
 			CheckTakeSkipParams(dc, false, additional);
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Linq/TestQueryCache.cs
```diff
@@ -164,7 +164,7 @@ private static MappingSchema CreateMappingSchema(string columnName, string? sche
 			return ms;
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Mapping/FluentMappingBuildTests.cs
```diff
@@ -111,7 +111,7 @@ await tmp
 				.MergeAsync();
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Update/MergeTests.Caching.cs
```diff
@@ -11,7 +11,7 @@ namespace Tests.xUpdate
 	// tests for empty enumerable source
 	public partial class MergeTests
 	{
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/Update/MergeTests.Issues.cs
```diff
@@ -1008,7 +1008,7 @@ public void Issue4584Test([MergeDataContextSource(false)] string context)
 			Assert.That(db.LastQuery!.Count(_ => _ == GetParameterToken(context)), Is.EqualTo(6));
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/UserTests/Issue3148Tests.cs
```diff
@@ -18,7 +18,7 @@ namespace Tests.UserTests
 	[TestFixture]
 	public class Issue3148Tests : TestBase
 	{
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/UserTests/Issue447Tests.cs
```diff
@@ -44,7 +44,7 @@ public void TestLinqToDBComplexQuery2([DataSources] string context)
 			_ = result.ToSqlQuery().Sql;
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/UserTests/Issue5302Tests.cs
```diff
@@ -28,7 +28,7 @@ public class Warehouse
 		/// <summary>
 		/// Reproduces the issue: query using ValueIsEqualToAny with array causes cache miss on every execution.
 		/// </summary>
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

Tests/Linq/UserTests/Issue973Tests.cs
```diff
@@ -112,7 +112,7 @@ public void Test([DataSources] string context)
 				GetParents(db, values2));
 		}
 
-		[Test]
+		[Test, NonParallelizable]
```
MaceWindu: reasoning for NonParallelizable missing in this file

MaceWindu:
## Follow-up commit `c144427c`

Extracted the generic parallel-dispatch machinery into a reusable, linq2db-free module so it can be reused or proposed upstream (nunit/nunit#165, #3122). Behaviour-identical.

`Tests/Base/Parallelization/` (namespace `NUnit.ParallelByResource`):

- **`ResourceLaneDispatcher`** (+ nested `SerialLane`) — the former `ParallelDatabaseWorkItemDispatcher`, now parameterized by an **`IResourceLaneStrategy`** that maps a test work item to a **`LaneAssignment`**: `GatedInline` / `SerialLane(+ secondary mutex)` / `Ungated`. The `[NonParallelizable]` → exclusive-lane and `CompositeWorkItem` → original-dispatcher handling stay in the dispatcher (pure NUnit). Diagnostics go through an injected **`IParallelDiagnostics`** (no-op default).
- **`ResourceReadinessLatch`** — generic keyed `MarkReady` / `WaitReady`; `CustomTestContext`'s database-readiness latch now delegates to it (public signatures unchanged).
- **`ResourceLaneDispatcherInstaller.TryInstall`** — wraps the `TestExecutionContext.Dispatcher` swap.

linq2db keeps its specifics behind the seam: **`DatabaseLaneStrategy`** keys lanes by provider (via `NUnitUtils`), routes `CreateDatabase` → `Ungated` and `LinqService` → secondary mutex; `TestsInitialization` installs via the helper, passing `ParallelDiag` as the sink.

Primitives: `System.Threading.Lock` for the lane-dictionary critical section, `SemaphoreSlim(1,1)` for the cross-lane secondary gate, `ThreadLocal<bool>` for the re-entrancy flag.

Verified locally on a parallel SQLite slice (SQLite.Classic + SQLite.MS): all routing dispositions fired (ungated `CreateDatabase`, exclusive lane for `[NonParallelizable]`, per-provider serial lanes running concurrently) and the readiness latch signalled correctly.

MaceWindu:
## Corpus doc edits dropped by the master sync (`81cd5f225`)

Master's submodule migration deleted `.agents/` (the corpus now lives in the `.claude/` submodule -> [linq2db/agents](https://github.com/linq2db/agents)). This branch had edited four files there; the merge resolved them to master's deletion, so the content is gone from the diff.

The edits replace the old "one distinct provider per process, build once then run the test exe directly" guidance with the in-process cross-provider parallelization this PR introduces. **They need porting into the corpus repo once this PR merges** - the corpus is not versioned per linq2db branch, so the change can only land there after the behaviour it documents is on master.

Files:

- `.agents/agents/test-runner.md` -> `.claude/agents/test-runner.md`
- `.agents/docs/testing.md` -> `.claude/docs/testing.md` (section rename: *Running providers in parallel* -> *Testing several providers in one run*)
- `.agents/skills/test/SKILL.md` -> `.claude/skills/test/SKILL.md`
- `.agents/skills/test-progress/SKILL.md` -> `.claude/skills/test-progress/SKILL.md`

<details>
<summary>Dropped diff (against merge-base `4f597b6b3`)</summary>

```diff
diff --git a/.agents/agents/test-runner.md b/.agents/agents/test-runner.md
index e5a6a8060..d40abeab0 100644
--- a/.agents/agents/test-runner.md
+++ b/.agents/agents/test-runner.md
@@ -53,7 +53,7 @@ Container availability is **not** checked here (that's `/test-providers`' job):
 | EFCore EF9 tests | `Tests/EntityFrameworkCore/Tests.EntityFrameworkCore.EF9.csproj` | `NET90` (net9.0) |
 | EFCore EF10 tests | `Tests/EntityFrameworkCore/Tests.EntityFrameworkCore.EF10.csproj` | `NET100` (net10.0) |
 
-A single invocation may span multiple targets; run them sequentially (most test DBs — SQL Server especially — don't survive parallel EF3/EF8/EF9/EF10 runs because they share database names like `TestData2016MS`).
+A single invocation may span multiple targets.
 
 ### Default to Playground + net10.0 for main-test runs
 
@@ -141,5 +141,4 @@ Rules:
 - Do not run tests in `Release` config unless the caller explicitly sets `config: "Release"`.
 - Do not prompt the user. Unclear input or a provider mismatch is an error the caller resolves.
 - Do not skip log lines. If the log is very long, still read it fully — the first failure's setup lines are usually the root cause, not the final summary.
-- Do not run multiple targets in parallel. EF3/EF8/EF9/EF10 against SQL Server share database names; parallel runs corrupt state.
 - Never kill processes by broad pattern. If a run is blocked by a file lock (`CS2012`, `Access denied`, a DLL "used by another process"), do **not** run `pkill -f dotnet`, `Stop-Process -Name dotnet`, or `taskkill /IM dotnet.exe` — a machine-wide kill terminates the user's other builds / IDEs / agents and is never authorized. Either abort with a `blocked` status carrying the lock detail, or clear it narrowly: `dotnet build-server shutdown` (VBCSCompiler / MSBuild server), or `Stop-Process -Id <pid>` for the **specific** PID named in the lock message (see [`../docs/windows-dev-gotchas.md`](../docs/windows-dev-gotchas.md) → *Iterative-build gotchas*).
diff --git a/.agents/docs/testing.md b/.agents/docs/testing.md
index 7f5ea2f23..0fd06da49 100644
--- a/.agents/docs/testing.md
+++ b/.agents/docs/testing.md
@@ -112,24 +112,20 @@ dotnet test --project Tests/Linq/Tests.csproj -f net10.0 --settings .runsettings
 
 `--provider` replaces the active main-test provider set (`TestConfiguration.UserProviders`) with exactly the names given, so **any provider with a connection string defined in `DataProviders.json` / `UserDataProviders.json` runs without editing the enabled-providers list** — it need not be enabled. (EF Core test runs instead intersect their curated `EFProviders` list, so `--provider` can only narrow EF runs to supported providers, never add an unsupported one.) Connection strings still come from those files; a `--provider` name with no connection logs a warning and fails to connect. An absent option leaves behavior unchanged (the providers enabled in `UserDataProviders.json` run).
 
-### Running providers in parallel
+### Testing several providers in one run
 
-Each provider config maps to a distinct database, so runs scoped to **distinct** providers don't collide and can run concurrently. Two cautions, both from the standing single-session rule:
+The test harness parallelizes across providers **in-process**: a single `dotnet test` (or test-exe) run given multiple providers executes their tests concurrently, each provider on its own database, within one process. To exercise several providers, pass them all to **one** run and let the harness parallelize internally:
 
-- **One distinct provider (= one database) per parallel run.** The same provider on two concurrent runs — including the *same* provider across two TFMs — hits the same database and corrupts state. The parallel unit is the database, not the process.
-- **Build once, and don't let the runs build concurrently.** Two `dotnet test` invocations on the same project race to write and lock the shared `linq2db.Tests.dll` (fails with `MSB3027`). Build once, then launch each run as the **test executable directly** — it never builds, and each process writes its own per-PID heartbeat:
-
-```bash
-dotnet build Tests/Linq/Tests.csproj -c Debug -f net10.0
-```
 ```bash
-.build/bin/Tests/Debug/net10.0/linq2db.Tests.exe --provider PostgreSQL.18 --test-progress --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest"
-```
-```bash
-.build/bin/Tests/Debug/net10.0/linq2db.Tests.exe --provider MySql.8.0 --test-progress --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest"
+dotnet test --project Tests/Linq/Tests.csproj -f net10.0 --settings .runsettings --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest" --provider PostgreSQL.18,MySql.8.0
 ```
 
-Each process writes its own `.build/.agents/test-progress.<tfm>.<pid>.json`; poll them with `/test-progress` or `test-status.ps1`. The runner's native filter flag is `--filter` (the same `FullyQualifiedName~` syntax), confirmed via `--help`.
+**Do not launch multiple test processes to cover several providers.** Two concurrent `dotnet test` / test-exe invocations:
+
+- race to build and lock the shared `linq2db.Tests.dll` (`MSB3027`); and
+- corrupt each other's state the moment they touch the **same** database — e.g. the *same* provider run concurrently across two TFMs. The parallel unit is the database, and the in-process runner already schedules distinct-database work concurrently, so a second process adds conflict risk, not coverage.
+
+Poll a run's progress with `/test-progress` (heartbeat at `.build/.agents/test-progress.<tfm>.<pid>.json`). The runner's native filter flag is `--filter` (the same `FullyQualifiedName~` syntax), confirmed via `--help`.
 
 ## Database initialization
 
diff --git a/.agents/skills/test-progress/SKILL.md b/.agents/skills/test-progress/SKILL.md
index 6335124f3..f893082fb 100644
--- a/.agents/skills/test-progress/SKILL.md
+++ b/.agents/skills/test-progress/SKILL.md
@@ -27,7 +27,7 @@ The file lands at `.build/.agents/test-progress.<tfm>.<pid>.json` (one per TFM /
 
    Relay its one-line output (state, completed/total, pass/fail/skip, rate, elapsed, ETA, current test). If it reports no heartbeat file, say so — either no run has started, or the run was launched without `--test-progress` (unusual for a Claude-launched run).
 2. Add `-Raw` if the user wants the full JSON, or `-Path <file>` to target a specific run's heartbeat.
-3. For a parallel per-provider run, each process writes its own `.build/.agents/test-progress.<tfm>.<pid>.json`; point `-Path` at the one you want to watch.
+3. Each run writes a heartbeat named `.build/.agents/test-progress.<tfm>.<pid>.json`; when more than one run is active (e.g. different TFMs), point `-Path` at the one you want to watch.
 
 ## Don'ts
 
diff --git a/.agents/skills/test/SKILL.md b/.agents/skills/test/SKILL.md
index c7e35c79c..f9470bf72 100644
--- a/.agents/skills/test/SKILL.md
+++ b/.agents/skills/test/SKILL.md
@@ -140,6 +140,6 @@ If containers were started during the session via `/test-providers`, remind the
 - Do not edit `UserDataProviders.json` or invoke `docker` commands from `/test`. Env changes route through `/test-providers` exclusively. If the run needs a different provider set or a stopped container started, tell the user — don't fix it.
 - Do not run tests in `Release` config by default — analyzers + banned-API checks are slow and rarely what the user wants for a single-test run.
 - Do not edit source files yourself. Writing is `test-writer`'s job; running is `test-runner`'s job. The skill only orchestrates and relays.
-- Do not run targets in parallel from `/test`. The agents won't anyway (sequential is built into `test-runner`), and don't fan out multiple agent invocations with overlapping target sets. Deliberately parallel runs are possible but only **one distinct provider (= one database) per run**, and only as a manual workflow — see `.agents/docs/testing.md` → **Running providers in parallel** (build once, then run the test exe directly with `--provider`).
+- Do not launch multiple test processes to cover several providers. The harness parallelizes across providers **in-process**, so pass the full set to a **single** `test-runner` invocation (`--provider a,b,c`) and let it parallelize internally; a second concurrent process only risks shared-database corruption, not more coverage (see `.agents/docs/testing.md` → **Testing several providers in one run**). Don't fan out multiple `test-runner` invocations with overlapping target sets.
 - Do not suppress or skim the agent's output. If the agent reports a failure, relay the verbatim error message; don't paraphrase.
 - Do not pre-validate the env state before invoking `test-runner`. Trust the user; let the agent's own provider-check abort surface any mismatch. Auto-fix attempts here defeat the boundary.
```

</details>

MaceWindu:
## Follow-up: replace the per-TFM `UserDataProviders.json` copies with `--provider`

Checked against the tree at `81cd5f225`:

- `Build/Azure/{netfx,net80,net90,net100}/` holds **92 files across 26 distinct config names**.
- **None of them contains a connection string.** Every connection comes from the tracked `DataProviders.json`; these files only select *which* providers run.
- **No config name's provider list differs between its TFM directories** (0 of 26), so one list per matrix leg would reproduce all four copies.
- 24 of the 26 are a bare `Providers` array. Two carry one extra setting each: `sqlserver.2005` → `DisableRemoteContext`, `sqlserver.2022.metrics` → `StoreMetrics`.
- (The three `hana2.json` files are empty, so HANA already resolves its providers elsewhere.)

The runner's `--provider a,b,c` replaces `TestConfiguration.UserProviders` with exactly the names given — precisely what these files do. A `providers:` string on each `test-matrix.yml` leg could therefore replace the whole tree, removing:

- `Build/Azure/{netfx,net80,net90,net100}/` (92 files) and the `artifact_test_configs` artifact
- `Extract test configs` — 1 step per leg
- `Copy <TFM> test config` — 1 step per TFM per leg, so 3-4 per leg

That is ~4-5 fewer steps per leg on top of the loop refactor, one fewer build artifact, and the provider set moves next to the leg definition instead of being spread across four per-TFM files.

Two things to settle first:

1. `DisableRemoteContext` and `StoreMetrics` have no CLI equivalent. They need runner options, or those two legs keep a config file.
2. EF Core runs *intersect* `--provider` with their curated `EFProviders` list rather than replacing it, so EF coverage needs a check that the intersection still matches today's set.

Tests/Linq/TestsInitialization.cs
```diff
@@ -168,6 +207,24 @@ public void TestAssemblySetup()
...
 
+		// Parallelize tests across database providers: route each provider's tests to a
+		// dedicated lane so the same database is never hit concurrently. Only swap when
+		// NUnit is actually running in parallel; a serial run keeps the original dispatcher.
+		var configuredLanes = TestConfiguration.MaxParallelLanes;
+		var maxLanes        = configuredLanes ?? (2 * Environment.ProcessorCount);
+		if (maxLanes < 1)
+			maxLanes = 1;
+
+		if (ResourceLaneDispatcherInstaller.TryInstall(new DatabaseLaneStrategy(), diagnostics: null, maxLanes, out var workers))
```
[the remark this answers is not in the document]
MaceWindu: **Inaccurate.** `diagnostics` is the second parameter of `TryInstall(strategy, diagnostics, maxLanes, out levelOfParallelism)`, so `diagnostics: null` occupies its own positional slot. A non-trailing named argument in its correct position has been legal since C# 7.2, so this compiled as written and no change was needed on that account.

The call has since changed for unrelated reasons: the sink is now non-null when `L2DB_PARALLEL_DIAG=1` (dd2efff89), and the signature gained an `out` for the installed dispatcher so the lanes can be shut down at teardown (3b87f898d).

Source/LinqToDB/Internal/SqlProvider/BasicSqlBuilder.cs
```diff
@@ -2068,6 +2068,12 @@ protected virtual void BuildFromExtensions(SelectQuery selectQuery)
 					if (appendParentheses)
 						AppendIndent().Append(')');
 
+					// TEMP DIAG (Issue1742, uncommitted-intent): table carries the scalar 'value' field but IsScalar is
+					// false here -> the t1(value) alias-list below is skipped, producing the broken `SELECT t1.value
+					// FROM (...) t1` that fails with 42703. Confirms the copy-ctor IsScalar loss.
+					if (!rawSqlTable.IsScalar && rawSqlTable.Fields.Exists(f => string.Equals(f.Name, "value", StringComparison.Ordinal)))
+						System.Console.Error.WriteLine($"RAWSQLDIAG render-lost-IsScalar sql={rawSqlTable.SQL}");
```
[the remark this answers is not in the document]
MaceWindu: **Fixed** in 80a4dbd45. Both halves of the temporary `RAWSQLDIAG` probe are removed — this one on the render path and its counterpart in `TableBuilder.RawSqlContext` on the build path. `Source/` is now byte-identical to `master`, so the PR's footprint is `Tests/` + `Build/` only.

Source/LinqToDB/Internal/Linq/Builder/TableBuilder.RawSqlContext.cs
```diff
@@ -48,6 +48,12 @@ static BuildSequenceResult BuildRawSqlTable(ExpressionBuilder builder, BuildInfo
 
 			isScalar ??= builder.MappingSchema.IsScalarType(entityType);
 
+			// TEMP DIAG (Issue1742, uncommitted-intent): a raw-SQL FromSql over a *class* type built as scalar
+			// is the anomaly behind the parallel `column t1.value does not exist` failure. Log it so CI shows
+			// whether ScalarResult<int> is misclassified scalar at build time (vs. inheriting a cached scalar table).
+			if (isScalar.Value && entityType is { IsPrimitive: false, IsEnum: false } && entityType != typeof(string))
+				System.Console.Error.WriteLine($"RAWSQLDIAG anomalous-scalar-build type={entityType.FullName}");
```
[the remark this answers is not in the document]
MaceWindu: **Fixed** in 80a4dbd45, together with the render-path half in `BasicSqlBuilder`. `Source/` is now byte-identical to `master`.

Build/Azure/pipelines/testing.yml
```diff
@@ -35,7 +35,16 @@ stages:
...
+        # TEMPORARY (PR #5614): make test-all mirror the release run - every TFM's main suite rather
+        # than net10 alone - so the reworked workflow templates and the 300-min release timeout are
+        # exercised on the full matrix. default.yml only enables that combination for PRs targeting
+        # the release branch, which this PR is not. REVERT BEFORE MERGE: restore the plain
+        # `full_run: false` below and drop the test-all db_filter entry.
+        ${{ if eq(variables['Build.DefinitionName'], 'test-all') }}:
+          full_run: true
+          db_filter: '[all]'
+        ${{ if ne(variables['Build.DefinitionName'], 'test-all') }}:
+          full_run: false
```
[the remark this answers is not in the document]
MaceWindu: **Fixed** in 5d446e072. The unconditional `full_run: false` is restored and both `${{ if }}` branches are gone, so `test-all` is back to the PR-run shape rather than the release matrix.

Tests/Linq/DataProvider/DB2Tests.cs
```diff
@@ -38,6 +38,67 @@ public class DB2Tests : DataProviderTestBase
...
+
+		// TEMPORARY (https://github.com/linq2db/linq2db/issues/5765): rows accumulate in AllTypes during a
+		// run - DB2's ResetAllTypesIdentity only restarts the identity, it never deletes - and TestType's
+		// "SELECT ID FROM AllTypes WHERE <field> IS NULL / = @p" then returns a leftover instead of the
+		// seeded row 1/2. Logs each leftover's ID plus the set of columns it has values in; that column
+		// signature identifies which test wrote the row. Remove once the writer is fixed.
+		static void DumpAllTypesLeftovers(DataConnection conn)
+		{
+			using var _ = new DisableBaseline("diagnostics");
+
```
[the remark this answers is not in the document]
MaceWindu: **Fixed** in 7024fc519 — the helper, its `AllTypesLeftover` support type and the call site are all removed, so `DB2Tests.cs` is identical to `master` again.

Removed rather than gated: the probe's own stated exit condition was "remove once the writer is fixed", and that fix (#5765) landed on this branch in a57b9c770.

Tests/Base/BaselinesWriter.cs
MaceWindu: **Done** in 0ef2d56e3 — the static `_context` slot is gone.

It was a last-writer-wins value that `WriteMetrics` read to name the metrics file, so under a parallel multi-provider run the process-wide `ActivityStatistics` report was attributed to an arbitrary provider, differing between runs. `TestUtils.GetConfigName()` already identifies the leg, so the provider component is dropped rather than moved into `CustomTestContext`. The empty-run guard the slot also served is now expressed against the overwrite-detection map.

This renames the existing `*.Metrics.txt` baselines, which the release flow regenerates from the anchor commit anyway.

Tests/Base/TestBase.Tables.cs
MaceWindu: **Done** — `Tests/Base/TestBase.Tables.cs` is no longer in this PR's diff. The `EnsureModel` / `_modelSync` double-checked lock landed on `master` separately, which is what this asked for.

Tests/Linq/Linq/EvaluationTests.cs
MaceWindu: **Done** — `Tests/Linq/Linq/EvaluationTests.cs` is no longer in this PR's diff, so the library fix is out of scope here as requested.

Tests/Base/TestBase.Concurrent.cs
```diff
@@ -78,37 +104,37 @@ protected void ConcurrentRunner<TParam, TResult>(DataConnection dc, string conte
...
+					// the provider's own exception, where the worker thread's original stack is the most
+					// useful part of the report and a rethrow would reset it to this line.
+					ExceptionDispatchInfo.Capture(result.Failure!).Throw();
 				}
 
-				try
+				if (result.Failure == null)
+					continue;
+
+				var testResult = queryFunc(dc, result.Param);
```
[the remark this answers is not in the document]
MaceWindu: **Accurate, and fixed** in 513c26a35 — though it is not a regression from this PR.

The re-run is now contained: its exception is logged, its result dump is skipped, and `result.Failure` is rethrown either way.

On the pre-existing part, for the record: `master` ran the same re-run unguarded too, as `queryFunc(dc, result!.Item1)` on the first line of the `catch` block. An exception raised inside a `catch` replaces the in-flight one, so the check failure was lost there identically — the restructuring in this PR moved the call out of the catch but did not change that property.

Fixed here anyway rather than tracked separately, because it is six lines in a file this PR already rewrote, and the failure mode is realistic rather than exotic: the re-run executes the very query that just produced the failing result, on the shared connection.

One deviation from the suggested shape — `testResult` is consumed by a `DumpObject` further down, so a swallowed re-run cannot simply fall through to the rethrow; it needs a flag to skip that dump, otherwise a `default(TResult)` would be printed as though it were a real, empty result.

Verified with a probe whose worker threads succeed and whose re-run throws: the test now fails with the check exception and only logs the re-run one.

MaceWindu:
### Deferred corpus change — parking note

Not for this PR's diff. Recording it here so it isn't lost between now and merge, per the convention that `.claude/` docs describing an unmerged change wait until it lands.

**Target:** `.claude/docs/testing.md`, appended to the existing *"Capturing a passing test's exception / stack"* section.

**Why deferred:** it documents behaviour this PR introduces (the lane dispatcher, `L2DB_PARALLEL_DIAG`, `ResourceLaneDispatcher.Shutdown`), so the shape could still shift in review.

**Gap it closes:** that section already covers getting a *caught exception's stack* out of a passing test — make the test fail so MTP prints it. It says nothing about the adjacent problem of confirming that a **code path ran at all** in a green run, which came up four separate times during this PR's review and needed a different technique each time it was rediscovered.

<details>
<summary>Proposed addition</summary>

#### Verifying a code path ran, in a green run

Same root cause as above — MTP routes per-test stdout only for failed / skipped tests — but a different question: not *what did this throw*, but *did this callback fire / was this component installed / was this branch taken*. `TestContext.Progress.WriteLine` from production or harness code is invisible on a green run, so a passing run is not evidence the path executed.

Three techniques, cheapest first:

- **Assert the state instead of logging it.** If the thing is observable (`TestBase.ParallelExecutionEnabled`, a latch's state), a throwaway `[Test]` asserting it is one build and a few seconds. This is how the parallel dispatcher's installation was confirmed.
- **Route the sink to a file under `.build/.agents/`.** When the thing under test *is* a logging seam or callback, temporarily replace the sink with `File.AppendAllText` to an absolute path, run, read it back, revert. This is what confirmed `IParallelDiagnostics` was reachable (`dispatch->ungated key=…` per provider) and what caught `Shutdown` self-joining its own lane thread — a defect the green log showed no trace of.
- **Make it fail on purpose.** For a path only reachable on failure (a diagnostic re-run, an error-reporting branch), a throwaway test that forces the failure is the only way to see the output. Name the two exceptions distinguishably so the assertion is unambiguous about which one surfaced.

All three are throwaway: revert before committing. Note the absolute-path caveat — a relative redirect follows the *invoking process's* working directory, which differs between a `--project <worktree>` run and a `Set-Location <worktree>` run (see `worktree.md`).

</details>

Tests/Linq/DataProvider/SqlServerFunctionsTests.cs
```diff
@@ -193,24 +196,29 @@ public void ConvertTest2([IncludeDataSources(TestProvName.AllSqlServer)] string
 			var result = db.Select(() => SqlFn.Convert(SqlType.Decimal, 123));
 
 			Assert.That(result, Is.EqualTo(123m));
+			db.LastQuery!.ShouldContain("CONVERT(decimal,");
 		}
 
+		[ActiveIssue]
```
MaceWindu, line 202: tests never (?) worked properly and now fail after asserts added. will be handled separately as pre-existing issue

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `6896766`.</sub>

PR #5619, opened 2026-06-15
Fix InvalidCastException for aggregates inside set operations

### LINKED ISSUE #5616: Some aggregate functions are failing inside Unions
### Description

Some aggregate functions like .Average() or custom extension as in example are failing inside Unions

### Reproduction Steps

```
var query1 = db.GetTable<Table1>();

var query2 =
    from tr in db.GetTable<Table2>()
    group tr by tr.Id into g
    select new
    {
        Count = DbExtensions.CountIf(g, x => x.Id > 123)
        // Count = g.Average(x => x.Id, Sql.AggregateModifier.All)
    };

var sql = query1
    .Select(x => new
    {
        Count = 0L // or 0d for .Average
    })
    .UnionAll(query2)
    .ToSqlQuery();

Console.WriteLine(sql.Sql);


public static class DbExtensions
{
    [Sql.Extension("count_if({predicate})", IsAggregate = true, ServerSideOnly = true)]
    public static long CountIf<TSource>(IEnumerable<TSource> src, [ExprParameter] Expression<Func<TSource, bool>> predicate)
    {
        throw new InvalidOperationException();
    }
}
```

### Exception details

```
Unhandled exception. System.InvalidCastException: Unable to cast object of type 'LinqToDB.Internal.Expressions.SqlPathExpression' to type 'LinqToDB.Internal.Expressions.SqlPlaceholderExpression'.
   at LinqToDB.Internal.Expressions.ExpressionVisitorBase.VisitSqlReaderIsNullExpression(SqlReaderIsNullExpression node) in /_/Source/LinqToDB/Internal/Expressions/ExpressionVisitorBase.cs:line 104
```

### Regression?

Did work in 5.4.1.9

### Linq To DB version

6.3.0

### .NET version

10.0.300

### Database

_No response_

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5616.

Aggregates dispatched through `MethodChainBuilder` — custom `[Sql.Extension(IsAggregate = true)]`
(e.g. `count_if`) and the analytic `Average(…, Sql.AggregateModifier)` overloads — threw
`InvalidCastException` (`SqlPathExpression` → `SqlPlaceholderExpression`) when placed in one branch
of a set operation (`UNION ALL`, …) against a constant in the other.

## Cause
For a non-nullable CLR return, `MethodChainBuilder.ChainContext.MakeExpression` wraps the aggregate
in a read-time null guard (`CheckNullValue` over `SqlReaderIsNullExpression`). When
`SetOperationBuilder.InitializeProjections` builds each branch's projection,
`MergeProjectionHelper.ExpressionPathVisitor` rewrites column placeholders into `SqlPathExpression`.
It has no `VisitSqlReaderIsNullExpression` override, so the base visitor's hard cast of the rewritten
placeholder back to `SqlPlaceholderExpression` threw.

## Fix
Suppress the guard when building a set projection — gate it on `!flags.IsForSetProjection()`. The
guard is a materialization concern, irrelevant to set-operation column-structure extraction; normal
materialization is unchanged. Mirrors the existing `IsExpression() && !IsForSetProjection()` idiom in
`BuildProxyBase`.

## Tests
- `Issue5616Tests.Average_InUnionAll` — executable `[DataSources]`, built-in aggregate via the same builder path.
- `Issue5616Tests.CustomAggregate_InUnionAll` — the issue's exact `count_if` extension; asserts the query builds.
- Regression: `ConcatUnionTests` + `GroupByTests` + `AnalyticTests` — 4489 passed, 0 failed.


MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Reviewed the one-line guard change in `MethodChainBuilder.ChainContext.MakeExpression` and the new `Issue5616Tests` fixture. The fix is correct and in-scope: `ForSetProjection` is set only during the SQL-shape projection phase, not runtime materialization, so gating the `CheckNullValue`/`SqlReaderIsNullExpression` block on `!flags.IsForSetProjection()` avoids the `SqlPathExpression -> SqlPlaceholderExpression` cast crash without removing the runtime null-check from the normal materialization path. The change mirrors the existing `IsExpression() && !IsForSetProjection()` idiom in `BuildProxyBase`.
- [x] No public-API surface change — the edit is confined to `Internal.*`.

## Out-of-scope observations

- **Fix is at the guard-producer, not the `SqlReaderIsNullExpression` consumer** — The PR body identifies the proximate cause as a missing `VisitSqlReaderIsNullExpression` override in `ExpressionPathVisitor` (the `SqlPathExpression -> SqlPlaceholderExpression` cast fails there during set-projection building). This PR fixes it at the producer side instead. Both are defensible; the chosen approach mirrors the `BuildProxyBase` idiom and keeps the null-guard block out of the SQL-shape projection phase entirely. The alternative (teaching `ExpressionPathVisitor` to handle the node) would also help any other builder that emits `SqlReaderIsNullExpression` into a set projection. Worth a maintainer note, not a blocker.
- **Runtime non-nullable-aggregate NULL guard does not apply across a set operation** — Outside set operations, a non-nullable aggregate (e.g. `Average -> double`) that returns SQL NULL triggers `CheckNullValue` and throws a descriptive "use the nullable version" error at read time. Inside a UNION ALL branch the value materializes from the merged set-operation column rather than the aggregate placeholder, so that guard is structurally absent — this PR makes that explicit. Consistent rather than a regression (the union with a constant/other branch makes the column conceptually nullable), but the behavioral asymmetry is worth awareness.
- **`CustomAggregate_InUnionAll` uses a substring SQL assertion** — It asserts `query.ToSqlQuery().Sql.ShouldContain("count_if")`. A substring SQL assertion is generally weak, but acceptable here: `count_if` is a distinctive token, the function isn't executable on the default SQLite providers, and the originally reported failure was a build-time `InvalidCastException` — so "the query builds and emits `count_if`" is exactly the regression being pinned. No change required.

## Baselines
No diff available. The baselines PR for this PR ([linq2db.baselines#1939](https://github.com/linq2db/linq2db.baselines/pull/1939), branch `baselines/pr_5619`) is **closed** and its branch pruned, so there's nothing to compare right now. linq2dbot reported that baselines changed, and `test-all` was re-triggered today — a fresh baselines PR should regenerate; re-check it before merge.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed an `InvalidCastException` thrown when an aggregate — a custom `[Sql.Extension(IsAggregate = true)]` function (e.g. `count_if`) or an analytic `Average(…, Sql.AggregateModifier)` overload — appeared in one branch of a set operation (`UNION ALL` etc.) against a constant in the other branch. Such queries now build correctly. ([#5619](https://github.com/linq2db/linq2db/pull/5619))


_(none)_


---
<sub>Generated from commit `ac79384`.</sub>

PR #5621, opened 2026-06-16
Add --provider and --test-progress test CLI options

sdanyliv:
## Summary

Adds two Microsoft.Testing.Platform command-line options to the test executables, via an `ICommandLineOptionsProvider` registered through a `TestingPlatformBuilderHook` (the mechanism NUnit itself uses):

- **`--provider <names>`** — run only the named provider(s). For the main tests it **replaces** `TestConfiguration.UserProviders`, so any provider with a connection string in `DataProviders.json` / `UserDataProviders.json` runs without editing the enabled list. EF Core tests **intersect** their curated `EFProviders` set (an EF-unsupported provider can't be selected). Repeatable; comma- or space-separated.
- **`--test-progress[=<dir|*.json>]`** — write the live progress heartbeat. **Replaces** the `LINQ2DB_TEST_PROGRESS` environment variable, which is removed; this option is now the sole control.

Together these let one fully-populated `UserDataProviders.json` drive many provider-scoped runs — including several in parallel, one distinct provider (= distinct database) each.

## Usage

Scope a run to a single provider (no `UserDataProviders.json` edit needed — the provider only needs a connection string):

```bash
dotnet test --project Tests/Linq/Tests.csproj -f net10.0 --settings .runsettings --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest" --provider PostgreSQL.18 --test-progress
```

Run several providers in parallel — build once, then launch the test executable directly with **one distinct provider per run** (each hits its own database):

```bash
dotnet build Tests/Linq/Tests.csproj -c Debug -f net10.0
.build/bin/Tests/Debug/net10.0/linq2db.Tests.exe --provider PostgreSQL.18 --test-progress --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest"
.build/bin/Tests/Debug/net10.0/linq2db.Tests.exe --provider MySql.8.0 --test-progress --filter "FullyQualifiedName~CreateData.CreateDatabase|FullyQualifiedName~MyTest"
```

Each process writes its own `.build/.claude/test-progress.<tfm>.<pid>.json`; read any of them with `pwsh -NoProfile -File .claude/scripts/test-status.ps1`.

## Notes

- `test-runner` always passes `--test-progress`; `/test-progress` is now a read-only status reporter; `/test` no longer toggles a session env var.
- Options are read from `Environment.GetCommandLineArgs()`, so they apply during NUnit discovery (where `TestContext.Parameters` isn't available) and are TFM-agnostic. The hook is scoped to C# test projects (the F# EF project doesn't reference `Tests.Base`).
- Separate commit fixes a pre-existing Release-mode analyzer error (MA0154) in `YdbTransientExceptionDetectorTests`, unrelated to the feature — isolated for easy review/drop.

## Test plan

- Full solution builds clean in **Release** (analyzers + banned-API) and across TFMs incl. **net462**.
- `--provider SQLite.MS` → 1 provider; `+ SQLite.Classic` → 2; non-enabled `SqlServer.2022.MS` runs (replace verified).
- `--test-progress` writes a per-PID heartbeat; env var set + no flag → no heartbeat (removal verified). `--help` lists both options.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -151,6 +151,7 @@
 
 	<ItemGroup Label="Testing">
 		<PackageVersion Include="Shouldly"                                              Version="4.3.0"                        />
+		<PackageVersion Include="Microsoft.Testing.Platform"                            Version="2.2.3"                        />
 		<PackageVersion Include="Microsoft.Testing.Platform.MSBuild"                    Version="2.2.3"                        />
 		<PackageVersion Include="Microsoft.Testing.Extensions.TrxReport"                Version="2.2.3"                        />
 		<PackageVersion Include="Microsoft.Testing.Extensions.HangDump"                 Version="2.2.3"                        />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -39,6 +39,8 @@
 		<AspNetCoreLegacyVersion>2.3.10</AspNetCoreLegacyVersion>
 		<!--bundled version for OpenTelemetry packages-->
 		<OpenTelemetryVersion>1.15.3</OpenTelemetryVersion>
+		<!--bundled version for Microsoft.Testing.Platform / Extensions packages-->
+		<MicrosoftTestingVersion>2.2.3</MicrosoftTestingVersion>
 	</PropertyGroup>
 
 	<ItemGroup Label="Build: Analyzers and Tools">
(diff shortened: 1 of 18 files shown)


Directory.Packages.props
```diff
@@ -151,6 +151,7 @@
 
 	<ItemGroup Label="Testing">
 		<PackageVersion Include="Shouldly"                                              Version="4.3.0"                        />
+		<PackageVersion Include="Microsoft.Testing.Platform"                            Version="2.2.3"                        />
```
MaceWindu: extract version number to property

MaceWindu (approved):

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Env-var removal is complete — no `LINQ2DB_TEST_PROGRESS` references remain in code, scripts, or docs at HEAD.
- [x] Static-init ordering in `Tests/Base/TestConfiguration.cs` is sound: `ProviderOverride` (field initializer) runs before `EFProviders`, per C# textual-order field-initializer rules. REPLACE (main) vs INTERSECT (EF) split and the no-connection warning are correct.
- [x] Option parsing previously matched only the exact `--option value` token, silently ignoring the inline `--option=value` form; fixed in `e276c623b` (both `GetValues` and `GetSingle` now accept it). The `MicrosoftTestingVersion` property extraction is also addressed (commit `a66e457`).

This is a static review — no local test run was performed (a Release build of `Tests.Base` was confirmed clean).

## Out-of-scope observations

- **F# EF test project would reject `--provider` / `--test-progress` if ever invoked with them** — The `TestingPlatformBuilderHook` registration in `Tests/linq2db.BasicTestProjects.props` is gated to `'$(MSBuildProjectExtension)' == '.csproj'`, so `Tests.EntityFrameworkCore.FSharp` imports the props but never registers the option provider. The MTP runner rejects unrecognized command-line options, so invoking that project with `--provider` or `--test-progress` would fail with an unknown-option error. The current EF-matrix flow only runs the four C# EF projects, so this is unreachable through the documented workflow — but a uniform CI step passing `--test-progress` to every test project (or a contributor running the F# project directly) would hit it. Worth a one-line note in the props comment.

## Baselines

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `e276c62`.</sub>

PR #5622, opened 2026-06-17
test infra: thread-safe shared Parent/Child test data model

MaceWindu:
Pre-flight extracted from #5614 (parallel test execution).

The `Parent` / `Parent1` / `Parent5` / `Child` / `GrandChild` / `GrandChild1` lists in `TestBase.Tables` form one cross-referenced in-memory graph (`Parent.Children` → `Child.Parent` → …). They were built piecemeal by each getter, which **published its backing field before fully wiring associations** to break the circular `Parent`↔`Child` init dependency. That's fine under serial execution; once test cases of a shared fixture run concurrently (the upcoming parallel-test execution) a second case can observe a list whose elements aren't wired yet (`null .Children` → NRE).

This builds the whole graph once in a single `EnsureModel()` under a **double-checked lock**:

- a `volatile bool _modelLoaded` keeps the lock off the hot path — it's taken only until the graph is built, then every subsequent access is a lock-free flag read;
- the wiring reads the backing fields directly (not the getters), so the cyclic init never re-enters a getter;
- one `TestDataConnection` loads all the raw lists in a single pass.

No behaviour change for serial runs. Minor: `GrandChild1` / `Parent5` now load together with the rest of the graph (the cyclic `Parent`/`Child`/`GrandChild` trio already co-loaded) — tiny seeded tables, so the extra query is negligible.

Test-infra only — no product code touched.

### DIFF
--- a/Tests/Base/TestBase.Tables.cs
+++ b/Tests/Base/TestBase.Tables.cs
@@ -98,134 +98,98 @@ namespace Tests
 
 		#region Parent/Child Model
 
-		private   List<Parent>?      _parent;
-		protected IEnumerable<Parent> Parent
+		// The Parent/Parent1/Parent5/Child/GrandChild/GrandChild1 lists form one cross-referenced
+		// in-memory graph (Parent.Children -> Child.Parent -> ...). They used to be built piecemeal by
+		// each getter, publishing a half-built backing field before wiring associations to break the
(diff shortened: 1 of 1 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `80e39ee`.</sub>

PR #5623, opened 2026-06-17
Make the LINQ expression mapping registry thread-safe

MaceWindu:
Pre-flight extracted from #5614 (parallel test execution).

### Problem

`Expressions.MapMember` / `MapBinary` mutate the shared static `Members` and `_binaries` dictionaries (`Add`, `dic[key] = …`) with **no synchronization**, while `ConvertMember` / `ConvertBinary` read them **lock-free** on the query-compilation path. The registry's documented contract was "configure at application init, single-threaded" — but concurrent registration, or a registration racing a concurrent read, corrupts a plain `Dictionary<,>`.

### Fix

Make both levels `ConcurrentDictionary`:

- the outer map (provider → inner map) and each inner map (member/binary key → expression);
- reads stay **lock-free** — every access is `TryGetValue` / `[""]`; nothing enumerates the dictionaries, so there's no read-side change beyond the type;
- writes use `GetOrAdd` instead of the racy `TryGetValue`-then-`Add`.

New registrations are very rare (application initialization), so the marginally higher `ConcurrentDictionary` read cost is negligible — and far cheaper than locking the per-query read path would be.

`LoadMembers` keeps its large plain-`Dictionary` initializer (and DEBUG validation) untouched, snapshotting into `ConcurrentDictionary` at the `return`. The `<remarks>Note that method is not thread safe …</remarks>` lines on `MapBinary` (now inaccurate) are updated.

No public API change. Builds clean on Release (analyzers) and netstandard2.0.

### DIFF
--- a/Source/LinqToDB/Linq/Expressions.cs
+++ b/Source/LinqToDB/Linq/Expressions.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Data.Linq;
@@ -66,8 +67,7 @@ namespace LinqToDB.Linq
 		{
 			memberInfoWithType = NormalizeMemeberInfo(memberInfoWithType);
 
-			if (!Members.TryGetValue(providerName, out var dic))
-				Members.Add(providerName, dic = new Dictionary<MemberHelper.MemberInfoWithType,IExpressionInfo>());
+			var dic = Members.GetOrAdd(providerName, static _ => new());
 
 			var expr = new LazyExpressionInfo();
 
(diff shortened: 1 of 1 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- The LINQ expression mapping registry (`Expressions.MapMember` / `MapBinary`) is now thread-safe. Registering a custom member/binary mapping concurrently with query compilation (or from multiple threads) could previously corrupt the internal registry; registrations and lookups are now safe under concurrent access. ([#5623](https://github.com/linq2db/linq2db/pull/5623))


_(none)_


---
<sub>Generated from commit `6f1055f`.</sub>

Issue #1991, opened 2019-11-28
Linq2Db is trying to do rollback on "invalid" MySQL connection


erikfroseth:
Hi,

When certain error happens inside a MySQL transaction, it seems like Linq2Db tries to do a rollback on an invalid connection. It throws the following exception: 

> Connection must be valid and open to rollback transaction

I'm not 100% sure what the correct behavior is here, but the exception seems odd. Here is a reproducible case:

```
# Set up a test schema/table.
CREATE SCHEMA test123;
USE test123;
CREATE TABLE my_table (column INTEGER);
```
```
  public class Program
  {
    [Table("my_table")]
    public class MyTable
    {
      [Column("column")]
      public int Column { get; set; }
    }

    public class Model : DataConnection
    {
      public Model(string connectionString)
        : base(new LinqToDB.DataProvider.MySql.MySqlDataProvider(),
            connectionString)
      { }
      public ITable<MyTable> TheTable { get { return GetTable<MyTable>(); } }
    }

    private static string GetConnectionString()
    {
      var connstring = new MySql.Data.MySqlClient.MySqlConnectionStringBuilder();
      connstring.UserID = "yourusername";
      connstring.Server = "yourserver";
      connstring.Database = "test123";
      connstring.Password = "yourpassword";
      connstring.DefaultCommandTimeout = 5;
      return connstring.ConnectionString;
    }

    static void RunTest()
    {
      var lockTablesSignal = new AutoResetEvent(false);
      var thread1 = new Thread(() =>
      {
        using (var connection = new Model(GetConnectionString()))
        {
          connection.Execute("LOCK TABLE my_table WRITE;");

          // Signal the other thread that it can start.
          lockTablesSignal.Set();
          Thread.Sleep(TimeSpan.FromSeconds(10));
        }
      });

      var thread2 = new Thread(() =>
      {
        try
        {
          using (var connection = new Model(GetConnectionString()))
          using (var transaction = connection.BeginTransaction())
          {
            var data = connection.TheTable.ToList();
          }
        }
        catch (Exception ex)
        {
          // Observe the exception here!
        }
      });

      thread1.Start();

      // Wait for the signal from thread 1 before launching thread 2.
      lockTablesSignal.WaitOne();
      thread2.Start();

      thread1.Join();
      thread2.Join();
    }


    static void Main(string[] args)
    {
      RunTest();
    }
```

MaceWindu:
[...]

PR #5624, opened 2026-06-17
Fix #195, #4646: automatic F# 'T option type mapping

### LINKED ISSUE #195: F# option types are not supported
FSharp has the concept of `Option` types, which can hold the value of `None|Some<T>`. They make handling values that can be null very easy:
```f#
let f = fun (x : string option) -> 
    match x with 
    | None -> "None"
    | Some(x) -> x

f(Some("foo")) //-> "foo"
f(None) //-> "None"
```

What I am thinking is adding support for colums being typed as `Option<T>`, and they are then implictly treated as a `<Nullable>` column. If the value is null in the database, they are set to `None`, if its not null they are set to `Some(Value)`.

See pr #194 for tests exercising this.

### LINKED ISSUE #4646: F# int option type with None value is stored as 0
Hello,
I've been using linq2db with EF Core for a few months now. I just encountered an error where `int option` types with `None` values get translated to 0 in the database. I find this weird because I've been using `string option` types in the models for the entities we are using, and they behave as expected, `None` values are stored as `null` values in the database.

I'm using:
linq2db 4.3.0
linq2db.EntityFrameworkCore 6.10.0
PostgreSQL 14.3

MaceWindu:
## Summary

Adds automatic mapping of F# `'T option` columns to the `linq2db.FSharp` package. After
`.UseFSharp()`, option columns round-trip with **no** manual `MappingSchema` configuration:

- `Some v` ↔ stored value, `None` ↔ `NULL`.
- Value-typed options (e.g. `int option`) route through `Nullable<'T>`, so `None` stores as
  `NULL` instead of the value-type default — fixes #4646 (`int option` `None` stored as `0`).

Closes #195
Fixes #4646

## How it works

A metadata reader (`FSharpOptionSupport` / `FSharpOptionMetadataReader`), combined into the
connection's mapping schema by `UseFSharp()`, supplies for every `'T option` member:

- `ScalarTypeAttribute` on `FSharpOption<_>` so option members are treated as **columns**;
- `DataTypeAttribute` derived from the element's DB type for correct DDL;
- `ValueConverterAttribute` with an expression-tree `ValueConverter` doing the `None`↔`NULL` /
  `Some`↔value conversion (value-typed elements use `Nullable<'T>` as the provider type).

No core `LinqToDB` changes; the new types are `internal`, so there is **no public-API surface change**.

## Tests

- New `Issue4646.fs` round-trips `int option` / `string option` (`Some` and `None`) via `CreateLocalTable`.
- The `#195` acceptance test (`LoadSingleWithOptions`) is un-gated and switched to the automatic path;
  an explicit test (`LoadSingleWithExplicitOptionsMapping`) retains manual `MappingSchema` registration
  to confirm it still coexists with auto-mapping.
- Verified green on SQLite (Classic + MS), PostgreSQL 16, and remote LinqService variants
  (201/201 F# tests). Multi-TFM Release build (`net462`, `netstandard2.0`, `net8.0`–`net10.0`) is clean.

## Not included

EF Core F# option columns flow through `linq2db.EntityFrameworkCore`'s EF value-converter translation
(`EFCoreMetadataReader`) — a separate path. Extending the option metadata reader there (so the EF model
would not need an explicit `HasConversion`) is a possible follow-up.

## Follow-up: schema-priority fix

After syncing with `master`, `ConcurrencyTests` failed across SqlServer (rowversion → `VarBinary`),
Oracle and SqlCe (`Binary(16)` → `BLOB`). Root cause: `UseFSharp()` combined the option schema via
`UseAdditionalMappingSchema`, which adds it at **higher** priority. That schema's embedded default
attribute reader then shadowed users' fluent column metadata — and since `EntityDescriptor` keeps the
first `ColumnAttribute` per member, an explicit fluent `DataType` (e.g. rowversion / `Binary`) on an
attributed entity was dropped in favour of the plain `[Column]` attribute. (The test base applies
`UseFSharp()` to every context, so this surfaced in non-F# tests too.)

Fixed by combining the option schema as a **lower-priority fallback** (`CombineSchemas(existing,
optionSchema)`): auto `'T option` mapping only fills in members the user has not mapped, and explicit
mappings keep priority. Verified: `TestDbStrategy*` on SqlServer 2022 + full `ConcurrencyTests` +
`FSharpTests` = 175/175 green.

## Follow-up: review (2026-06-20)

- **Regression test for the schema-priority fix.** Added `OptionMappingPrecedence.fs`
  (`FSharpTests.OptionMapping_ExplicitDataTypePreserved`): pins an explicit fluent `DataType`
  (`VarChar`) on an option member that also carries a plain `[<Column>]`, and asserts the entity
  descriptor keeps `VarChar`. Verified red→green — with the option schema combined at higher
  priority the column resolves to the derived `NVarChar` instead.
- **Documented the option-column DB-type limitation.** Auto-mapped `'T option` columns derive their
  DB type from `MappingSchema.Default` (a metadata reader has no connection/provider-schema access),
  so provider-specific / user-custom element overrides aren't applied automatically (e.g. `string
  option` → `NVarChar`, not a provider's preferred `VarChar`). Noted in the reader comment + readme;
  round-trip is unaffected — set `DataType` explicitly when a provider-faithful type is needed.

## Follow-up: review (2026-06-22)

- **F# `'T voption` support.** The value converter and metadata reader now also auto-map struct
  value-options (`FSharpValueOption<_>`), not just reference options.
- **Scalar-element gate.** Only options whose element is a scalar type are auto-mapped, so an
  option over a complex/entity element is no longer force-scalarized.
- **`Nullable<_>` element guard.** A `Nullable<int> option` no longer builds an invalid
  `Nullable<Nullable<int>>` provider type.
- **Nullable option columns.** Auto-mapped option columns are marked `CanBeNull` - a struct
  value-option is a non-nullable value type, so the column would otherwise emit `NOT NULL` and
  reject the "none" case.
- New F# round-trip tests for `Nullable<_>`-element options and value-options.


Source/LinqToDB.FSharp/FSharpOptionSupport.fs
```diff
@@ -0,0 +1,111 @@
...
+        | _                    -> typeof<obj>
+
+    interface IMetadataReader with
+        // Mark FSharpOption<_> as a scalar type so option members are treated as columns rather than
+        // nested entities; the per-member ValueConverter (below) supplies the actual conversion.
+        member _.GetAttributes(_type: Type) =
+            if FSharpOptionSupport.IsOption _type then
+                [| ScalarTypeAttribute() :> MappingAttribute |]
+            else
+                Array.empty<MappingAttribute>
```
MaceWindu, line 106: **Minor · MIN001** — The type-level reader marks *every* `FSharpOption<_>` as a scalar type regardless of element type. For an option over a complex/entity type (`MyEntity option`), this now force-scalarizes the member: the member-level reader derives `MappingSchema.Default.GetDbDataType(MyEntity)` (typically `Undefined`) and attaches a converter treating the whole entity as one column — a behavior change that can break an unmapped `Entity option` member that previously wasn't auto-scalarized. The element-wrapping converter is only meaningful when the element is itself a scalar/column type.

Fix: gate the scalar treatment (in **both** `GetAttributes` overloads) on the element type actually being a scalar — e.g. a value type or `string`, or `MappingSchema.Default.GetDbDataType(elementType).DataType <> DataType.Undefined` — rather than blanket-applying it to any option.

— ✓ Fixed in 8739088

Source/LinqToDB.FSharp/FSharpOptionSupport.fs
```diff
@@ -0,0 +1,111 @@
...
+
+    static let cache = ConcurrentDictionary<Type, IValueConverter>()
+
+    // Builds the bidirectional value converter for a closed 'T option type using explicit expression
+    // trees. F# None is represented as a null reference, which makes IsSome / None construction trivial.
+    // For a value element 'a the DB-facing type is Nullable<'a> so that None stores as NULL rather than
+    // default('a) - the root cause of issue #4646 (int option None stored as 0).
+    static let build (optionType: Type) : IValueConverter =
+        let elementType  = optionType.GetGenericArguments().[0]
+        let providerType = if elementType.IsValueType then typedefof<Nullable<_>>.MakeGenericType(elementType) else elementType
```
MaceWindu: **Suggestion · SUG001** — When the element type is itself `Nullable<'a>` (expressible as `Nullable<int> option`), `elementType.IsValueType` is true, so `providerType` becomes `Nullable<Nullable<int>>` and `MakeGenericType` throws `ArgumentException` at converter-build time. Pathological and non-idiomatic (`int option` already gives the desired semantics) but a reachable crash path.

Fix: unwrap an already-`Nullable<>` element before re-wrapping.

```suggestion
        let elementType  = optionType.GetGenericArguments().[0]
        let providerType = if elementType.IsValueType && Nullable.GetUnderlyingType(elementType) = null then typedefof<Nullable<_>>.MakeGenericType(elementType) else elementType
```

— ✓ Fixed in 8739088

Source/LinqToDB.FSharp/FSharpOptionSupport.fs
```diff
@@ -0,0 +1,111 @@
...
+                // Limitation: a metadata reader has no access to the connection/provider mapping schema, so
+                // this resolves against MappingSchema.Default only - provider-specific or user-custom DB-type
+                // overrides for the element type are not honored (e.g. 'string option' maps to NVarChar, not
+                // a provider's preferred VarChar). Set the column's DataType explicitly (attribute or fluent
+                // mapping) when a provider-faithful type is required.
+                let dbDataType  = MappingSchema.Default.GetDbDataType(elementType)
+                [|
+                    DataTypeAttribute(dbDataType.DataType, DbType = dbDataType.DbType) :> MappingAttribute
+                    ValueConverterAttribute(ValueConverter = FSharpOptionSupport.GetConverter mt) :> MappingAttribute
+                |]
```
MaceWindu: **Minor · MIN002** — `DataTypeAttribute` carries only `DataType` + `DbType`, never `Length`/`Precision`/`Scale`, and because it sets `DataType` non-`Undefined`, `ColumnDescriptor` skips `CalculateDbDataType` and never consults the provider schema. So a `decimal option` column gets a bare `Decimal` → SQL Server's engine-default `DECIMAL(18,0)`, silently truncating scale, whereas the non-option `decimal` sibling gets `DECIMAL(18,10)` (and 28/10, 38/10 on Oracle/SAP HANA). `char option` likewise loses its `NChar` length. `Issue4646.fs` only exercises `int`/`string` option, so this round-trips silently wrong and untested.

Fix: strengthen the readme/XML-doc limitation note to call out precision/scale/length loss (not only the benign `string → NVarChar` naming), and add a `decimal option` round-trip test (non-zero scale, asserting the read-back is not truncated). Provider-faithful precision can't be expressed through `DataTypeAttribute` — it would need a `ColumnAttribute` carrying `Precision`/`Scale`, or document that decimal/datetime/char options require an explicit mapping.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] No public-API surface change. The new `FSharpOptionSupport` / `FSharpOptionMetadataReader` types are `internal` and `UseFSharp()` keeps its signature — no `PublicAPI.*` or API-baseline refresh needed.
- [ ] Test baselines changed (122 new SQL files across 61 providers, all additions). Merge/close the companion baselines PR after this PR merges.

## Out-of-scope observations

- **F# `ValueOption` (`'T voption`) is silently unsupported** — `FSharpOptionSupport.IsOption` matches only `FSharpOption<_>`, so struct-option (`voption`) columns get no auto-mapping and no diagnostic. Consistent with the PR's stated `'T option` scope; worth a one-line readme note or a follow-up if voption is wanted.
- **`Some null` cannot be distinguished from `None` for reference elements** — a reference-typed `Some (null:string)` round-trips as `None`, since both store as a NULL column. Inherent to representing reference options via a nullable column; unusual in idiomatic F# but worth documenting.
- **Auto-derived column DB type resolves against `MappingSchema.Default` only** — as the in-code comment documents, the reader has no access to the connection/provider schema, so provider-specific element overrides aren't honored (`string option` → `NVarChar` everywhere; on Oracle `NVarChar2(2000)` vs plain string's `VarChar2`, on PostgreSQL vs `text`). Round-trip is unaffected and the explicit-mapping escape hatch is verified by `OptionMapping_ExplicitDataTypePreserved`. MIN002 covers the silent facet-loss subset of this.
- **`CombineSchemas` with the non-lockable option schema isn't cached** — `optionMappingSchema` is a plain (non-lockable) `MappingSchema`, so the combined-schema cache never engages and each `UseFSharp()` rebuilds a combined schema. Bounded impact (static singleton, connection-setup not query path); locking the shared option schema once would let the combine cache engage.
- **Build note — `nonNull` intrinsic under `LangVersion 9`** — `FSharpOptionSupport.fs:28-29` use the FSharp.Core 9 `nonNull` intrinsic while the project pins `LangVersion 9` (and netfx uses an older FSharp.Core). The PR reports a clean multi-TFM Release build and `test-all` has run, so this is informational rather than a concern.

## Baselines
Delta: [linq2db.baselines PR #1943](https://github.com/linq2db/linq2db.baselines/pull/1943) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5624)

122 new SQL baseline files, **0 modified, 0 deleted** — all additions for two new tests across 61 providers:

### `Issue4646_OptionRoundtrip` (61 providers)
INSERT of two rows (one with values, one all-NULL) into a local `Issue4646Table`, then two `WHERE Id = N` reads. `None` correctly renders as a NULL parameter on the second INSERT across the family. Per-provider variation is purely parameter syntax / quoting / paging (e.g. Oracle 11 `ROWNUM <= 2` vs Oracle 12+ `FETCH NEXT 2 ROWS ONLY`; ClickHouse inlines literals with no `DECLARE`). No structural concern.

### `LoadSingleWithExplicitOptionsMapping` (61 providers)
Single parameterized `SELECT` of the `Person` columns. Confirms the explicit user mapping survives auto-option-mapping (the lower-priority-fallback fix).

- **Cross-provider note (not a finding):** Oracle 12-23 emit `p."PersonID" as ID` and SqlCe `[p].[PersonID] as [ID]`, while Oracle 11 and all other providers project `PersonID` without an alias. Verified against the actual `.sql` files. This is pre-existing provider rendering of the explicit `ID`-member ↔ `PersonID`-column mapping (Oracle 11 lacks the alias because it uses the `ROWNUM` wrapping path), surfaced by a brand-new baseline — not introduced by this PR's option code.

Source/LinqToDB.FSharp/FSharpOptionSupport.fs
MaceWindu: **Partial fix · MIN002 → MIN003** — The limitation note was strengthened to cover the DB-type-**name** case (`string option` → `NVarChar`), but the precision/scale/length loss this finding flagged is still present and still undocumented.

Auto-mapped option columns receive only `DataType` + `DbType` from `MappingSchema.Default.GetDbDataType(elementType)` (`FSharpOptionSupport.fs:124-125`); `Length`/`Precision`/`Scale` are never copied. And because `DataType` is now non-`Undefined`, `ColumnDescriptor` skips `CalculateDbDataType`, so the provider schema can't supply them either. A `decimal option` therefore creates as the engine default `DECIMAL(18,0)` (scale silently truncated), and a `char option` loses its length. The new tests still only exercise `int` / `string` / `Nullable<int>` / `voption`, so this round-trips silently wrong and is unverified.

Fix: extend the readme limitation note (`readme.md:16`) and the matching XML-doc note to call out length/precision/scale loss (not only the benign `string → NVarChar` naming), and add a `decimal option` round-trip test with a non-zero scale (e.g. `Some 12.34m`) asserting the read-back value is not truncated.

MaceWindu (commented):
> [!IMPORTANT]

## Verification update — 2026-06-23, against HEAD `8739088`

Re-verified the three prior findings against current HEAD.

### Fixed
- [x] **MIN001** — Scalar gate for `'T option` over a complex/entity element. `IsScalarOption` (`IsOption && MappingSchema.Default.IsScalarType elementType`) now gates **both** `IMetadataReader.GetAttributes` overloads, so an option over a non-scalar/entity element is no longer force-scalarized (falls through to `Array.empty`).
- [x] **SUG001** — `Nullable<>`-element option crash. The provider type is now computed with `wrapInNullable = elementType.IsValueType && isNull (Nullable.GetUnderlyingType elementType)`, so a `Nullable<int> option` keeps `Nullable<int>` rather than building `Nullable<Nullable<int>>`. Covered by the new `Nullable<int> option` round-trip test.

### Partially fixed
- [~] **MIN002** — Auto-mapped option-column DB-type limitation. The doc note was strengthened for the DB-type-**name** case (`string option` → `NVarChar`), but the precision/scale/length truncation this finding flagged is still present and undocumented, and no `decimal option` test was added. Carried forward as **MIN003** on the original thread.

## Baselines
Delta: [linq2db.baselines PR #1943](https://github.com/linq2db/linq2db.baselines/pull/1943) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5624)

Now **244 new SQL baseline files, 0 modified, 0 deleted** — all additions for **four** new tests across 61 providers (up from 122/two tests at the prior review, because the fix commit added the voption and `Nullable<>`-element round-trip tests):

- `Issue4646_OptionRoundtrip` — INSERT a `Some` row + an all-NULL `None` row, then read each back. `None` renders as a NULL parameter across the family.
- `LoadSingleWithExplicitOptionsMapping` — single parameterized `SELECT` confirming the explicit user mapping survives auto-option-mapping.
- `Option_NullableElementRoundtrip` — `Nullable<int> option` Some(42)/None round-trip.
- `Option_ValueOptionRoundtrip` — `voption` round-trip; SQL is structurally identical to the reference-option test, confirming voption and option map identically at the SQL level.

Per-provider variation is purely paging syntax / quoting / parameter style (Oracle 11 `ROWNUM <= 2` vs 12+ `FETCH NEXT 2 ROWS ONLY`; ClickHouse inlines literals; Oracle 12+/SqlCe emit the pre-existing `PersonID as ID` alias on the `ID`-member mapping). No structural concern, no cross-provider anomaly.

Source/LinqToDB.FSharp/FSharpOptionSupport.fs
MaceWindu: Fixed in b81434c69. The reader no longer sets an explicit DataType; with none set, ColumnDescriptor now resolves the column's DB type from the value converter's provider type against the connection's provider-inclusive schema (#5645), so decimal precision/scale and string length are preserved instead of being dropped by MappingSchema.Default. A `decimal option` now maps to the provider's `decimal(18,10)` rather than a bare `decimal(18,0)`.

Verified red->green on SQL Server 2022 with a new `Option_DecimalRoundtrip` test: `Some 12.34m` read back as `12` before the fix, `12.34` after. The readme limitation note was corrected accordingly (the caveat no longer applies).

sdanyliv (dismissed):

Source/LinqToDB.FSharp/DataOptionsExtensions.fs
```diff
@@ -1,14 +1,31 @@
...
 type Methods() =
+    // Shared schema carrying the F# option metadata reader; combined into the user's options below.
+    static let optionMappingSchema =
+        let ms = MappingSchema()
+        ms.AddMetadataReader(FSharpOptionMetadataReader())
+        ms
+
     /// <summary>Enables support for F#-specific features.</summary>
-    /// <remarks>Currently it is limited to record types support, but we plan to add more support in future releases.</remarks>
+    /// <remarks>Adds support for F# record types and automatic mapping of F# <c>'T option</c> columns.</remarks>
```
MaceWindu: convert to list and add voption support

Source/LinqToDB.FSharp/FSharpOptionSupport.fs
```diff
@@ -0,0 +1,128 @@
...
+        | _ -> raise (InvalidOperationException $"Failed to create F# option value converter for '{optionType}'")
+
+    /// Returns <c>true</c> when <paramref name="t"/> is <c>FSharpOption&lt;_&gt;</c> or <c>FSharpValueOption&lt;_&gt;</c>.
+    static member IsOption(t: Type) =
+        t.IsGenericType &&
+        (let d = t.GetGenericTypeDefinition() in d = typedefof<_ option> || d = typedefof<_ voption>)
+
+    /// Returns <c>true</c> when <paramref name="t"/> is an option type whose element is a scalar (column)
+    /// type. An option over a complex/entity element is not treated as a column.
+    static member IsScalarOption(t: Type) =
```
MaceWindu, line 81: **Suggestion · SUG003** — `IsScalarOption` gates auto-mapping on `MappingSchema.Default.IsScalarType(elementType)`. An option whose element is scalar *only* in a specific provider's schema (e.g. a provider-native type like `NpgsqlPoint`) isn't in `Default`, so it's treated as a complex/entity type and won't auto-map. Standard scalar elements (primitives, `string`, `decimal`, `DateTime`, `Guid`, enums) are unaffected. Root cause: `IMetadataReader` has no provider-schema access at attribute-resolution time — not fixable without a schema-aware reader.

Fix: document this provider-only-scalar limitation alongside the existing "only scalar-element options are auto-mapped" note (readme + XML doc).

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] No public-API surface change. `FSharpOptionSupport` / `FSharpOptionMetadataReader` are `internal` and `UseFSharp()` keeps its signature — no `PublicAPI.*` / baseline refresh needed. Latest commit relies on merged #5645 for provider-faithful facets.
- [ ] Test baselines: **305 new SQL files, 0 modified, 0 deleted** across 61 providers × 5 tests. `None`→`NULL` renders uniformly; per-provider quirks (Access decimal-as-AnsiString, DB2/SQLite blank type name, YDB fixed-precision literal + `Value_1` alias) all verified pre-existing on master. Merge/close the companion baselines PR after this merges.

Code correctness, thread-safety, performance, and the SQL/provider dimension are clean: the DB type derived per option element equals that of the corresponding plain nullable column, and facet fidelity (decimal scale, string length) is delegated to merged #5645 and covered by the decimal-scale round-trip test.

## Suggestions
- [x] **SUG002** — No test covered the documented negative branch: an option over a complex/entity element is intentionally *not* auto-mapped (`IsScalarOption` returns false, so no column/converter attributes are emitted). Every positive branch (int/string option, voption, `Nullable<_>` element, decimal scale) had a round-trip test, but this branch had none. Added regression test `Option_ComplexElementNotScalarized`: asserts a scalar-element option carries the F# value converter while a complex-element option does not. Verified green on SQLite.

## Baselines
Delta: [linq2db.baselines PR #1991](https://github.com/linq2db/linq2db.baselines/pull/1991) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5624)

**305 new SQL baseline files, 0 modified, 0 deleted** — all additions for five new tests across 61 providers (`Issue4646_OptionRoundtrip`, `LoadSingleWithExplicitOptionsMapping`, `Option_DecimalRoundtrip`, `Option_NullableElementRoundtrip`, `Option_ValueOptionRoundtrip`). Every body is a plain `DECLARE`/`SET`/`INSERT`/`SELECT` — no `CAST`/`COALESCE`/`CASE`; `None` renders as `NULL` across the family. No cross-provider anomaly.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

F# `'T option` and `'T voption` (value-option) columns now map automatically in the `linq2db.FSharp` package — after `.UseFSharp()`, option columns round-trip with no manual `MappingSchema` configuration: the *some* case (`Some`/`ValueSome`) stores the value and the *none* case (`None`/`ValueNone`) stores `NULL`. Value-typed options such as `int option` route through `Nullable<'T>`, so *none* is stored as `NULL` instead of the value-type default (previously an `int option` `None` was stored as `0`). Only options over a scalar element type are auto-mapped, and the column's DB type — including facets like decimal precision/scale and string length — is resolved from the element type against the connection's provider schema, so it matches the equivalent non-option column. Auto-mapping is a lower-priority fallback and never overrides mappings you've configured explicitly.


Automatic mapping of F# `'T option` and `'T voption` columns after `.UseFSharp()`, with *none* correctly stored as `NULL`.


---
<sub>Generated from commit `fc8a4dc`.</sub>

PR #5627, opened 2026-06-17
Fix #5598: F# { record with ... } update sets only changed columns

### LINKED ISSUE #5598: F# "{ record with ... }" update emits every column (including the PK) instead of only the changed one
An F# record-copy update (`{ record with X = ... }`) is translated to an UPDATE that sets **every** column - including the primary key - rather than only the changed column. YDB rejects assigning a PK column in SET.

Expected: the F# update path should set only the changed column(s), matching the C# projection-update behavior.

Surfaced by `Linq/FSharpTests` while completing the YDB provider (#5564).

MaceWindu:
## Summary

Fixes the F# record-copy update bug: `query.Update(predicate, fun r -> { r with Field = v })` was
translated to an UPDATE that set **every** column (including the primary key), which YDB rejects.

Now only the changed column(s) are written.

Closes #5598

## How it works

F# `{ r with Field = v }` lowers to a full record reconstruction — `new R(r.Id, r.Number, v)` — so by
the time linq2db sees it there is no marker distinguishing it from an explicit full construction. The
fix introduces an F# `IQueryExpressionInterceptor` (registered by `UseFSharp`) that:

- **moves the F#-specific block-inlining out of core** (`ExposeExpressionVisitor.VisitBlock`) into the
  F# library, where F# quirks belong; and
- rewrites an F# `Update(predicate, setter)` whose setter is a record construction into the explicit
  partial-update form `query.Where(predicate).Set(x => x.Field, x => v)…​.Update()`, keeping only the
  constructor arguments that are **not** a self-copy `Field = r.Field`.

Detection keys off the constructor arguments (`arg == r.SameMember`), **not** F#'s block-variable
codegen, which varies between compiler releases. Explicit field-by-field constructions (including a
deliberate `Field = r.Field`) are left untouched; the only collateral is that a literal no-op
`{ r with Field = r.Field }` is dropped, which is harmless.

The rewrite produces only standard expression nodes (`Where`/`Set`/`Update`), so it flows through the
translator like any C# partial update. No new core API is introduced, and future F# AST quirks can
extend the same interceptor.

## Tests

- New `Issue5598.fs` / `FSharpTests.Issue5598_UpdateSetsOnlyChangedColumn` (SQLite): asserts the
  generated UPDATE contains the changed column and **not** the unchanged one, and that the row
  round-trips correctly.
- `Issue4132Test2`'s `[ActiveIssue(5598, … YDB …)]` gate is removed; verified passing on YDB.
- Full `FSharpTests` fixture green on SQLite (incl. remote/LinqService) — no regression from moving
  `VisitBlock` out of core. Multi-TFM Release build (`net462`, `netstandard2.0`, `net8.0`–`net10.0`) clean.

## Follow-up commits (review)

- Tighten the empty-change-set fallback comment in `FSharpQueryExpressionInterceptor` to state that a literal no-op `{ r with Field = r.Field }` re-emits the full all-column UPDATE (PK included), instead of describing the no-op as "lost".
- Tighten the SQLite assertion in `Issue5598.fs` to the bracket-quoted `[Text]` form (symmetric with the `[Number]` negative check).
- Add `Issue5598_UpdateSetsOnlyChangedColumnYdb`: the same insert/update/read-back round-trip on YDB (the provider that originally rejected the all-column UPDATE), with the SQLite-specific SQL-shape assertion gated behind an `assertSql` flag.
- Add async coverage (`Issue5598_UpdateSetsOnlyChangedColumnAsync`, SQLite + YDB). `UpdateAsync(predicate, setter)` builds its expression from the **synchronous** `UpdateSetter`/`UpdatePredicateSetter` method infos (async-ness is handled at execution), so the F# record-copy rewrite already covers the async path — the new tests lock that in. No interceptor change needed.
- Scope the SQLite SQL-shape assertion to the SET clause (text before `WHERE`) so the PK (`[Id]`) exclusion is asserted directly — `[Id]` legitimately appears in the WHERE clause, so the prior whole-statement check couldn't assert it.

### Review follow-ups (commit b124678)

- **No-op record copy** (`{ r with Id = r.Id }`): the empty-change-set path no longer re-emits the full all-column UPDATE (PK included, rejected by YDB). It now assigns every non-PK column to its own value, keeping the PK out of SET (supersedes the "no-op is dropped / harmless" note above). The `MappingSchema` is threaded into the rewrite visitor to identify PK columns. New `UpdateNoOpExcludesPrimaryKey` test (SQLite SQL-shape + YDB round-trip).
- **No-predicate overload**: new `UpdateSetsOnlyChangedColumnNoPredicate` test exercises `table.Update(fun r -> {...})` (the `UpdateSetter` branch the predicate tests didn't reach).
- **SET-clause extraction**: extracted `setClauseOf`, which tolerates a WHERE-less UPDATE (the no-predicate case) instead of throwing on `IndexOf` returning -1.
- **`Issue4132Test2`**: now inserts a row and reads it back, so the YDB gate removal is backed by a data-correctness assertion rather than smoke-level execution only.

Verified: `FSharpTests` `Issue5598`/`Issue4132` green on SQLite + YDB.

### DIFF
--- a/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
+++ b/Source/LinqToDB.FSharp/DataOptionsExtensions.fs
@@ -9,4 +9,6 @@ type Methods() =
     /// <remarks>Currently it is limited to record types support, but we plan to add more support in future releases.</remarks>
     [<Extension>]
     static member UseFSharp(options : DataOptions) =
-        options.UseInterceptor FSharpEntityBindingInterceptor.Instance
+        options
+            .UseInterceptor(FSharpEntityBindingInterceptor.Instance)
+            .UseInterceptor(FSharpQueryExpressionInterceptor.Instance)
--- a/Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
+++ b/Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
@@ -12,8 +12,10 @@ open LinqToDB.Internal.Reflection
 ///       { var x = expr1; new type(x, expr2) }  ->  new type(expr1, expr2)
 ///   * turns an F# record-copy update (`q.Update(p, fun r -> { r with Field = v })`) into the explicit
 ///     partial-update form `q.Where(p).Set(x => x.Field, x => v).Update()`, so only the changed columns
-///     are written. A ctor argument that is a self-copy `r.SameField` is dropped (collateral: a literal
-///     `{ r with Field = r.Field }` no-op is lost, which is acceptable).
+///     are written. A ctor argument that is a self-copy `r.SameField` is dropped; when *every* argument
+///     is a self-copy (a literal no-op such as `{ r with Field = r.Field }`), the change set is empty and
+///     the original full-record construction is left as-is, so that no-op input still emits an all-column
+///     UPDATE (PK included) - acceptable, since such input changes no column anyway.
 /// (Kept out of core so F# quirks live in the F# library.)
 type private FSharpRewriteVisitor() =
     inherit ExpressionVisitor()
--- a/Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
+++ b/Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
@@ -1,0 +1,125 @@
+namespace LinqToDB.FSharp
+
+open System.Linq.Expressions
+
+open LinqToDB.Expressions
+open LinqToDB.Interceptors
+open LinqToDB.Reflection
+open LinqToDB.Internal.Reflection
+
+/// Rewrites F#-specific expression-tree shapes the core translator does not understand:
+///   * inlines the unnecessary block F# emits for record construction
+///       { var x = expr1; new type(x, expr2) }  ->  new type(expr1, expr2)
+///   * turns an F# record-copy update (`q.Update(p, fun r -> { r with Field = v })`) into the explicit
+///     partial-update form `q.Where(p).Set(x => x.Field, x => v).Update()`, so only the changed columns
+///     are written. A ctor argument that is a self-copy `r.SameField` is dropped (collateral: a literal
+///     `{ r with Field = r.Field }` no-op is lost, which is acceptable).
+/// (Kept out of core so F# quirks live in the F# library.)
+type private FSharpRewriteVisitor() =
+    inherit ExpressionVisitor()
+
+    // Rewrites an F# Update(... , setter) call into a Where(...)?.Set(...).Update() chain that assigns
+    // only the members whose ctor argument is not a self-copy `row.SameMember`.
+    member private _.RewriteUpdate(mc: MethodCallExpression, hasPredicate: bool) : Expression =
+        let entityType = mc.Method.GetGenericArguments().[0]
+        let source     = mc.Arguments.[0]
+        let setterArg  = mc.Arguments.[mc.Arguments.Count - 1]
+        match setterArg with
+        | :? UnaryExpression as q when q.NodeType = ExpressionType.Quote ->
+            match q.Operand with
+            | :? LambdaExpression as setter when
+                    setter.Parameters.Count = 1
+                    && (setter.Body :? NewExpression)
+                    && FSharpEntityBindingInterceptor.isRecord setter.Body.Type ->
+                let ne = setter.Body :?> NewExpression
+                match FSharpEntityBindingInterceptor.TryMapMembersToConstructor(TypeAccessor.GetAccessor ne.Type) with
+                | Some map ->
+                    let rowParam = setter.Parameters.[0]
+                    let changed  = ResizeArray<MemberAccessor * Expression>()
+                    for i in 0 .. ne.Arguments.Count - 1 do
+                        match map.TryGetValue i with
+                        | true, ma ->
+                            let arg = ne.Arguments.[i]
+                            let isSelfCopy =
+                                match arg with
+                                | :? MemberExpression as me -> me.Member = ma.MemberInfo && obj.ReferenceEquals(me.Expression, rowParam)
+                                | _ -> false
+                            if not isSelfCopy then changed.Add((ma, arg))
+                        | _ -> ()
+
+                    if changed.Count = 0 then
+                        mc :> Expression
+                    else
+                        let mutable chain : Expression = source
+                        if hasPredicate then
+                            chain <- Expression.Call(Methods.Queryable.Where.MakeGenericMethod(entityType), source, mc.Arguments.[1])
+
+                        let mutable first = true
+                        for (ma, valueExpr) in changed do
+                            let x       = Expression.Parameter(entityType, "x")
+                            let extract = Expression.Lambda(Expression.MakeMemberAccess(x, ma.MemberInfo), [| x |])
+                            let update  = Expression.Lambda(valueExpr.Replace(rowParam, x), [| x |])
+                            let setM    = if first then Methods.LinqToDB.Update.SetQueryablePrev else Methods.LinqToDB.Update.SetUpdatablePrev
+                            chain <- Expression.Call(setM.MakeGenericMethod(entityType, ma.Type), chain, Expression.Quote extract, Expression.Quote update)
+                            first <- false
+
+                        Expression.Call(Methods.LinqToDB.Update.UpdateUpdatable.MakeGenericMethod(entityType), chain) :> Expression
+                | None -> mc :> Expression
+            | _ -> mc :> Expression
+        | _ -> mc :> Expression
+
+    override this.VisitBlock(node: BlockExpression) =
+        // block items must be: N assignments to block variables + a result expression
+        if node.Variables.Count > 0
+           && node.Variables.Count + 1 = node.Expressions.Count
+           && obj.ReferenceEquals(node.Result, node.Expressions.[node.Expressions.Count - 1]) then
+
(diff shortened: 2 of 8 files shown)


MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] **New public API.** `LinqToDB.FSharp.FSharpQueryExpressionInterceptor` (type) and its `Instance` member are new public surface. Both carry XML docs and match the sibling `FSharpEntityBindingInterceptor` shape — additive, appropriate for the 6.4.0 milestone.
- [ ] **Core premise checks out.** The rewritten `Where(predicate).Set(…).Update()` chain lowers to the same `SqlUpdateStatement` as the original `Update(predicate, setter)` (`UpdateBuilder` does `BuildWhere` + `ParseSetter` internally), differing only by dropping self-copy columns from SET. The YDB gate removal on `Issue4132Test2` is justified — `{ row with Text = v }` no longer emits the PK in SET. The block-inlining moved out of core is a faithful port of the removed `ExposeExpressionVisitor.VisitBlock`.
- [x] **Two minor findings addressed in follow-up commits on this branch.**
  - The new test's positive assertion now matches the bracket-quoted `[Text]` form (symmetric with the `[Number]` negative check) instead of the looser bare `Text` substring.
  - The empty-change-set fallback comment now states explicitly that a literal no-op `{ r with Field = r.Field }` re-emits the full all-column UPDATE (PK included), rather than describing the no-op as "lost".
- [x] **YDB coverage added.** `Issue5598_UpdateSetsOnlyChangedColumnYdb` runs the insert → update-one-column → read-back round-trip on YDB — the provider that originally rejected the all-column UPDATE — as a data-correctness check. The SQLite-only SQL-shape assertion is now gated behind an `assertSql` flag, since it relies on SQLite bracket quoting. Previously YDB was exercised only as "does not throw" via the now-ungated `Issue4132Test2`.

## Baselines
Delta: [linq2db.baselines PR #1945](https://github.com/linq2db/linq2db.baselines/pull/1945) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5627)

Analysis below reflects the baselines generated from the pre-follow-up commits; the added YDB test will produce one new baseline on the next CI run. 63 SQL baselines changed (5 added, 58 modified, 0 deleted), all accounted for by the PR's in-scope tests except one benign unrelated cleanup.

### `Issue4132Test2` — no-op SET lines removed (57 modified + 1 YDB added)
Every provider previously emitted `Id = <alias>.Id` and `Number = <alias>.Number` before the changed column in the UPDATE SET clause; the rewrite drops both no-op self-assignments. The diff is purely subtractive (nothing added to SET), WHERE unchanged — the correct post-fix shape. The YDB entry is newly added (gate removed) and emits only `Text = 'updated recently'u` in SET (backtick quoting + `u` suffix are YDB-specific, expected).

<details><summary>sample diff (DuckDB)</summary>

```diff
@@ -3,8 +3,6 @@
 UPDATE
 	Issue4132Table
 SET
-	Id = Issue4132Table.Id,
-	"Number" = Issue4132Table."Number",
 	Text = 'updated recently'
 WHERE
 	Issue4132Table."Number" = 1
```
</details>

### `Issue5598_UpdateSetsOnlyChangedColumn` — regression test (4 SQLite providers added)
INSERT (all 3 columns) → `UPDATE … SET [Text] = 'after' WHERE [Id] = 1` (only the changed column) → SELECT read-back. Shape matches the fix's intent.

### Unrelated cleanup
`Issue1174Tests.TestConcurrentSelect` (ClickHouse.Octonica only) drops a duplicate second `SELECT … LIMIT 1` block. Not caused by this PR — the master baseline carried a duplicate-capture artifact and this run captured the clean single-query form. Verified against the branch file (single SELECT at HEAD); benign, but it rides along in the baselines PR.

No regression candidates, no `.sql.other` files, no cross-provider anomalies.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] **New public API.** `LinqToDB.FSharp.FSharpQueryExpressionInterceptor` (type) and its static `Instance` member are new public surface — both XML-documented, matching the sibling `FSharpEntityBindingInterceptor` shape. Additive, appropriate for the 6.4.0 milestone. (F# project, so there is no Roslyn `PublicAPI.*.txt` baseline to update.)
- [ ] **Rewrite premise, core move, and async coverage all confirmed.** The rewritten `Where(p).Set(…).Update()` chain lowers to the same `SqlUpdateStatement` as `Update(p, setter)`, differing only by dropping the self-copy columns — partial-column UPDATE is the canonical supported path across providers (YDB is just the strictest, rejecting the PK in SET). The block-inlining moved out of `ExposeExpressionVisitor.VisitBlock` is a faithful, behavior-preserving port (only F# emits these assignment-blocks in LINQ trees). The `Issue4132Test2` YDB gate removal is justified — that is the sync `Update(p, setter)` shape the fix rewrites. The async `UpdateAsync(predicate, setter)` path builds its expression from the **synchronous** `UpdateSetter` / `UpdatePredicateSetter` method infos (async-ness is handled at the execution layer, not via a distinct expression node), so the same rewrite already covers it — confirmed by the async SQLite + YDB regression tests added in a follow-up commit (the SET clause carries only the changed column on both paths).

## Out-of-scope observations

- **No-op record copy still emits an all-column UPDATE.** When every constructor argument is a self-copy (`{ row with Field = row.Field }`), the change set is empty and the original full-record construction is left as-is, so the all-column UPDATE (PK included) is still emitted and still throws on YDB. The doc comment acknowledges this as acceptable since no column changes — true, but the "no rows actually change" rationale doesn't hold on YDB, where the statement errors rather than silently no-op'ing. Degenerate input, not the #5598 scenario; a future hardening could short-circuit a no-op update to emit nothing.
- **`Set<T,TV>` generic-arg type fidelity is untested.** The rewrite types the value lambda by the constructor argument's type while `Set`'s `TV` is fixed to the mapped member type. For ordinary F# record constructors these coincide, but a ctor argument whose static type diverges from the column type (implicit widening, option/nullable mismatch) would make `Expression.Call` throw at rewrite time rather than yielding a clean translation error. The current test type (all `int`/`string`) doesn't exercise divergence — untested rather than known-broken.
- **No-predicate `Update` overload path is untested.** `RewriteUpdate` handles both the predicate and no-predicate (`table.Update(fun r -> {…})`) cases, but every test exercises only the predicate form. Low risk given structural similarity; a full-table single-column update test would close the gap.
- **`Issue4132Test2` retains only smoke-level YDB coverage after gate removal.** Its body performs the update but never reads rows back, so re-enabling it on YDB only asserts the UPDATE no longer throws. The substantive correctness assertion lives in the new `Issue5598_UpdateSetsOnlyChangedColumnYdb` round-trip — adequate, but the gate removal's real safety net is the new test, not the re-enabled `Issue4132Test2`.

## Baselines
Delta: [linq2db.baselines PR #1945](https://github.com/linq2db/linq2db.baselines/pull/1945) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5627)

63 SQL baselines from the earlier commits — **57 modified, 6 added, 0 deleted** — all explained by the PR scope (independently confirmed against the files via a three-dot diff; every changed file is an `Issue4132Test2` or `Issue5598` baseline, no stray clusters). The modified set drops the no-op `Id = …` / `Number = …` self-assignments from `Issue4132Test2`'s SET clause across every provider (WHERE unchanged — the correct subtractive shape); the added set is the sync `Issue5598_UpdateSetsOnlyChangedColumn` (SQLite) + `…Ydb` (YDB) regression baselines plus the newly-ungated `Issue4132Test2(YDB)`. No cross-provider anomalies. The async regression tests in the latest follow-up will add further SQLite + YDB baselines on the next CI run, with SQL shape identical to the sync versions (locally verified: `SET [Text] = … WHERE [Id] = 1`).

igor-tkachev (dismissed):
Overall the direction looks right to me.

Moving the F# block-normalization out of `ExposeExpressionVisitor` and into the F# query interceptor is a good architectural cleanup: the core pipeline no longer carries an F# compiler-output workaround, and the new rewrite lowers the F# record-copy update into the already-supported explicit partial-update form.

The main `Update(predicate, setter)` fix looks correct: constructor arguments that are exact self-copies are dropped from the generated `Set(...)` chain, so `{ row with Text = ... }` updates only `Text` instead of re-emitting `Id`/`Number` into `SET`.

The new SQLite + YDB coverage, including async, is also the right test direction.

Two small follow-ups I would still consider:

1. The no-op fallback comment is a bit too strong. For `{ row with Field = row.Field }` the code intentionally falls back to the old full-record update path, which can still emit PK assignment and still fail on YDB. That is fine as out-of-scope for #5598, but the comment should not describe it as generally harmless.

2. `RewriteUpdate` handles both `UpdateSetter` and `UpdatePredicateSetter`, but tests currently cover only the predicate overload. A small SQLite test for `table.Update(fun row -> { row with Text = "after" })` would close that branch. For this test the SET-clause extraction should tolerate absence of `WHERE`.

Not a blocker from my side.

sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed F# record-copy updates. Both `table.Update(fun r -> { r with Field = v })` and `query.Update(predicate, fun r -> { r with Field = v })` — including the async `UpdateAsync` variants — previously generated an `UPDATE` that wrote **every** column (including the primary key), which some providers reject. Only the changed (non-PK) column(s) are now written.


_(none)_


---
<sub>Generated from commit `b124678`.</sub>

PR #5630, opened 2026-06-17
Support PostgreSQL/DuckDB DISTINCT ON via DistinctBy

### LINKED ISSUE #2075: using "distinct on" in postgres, is this possible?
is it possible to a query like here:

https://www.sisense.com/blog/first-row-per-group-5x-faster/

in oracle it seems to be possible via first_value

https://stackoverflow.com/questions/10515391/oracle-equivalent-of-postgres-distinct-on

maybe we have a extension method for example "DistinctBy<>" where we could hand over a column or a new class with multiple columns?

sdanyliv:
Closes #2075

## What

`DistinctBy` now emits native `SELECT DISTINCT ON (...)` on providers that
support it (PostgreSQL, DuckDB) instead of the `ROW_NUMBER()` emulation, which
stays as the fallback everywhere else — no SQL/baseline churn off PG/DuckDB.

The key selector becomes the `ON` expression list; the preceding `OrderBy` is
forced to lead with those keys (the syntax requires it), and the natural
`DISTINCT ON` ordering is kept.

## Changes

- AST: `SqlSelectClause.DistinctOn` + `IsDistinctOn`; threaded through the ctor,
  hash, `ToString`, `Cleanup`, and the query-element visitor (walk + clone).
- New `SqlProviderFlags.IsDistinctOnSupported` (default `false`), set on
  PostgreSQL (all versions) and DuckDB.
- `BasicSqlBuilder.BuildDistinctModifier` hook; PG/DuckDB builders render
  `DISTINCT ON (...)` via a shared helper that reuses the ORDER BY path.
- `DistinctByBuilder`: new branch before the ROW_NUMBER path, gated on the flag.
- Optimizer guards so the ON list and its ORDER BY survive flatten/merge/EXISTS:
  `SqlQueryOrderByOptimizer`, `SelectQueryOptimizerVisitor`,
  `BasicSqlOptimizer.SeparateDistinctFromPagination`, and
  `QueryHelper.CollectUniqueKeys` (ON tuple isn't a full-projection unique key).
- Remote round-trip in `LinqServiceSerializer`.

## Tests

`Tests/Linq/Linq/DistinctByMethodTests.cs`: existing behavioral tests now
exercise DISTINCT ON on PG/DuckDB; added SQL-shape assertions (`DISTINCT ON` on
PG/DuckDB, `ROW_NUMBER` + no `DISTINCT ON` on SQL Server) and edge cases (`Take`,
`Where`, `AsSubQuery`, nested `DistinctBy`, composite key). Green on PostgreSQL 16,
DuckDB, and SQL Server 2019.

## Follow-up

- ClickHouse has a `DISTINCT ON` clause with different ordering guarantees —
  left as a separate task (flag stays `false`).

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
@@ -1572,7 +1572,7 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 				parentQuery.Select.OptimizeDistinct = parentQuery.Select.OptimizeDistinct || subQuery.Select.OptimizeDistinct;
 				parentQuery.Select.IsDistinct       = true;
 				// carry DISTINCT ON keys with the modifier so a flattened parent keeps the correct distinct semantics
-				if (subQuery.Select.DistinctOn != null)
+				if (subQuery.Select.IsDistinctOn)
 					parentQuery.Select.DistinctOn = subQuery.Select.DistinctOn;
 			}
 
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
@@ -1076,6 +1076,10 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			if (!selectQuery.Select.IsDistinct || !selectQuery.Select.OptimizeDistinct)
 				return false;
 
+			// DISTINCT ON is never "redundant": removing it changes which row survives per ON key. Leave it intact.
+			if (selectQuery.Select.IsDistinctOn)
+				return false;
+
 			if (IsComplexQuery(selectQuery, false))
 				return false;
 
@@ -1386,7 +1390,7 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 						else
 							searchCondition.Add(new SqlPredicate.ExprExpr(rnColumn, SqlPredicate.Operator.LessOrEqual, takeValue, null));
 					}
-					else if (sql.Select.IsDistinct)
+					else if (sql.Select.IsDistinct && !sql.Select.IsDistinctOn)
 					{
 						sql.Select.IsDistinct = false;
 						searchCondition.Add(new SqlPredicate.ExprExpr(rnColumn, SqlPredicate.Operator.Equal, new SqlValue(1), null));
@@ -1567,6 +1571,9 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			{
 				parentQuery.Select.OptimizeDistinct = parentQuery.Select.OptimizeDistinct || subQuery.Select.OptimizeDistinct;
 				parentQuery.Select.IsDistinct       = true;
+				// carry DISTINCT ON keys with the modifier so a flattened parent keeps the correct distinct semantics
+				if (subQuery.Select.DistinctOn != null)
+					parentQuery.Select.DistinctOn = subQuery.Select.DistinctOn;
 			}
 
 			if (subQuery.Select.TakeValue != null)
@@ -2085,6 +2092,11 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 
 			if (subQuery.IsDistinct)
 			{
+				// A DISTINCT ON subquery must not be flattened into its parent: moving the modifier up would drop the
+				// ON list and break its ORDER BY dependency. Keep it as a nested derived table.
+				if (subQuery.Select.IsDistinctOn)
+					return false;
+
 				if (parentQuery.HasOrderBy && !parentQuery.OrderBy.Items.TrueForAll(oi => oi.Expression is SqlColumn col && subQuery.Select.Columns.Contains(col)))
 				{
 					return false;
@@ -3294,6 +3306,9 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			{
 				subQuery.Select.IsDistinct = true;
 				query.Select.IsDistinct    = false;
+				// move DISTINCT ON keys down with the modifier and clear them from the source query
+				subQuery.Select.DistinctOn = query.Select.DistinctOn;
+				query.Select.DistinctOn    = null;
 			}
 
 			_columnNestingCorrector.CorrectColumnNesting(query);
@@ -3395,7 +3410,7 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 				}
 			}
 
-			if (joinQuery.Select.Columns.Count > 1 && joinQuery.Select.IsDistinct)
+			if (joinQuery.Select.Columns.Count > 1 && joinQuery.Select.IsDistinct && !joinQuery.Select.IsDistinctOn)
 			{
 				if (!SqlProviderHelper.IsValidQuery(joinQuery, parentQuery: parentQuery, fakeJoin: null, columnSubqueryLevel: 1, _providerFlags, out _))
 					return false;
@@ -3568,8 +3583,9 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 
 			var sq = predicate.SubQuery;
 
-			// We can safely optimize out Distinct
-			if (sq.Select.IsDistinct)
+			// We can safely optimize out Distinct (but not DISTINCT ON — clearing only IsDistinct would leave the ON
+			// list dangling; EXISTS ignores distinctness/ordering anyway, so leaving it intact is harmless).
+			if (sq.Select.IsDistinct && !sq.Select.IsDistinctOn)
 			{
 				sq.Select.IsDistinct = false;
 			}
--- a/Tests/Linq/Linq/DistinctByMethodTests.cs
+++ b/Tests/Linq/Linq/DistinctByMethodTests.cs
@@ -220,6 +220,7 @@ namespace Tests.Linq
 				.Take(2);
 
 			AssertQuery(query);
+			Assert.That(query.ToSqlQuery().Sql, Does.Contain("DISTINCT ON"));
 		}
 
 		[Test]
@@ -236,6 +237,7 @@ namespace Tests.Linq
 				.Where(x => x.Amount > 100m);
 
 			AssertQuery(query);
+			Assert.That(query.ToSqlQuery().Sql, Does.Contain("DISTINCT ON"));
 		}
 
 		[Test]
@@ -254,6 +256,7 @@ namespace Tests.Linq
 				.Where(x => x.Amount > 100m);
 
 			AssertQuery(query);
+			Assert.That(query.ToSqlQuery().Sql, Does.Contain("DISTINCT ON"));
 		}
 
 		[Test]
@@ -270,6 +273,7 @@ namespace Tests.Linq
 				.DistinctBy(x => x.Name);
 
 			AssertQuery(query);
+			Assert.That(query.ToSqlQuery().Sql, Does.Contain("DISTINCT ON"));
 		}
 	}
 }
--- a/Tests/Linq/Linq/DistinctByMethodTests.cs
+++ b/Tests/Linq/Linq/DistinctByMethodTests.cs
@@ -152,6 +152,125 @@ namespace Tests.Linq
 
 			AssertQuery(query);
 		}
+
+		[Test]
+		public void DistinctByEmitsDistinctOn([IncludeDataSources(TestProvName.AllPostgreSQL, TestProvName.AllDuckDB)] string context)
+		{
+			using var db    = GetDataContext(context);
+			using var table = db.CreateLocalTable(TestData.Seed());
+
+			var query = table
+				.OrderBy(t => t.Group)
+				.ThenBy(t => t.Date)
+				.DistinctBy(x => x.Group);
+
+			AssertQuery(query);
+
+			var sql = query.ToSqlQuery().Sql;
+			Assert.That(sql, Does.Contain("DISTINCT ON"));
+			Assert.That(sql, Does.Not.Contain("ROW_NUMBER"));
+		}
+
+		[Test]
+		public void DistinctByCompositeKeyEmitsDistinctOn([IncludeDataSources(TestProvName.AllPostgreSQL, TestProvName.AllDuckDB)] string context)
+		{
+			using var db    = GetDataContext(context);
+			using var table = db.CreateLocalTable(TestData.Seed());
+
+			var query = table
+				.OrderBy(t => t.Name)
+				.ThenByDescending(t => t.Date)
+				.DistinctBy(x => new { x.Id, x.Name });
+
+			AssertQuery(query);
+
+			Assert.That(query.ToSqlQuery().Sql, Does.Contain("DISTINCT ON"));
+		}
+
+		[Test]
+		public void DistinctByUsesRowNumberWhenDistinctOnUnsupported([IncludeDataSources(TestProvName.AllSqlServer)] string context)
+		{
+			using var db    = GetDataContext(context);
+			using var table = db.CreateLocalTable(TestData.Seed());
+
+			var query = table
+				.OrderBy(t => t.Group)
+				.ThenBy(t => t.Date)
+				.DistinctBy(x => x.Group);
+
+			AssertQuery(query);
+
+			var sql = query.ToSqlQuery().Sql;
+			Assert.That(sql, Does.Not.Contain("DISTINCT ON"));
+			Assert.That(sql, Does.Contain("ROW_NUMBER"));
+		}
+
+		[Test]
+		public void DistinctByThenTake([IncludeDataSources(TestProvName.AllPostgreSQL, TestProvName.AllDuckDB)] string context)
+		{
+			using var db    = GetDataContext(context);
+			using var table = db.CreateLocalTable(TestData.Seed());
+
+			// DISTINCT ON followed by an outer ORDER BY + LIMIT: the inner ON ordering must survive.
+			var query = table
+				.OrderBy(t => t.Group)
+				.ThenBy(t => t.Date)
+				.DistinctBy(x => x.Group)
+				.OrderBy(x => x.Group)
+				.Take(2);
+
+			AssertQuery(query);
+		}
+
+		[Test]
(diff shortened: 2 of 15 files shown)


Tests/Linq/Linq/DistinctByMethodTests.cs
```diff
@@ -152,6 +152,125 @@ public void DistinctByWithComparerOrderShouldFail([DataSources] string context)
...
+				.DistinctBy(x => x.Group);
+
+			AssertQuery(query);
+
+			var sql = query.ToSqlQuery().Sql;
+			Assert.That(sql, Does.Not.Contain("DISTINCT ON"));
+			Assert.That(sql, Does.Contain("ROW_NUMBER"));
+		}
+
+		[Test]
```
MaceWindu, line 233: **Minor · MIN001** — Optimizer-preservation tests pin SQL shape in comments but only assert on rows.

`DistinctByThenTake` (and its siblings `DistinctByThenWhere`, `DistinctByInSubQuery`, `NestedDistinctBy`) carry comments asserting a specific SQL shape — "the inner DISTINCT ON must survive", "must wrap the DISTINCT ON query as a subquery, not push into it" — but the test body only calls `AssertQuery`, which validates the result set, not the structure. A future optimizer change that flattens/strips the ON-list while still returning correct rows would pass these tests undetected.

Fix: add a `ToSqlQuery().Sql` assertion that pins the structural claim (e.g. `Does.Contain("DISTINCT ON")`, and the expected nesting for the subquery/nested cases), the way `DistinctByEmitsDistinctOn` already does.

— ✓ Fixed in 3b6dcec

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
```diff
@@ -1567,6 +1571,9 @@ bool MoveSubQueryUp(SelectQuery parentQuery, SqlTableSource tableSource)
 			{
 				parentQuery.Select.OptimizeDistinct = parentQuery.Select.OptimizeDistinct || subQuery.Select.OptimizeDistinct;
 				parentQuery.Select.IsDistinct       = true;
+				// carry DISTINCT ON keys with the modifier so a flattened parent keeps the correct distinct semantics
+				if (subQuery.Select.DistinctOn != null)
+					parentQuery.Select.DistinctOn = subQuery.Select.DistinctOn;
```
MaceWindu, line 1576: **Suggestion · SUG001** — This `DistinctOn` carry-up in `MoveSubQueryUp` is unreachable for a real DISTINCT ON subquery.

`MoveSubQueryUp` only proceeds past `IsMovingUpValid`, which now returns `false` whenever `subQuery.Select.IsDistinctOn` is true. The only way to reach this assignment with a non-null `subQuery.Select.DistinctOn` is a non-null but *empty* list — a state `ApplyDistinctOn` never produces (it sets `DistinctOn` only when there is at least one ON expression). So the block is effectively dead. It is harmless, but its `DistinctOn != null` guard is also inconsistent with the `IsDistinctOn` (Count > 0) predicate every other guard the PR adds uses, which can read as a latent disagreement to a future maintainer.

Fix: drop the block (it cannot fire for a real DISTINCT ON subquery), or — if kept as belt-and-suspenders — use the `IsDistinctOn` predicate to stay consistent with the rest of the PR's guards.

— ✓ Fixed in 3b6dcec

Tests/Linq/Linq/DistinctByMethodTests.cs
MaceWindu, line 1: **Major · MAJ001** — The `DistinctOn` remote/LinqService serialization round-trip is untested.

The PR adds `DistinctOn` to `LinqServiceSerializer` (write + read/`ReadArray`), but no test exercises that round-trip. `DistinctOn` is only populated when `IsDistinctOnSupported` is true (PostgreSQL/DuckDB). Every DISTINCT ON test here uses `[IncludeDataSources(...)]`, whose default constructor passes `includeLinqService:false`, so none run over the remote context; the `[DataSources]` tests that *do* enable the remote context run on providers where the flag is false. Net result: the new serializer code has zero coverage — a dropped ON expression on the wire would pass CI silently. (This same deserialization line is also one of the BLK001 build-break sites.)

Fix: add a test that forces a `DistinctBy` query through the LinqService round-trip on PostgreSQL/DuckDB — e.g. `[IncludeDataSources(true, TestProvName.AllPostgreSQL, TestProvName.AllDuckDB)]` wrapping a `DistinctBy` query and `AssertQuery` (the remote context exercises the serializer end to end).

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] Solid, well-contained feature. The new `DistinctOn` AST member is threaded through every fan-out path (hash, `ToString`, `Cleanup`, the visitor's read/in-place/clone branches, the remote serializer), the optimizer guards are comprehensive, and the PostgreSQL/DuckDB SQL shapes are correct. The one issue that must be fixed before merge is the Release build break (BLK001).
- [ ] Provider scoping is complete: PostgreSQL (all versions) + DuckDB enabled; ClickHouse intentionally deferred (different `DISTINCT ON` ordering semantics); no other provider supports `DISTINCT ON` (H2 does not; there are no separate CockroachDB/Redshift providers). See the out-of-scope note.

### Blockers (BLK)
- [x] **BLK001** — Release build fails: `IDE0306` (collection-expression) analyzer error at three new sites
  Why: A Release build (`dotnet build Source/LinqToDB/LinqToDB.csproj -c Release`) fails with 6 `IDE0306` errors. Analyzers run as errors only in Release, so a Debug/Testing build is green — which is why this passed locally. All three are PR-added `new List<ISqlExpression>(...)` calls in **target-typed positions**, where the analyzer requires a collection expression. (The same construct at `QueryElementVisitor.cs:2121` uses `var`, has no collection-expression target, and correctly does not trip the rule.) CI's `build` leg will fail as-is.
  Fix:

  `Source/LinqToDB/Internal/SqlQuery/Visitors/QueryElementVisitor.cs:2090`
  ```diff
  -    (newDistinctOn ??= new List<ISqlExpression>(element.DistinctOn))[i] = newOn;
  +    (newDistinctOn ??= [..element.DistinctOn])[i] = newOn;
  ```
  `Source/LinqToDB/Internal/SqlQuery/Visitors/QueryElementVisitor.cs:2196`
  ```diff
  -    query.Select.DistinctOn != null ? new List<ISqlExpression>(query.Select.DistinctOn) : null,
  +    query.Select.DistinctOn != null ? [..query.Select.DistinctOn] : null,
  ```
  `Source/LinqToDB/Internal/Remote/LinqServiceSerializer.cs:2482`
  ```diff
  -    DistinctOn       = distinctOn == null ? null : new List<ISqlExpression>(distinctOn),
  +    DistinctOn       = distinctOn == null ? null : [..distinctOn],
  ```

## Out-of-scope observations

- **ClickHouse / other providers correctly not enabled** — ClickHouse has a `DISTINCT ON` clause but with non-standard semantics (no ORDER-BY-leading-prefix requirement, different row selection), so leaving `IsDistinctOnSupported = false` there (ROW_NUMBER fallback) is correct; the follow-up note in the PR body captures it. No other supported provider qualifies (H2 has no `DISTINCT ON`; no Cockroach/Redshift providers exist). Provider coverage is complete.
- **`MoveSubQueryUp` into a `DISTINCT ON` *parent* is not specifically guarded** — the PR rejects flattening a `DISTINCT ON` *subquery* upward (`IsMovingUpValid`), but the symmetric case (a non-distinct child moved up into a `DISTINCT ON` parent) relies on `MoveSubQueryUp`'s `NotifyReplaced` pass propagating into the parent's `DistinctOn` list via the visitor. This appears handled but could not be proven across all flatten shapes by static reading. Worth a targeted test: a `DISTINCT ON` outer query whose sole FROM is a flattenable derived table, asserting the ON expressions still bind correctly after optimization.
- **`DISTINCT ON` forces the ON keys to the front of the final ORDER BY** — `ApplyDistinctOn` builds ORDER BY as `[ON keys] + [remaining user order-by]`, synthesizing ASC for any ON key the user did not order by. This is inherent to PostgreSQL/DuckDB `DISTINCT ON` and the result-set tests pass against the in-memory baseline, so it is correct behavior — noted only because a user relying on a specific post-`DistinctBy` ordering gets the ON-led order unless they wrap in an outer OrderBy.
- **Positional ORDER BY vs `DISTINCT ON` rendering** — the `DISTINCT ON` list always renders bare expressions, while the ORDER BY clause substitutes an ordinal (`ORDER BY 1`) for positioned items. If a positional ORDER BY item matching an ON key ever reaches a `DISTINCT ON` query, the two renderings diverge syntactically; PostgreSQL/DuckDB resolve the ordinal so it should still satisfy the leftmost-match rule. Fragility note, not a reproduced defect.
- **`PublicAPI.Unshipped.txt` not yet updated** — the new `LinqToDB.Internal.*` members (`SqlProviderFlags.IsDistinctOnSupported`, `SqlSelectClause.DistinctOn`/`IsDistinctOn`, `BasicSqlBuilder.BuildDistinctModifier`) are not in `PublicAPI.Unshipped.txt` at HEAD. Per repo convention these lag source within a release cycle; flagged only so the pre-release PublicAPI refresh accounts for the new surface.
- **`DistinctByInSubQuery` and `DistinctByThenWhere` emit identical SQL** — both produce a WHERE-filtered outer query over a `DISTINCT ON` subquery (same baseline). Expected, but worth confirming the two test bodies exercise genuinely different LINQ operator sequences rather than duplicate coverage.

## Baselines
Delta: [linq2db.baselines PR #1947](https://github.com/linq2db/linq2db.baselines/pull/1947) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5630)

106 SQL baselines changed (64 added, 42 modified, 0 deleted), no metrics changes, **zero cross-provider anomalies**. The PR's "no churn off PG/DuckDB" claim holds.

### DuckDB + PostgreSQL 13–18: ROW_NUMBER() emulation → DISTINCT ON (42 modified)
Existing `DistinctBy` / `DistinctByOrderByNulls` / `DistinctByDefaultNullsPosition` baselines now emit `SELECT DISTINCT ON (keys) … ORDER BY keys, sort` instead of the `ROW_NUMBER() … WHERE RowNumber = 1` subquery. The ON key matches the prior PARTITION BY; ORDER BY preserves direction including NULLS FIRST/LAST. Correct.

### DuckDB + PostgreSQL 13–18: new DISTINCT ON test baselines (64 added)
8 new test methods cover single key, composite key, subquery, nested, Take, Where — all correct `DISTINCT ON` shapes.

### SqlServer 2005–2025 + Contained + SA: new ROW_NUMBER fallback baselines (22 added)
New `DistinctByUsesRowNumberWhenDistinctOnUnsupported` documents the fallback path. No existing SqlServer baseline was modified, so the "no churn off PG/DuckDB" claim is literally met.

Tests/Linq/Linq/DistinctByMethodTests.cs
MaceWindu: Correction: MAJ001 is withdrawn — the remote/LinqService `DistinctOn` round-trip is covered, not untested.

The `DistinctBy` test above uses `[DataSources]`, whose no-arg ctor sets `includeLinqService: true` (`DataSourcesAttribute.cs:11`), and `DataSourcesBaseAttribute.cs:45-46` appends a `.LinqService` variant for every provider. On a `PostgreSQL.*.LinqService` / `DuckDB.LinqService` context `IsDistinctOnSupported` is true, so the query populates `SqlSelectClause.DistinctOn` and serializes it through `LinqServiceSerializer` end to end; a dropped ON expression would change the result set (10 seed rows vs deduped) and fail `AssertQuery`. The premise that every DISTINCT ON test runs with `includeLinqService:false` does not hold for this `[DataSources]` test.

Residual (minor, not blocking): the coverage is result-only — no assertion pins the round-tripped `DISTINCT ON` SQL shape on the remote path specifically. An explicit `.LinqService` PG/DuckDB test asserting the SQL would pin the serializer contract directly, but functional coverage already exists.

Tests/Linq/Linq/DistinctByMethodTests.cs
sdanyliv: Addressed in 73fddfd2cf: `DistinctByThenTake`, `DistinctByThenWhere`, `DistinctByInSubQuery`, and `NestedDistinctBy` now assert `query.ToSqlQuery().Sql` contains `DISTINCT ON` alongside `AssertQuery`, so a future flatten/strip that still returns correct rows would fail the SQL-shape check.

Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs
sdanyliv: Addressed in 73fddfd2cf: the `MoveSubQueryUp` carry guard now uses `IsDistinctOn` instead of `DistinctOn != null`, consistent with the PR's other guards. Kept as belt-and-suspenders rather than dropped.

sdanyliv:
Pushed 73fddfd2cf addressing the review:

- **BLK001** - the three target-typed `new List<ISqlExpression>(...)` are now collection expressions (`[..x]`); `dotnet build -c Release` is clean. The `var`-ternary clone-path site is left as-is (no collection-expression target, matching the existing `var` site that correctly doesn't trip IDE0306).
- **MIN001** - `DISTINCT ON` SQL-shape assertions added to the four optimizer-preservation tests.
- **SUG001** - `MoveSubQueryUp` carry guard now uses `IsDistinctOn`.
- **MAJ001** - already self-withdrawn; the `[DataSources]` LinqService variants exercise the serializer round-trip.

Re-validated on DuckDB (22/22) and SQL Server 2019; Release build green.

igor-tkachev (dismissed):
Looks good to me.

The implementation is well-scoped: `IsDistinctOnSupported` defaults to false, is enabled only for PostgreSQL/DuckDB, and unsupported providers keep the existing `ROW_NUMBER()` fallback path. The added SQL Server coverage also explicitly asserts no `DISTINCT ON` and verifies the fallback shape.

Compatibility note: `DistinctBy` is already released and currently has an existing SQL translation. For PostgreSQL/DuckDB this PR changes existing generated SQL from `ROW_NUMBER()` emulation to native `DISTINCT ON`. The new SQL is valid and is probably the better native translation for those providers, but it is still an observable SQL-generation change: baselines, query plans, and possibly final result ordering can change unless users apply an outer `OrderBy`.

I don't think this should block the PR. We should either document it explicitly as an intentional SQL-generation change in release notes, or consider whether it belongs behind a compatibility/SQL-shape option for one release cycle. My preference would be at least a release note entry; a flag is a product/release-policy decision, not a required change for this PR.

igor-tkachev (commented):
Follow-up on the compatibility note: if we decide to make the old SQL form configurable for one release cycle, I would prefer a specific option name rather than a vague provider-optimization switch.

Possible names:

- `UseNativeDistinctOnForDistinctBy`
- `UseDistinctOnForDistinctBy`

I would not treat this as required for this PR. The implementation is correct and provider-scoped. This is mainly a release-policy/API-contract decision: either document the PostgreSQL/DuckDB SQL-shape change explicitly, or expose a focused opt-out/compatibility option for users who need the previous `ROW_NUMBER()` translation temporarily.

MaceWindu (commented):
## Verification update — 2026-06-22, against HEAD `4baffaac`

> [!IMPORTANT]
> **Agentic review — treat with care.** Produced by an LLM agent, not a human reviewer; individual findings can be wrong or miss context. Findings are starting points for discussion, not verdicts.

All three findings from the prior review are fixed. One new defect was found while verifying the constant-key coverage and has been fixed in `4baffaac` (pushed to this branch).

### Fixed
- [x] **BLK001** — Release build `IDE0306`: the three target-typed `new List<ISqlExpression>(...)` sites are now collection expressions (`[..x]`). Confirmed by a clean `dotnet build -c Release -f net10.0` (0 warnings, 0 errors) at HEAD.
- [x] **MIN001** — All four optimizer-preservation tests now assert `ToSqlQuery().Sql`: `DistinctByThenTake/ThenWhere/InSubQuery` use `Does.Contain("DISTINCT ON")`; `NestedDistinctBy` asserts the DISTINCT-ON count `== 2`.
- [x] **SUG001** — `SelectQueryOptimizerVisitor` `MoveSubQueryUp` carry guard now uses `IsDistinctOn`, consistent with the PR's other guards.

(MAJ001 was self-withdrawn earlier; not re-verified.)

### New finding — found and fixed in `4baffaac`
- [x] **MAJ002** — `DistinctByBuilder` emitted a cartesian product for a constant / non-SQL-ready key on DISTINCT-ON providers.

  On PostgreSQL/DuckDB the new DISTINCT ON branch always builds a sequence over `orderedExpression` to derive the ON list. For an empty ON list (`DistinctBy(x => 1)`) — or any key where `IsSqlReady` is false — it fell through to the ROW_NUMBER fallback, which built a **second** `TryBuildSequence` over the same source. The discarded first build leaked an extra table reference, so the ROW_NUMBER subquery emitted `FROM TestData e, TestData e_1` — a self cross-join (O(N²)). Results stayed correct, so `AssertQuery` passed, but the PR's own new `DistinctByConstantKeyFallsBackToRowNumber` baseline enshrined the cartesian product as expected.

  **Fix (`4baffaac`):** reuse the already-built sequence for the ROW_NUMBER fallback instead of building a second one. Added a regression assertion to `DistinctByConstantKeyFallsBackToRowNumber` pinning a single source-table reference (`sql.Split("TestData").Length - 1 == 1`). Verified red→green on DuckDB (the assertion fails on the prior builder, passes after the fix); full `DistinctByMethodTests` fixture green (25/25) on DuckDB; Release build clean.

  Note: the `DistinctByConstantKeyFallsBackToRowNumber` DuckDB baseline will regenerate on CI (the cross-join `FROM TestData e, TestData e_1` becomes a single-table ROW_NUMBER query).

### Still actual
None.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- `DistinctBy` now generates native `SELECT DISTINCT ON (...)` on PostgreSQL and DuckDB instead of the `ROW_NUMBER()` emulation, producing simpler SQL. The key selector becomes the `ON` list and the query's `OrderBy` is arranged to lead with those keys (as the syntax requires); other providers keep the existing emulation unchanged. ([#5630](https://github.com/linq2db/linq2db/pull/5630))


- `DistinctBy` emits native `DISTINCT ON` on PostgreSQL and DuckDB ([#5630](https://github.com/linq2db/linq2db/pull/5630)).


---
<sub>Generated from commit `570be39`.</sub>

PR #5632, opened 2026-06-18
Fix #5626: ORDER BY dropped for OrderBy().Distinct()/GroupBy().Take()

### LINKED ISSUE #5626: `ORDER BY` clause omitted for certain `.OrderBy(…).Distinct().Take(…)….InsertAsync(…)` expressions
### Description

In some circumstances, all or some of the `ORDER BY` clauses are omitted when the order is only observed via a `.Take(…)`, and the `.Take(…)` doesn't immediately follow the `.OrderBy(…)`/`.ThenBy(…)` in the method chain (e.g. with a `.Distinct()` between them).

Repro project: https://github.com/megaslice/p1-upstream-bug-repro-linq2db-6.3.0

CC @coudy

### Reproduction Steps

https://github.com/megaslice/p1-upstream-bug-repro-linq2db-6.3.0/blob/master/SpecificReproductions/OrderByDistinctTakeInsertTest.cs

### Exception details

Test results: https://github.com/megaslice/p1-upstream-bug-repro-linq2db-6.3.0/blob/master/linq2db-6.trx

### Regression?

5.4.1.9: https://github.com/megaslice/p1-upstream-bug-repro-linq2db-6.3.0/blob/master/linq2db-5.trx

### Linq To DB version

6.3.0

### .NET version

10

### Database

SQL Server 2019

### Database provider

Microsoft.Data.SqlClient 6.1.3

### Operating system

Windows

### Configuration

https://github.com/megaslice/p1-upstream-bug-repro-linq2db-6.3.0/blob/master/setup.sql

sdanyliv:
Fixes #5626.

`OrderBy(expr).Distinct().Take(n)` dropped the ORDER BY. `SqlQueryOrderByOptimizer`
removed an order feeding a DISTINCT/GROUP BY when its expression wasn't a literal
output column — but a `Take`/`Skip` above the DISTINCT/grouping observes which rows
are kept, so the limit returned arbitrary rows.

## Changes — `SqlQueryOrderByOptimizer`
- **DISTINCT**: keep the order above the DISTINCT when every column it references is
  produced by the DISTINCT projection (so it's functionally determined by the output),
  instead of requiring an exact column match.
- **GROUP BY**: same check against the grouping keys (plain `GROUP BY` only — ROLLUP/
  CUBE/GROUPING SETS excluded); the legacy exact-column form is retained.
- Move the validated order up **as-is** rather than materializing it as a synthetic
  column; the column-nesting corrector re-levels references onto existing columns.

`SelectQueryOptimizerVisitor` was audited — its merge guards already keep these orders
correctly nested; no change needed.

## Tests
`OrderByDistinctTests`: `OrderByDistinctTakeOrdering` (DISTINCT) and
`OrderByGroupByTakeOrdering` (GROUP BY). Asserted against explicit expected rows —
`AssertQuery` can't catch this class of bug (its in-memory baseline drops the same
ORDER BY).

## Validation
- SQLite (direct + remote): 1262/0 across Distinct/OrderBy/GroupBy/Union/TakeSkip/
  EagerLoading/SelectMany.
- SQL Server 2019 (Microsoft.Data.SqlClient, the reported provider): 788/0 — confirms
  valid SQL on a provider that rejects `SELECT DISTINCT … ORDER BY <compound>`.

Generated SQL for affected `…Distinct()/GroupBy().Take()` queries now includes the
ORDER BY, so some SQL baselines will change.

## Follow-up commit

Review follow-up (bb69ae09, 4be07a48):
- Extracted the duplicated order-expression leaf-collection block in `SqlQueryOrderByOptimizer` (`AllOrderColumnsProducedByDistinct` / `AllOrderColumnsAreGroupingKeys`) into a shared `CollectReferencedColumns` helper.
- `OrderByDistinctTests`: switched `OrderByDistinctTakeOrdering` / `OrderByGroupByTakeOrdering` to order-sensitive `Assert.That(..., Is.EqualTo(...))`, and added `OrderByGroupByTakeOrderingKeyNotProjected` whose ORDER BY expression is absent from the SELECT projection, exercising the grouping-key path unambiguously.

### DIFF
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
@@ -242,12 +242,10 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 		}
 
 		/// <summary>
-		/// Returns <see langword="true"/> when every column/field referenced by <paramref name="orderExpression"/>
-		/// is produced as an output column of the DISTINCT <paramref name="distinctQuery"/>. In that case the
-		/// expression is functionally determined by the DISTINCT projection, so it can be ordered above the
-		/// DISTINCT and exposed as an additional output column without changing which rows are kept.
+		/// Collects the <see cref="QueryElementType.Column"/> / <see cref="QueryElementType.SqlField"/> leaf
+		/// references of <paramref name="orderExpression"/> without descending into their definitions.
 		/// </summary>
-		static bool AllOrderColumnsProducedByDistinct(SelectQuery distinctQuery, ISqlExpression orderExpression)
+		static List<ISqlExpression> CollectReferencedColumns(ISqlExpression orderExpression)
 		{
 			var referenced = new List<ISqlExpression>();
 
@@ -261,6 +259,19 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 
 				return true;
 			});
+
+			return referenced;
+		}
+
+		/// <summary>
+		/// Returns <see langword="true"/> when every column/field referenced by <paramref name="orderExpression"/>
+		/// is produced as an output column of the DISTINCT <paramref name="distinctQuery"/>. In that case the
+		/// expression is functionally determined by the DISTINCT projection, so it can be ordered above the
+		/// DISTINCT and exposed as an additional output column without changing which rows are kept.
+		/// </summary>
+		static bool AllOrderColumnsProducedByDistinct(SelectQuery distinctQuery, ISqlExpression orderExpression)
+		{
+			var referenced = CollectReferencedColumns(orderExpression);
 
 			if (referenced.Count == 0)
 				return false;
@@ -290,18 +301,7 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 			if (groupedQuery.GroupBy.GroupingType != GroupingType.Default)
 				return false;
 
-			var referenced = new List<ISqlExpression>();
-
-			orderExpression.VisitParentFirst(referenced, static (list, e) =>
-			{
-				if (e.ElementType is QueryElementType.Column or QueryElementType.SqlField)
-				{
-					list.Add((ISqlExpression)e);
-					return false; // leaf reference - don't descend into the column/field definition
-				}
-
-				return true;
-			});
+			var referenced = CollectReferencedColumns(orderExpression);
 
 			if (referenced.Count == 0)
 				return false;
--- a/Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
@@ -122,36 +122,47 @@ namespace LinqToDB.Internal.SqlQuery.Visitors
 
 							if (canPopulateUpperLevel && parentSelectQuery.Select.IsDistinct)
 							{
-								canPopulateUpperLevel = parentSelectQuery.Select.Columns.Exists(c =>
-								{
-									if (c.Expression is SqlColumn column)
-										return QueryHelper.SameWithoutNullablity(column.Expression, orderByItem.Expression);
-									return false;
-								});
+								// The order can move above a DISTINCT parent only when its expression is built entirely
+								// from columns/fields the DISTINCT projects (#5626). Then it is functionally determined by
+								// the DISTINCT output, so ordering by it above the DISTINCT is well-defined and exposing it
+								// as an extra output column cannot change which rows are distinct. Ordering by anything
+								// outside the projection stays unsupported and the order is dropped.
+								canPopulateUpperLevel = AllOrderColumnsProducedByDistinct(parentSelectQuery, orderByItem.Expression);
 							}
 
 							if (canPopulateUpperLevel && !parentSelectQuery.GroupBy.IsEmpty)
 							{
-								canPopulateUpperLevel = selectQuery.Select.Columns.Exists(c => QueryHelper.SameWithoutNullablity(c.Expression, orderByItem.Expression));
+								// The order can move above a GROUP BY parent only when its expression is built entirely
+								// from the grouping keys (#5626): `GROUP BY a, b ... ORDER BY a * 100 + b` is valid, but
+								// ordering by a non-grouped column is not. The exact-column form is kept for back-compat.
+								canPopulateUpperLevel =
+									selectQuery.Select.Columns.Exists(c => QueryHelper.SameWithoutNullablity(c.Expression, orderByItem.Expression))
+									|| AllOrderColumnsAreGroupingKeys(parentSelectQuery, orderByItem.Expression);
 							}
 
 							if (canPopulateUpperLevel)
 							{
-								if (orderByItem.Expression is SqlExpression or SqlFragment)
+								// Raw-template AST nodes (Sql.Expr / Sql.Fragment) may carry trailing direction
+								// modifiers in their template text (e.g. "{0} NULLS FIRST"). Wrapping them as a synthetic
+								// subquery column captured the modifier inside the column expression and emitted invalid
+								// SQL like `... END NULLS FIRST as c1`. Push such expressions directly to the parent ORDER
+								// BY so the modifier stays in the outer clause.
+								// When a DISTINCT or GROUP BY is involved we also push the expression as-is instead of
+								// materializing it as a column (#5626): the order has been validated to use only columns the
+								// DISTINCT projects / grouping keys, so adding a synthetic column would needlessly widen the
+								// DISTINCT key or the projection. The column-nesting corrector re-levels the references onto
+								// the existing output columns.
+								if (orderByItem.Expression is SqlExpression or SqlFragment
+									|| parentSelectQuery.Select.IsDistinct
+									|| selectQuery.Select.IsDistinct
+									|| !parentSelectQuery.GroupBy.IsEmpty)
 								{
-									// Raw-template AST nodes (Sql.Expr / Sql.Fragment) may carry trailing
-									// direction modifiers in their template text (e.g. "{0} NULLS FIRST").
-									// Wrapping them as a synthetic subquery column captured the modifier
-									// inside the column expression and emitted invalid SQL like
-									// `... END NULLS FIRST as c1`. Push such expressions directly to the
-									// parent ORDER BY so the modifier stays in the outer clause where it
-									// belongs. Every other AST node (structured expressions, columns, fields,
-									// functions, case, etc.) keeps the existing AddColumn dedup path — their
-									// SQL output is always a valid scalar.
 									parentSelectQuery.OrderBy.Items.Add(new SqlOrderByItem(orderByItem.Expression, orderByItem.IsDescending, orderByItem.IsPositioned, orderByItem.NullsPosition));
 								}
 								else
 								{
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
```diff
@@ -230,6 +241,85 @@ static bool CanRemoveOrderBy(SelectQuery selectQuery)
...
+			orderExpression.VisitParentFirst(referenced, static (list, e) =>
+			{
+				if (e.ElementType is QueryElementType.Column or QueryElementType.SqlField)
+				{
+					list.Add((ISqlExpression)e);
+					return false; // leaf reference - don't descend into the column/field definition
+				}
+
+				return true;
+			});
```
MaceWindu: **Minor · MIN001** — The leaf-reference collection block here — the `VisitParentFirst` walk that gathers `Column`/`SqlField` leaf nodes plus the `referenced.Count == 0` early-out — is duplicated verbatim from `AllOrderColumnsProducedByDistinct` above (the copy starting at line 252). The two methods differ only in the final per-reference predicate (produced-by-DISTINCT-output vs. is-a-grouping-key); everything up to that predicate is identical copy-paste introduced by this PR.

Fix: extract the collection into a single private static helper, e.g. `static List<ISqlExpression> CollectReferencedColumns(ISqlExpression orderExpression)`, and have both methods call it before applying their respective predicate. (Structural extraction spanning both methods — no inline suggestion.)

Tests/Linq/Linq/OrderByDistinctTests.cs
```diff
@@ -483,5 +483,52 @@ public void OrderByDistinctAll2([DataSources] string context)
...
+			// from linq2db's own exposed expression, which drops the same ORDER BY and would mask the bug.
+			var result = table
+				.Concat(table)
+				.OrderByDescending(x => x.OrderData1 * 100 + x.Id)
+				.Distinct()
+				.Take(3)
+				.Select(x => x.Id)
+				.ToArray();
+
+			AreEqual(new[] { 600, 500, 400 }, result);
```
MaceWindu: **Suggestion · SUG001** — `TestBase.AreEqual` compares set-wise (via `Except`/count), so it is order-insensitive. Both new tests are specifically about the ORDER BY surviving so `Take(3)` keeps the correct rows in the correct sequence; the set comparison does catch the headline bug, but it would miss a regression that returns the right 3 rows in the wrong order. Since these queries have a well-defined final order, an order-sensitive assertion is valid and strictly stronger, and matches the comment's stated intent.

Fix: use an order-sensitive assertion. Apply the same change at line 531 for `OrderByGroupByTakeOrdering` (`Assert.That(result.Select(x => x.Key).ToArray(), Is.EqualTo(new[] { 505, 404, 303 }));`).

```suggestion
			Assert.That(result, Is.EqualTo(new[] { 600, 500, 400 }));
```

MaceWindu (commented):
> [!IMPORTANT]

The fix is well-scoped and the core logic is sound: `AllOrderColumnsProducedByDistinct` / `AllOrderColumnsAreGroupingKeys` correctly require every leaf column/field in the order expression to be functionally determined by the DISTINCT projection / grouping keys, the GROUP BY check gates on `GroupingType.Default` to exclude ROLLUP/CUBE/GROUPING SETS, and the as-is-move vs `AddColumn` split is reasoned. No correctness, SQL, public-API, or test-coverage issues found. Two non-blocking notes are inline.

## Out-of-scope observations

- **GROUP BY test may exercise the back-compat disjunct rather than the new grouping-key path** — In `OrderByGroupByTakeOrdering` the order expression `OrderData1 * 100 + OrderData2` is also projected as the inner `Key` column, so the pre-existing exact-column disjunct (`Select.Columns.Exists` of an order-matching column) can satisfy the keep-order check before `AllOrderColumnsAreGroupingKeys` is consulted. Both disjuncts fix the reported bug, so this is not a defect — but a variant whose ORDER BY expression is **not** also present verbatim in the SELECT projection would exercise the new grouping-key path unambiguously.

## Baselines
Delta: [linq2db.baselines #1952](https://github.com/linq2db/linq2db.baselines/pull/1952) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5632)

177 SQL baselines changed across 59 providers: **118 added, 59 modified, 0 deleted** — all consistent with the fix (ORDER BY now survives DISTINCT / GROUP BY + Take). Three-dot diff counts and the modified/added file contents were verified against the branch.

### `DistinctTests.OrderByExpressionDistinct` — second ORDER BY key restored (59 providers, modified)
Pre-existing test, all 59 modified files. The ORDER BY previously emitted only the first key (`Coalesce(MiddleName) DESC`); it now also emits the secondary key (`Coalesce(LastName)`) that the old optimizer silently dropped — the direct consequence of the fix.

<details><summary>sample diff (SqlServer.2019.MS)</summary>

```diff
 ORDER BY
-	Coalesce([x_1].[MiddleName], N'-2') DESC
+	Coalesce([x_1].[MiddleName], N'-2') DESC,
+	Coalesce([x_1].[LastName], N'-1')
```
</details>

### `OrderByDistinctTests.OrderByDistinctTakeOrdering` — new test (59 providers, added)
`OrderBy(expr).Distinct().Take(n)`: outer `SELECT TOP/LIMIT/FETCH … ORDER BY OrderData1 * 100 + Id DESC` over a DISTINCT subquery. Correct on every provider family (paging/quoting variation only); Oracle 11 uses the expected ROWNUM wrap.

### `OrderByDistinctTests.OrderByGroupByTakeOrdering` — new test (59 providers, added)
`OrderBy(expr).GroupBy(keys).Take(n)`: flat `GROUP BY … ORDER BY OrderData1 * 100 + OrderData2 DESC` + provider paging. Correct on every provider family; Oracle 11 re-applies the ORDER BY on the outer ROWNUM wrapper, YDB orders by the column alias (semantically equivalent).

No cross-provider anomalies.

Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryOrderByOptimizer.cs
MaceWindu: Fixed in bb69ae09 — extracted the shared leaf-collection block into a `CollectReferencedColumns` helper; both `AllOrderColumnsProducedByDistinct` and `AllOrderColumnsAreGroupingKeys` now call it before applying their own predicate.

Tests/Linq/Linq/OrderByDistinctTests.cs
MaceWindu: Fixed in 4be07a48 — both tests now use an order-sensitive `Assert.That(..., Is.EqualTo(...))`. Also added `OrderByGroupByTakeOrderingKeyNotProjected`, whose ORDER BY expression is absent from the SELECT projection, so it exercises the new grouping-key path (`AllOrderColumnsAreGroupingKeys`) rather than the back-compat exact-column disjunct.

MaceWindu (approved):

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review, so this complements — rather than replaces — human judgement. Individual findings can be wrong, overconfident, or miss context a human would catch. Feel free to disagree, dismiss, or ask for clarification on any finding — the agent's judgement is not authoritative.

The fix holds up at current HEAD. The two findings from the earlier review are correctly resolved:

- **MIN001** — the duplicated leaf-collection block is now a single `CollectReferencedColumns` static helper, consumed by both `AllOrderColumnsProducedByDistinct` and `AllOrderColumnsAreGroupingKeys` before each applies its own predicate.
- **SUG001** — all three new tests use order-sensitive `Assert.That(result, Is.EqualTo(new[]{…}))`, and the added `OrderByGroupByTakeOrderingKeyNotProjected` exercises the grouping-key path (its ORDER BY expression is absent from the outer projection), closing the earlier note about the back-compat exact-column disjunct masking the new path.

No new correctness, SQL, public-API, or test-coverage issues. Spot-confirmed: `GroupingType.Default` correctly gates out ROLLUP/CUBE/GROUPING SETS; the `VisitParentFirst` walk collects leaf references without descending into definitions; empty-reference lists return `false` (conservative — constant/aggregate-only orders stay dropped); the as-is-move (computed expressions) vs `AddColumn`-dedup (plain columns/fields) split is intact. No public API surface change — the three new methods are private `static`.

## Out-of-scope observations

Surfaced during review but outside this PR's scope. FYI, not findings.

- **`SELECT DISTINCT … ORDER BY <non-projected computed expr>` validity rests on the column-nesting corrector.** Computed order expressions are now pushed to the parent ORDER BY as-is rather than materialized as a synthetic column. Strict engines reject `SELECT DISTINCT … ORDER BY <expr>` when `<expr>` isn't in the select list; validity here depends on `AllOrderColumnsProducedByDistinct` guaranteeing the referenced columns are projected, plus the corrector re-leveling the ORDER BY onto those output columns. This is the design intent and the new `[DataSources]` tests verify it across 61 providers in CI — noted only so the cross-provider ORDER-BY-validity dimension stays visible.

## Baselines
Delta: [linq2db.baselines #1952](https://github.com/linq2db/linq2db.baselines/pull/1952) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5632)

244 SQL baselines changed across 61 providers: **183 added, 61 modified, 0 deleted** — counts verified against the branch (three-dot). All confined to 4 test groups, all consistent with the fix.

### `DistinctTests.OrderByExpressionDistinct` — restored secondary ORDER BY key (61 providers, modified)
Pre-existing test. The optimizer was dropping the secondary key; it now emits both.

<details><summary>sample diff (SqlServer.2019.MS)</summary>

```diff
 ORDER BY
-	Coalesce([x_1].[MiddleName], N'-2') DESC
+	Coalesce([x_1].[MiddleName], N'-2') DESC,
+	Coalesce([x_1].[LastName], N'-1')
```
</details>

### `OrderByDistinctTests.OrderByDistinctTakeOrdering` — new test (61 providers, added)
`OrderBy(expr).Distinct().Take(n)`: outer paging over a `SELECT DISTINCT` subquery, `ORDER BY OrderData1 * 100 + Id DESC` preserved. Correct across every provider family (paging/quoting variation only; Oracle 11 ROWNUM wrap as expected).

### `OrderByDistinctTests.OrderByGroupByTakeOrdering` — new test (61 providers, added)
`OrderBy(expr).GroupBy(keys).Take(n)`: flat `GROUP BY … ORDER BY … DESC` + provider paging. Correct everywhere; YDB orders by the key alias (semantically equivalent).

### `OrderByDistinctTests.OrderByGroupByTakeOrderingKeyNotProjected` — new test (61 providers, added)
The grouping-key-path variant: `OrderData2` appears only in the inner DISTINCT projection, not the outer SELECT, yet the ORDER BY survives because every leaf column is functionally determined by the projection.

No cross-provider anomalies.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed `ORDER BY` being dropped from `OrderBy(...).Distinct().Take(n)` and `...GroupBy(...).Take(n)` queries, which let `Take` / `Skip` return arbitrary rows. The ordering is now preserved when every column it references is produced by the `DISTINCT` projection or the `GROUP BY` keys. ([#5632](https://github.com/linq2db/linq2db/pull/5632))


_(none)_


---
<sub>Generated from commit `4be07a4`.</sub>

PR #5633, opened 2026-06-19
[NO MERGE] Revert "Refactor query cache: global bucketed dictionary with eviction (#5501)"

MaceWindu:
Diagnostic revert of #5501 (commit c58b4465b) to test the hypothesis that
the global bucketed query-cache refactor is responsible for the Azure
CI test-run slowdown.

This is an investigation PR, not a merge candidate — it exists to get a
clean CI timing comparison against master with the prior per-Query<T>
single-bucket cache restored.

Scope: pure `git revert` of c58b4465b, no manual edits. Restores the
per-Query<T> cache, the MemoryCache-based CompiledTable<T> path, and
removes QueryCache.cs / QueryCacheEvictionTests.cs / CacheActivityBenchmark.cs.

Verified locally: core library builds clean in Release (net10.0); Linq
test project compiles. No PublicAPI surface change (the QueryCache type
was internal).


MaceWindu:
nope, just azure outage

PR #5634, opened 2026-06-19
test infra: isolate + restore tests mutating the global Common.Convert<,> converter

MaceWindu:
Pre-flight cleanup for #5614 (parallel test execution).

### Problem

Client-side `Sql.Convert(...)` / `ConvertTo<>.From` read the **process-global** `Common.Convert<TFrom,TTo>` static converter. Two tests overwrite it globally:

- **`MappingSchemaTests.BaseSchema2`** sets `Convert<DateTime,string>.Lambda` and `Convert<string,DateTime>.Expression` and **never restores them**. That permanently flips e.g. string→`DateTime` to `DateTime.Parse`, so a time-only string ("9:01:01") resolves to **today's date** instead of min-date.
- **`Common.ConvertTests.SetExpression` / `ToStringTest`** set `Convert<int,string>` / `Convert<DateTime,string>` (these already reset, but weren't isolated).

Serially this is masked by fixture ordering (`Tests.Linq.ConvertTests` runs before `Tests.Mapping.MappingSchemaTests`); under **parallel** execution it corrupts concurrent conversion-sensitive tests — `ConvertTests.ToSqlTime`/`ToSqlTimeSql` resolved a today-dated value against the DB's min-date on every real-`TIME` provider.

### Change

Mark the three tests `[NonParallelizable]` (so they don't run alongside conversion-sensitive tests) **and** restore the global converters after the test — `BaseSchema2` was missing the restore entirely (a latent bug even serially).

Verified red→green locally on MySQL under the parallel harness: `ConvertTests` + `MappingSchemaTests` + `Common.ConvertTests` went from `ToSqlTime` failing → **432/432 passing**. No production code touched.

### DIFF
--- a/Tests/Linq/Common/ConvertTests.cs
+++ b/Tests/Linq/Common/ConvertTests.cs
@@ -31,21 +31,28 @@ namespace Tests.Common
 
 		// NonParallelizable: mutates the process-global Common.Convert<TFrom,TTo> converter, which
 		// Sql.Convert / ConvertTo read across all tests - would corrupt concurrent conversion-sensitive
-		// tests under parallel execution. Restored at the end of the test.
+		// tests under parallel execution. Restored in a finally so a failed assertion can't leak it.
 		[Test, NonParallelizable]
 		public void SetExpression()
 		{
-			Convert<int,string>.Lambda = i => (i * 2).ToString();
+			try
+			{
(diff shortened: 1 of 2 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `d9aa52f`.</sub>

PR #5639, opened 2026-06-20
Make OptimizeForSequentialAccess a context-level option

MaceWindu:
## Problem

`Configuration.OptimizeForSequentialAccess` is read at query-**build** time (`QueryRunner`, `RecordReaderBuilder`, `ConvertFromDataReaderExpression`, `CommandInfo`) to decide whether the compiled data-reader materialization reads columns in strict ordinal order. That choice changes the compiled plan.

Because the setting was a process-global static, a plan compiled while it was **off** (random column access) could be reused by a context that opens its reader with `CommandBehavior.SequentialAccess` (or vice-versa):

```
System.InvalidOperationException : Invalid attempt to read from column ordinal '0'.
With CommandBehavior.SequentialAccess, you may only read from column ordinal 'N' or greater.
```

Being process-global also makes the option impossible to vary per context — which blocks parallel test execution (#5614), where the `SqlServer.SA` / `MySql.5.7` lanes toggle the global option while other lanes share the query cache.

## Fix

Move the setting from a process-global static to a per-context `LinqOptions.OptimizeForSequentialAccess` option, with `WithOptimizeForSequentialAccess` / `UseOptimizeForSequentialAccess` builders.

`LinqOptions` already participates in `DataOptions.ConfigurationID`, which `Query.Compare` gates the query-cache lookup on, so sequential-access and non-sequential plans now land in distinct cache slots automatically — no separate `QueryFlags` bit is needed.

`Configuration.OptimizeForSequentialAccess` is kept as a back-compat proxy property over the global default `LinqOptions`, so existing global usage keeps working.

## Test infra

The `.SA` test provider sets the option per-context in the `GetDataConnection` / `GetDataContext` factory (direct and the remote server-side connection) instead of flipping the process-global flag in `OnBeforeTest` / `OnAfterTest`, making it safe under parallel execution. The `SelectTests` SequentialAccess tests use a per-context flag, and a regression test asserts the option changes `DataOptions.ConfigurationID`.

## Follow-up commit `6bb0b666c`

- **`SequentialAccessTest_CacheKey`** — regression test that warms the query cache for a configuration with `OptimizeForSequentialAccess` off (random column-access plan), then re-runs the same query on the same configuration with the option on and a `SequentialAccess` reader, asserting the two materialization plans occupy distinct query-cache slots (the collision the per-context option fixes).
- **Remote `.SA` parity** — the `GetDataContext(string, Func<DataOptions,DataOptions>, RemoteTransport)` overload now applies the `AllSqlServerSequentialAccess` handling (option + interceptor + mapping schema) on the remote server connection, matching the sibling remote overload and the non-remote `GetDataConnection` path.

## Follow-up commit `4631688a6`

- **`SequentialAccessTest_CacheKey` warm-up comment** — clarified that on the SqlServer `.SA` lane the factory forces `UseOptimizeForSequentialAccess(true)` for every context, so the warm-up there is already a sequential-access plan; the warm-OFF→collide-ON cache-key divergence is exercised by the other `[DataSources]` providers. Comment-only change.

## Follow-up commit `d20dd1bd7`

- **`LinqOptions` binary-compat shim** — pin the shim constructor's `UpsertEmulationPolicy` forward to the explicit `.Allow` member instead of `default` (equivalent today since `Allow` is the zero value; resilient if the enum is ever reordered).
- **Copy-constructor doc** — note that the user-declared `LinqOptions(LinqOptions original)` copy constructor is not dead code: it replaces the compiler-synthesized copy ctor every `with` expression invokes and intentionally omits the obsolete no-effect parameters.


igor-tkachev (dismissed):

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] **Core mechanism is sound.** `OptimizeForSequentialAccess` now participates in `LinqOptions` → `DataOptions.ConfigurationID`, which `Query.Compare` gates the cache lookup on, so sequential-access and random-access materialization plans land in distinct cache slots. All four query-build read sites (`QueryRunner`, `RecordReaderBuilder`, `ConvertFromDataReaderExpression`, `CommandInfo`) read the option from the in-scope context's options; the only nullable site (`ConvertFromDataReaderExpression`) is guarded by an earlier `dataContext == null` return.
- [x] **Public API recorded correctly.** `LinqOptions.OptimizeForSequentialAccess` (get/init), `DataOptionsExtensions.WithOptimizeForSequentialAccess` / `UseOptimizeForSequentialAccess`, and the grown `LinqOptions` primary-ctor + `Deconstruct` signatures are all present in `PublicAPI.Unshipped.txt` with XML docs on each new member. The `Configuration.OptimizeForSequentialAccess` field→property change is encoded as a `*REMOVED*` field + get/set and needs no `CompatibilitySuppressions.xml` entry. The change is source-compatible but **not** binary-compatible (assemblies compiled against the old static field must recompile) — worth a release-notes line for 6.4.0.
- [x] **Test warm-up comment fidelity** addressed in follow-up commit `4631688a6` — `SequentialAccessTest_CacheKey`'s warm-up comment claimed a random column-access plan, which doesn't hold on the SqlServer `.SA` lane (the factory forces the option ON for every `AllSqlServerSequentialAccess` context). The comment now notes that the warm-OFF→collide-ON divergence is exercised by the other `[DataSources]` providers.

## Out-of-scope observations

- **Binary-compat `LinqOptions` constructor forwards `UpsertEmulationPolicy: default`** — the pre-existing binary-compatibility constructor passes `UpsertEmulationPolicy: default` to the full ctor; if `UpsertEmulationPolicy.Allow` is not the zero value, callers compiled against the old signature would silently get a non-`Allow` policy. This predates this PR (which only extends the matching `Deconstruct`), and the new `OptimizeForSequentialAccess` param correctly defaults to `false` through this path. Worth a look during the v7 shim removal.
- **Unused private `LinqOptions(LinqOptions original)` copy constructor** — the explicit field-by-field private copy ctor has no callers in `Source/LinqToDB` (record `with` uses the compiler-generated copy ctor). This PR correctly adds the new field to it, but it appears to be dead code; a maintainer may want to remove it rather than keep it in sync with every future field.

## Baselines
Delta: [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5639). 59 SQL baseline changes:

- **New: `SequentialAccessTest_CacheKey` (58 providers)** — the new regression test; each file records two identical `SELECT … FROM InheritanceParent` blocks (the query run once per cache slot). SQL shape is uniform modulo routine per-provider syntax (YDB splits to its own pattern for `col AS col` aliasing). The baseline's point is the cache-key split, not SQL shape — correct.
- **Modified: `Issue1174Tests.TestConcurrentSelect` (ClickHouse.Octonica)** — loses its duplicate second query block. This is a `[DataSources]` *concurrent*-select test whose baseline captures however many query blocks were logged during a concurrent run, so the block count is non-deterministic run-to-run. The PR touches neither ClickHouse nor concurrency — a benign baseline-flake artifact, not a PR-caused change.

sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- `OptimizeForSequentialAccess` is now a per-context option (`LinqOptions.OptimizeForSequentialAccess`, with `DataOptions.UseOptimizeForSequentialAccess(...)`) instead of only a process-global static. This fixes an `InvalidOperationException` ("Invalid attempt to read from column ordinal … With CommandBehavior.SequentialAccess …") that could occur when a cached materialization plan compiled with the option off was reused by a context reading with `SequentialAccess` (or vice-versa) — sequential and non-sequential plans now occupy distinct query-cache slots. The global `Configuration.OptimizeForSequentialAccess` is kept as a back-compat default. ([#5639](https://github.com/linq2db/linq2db/pull/5639))


- `OptimizeForSequentialAccess` is now a per-context option, fixing an `InvalidOperationException` from query-cache collisions between sequential and non-sequential reader plans ([#5639](https://github.com/linq2db/linq2db/pull/5639)).


---
<sub>Generated from commit `963d158`.</sub>

PR #5482, opened 2026-04-21
Fluent Upsert + Insert/Update entity API

sdanyliv:
Part of #2558.
Fix #2528.
Fix #4153.
Fix #3721.
Closes #1480 as alternative design

Adds three fluent entity-level DML APIs and unifies the chain-builder shape across them.

## Upsert

Insert-or-update with a single configure lambda. Replaces the three-expression `InsertOrUpdate` overloads.

```csharp
db.Users.Upsert(user, u => u
    .Match ((t, s) => t.Id == s.Id)
    .Insert(i => i.Set(x => x.CreatedAt, () => DateTime.UtcNow))
    .Update(v => v.Set(x => x.UpdatedAt, () => DateTime.UtcNow)));

// Bulk:
db.Users.Upsert(items, u => u.Match((t, s) => t.Id == s.Id));   // IEnumerable<T>
db.Users.Upsert(query, u => u.Match((t, s) => t.Id == s.Id));   // IQueryable<T>
```

## Insert / Update by entity

```csharp
db.Users.Insert(user, b => b.Set(x => x.CreatedAt, () => DateTime.UtcNow).Ignore(x => x.Notes));
db.Users.Update(user, b => b.Set(x => x.UpdatedAt, () => DateTime.UtcNow));   // PK match
```

Every mapped column from `item` is the default; `.Set` / `.Ignore` overlay on top. Sync + async (`InsertAsync` / `UpdateAsync`). The standalone builders only expose `Set` / `Ignore` — `When` / `DoNothing` are reachable only inside an Upsert branch by design.

## Provider support — Upsert

| Provider | Native path | `.When` predicate | MERGE-lowering (bulk / SkipInsert / non-PK match) |
|---|---|---|---|
| SQLite | ON CONFLICT | ON CONFLICT … WHERE | MERGE |
| PostgreSQL 9.5+ | ON CONFLICT | ON CONFLICT … WHERE | MERGE (15+) |
| SQL Server 2008+ | MERGE | WHEN MATCHED AND | MERGE |
| SQL Server 2005 | UPDATE+INSERT | alt-path (3-query) | alt-path |
| Oracle | MERGE | UPDATE SET … WHERE | MERGE |
| DB2 | MERGE | UPDATE SET … WHERE | MERGE |
| MySQL / MariaDB | ON DUPLICATE KEY UPDATE | alt-path (3-query) | not supported |
| Sybase | UPDATE+INSERT | alt-path (3-query) | not supported |
[...]

PR #3886, opened 2022-12-18
LinqToDB.Identity migration

MaceWindu:
- `LinqToDB.Identity` project moved to main [mono-repo](https://github.com/linq2db/linq2db)
- project assembly name changed from `LinqToDB.Identity` to `linq2db.Identity` to match naming pattern, used by other project
- dropped support for legacy `Microsoft.AspNetCore.Identity` membership model and switched to `Microsoft.Extensions.Identity.Core` model
- add `UserOnlyStore` store support
- `IConnectionFactory` removed (change my mind)
- reuse more code from `Microsoft.Extensions.Identity.Core` instead of re-implementation

Remaining tasks:
- define default mappings
- add tests
- update various readme files
- review issues in old identity repo
- archive old identity repo

Superseeds  #2533

Issue #21, opened 2013-06-05
SQLite.Interop.dll And System.Data.SQLite.dll old versions support.


fmfer:
Hello. First let me thank you for your responsiveness and all the good work you are doing over linq2db.

I was experimenting with linq2db a little more, trying to adapt it to our project and not the other way around.
We happen to be using an old version of System.Data.SQLite.dll (1.0.82.0) to target old frameworks.
- Is there a particular reason linq2db (Linq To SQLite flavor ) is shipped with the newest version of System.Data.SQLite.dll (1.0.86.0)? it seems to work so far with the old one (82), and I'm hesitant to switching to an older version of linq2db since I cannot find any change log for linq2db to see what has been fixed/added in newer versions. 
- If there's a change log for lin2db / Linq To SQLite could you point me to it?
- If this isn't the right place for this questions please tell me so I don't flood your issues tracker with my questions :)

Best regards.
Marcelo.

Issue #18, opened 2013-05-29
Linq2db (linq to sqlite) on .NET 4.0 ?


fmfer:
My company requires our product to target .Net 4.0, the compiler complains lin2db needs System.Data.Services 4.0.0.0, which seems to require .Net 4.5. Is there any chance of getting this working? or maybe this is some problem with my setup? To help answer this questions, please I would like to know if it's possible at all to work with linq2db on .NET 4.0 or if it was made depending heavily on 4.5. 

Best regards.
Marcelo.

PR #5640, opened 2026-06-20
Migrate LinqToDB.Identity into the monorepo

MaceWindu:
Migrates the standalone `linq2db.Identity` provider into `Source/LinqToDB.Identity`, modernized onto the `Microsoft.Extensions.Identity.Stores` base types and the current .NET / ASP.NET Core Identity contract. Supersedes #3886.

## What's included
- Layered `UserStore` / `UserOnlyStore` / `RoleStore` over the `Microsoft.AspNetCore.Identity` base store classes; `IdentityDataConnection` / `IdentityDataContext`; `AddLinqToDBStores<TContext>` DI extension (replaces the old `IConnectionFactory`).
- Default `AspNet*` fluent mappings — **DB-schema compatible** with the EF Core provider, so existing Identity databases keep working unchanged.
- Navigation associations — `Roles` / `Claims` / `Logins` on `IdentityUser<TKey>` and `Users` / `Claims` on `IdentityRole<TKey>`, as association **extension methods** on the standard MS types (no derived model needed).
- `IUserPasskeyStore` implemented on .NET 10 (`IdentityPasskeyData` stored as a JSON column).
- `Microsoft.Extensions.Identity.Stores` pinned to the lowest supported runtime floor — `8.0` on `net462`/`netstandard2.0`/`net8.0`, `9.0` on net9, `10.0` on net10 (same pattern as the other `Microsoft.Extensions.*` abstractions; the package ships net462/netstandard2.0 assets on all those versions).
- Tests under `Tests/Linq/Identity`, run across all providers — direct and remote (LinqService).

## Breaking change vs standalone 6.0.0
New major version: the custom `IIdentityUser` / `IIdentityRole` / `IConnectionFactory` surface is replaced by the `Microsoft.AspNetCore.Identity` base types. The default **database schema is unchanged**. The migration guide is in the project readme (`Source/LinqToDB.Identity/readme.md`).

## Bugs fixed (carried from #3886, which never had tests)
- `Entity<T>("AspNetUsers")` passed the table name to the *configuration* parameter, so the fluent mappings were silently ignored → now `Entity<T>().HasTableName(...)`.
- `Extensions.BulkCopyAsync` re-bound to itself and recursed (stack overflow) → now dispatches via `GetTable<T>().BulkCopyAsync`.
- Claim `Id` was not marked identity → now `.IsIdentity()`, matching the EF Core schema.

## Expanded test coverage & further fixes (follow-up commits)
Positive + negative + variant coverage (string/Guid/int keys, `UserOnlyStore`, `IdentityDataContext`, DI + `UserManager`), run both direct and over LinqService. This surfaced and fixed three more issues:
- `SetTokenAsync` didn't persist updates to an existing token (it relied on EF change tracking) → overridden to update/insert.
- The passkey `Data` JSON column had no explicit DB type → DDL failed on .NET 10 → `HasDataType(NVarChar)`.
- `UserStore` didn't declare `IUserRoleStore<TUser>`, so `UserManager` role operations threw `NotSupportedException` → declared it.
- `AddToRoleAsync` now uses the #5482 Upsert API (`SkipUpdate`, insert-or-ignore), so re-adding an existing role is an idempotent no-op rather than a PK violation.
- Replaced the orphaned `Model.IdentityUser` / `Model.IdentityRole` shadow nav types (shadowed by the MS types, unreachable via the default store path, untested) with the navigation associations above — extension methods rather than properties because C# forbids extension-property access in expression trees (CS9296).

Plus EF parity: the in-memory `ConcurrencyStamp` is refreshed after a successful optimistic `UpdateAsync`, so the same instance stays usable for a further `Update`/`Delete`.

## Resolves (standalone `linq2db/LinqToDB.Identity`)
- linq2db/LinqToDB.Identity#15 — `CreateTable` no longer emits `` IdentityUserClaim`1 ``; the default mappings name the `AspNet*` tables (regression test added).
- linq2db/LinqToDB.Identity#12 — `IUserStore` DI resolution covered end-to-end (DI + `UserManager` round-trip test).
- linq2db/LinqToDB.Identity#18 — arbitrary key types (string/Guid/int) supported and tested.
- linq2db/LinqToDB.Identity#21 — int-keyed `UserManager` works; `FindByIdAsync(string)` converts to the key type (regression test added).

_(GitHub doesn't auto-close cross-repo issues from a PR; these will be closed when the standalone repo is retired.)_

## Follow-up commit - `ConcurrencyStamp` refresh via `UpdateOptimisticWithRefresh`

Synced with master and replaced the stores' manual post-update re-`SELECT` with `ConcurrencyExtensions.UpdateOptimisticWithRefreshAsync` (#5643). `UserOnlyStore.UpdateAsync` / `RoleStore.UpdateAsync` now read the regenerated `ConcurrencyStamp` back from the `UPDATE` itself via `OUTPUT` / `RETURNING` where the provider supports it, falling back to a post-update `SELECT` elsewhere, and collapse to the same shape as their sibling `DeleteAsync`.

Behaviour change on ClickHouse: the new API throws rather than report a concurrency verdict it cannot establish, on a provider that reports neither OUTPUT rows nor an affected-row count. The stores were already unusable there - `UpdateOptimisticAsync` returns `0` on every call, so every `UpdateAsync` / `DeleteAsync` reported a concurrency failure.

Verified green direct + LinqService on SQLite.MS (OUTPUT path) and MySql.8.0 (`SELECT` read-back path).

## Follow-up commits - CI fixes

Three independent failures on the previous head, fixed as separate commits.

**`LockoutEnd` had no renderable DDL type on 8 provider configurations** — 556 of the 557 real test failures, across 9 legs. `IdentityUser.LockoutEnd` is `DateTimeOffset?` mapped with no `DataType`, so `CreateTable` asked each provider's builder to render `DataType.DateTimeOffset`; seven have no case for it and `BasicSqlBuilder.BuildDataTypeFromDataType` falls back to printing the enum name, putting the literal token `DateTimeOffset` in the DDL. (Firebird 2.5/3 differ: the builder has a case, but emits the FB4+ `TIMESTAMP WITH TIME ZONE` unconditionally.)

Fixed with one configuration-scoped `DataTypeAttribute` per affected provider, pinning the column to the datetime type that provider's builder already produces. `DataTypeAttribute` rather than a scoped `ColumnAttribute`: `ColumnDescriptor` consults it only when the column attribute left `DataType` undefined, so the shared unscoped mapping stays untouched — a scoped `ColumnAttribute` would replace it wholesale. `DB2.LUW`, `Firebird.2.5` and `Firebird.3` are pinned per dialect rather than per family, because DB2 z/OS and Firebird 4+ render `DateTimeOffset` natively under the same base configuration name.

Verified on SqlCe (pinned): `[LockoutEnd] DateTime NULL`, previously rejected. Oracle 11 as a non-pinned control is byte-identical — `"LockoutEnd" timestamp with time zone`.

**The package suppressed the `linq2db.Analyzers` subtree for its consumers.** The `ProjectReference` to `LinqToDB` carried no `PrivateAssets`, so pack wrote `exclude="Build,Analyzers"` in every TFM group and `verify-analyzer-delivery.ps1` failed `Build and pack` with one violation per group. Now `PrivateAssets="contentfiles;build"`, matching the seven other packable projects referencing `LinqToDB.csproj`.

**The Oracle legs failed and hung on a stale cached cursor.** `Schema` / `KeyedSchema` execute a byte-identical drop block twice on one connection with `CreateTable` in between, so ODP.NET re-executes a cached cursor after a server-side metadata change — `ORA-24449`, plus `ORA-00932` on a block with no binds and `NullReferenceException`s from corrupted driver state. Measured over the 45 identity tests on Oracle 11, same binaries, only the connection string differing:

| Oracle connection string | Failed / 45 |
|---|---|
| default | 8 |
| `Statement Cache Size=0` | 9 |
| `Self Tuning=false;Statement Cache Size=0` | 0 |

Both keywords are needed — self-tuning re-enables the cache when only the size is pinned. `ORA-24449` is also what the Oracle 21c/23c leg reported, where the same tests passed on retry. This is a test-configuration change: the drop-create-drop behaviour is unchanged for users, and CI no longer exercises the driver's default cache setting. Whether it also clears the 5-minute hang on the Oracle 11/12/18/19 legs is unconfirmed — the hang did not reproduce locally; only the `ORA-24449` family did.

## Follow-up commits - provider mapping and gates

Second CI round on `77a981e38`: Azure `test-all` green, GitHub Actions `tests [all]` red on 8 of 22 legs - 410 failing rows, all inside the new Identity fixtures. Six causes, fixed as three commits and verified locally against every affected provider.

**`LockoutEnd` lost its offset on every pinned configuration.** The `DataType` pin from the previous round fixed only the DDL; the value still reached the driver unconverted, so a column that cannot hold an offset round-trips through the *machine's local* offset - Firebird 3 wrote `13:52:49Z` and read back `13:52:49+02:00`. Sybase and SAP HANA never got that far, rejecting the `DateTimeOffset` parameter outright (`Specified cast is not valid. From: DateTimeOffset to: DateTime`), which is what CI reported. The two are the same defect. **CI cannot see the offset half of it, because CI runs in UTC** - it surfaced only on a UTC+2 machine. Each pinned configuration now also carries a configuration-scoped `ValueConverterAttribute` storing UTC and reading back at zero offset.

**Access could not create its own schema.** Its text columns cap at 255 characters and the Identity name/email columns are 256, so `CREATE TABLE` failed with `Size of field 'UserName' is too long` before any row existed - 141 rows across the three Access legs.

**Three engines cannot index the keys the EF-compatible schema declares.** Those tables now emit without the constraint, which beats not being able to create them:

| Provider | Key | Failure |
|---|---|---|
| Access | `CredentialId` (1024 B) | `VarBinary` caps at 255 B - the column itself is rejected; now `LongBinary` |
| Informix | `CredentialId` (1024 B) | `-103 illegal key descriptor` - cannot index a `BYTE` blob |
| Firebird 2.5 | `CredentialId`, and `AspNetUserTokens` (`UserId`+`LoginProvider`+`Name`) | `cannot create index` - 292 chars at `CHARACTER SET UNICODE_FSS` overruns the index-key limit |

Note `IsPrimaryKey = false` rather than an omitted flag: fluent `.IsPrimaryKey()` registers a separate `PrimaryKeyAttribute` that `ColumnDescriptor` consults whenever the column attribute leaves the flag unset, so a scoped attribute has to answer the question rather than decline it. `AddOrUpdatePasskeyAsync` consequently keys its update on the predicate - `Context.UpdateAsync` on a keyless entity has no fields left to set and throws, or silently no-ops under `IgnoreEmptyUpdate`. `Context.DeleteAsync(entity)` degrades safely by the same mechanism, so the token table needed no store change.

**Firebird DDL serialization.** `Schema` / `KeyedSchema` used bare `CreateTable` / `DropTable` pairs; a closed-but-pooled connection keeps a reference and the next `DROP` reports `object TABLE ... is in use`, reddening all four Firebird versions end to end (148 rows). Both now build on `CreateLocalTable`, which carries the per-database pool eviction and drop-retry.

**`ConcurrencyFailure` was asserting nothing on Access.** It corrupted the stored `ConcurrencyStamp` through the setup connection while the rows were inserted through `ctx`; Access does not show one connection the rows another just inserted, so that statement matched 0 rows and the stale-update assertion passed vacuously. Routed through `ctx`; the stamp is still corrupted out-of-band relative to the store's cached entity.

**Gates.** ClickHouse reports no affected-rows count and has no `UPDATE ... RETURNING`, so the store's optimistic update and delete cannot tell a conflict from a success, and `InsertOrUpdate` is unsupported outright. linq2db's YDB provider implements no `InsertOrUpdate`, which the role and token upserts need. Informix can create the passkey table now, but `BYTE` values cannot appear in a comparison, so every passkey lookup filters on a column it refuses to filter on. Gated per test rather than per provider - the attribute is the run-and-verify kind, so each gate asserts the case still fails for the stated reason.

Local verification, all four Firebird versions included:

| Run | Result |
|---|---|
| Firebird 2.5 + Firebird 3 + Sybase + Access ACE (ODBC) | 174/174 |
| Access ACE (OleDb + ODBC) + DuckDB | 131/131 |
| Informix + SAP HANA + SqlCe + DuckDB | 172 green, 2 held by the Informix gate |
| ClickHouse x3 + YDB + DuckDB | 132 green, 85 held, none over-firing |

## Follow-up commit - Firebird 4/5 LockoutEnd pin

Third CI round on `f698a39c3`: Azure `test-all` green; GitHub Actions `tests [all]` red on one leg, `Lin k_Firebird`, with four rows - `UserLockout` on `Firebird.4`, `Firebird.5` and both their LinqService variants. Every other row in that leg was a gate holding as declared.

The earlier round left Firebird 4+ unpinned because its builder renders `TIMESTAMP WITH TIME ZONE`. The DDL was never the whole problem: `FirebirdDataProvider.SetParameterType` maps `DataType.DateTimeOffset` to `FbDbType.TimeStampTZ`, and FirebirdClient 10.3.4 accepts only `FbZonedDateTime` for that type, so the parameter write throws `System.InvalidOperationException : Incorrect time zone value.` at `FirebirdSql.Data.Common.DbValue.GetTimeZoneId()`. `UserLockout` is the only fixture that writes a non-null `LockoutEnd`, which is why it was the only casualty. The provider gap itself is #5915 - blocked upstream on the client - and not this PR's to fix. This supersedes the note above that Firebird was pinned per dialect because 4+ renders the type natively.

The four per-dialect rows collapse to one `ProviderName.Firebird` row: the family name sits in every dialect's mapping-schema chain, so 2.5 through 5 match it and a future dialect will too.

| Dialect | Identity suite (net10.0, local) |
|---|---|
| Firebird.2.5 | green |
| Firebird.3 | 45/45 |
| Firebird.4 | green, incl. the `UserLockout` rows CI reported red |
| Firebird.5 | 45/45; red before the change with the trace above, green after |
## Follow-ups (deferred)
- Promote the readme (incl. migration guide) to the wiki.
- Retire the standalone `LinqToDB.Identity` repo and deprecate its NuGet package.


Source/LinqToDB.Identity/Model/IdentityUser.cs
```diff
@@ -0,0 +1,110 @@
+using System;
+using System.Collections.Generic;
+using ASP = Microsoft.AspNetCore.Identity;
+using LinqToDB.Mapping;
+
+namespace LinqToDB.Identity.Model
```
MaceWindu: **Major · MAJ001** — The `LinqToDB.Identity.Model.IdentityUser` / `IdentityRole` types — the only types carrying the `[Association]` navigation properties (`Roles`, `Claims`, `Logins`, etc.) — are public but unreferenced anywhere in the assembly and untested. A repo-wide search finds `LinqToDB.Identity.Model` only in these two namespace declarations and in `PublicAPI.Unshipped.txt`. The default `IdentityDataConnection` / `IdentityDataContext` declare `: IdentityDataConnection<IdentityUser, IdentityRole, string>` inside `namespace LinqToDB.Identity` with only `using Microsoft.AspNetCore.Identity;`, so those `IdentityUser` / `IdentityRole` names bind to the ASP.NET base types, **not** these `Model` types — and the tests do the same. The migration-guide draft tells users these ready-made navigation-property models are available, yet nothing exercises them, and they're shadowed by same-named ASP.NET types.

Fix: Decide the intent and make it consistent. If the default contexts should expose navigation properties, bind the `LinqToDB.Identity.Model` types in `IdentityDataConnection` / `IdentityDataContext` (add `using LinqToDB.Identity.Model;` or fully-qualify) and add a test that round-trips a navigation property (e.g. load a user and assert `user.Roles`). If they're an opt-in convenience only, add at least one test that constructs a context/store over them and exercises an association. Either way, shipping public, untested surface that callers can't reach through the default context is a trap.

Source/LinqToDB.Identity/Stores/UserOnlyStore.cs
```diff
@@ -0,0 +1,590 @@
...
+			if (result != 1)
+				return IdentityResult.Failed(ErrorDescriber.ConcurrencyFailure());
+
+			// The optimistic update generated a new concurrency stamp server-side; refresh the in-memory entity
+			// so it stays usable for further operations (EF Core refreshes the tracked entity the same way).
+			user.ConcurrencyStamp = await Users
+				.Where(u => u.Id.Equals(user.Id))
+				.Select(u => u.ConcurrencyStamp)
+				.FirstOrDefaultAsync(cancellationToken)
+				.ConfigureAwait(false);
```
MaceWindu: **Minor · MIN001** — After a successful optimistic `UpdateAsync`, the in-memory `ConcurrencyStamp` is refreshed via a `FirstOrDefaultAsync` read-back. If the row is concurrently deleted in the window between the update and this read, the query returns `null` and `user.ConcurrencyStamp` is silently set to `null` while the method still returns `IdentityResult.Success`. Reusing that instance for a later `Update`/`Delete` then compares against a null stamp and fails with a spurious `ConcurrencyFailure` rather than surfacing the real not-found condition. The same read-back exists in `RoleStore.UpdateAsync`. (Separately, the comment above says the new stamp was "generated server-side"; with `VersionBehavior.Guid` the value is produced client-side via `Guid.NewGuid()` on providers that don't translate `NEWID()`, so the wording overstates where it originates.)

Fix: Guard the assignment so a null read-back (row vanished) doesn't overwrite a non-null stamp — capture into a local and only assign when non-null, or treat a null result as a concurrency failure. Apply the matching fix in `RoleStore.UpdateAsync`.

Source/LinqToDB.Identity/Context/IdentityDataConnection.cs
```diff
@@ -0,0 +1,281 @@
...
+	public class IdentityDataConnection<TUser, TKey, TUserClaim, TUserLogin, TUserToken> :
+		DataConnection
+		where TUser      : IdentityUser<TKey>
+		where TKey       : IEquatable<TKey>
+		where TUserClaim : IdentityUserClaim<TKey>
+		where TUserLogin : IdentityUserLogin<TKey>
+		where TUserToken : IdentityUserToken<TKey>
+	{
+		private static readonly Lock _syncRoot = new ();
+		private static MappingSchema?  _mappingSchema;
```
MaceWindu: **Minor · MIN002** — `_mappingSchema` is a `static` field on the 5-parameter base `IdentityDataConnection<TUser, TKey, TUserClaim, TUserLogin, TUserToken>`, but the role mappings (`TRole`/`TUserRole`/`TRoleClaim`) are added by the 8-parameter derived class's `ConfigureMappings` override. A static field of a generic type is shared per closed **base** type, whose type arguments don't include the role types — so two contexts that share the same `<TUser, TKey, TUserClaim, TUserLogin, TUserToken>` but differ only in role types collide on one `_mappingSchema`: whichever is constructed first wins, and the second silently gets role mappings built for the other role type. It's also process-global, so the first configuration to initialize fixes the schema for all later instances of that closed base type.

Fix: Build the schema per instance, or key the cache by the most-derived type (e.g. a `ConcurrentDictionary<Type, MappingSchema>` keyed by `GetType()`) so role-type variation participates. Same issue in `IdentityDataContext` (MIN003).

Source/LinqToDB.Identity/Context/IdentityDataContext.cs
```diff
@@ -0,0 +1,278 @@
...
+	public class IdentityDataContext<TUser, TKey, TUserClaim, TUserLogin, TUserToken> :
+		DataContext
+		where TUser      : IdentityUser<TKey>
+		where TKey       : IEquatable<TKey>
+		where TUserClaim : IdentityUserClaim<TKey>
+		where TUserLogin : IdentityUserLogin<TKey>
+		where TUserToken : IdentityUserToken<TKey>
+	{
+		private static readonly Lock _syncRoot = new ();
+		private static MappingSchema?  _mappingSchema;
```
MaceWindu: **Minor · MIN003** — Same static mapping-schema cache problem as MIN002, in the `DataContext`-based hierarchy: `_mappingSchema` is `static` on the 5-parameter base while role mappings are layered on by the 8-parameter derived override, so the cache key omits the role types and two contexts differing only in role types share one schema (first-constructed wins, process-global).

Fix: Apply the same fix as MIN002 — per-instance schema, or key the cache by the most-derived `GetType()`.

Source/LinqToDB.Identity/Stores/UserOnlyStore.cs
```diff
@@ -0,0 +1,590 @@
...
+				.ToListAsync(cancellationToken)
+				.ConfigureAwait(false);
+		}
+
+		/// <inheritdoc cref="FindByEmailAsync(string, CancellationToken)"/>
+		public override Task<TUser?> FindByEmailAsync(string normalizedEmail, CancellationToken cancellationToken = default)
+		{
+			ThrowIfDisposed();
+
+			return Users.SingleOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
```
MaceWindu: **Suggestion · SUG001** — `FindByEmailAsync` uses `SingleOrDefaultAsync` while the sibling lookups (`FindByIdAsync`, `FindByNameAsync`, `FindByLoginAsync`) use `FirstOrDefaultAsync`. The default schema places no unique constraint on `NormalizedEmail`, and ASP.NET Identity permits duplicate emails unless `RequireUniqueEmail` is set, so when two users share a normalized email this throws `InvalidOperationException` instead of returning a user.

Fix: Use `FirstOrDefaultAsync` for consistency and to tolerate the non-unique-email configuration the schema allows. If single-match is intentional (to surface a misconfiguration), note it in a comment.

```suggestion
			return Users.FirstOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
```

Tests/Linq/Identity/IdentityStoreVariantTests.cs
```diff
@@ -0,0 +1,212 @@
...
+			Assert.That(role.Id, Is.GreaterThan(0));
+
+			await users.AddToRoleAsync(user, "IADMIN");
+			Assert.That(await users.IsInRoleAsync(user, "IADMIN"), Is.True);
+
+			Assert.That((await users.DeleteAsync(user)).Succeeded, Is.True);
+		}
+
+		[Test]
+		public async Task DependencyInjection_ResolvesStores_AndUserManagerRoundTrips([IncludeDataSources(TestProvName.AllSQLite)] string context)
```
MaceWindu: **Suggestion · SUG002** — The PR carries fixes for four standalone-repo issues (#15, #18, #21, #12), but only #15 (`DefaultMappingsUseAspNetTableNames`) and #21 (`DependencyInjection_IntKeys_UserManagerRoundTrips`) are annotated with their issue numbers. This test covers #12 (DI store resolution) and the Guid/int-key variant tests cover #18 (non-string key types), yet neither names the issue it regresses.

Fix: Add a `// Regression for LinqToDB.Identity#12: ...` comment here and a `// Regression for LinqToDB.Identity#18: ...` comment on the Guid/int-key tests, matching the style already used for #15 and #21. Comment-only change.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [ ] New public assembly `LinqToDB.Identity`: the public surface is tracked in `PublicAPI.Unshipped.txt` (plus a `net10.0` overlay for the passkey API) and matches the actual public types; the `internal` helpers (`DefaultMappings`, `Extensions`, `Resources`) are correctly hidden. No `CompatibilitySuppressions.xml` work applies — there's no prior in-repo shipped version of this assembly.
- [ ] `Microsoft.Extensions.Identity.Stores` 8.x ships `net462`/`netstandard2.0` assets, so the lowest-runtime-floor pin (8.0 / 9.0 / 10.0) builds across the project's TFMs — confirmed against the package on nuget.org.

All findings below are attached to specific lines as inline comments.

## Out-of-scope observations

Surfaced during review but fall outside this PR's core intent. Not findings on this PR — included as FYI.

- **`AddToRoleAsync` doesn't pre-check existing membership** — it inserts the user-role row directly. `AspNetUserRoles` has a composite `(UserId, RoleId)` primary key and no identity column, so re-adding a role the user already has surfaces as a primary-key violation from the provider rather than a clean `IdentityResult` failure. EF Core's store doesn't guard here either (`UserManager` dedups one layer up), so this matches upstream — flagged only so the divergence from a defensive insert is a conscious choice.
- **Passkey `CredentialId` is a 1024-byte `VarBinary` primary key — unverified beyond SQLite** — `SetupIdentityUserPasskey` maps `CredentialId` (`byte[]`) as the primary key with `HasLength(1024)`. SQL Server's maximum index key size is 900 bytes, so a `varbinary(1024)` PRIMARY KEY can fail at `CREATE TABLE`. The new tests exercise the schema only on SQLite, so this is unverified on SQL Server / other providers; worth confirming `CreateTable<IdentityUserPasskey<TKey>>()` produces a valid PK there and reconciling the 1024 length against EF Core's own passkey column definition. (.NET 10 only.)
- **Passkey `Data` JSON column has no length** — mapped `HasDataType(NVarChar)` with no `HasLength`. On SQL Server this is correct (emits `nvarchar(max)`, matching EF Core), but serialized passkey JSON runs to kilobytes and bounded-NVARCHAR providers (e.g. Oracle `NVARCHAR2`, capped at 4000/32767) may truncate or reject it. Unverified beyond SQLite.
- **`LinqToDB.Identity.Model` sub-namespace placement** — independently of the orphaned-types finding (MAJ001 inline), placing the model types under `.Model` while the rest of the surface sits directly under `LinqToDB.Identity` avoids the simple-name collision with `Microsoft.AspNetCore.Identity.IdentityUser`/`IdentityRole`, but keeps that collision latent for callers (a bare `using LinqToDB.Identity;` resolves those names to the ASP.NET types). A namespace-organization judgment call for the maintainer.
- **A few negative/validation tests provision full DDL they don't need** — `NullArguments`, `RoleNameValidation`, `DisposedStore`, `IdempotentRemovals` build the full schema even though several of their assertions throw during argument validation before touching the database. Harmless and keeps fixtures uniform; could drop the schema setup for the pure-validation cases if SQLite table creation ever becomes a bottleneck.

## Baselines
No `baselines/pr_5640` branch exists yet — the CI baseline run hasn't landed. This PR adds an isolated new project with new Identity tests, so once CI runs, expect *added* baselines for the new tests and no modifications to existing baselines.

PR #5643, opened 2026-06-21
ConcurrencyExtensions: UpdateOptimisticWithRefresh (#5642, fixes #4194)

### LINKED ISSUE #4194: UpdateOptimistic doesn't change concurrency token on entity
```
builder.Property(x => x.ConcurrencyToken).HasAttribute(new OptimisticLockPropertyAttribute(VersionBehavior.Guid));
```

```
var entity = await dataConnection.GetTable<TEntity>().FirstAsync();
entity.Test = "123";
// entity.ConcurrencyToken == 7731dbb6-a04b-421d-979b-fdb18375b922
await dataConnection.UpdateOptimisticAsync(entity); // in DB guid changed to 2796e69c-2c87-4b47-856f-e507f78b9028
// entity.ConcurrencyToken == 7731dbb6-a04b-421d-979b-fdb18375b922
```

I expect that value of entity.ConcurrencyToken changed to  2796e69c-2c87-4b47-856f-e507f78b9028 after UpdateOptimisticAsync, but value in object doesn't changed

### Environment details

`Linq To DB` version: 5.1.0

### LINKED ISSUE #5642: ConcurrencyExtensions: UpdateOptimistic/DeleteOptimistic overloads returning the regenerated version via OUTPUT/RETURNING
## Summary

Add an `UpdateOptimistic` overload to `ConcurrencyExtensions` that writes the **regenerated** optimistic-lock
value(s) back onto the entity, reading them in the same statement via `UpdateWithOutput` (`OUTPUT` / `RETURNING`)
where the provider supports it — removing the extra `SELECT` callers currently need to refresh the version after an
optimistic update.

## Motivation

`OptimisticLockPropertyAttribute(VersionBehavior.Guid/AutoIncrement)` regenerates the lock column on each
`UpdateOptimisticAsync`. Today, a caller that needs the new value after the update must issue a **separate `SELECT`** to
read it back. That is an extra round-trip and is racy (the row can change between the `UPDATE` and the `SELECT`).

Concrete example — `linq2db.Identity`'s `UserStore` / `RoleStore` `UpdateAsync` (PR #5640) refreshes `ConcurrencyStamp`
exactly this way:

```csharp
await Users.UpdateOptimisticAsync(user, token);
var stamp = await Users.Where(u => u.Id.Equals(user.Id))
    .Select(u => u.ConcurrencyStamp).FirstOrDefaultAsync(token);   // <-- extra SELECT + race window
if (stamp == null) return IdentityResult.Failed(ErrorDescriber.ConcurrencyFailure());
user.ConcurrencyStamp = stamp;
```

## Proposal

New overloads that fold the read-back into the modifying statement using the existing `*WithOutput` machinery:

```csharp
// returns affected count; sets the regenerated version member(s) on `obj` from OUTPUT/RETURNING
int       UpdateOptimisticWithRefresh<T>(this IDataContext db, T obj);
Task<int> UpdateOptimisticWithRefreshAsync<T>(this IDataContext db, T obj, CancellationToken token = default);

// same, for a query source with optional filtering / table-name override
int       UpdateOptimisticWithRefresh<T>(this IQueryable<T> source, T obj);
Task<int> UpdateOptimisticWithRefreshAsync<T>(this IQueryable<T> source, T obj, CancellationToken token = default);
```

The overload outputs the optimistic-lock column(s) and writes them straight back onto `obj`, so the caller's entity
carries the fresh version with no follow-up query.

## Provider considerations

- Read-back uses `UPDATE … OUTPUT` / `RETURNING` where the provider supports it, declared by a new
  `SqlProviderFlags.IsUpdateOutputSupported` and kept honest by a probing guard test.
- Where it is unavailable, the value is read back with a follow-up `SELECT`, gated on a reliable affected-rows count
  (`SqlProviderFlags.IsAffectedRowsCountSupported`).
- Providers that support **neither** cannot guarantee the optimistic-concurrency result at all, so the method throws
  for them rather than guessing best-effort. This is a deliberate departure from the original "works everywhere"
  requirement: a silently wrong concurrency verdict is worse than an explicit failure. Affects ClickHouse.
- Behaviour otherwise matches the existing `UpdateOptimistic` concurrency semantics (0 rows affected ⇒ concurrency
  failure), except that an entity declaring no optimistic-lock column degrades to a plain update and returns the raw
  provider count.

## Out of scope

- A projected-output variant (`UpdateOptimisticWithOutputAsync<T, TOut>` taking an output expression) — not needed by
  the `linq2db.Identity` migration that motivated this issue.
- `DeleteOptimistic` equivalents returning the deleted row's values — `DeleteOptimistic` / `DeleteOptimisticAsync`
  already exist, and returning the deleted values is not required here.

## Benefit

- One round-trip instead of two for optimistic update + version refresh.
- Atomic read-back on providers with `OUTPUT` / `RETURNING` (no `UPDATE`→`SELECT` race).
- Cleaner consumer code (removes the manual re-`SELECT` in the Identity stores and any user code doing the same).

_Filed as a follow-up from the `linq2db.Identity` migration (#5640)._

MaceWindu:
Adds `UpdateOptimisticWithRefresh` / `UpdateOptimisticWithRefreshAsync` overloads
(`IDataContext` + `IQueryable`) that perform an optimistic update and write the
regenerated optimistic-lock column(s) back onto the entity.

The new value is read from the same statement via `OUTPUT` / `RETURNING` where the
provider supports it (SqlServer, PostgreSQL, SQLite, Firebird, YDB), falling back
to a post-update `SELECT`-by-PK elsewhere so the API works on every provider.
`0` affected rows ⇒ concurrency failure (entity left untouched).

Fixes #4194 — `UpdateOptimistic` never refreshed the entity's version column.
Removes the manual re-`SELECT` the `linq2db.Identity` stores do (#5640).

Provider `OUTPUT` support is declared by the per-provider
`SqlProviderFlags.IsUpdateOutputSupported` flag and kept honest by the
`ConcurrencyRefreshTests.OutputSupportSurface` guard test, which probes each
provider's actual UPDATE `OUTPUT` support and fails on divergence.

Closes #5642

## Follow-up commit — no-rowcount providers + guard-test repair

Two defects found in review and verified live on YDB + ClickHouse:

- **Fallback refresh skipped on no-rowcount providers.** The `SELECT`-fallback read-back was gated on `UPDATE`'s affected-row count, which is `0` even on success for providers that don't report it (ClickHouse) — so the lock column was never refreshed and `0` was returned as a false concurrency failure. Added `SqlProviderFlags.IsAffectedRowsCountSupported` (default `true`; `false` on ClickHouse and YDB) and gated the read-back on `count > 0 || !IsAffectedRowsCountSupported`, making the refresh **best-effort** where the count is unreliable. Kept honest by a new `AffectedRowsCountSurface` guard test.
- **`OutputSupportSurface` probe inserted into the wrong table.** It created `ConcurrencyRefreshProbe` via `CreateLocalTable` but inserted via `db.Insert` into the entity's default-named table, so the probe always threw and returned `false` — failing on **every** `IsUpdateOutputSupported=true` provider (SQLite/SqlServer/PostgreSQL/Firebird/YDB), which is what the prior red CI run reflected. Fixed to insert into the probe table by name so it actually exercises `UPDATE … OUTPUT`.

Note: on providers that don't report affected rows, the refresh is best-effort (no single-statement or concurrency-detection guarantee), consistent with the existing optimistic-concurrency API's stance on those providers.

## Follow-up commit — verify-based no-rowcount refresh + plain-API docs

Replaces the best-effort no-rowcount fallback described above. In review, @igor-tkachev flagged that returning raw `0` on success contradicted the documented `0 == concurrency failure` contract, and that the PK-only read-back mutated the entity even on a stale-token failure.

- **Success verified by written columns.** When `IsAffectedRowsCountSupported` is `false`, success is now confirmed by re-reading the row filtered by primary key **and** the values written to the non-lock columns: a match means our write is the persisted state -> refresh the lock column(s) and return `1`; no match -> return `0` and leave the entity untouched. This restores the `1/0` contract without a `-1` sentinel. Scope is ClickHouse-only (YDB takes the OUTPUT/RETURNING path and returns an accurate count). Best-effort where the provider's UPDATE is asynchronous (a ClickHouse mutation can be read before it applies), documented on the `<returns>` docs. Covered by a new `UpdateOnNoRowcountProviderReportsResult` test, green on ClickHouse.
- **Plain optimistic API.** Documented the same no-rowcount count caveat on the pre-existing `UpdateOptimistic` / `DeleteOptimistic` overloads and aligned their argument null-checks (no behavior change).

## Follow-up commit — reject on providers that can't report the result

Review follow-up (@igor-tkachev): the no-rowcount `SELECT`-fallback still broke the `1/0` contract when the entity had only PK + lock column (`comparable.Length == 0`) — it refreshed the entity yet returned raw `0`. Rather than extend the best-effort heuristic, `UpdateOptimisticWithRefresh` now **throws** `LinqToDBException` (before executing any UPDATE) on providers that support neither UPDATE `OUTPUT` / `RETURNING` nor a reliable affected-rows count (ClickHouse): such providers cannot guarantee the optimistic-concurrency result, so best-effort is not offered. Removes the written-column verify path (`GetComparableColumns`); the reliable affected-rows read-back is unchanged. Test `UpdateOnNoRowcountProviderReportsResult` replaced by `UpdateOnUnsupportedProviderThrows` (gated by the capability flags, not a provider name; red on the old code, green on ClickHouse direct + remote).

## Follow-up commits — review findings

Five commits from a review pass. Each defect was reproduced red before the fix and re-verified green after, on ClickHouse / MySQL / SQLite / DuckDB in both direct and LinqService modes (128 tests, 0 failed).

- **The fallback read-back inherited the caller's predicates.** The `SELECT`-fallback composed the primary-key filter onto the caller's `source`, so it re-applied predicates that the `UPDATE` itself had just invalidated — it matched nothing, skipped the refresh, and still returned a non-zero count, silently restoring the #4194 behaviour this API removes. It now reads back through the table `source` is rooted on, preserving `TableName(...)` overrides while dropping the caller's query operators. Regression test `UpdateViaQueryWithFilterOnUpdatedColumn`, red on MySQL direct and remote.
- **Five tests called the API on a provider where it throws.** They gated on `SupportsRowcount()` (i.e. `IsAffectedRowsCountSupported`), but the effective capability is `IsUpdateOutputSupported` OR `IsAffectedRowsCountSupported`. The gate suppressed only the count assertion, leaving the call unguarded, so they threw on ClickHouse — 10 failures across direct and LinqService. Replaced with a capability gate, and `UpdateConcurrencyFailureLeavesEntityUnchanged` no longer over-skips YDB, where the OUTPUT path makes the contract enforceable. This is @igor-tkachev's remaining note on `4ddd0a553`, which turned out to be a hard failure rather than an imprecision.
- **Constructor-mapped entities threw.** The lock-column projection used `new T { ... }`, which needs a parameterless constructor, so an entity mapped through a constructor failed with `ArgumentException` on both the OUTPUT and the fallback path. Those entities now project the whole row and materialize through the normal constructor mapping. Found by Copilot review; regression test `UpdateRefreshesEntityWithoutDefaultConstructor`.
- **Tests and infrastructure.** Async counterparts for the three async branches that had no coverage (concurrency failure, unsupported-provider throw, no-lock-column short-circuit). The capability probes no longer swallow setup failures as "unsupported", which had made the guard pass vacuously for every provider whose flag is `false`. `SupportsRowcount` now reads `IsAffectedRowsCountSupported` instead of duplicating it. DuckDB added to the `UpdateWithOutput` feature constants — 42/42 pass, which confirms its `IsUpdateOutputSupported` empirically rather than by inference.
- **Documentation.** `<returns>` no longer states unconditionally that `0` means a concurrency failure: the no-lock-column short-circuit returns the raw provider count. Also documented that the fallback read-back is a separate statement and so needs an ambient transaction to guarantee it observes this update's own value, and that on SQL Server the OUTPUT path cannot target a table with an enabled UPDATE trigger.
- **Firebird before v5 reported a concurrency failure as success.** `UPDATE … RETURNING` is a singleton there, yielding one record whatever the statement matched, so a zero-row update was indistinguishable from a one-row update and the entity was stamped with a value that was never persisted. Caught by `test-all` build 22880, which failed only on Firebird 2.5/3/4. The capability flag is now gated on `Version >= FirebirdVersion.v5` so those versions use the affected-rows plus read-back path, `IsUpdateOutputSupported` is renamed to `IsUpdateOutputRowsSupported` with docs stating the real contract (the output must be a result set of the rows actually updated, so a non-matching `UPDATE` returns none), and the capability probe now enforces that contract as well. This defect predates the review commits: the failing test has been in the PR since `4208f396a`.

## Follow-up commit - Oracle 11 PK collision, MIN006, NIT001

- **Oracle 11 rejected the new source-shape test's schema.** `UpdateThroughUnsupportedSourceShapeThrows` creates two tables, and `BasicSqlBuilder` names each PK constraint `"PK_" + table name`, truncating it through `IdentifiersHelper.TruncateIdentifier` - a 30-char budget on Oracle 11, 128 from Oracle 12. `PK_ConcurrencyRefreshShapeOuter` is 31 chars and the truncation drops `sizeDecrement + 4` = 5 of them, landing exactly on `PK_ConcurrencyRefreshShape`, the other table's own untruncated PK name. Oracle's constraint namespace is schema-global, so the second `CREATE TABLE` failed with `ORA-02264` on both the direct and LinqService contexts - the only red leg in build 23083. The outer table is renamed to `ConcurrencyRefreshOuter`, whose PK name is 26 chars and so neither truncates nor collides; green on Oracle 11 locally.
- **MIN006: the `IgnoreFilters` re-issue branch had no coverage.** `ReadBackExpression` re-issues `IgnoreFilters` over the reduced inner source, but `UpdateViaQueryWithIgnoreFilters` builds its source as `t.IgnoreFilters()` with nothing in between, so the inner expression is an `ITable`, the `ReferenceEquals` check holds, and the method returns early. New test `UpdateViaQueryWithIgnoreFiltersAboveWhere` puts a droppable operator between the table and `IgnoreFilters`. Neutering the re-issue to `return inner` turns it red on both Oracle 11 contexts while the existing test stays green - which is what shows the branch was previously unreached. The failure is the stale-token shape: `cnt` is 1 and `record.Stamp` is still all-zeros.
- **NIT001: comments.** The comment introducing that branch described `IgnoreFilters` as "the one dropped operator" inside the branch that keeps it, and the block comment above `ReadBackSource` stated the drop rule without its exception. Both corrected.

## Follow-up commit - error messages to ErrorHelper

Review follow-up (@sdanyliv): the unsupported-provider message named `SqlProviderFlags.IsUpdateOutputRowsSupported` / `IsAffectedRowsCountSupported`, both of which live in `LinqToDB.Internal.SqlProvider` - names a caller cannot act on. It now describes the capability instead. All three of the API's messages move to `ErrorHelper` (`Error_Concurrency_UpdateWithRefresh_NotSupported` / `_ReadOnlyLockMember` / `_UnsupportedSource`, the last two as format strings), and the repeated `"Linq To DB query expected"` literal becomes `Error_LinqToDBQueryExpected`, applied to all five sites in the file. `EnsureLockColumnsWritable`'s throw is now covered by the `<exception>` block on all four overloads, which previously documented only the other two conditions. Tests assert against the constants rather than message substrings.


sdanyliv:
@MaceWindu, we have new Update API and you can extend it something like this (just idea)
```cs
table.Update(obj, b => b
   .UseConcurrencyField(x => x.ConcurrencyToken)
   .Fetch(x => x.UpdatedAt)
   .Fetch(x => x.ConcurrencyToken)
)
```

MaceWindu:
This API (except existing issue) I need to finish Identity provider migration to avoid manual update of token and extra select when database supports RETURNING.
I will put new Update API extension to separate feature request

igor-tkachev (changes requested):
I think this PR needs one contract fix before merge.

The general idea of `UpdateOptimisticWithRefresh` looks good: after a successful optimistic update, the regenerated optimistic-lock value should be written back to the entity. This fixes the current problem where the database row gets a new token, but the in-memory entity keeps the old one.

The issue is the fallback path for providers that don't support reliable affected-row counts.

Currently the code does roughly this:

```cs
var count = updatable.Update();

if (count > 0 || !dc.SqlProviderFlags.IsAffectedRowsCountSupported)
{
    var fresh = FilterByPrimaryKey(...)
        .Select(...)
        .FirstOrDefault();

    if (fresh != null)
        CopyColumns(lockColumns, fresh, obj);
}

return count;
```

For providers with `IsAffectedRowsCountSupported == false`, raw `count == 0` cannot be interpreted as optimistic concurrency failure. It just means that the provider doesn't report the affected-row count. In that case the current implementation can refresh the entity from the database and still return `0`, while the XML docs say that `0` means concurrency failure. That gives callers a false failure signal and mutates the entity at the same time.

I don't think we should introduce a special `-1` result here. It would be awkward for callers and hard to explain. Instead, we can preserve the usual `1/0` API contract by using documented heuristic verification for providers without affected-row count support.

Suggested fallback semantics for `!IsAffectedRowsCountSupported`:

1. Execute the optimistic update attempt.
2. Re-read the database row by primary key.
3. Compare the persisted updated columns with the values we attempted to write.
4. If the row matches the expected persisted state:
   - refresh the optimistic-lock column(s) on the entity;
   - return `1`.
5. If the row doesn't match the expected persisted state:
   - leave the entity unchanged;
   - return `0`.

This makes the contract practical and predictable:

- `1` means the requested persisted state is present in the database after the update attempt.
- `0` means the requested persisted state is not present, so the update should be treated as a concurrency failure.
- On normal providers with reliable affected-row count, the existing strict semantics remain unchanged.
- On providers with `OUTPUT`/`RETURNING`, we still get the best behavior because the refreshed token comes from the same statement.

This heuristic is weaker than `OUTPUT`/`RETURNING` and less efficient than reliable affected-row count, but that is expected for providers that cannot report affected rows. It should be documented explicitly as degraded fallback behavior.

Important detail: for this heuristic, comparing only the refreshed token is not always enough. If the new token is known or computable by linq2db, comparing the token can be sufficient. But for database-generated tokens, the expected new token might be unknown without `OUTPUT`/`RETURNING`. In that case the fallback should verify the persisted updated columns instead of treating the re-read token alone as proof of success.

Also, this should be covered by tests for providers with `IsAffectedRowsCountSupported == false`: successful update should return `1` and refresh the token; stale-token update should return `0` and should not refresh the entity as if the update succeeded.

Alternative: if we don't want this heuristic, then `UpdateOptimisticWithRefresh` should be unsupported for provider/mapping combinations where neither `OUTPUT`/`RETURNING` nor reliable affected-row count is available.

But the current behavior should not be merged as-is: returning raw `0` from no-rowcount providers while also refreshing the entity contradicts the documented `0 == concurrency failure` contract.

MaceWindu:
@igor-tkachev fixed in 6ba666536. The `!IsAffectedRowsCountSupported` fallback no longer returns raw `0` on success: it now verifies the update by re-reading the row filtered by PK **and** the values written to the non-lock columns. A match means our write is the persisted state -> refresh the lock column(s) and return `1`; no match -> return `0` and leave the entity untouched. This restores the `1/0` contract you described without a `-1` sentinel.

Scope turned out to be ClickHouse-only: YDB sets `IsUpdateOutputSupported=true`, so it takes the OUTPUT/RETURNING path and returns an accurate count without needing affected-rows support. The heuristic is best-effort where the provider's UPDATE is asynchronous (ClickHouse mutations can be read before they apply) — documented on the `<returns>` docs. Covered by a new `UpdateOnNoRowcountProviderReportsResult` test (success -> `1` + refresh, stale -> `0` + unchanged), green on ClickHouse.

Unrelated aside re the new Update/Upsert API: Upsert doesn't help here — this needs an update-or-fail guarded on the version plus a version read-back over an arbitrary entity, whereas Upsert is insert-or-update matched on the key (a version mismatch would fall to the INSERT branch, not report a clean `0`). The builder-style Update API is a separate feature request as noted above.

igor-tkachev (commented):
The main no-rowcount issue looks addressed now. The verify-by-written-columns path matches the expected `1/0` contract much better than returning the raw provider count.

One remaining edge case: when `comparable.Length == 0`, the code falls back to the previous behavior: re-read by PK, copy the lock column to the entity, and return the raw provider count.

On no-rowcount providers that raw count is usually `0`, so this again means: entity was mutated while the method returned `0`. That is the same contract smell as before, just limited to entities where there are no deterministic non-lock columns to compare.

I think this branch should not perform best-effort refresh under the same `1/0` contract. If there are no comparable written columns, we cannot verify success. I would either return `0` without touching the entity, or better, make `UpdateOptimisticWithRefresh` unsupported for this provider/mapping combination because neither `OUTPUT`/`RETURNING`, nor affected-row count, nor written-column verification is available.

igor-tkachev (dismissed):
Thanks, this addresses my concern.

Throwing before `UPDATE` on providers that support neither `UPDATE OUTPUT` / `RETURNING` nor reliable affected-row count is a better contract than the previous best-effort fallback. It keeps the `1/0` semantics clean and avoids mutating the entity when the method cannot actually guarantee the optimistic-concurrency result.

Minor: some tests still use `SupportsRowcount()` as the condition, but for `UpdateOptimisticWithRefresh` the effective capability is now `IsUpdateOutputSupported || IsAffectedRowsCountSupported`. Not a blocker, but capability-based test conditions would describe the new contract more precisely.

igor-tkachev (dismissed):
Looks good to me.

The latest changes make sense:

- the fallback read-back now uses the original `source` instead of `dc.GetTable<T>()`, so `TableName(...)` / source-level table mapping is preserved;
- the renamed-table regression test covers that path;
- DuckDB `RETURNING` support is now reflected through `IsUpdateOutputSupported`;
- the unsupported no-output/no-rowcount contract remains strict and explicit.

No further concerns from my side.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +291,259 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+		// new T { lockCol = <source>.lockCol, ... } — only the optimistic-lock column(s), nothing else
+		private static MemberInitExpression InitLockColumns(Type objType, ColumnDescriptor[] lockColumns, Expression source)
+		{
+			var bindings = new MemberBinding[lockColumns.Length];
+
+			for (var i = 0; i < lockColumns.Length; i++)
+				bindings[i] = Expression.Bind(lockColumns[i].MemberInfo, Expression.MakeMemberAccess(source, lockColumns[i].MemberInfo));
+
+			return Expression.MemberInit(Expression.New(objType), bindings);
+		}
```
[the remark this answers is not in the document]
MaceWindu: Fixed in `8799739f6`. `CanInitLockColumns` now gates both projection helpers — `LockColumnsOutput` (OUTPUT path) and `LockColumnsSelector` (read-back path) — so `Expression.New(objType)` is unreachable for a type without a public parameterless constructor on either path; those entities project the whole row and materialize through the normal constructor mapping. `CopyColumns` still writes back only `lockColumns`, so non-lock columns on the caller's entity are untouched. Covered by `UpdateRefreshesEntityWithoutDefaultConstructor`.

One residual gap the report didn't cover, found while verifying this: the gate asks whether the *type* is constructible, not whether the *lock member* is writable. A get-only lock member made `MemberAccessor.SetValue` a silent no-op, so the call returned `1` with a stale token on the entity — reproduced on both paths, direct and remote. Now rejected up front with a `LinqToDBException` before the UPDATE runs, plus regression test `UpdateOnReadOnlyLockMemberThrows`.

Tests/Linq/Linq/ConcurrencyRefreshTests.cs
```diff
@@ -0,0 +1,365 @@
...
+using NUnit.Framework;
+
+using Shouldly;
+
+namespace Tests.Linq
+{
+	[TestFixture]
+	public class ConcurrencyRefreshTests : TestBase
+	{
+		// table name overriden for each test to workaround
```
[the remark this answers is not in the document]
MaceWindu: Fixed in `8799739f6` — both "overriden" → "overridden" and "to workaround" → "to work around" are applied at the top of this fixture. The original spelling still exists in `Tests/Linq/Linq/ConcurrencyTests.cs`, which this PR doesn't touch.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
```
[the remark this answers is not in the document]
MaceWindu: The directive is indeed unused — removed. No `System.Collections.Generic` type is named in the file: `ToList` resolves through `System.Linq.Enumerable`, `ToListAsync` on the `IAsyncEnumerable<T>` from `UpdateWithOutputAsync` through the `System.Linq.AsyncEnumerable` polyfill, and `FirstOrDefaultAsync` through `LinqToDB.Async`.

The build-failure rationale is wrong, though, and worth correcting for future reports: CS8019 is a *hidden* compiler diagnostic that `csc` never emits, and `.editorconfig` sets `dotnet_diagnostic.IDE0005.severity = none` repo-wide. `TreatWarningsAsErrors` therefore cannot promote it — which is why the Release leg of build 22896 was green on this commit rather than green by luck.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +290,351 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+		{
+			var expression = source.Expression;
+
+			while (expression is MethodCallExpression call
+				&& !typeof(ITable<T>).IsAssignableFrom(call.Type)
+				&& call.Arguments.Count > 0
+				&& typeof(IQueryable<T>).IsAssignableFrom(call.Arguments[0].Type))
+			{
+				expression = call.Arguments[0];
+			}
```
sdanyliv: **Minor · MIN001** — The peel classifies operators by *type shape* (`IQueryable<T>` in, `IQueryable<T>` out) rather than by what they do, so it also strips operators that carry translation intent rather than a predicate. `IgnoreFilters` is the load-bearing case: `LinqExtensions.IgnoreFilters<TSource>(this IQueryable<TSource>, params Type[])` returns `IQueryable<TSource>` and builds `Expression.Call(null, method, currentSource.Expression, …)`, so `call.Type` is not `ITable<T>` while `Arguments[0].Type` is the `ITable<T>` root — the loop peels straight past it.

The consequence lands on the `SELECT`-fallback path (no UPDATE `OUTPUT`/`RETURNING` but a reliable row count — MySQL, Oracle, DB2, Firebird < 5, Access, SQL CE, Informix, Sybase, SAP HANA). For an entity carrying a query filter, `db.GetTable<T>().IgnoreFilters().UpdateOptimisticWithRefresh(obj)` applies the UPDATE to the filtered-out row — `updatable` is built from the caller's `source`, which still carries `IgnoreFilters` — but the read-back is rebuilt without it, so the filter hides the row, `FirstOrDefault()` returns `null`, and the method returns `1` with `obj` still holding the stale token. The next optimistic update on that instance then affects 0 rows and reads as a spurious concurrency conflict.

The `<para>` caveat on the four public overloads covers the neighbouring case ("an entity query filter no longer accepts the updated values") but not this one, where the filter never accepted the row and the caller explicitly asked to ignore it. Of the other operators of the same shape that are stripped — `InlineParameters`, `DisableGuard`, `QueryHint`, `TagQuery` — none changes which rows are visible; `IgnoreFilters` is the exception.

Reachability is derived from the peel condition plus `IgnoreFilters`' expression shape; I did not run it end to end, which is why this is Minor rather than Major.

Fix: either remember an `IgnoreFilters` call seen while peeling and re-issue it (with the same `entityTypes` / `filterKeys` arguments) on the query returned by `Internals.CreateExpressionQueryInstance<T>`, or replace the type-shape test with an explicit classifier that drops only row-selecting operators (`Where` / `OrderBy` / `Take` / `Skip` / tag and hint calls) and stops on everything else. If neither is wanted, extend the `<para>` caveat to say the read-back re-enables entity query filters even when the caller disabled them. No inline suggestion — the fix changes the helper's shape rather than these lines.

— ✓ Fixed in 575c6d204, verified at HEAD 164f697d. The re-issue preserves the original argument list, so both `IgnoreFilters` overloads keep their `entityTypes` / `filterKeys`, and the OUTPUT/RETURNING path is untouched. One gap remains and is raised separately as MIN006: the regression test exercises only the reference-equal short-circuit, never the branch that rebuilds the call over a reduced inner source.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +290,351 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+				? InitLockColumns(objType, lockColumns, x)
+				: x;
+
+			return Expression.Lambda<Func<T, T>>(body, x);
+		}
+
+		// Table-shaping calls on the caller's source (TableName / SchemaName / ...) are kept so the read-back hits
+		// the same table the UPDATE targeted; the caller's query operators are dropped, because their predicates
+		// may test columns the UPDATE itself rewrites - the updated row would then no longer match and the
+		// refresh would silently not happen.
```
sdanyliv: **Minor · MIN002** — The comment states the guarantee unconditionally — "the caller's query operators are dropped" — but the loop below only unwinds while the chain stays `IQueryable<T>` → `IQueryable<T>`. The fourth condition, `typeof(IQueryable<T>).IsAssignableFrom(call.Arguments[0].Type)`, stops the peel at the first operator whose own source element type is not `T`: `SelectMany` / `Join` whose outer sequence is a different entity, or `OfType<TDerived>()` / `Cast<TDerived>()` over an `IQueryable<TBase>`. In those shapes the loop never advances, `ReferenceEquals(expression, source.Expression)` holds, and the caller's whole expression — predicate included — becomes the read-back source. If that predicate tests a column the UPDATE rewrote, the read-back matches nothing and the entity is left unrefreshed while a non-zero count is returned: exactly the failure this comment says it prevents.

The observable outcome is at least partly covered by the `<para>` caveat on the public overloads; the comment's absolute phrasing is what invites a future reader to rely on a guarantee the code does not give.

Fix: narrow the comment to what the code actually guarantees. The alternative, larger fix is to make the peel classify operators explicitly instead of by type shape (see MIN001), which would let the comment keep its stronger claim.

```suggestion
		// Table-shaping calls on the caller's source (TableName / SchemaName / ...) are kept so the read-back hits
		// the same table the UPDATE targeted. Query operators are dropped only while the chain stays
		// IQueryable<T> -> IQueryable<T>, because their predicates may test columns the UPDATE itself rewrites -
		// the updated row would then no longer match and the refresh would silently not happen. A predicate sitting
		// under an operator whose own source element type differs (SelectMany / Join with a different outer,
		// OfType / Cast) is not dropped and can still hide the updated row from the read-back.
```

— ✓ Fixed in 575c6d204, verified at HEAD 164f697d. Rejection happens at `ConcurrencyExtensions.cs:391-392`, reached from `:444` / `:492` — both above `updatable.Update()` at `:446` / `:494` — so an unsupported source is refused without having written anything. Table-shaping sources and a plain `GetTable<T>()` are unaffected via the `ITable<T>` short-circuit at `:385`.

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDataProvider.cs
```diff
@@ -68,6 +68,7 @@ protected SqlServerDataProvider(string name, SqlServerVersion version, SqlServer
 			SqlProviderFlags.OutputInsertUseSpecialTable        = true;
 			SqlProviderFlags.OutputUpdateUseSpecialTables       = true;
 			SqlProviderFlags.OutputMergeUseSpecialTables        = true;
+			SqlProviderFlags.IsUpdateOutputRowsSupported        = true;
```
sdanyliv, line 71: **Minor · MIN003** — On SQL Server the capability this flag models is a property of the *target table*, not of the provider, so a per-provider constant cannot express it. From Microsoft's OUTPUT-clause reference: "If the `OUTPUT` clause is specified without also specifying the `INTO` keyword, the target of the DML operation can't have any enabled trigger defined on it for the given DML action. For example, if the `OUTPUT` clause is defined in an `UPDATE` statement, the target table can't have any enabled `UPDATE` triggers." (<https://learn.microsoft.com/en-us/sql/t-sql/queries/output-clause-transact-sql>)

Because `UpdateOptimisticWithRefreshCore` checks `IsUpdateOutputRowsSupported` first and never falls back, every `UpdateOptimisticWithRefresh` call against a SQL Server table carrying any enabled UPDATE trigger — an audit trigger, a history trigger, anything unrelated to versioning — fails with `Msg 334`, even though the second path in the same method (reliable affected-rows count plus `SELECT`-by-PK, which SQL Server fully supports) would produce the correct answer. The caller has no way to select it.

The new `ConcurrencyRefreshTests.OutputSupportSurface` probe cannot observe this: it creates a fresh table with no triggers, so the restriction is structurally outside what it exercises — which is why the green CI run says nothing about it. The limitation *is* documented on all four overloads, so this is a deliberate trade-off rather than an oversight; the question is whether a hard failure is the right disposition when a working path exists in the same method.

Fix: two options, both leaving the flag itself alone — it is the correct provider-level answer. (1) Let the caller opt out: an overload parameter or a `DataOptions` switch that forces the two-statement affected-rows path, so a trigger-bearing table stays usable. (2) Fall back automatically by catching the `Msg 334` condition on the OUTPUT attempt and re-running through the affected-rows path — cheaper for users, but it couples `ConcurrencyExtensions` to a provider-specific error number, so (1) is likely preferable. If neither is wanted, consider promoting the trigger caveat out of the `<para>` prose into an explicit `<exception>` remark so it is visible at the call site, since the resulting failure is a raw provider error with no linq2db context.

See also MIN005, which asks for a regression anchor on this same caveat — if a fallback is added here, that finding is moot.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -116,7 +109,10 @@ private static IUpdatable<T> MakeUpdateOptimistic<T>(IQueryable<T> query, IDataC
 		/// <typeparam name="T">Entity type.</typeparam>
 		/// <param name="dc">Database context.</param>
 		/// <param name="obj">Entity instance to update.</param>
-		/// <returns>Number of updated records.</returns>
+		/// <returns>
+		/// Number of updated records. On providers that do not report affected rows (e.g. ClickHouse) the count is
+		/// always <c>0</c>, so it cannot be used to detect an optimistic-concurrency failure on those providers.
+		/// </returns>
```
sdanyliv, line 116: **Minor · MIN004** — This PR rewrites the `<returns>` block on all eight shipped `UpdateOptimistic` / `DeleteOptimistic` overloads to say that on providers which do not report affected rows the count is "always `0`". That asserts the outcome for the whole class of providers, but the PR's own new `UpdateOptimisticWithRefresh` docs deliberately split the two cases: "unreliable on every provider that does not report affected rows - including YDB, and always `0` on ClickHouse".

YDB is exactly what separates them. This PR sets `SqlProviderFlags.IsAffectedRowsCountSupported = false` for it in `Source/LinqToDB/Internal/DataProvider/Ydb/YdbDataProvider.cs`, so YDB is inside the class the rewritten sentence covers — yet the new docs pointedly decline to claim its count is `0`. Both wordings ship in the same file from the same PR, so one of them is wrong for YDB, and as written the older overloads carry the more absolute of the pair.

Fix: adopt the phrasing already used on the new methods — unreliability for the class, `always 0` only for ClickHouse. The identical edit applies at lines 112-115, 133-136, 153-156 and 176-179 (the `UpdateOptimistic` family) and at 198-201, 219-222, 239-242 and 260-263 (the `DeleteOptimistic` family, with "updated" swapped for "deleted").

```suggestion
		/// <returns>
		/// Number of updated records. On providers that do not report affected rows the count is unreliable - always
		/// <c>0</c> on ClickHouse - so it cannot be used to detect an optimistic-concurrency failure there.
		/// </returns>
```

— ✓ Fixed in 575c6d204, verified at HEAD 164f697d. All eight blocks carry the corrected phrasing: `:113`, `:134`, `:154`, `:177` ("updated") and `:199`, `:220`, `:240`, `:261` ("deleted").

Tests/Linq/Linq/ConcurrencyRefreshTests.cs
```diff
@@ -0,0 +1,663 @@
...
+			var failed = db.UpdateOptimisticWithRefresh(stale);
+
+			failed.ShouldBe(0);
+			stale.Stamp.ShouldBe(beforeGuid);
+			stale.Revision.ShouldBe(5);
+		}
+
+		[Test]
+		public void UpdateRefreshesDatabaseGenerated([IncludeDataSources(true, TestProvName.AllSqlServer)] string context)
+		{
```
sdanyliv, line 534: **Minor · MIN005** — The new XML docs state that on SQL Server the OUTPUT path cannot be used against a table carrying any enabled UPDATE trigger, and that this rules out the database-trigger variant of `VersionBehavior.Auto`. The claim is correct, but nothing in the fixture exercises it.

`SqlServerDataProvider` sets `IsUpdateOutputRowsSupported = true` unconditionally and the core has no fallback, so on SQL Server `UpdateOptimisticWithRefresh` always takes the OUTPUT path and fails with a raw `Msg 334` on any table with an enabled UPDATE trigger — where the sibling `UpdateOptimistic` succeeds. Audit triggers are common on SQL Server tables, and `VersionBehavior.Auto` explicitly advertises the trigger variant (`Source/LinqToDB/Mapping/VersionBehavior.cs:14`, "E.g. using SQL Server rowversion/timestamp column or database trigger"). This test is the natural place to pin it: it already covers the *other* half of `VersionBehavior.Auto` (rowversion) on SQL Server, and the trigger half is uncovered.

Fix: add a SQL-Server-only companion next to `UpdateRefreshesDatabaseGenerated` that creates an `AFTER UPDATE` trigger on the local table (`db.Execute("CREATE TRIGGER ... ON ConcurrencyRefreshRowVersion AFTER UPDATE AS SET NOCOUNT ON")`) and asserts the current failure mode, so the documented caveat has a regression anchor and any future change — routing through `OUTPUT ... INTO`, or falling back to the read-back path — shows up as a test change rather than silently. If pinning the raw provider error is judged too brittle, the alternative is to catch it and re-wrap it in a `LinqToDBException` naming the trigger limitation, which is then testable by message.

Paired with MIN003: if a fallback is added there, this finding goes away.

— ✓ Fixed in 575c6d204, verified at HEAD 164f697d as `UpdateOnTriggeredTableThrowsOnSqlServer` (`ConcurrencyRefreshTests.cs:526-546`). Reviewed statically as never-executed code — provider gating, the single-statement `CREATE TRIGGER` batch, teardown ordering and the `SET NOCOUNT ON` interaction all check out — but it has not yet run: the SQL Server matrix legs are still the gate.

Tests/Linq/Linq/ConcurrencyRefreshTests.cs
```diff
@@ -0,0 +1,663 @@
...
+
+			try
+			{
+				return t.Where(r => r.Id == 1).Set(r => r.Stamp, 2).Update() == 1;
+			}
+			catch
+			{
+				return false;
+			}
+		}
```
sdanyliv, line 814: **Suggestion · SUG001** — `ProbeUpdateOutput` probes both directions on purpose: a matching UPDATE must return exactly one output row, *and* a non-matching one must return none, because the second half is what makes the returned row count usable as a concurrency verdict. `ProbeAffectedRows` only probes the first half — it asserts that a matching UPDATE reports `1` and never checks that a non-matching UPDATE reports `0`.

The zero direction is precisely what `UpdateOptimisticWithRefreshCore` relies on when `IsAffectedRowsCountSupported` is `true`: it treats `count == 0` as "the optimistic filter matched nothing". A provider that reported `1` for a matched row and anything non-zero for an unmatched one would satisfy this guard while silently violating the contract the flag asserts. `UpdateConcurrencyFailureLeavesEntityUnchanged` does exercise the zero direction today, but as a behaviour test on the public API rather than a flag guard — and it takes the OUTPUT path, never touching the count, on every provider where `IsUpdateOutputRowsSupported` is `true`.

Fix: mirror `ProbeUpdateOutput`'s two-step shape — keep the matched-row check, then run a second UPDATE whose predicate matches nothing and require the reported count to be `0`. Both statements stay inside the existing `try`; setup stays outside it as it does now. Worth a full test run across the matrix before merging, since the whole point of the change is that it can go red on a provider that currently passes.

```suggestion
		private static bool ProbeAffectedRows(IDataContext db)
		{
			// setup stays outside the try - see ProbeUpdateOutput
			using var t = db.CreateLocalTable<RefreshTable<int>>("ConcurrencyRowcountProbe");
			db.Insert(new RefreshTable<int> { Id = 1, Stamp = 1, Value = "x" }, tableName: "ConcurrencyRowcountProbe");

			try
			{
				if (t.Where(r => r.Id == 1).Set(r => r.Stamp, 2).Update() != 1)
					return false;

				// the count must also distinguish "no row matched" - that is the direction
				// UpdateOptimisticWithRefresh relies on to report a concurrency failure
				return t.Where(r => r.Id == -1).Set(r => r.Stamp, 3).Update() == 0;
			}
			catch
			{
				return false;
			}
		}
```

— ✓ Fixed in 575c6d204, verified at HEAD 164f697d. `ProbeAffectedRows` (`ConcurrencyRefreshTests.cs:753-772`) now mirrors `ProbeUpdateOutput`'s two-step shape, with the negative probe on a predicate that matches nothing. As flagged, this widens a guard test matrix-wide and has not yet been through `test-all`.

sdanyliv (commented):
> [!IMPORTANT]

## Review notes

- [ ] `Tests/Base/ProviderNameHelpers.SupportsRowcount` now derives from `SqlProviderFlags.IsAffectedRowsCountSupported` instead of a hard-coded provider list. Answer parity holds today (the flag is `false` in exactly the two places the old list excluded), but the consequence is that flipping that flag for any provider silently re-gates ~120 rowcount assertions across 15 files under `Tests/Linq`. Worth knowing before the flag is next changed.

## Prior-review audit

- **Fixed** · Copilot — `InitLockColumns` builds `new T { … }` via `Expression.New(objType)`, which requires a parameterless constructor. `CanInitLockColumns` (`ConcurrencyExtensions.cs:338`) now gates both `LockColumnsOutput` and `LockColumnsSelector`, so constructor-mapped entities project the whole row and materialize through the normal constructor mapping.
- **Fixed** · Copilot — "overriden" / "to workaround" typos in the new fixture; both corrected at `ConcurrencyRefreshTests.cs:18`.
- **Fixed** · Copilot — unused `using System.Collections.Generic;`, removed. Its stated rationale was **inaccurate**, and the author already corrected it on the thread: CS8019 is a hidden diagnostic `csc` never emits, and `IDE0005` is set to `none` repo-wide, so `TreatWarningsAsErrors` could not have promoted it to a build failure.
- **Inaccurate at HEAD** · Copilot (suppressed review-body comment) — that `ReadBackSource` "only peels Queryable operators when the inner expression is assignable to `IQueryable<T>`", stopping at element-type-changing operators. The loop no longer tests `call.Method.DeclaringType == typeof(Queryable)`; it stops on `ITable<T>` instead, so the description does not match the current code. The residual gap it gestures at is real, though, and is carried as MIN002 — along with a second, opposite gap in the same helper as MIN001.
- **Fixed** · igor-tkachev — the no-rowcount fallback could refresh the entity and still return `0`, contradicting the documented `0 == concurrency failure` contract. HEAD throws before executing any UPDATE (`ConcurrencyExtensions.cs:418` / `:463`).
- **Fixed** · igor-tkachev — the `comparable.Length == 0` branch kept the same best-effort contract. The written-column verify path was removed entirely; the same capability throw now covers that case.
- **Fixed** · igor-tkachev — tests gated on `SupportsRowcount()` where the effective capability is `IsUpdateOutputRowsSupported || IsAffectedRowsCountSupported`. Six sites now gate on the flags directly. The two remaining `SupportsRowcount()` uses (`ConcurrencyRefreshTests.cs:493`, `:519`) are correct as-is: those tests cover the no-lock-column path, which short-circuits to a plain update and returns the raw provider count, so the affected-rows flag alone *is* the right gate there.

## Out-of-scope observations

- **`UpdateWithOutput`'s "Database support" list omits DuckDB and YDB.** All 40 `Database support` blocks across the `UpdateWithOutput` / `UpdateWithOutputAsync` overloads in `Source/LinqToDB/LinqExtensions/LinqExtensions.Update.cs` name only SQL Server 2005+ and Firebird 2.5+, some adding SQLite 3.35+ and PostgreSQL. Neither DuckDB nor YDB appears, yet this PR declares both `IsUpdateOutputRowsSupported = true` and lands 20 green DuckDB `UpdateWithOutputTests` baselines. Provider support predates this PR — only the coverage and the flag are new — but those `<remarks>` are what a user reads to decide whether the API is available on their provider.
- **DB2 has the underlying capability but no builder support.** `SELECT … FROM FINAL TABLE (UPDATE …)` yields exactly the rows an UPDATE touched, and linq2db already emits the sibling form — `SELECT <identity> FROM NEW TABLE (INSERT …)` at `Source/LinqToDB/Internal/DataProvider/DB2/DB2SqlBuilderBase.cs:155`. Leaving `IsUpdateOutputRowsSupported = false` is correct today, because there is no `UpdateWithOutput` rendering for DB2 and flipping the flag would turn the new guard test red. Adding that rendering would move DB2 from the two-statement path onto the single-statement one.
- **"Affected-row count is unreliable" now has two sources of truth.** `ClickHouseDataProvider.cs:47-48` already carried the fact as prose — "we enable InsertOrUpdate deliberately here and then throw exception from SqlBuilder / as emulation doesn't work properly due to missing rowcount functionality" — and handles it with a ClickHouse-specific throw from the SQL builder. The new `IsAffectedRowsCountSupported` states the same fact on the next line, but nothing outside `ConcurrencyExtensions` and the test helper reads it, so the `InsertOrUpdate` emulation stays gated on a hard-coded provider decision. No provider is inconsistent today; routing that special case through the flag would remove the duplication.
- **`ConcurrencyExtensions` cannot handle a nested (complex) mapped column.** Pre-existing on `master`, not introduced here: `MakeUpdateOptimistic` and `FilterByColumns` build member access with `Expression.MakeMemberAccess(param, cd.MemberInfo)`, and for a column mapped through a nested path (`MemberName = "Residence.Street"`) `ColumnDescriptor.MemberInfo` is the *leaf* member declared on the nested type, so resolving it against the entity root throws. The canonical helper for this is `GetMemberAccessExpression` (`Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs:37`), which walks the dotted path and also handles dynamic columns. The PR's new `InitLockColumns` follows the existing pattern rather than adding a new failure — and `Expression.Bind` could not express a nested path in a member-init anyway.

## Baselines

Delta: [linq2db.baselines#2067](https://github.com/linq2db/linq2db.baselines/pull/2067) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5643)

**221 added · 0 modified · 0 deleted** across 68 providers — 9,971 added lines, none removed. Zero modified and zero deleted is exactly the right shape: both new flags are capability declarations read only by `ConcurrencyExtensions`, so no pre-existing test's SQL should move, and none did.

The 221 reconcile exactly against the fixture. It has 20 `[Test]` methods, 17 of which carry `DisableBaseline`, leaving 3 that capture: 68 × `UpdateWithoutLockColumnBehavesLikePlainUpdate{,Async}` + 65 × `UpdateRefreshesAutoIncrement` = 201, plus 20 DuckDB `UpdateWithOutputTests` baselines newly enabled by the feature-constant change.

### Capability matrix, read off the baseline files

- **SQL Server 2005 → 2025** — `UPDATE … OUTPUT INSERTED.[Stamp]`, no follow-up SELECT.
- **PostgreSQL 9.2 → 17** — `RETURNING "ConcurrencyRefreshAutoInc"."Stamp"`; **18 / 19** — `RETURNING NEW."Stamp"`, the v18 `OLD`/`NEW` form.
- **SQLite ×4, DuckDB, YDB** — `RETURNING`.
- **Firebird 2.5 / 3 / 4** — plain `UPDATE`, then a separate `SELECT "obj"."Stamp" … FETCH NEXT 1 ROWS ONLY`; **Firebird 5** — `UPDATE … RETURNING NEW."Stamp"`. The `Version >= FirebirdVersion.v5` gate lands precisely on the version boundary.
- **Oracle ×6, MySQL family ×5, DB2, Informix, SAP HANA, SQL CE, Sybase, Access ×4** — fallback `UPDATE` followed by `SELECT`-by-PK.
- **ClickHouse ×3** — no `UpdateRefreshesAutoIncrement` baseline at all (65 of 68 providers have one), while all three do appear in the no-lock-column tests (68 of 68). The absence is path-specific — the capability throw — not a provider-wide skip.

No cross-provider anomalies.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
MaceWindu: Confirmed and fixed in 575c6d204.

Reproduced red on MySql.8.0 first, direct and via LinqService: the UPDATE ran with `IgnoreFilters` applied (`Records Affected: 1`), the read-back was rebuilt without it and emitted `WHERE e.Id <> 1 AND e.Id = @Id` (`Rows Count: 0`), so the call returned 1 with the entity still holding `00000000-0000-0000-0000-000000000000`.

Took the first of your two options: `ReadBackExpression` records the `IgnoreFilters` call while peeling and re-issues it with its original arguments on the rebuilt query. Regression test `UpdateViaQueryWithIgnoreFilters`.

One note on severity - this is a silent stale token on a reported success, reachable from plain public API, so it ranked above Minor. The unverified-reachability hedge is what held it down, and running it is what settled it.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
MaceWindu: Confirmed, but the `suggestion` was deliberately not applied: the finding is understated, and a comment change would have documented a wrong answer.

Probed the shape rather than reasoning about it. `outer.SelectMany(o => t.Where(g => g.Value == "initial"))` on MySql.8.0 gives `cnt=1 refreshed=False` - the UPDATE lands (`Records Affected: 1`) and the read-back keeps the stale predicate (`WHERE g_1.Value = 'initial' AND g_1.Id = @Id`, `Rows Count: 0`). So this is a wrong-result bug of the same class as MIN001, not an over-strong comment.

Fixed in 575c6d204 by rejecting sources the peel cannot reduce to the updated table, with the read-back query built before the UPDATE so an unsupported source is rejected without having modified anything. Regression test `UpdateThroughUnsupportedSourceShapeThrows`.

Rejecting rather than classifying operators, because the target table sits inside the `SelectMany` lambda: an allow-list of row-selecting operators still would not find it, and falling back to `dc.GetTable<T>()` would turn a missing refresh into a wrong-table read for `TableName(...)` sources.

Source/LinqToDB/Internal/DataProvider/SqlServer/SqlServerDataProvider.cs
MaceWindu: Accurate on every point, and declined for 6.5.0 - resolving as won't-fix, not fixed.

The limitation stands as documented on all four overloads. Option 1 (caller opt-out) is the right eventual answer but adds public API surface to a PR already introducing four overloads, and the flag is correct either way. Option 2 is out: coupling `ConcurrencyExtensions` to error 334 is not worth it. No follow-up issue filed, by maintainer decision.

MIN005 was taken instead, so the caveat now has a regression anchor.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
MaceWindu: Fixed in 575c6d204, all eight blocks.

One correction: "one of them is wrong for YDB" is too strong. If YDB does return 0, the older wording is merely stronger than needed, not false. Nothing in the repo pins YDB's actual value - the tests only skip the assertion via `SupportsRowcount` - so I adopted your phrasing, which is accurate under either behaviour.

Tests/Linq/Linq/ConcurrencyRefreshTests.cs
MaceWindu: Added in 575c6d204 as `UpdateOnTriggeredTableThrowsOnSqlServer`.

Asserts only that it throws, not error 334: the message is localized and the exception type differs between the `System.Data.SqlClient` and `Microsoft.Data.SqlClient` contexts. It also asserts the sibling `UpdateOptimistic` still returns 1 against the same trigger-bearing table, which is the contrast that makes the limitation concrete.

Not verified locally - it cannot run on MySQL and no SQL Server container was up, so CI is the gate here. Also worth flagging that this is a new pattern for the suite: no other test under `Tests/` creates a trigger.

Tests/Linq/Linq/ConcurrencyRefreshTests.cs
MaceWindu: Applied as suggested in 575c6d204.

Flagging that this is a matrix-wide change to a guard test whose whole point is that it can go red on a provider that currently passes, so `test-all` is the real gate rather than the local MySql.8.0 run.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +290,391 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+			if (call.Method.DeclaringType == typeof(LinqExtensions) && string.Equals(call.Method.Name, nameof(LinqExtensions.IgnoreFilters), StringComparison.Ordinal))
+			{
+				if (ReferenceEquals(inner, call.Arguments[0]))
+					return call;
+
+				var arguments = call.Arguments.ToArray();
+				arguments[0]  = inner;
+
+				return Expression.Call(call.Method, arguments);
+			}
```
sdanyliv, line 403: **Minor · MIN006** — The MIN001 fix has two halves and only one of them is reachable from a test.

`UpdateViaQueryWithIgnoreFilters` (`ConcurrencyRefreshTests.cs:275-308`) builds its source as `t.IgnoreFilters()`, with nothing between the operator and the table. `ReadBackExpression` on `call.Arguments[0]` therefore hits the `ITable<T>` short-circuit at line 385 and returns the *same reference*, so `ReferenceEquals(inner, call.Arguments[0])` holds at line 400 and the method exits at line 401 through `return call`. `ReadBackSource` then short-circuits in turn on `ReferenceEquals(expression, source.Expression)` at line 377 and hands back the original source untouched.

Lines 403-406 — the re-issue that actually reconstructs the `IgnoreFilters` call over a *reduced* inner source — are executed by no test in the fixture. That branch is precisely the case the finding was about: a caller who both disables a query filter and applies a predicate, e.g. `t.Where(r => r.Value == "initial").IgnoreFilters()` or `t.IgnoreFilters().Where(r => r.Value == "initial")`, where the `Where` must be dropped because it tests a column the UPDATE rewrites while the filter-disable must survive. That shape is ordinary user code, reachable on every `SELECT`-fallback provider, and a mistake there — dropping the operator, mis-ordering it, losing an `entityTypes` / `filterKeys` argument — fails silently: a non-zero count with a stale token on the entity, which is the bug MIN001 reported.

Fix: add a variant of `UpdateViaQueryWithIgnoreFilters` whose source puts a row-selecting operator between the table and `IgnoreFilters` — `t.Where(r => r.Value == "initial").IgnoreFilters()` — keeping the same `HasQueryFilter(e => e.Id != 1)` mapping, the same `SELECT`-fallback gate, and the same `record.Stamp.ShouldNotBe(before)` / `ShouldBe(t.IgnoreFilters().Single().Stamp)` assertions. That combination is red both when the operator is dropped (the filter hides the row) and when the `Where` is wrongly kept (the predicate no longer matches the updated row). A second variant passing `IgnoreFilters(typeof(RefreshTable<Guid>))` would additionally pin that the original arguments are carried through.

No inline suggestion — this is a new test, not an edit to these lines.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +290,391 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+			// the operator doesn't expose the updated table as its own source (SelectMany / Join with a different
+			// outer, OfType / Cast over a base type), so the caller's filters cannot be dropped from the read-back:
+			// a predicate over a column the UPDATE rewrites would hide the updated row and leave the entity stale
+			if (call.Arguments.Count == 0 || !typeof(IQueryable<T>).IsAssignableFrom(call.Arguments[0].Type))
+				throw new LinqToDBException($"UpdateOptimisticWithRefresh cannot refresh through '{call.Method.Name}': the query operator does not expose the updated table as its source, so the regenerated value cannot be read back reliably.");
+
+			var inner = ReadBackExpression<T>(call.Arguments[0]);
+
+			// IgnoreFilters is the one dropped operator that changes which rows are visible: without it the
+			// read-back re-enables entity query filters the caller disabled, hiding the row the UPDATE reached
```
sdanyliv: **Nit · NIT001** — The comment calls `IgnoreFilters` "the one dropped operator", but the branch it introduces is the branch that *keeps* it — the whole point of the block below is that this operator is the exception to the drop rule. As written the first clause states the opposite of what the code does, and a reader has to reach line 406 to resolve it.

The block comment at lines 368-371 has the matching gap in the other direction: it still says flatly "the caller's query operators are dropped", with no mention that `IgnoreFilters` is re-issued. Appending "(except `IgnoreFilters`, which is re-issued - see below)" there would keep the two comments consistent.

```suggestion
			// IgnoreFilters is the one operator that must not be dropped: it changes which rows are visible, and
			// without it the read-back re-enables entity query filters the caller disabled, hiding the row the UPDATE reached
```

sdanyliv (commented):
> [!IMPORTANT]

## Verification update — 2026-08-27, against HEAD `164f697d`

Re-verification of the six findings posted on 2026-08-26 against `877706eb`. The PR's own commits since then are `575c6d204` (the fixes) and `be7b56c0b` (the `UpdateWithOutput` doc additions); `164f697d` is a `master` merge.

### Fixed

- [x] **MIN001** — `IgnoreFilters` dropped from the read-back. `ReadBackExpression` (`ConcurrencyExtensions.cs:382-410`) is now a recursion that re-issues the call with its original argument list (`arguments[0] = inner`), so both shipped `IgnoreFilters` overloads keep their `entityTypes` / `filterKeys` arguments, and multiple calls in one chain keep their relative order. The OUTPUT/RETURNING path is untouched — it returns at `:435` / `:483`, before `ReadBackSource` is reached.
- [x] **MIN002** — comment asserted a guarantee the peel did not give. Taken further than the finding asked: sources the peel cannot reduce to the updated table now throw (`:391-392`), and the read-back query is built at `:444` / `:492`, above `updatable.Update()` at `:446` / `:494`, so an unsupported shape is rejected without having written anything. The `ITable<T>` short-circuit at `:385` leaves `TableName` / `SchemaName` / `DatabaseName` / `ServerName` sources whole, and a plain `GetTable<T>()` returns unchanged. New `<exception>` block on all four overloads; the block comment at `:368-371` now ends "A source the peel cannot reduce to that table is rejected outright."
- [x] **MIN004** — `<returns>` wording. All eight shipped overloads carry the corrected phrasing at HEAD: `:113`, `:134`, `:154`, `:177` ("Number of updated records"), `:199`, `:220`, `:240`, `:261` ("Number of deleted records"). Eight of eight, and the delete family says "deleted".
- [x] **MIN005** — no regression anchor for the SQL Server trigger caveat. `UpdateOnTriggeredTableThrowsOnSqlServer` at `ConcurrencyRefreshTests.cs:526-546`. Reviewed as never-executed code: `[IncludeDataSources(false, TestProvName.AllSqlServer)]` excludes the remote context so `db.Execute` issues DDL directly; the `CREATE TRIGGER` is the only statement in its command, satisfying SQL Server's first-statement-in-batch rule; the `using` declaration order disposes the table before the connection, so `DROP TABLE` takes the trigger with it. The sibling `UpdateOptimistic(record).ShouldBe(1)` assertion is consistent — a DML-free trigger body's `SET NOCOUNT ON` suppresses the trigger's own counts, not the triggering statement's.
- [x] **SUG001** — `ProbeAffectedRows` probed only one direction. `ConcurrencyRefreshTests.cs:753-772` now mirrors `ProbeUpdateOutput`: setup outside the `try`, the matched-row check retained, and a negative probe on `Id == -1` against a table holding only `Id == 1`. `catch { return false; }` unchanged, and `AffectedRowsCountSurface` consumes the result as before.

### Still actual

- [ ] **MIN003** — SQL Server's OUTPUT-without-INTO restriction is a property of the target table, not the provider. Verified unchanged at HEAD: `UpdateOptimisticWithRefreshCore` still tests `IsUpdateOutputRowsSupported` first (`:426` / `:474`) and returns from that branch with no fallback, and `SqlServerDataProvider.cs:71` still sets the flag unconditionally. The caveat text has not drifted — it is present verbatim on all four overloads (`:517-519`, `:558-560`, `:600-602`, `:643-645`).

  Recorded as **declined for 6.5.0 by maintainer decision**, not as an unaddressed finding. Left as-is and not re-argued; MIN005's new test is the agreed regression anchor.

### New findings

Both are posted as inline comments on `Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs`:

- **MIN006** (Minor) — the MIN001 fix has two halves and only one is covered by a test.
- **NIT001** (Nit) — the comment at `:396-397` describes `IgnoreFilters` as dropped, in the branch that keeps it.

## Review notes

- [ ] **`test-all` has not run on this HEAD.** The most recent full matrix run was on 2026-08-25 against `12c96472a`. Everything after it is unexercised: the `ConcurrencyExtensions` changes in `877706eb` *and* `575c6d204`, six new tests — `UpdateViaQueryWithFilterUnderLinqToDBOperator`, `UpdateViaQueryWithIgnoreFilters`, `UpdateThroughUnsupportedSourceShapeThrows`, `UpdateOnReadOnlyLockMemberThrows`, `UpdateRefreshesMultipleLockColumns` (all `[DataSources]`, so matrix-wide) and `UpdateOnTriggeredTableThrowsOnSqlServer` — and the widened `ProbeAffectedRows` guard, which now requires every provider declaring `IsAffectedRowsCountSupported = true` to report exactly `0` for a non-matching UPDATE or `AffectedRowsCountSurface` fails. `UpdateOnTriggeredTableThrowsOnSqlServer` has no local coverage at all. Worth an `/azp run test-all` before merge.

  (For the record: the `SupportsRowcount` rewrite and the DuckDB `UpdateWithOutputTests` enablement are **not** in this set — both landed in `666d532266` on 2026-08-18 and were covered by the 08/25 matrix.)

## Prior-review audit

The Copilot threads and igor-tkachev's four reviews were audited in the previous review and each verdict still holds at HEAD — nothing in `575c6d204` or `be7b56c0b` reopens any of them, and all are resolved or explicitly withdrawn. Not re-listed here.

## Out-of-scope observations

- **The source-shape guard makes the new API's throw contract provider-dependent.** `ReadBackSource` — and therefore the new `LinqToDBException` for `SelectMany` / `Join` / `OfType` / `Cast` sources — is only reached on the `SELECT`-fallback path, because the OUTPUT/RETURNING branch returns at `:435` / `:483` first. The same caller code therefore throws on MySQL, Oracle, DB2, Access, SQL CE, Informix, Sybase and SAP HANA while succeeding on SQL Server, PostgreSQL, SQLite, DuckDB, YDB and Firebird 5+. That is defensible — on the OUTPUT path there is no read-back to get wrong, so rejecting there would refuse shapes that genuinely work — but it is a portability trap in a brand-new API, and the `<exception>` text describes the condition as "on the read-back path", which a caller cannot map to a provider without also reading the `<summary>`. Worth a conscious decision either way: state in the `<exception>` block that this applies only where OUTPUT / RETURNING is unavailable, or apply the guard unconditionally so the contract is provider-independent.

## Baselines

Not reviewed this round. The `baselines/pr_5643` branch was last written 2026-08-26 03:25 UTC by the 08/25 matrix run at `12c96472a`, ~14 hours before `575c6d204`, so it describes SQL the current code no longer emits and carries nothing for the six new tests. The previous review's baselines analysis still covers what that branch holds. Re-run the matrix on this HEAD and the branch will regenerate.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
MaceWindu: Confirmed and fixed in b485d05ab.

Verified the mechanism rather than taking it on reading: neutering the re-issue at `ConcurrencyExtensions.cs:403-406` to `return inner` turns the new `UpdateViaQueryWithIgnoreFiltersAboveWhere` red on both Oracle 11 contexts while `UpdateViaQueryWithIgnoreFilters` stays green. The existing test staying green with the branch destroyed is the direct evidence it never reached it, as you described.

Failure shape on the neutered build is the stale-token one this API has to guard against: `cnt.ShouldBe(1)` passes, then `record.Stamp` is still `00000000-0000-0000-0000-000000000000`.

Skipped the second `IgnoreFilters(typeof(RefreshTable<Guid>))` variant.

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
MaceWindu: Applied in b485d05ab, including the `:368-371` amendment you flagged - the block comment now reads "the caller's query operators are dropped (except `IgnoreFilters`, which is re-issued - see below)".

Source/LinqToDB/Concurrency/ConcurrencyExtensions.cs
```diff
@@ -265,5 +290,392 @@ public static IQueryable<T> WhereKeyOptimistic<T>(this IQueryable<T> source, T o
...
+		private static ColumnDescriptor[] GetOptimisticLockColumns(EntityDescriptor ed, Type objType)
+		{
+			return ed.Columns
+				.Where(c => ed.MappingSchema.GetAttribute<OptimisticLockPropertyBaseAttribute>(objType, c.MemberInfo) != null)
+				.ToArray();
+		}
+
+		private const string UpdateWithRefreshNotSupportedMessage =
+			"UpdateOptimisticWithRefresh requires the provider to support UPDATE OUTPUT / RETURNING " +
+			"(SqlProviderFlags.IsUpdateOutputRowsSupported) or a reliable affected-rows count " +
```
sdanyliv: Why we mention internals in public message?
Also we have ErrorHelper class to hold error messages.

MaceWindu: Both fixed in `93a99bbfd`.

`SqlProviderFlags` is `LinqToDB.Internal.SqlProvider`, so the flag names told a caller nothing they could act on — the message now describes the capability instead: "requires the provider to either return updated rows from the UPDATE statement (OUTPUT / RETURNING) or report the number of affected rows".

All three of the API's messages moved to `ErrorHelper`, not just this one:

- `Error_Concurrency_UpdateWithRefresh_NotSupported`
- `Error_Concurrency_UpdateWithRefresh_ReadOnlyLockMember` (`{0}.{1}`)
- `Error_Concurrency_UpdateWithRefresh_UnsupportedSource` (`{0}`)

Also moved the repeated `"Linq To DB query expected"` literal to `Error_LinqToDBQueryExpected` and applied it to all five sites in the file, so the same text isn't both a constant and a literal. The tests now assert against the constants rather than substrings.

Verified: `ConcurrencyRefreshTests` 44/44 on MySQL (SELECT-fallback path, 40 passed / 4 skipped) and on SQLite.MS (RETURNING path, 28 passed / 16 skipped); Release net10.0 build clean.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- `UpdateOptimisticWithRefresh` / `UpdateOptimisticWithRefreshAsync` — an optimistic update that writes the regenerated lock token back onto your entity, so the same instance can be updated again without a manual re-`SELECT`. `UpdateOptimistic` left the entity holding the old token, which made the next update fail as a false concurrency conflict. The new value is read from the same statement via `UPDATE … OUTPUT` / `RETURNING` on SQL Server, PostgreSQL, SQLite, DuckDB, YDB and Firebird 5+, and via a follow-up `SELECT` by primary key elsewhere; the return value is the number of updated rows, `0` meaning a concurrency failure with the entity left untouched. Not available on ClickHouse, which reports no affected-row count. ([#5643](https://github.com/linq2db/linq2db/pull/5643))


`UpdateOptimisticWithRefresh` refreshes the concurrency token on your entity after an optimistic update, so you can reuse the same instance without re-reading the row.


---
<sub>Generated from commit `93a99bb`.</sub>

PR #5645, opened 2026-06-21
Resolve value-converter column DB type from the converter provider type

### LINKED ISSUE #195: F# option types are not supported

MaceWindu:
## Summary

A column with a `ValueConverter` but no explicit `DataType` resolved its DB type from the
**member (model) type**, which usually has no DB type of its own — so the column fell back to
`Undefined` and dropped the element's facets (precision/scale/length).

Most visibly, an F# `decimal option` (provider type `Nullable<decimal>`) collapsed to a bare
`Decimal` and silently truncated scale on strict providers instead of the provider's
`decimal(18,10)`; `string option` likewise ignored the provider's preferred string type.

## Fix

In `ColumnDescriptor.GetDbDataType`, when no explicit `DataType` is set, resolve the DB type from
the converter's **provider type** (`ToProviderExpression` result type) against the descriptor's
active, provider-inclusive `MappingSchema` — falling back to the member type when there is no
converter.

## Impact

- Affects every value-converter column with an undefined `DataType` (enums, custom scalar
  converters, F# option columns). Such columns now pick up the provider-faithful element type and
  its facets. **Baselines will move** accordingly.
- No public-API surface change.

## Context

Surfaced reviewing #5624 (automatic F# `'T option` mapping, fixes #195 / #4646): the metadata
reader can only resolve against `MappingSchema.Default`, so provider precision was lost. This core
change lets the F# reader drop its explicit `DataType` and get provider-faithful option columns;
that follow-up rides on #5624's branch once this merges.

### DIFF
--- a/Source/LinqToDB/Mapping/ColumnDescriptor.cs
+++ b/Source/LinqToDB/Mapping/ColumnDescriptor.cs
@@ -417,7 +417,13 @@ namespace LinqToDB.Mapping
 			DbDataType? dbDataType = null;
 
 			if (completeDataType && dataType == DataType.Undefined)
-				dbDataType = CalculateDbDataType(MappingSchema, systemType);
+				// When a ValueConverter is present, resolve the DB type from the converter's *provider* type
+				// against this descriptor's (active, provider-inclusive) MappingSchema rather than the member
+				// type - so e.g. an F# 'decimal option' (provider type Nullable<decimal>) picks up the
+				// provider's decimal(18,10) and 'string option' the provider's preferred string type. The
+				// member type often has no DB type of its own (the converter is what maps it to a column).
+				// Falls back to the member type when there is no converter.
+				dbDataType = CalculateDbDataType(MappingSchema, ValueConverter?.ToProviderExpression.Body.Type ?? systemType);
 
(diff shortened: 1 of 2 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed a column that has a `ValueConverter` but no explicit `DataType` resolving its database type from the model member type (which usually has none), so it fell back to `Undefined` and dropped precision/scale/length facets. The DB type is now resolved from the converter's provider type — for example an F# `decimal option` now maps to the provider's `decimal(18,10)` instead of collapsing to a bare `Decimal` and truncating scale. ([#5645](https://github.com/linq2db/linq2db/pull/5645))


_(none)_


---
<sub>Generated from commit `5c4e232`.</sub>

PR #5647, opened 2026-06-22
Fix PostgreSQL schema identity detection for sequence defaults

### LINKED ISSUE #5628: PostgreSQL scaffold fails with "Duplicate identity found" when multiple column defaults contain nextval(...)
## Summary

Scaffolding fails for a PostgreSQL table when more than one column has a default value that involves `nextval(...)`, and one of those defaults is part of a larger expression (for example, string concatenation).

LinqToDB appears to interpret both columns as identity-like and then throws a duplicate identity error during schema parsing.

## Environment

- LinqToDB CLI / Scaffold
- LinqToDB version: 6.3.0
- PostgreSQL 15

## Repro schema

```sql
CREATE TABLE sample_messages (
    code character varying(255) DEFAULT ('PREFIX_'::text || nextval('code_seq'::regclass)) NOT NULL,
    item_id integer DEFAULT nextval('item_id_seq'::regclass) NOT NULL,
    status smallint DEFAULT 0 NOT NULL,
    locale_id integer NOT NULL,
    item_type integer DEFAULT 1 NOT NULL
);

ALTER TABLE ONLY sample_messages
    ADD CONSTRAINT sample_messages_pkey PRIMARY KEY (item_id, item_type);

ALTER TABLE ONLY sample_messages
    ADD CONSTRAINT sample_messages_code_key UNIQUE (code);
```

## Observed behavior

Scaffold fails during schema parsing with an error like:

```text
Unhandled exception: Duplicate identity found on table sample_messages
   at LinqToDB.Schema.LegacySchemaProvider.ParseTable(...)
```

## PostgreSQL metadata

According to `information_schema.columns`:

- `item_id`: `is_identity = NO`, `column_default = nextval('item_id_seq'::regclass)`
- `code`: `is_identity = NO`, `column_default = ('PREFIX_'::text || nextval('code_seq'::regclass))`

So PostgreSQL does not classify either column as a true `IDENTITY` column.

## Notes

- Primary key is composite: `(item_id, item_type)`
- `code` is `UNIQUE`, not part of the primary key
- The issue seems related to how LinqToDB interprets sequence-based default expressions

### Reproduction Steps

1. Create the table
2. Execute scaffolding tool

### Error details

Unhandled exception: Duplicate identity found on table u_ggxscp01_00.messages
   at LinqToDB.Schema.LegacySchemaProvider.ParseTable(TableSchema table, SchemaObjects loadedObjects) in C:\GitHub\linq2db.claude.fix-nuget-size-guards\Source\LinqToDB.Scaffold\Schema\LegacySchemaProvider.cs:line 389
   at LinqToDB.Schema.LegacySchemaProvider.ParseSchema(DatabaseSchema schema, SchemaObjects loadedObjects) in C:\GitHub\linq2db.claude.fix-nuget-size-guards\Source\LinqToDB.Scaffold\Schema\LegacySchemaProvider.cs:line 89
   at LinqToDB.Schema.LegacySchemaProvider..ctor(DataConnection connection, SchemaOptions options, ILanguageProvider languageProvider) in C:\GitHub\linq2db.claude.fix-nuget-size-guards\Source\LinqToDB.Scaffold\Schema\LegacySchemaProvider.cs:line 68
   at LinqToDB.CommandLine.ScaffoldCommand.Scaffold(ScaffoldOptions settings, ScaffoldInterceptors interceptors, String provider, String providerLocation, String connectionString, String additionalConnectionString, String output, Boolean overwrite) in C:\GitHub\linq2db.claude.fix-nuget-size-guards\Source\LinqToDB.CLI\CommandLine\Commands\ScaffoldCommand.Execute.cs:line 122
   at LinqToDB.CommandLine.ScaffoldCommand.Execute(CliController controller, String[] rawArgs, Dictionary2 options, IReadOnlyCollection1 unknownArgs) in C:\GitHub\linq2db.claude.fix-nuget-size-guards\Source\LinqToDB.CLI\CommandLine\Commands\ScaffoldCommand.Execute.cs:line 102

### Regression?

It works in `3.6.0` using T4

### Linq To DB CLI version

6.3.0

### .NET version

10.0.301

### Database

PostgreSQL 15

### Operating system

Windows

### Configuration

_No response_

igor-tkachev:
## Summary
- Stop treating arbitrary `DEFAULT nextval(...)` expressions as native identities during PostgreSQL schema discovery.
- Prefer real PostgreSQL identity columns when they exist.
- When a table has multiple sequence-backed defaults, select a primary-key column first, otherwise keep the first matching column.
- Add regression coverage for mixed defaults, duplicate sequence defaults, and native identity columns.

## Testing
- Added PostgreSQL schema provider tests covering the affected identity-detection cases.
- Not run (not requested).


MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Fix verified correct against all four documented repro scenarios from #5628. The outer `IsIdentity` expression (`columns.IsIdentity OR NOT columns.HasIdentity AND columns.IsSequenceDefault AND columns.Ordinal = columns.IdentityFallbackOrdinal`) parses as intended (`AND` binds tighter than `OR`); the nested subqueries compute `HasIdentity`, `IsSequenceDefault`, `HasPrimaryKeySequenceDefault`, and `IdentityFallbackOrdinal` (`min(ordinal) OVER (PARTITION BY ...)`, primary-key-preferring) correctly.
- [x] No output-mapping regression: the outer `SELECT` still projects the same 15 columns in the same order, so reader indices align. The `SkipOnUpdate` `NOT (A OR B AND C AND D)` predicate is semantically identical to base across the reformat.
- [x] No SQL-injection surface: only internal version-derived constants and the existing schema filter are interpolated; test names derive from integer counters.

### Suggestion (SUG)
- [ ] **SUG001** — Rationale for dropping (vs. extending) detection of sequence-backed defaults embedded in larger expressions
  Why: The new anchored match (`^[[:space:]]*nextval[[:space:]]*\(`) intentionally stops treating a column like `code varchar DEFAULT ('PREFIX_'::text || nextval('code_seq'::regclass))` as an identity. That fixes the duplicate-identity crash, but it does so by *ignoring* such a column rather than handling it — the column is still server-generated from a sequence on insert, just not a bare integer identity.
  Fix: Could you note the rationale for the removal in the PR? Specifically: is dropping detection of embedded-`nextval` defaults the intended final behavior, or should detection instead be *extended* so these columns are still recognized as sequence-backed (e.g. so they keep an appropriate skip-on-insert marking) rather than silently demoted to ordinary columns? Either answer is fine — flagging so the choice is explicit and recorded.

MaceWindu:
**Re SUG001** — some history on the `nextval` heuristic this PR narrows.

The `column_default ~* 'nextval'` detection was introduced in commit `448cb3e7a` ("Working on PostreSQL schema provider.", 2013-05-14) — a direct push with no associated PR or issue (it predates the PR-based workflow). It has stood unchanged for ~13 years until now, and `git log -S nextval` shows only that commit and this PR have ever touched it.

Crucially, it was written *before* PostgreSQL had `GENERATED ... AS IDENTITY` (that arrived in PG 10, 2017). At the time the loose `nextval` substring match was effectively the *only* identity signal available, which is why it was written so broadly — and why it now over-matches embedded expressions like `'PREFIX_' || nextval(...)`.

I agree with removing the loose `nextval` heuristic. But that decision is coupled to one we should make explicitly: **do we still support pre-10 PostgreSQL?** Native identity detection (`is_identity = 'YES'`) requires PG 10+, so on older servers — and for plain `serial` columns, which are not native identities even on PG 10+ — the anchored bare-`nextval(...)` fallback is the only thing keeping identity detection working. If we want to fully drop the `nextval` path eventually, we'd be dropping serial/pre-10 identity detection with it, so let's settle the supported-version policy as part of this change.

igor-tkachev:
@MaceWindu 

Thanks, fully agreed.

For this PR, the intended scope is narrower: keep the `nextval` fallback for PostgreSQL versions/objects where native identity metadata is not available, but stop treating arbitrary expressions containing `nextval` as identity columns.

So the current behavior is intentional:
- native PostgreSQL identity metadata wins when available;
- bare `DEFAULT nextval(...)` remains supported for pre-10 PostgreSQL and `serial` columns;
- embedded expressions like `'PREFIX_' || nextval(...)` are no longer treated as identity, because linq2db/scaffold currently has only a single-column identity model and derives `SkipOnInsert` from `IsIdentity`.

I agree that sequence-backed/default-generated columns are a broader metadata topic. Ideally, linq2db should eventually model server defaults separately from identity, so columns with arbitrary default expressions can be represented without overloading `IsIdentity` or forcing `SkipOnInsert`.

That feels like a separate PR/issue though. This PR should only fix the incorrect identity detection and duplicate-identity failure from #5628 while preserving existing serial/pre-10 support. We should record the broader server-default metadata follow-up somewhere so it does not get lost.

sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed PostgreSQL schema discovery misclassifying identity columns: a `nextval(...)` call embedded in a larger default expression (e.g. `'PREFIX_' || nextval(...)`) is no longer treated as a native identity column — only a direct serial-style `DEFAULT nextval(...)` still is. A true identity column is preferred when present, and when a table has multiple sequence-backed defaults a primary-key column is selected first.


_(none)_


---
<sub>Generated from commit `7e33354`.</sub>

PR #5648, opened 2026-06-22
Add UUIDv7 support: Sql.NewGuid7() + map Guid.CreateVersion7()

### LINKED ISSUE #5646: Add UUIDv7 support: Sql.NewGuid7() API + map Guid.CreateVersion7()
### Summary

Add UUIDv7 generation support: a new `Sql.NewGuid7()` extension and translation of
.NET 9+ `Guid.CreateVersion7()` to provider-native UUIDv7 generators where available.

### Background

.NET 9 adds `Guid.CreateVersion7()` (RFC 9562 UUIDv7, time-ordered). Several databases
now expose server-side UUIDv7 — e.g. Firebird 6's `GEN_UUID([<version>])` (default 4)
accepts version 7. linq2db already maps `Guid.NewGuid()` to UUIDv4 generators.

### Current behavior

- `Source/LinqToDB/Internal/DataProvider/Firebird/Translation/FirebirdMemberTranslator.cs:375`
  — `Guid.NewGuid()` → `Gen_Uuid()` (UUIDv4 only). No UUIDv7 path, and no `Sql.NewGuid7` API.

### Expected behavior

- New `Sql.NewGuid7()` extension producing a server-side UUIDv7 on supporting providers.
- Translate `Guid.CreateVersion7()` (.NET 9+) to the same server-side UUIDv7 where supported
  (e.g. Firebird 6 → `GEN_UUID(7)`); fall back to client-side generation otherwise.

### Notes

- Cross-provider: enumerate which providers expose server-side UUIDv7 and define the fallback
  for those that don't.
- Split out of Firebird 6 support work (#5483 / PR #5485) to keep that PR provider-scoped.

MaceWindu:
Fixes #5646

Adds server-side UUIDv7 generation: a new `Sql.NewGuid7()` extension and translation of .NET 9+ `Guid.CreateVersion7()` to provider-native UUIDv7 generators, mirroring the existing `Sql.NewGuid()` / `Guid.NewGuid()` v4 mechanism.

## API

- `Sql.NewGuid7()` — returns a v7 GUID. On net9+ it delegates to `Guid.CreateVersion7()`; on `net462` / `netstandard2.0` / `net8.0` (no `Guid.CreateVersion7`, no polyfill) it builds a valid RFC 9562 v7 GUID client-side.
- `Guid.CreateVersion7()` (net9+) is registered for translation alongside `Sql.NewGuid7()`.

Both route through a new `protected virtual TranslateNewGuid7Method` on `ProviderMemberTranslatorDefault` (returns `null` by default → expression falls back to client-side generation, exactly as `NewGuid` does).

## Server-side translation

| Provider | Function | Min DB version |
|---|---|---|
| PostgreSQL | `uuidv7()` | 18 (new `PostgreSQL18MemberTranslator`, `>= v18` arm) |
| DuckDB | `uuidv7()` | 1.3.0 |
| ClickHouse | `generateUUIDv7()` | 24.5 |
| MariaDB | `UUID_v7()` | 11.7 (new `MariaDBMemberTranslator`, split off the MySQL-8-shared translator since MySQL 8 has no `UUID_v7()`) |

DuckDB / ClickHouse / MariaDB have no version-dialect split in linq2db, so the function is emitted unconditionally (older DB versions predate practical support; noted in code comments). Providers without server-side support generate client-side.

## Tests

`FunctionTests.NewGuid7` and `FunctionTests.CreateVersion7` (the latter under `#if NET9_0_OR_GREATER`) assert the returned GUID's version nibble is `7`, on **all** providers (`[DataSources]`) — native UUIDv7 where supported, client-side fallback elsewhere. `Sql.AsSql` forces SQL evaluation (not silent client parameterization) and the baseline is disabled (non-deterministic value). Verified green server-side on PostgreSQL 18, DuckDB, ClickHouse, MariaDB and via client fallback on SQLite, MySQL 8.0/5.7, PostgreSQL 16 (the net8.0 fallback path also exercises the RFC 9562 builder).

## Notes

- **Merge is blocked until Firebird 6 support (#5483 / PR #5485) lands** — `FirebirdVersion.v6` does not exist on `master` yet. Firebird's `GEN_UUID(7)` arm will be added as a follow-up.
- Release build verified clean on `net10.0` (analyzers + PublicAPI) and `netstandard2.0` (oldest-TFM client fallback).

## Follow-up commit

Inlined the meaningless `timePart` local (a copy-paste from an unrelated datetime override) into a direct `return` across the `NewGuid` and new `NewGuid7` translator overrides. Cosmetic only; Access keeps its genuine `timePart`.

## Follow-up commits (review)

- Added XML doc to `Sql.NewGuid7()` (required for new public API).
- Corrected the ClickHouse min-version comment to 24.5+ (`generateUUIDv7` landed in ClickHouse 24.5, not 24.1).


igor-tkachev (dismissed):
Direction looks good and the general translator hook mirrors the existing `Sql.NewGuid()` / `Guid.NewGuid()` mechanism well.

I think we need one compatibility change before merge: MariaDB should not get unconditional server-side translation in this PR.

`UUID_v7()` is available only starting with MariaDB 11.7, but current linq2db dialect selection has only `MySqlVersion.MariaDB10`. Mapping all MariaDB dialects to `MariaDBMemberTranslator` means MariaDB 10.x / 11.0-11.6 users will get a runtime SQL error instead of the documented client-side fallback.

My preferred option is to keep MariaDB on fallback for this PR:

- keep `MySqlVersion.MariaDB10 => new MySql80MemberTranslator()` as before;
- remove `MariaDBMemberTranslator`;
- add MariaDB 11.7+ native UUIDv7 support later with a real version split/detection and provider tests.

That keeps the new API usable on MariaDB without introducing a version-specific runtime break.

A couple of follow-up points:

- For ClickHouse, please verify repeated calls in the same query, e.g. `new { A = Sql.NewGuid7(), B = Sql.NewGuid7() }`, because UUID functions have the optional `expr` argument to bypass CSE.
- Current tests validate only the returned GUID version. It would be better to add SQL-generation assertions for native providers and fallback providers, especially PostgreSQL 18 vs PostgreSQL < 18 and MariaDB fallback.
- XML doc should avoid implying that server-side fallback is based on actual connected server capability when the provider translator emits the function unconditionally.

sdanyliv (dismissed):
Looks good, but merge after Eager Loading. There are MemberTranslator conflicts.

MaceWindu:
Remaining review issues are out-of-scope:
- MariaDB version: currently we don't have versioning for MariaDB and taking into account how they release features in non-major releases I have no intention to add it without user request. As for this feature - nobody force user to use unsupported functionality. We can review it e.g. in v7 alongside with some other actively developed databases like pgsql and maybe add some explicit support policy for each.
- ClickHouse: guid functions is a PITA and this is not a new situation here, same story with regular guid `generateUUIDv4` function. If needed we can add additional Sql.xxx overloads with parameter to support random guids, but it is out of scope here.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Added server-side UUIDv7 generation — a new `Sql.NewGuid7()` extension and translation of .NET 9+ `Guid.CreateVersion7()`, mirroring the existing `Sql.NewGuid()` / `Guid.NewGuid()` (v4) mechanism. Providers with a native UUIDv7 function emit it server-side (PostgreSQL 18+, DuckDB, ClickHouse, MariaDB); every other provider generates a client-side RFC 9562 v7 GUID.

##### deep-dive: UUIDv7 generation

```cs
var id  = db.GetTable<Person>().Select(_ => Sql.NewGuid7()).First();
var id2 = db.GetTable<Person>().Select(_ => Guid.CreateVersion7()).First(); // .NET 9+
```

Native server-side functions: PostgreSQL 18+ `uuidv7()`, DuckDB 1.3.0+ `uuidv7()`, ClickHouse 24.5+ `generateUUIDv7()`, MariaDB 11.7+ `UUID_v7()`. PostgreSQL is version-aware — below 18 it falls back to client-side generation. For DuckDB, ClickHouse and MariaDB the native function is emitted unconditionally (linq2db doesn't probe the server version), so targeting a server older than the function's introduction raises a server-side error rather than falling back. Providers without any native generator always produce the v7 GUID client-side.


New `Sql.NewGuid7()` + `Guid.CreateVersion7()` mapping for server-side (or client-side fallback) UUIDv7 generation.


---
<sub>Generated from commit `52bd47d`.</sub>

PR #5651, opened 2026-06-23
Tests: stop polluting global MappingSchema.Default in ConverterInsertTests/Issue531Tests

MaceWindu:
### Problem

`ConverterInsertTests.Test` and `Issue531Tests.Test` registered type converters on the **process-global** `MappingSchema.Default` and never restored it:

```csharp
MappingSchema.Default.SetConverter<Dictionary<string, string>?, DataParameter?>(
    obj => obj == null ? null : new DataParameter { Value = obj.Keys.FirstOrDefault(), DataType = DataType.NVarChar });
```

That leaked converter turns a `Dictionary<string,string>` into the first key as `NVarChar`. Any later test that maps the same CLR type differently then picks up the polluted converter. Concretely, the PostgreSQL `Issue4556` json tests map `Dictionary<string,string>` to `json`/`jsonb` columns; once polluted they emit `character varying` where json is required and fail with:

```
42804: VALUES types json and character varying cannot be matched
```

In natural alphabetical test order the polluters (`Tests.UserTests.*`) run **after** their victims (`Tests.DataProvider.PostgreSQLTests`), so the leak is invisible in a serial run. Under **parallel execution** they run concurrently across provider lanes and corrupt the shared global converter cache mid-query — which is how this surfaced while enabling parallel test execution.

### Fix

Use a **local** `MappingSchema` instead of mutating `MappingSchema.Default`, and attach it to the connection:

- `ConverterInsertTests.Test` → `GetDataContext(context, ms)` — mirrors the sibling `TestEnumString` in the same file, which already does it correctly.
- `Issue531Tests.Test` → `db.AddMappingSchema(ms)` — the established local-schema pattern used by `Issue464Tests`, `Issue1438Tests`, `MiniProfilerTests`, etc.

No production code changes — test-only global-state hygiene.

### Verification

- Reproduced the failure deterministically (sequential pollute-then-merge) on PostgreSQL 16 — exact `42804` error; the local-schema variant passes.
- `ConverterInsertTests` (all 5 cases) green on SQLite.MS after the fix.
- `Tests/Linq` builds clean (`net10.0`).


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `ffdc833`.</sub>

PR #5653, opened 2026-06-23
Tests: build identity merge fixtures per-invocation instead of shared static

MaceWindu:
### Problem

`MergeTests` held its identity test fixtures as **`static readonly` arrays of mutable objects**:

```csharp
private static readonly MPerson[]  IdentityPersons  = { ID=101..106 ... };
private static readonly MPatient[] IdentityPatients = { ... };
private static readonly MDoctor[]  IdentityDoctors  = { ... };
```

`PrepareIdentityData` — called by every identity/association merge test — *mutates* these statics on each run (assigns DB-generated identities via `InsertWithIdentity`, wires `IdentityPatients[i].PersonID = IdentityPersons[j].ID`, etc.), and the test assertions read them back.

Under parallel test execution each provider context runs on its own resource lane. The failing CI legs each run **two contexts concurrently on separate databases** — SQL Server `.MS` (Microsoft.Data.SqlClient → `TestDataMS`) + non-`.MS` (System.Data.SqlClient → `TestData`), and "MySQL 5.7 **both providers**". The two lanes raced the shared arrays, so:

- a `Patient`/`Doctor` got a `PersonID` from the *other* lane's database → `The INSERT statement conflicted with the FOREIGN KEY constraint "FK_Patient_Person"`, or
- an assertion read an `ID` the other lane had just overwritten → `Assert.That(actual.ID, Is.EqualTo(expected.ID))` mismatch.

This is invisible to the `AssertState` drift detector (the fixtures are in-memory .NET objects and the tests roll back their DB writes), and it needs ≥2 concurrent lanes — which is exactly why it surfaced only on the multi-context legs and never on single-lane runs.

### Fix

`PrepareIdentityData` now **builds the three arrays locally on each call and returns them** (`IdentityData` holder); each test uses the returned copy. Each lane works on its own data, so there is no shared mutable state to race. Test-only — no production change, no thread-affinity, no fixture-lifecycle change.

Adds `IdentityFixtures_ArePerInvocation`, which fails if the fixtures are ever made shared again (asserts two invocations return independent arrays).

### Verification

- Reproduced the shared-state race deterministically (an isolation probe failed on the shared static, passed once isolated).
- **363/363 `MergeTests` pass on PostgreSQL 16**; regression + `TestAssociationsData` pass on SQLite.MS.
- `Tests/Linq` builds clean in **Testing and Release** (`net10.0`) — analyzers pass.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `fec0384`.</sub>

PR #5656, opened 2026-06-25
StackGuard: probe stack every 8 levels to fix flaky StackOverflow

MaceWindu:
## Problem

The deep-recursion stack guard used by the expression / SQL visitors intermittently crashes CI with an uncatchable `StackOverflowException` (e.g. `StackUseTests` on the `net462` / Linux legs) instead of recovering via its thread-hop mechanism.

## Root cause

`StackGuard.Enter` probed `RuntimeHelpers.TryEnsureSufficientExecutionStack()` only **every 64 recursion levels**. Between two probes up to ~64 × per-level stack is consumed with no check. Captured at the hop point on x64 there are 3 frames per level (`ExpressionVisitorBase.Visit` → `ExpressionVisitorUtils.VisitArguments` → `ExpressionVisitor.VisitMethodCall`), depth **4737** before the first hop ⇒ **~220 bytes/level**, so 64 levels ≈ **~14 KB**.

`TryEnsureSufficientExecutionStack` only guarantees a fixed margin. On tighter configurations (`net462` / Linux, larger frames) those ~14 KB can exceed the margin, so the stack overflows **between probes** — a hard `StackOverflowException` that crashes the process — rather than the intended graceful hop.

## Fix

Probe **every 8 levels** (named `StackProbeInterval` constant), bounding the between-probe stack to ~2 KB — comfortably inside the guaranteed margin on every TFM. The probe is a cheap JIT intrinsic and only matters for deep recursion (shallow queries probe ~once), so the cost is negligible. Hop semantics are unchanged; it just hops slightly sooner.

## Verification

`StackUseTests` hop tests pass locally on net10.0 — `TestExpressionVisitorHops` / `TestSqlVisitorHops` (0/1/2 → nested `InsufficientExecutionStackException`, 10 → completes) and `TestPreserveExceptionOnHop`. The hard `StackOverflow` only reproduces on the tight-margin legs (net10.0/x64 hops gracefully at depth 4737), so CI on `net462` / Linux is the final confirmation.

### DIFF
--- a/Source/LinqToDB/Internal/Common/StackGuard.cs
+++ b/Source/LinqToDB/Internal/Common/StackGuard.cs
@@ -9,6 +9,11 @@ namespace LinqToDB.Internal.Common
 {
 	public sealed class StackGuard
 	{
+		// How often (in recursion levels) Enter probes the remaining stack. Small enough that the stack
+		// consumed between two probes stays well inside the margin TryEnsureSufficientExecutionStack
+		// guarantees, so deep recursion always hops gracefully instead of overflowing between probes.
+		const int StackProbeInterval = 8;
+
 		// Use interlocked to be safe when execution switches threads.
 		int _hopCount;
 		int _internalDepth;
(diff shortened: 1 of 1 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed an uncatchable `StackOverflowException` that could crash the process on deeply nested / recursive queries instead of recovering gracefully. The deep-recursion stack guard now probes the remaining stack more frequently, so it reliably switches to its thread-hop fallback (or throws a catchable `InsufficientExecutionStackException`) before the stack is exhausted. ([#5656](https://github.com/linq2db/linq2db/pull/5656))


_(none)_


---
<sub>Generated from commit `81b0f9b`.</sub>

Issue #5169, opened 2025-10-30
Discrepancy between direct and remote context generated SQL


MaceWindu:
### Description

Implementation of remote/non-remote baselines comparison found several cases where we generate different SQL. Some of them were fixed, others are tracked by this issue.

- [PostgreSQL] `TestTypesInsertByMerge()` test deduplicate enumerable source parameters, but direct call - don't. Caused by different call sites to `SqlValuesTable.BuildRows(...)`
- [ClickHouse] `InTest2, Null_In_Null_Test1, Null_In_Null_Test2, Null_NotIn_Null_Test1` tests generate same alias for different tables without remote context, because aliases assinged by `PrepareQueryAndAliases` and then `BuildSql` finalize `IN (subquery)` predicate which introduce additional subquery. Other databases not affected as they can use correlated `EXISTS(subquery)` predicate
- [Access, OleDB] `ToSqlQuery_WithParametersDeduplication(false)` generate two parameters with same value (expected), but for remote context it generates them with same name
- [Access] `Truncate2` while it is easy to fix by adding extra optimization, it demonstrates issue with optimizations order. Expression `-(-1 * x)` should be reduced as `-(-1 * x)` -> `1 * x` -> `x`, but last step executed only for remote context leaving non-remote sql having `1 * x` expression
- [SQL CE] `InsertFromSelectWithNullableFilter` : remote context doesn't has column alias for top level field. Some aliases generation ordering issue
- [PostgreSQL] `MappingTypingByConstant_FromQuery_UInt64(true)` : in remote context we evaluate parameter type before expression converter check necessity of Floor function

Currently baselines for those tests are disabled to avoid CI failures.

None of those are critical issues as both contexts generate proper SQL (maybe except ClickHouse one having risks to generate invalid SQL with conflicting aliases)

### Regression?

No

### Linq To DB version

6.0.0

Issue #4437, opened 2024-02-28
Inconsistent object mapping behavior when querying using LINQ  vs `QueryToListAsync`


jeffijoe:
### Describe your issue

When using `QueryToListAsync<Test>` where `Test` is a record **with a primary constructor** and `[property: Column("some_column")]` attributes, it appears the column names specified in the attributes are not being used. If the record does not have a constructor, then it works fine. Additionally, using `connection.GetTable<MyType>()` to query, then the record with a primary constructor also works. Removing the constructor makes it work with `QueryToListAsync`.

### Steps to reproduce

Using the following as our table/query result type

```csharp
[Table("test")]
public record Test(
    [property: Column("some_column")]
    string SomeColumn);
```

When using the following:

```csharp
await connection.QueryToListAsync<Test>("select some_column from test")
```

The resulting records will have `null` in `SomeColumn` because the `[Column]` attribute which provides the correct name is not being used.

If instead I do

```csharp
await db.GetTable<Test>().FirstOrDefaultAsync();
```

Then the column mapping works and I get the expected value in `SomeColumn`.

If I change my record to not have a constructor, then everything works fine with `QueryToListAsync` as well:

```csharp
[Table("test")]
public record Test
{
    [Column("some_column")]
    public string SomeColumn { get; init; }
}
```

Is there a difference in how Linq2Db handles object mapping for LINQ-based queries vs SQL-based ones?

### Environment details

`Linq To DB` version: 5.4

Database (with version): Postgres 13

ADO.NET Provider (with version): Npgsql 8.0

Operating system: macOS

.NET Version: 8.0

MaceWindu:
Yes, currently mapping for raw queries works in a different way. We will unify it in upcoming 6.0 release

jeffijoe:
Thank you @MaceWindu, and thank you for all the work you guys are doing on Linq2Db, it's an incredible library! 🙏

MaceWindu:
We reviewed this issue and currently it is expected behavior. We map raw SQL by name in C#, so proper query should be 

```cs
// note alias added for column
await connection.QueryToListAsync<Test>("select some_column AS ColumnName from test")
```

Changing this will be:
- breaking change to existing users
- will break multi-table queries with same column name (`SELECT table1.field, table2.field FROM ...`)

We don't like current situation, but changes to it should be carefully designed

jeffijoe:
[...]

PR #5657, opened 2026-06-25
Fix concurrent-execution SQL alias corruption (non-mutating aliasing)

### LINKED ISSUE #5599: Final-alias rewriting breaks by-name column mapping for raw SQL
When final aliases are applied, by-name column mapping for raw-SQL queries breaks. Affects SqlCe and YDB.

Need to determine whether by-name mapping should resolve against the final aliases or the original column names, and fix the mismatch.

Surfaced by `Linq/QueryGenerationTests` while completing the YDB provider (#5564).

MaceWindu:
## Problem

Under concurrent execution, a cached **parameter-dependent** query could render corrupted SQL — swapped/duplicated table aliases and column names (on PostgreSQL surfacing as `42703 column does not exist`, elsewhere as wrong values). Root cause: for parameter-dependent queries (`optimizeAndConvertAll == false`, i.e. every `continuousRun`), `AliasesHelper.PrepareQueryAndAliases` **mutated the shared cached `SqlStatement`'s aliases in place**, lock-free. Two threads executing the same cached query raced on that mutation.

This was surfaced by the parallel-test-execution work (#5614).

## Fix — make aliasing non-mutating

Finalized identifier names now live in `AliasesContext` (keyed off the statement), not on the AST nodes, so a cached statement is immutable across executions and can be rendered concurrently. Four parts:

1. **Out-of-node alias storage + context-aware derivation.** `AliasesContext` holds finalized table/column/field names; `GetColumnAlias`/`GetTableAlias` derive a node's alias by recursing through sub-expression names *via the context*, not the node's own `GetAlias`.
2. **Table aliases keyed by `SourceID`, not object identity.** One logical source is wrapped by several `SqlTableSource` instances in a single statement (a correlated reference resolves to a different wrapper than the FROM clause); all must resolve to the same finalized alias.
3. **Nested builders inherit the parent's `AliasesContext`** (in the parent-copy ctor). Subqueries, CTE bodies and set-operation parts are rendered by a fresh `BasicSqlBuilder` whose context otherwise defaulted to empty.
4. **Remote (LinqService) path aliases the *prepared* statement.** `PrepareStatementForSql` produces new nodes; aliasing now runs after it, on the statement actually rendered.

## Guard — `LINQ2DB0001` analyzer

A new analyzer in the `CodeGenerators` project flags reads of `SqlColumn.Alias` / `SqlTableSource.Alias` **only inside `BasicSqlBuilder`-derived types** (the render path), where the node value is not the finalized alias. Optimizers/builders/diagnostics that legitimately read the raw node alias are unaffected; `RawAlias` is the escape hatch. This replaces a development-time `[Obsolete]` guard.

## Test

`ConcurrencyTests.ParallelExecutionDoesNotCorruptAliases` primes a parameter-dependent `FromSql` query (distinct value per quoted, mixed-case column) and runs it from 4 threads × 64 iterations, asserting the column→value mapping stays intact. Light load on purpose — a sometimes-red CI signal pre-fix, reliably green with the fix. Runs on PostgreSQL (strong signal — quoting matters) and SQLite.

## Verification (local)

- 2606 SQLite tests green (direct + LinqService), up from 15/16 failing mid-development.
- Regression test stably green across repeated runs on PostgreSQL 18 + SQLite.MS.
- Release `net10.0` (analyzers on) + `netstandard2.0` build clean — 0 `LINQ2DB0001`, 0 `CS0618`.

## Follow-up commits

Two failures the full `test-all` matrix surfaced on this branch, both fixed:

- **`9cd295360` — restore per-execution convert for parameter-dependent queries.** Passing `isAlreadyOptimizedAndConverted: true` unconditionally turned the build-time `OptimizationContext.OptimizeAndConvert` in `BuildSql` into a no-op. The single upfront `OptimizeAndConvertAll` does not cover every localized refinement, so parameter-dependent queries lost them on the direct path while the remote runner (which rebuilds per execution) kept them — producing direct/remote baseline divergence and, on some providers, invalid SQL: the InsertOrUpdate (`ON CONFLICT`/merge) UPDATE-SET value cast fell back to the default string length instead of the column's `DbDataType` length (`NVarChar(255)` vs `NVarChar(11)`), and ClickHouse `MapInvertOperator` emitted a bare `&` instead of `bitAnd(...)`. Gating the flag on `optimizeAndConvertAll` keeps the upfront pass authoritative for non-parameter-dependent queries while letting parameter-dependent ones re-run convert in `BuildSql`. Safe because their visitors are non-mutating (Transform mode), so the build-time pass produces new nodes without mutating the shared cached statement and cannot reintroduce the aliasing race — `ParallelExecutionDoesNotCorruptAliases` stays green (verified 3× on PostgreSQL).
- **`91a432720` — `TestInvaildAliasExprUsage` expects DB rejection.** `Sql.AliasExpr()` placed where it cannot become a table alias produces invalid SQL; linq2db cannot inspect placeholder position in opaque raw SQL, so the database rejects it. The test originally (#1863) asserted `Npgsql.PostgresException`; the `InvalidOperationException` expectation was incidental to the #3401 query-parsing refactor. The non-mutating aliasing again surfaces the misuse at the database, so the original contract is restored.
- **`f242c92` — fix direct/remote alias-numbering divergence for parameter-dependent queries.** The non-mutating model keys finalized aliases by node identity, but parameter-dependent queries rebuild the statement each execution (Transform mode → fresh node identities), so re-feeding the previous run's `query.Aliases` as the alias seed produced non-deterministic alias numbering that drifted between executions and diverged from the remote path (which always aliases fresh server-side over a freshly deserialized statement). Surfaced as 7 Oracle 11 `TakeSkipTests` failures on `*.LinqService`: the ROWNUM nested-paging wrapper aliased `t3` on the first execution and `t1` on the second, while remote stayed `t3`. Alias fresh on every execution instead — deterministic for a given statement shape, and identical to the remote rendering. Non-parameter-dependent queries are unaffected (they cache their commands and short-circuit before aliasing). `ParallelExecutionDoesNotCorruptAliases` stays green.
- **`62233c66` — fix #5599: preserve physical column names for bare-field root columns.** Raw-SQL by-name entity materialization resolves result columns on their physical column name, but a bare entity-field root column was aliased to its CLR member name (`PersonID` rendered `as ID`). Alias-skipping providers never emit it (result column stays physical → mapping works); providers that force root aliases (SqlCe, YDB) emitted the member-name alias, renaming the result column and breaking by-name mapping (`ToSqlQuery → Query<T>` returned defaults). The root-select alias uniquifier now seeds bare-field columns from the field's physical name, so result columns keep physical names across all providers; the uniquifier still suffixes genuine duplicates. **Complements #5659 (#4437):** that fixed the read side (honor `[Column]`/`[ValueConverter]` in raw-SQL materialization); this fixes the write side (emit physical result-column names) — together the `ToSqlQuery → Query<T>` round-trip is correct on alias-forcing providers. Verified: full YDB (8255 tests, 0 failed) + full SqlCe (7832 tests, only 2 pre-existing env-timezone failures).
- **`183500681` — ungate the direct/remote-divergence tests the above resolves.** The fresh-aliasing fix resolves the derived-table alias-numbering class: ungated on ClickHouse (`InTest2`, `Null_In_Null_Test1/2`, `Null_NotIn_Null_Test1`), YDB (`Null_NotIn_Null_Test3`, `Issue4520Test`), PostgreSQL (`MappingTypingByConstant_FromQuery_UInt64`), SqlCe (`InsertFromSelectWithNullableFilter`), Access (`Truncate2`). Remaining `#5169` gates are distinct divergence classes the fix does not address, narrowed to the affected provider: YDB CTE-alias collision (`t1_1` vs `t1`), PostgreSQL MERGE-source parameter dedup, Access `ToSqlQuery_WithParametersDeduplication`, and the YDB CAST-elision cases.
- **`9d777c676` — fix recursive-CTE column references after merging master (`SqlField` → `SqlCteField`).** The master merge brought #5450's CTE-field refactor: CTE columns are now `SqlCteField` / `SqlCteTableField`, not `SqlField`. `BasicSqlBuilder` rendered `SqlCteTableField` references from the delegated `CteField.Name` directly, bypassing the finalized name held in `AliasesContext`. Under the non-mutating model the finalized name lives only in the context, so once the aliasing pass uniquified a CTE column (`Level` → `Level_1`) the `WITH`-clause declaration used the finalized name while the recursive self-reference and the outer references still emitted the stale `Level`, producing `Invalid column name 'Level'` (SQL Server) / `no such column: t3.Level` (SQLite) / `Dynamic SQL Error` (Firebird) — 4714 `CteTests` failures across the provider matrix. Both `SqlCteTableField` render sites (aliased and `_disableAlias`) now resolve through `AliasesContext.GetFieldName`, which falls back to the delegated name for non-renamed CTEs so unaffected queries stay byte-identical. Verified: full `CteTests` (105) green on SQLite.


MaceWindu (commented):
> [!IMPORTANT]

Reviewed the non-mutating aliasing refactor across all 13 changed files. **No defects found.** The change is behaviour-preserving for single-threaded rendering and removes the concurrent-execution alias-corruption race. The notes below record what was verified.

## Review notes
- [x] **SourceID-keyed aliasing is safe.** Aliasing runs after the structural (query-level) optimizers, so the set of sources is frozen; the post-aliasing per-command convert is expression-level only — it returns FROM-source field/column references by identity (`QueryElementVisitor.VisitSqlFieldReference` / `VisitSqlColumnReference` return the element unchanged) and never reconstructs a table's `SourceID`, and the FROM clause renders aliases on the original sources (`BasicSqlBuilder.BuildFromClause`). So the `AliasesContext` lookups, keyed by `SourceID`, always resolve. The only path that mints a new `SourceID` (the `SqlTable` copy-ctor in `QueryElementVisitor.VisitSqlTable` Transform mode) is reachable only during the structural passes that precede aliasing.
- [x] **No alias-text shift vs `master`.** A correlated self-reference subquery plus three association-correlation queries render byte-identical SQL with and without this change (checked on PostgreSQL), so the refactor itself introduces no baseline alias churn.
- [x] **`LINQ2DB0001` coverage is complete for the render path.** Only `BasicSqlBuilder` and `YdbSqlBuilder` read the guarded `SqlColumn.Alias` / `SqlTableSource.Alias` getters, and both are converted, so the Release build stays clean. `SqlTable.Alias` is intentionally outside the guard (its alias stays on the node, and `AliasesContext.GetTableAlias` falls back to it); this is now documented in the analyzer's remarks (commit f3f2f09).
- [x] **Concurrency regression coverage is appropriate.** `ParallelExecutionDoesNotCorruptAliases` is the right guard for the race; a deterministic SQL-rendering test would add nothing here, since the refactor produces identical SQL with and without the fix.
- [x] Two documentation-only clarifications were added on the branch during review (commit f3f2f09): the `IQueryContext.Aliases` reuse-seed semantics and the `SqlTable` analyzer-exclusion rationale.

## Baselines
Baselines review skipped per request. No alias-text shift was observed on representative correlated / association queries, but the full baseline set was not diffed here — rely on CI's generated baselines for the complete picture.

Source/LinqToDB/Internal/SqlQuery/AliasesContext.cs
```diff
@@ -10,6 +10,20 @@ public sealed class AliasesContext
...
 
+		// Finalized identifier names. Holding the names here (instead of mutating the AST nodes in
+		// place) keeps the statement immutable across executions, so the same cached statement can be
+		// aliased / rendered concurrently without a data race.
+		//
+		// Table aliases are keyed by the source's stable SourceID, not object identity: one logical
+		// source can be wrapped by several SqlTableSource instances within a single statement (a
+		// correlated reference resolves to a different wrapper than the one in the FROM clause), and
+		// every wrapper must resolve to the same finalized alias.
+		readonly Dictionary<int, string>            _tableAliases  = new ();
```
MaceWindu: fix alignment

Source/LinqToDB/Internal/SqlQuery/AliasesContext.cs
```diff
@@ -20,20 +34,95 @@ public void RegisterAliased(IReadOnlyCollection<IQueryElement> elements)
...
+		public IReadOnlyCollection<IQueryElement> GetAliased()
 		{
-			return _aliasesSet.Contains(element);
+			return _aliasesSet;
 		}
 
-		public IReadOnlyCollection<IQueryElement> GetAliased()
+		#region Finalized names
+
+		public void SetTableAlias(SqlTableSource tableSource, string alias) => _tableAliases [tableSource.SourceID] = alias;
```
MaceWindu: fix alignment

Source/LinqToDB/Internal/SqlQuery/AliasesContext.cs
MaceWindu: Fixed in e20c84b83 (re-aligned the readonly dictionary declarations).

Source/LinqToDB/Internal/SqlQuery/AliasesContext.cs
MaceWindu: Fixed in e20c84b83 (re-aligned the Set* one-liner columns).

sdanyliv (dismissed):

Tests/Linq/Linq/ConcurrencyTests.cs
```diff
@@ -999,5 +1002,77 @@ void AssertData(ConcurrencyTable<byte[]> record, bool equals = false)
...
+			// prime once so subsequent executions hit the cached (re-alias) path
+			using (var prime = GetDataContext(context))
+				prime.FromSql<AliasRaceRow>(sql, NewParam()).Single();
+
+			Exception? error = null;
+			var        bad   = 0;
+
+			void Worker()
+			{
+				for (var i = 0; i < 64 && error == null && Volatile.Read(ref bad) == 0; i++)
```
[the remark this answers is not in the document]
MaceWindu: Fixed in 38b39b8c5 — the worker loop now reads `error` via `Volatile.Read(ref error)`, matching the sibling `Volatile.Read(ref bad)` read.

Tests/Linq/Linq/FromSqlTests.cs
```diff
@@ -556,7 +556,9 @@ public void TestInvaildAliasExprUsage(
 					from c in db.FromSql<int>($"select {1} {Sql.AliasExpr()}")
 					select c;
 
-			Assert.Throws<InvalidOperationException>(() => query.ToArray());
+			// Sql.AliasExpr() placed where it cannot become a table alias produces invalid SQL that the
+			// database rejects (there is no build-time validation for placeholder position in raw SQL).
+			Assert.Throws<Npgsql.PostgresException>(() => query.ToArray());
```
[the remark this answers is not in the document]
MaceWindu: Fixed in 38b39b8c5 — converted to `Should.Throw<Npgsql.PostgresException>(...)` for consistency with the file's Shouldly usage.

Source/LinqToDB/Data/DataConnection.QueryRunner.cs
```diff
@@ -248,17 +253,33 @@ static PreparedQuery GetCommand(DataConnection dataConnection, IQueryContext que
...
 					var serviceProvider = ((IInfrastructure<IServiceProvider>)dataConnection.DataProvider).Instance;
-					AliasesHelper.PrepareQueryAndAliases(serviceProvider.GetRequiredService<IIdentifierService>(), statement, query.Aliases, out var aliases);
-
-					query.Aliases = aliases;
+					// Alias fresh every execution. Parameter-dependent / continuous-run queries (the only ones
+					// that re-reach this point - non-parameter-dependent ones cache their commands above and
+					// short-circuit) rebuild the statement each execution via OptimizeAndConvertAll in Transform
+					// mode, producing fresh node identities. Fresh aliasing is deterministic for a given statement
+					// shape, so it is both stable across executions and identical to the remote rendering (which
+					// always aliases fresh server-side over a freshly deserialized statement).
```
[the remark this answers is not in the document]
MaceWindu: Fixed in cb601f4e3 — reworded the comment to state that non-parameter-dependent queries reach the aliasing block on their first execution (then cache their commands and short-circuit on later runs), while parameter-dependent / continuous-run queries re-reach it every execution.

igor-tkachev (commented):
Reviewed current state of the PR. Overall: the main aliasing fix looks correct to me.

The non-mutating aliasing model is the right direction for this bug: finalized table/column/field names are moved out of AST nodes into `AliasesContext`, parameter-dependent queries no longer reuse a previous object-identity-based alias context, direct path aliases after upfront prepare/convert, and remote path aliases the prepared statement that is actually rendered. That addresses the original shared cached statement race without introducing obvious direct/remote asymmetry.

What I checked:

- Removing `query.Aliases` reuse makes sense. With Transform-mode rebuilds, previous-run alias context is unsafe because node identities change, which explains the observed alias-numbering drift.
- SourceID-keyed table aliases are reasonable for correlated wrappers / CTE wrappers of the same logical source, assuming `SourceID` remains the identity of one logical source within a rendered statement.
- Nested builders inheriting `AliasesContext` is required and correctly handled. Without it, subqueries, CTE bodies and set-operation parts would fall back to raw node names.
- `BuildColumns` correctly reads aliases from the original aliased projection when `ConvertElement(selectQuery.Select)` returns rebuilt `SqlColumn` instances. The count check is a good fail-fast invariant.
- CTE field rendering through `AliasesContext.GetFieldName` looks correct after the `SqlField` -> `SqlCteField` refactor.
- `LINQ2DB0001` is a useful guardrail for the render path. Excluding `SqlTable.Alias` is acceptable as long as `SqlTable` alias remains stored on the node and `AliasesContext.GetTableAlias` falls back to it.

One thing I would like clarified before merge: root bare-field alias normalization in `AliasesHelper.GetColumnAliasSeed()`.

Currently, for any root `SqlField`, the final alias seed is forced to `field.PhysicalName`. This fixes the `ToSqlQuery -> Query<T>` entity round-trip for alias-forcing providers, but it also means explicit root projection aliases are silently dropped. The new test even pins this behavior:

```csharp
var query = from p in db.Person
            select new { Renamed = p.FirstName };

var command = query.ToSqlQuery();
command.Sql.ShouldNotContain("Renamed");
```

That is observable behavior of `ToSqlQuery`, not just an internal rendering detail. If a user explicitly writes `Renamed = p.FirstName`, it is reasonable to expect the generated SQL to preserve `Renamed` as the result column alias, especially when the SQL is later used as raw SQL for DTO/non-entity projection or diagnostics.

The comment says explicit member rename of a bare field is indistinguishable from an implicit member alias in the AST. That is true for some cases, but for `Renamed = p.FirstName` the alias value itself differs from the mapped member/field name, so it might be possible to restrict the physical-name normalization only to implicit member aliases needed for entity raw-SQL round-trip, instead of all root bare-field projections.

So my only concern is: is dropping explicit root projection aliases intentional public behavior? If yes, please call it out explicitly as accepted behavior in the PR description/release notes/test name. If not, I would prefer narrowing the normalization rule before merge.

I do not see a blocker in the core non-mutating aliasing design. I would merge after this behavior is clarified and the current head has a green full test run.

MaceWindu:
> So my only concern is: is dropping explicit root projection aliases intentional public behavior?

yes, it is interntional, see #4437

sdanyliv (dismissed):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed a concurrency bug where executing the same cached, parameter-dependent query on multiple threads at once could render corrupted SQL — table aliases and column names could be swapped or duplicated, surfacing as `42703 column does not exist` on PostgreSQL or as wrong results on other databases. Query alias finalization no longer mutates the shared cached statement, so a cached query is now safe to render concurrently.

Fixed raw-SQL / `ToSqlQuery()` entity materialization that maps result columns by name returning defaults or failing on providers that force aliases on the root SELECT (SqlCe, YDB, Access): the root SELECT now keeps each column's physical name instead of renaming it to the member name, and colliding root columns are given unique aliases (e.g. `PersonID` / `PersonID_1`).

Fixed the generated SQL differing between the direct and remote (LinqService) execution paths; table aliases are now uniquified per logical source rather than per table-source wrapper, so both paths produce identical aliasing.


Fixed corrupted SQL (swapped/duplicated table aliases and column names) when the same cached parameter-dependent query executed on multiple threads concurrently.


---
<sub>Generated from commit `407b021`.</sub>

PR #5658, opened 2026-06-26
Fix InvalidCastException in LoadWith/ThenLoad on non-linq2db IQueryable

### LINKED ISSUE #5649: Breaking change between linq2db v6.1.0 and 6.2.0 in LoadWith
### Description

There is a breaking change in LoadWith over IQuerable -- in v6.2.0 code `Enumerable.Empty<A>().AsQueryable().LoadWith(a => a.Prop)` throws `System.InvalidCastException: 'Unable to cast object of type 'System.Linq.EnumerableQuery1[A] to type LinqToDB.Internal.Linq.IExpressionQuery1[A]`

### Reproduction Steps

```C#
var t = Enumerable.Empty<A>().AsQueryable().LoadWith(a => a.Prop);

class A
{
    public B Prop { get; set; }
}

class B
{ }
```

This code works fine in 6.1.0 and throws InvalidCastException in 6.2.0

### Exception details

```
Unable to cast object of type 'System.Linq.EnumerableQuery`1[A]' to type 'LinqToDB.Internal.Linq.IExpressionQuery`1[A]'.

   at System.Runtime.CompilerServices.CastHelpers.ChkCastAny_NoCacheLookup(Void* toTypeHnd, Object obj)
   at LinqToDB.LinqExtensions.LoadWith[TEntity,TProperty](IQueryable`1 source, Expression`1 selector)
```

### Regression?

Yes

### Linq To DB version

6.2.0

### .NET version

10

### Database

_No response_

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5649

Since #5341 (v6.2.0), `LoadWith` required the source to be a linq2db `IExpressionQuery<T>`, causing an `InvalidCastException` when called on a plain `IQueryable` (e.g. `Enumerable.Empty<A>().AsQueryable()`).

## Changes

- Widened `LoadWithQueryableBase<TEntity>` and `LoadWithQueryable<TEntity,TProperty>` constructor parameters from `IExpressionQuery<TEntity>` to `IQueryable<TEntity>`. linq2db-specific members (`DataContext`, `GetSqlQueries`, `DebugView`) guard with `is IExpressionQuery<TEntity>` and throw a clear `LinqToDBException` if accessed without a linq2db source.
- All 9 `LoadWith`/`ThenLoad` overloads: when the source provider is not a linq2db `IQueryProviderAsync`, skip expression-tree injection and return a passthrough wrapper that delegates enumeration to the original source — mirrors EF Core `Include` behavior.
- Nine tests added, one per `LoadWith`/`ThenLoad` overload.

### DIFF
--- a/Source/LinqToDB/LinqExtensions/LinqExtensions.LoadWith.cs
+++ b/Source/LinqToDB/LinqExtensions/LinqExtensions.LoadWith.cs
@@ -8,6 +8,7 @@ using System.Threading;
 
 using JetBrains.Annotations;
 
+using LinqToDB.Internal.Async;
 using LinqToDB.Internal.Linq;
 
 namespace LinqToDB
@@ -61,23 +62,29 @@ namespace LinqToDB
 			return newTable;
 		}
 
-		abstract class LoadWithQueryableBase<TEntity>(IExpressionQuery<TEntity> query) : IExpressionQuery
+		abstract class LoadWithQueryableBase<TEntity>(IQueryable<TEntity> query) : IExpressionQuery
 		{
 			[DebuggerBrowsable(DebuggerBrowsableState.Never)]
-			public IExpressionQuery<TEntity> Query { get; } = query;
+			public IQueryable<TEntity> Query { get; } = query;
 
 			[DebuggerBrowsable(DebuggerBrowsableState.Never)]
-			Expression IExpressionQuery.Expression                                   => Query.Expression;
+			Expression IExpressionQuery.Expression => Query.Expression;
 
 			[DebuggerBrowsable(DebuggerBrowsableState.Never)]
-			IDataContext IExpressionQuery.DataContext                                  => ((IExpressionQuery)Query.GetLinqToDBSource()).DataContext;
+			IDataContext IExpressionQuery.DataContext =>
+				Query is IExpressionQuery<TEntity> exprQuery
+					? ((IExpressionQuery)exprQuery.GetLinqToDBSource()).DataContext
+					: throw new LinqToDBException($"LoadWith source '{Query.GetType()}' is not a linq2db query.");
 
 			public abstract QueryDebugView DebugView { get; }
 
-			IReadOnlyList<QuerySql> IExpressionQuery.GetSqlQueries(SqlGenerationOptions? options) => ((IExpressionQuery)Query.GetLinqToDBSource()).GetSqlQueries(options);
+			IReadOnlyList<QuerySql> IExpressionQuery.GetSqlQueries(SqlGenerationOptions? options) =>
+				Query is IExpressionQuery<TEntity> exprQuery
+					? ((IExpressionQuery)exprQuery.GetLinqToDBSource()).GetSqlQueries(options)
+					: throw new LinqToDBException($"LoadWith source '{Query.GetType()}' is not a linq2db query.");
 		}
 
-		sealed class LoadWithQueryable<TEntity, TProperty>(IExpressionQuery<TEntity> query) : LoadWithQueryableBase<TEntity>(query), ILoadWithQueryable<TEntity, TProperty>, IAsyncEnumerable<TEntity>
+		sealed class LoadWithQueryable<TEntity, TProperty>(IQueryable<TEntity> query) : LoadWithQueryableBase<TEntity>(query), ILoadWithQueryable<TEntity, TProperty>, IAsyncEnumerable<TEntity>
 		{
 			[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 			Type IQueryable.ElementType => Query.ElementType;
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/LinqExtensions/LinqExtensions.LoadWith.cs
MaceWindu, line 1: **Suggestion · SUG001** — When the source provider is not an `IQueryProviderAsync` (an EF Core query, or any third-party `IQueryable`), the new passthrough branch wraps the original source and discards the `selector`/`loadFunc` — the eager-load directive is silently dropped rather than throwing. This is the deliberate "mirror EF Core `Include`" intent and resolves the #5649 regression, but it changes the failure mode from a loud `InvalidCastException` to a silent no-op: a caller who mistakenly chains linq2db's `LoadWith` onto a foreign query (e.g. in provider-generic code) now gets results with the navigation un-loaded and no signal that the directive was ignored.

Fix: add a one-line note to the `LoadWith`/`ThenLoad` XML docs (and/or release notes) stating that on a non-linq2db source the call is a passthrough no-op, so the contract is discoverable.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Fix is correct and consistent. The `(IExpressionQuery<TEntity>)result` cast is removed, the wrapper ctors widened to `IQueryable<TEntity>`, and the `currentSource.Provider is IQueryProviderAsync ? … : currentSource` branch is applied identically across all 9 `LoadWith`/`ThenLoad` overloads — none silently misses the non-linq2db path. The linq2db-specific members (`DataContext`, `GetSqlQueries`, `DebugView`, async enumerator) all guard with `is IExpressionQuery<TEntity>` and throw `LinqToDBException`. A chained `ThenLoad` on a passthrough wrapper re-takes the non-async branch (the wrapper exposes the foreign provider via `Query.Provider`), so wrappers nest correctly down to the original source.
- [x] No public-API change. `LoadWithQueryableBase<TEntity>` and `LoadWithQueryable<TEntity,TProperty>` are `private` nested types, so the ctor widening is not part of the public surface.
- [x] Test strengthening applied in commit 0dbbb20. The 9 regression tests originally ran against an empty source (`Enumerable.Empty<...>().AsQueryable()`) and asserted only `Is.Empty`, which proves "no `InvalidCastException`" but not element passthrough — with zero elements a wrapper that dropped, duplicated, or mis-typed elements would still pass. The follow-up commit seeds the source with a real instance and asserts the element comes through unchanged (`Has.Count.EqualTo(1)` + `Is.SameAs`), so the tests now distinguish a correct passthrough from a broken one, including through a chained `ThenLoad`.

## Findings

One non-blocking suggestion is posted as a file comment (SUG001) on `Source/LinqToDB/LinqExtensions/LinqExtensions.LoadWith.cs`.

## Baselines
Baselines review skipped per user request (no baseline files in the diff).

Source/LinqToDB/LinqExtensions/LinqExtensions.LoadWith.cs
MaceWindu: Addressed in `d441e571` — added a `<remarks>` note to all 9 `LoadWith`/`ThenLoad` overloads stating that on a non-linq2db source the eager-load directive is ignored and the query is returned unchanged as a passthrough. The contract is now discoverable from IntelliSense.

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review, so this complements — rather than replaces — human judgement. Individual findings can be wrong, overconfident, or miss context a human would catch. Feel free to disagree, dismiss, or ask for clarification on any finding.

## Review notes
- [x] Re-reviewed at HEAD. The #5649 fix is correct and complete for the reported use case (a plain mocked `IQueryable` with no EF Core bridge active): `ProcessIQueryable()` is a no-op when the source hook is unset, the `is IQueryProviderAsync` discriminator then selects the passthrough wrapper, and enumeration delegates to the original source. Verified across all 9 `LoadWith`/`ThenLoad` overloads.

## Follow-up commits
Three small follow-ups pushed to this branch:
- `ff994bb` — adds a regression test for the async-enumeration guard: async enumeration of a passthrough wrapper over a non-linq2db source must throw `LinqToDBException`. The 9 existing tests only cover synchronous `.ToList()` passthrough, leaving the `GetAsyncEnumerator` throw path uncovered.
- `337ba61` — extracts the `"… is not a linq2db query."` guard message, repeated verbatim in `DataContext`, `GetSqlQueries`, and `DebugView`, into a single `NotLinqToDbSource()` helper, so a future wording change touches one site instead of three.
- `d441e571` — documents the non-linq2db passthrough no-op in the `<remarks>` of all 9 `LoadWith`/`ThenLoad` overloads, so the contract (the eager-load directive is silently ignored on a foreign source) is discoverable from IntelliSense. Addresses the earlier SUG001 thread.

## Out-of-scope observations
Surfaced during review but outside this PR's scope. FYI only.
- **EF-Core-bridge-loaded passthrough is gated by the EF hook, not this code.** When `LinqToDB.EntityFrameworkCore` is initialized, `ProcessIQueryable()` routes through the EF `ProcessSourceQueryable` hook, which throws `LinqToDBForEFToolsException` for a plain (non-EF) `IQueryable` before this PR's passthrough branch runs. This is pre-existing, arguably-intended bridge behavior and does not affect the reported #5649 use case (which runs without the bridge).

## Baselines
No baseline changes — the baselines branch carries an empty delta, consistent with no generated-SQL change.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed an `InvalidCastException` (a regression since 6.2.0) when `LoadWith` / `ThenLoad` was called on a plain `IQueryable` that isn't a linq2db query (e.g. `Enumerable.Empty<T>().AsQueryable()`). Synchronous enumeration of such a source now passes through to the original query (mirroring EF Core `Include`); async enumeration — and accessing the linq2db `DataContext`, `GetSqlQueries`, or the debug SQL view on it — now raises a clear `LinqToDBException` rather than the previous `InvalidCastException`.


_(none)_


---
<sub>Generated from commit `d441e57`.</sub>

Issue #1974, opened 2019-11-11
Association CanBeNull changed behaviour after update to 2.9.1


Franki1986:
Current working version:  2.9.3 and 3.0
version testet that brings a change: 2.9.1

After version 2.9.1 an Association with CanBeNull does not return null for associated Property:

```
public class Person{
    public int ID {get; set;}   
    public string Name {get; set;}
    [Association(QueryExpressionMethod = nameof(ArticleExpr), CanBeNull = true, Relationship = Relationship.OneToOne)]
    public Artivle Bought {get; set}

public static Expression<Func<Person, IDataContext, IQueryable<Article>>> ArticleExpr()
      {
         return (p, db) => db.GetTable<Article>().Where(a => a.PersonId == c.Id);
      }
}

public class Article{
    public string ID {get; set;}  
    public int PersonId {get; set;}  
    public double Price {get; set;}  
}
```

Franki1986:
Changed the title, because 'broken' is my interpretation..

Franki1986:
Any way to work around this? 
I have many situations where this does not work anymore, because the result is not null.

sdanyliv:
How to reproduce? No linq queries provided.

Franki1986:
public class Person{
    public int ID {get; set;}   
    public string Name {get; set;}
    [Association(QueryExpressionMethod = nameof(ArticleExpr), CanBeNull = true, Relationship = Relationship.OneToOne)]
    public Article Bought {get; set}

```
public static Expression<Func<Person, IDataContext, IQueryable<Article>>> ArticleExpr()
      {
         return (p, db) => db.GetTable<Article>().Where(a => a.PersonId == c.Id);
      }
}

public class Article{
    public string ID {get; set;}  
    public int PersonId {get; set;}  
    public double Price {get; set;}  
}
```


using(var db = new MydataContext()){
    db.GetTable<Person>().LoadWith(p => p.Bought).ToList();
}

if no Article is found (property Bought) in previous versions it was null, now it is initialized with default property values.

sdanyliv:
Ok, got it.

PR #5659, opened 2026-06-26
Honor [Column]/[ValueConverter] in raw-SQL constructor materialization (#4437)

### LINKED ISSUE #4437: Inconsistent object mapping behavior when querying using LINQ  vs `QueryToListAsync`

MaceWindu:
Fixes #4437

## Problem

`Query<T>(string sql)` materialized result columns onto positional-record / constructor parameters by **parameter name only**, ignoring the column mapping. So a type whose member name differs from its DB column —

```csharp
[Table("test4437")]
sealed record Issue4437Record([property: Column("some_column")] string SomeColumn);

db.Query<Issue4437Record>("select some_column from test4437"); // SomeColumn == null
```

— came back `null`/default unless the raw SQL aliased the column to the member name (`... as SomeColumn`). The LINQ path (`GetTable<T>()`) and the property/field materialization path were already correct; only the raw-SQL **constructor** path was wrong. A `[ValueConverter]` on a constructor-bound column was also never applied (the converter argument was hard-coded `null`).

## Fix

In `CommandInfo.CreateObjectReader<T>` (raw-SQL constructor path):

- Resolve each constructor parameter to its `ColumnDescriptor` by member name (ordinal, then case-insensitive — mirroring `EntityConstructorBase.MatchParameter` on the LINQ path).
- Match a result column by the mapped `ColumnName` **or** the parameter name, so both `select some_column` and `select some_column as SomeColumn` work.
- Forward the column's `ValueConverter` into the member-read expression.

## Tests

`MappingTests` (SQLite, all configs):

- `Issue4437Test2` — raw SQL, no alias — un-gated from `[ActiveIssue]` (now passes).
- `Issue4437Test4` — new — positional record with `[ValueConverter]` queried via raw SQL.
- `Issue4437Test1` (LINQ) and `Issue4437Test3` (raw SQL aliased) continue to pass — Test3 confirms the column-name change didn't regress the alias case.

## Scope note

This touches shared raw-SQL materialization (`CommandInfo`), used by every provider. The change is additive — the parameter-name fallback preserves all previously-working matches — but a full CI run is warranted. No public-API surface change.

## Follow-up commits

- `9ce67fcbd` — clarify the `GetColumn` comment: the cited `EntityConstructorBase.MatchParameter` also applies a member-type check that `GetColumn` intentionally omits (a no-op for positional records, where a parameter and its mapped member always share a type). No behavior change.
- `07e0b8f86` — document the column-resolution precedence (mapped `ColumnName` wins, parameter name kept as an alias fallback) and the multi-table first-match caveat in code. No behavior change.

## Follow-up: unified with the LINQ path

The constructor path is no longer a bespoke reimplementation — `CreateObjectReader` now delegates all non-scalar materialization to `RecordReaderBuilder` (`EntityConstructorBase`), the same builder `GetTable<T>()` uses. Consequences vs. the description above:

- The **mapped column name is authoritative**; the parameter-name alias fallback is **removed**. `select some_column as SomeColumn` no longer binds — select the mapped column name (`select some_column`). `Issue4437Test3` is updated to assert this.
- `RecordReaderBuilder` now binds a typed default for a column absent from the result set, and reduces reads in fast mode (no slow-mode dispatch for raw SQL; the inheritance / complex-column case is faster too).
- Supersedes the per-comment follow-up commits listed above.

### DIFF
--- a/Source/LinqToDB/Data/CommandInfo.cs
+++ b/Source/LinqToDB/Data/CommandInfo.cs
@@ -1926,11 +1926,12 @@ namespace LinqToDB.Data
 
 				if (ctors.Count > 0 && ctors.TrueForAll(c => c.ps.Length > 0))
 				{
-					// Resolve each ctor parameter to its mapped column (parameter name == member name, ordinal then
-					// case-insensitive — mirrors EntityConstructorBase.MatchParameter on the LINQ path), then match a
-					// result column by the mapped ColumnName OR the parameter name and convert via the column's
-					// ValueConverter, so [Column]/[ValueConverter] are honored for raw SQL whether the query uses the
-					// DB column name or aliases it to the member name (#4437).
+					// Resolve each ctor parameter to its mapped column by name (parameter name == member name, ordinal
+					// then case-insensitive — the name-matching of EntityConstructorBase.MatchParameter on the LINQ
+					// path; its additional member-type check is omitted, as positional-record params and their members
+					// always share a type). Then match a result column by the mapped ColumnName OR the parameter name
+					// and convert via the column's ValueConverter, so [Column]/[ValueConverter] are honored for raw SQL
+					// whether the query uses the DB column name or aliases it to the member name (#4437).
 					ColumnDescriptor? GetColumn(ParameterInfo p)
 						=> td.Columns.FirstOrDefault(c => string.Equals(c.MemberName, p.Name, StringComparison.Ordinal))
 						?? td.Columns.FirstOrDefault(c => string.Equals(c.MemberName, p.Name, StringComparison.OrdinalIgnoreCase));
--- a/Source/LinqToDB/Data/CommandInfo.cs
+++ b/Source/LinqToDB/Data/CommandInfo.cs
@@ -1926,9 +1926,38 @@ namespace LinqToDB.Data
 
 				if (ctors.Count > 0 && ctors.TrueForAll(c => c.ps.Length > 0))
 				{
+					// Resolve each ctor parameter to its mapped column (parameter name == member name, ordinal then
+					// case-insensitive — mirrors EntityConstructorBase.MatchParameter on the LINQ path), then match a
+					// result column by the mapped ColumnName OR the parameter name and convert via the column's
+					// ValueConverter, so [Column]/[ValueConverter] are honored for raw SQL whether the query uses the
+					// DB column name or aliases it to the member name (#4437).
+					ColumnDescriptor? GetColumn(ParameterInfo p)
+						=> td.Columns.FirstOrDefault(c => string.Equals(c.MemberName, p.Name, StringComparison.Ordinal))
+						?? td.Columns.FirstOrDefault(c => string.Equals(c.MemberName, p.Name, StringComparison.OrdinalIgnoreCase));
+
+					int MatchIndex(ParameterInfo p, ColumnDescriptor? column)
+					{
+						var comparer = dataConnection.MappingSchema.ColumnNameComparer;
+						var idx      = -1;
+
+						if (column != null)
+						{
+							var columnName = column.ColumnName;
+							idx = Array.FindIndex(names, n => comparer.Compare(n, columnName) == 0);
+						}
+
+						if (idx < 0)
+						{
+							var name = p.Name;
+							idx = Array.FindIndex(names, n => comparer.Compare(n, name) == 0);
+						}
+
+						return idx;
+					}
+
 					var q =
 						from c in ctors
-						let count = c.ps.Count(p => names.Contains(p.Name, dataConnection.MappingSchema.ColumnNameComparer))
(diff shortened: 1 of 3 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Raw SQL queries materialized into a record or constructor-based type — `Query<T>(string)`, `QueryToList` / `QueryToListAsync`, etc. — now bind result columns by the same rules as LINQ queries: a member's `[Column("some_column")]` name is authoritative and a `[ValueConverter]` on a constructor-bound column is applied. Previously a constructor / positional-record parameter was matched to a result column **by parameter name only**, so a `[Column("some_column")]`-mapped member came back `null`/default and its `[ValueConverter]` was never applied. ⚠ Because binding now follows the mapped column name, the previous workaround of aliasing a column to the member name (`select some_column as SomeColumn`) no longer binds — select the column under its mapped name instead. Duplicate or empty column names in a raw result set are tolerated.


Raw-SQL materialization into records/constructor types now honors `[Column]` mapping and `[ValueConverter]`, binding by the mapped column name rather than the parameter name.


---
<sub>Generated from commit `5788bd7`.</sub>

MaceWindu (commented):
> [!IMPORTANT]

## Review notes
- [x] Scope confirmed — honors `[Column]` / `[ValueConverter]` in the raw-SQL constructor materialization path (`CommandInfo.CreateObjectReader<T>`), fixing #4437. The fix is consistent with the property/field (`MemberInit`) path, which already forwarded the column's `ValueConverter`; no double-conversion or null-safety risk. The un-gated `Issue4437Test2` is the red→green proof, and `Issue4437Test4` covers the positional-record-plus-`[ValueConverter]` case.
- [x] CI green — `build` and `test-all` both pass on the PR head.
- [x] No findings to assert. Two follow-up commits were pushed during review: `9ce67fcbd` clarifies the `GetColumn` comment (the cited `EntityConstructorBase.MatchParameter` also applies a member-type check that `GetColumn` intentionally omits — a no-op for positional records, where a parameter and its mapped member always share a type); `07e0b8f86` documents the column-resolution precedence and the multi-table first-match caveat in code. Both comment-only, no behavior change.
- [x] Behavior change for release notes — raw-SQL constructor mapping now resolves a result column by the mapped `[Column]` name (the member-name-mapping match wins), with the previous parameter-name match kept as an alias fallback. Worth a 6.4.0 release-notes entry, including the multi-table same-column-name first-match caveat the maintainer raised on #4437.

## Baselines
Delta: [linq2db.baselines PR #1974](https://github.com/linq2db/linq2db.baselines/pull/1974) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5659) — **9 added, 0 modified, 0 deleted** (verified against a fresh three-dot diff against the merge-base).

### New SQLite baselines for Issue4437 tests
- **`Issue4437Test2`** (un-gated) — `select some_column from test4437`, across 4 SQLite variants (Classic, Classic.MPM, Classic.MPU, MS). Trivial single-table pass-through; shape correct.
- **`Issue4437Test4`** (new) — `select some_column from test4437vc`, across the same 4 SQLite variants. Shape correct.

### Cross-provider anomalies
- `AnalyticTests.TestVarPopOracle(Oracle.23.Managed)` — a new `.sql.other` (direct-vs-remote SQL mismatch) artifact of the known flaky analytic-aggregate pattern (#5513). The PR touches no Oracle or analytic code; not caused by this PR.

sdanyliv (dismissed):

PR #5660, opened 2026-06-27
Unify raw-SQL entity mapping with the LINQ path (#4437)

### LINKED ISSUE #4437: Inconsistent object mapping behavior when querying using LINQ  vs `QueryToListAsync`

MaceWindu:
Fixes #4437.

Routes all entity types in raw-SQL queries (`Query<T>` / `QueryToList`) through `RecordReaderBuilder`, so they map reader fields by `[Column]` metadata and select constructors with the same rules as LINQ queries / `GetTable<T>` — instead of the old behavior of matching constructor parameters by reader field name.

- `CommandInfo.CreateObjectReader`: removed the bespoke field-name ctor matcher; all entity types now use `RecordReaderBuilder` (previously only inheritance / complex-column types did).
- `RecordReaderBuilder`: reader-index map now keeps the first occurrence of a duplicated field name (multi-table `SELECT t1.f, t2.f`) instead of throwing.
- Regression tests in `DataExtensionsTests`.

## ⚠️ Breaking change
Raw-SQL result columns must now match the `[Column]` name (the database column name), **not** the C# member name. The previously-recommended workaround `SELECT some_column AS SomeColumn` no longer maps — use `SELECT some_column`. This makes raw-SQL mapping consistent with LINQ.

### DIFF
--- a/Source/LinqToDB/Data/CommandInfo.cs
+++ b/Source/LinqToDB/Data/CommandInfo.cs
@@ -1907,71 +1907,11 @@ namespace LinqToDB.Data
 			}
 			else
 			{
-				var td = dataConnection.MappingSchema.GetEntityDescriptor(typeof(T), dataConnection.Options.ConnectionOptions.OnEntityDescriptorCreated);
-
-				if (td.InheritanceMapping.Count > 0 || td.HasComplexColumns)
-				{
-					var    readerBuilder = new RecordReaderBuilder(dataConnection, typeof(T), dataReader, converterExpr);
-					return readerBuilder.BuildReaderFunction<T>();
-				}
-
-				var names = new string[dataReader.FieldCount];
-
-				for (var i = 0; i < dataReader.FieldCount; i++)
-					names[i] = dataReader.GetName(i);
-
-				expr = null;
-
-				var ctors = typeof(T).GetConstructors().Select(c => new { c, ps = c.GetParameters() }).ToList();
-
-				if (ctors.Count > 0 && ctors.TrueForAll(c => c.ps.Length > 0))
-				{
-					var q =
-						from c in ctors
-						let count = c.ps.Count(p => names.Contains(p.Name, dataConnection.MappingSchema.ColumnNameComparer))
-						orderby count descending
-						select c;
-
-					var ctor = q.FirstOrDefault();
-
-					if (ctor != null)
-					{
-						expr = Expression.New(
-							ctor.c,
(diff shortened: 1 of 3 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

##### ⚠ Breaking changes
- Raw SQL queries (`Query<T>`, `QueryToList`, etc.) now map result columns to entity members by the `[Column]` name (the database column name) and select constructors using the same rules as LINQ queries — instead of matching by the C# member/parameter name. Records with a primary constructor and `[Column]` attributes now map correctly from raw SQL. If you previously aliased a column to the C# member name (`SELECT some_column AS SomeColumn`) as a workaround, drop the alias — the result column must match the `[Column]` name. ([#4437](https://github.com/linq2db/linq2db/issues/4437))


Raw-SQL entity mapping (`Query<T>`) now uses `[Column]` names and LINQ-style constructor selection — records with primary constructors map correctly. ⚠ Breaking: column aliases that matched C# member names are no longer needed.


---
<sub>Generated from commit `79b73c0`.</sub>

MaceWindu:
Superseded by #5659 (already approved, non-breaking, and also fixes [ValueConverter] on constructor-bound columns). Closing this duplicate.

PR #5662, opened 2026-06-27
Self-heal baselines branch so test-job restarts work

MaceWindu:
## Problem

When a test run produces no baseline changes, `create_baselines_pr` deletes the
per-PR baselines branch — correct for a completed run, but it breaks **partial
restarts**. On *"rerun failed jobs"* (e.g. a flaky failure, build 21555), Azure does
not re-run the already-succeeded `create_baselines_branch` job, so its output
variables still point at `baselines/pr_<id>` — but that branch was already deleted.
The restarted test job's `git checkout -b … origin/$(baselines_branch)` then fails
and the restart never runs.

## Fix

Extract the "ensure baselines branch exists" logic into a reusable
`Build/Azure/scripts/ensure-baselines-branch.ps1`, shared by both callers:

- **Test jobs** call it (ensure-only) before their existing clone — re-creating the
  branch at the recorded `baselines_head` hash if a previous run deleted it.
- **`create_baselines_branch`** calls it (`-Rebase -EmitOutputs`) in place of its
  inline body — same create + rebase + output-variable behaviour as before.

Branch creation is race-tolerant (parallel test jobs re-query rather than fail on a
concurrent create). No change to `create_baselines_pr`; the happy path is unchanged
(the ensure step is a no-op `ls-remote` when the branch exists).

## Verification

- YAML validated (`yaml.safe_load`) on all four edited templates; script passes the
  PowerShell parser.
- End-to-end is CI-only (`BASELINES_GH_PAT`): a no-baseline run creates+deletes the
  branch as before; rerunning a test job then re-creates the branch and completes.

### DIFF
--- a/Build/Azure/pipelines/templates/test-jobs.yml
+++ b/Build/Azure/pipelines/templates/test-jobs.yml
@@ -21,102 +21,16 @@ jobs:
 
   steps:
   - checkout: none
+  - task: DownloadPipelineArtifact@2
+    inputs:
+      artifactName: $(artifact_test_scripts)
+      targetPath: '$(System.DefaultWorkingDirectory)/scripts'
+    displayName: Extract test scripts
   - task: PowerShell@2
     inputs:
-      targetType: 'inline'
-      script: |
-        $prId = "$(source_pr_id)"
-        $orgName = "linq2db"
-        $baselinesRepo = "linq2db.baselines"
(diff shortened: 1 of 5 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `37cbe39`.</sub>

PR #5665, opened 2026-06-30
Fix non-deterministic VALUES key order in keyed eager-load (direct vs remote SQL divergence)

### LINKED ISSUE #5664: Keyed eager-load inlines master keys as VALUES in non-deterministic order → direct vs remote SQL divergence
### Summary

When an eager load uses the keyed strategy (`WithKeyedLoadStrategy()`) and the master query has no stable `ORDER BY`, the master keys are collected and pasted into the detail query as an inline `VALUES` table. Those rows come out in whatever order the master query happened to return them, and that order isn't the same between a direct connection and a remote (LinqService) one. So the same LINQ query generates different SQL text on the two paths.

### Current behavior

The keyed strategy takes the distinct master keys and inlines them as a `VALUES` table for the detail query:

- `Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoad.cs:316` — `SelectDistinct` over the main keys. That set gets materialized on the client and written out as `(VALUES …) k_1(Item1, Item2)` rows in iteration order, with nothing sorting them first.

This shows up on the remote YDB leg of PR #5468, in `EagerLoadingStrategyKeyedQueryTests.Concat_KeyedQuery_EagerLoadDifferentDetails`. The master query there has no `ORDER BY`, and the detail SQL looks like:

```sql
(VALUES
    (102,'Inactive'u), (202,'Inactive'u), (204,'Inactive'u),
    ...
    (101,'Active'u), (103,'Active'u), ...
) k_1(Item1, Item2)
    INNER JOIN Employee d ON k_1.Item1 = d.DepartmentId
```

Direct and remote runs share one baseline file (`Tests/Base/BaselinesWriter.cs:66-87`): the direct run writes the `.sql`, then the remote run regenerates its SQL and compares. On a mismatch it dumps a `.sql.other` and throws *"Baselines for remote context doesn't match direct access baselines."* Since the VALUES rows come out in a different order on each path, the remote run fails.

### Expected behavior

The key VALUES should be emitted in a deterministic order that doesn't depend on how the master query returned its rows or which path executed it, so direct and remote generate identical SQL.

### Notes

- The query still returns the right data — the detail query matches keys by membership and has its own `ORDER BY d.Id`; only the generated SQL text drifts. So this is about SQL-generation determinism, not wrong results.
- Not reproduced end to end. The two confirmed mechanics are: baselines are shared between direct and remote, and the VALUES rows aren't ordered before emission. The actual `.sql.other` from the failing remote-YDB run was not captured.
- Not really YDB-specific — YDB just makes it visible because it inlines the keys as a literal VALUES list.

MaceWindu:
Fixes #5664.

### Problem

The KeyedQuery eager-loading strategy inlines collected master keys as a `(VALUES …)` table in the detail queries. The key array order depended on which extraction path produced it (client-side buffer vs SQL `SelectDistinct`) and on the order the database returned rows — neither stable across the direct ADO and remote (LinqService) paths. The same LINQ query could thus emit textually different SQL on the two paths, surfacing as the intermittent *"Baselines for remote context doesn't match direct access baselines"* failure (reproduced on YDB, ~27% of runs).

### Fix

Canonicalize the key order once at the single choke point both paths funnel through — `KeyedQueryKeysHolder.SetKeys` — via a guarded sort:

- `Comparer<TKey>.Default` so the inlined `VALUES` rows are deterministic regardless of extraction path or DB row order.
- Falls back to source order (clone-then-sort, no partial reorder) when `TKey` isn't orderable (e.g. a composite with a `byte[]` component), preserving eager load's equality-only / structural-key contract.
- Keeps the buffer optimization intact — no extra key-extraction round-trip.

### Verification

Reproduced on YDB (direct vs remote, ~4/15 runs), then green across 20/20 consecutive runs of `EagerLoadingStrategyKeyedQueryTests.Concat_KeyedQuery_EagerLoadDifferentDetails`.

### Baselines

Changes the `VALUES` row order (→ sorted by key) for every provider's keyed eager-load baseline that inlines multi-row keys. Baselines will be regenerated by CI.

## Follow-up commit (0681973)

`Comparer<TKey>.Default` sorted string keys (and string components of composite keys) with `string.CompareTo`, a culture-sensitive linguistic comparison — neither a stable total order over distinct strings (canonically-equal but distinct values tie, and `Array.Sort` is not stable) nor machine-culture-independent, so direct and remote could still diverge and string-keyed baselines could reorder across cultures. Switched to an ordinal, culture-invariant comparer: `StringComparer.Ordinal` for bare string keys, element-wise ordinal comparison via `IStructuralComparable` for composite keys at any nesting depth; `byte[]`/array key components keep the not-orderable source-order fallback. Adds a SQLite regression test asserting the inlined string keys appear in ordinal order.

### DIFF
--- a/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
@@ -707,7 +707,7 @@ namespace LinqToDB.Internal.Linq.Builder
 				{
 					// Sort in place — every caller passes a freshly-allocated array (.ToArray()) that
 					// nothing else holds, so mutating it is safe.
-					Array.Sort(keys, Comparer<TKey>.Default);
+					Array.Sort(keys, _keyComparer);
 				}
 				catch (InvalidOperationException)
 				{
@@ -719,6 +719,60 @@ namespace LinqToDB.Internal.Linq.Builder
 				}
 
 				return keys;
+			}
+
+			// Ordinal, culture-invariant key comparer. Comparer<TKey>.Default orders string keys (and
+			// string components of composite ValueTuple keys) with string.CompareTo, i.e. a culture-sensitive
+			// linguistic comparison. That is unsuitable for canonicalizing SQL text: it is not a stable total
+			// order over distinct strings (canonically-equal but distinct values — e.g. "å" vs "å"
+			// — tie, and Array.Sort is not stable, so the two extraction paths could still order them
+			// differently), and it is machine-culture-dependent (baselines would reorder across cultures).
+			// Both defeat the deterministic ordering this canonicalization exists to provide (#5664), so
+			// strings are compared ordinally.
+			static readonly IComparer<TKey> _keyComparer = CreateKeyComparer();
+
+			static IComparer<TKey> CreateKeyComparer()
+			{
+				if (typeof(TKey) == typeof(string))
+					return (IComparer<TKey>)(object)StringComparer.Ordinal;
+
+				// Composite key (ValueTuple implements IStructuralComparable) that may contain string
+				// components: compare element-wise with ordinal string handling. Arrays also implement
+				// IStructuralComparable but are excluded — a byte[] key keeps the not-orderable fallback,
+				// as before. Non-tuple, non-string keys use the default order, which is already
+				// culture-invariant and a total order for the scalar types used as keys.
+				if (typeof(IStructuralComparable).IsAssignableFrom(typeof(TKey)) && !typeof(TKey).IsArray)
+					return new OrdinalKeyComparer();
+
+				return Comparer<TKey>.Default;
+			}
+
+			// IComparer (non-generic) is implemented so IStructuralComparable.CompareTo can drive the
+			// element-wise comparison recursively — it passes each component pair back through Compare.
+			sealed class OrdinalKeyComparer : IComparer<TKey>, IComparer
+			{
+				public int Compare(TKey? x, TKey? y) => CompareElement(x, y);
+
+				int IComparer.Compare(object? x, object? y) => CompareElement(x, y);
+
+				int CompareElement(object? a, object? b)
+				{
+					if (a is null) return b is null ? 0 : -1;
+					if (b is null) return 1;
+
+					if (a is string sa && b is string sb)
+						return string.CompareOrdinal(sa, sb);
+
+					// Recurse into nested tuples (ValueTuple/Tuple implement IStructuralComparable) so
+					// string components at any depth compare ordinally. Arrays are excluded: a byte[]
+					// component is left to Comparer<object>.Default, which throws — Array.Sort wraps that
+					// as InvalidOperationException, which SortKeysDeterministically catches to fall back to
+					// source order, preserving the prior behaviour for non-orderable key classes.
+					if (a is IStructuralComparable structural and not Array)
+						return structural.CompareTo(b, this);
+
+					return Comparer<object>.Default.Compare(a, b);
+				}
 			}
 
 			// execution is null only if the keys accessor is evaluated without a live IQueryExpressions
--- a/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
@@ -686,8 +686,35 @@ namespace LinqToDB.Internal.Linq.Builder
 				if (execution == null)
 					return;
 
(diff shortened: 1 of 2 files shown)


Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
```diff
@@ -686,10 +686,37 @@ public void SetKeys(object? execution, TKey[] keys)
...
 			}
 
+			static TKey[] SortKeysDeterministically(TKey[] keys)
+			{
+				if (keys.Length < 2)
+					return keys;
+
+				try
+				{
+					var sorted = (TKey[])keys.Clone();
```
sdanyliv: Do we really need Clone?

MaceWindu: protection from keys changed by Sort before exception thrown. Probably not needed

MaceWindu: Removed in 79942fb. To answer the underlying question: yes, `Array.Sort` *can* leave the array partially reordered before throwing — for a composite key like `(int, byte[])`, `ValueTuple.CompareTo` compares the leading `int` first (doing swaps) and only throws later when two keys share an equal `int` and it falls through to the `byte[]`, and the catch doesn't roll those swaps back. But it's harmless here: that's the non-orderable key class whose order is already non-deterministic between the direct and remote paths, and `Sort` only reorders — never drops — elements. The clone was an O(n) allocation on the common (orderable) path guarding an outcome that's a no-op for correctness, so it's gone; every `SetKeys` caller passes a fresh `.ToArray()` array, so the in-place sort is safe.

sdanyliv (dismissed):

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.EagerLoadKeyedQuery.cs
```diff
@@ -686,10 +686,95 @@ public void SetKeys(object? execution, TKey[] keys)
...
+				{
+					// TKey (or one of its composite components) does not implement IComparable.
+					// Such keys are not deterministically orderable, so they keep source order — eager
+					// load compares keys structurally for equality only, no ordering contract is required.
+					// A partial Array.Sort here is harmless: it only reorders (never drops) elements, and
+					// the order of a non-orderable key class is non-deterministic between paths regardless.
+				}
+
+				return keys;
+			}
```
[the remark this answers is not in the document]
MaceWindu: Not a defect. The clone was deliberately removed in 79942fb: every `SetKeys` caller passes a freshly-allocated `.ToArray()` that nothing else holds, so the in-place sort is safe. A partial `Array.Sort` reorder on throw is reachable only for non-orderable key classes (e.g. a composite with a `byte[]` component), which were already non-deterministic between the direct and remote paths before this PR — so it introduces no new divergence, and `Array.Sort` only permutes (never drops) elements, leaving the key set valid for structural-equality matching. `ArgumentException` handling isn't needed: `Array.Sort` wraps all comparer faults (including the `ArgumentException` from `Comparer<object>.Default` on `byte[]`) as `InvalidOperationException` on every target TFM, which the `catch` already covers.

Tests/Linq/Linq/EagerLoadingStrategyKeyedQueryTests.cs
```diff
@@ -2057,6 +2057,94 @@ from p in tP
...
+			{
+				new StringKeyChild { Id = 1, ParentCode = "zza", Value = "a1" },
+				new StringKeyChild { Id = 2, ParentCode = "zzB", Value = "B1" },
+			};
+
+			var collector = new CommandTextCollector();
+
+			using var db = GetDataContext(context);
+			db.AddInterceptor(collector);
+
```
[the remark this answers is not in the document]
MaceWindu: The keyed eager-load inlines master keys into the detail query's VALUES table as `SqlValue` literals, not `SqlParameter`s, so they render as literals independent of `InlineParameters`. The generated SQLite baseline for this test confirms `VALUES ('zzB'), ('zza')`. The test is `[IncludeDataSources(false, TestProvName.AllSQLite)]`-scoped, so the `CommandInterceptor` captures the literal SQL on the direct path directly — `InlineParameters = true` isn't needed and the `Contains("'zzB'")` / `Contains("'zza'")` assertions hold as written.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_Omitted: intra-release fix to the KeyedQuery eager-loading strategy introduced this same release (#5450) — no user-observable regression from a prior version._


_(none)_


---
<sub>Generated from commit `767bdeb`.</sub>

PR #5667, opened 2026-06-30
Fix #5666: enum-vs-null comparison emits IS [NOT] NULL instead of <> 0

### LINKED ISSUE #5666: Wrong association SQL is generated when enum property is not nullable
### Description

When enum property is not nullable, linq2db adds extra unnecessary expression to association expression.

### Reproduction Steps

```cs
  public enum Enum1
  {
    Ok,
    Fail,
  }

  public class ReceiptBaseTest
  {
    [Column("RECEIPT_NO")]            public string           ReceiptNo             { get; set; }
    [Column("SERV_CATG")]             public Enum1            Enum1               { get; set; }
    [Column("PROPERTY_ID")]           public int?             PropertyId            { get; set; }

    [Association(ThisKey=nameof(PropertyId), OtherKey=nameof(EnumTest.Id))]
    public EnumTest EnumTest { get; }
  }

  public class EnumTest
  {
    public int Id { get; set; }
  }

    static void TestEnum()
    {
      using var dc = new Dc();

      var query = dc.GetTable<ReceiptBaseTest>()
        .TableName("vwCASHIERS_DET")
        .AsQueryable();

      query = query.Where(i => i.EnumTest.Id == 123);

      var items = query
        .Select(
          i =>
            new
            {
              i.ReceiptNo,
            })
        .ToList();
    }

```
Generated sql:
```sql
SELECT
	[i].[RECEIPT_NO]
FROM
	[vwCASHIERS_DET] [i]
		INNER JOIN [EnumTest] [a_EnumTest] ON [i].[SERV_CATG] <> 0 AND [i].[PROPERTY_ID] = [a_EnumTest].[Id]
WHERE
	[a_EnumTest].[Id] = 123
```
Expected sql:
```sql
SELECT
	[i].[RECEIPT_NO]
FROM
	[vwCASHIERS_DET] [i]
		INNER JOIN [EnumTest] [a_EnumTest] ON [i].[PROPERTY_ID] = [a_EnumTest].[Id]
WHERE
	[a_EnumTest].[Id] = 123
```

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.4.0.0

### .NET version

4.8

### Database

SqlServer 2022

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5666.

## Problem

For an optional association, the builder adds a row-present guard built from the parent's first non-nullable key column (`ExtractNotNullCheck`), i.e. `column != null`. When that column is an enum, `ConvertEnumConversion` mapped the `null` constant to the enum's underlying default (`0`), emitting `column <> 0` instead of `column IS NOT NULL`. As a value filter, `<> 0` polluted the join ON clause and silently dropped rows whose enum equals its zero-valued member.

Association INNER JOIN ON clause for the repro:

```diff
- ON [i].[SERV_CATG] <> 0 AND [i].[PROPERTY_ID] = [a_EnumTest].[Id]
+ ON [i].[PROPERTY_ID] = [a_EnumTest].[Id]
```

## Fix

`ConvertEnumConversion` now treats a `null` constant as a null check and emits `IS [NOT] NULL`, consistent with non-enum column handling. This also fixes user-written `enumColumn == null` / `enumColumn != null`.

## Tests

- `Tests/Linq/UserTests/Issue5666Tests.cs` — behavioral + SQL-shape; fails before, passes after.
- Regression: 1,149 enum-mapping + association tests pass (SQLite Classic + MS).

### DIFF
--- a/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
+++ b/Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
@@ -4655,15 +4655,25 @@ namespace LinqToDB.Internal.Linq.Builder
 				case ExpressionType.Constant:
 				{
 					var origValue = ((ConstantExpression)value).Value;
+
+					// Comparing an enum to null is a null check, not a value comparison. Mapping null to the
+					// enum's underlying default (e.g. 0) would pollute the SQL and turn the check into a value
+					// filter that drops rows whose enum equals that default — see issue #5666. Emit IS [NOT] NULL.
+					if (origValue == null)
+					{
+						var operandSql = (Visit(operand) as SqlPlaceholderExpression)?.Sql;
+						if (operandSql == null)
+							return null;
+
+						return new SqlPredicate.IsNull(operandSql, op == SqlPredicate.Operator.NotEqual);
+					}
+
 					var mapValue  = origValue;
 
-					if (origValue != null)
+					foreach (var enumVal in MappingSchema.GetMapValues(type.UnwrapNullableType())!)
 					{
-						foreach (var enumVal in MappingSchema.GetMapValues(type.UnwrapNullableType())!)
-						{
-							if (origValue.Equals(enumVal.OrigValue) && enumVal.MapValues.Length > 0)
-								mapValue = enumVal.MapValues[0].Value;
-						}
+						if (origValue.Equals(enumVal.OrigValue) && enumVal.MapValues.Length > 0)
+							mapValue = enumVal.MapValues[0].Value;
 					}
 
 					SqlValue sqlvalue;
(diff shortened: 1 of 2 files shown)


MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Fixed enum-column null handling in association joins: an optional association whose join condition involves an enum column emitted `column <> 0` instead of `column IS [NOT] NULL` (the enum-to-`int` conversion was compared against `0` rather than null-checked), which polluted the join condition and could silently drop rows whose enum value equals its zero-valued member. The conversion-wrapped enum null check now correctly generates `IS [NOT] NULL`.


_(none)_


---
<sub>Generated from commit `692635d`.</sub>

Issue #1404, opened 2018-11-15
Add support of Futures


loudadyassine:
Hi,

It would be a great value for the library to have Future support. What I mean by Future is the possibility to batch multiple Sql queries in a single request to the server and parse the result one time.

You can see a practical case in [Ayend's blog](https://ayende.com/blog/3979/nhibernate-futures), But nowadays it's more important to have such a feature.

As many applications, we have a Saas application deployed on the Cloud, a Single Page Application :
- When the application first loads, it fetchs a lot of data from the server, just a bunch of Selects to get the current user name, the notifications, some elements on the dashboard, and other business elements etc... The Sql server is hosted in the cloud too, sometimes not the same Network than the server so there is a lot of firewall between the two. Without Future, we would need to send each request, wait for the result, parse the result, and then send the second request, during all this time Sql Server is in Idle mode ( Because of Session pooling, we won't face TCP Warm up problem ).
- In our application, we follow UnitOfWork Pattern, at the beginning of the request we open a transaction, at the end of the request we flush all the Updates/Inserts/Deletes, that make a big request of many queries, a fire&forget query because we don't read anything from this point, without Future, we would need to wait for each Update/Insert/Delete to finish to run the next one, and during this time Sql Server is in Idle mode.
- We have a SPA, so the reactivity of our application is directly related to the performance of each small webservice, and because we precalculate everything, the webservices will have just to fetch data from sql server. Think of it like a Dashbord showing a lot of data, User experience depends directly on the speed of fetching the data from the Sql Server and serving it to the App.

If we can do something like this that would be great :

```
public static List<Product> Search(string searchFor, int currentPage, int pageSize, out int totalRecords)
{
  using (var db = new DbNorthwind())
  {
    var products = from p in db.Product 
                   select p;

    if (searchFor != null)
    {
      products = from p in products 
                 where p.Name.Contains(searchFor) 
                 select p;
    }

    var lTotalRecords = products.Future().Count();  // That would return a Lazy<int>

    var ret = products.Skip((currentPage - 1) * pageSize).Take(pageSize).Future().ToList();  // That would return a Lazy<List<Product>>

    totalRecords = lTotalRecords.Value; // At this moment the ret.Value will have also the value set.

    return ret.Value;      
  }
}
```

sdanyliv:
So, it looks like `DataConnection` is `Future` context. I think it should be also configurable.
```cs
products.Future(ctx).Count()
```
Also `async` processing is needed.

[...]

PR #5673, opened 2026-07-03
Optimize the render pipeline; add opt-in combined-command execution and eager loading

sdanyliv:
> **Umbrella PR.** Two related reworks: the combined-command execution engine (immediately below) and the query render-pipeline redesign (§ *Render pipeline redesign*, folded in from #5688).
>
> **Two halves, only one of them gated.** The render-pipeline redesign is **always on** — that is where the measured time and memory wins come from. The combined-command engine is **opt-in**, behind `LinqOptions.UseCombinedCommands` (`false` in 6.x, planned `true` in 7), because enabling it changes observable behaviour; see § *Opting in*.

## What

Unifies linq2db's two ad-hoc multi-command mechanisms into one model and one execution seam, and adds a combined multi-result-set engine used by both DML and eager loading.

- **`SqlCommandScenario`** — ordered steps (`SqlCommandStep`) + `SqlCommandGroupPlan` + `SqlCommandExecutionContext`, with `IDmlService.BuildCommandScenario` / `PlanScenario` as the single grouping authority. All 10 providers migrated off the legacy `ISqlBuilder.CommandCount` / `BuildCommand` splitting.
- **Combined multi-result-set engine** — a grouped scenario runs as one command whose result sets are harvested per step (`WalkCombinedResultSets`). Enabled on SQLite / SQL Server / MySQL / PostgreSQL.
- **Eager loading** — the main query + its combinable child collections collapse from N+1 round-trips to 1 via the combined engine (`EagerResultEnumerable`); non-combinable strategies (KeyedQuery / CteUnion / detached / buffer) run sequentially first. Both enumerable materialization (`ToList()` / foreach) and single-element queries (`First` / `FirstOrDefault` / `Single` / `SingleOrDefault`) route through this engine.
- **One execution seam** — `ExecuteCombined(+Async)` backs both DML and eager combined execution, with two interchangeable backends: a semicolon-concatenated `DbCommand` or an ADO.NET `DbBatch`.
- **`Harvester`** split: base surface (`Execute` / `GetUsedParametersAndValues` / `IsInlined`) + an `IStepMaterializer` interface implemented only by reader-combinable harvesters. The public `IQueryRunner.Preambles` property and the `GetQueryRunner` `preambles` parameter keep their names.
- **Render cache** — both paths cache rendered commands on `QueryInfo` off one shared `PreparedCommand` template type, replacing the two ad-hoc cache types: the main query as a statement-free `PreparedQuery` (`Prepared`), the combined eager load as a `PreparedScenario` (`EagerCommandCache`) that also carries its step facts and group list. A `LoadWith` query uses both slots — the eager executor for its data, `GetCommand` for `ToString()` / `GetSqlText`. A statement that isn't parameter-dependent is rendered once and only re-bound per run, and the grain is **per-command** — so a stable main SELECT stays cached while a `string.Contains` (→ `LIKE`) detail collection re-renders.
- **Structure** — `Query.Queries` (always a single entry; multi-statement expansion lives in `SqlCommandScenario`) collapses to one `QueryInfo` field; the eager executor and DML share one group-plan interpreter (`RunGroups`) over one scenario model — self-executing harvesters become `SqlStepKind.SelfExecuting` steps and reader-combinable ones become reader steps, driven by both the sync and async paths; and `IQueryRunner.QueryNumber` is now `[Obsolete]` (always `0`, signature-neutral).

## Opting in

The **combined-command engine** is gated by **`LinqOptions.UseCombinedCommands`**, `false` in 6.x. (The render-pipeline redesign is not gated — it is always on.) Three ways to set the switch:

```csharp
db.Options.UseCombinedCommands(true)                       // per context (DataOptionsExtensions)
LinqToDB.Common.Configuration.UseCombinedCommands = true;  // process-global default
new LinqOptions(UseCombinedCommands: true)                 // the record property directly
```

It is off by default because enabling it is **observable, not merely faster**: a combined eager load runs inside an implicit read-consistency transaction where sequential loading used none, the emitted SQL changes shape (statements joined by `;`, sharing one parameter scope with deduplicated names), and on the `DbBatch` backend the command travels a different ADO.NET path, which changes traced output. A minor release should not turn that on for everyone.

The switch is **policy**, layered on top of — never instead of — provider **capability**: a provider that does not report `IsMultiStatementBatchSupported` and `IsMultipleResultSetsSupported` stays sequential however the option is set. That is also why the check lives at the callers of `IDmlService.PlanScenario` rather than inside it: that method answers "what can this provider merge", and folding policy in would make a provider misreport its own abilities.

Within DML, only **identity inserts** actually change shape. The `InsertOrUpdate` / `Upsert` emulation builds gated steps and the planner always flushes before a gated step, and a truncate-with-reset is all non-query, so both already planned as one command per step.

## SQL example: eager load in one round-trip

For `db.GetTable<Parent>().LoadWith(p => p.Children).Where(p => p.Id > minId)` on SQL Server, the parent query and its child collection previously ran as **separate round-trips**. The combined engine now sends them in **one** round-trip (two result sets — children, then parents — harvested per step), rendered one of two ways. Captured SQL (table prefix simplified):

**Semicolon-concatenated** — `UseDbBatch` off, or the provider connection can't create a batch, or an `ICommandInterceptor` is registered. One `DbCommand`; the statements are joined by `;` and share **one** parameter scope, so `@minId` is emitted once and reused (deduplicated across the merged statements to match the separate-command path):

```sql
SELECT [m_1].[Id], [d].[Id], [d].[ParentId], [d].[Title]
FROM   [Parent] [m_1]
       INNER JOIN [Child] [d] ON [m_1].[Id] = [d].[ParentId]
WHERE  [m_1].[Id] > @minId
;
SELECT [x].[Id], [x].[Name]
FROM   [Parent] [x]
WHERE  [x].[Id] > @minId
```

**DbBatch** — `UseDbBatch` on (the default) where the provider connection supports ADO.NET `DbBatch` (SQL Server, PostgreSQL, MySQL, …). The same two statements are sent as separate `DbBatchCommand`s in one batch, **each with its own parameter scope** — so each command carries its own `@minId` and no cross-statement renaming is ever needed:

```sql
-- DbBatchCommand #1
SELECT [m_1].[Id], [d].[Id], [d].[ParentId], [d].[Title]
FROM   [Parent] [m_1]
       INNER JOIN [Child] [d] ON [m_1].[Id] = [d].[ParentId]
WHERE  [m_1].[Id] > @minId

-- DbBatchCommand #2
SELECT [x].[Id], [x].[Name]
FROM   [Parent] [x]
WHERE  [x].[Id] > @minId
```

The only functional difference between the two is parameter packaging: the concatenated command merges statements into one scope (hence the shared/uniquified `@minId`), while the DbBatch keeps each statement isolated. Both are a single server round-trip; the identity-insert DML path (`INSERT …; SELECT SCOPE_IDENTITY()`) collapses the same way.

## DbBatch backend

Which backend a combined command uses — it selects nothing on its own, since without `UseCombinedCommands` there are no multi-statement commands to send either way.

The `DbBatch` backend is gated by `LinqOptions.UseDbBatch` and auto-falls back to a concatenated command where the provider connection can't create a batch or an `ICommandInterceptor` is registered (interceptors are `DbCommand`-typed). Measured on MySQL 9.6: `MySqlBatch` sends one combined request (≈ concat, actually fewer bytes), so no round-trip regression there; a genuine win on SQL Server / PostgreSQL / MariaDB. The concatenated fallback requires the connection to accept multi-statement text — unconditional on SQL Server / SQLite / Npgsql and MySqlConnector, but `MySql.Data` needs it enabled in the connection string (or `UseDbBatch` left on, the default).

## Eager read-consistency transaction

Kept at `SqlProviderFlags.DefaultMultiQueryIsolationLevel` (RepeatableRead by default) for every combined eager load. A per-ADO-provider check confirmed no mainstream provider reproduces a cross-statement snapshot in a single round-trip — PostgreSQL's single-`Sync` batch is atomic but runs at READ COMMITTED (re-snapshots per statement); SQL Server / MySQL / SQLite autocommit per statement — so skipping the transaction would silently downgrade the guarantee. It is reused, never doubled, when a transaction is already active.

## Render pipeline redesign (folded in — was #5688)

Reworks the query render pipeline end to end for speed and lower allocations:

- **Structural prepare.** Optimize + convert + alias + scenario + plan are extracted into `ScenarioCommandRenderer.PrepareStructure`, run once under the query lock and consumed by the render in that same call; what outlives it — and what a cache hit skips re-optimizing and re-converting for — is the statement-free `BakedQuery` below.
- **Statement-free render cache.** A non-parameter-dependent query bakes once into a statement-free `BakedQuery` / `PreparedQuery`, caches it, and drops its `SqlStatement` / `AliasesContext` graph (collectible by type). The scenario interpreter runs off `ExecutionStep[]` plus the rendered physical commands and never re-renders at execution.
- **One optimize+convert pass.** `SqlExpressionConvertVisitor` now derives from `SqlExpressionOptimizerVisitor`; a single `Convert` traversal optimizes, lowers, and runs the CompareNulls null-guard reduce per node — collapsing the old five passes to one. A SearchCondition fixed-point fix keeps the single direct pass byte-identical to the remote double-convert.
- **SQL builder decoupled** from `OptimizationContext` via `ISqlBuilderRenderContext`; the builder converts on the fly without mutating the AST.
- Dead-parameter and orphan cleanup on the optimize/convert entry points.

Parameter-dependent queries keep the per-execution convert+render; their emitted SQL is byte-identical to before.

### Performance

BenchmarkDotNet, net10.0, InProcessEmit (6 warmups / 20 iterations). **Three** configurations, measured back-to-back in one session with the same benchmark code built against both libraries:

- **master** — `origin/master`; no combined engine at all.
- **6.x default** — this branch with `UseCombinedCommands` **off**: the render-pipeline redesign only. What an application gets on upgrade.
- **opt-in** — the same binary with `UseCombinedCommands` **on**.

Splitting the two keeps the effects separable; a single master-to-PR number would mix them.

#### Pipeline (always on) — master → 6.x default

| Benchmark | Δ time | Δ memory |
|---|---|---|
| Render_ComplexParam | −36% | −51% |
| Render_ValueDependent | −31% | −34% |
| Execute_Insert | −28% | −22% |
| Prepare_Cold | ~flat | −7% |
| Render_Hot / Render_SkipTake / Execute_Select | ~flat | ~flat |

Carried over from an earlier two-way run and **not** re-measured in this three-way pass: Compiled INSERT / query (Issue3253) −34…−54% / −40…−53%; Update / Compiled_Update −24…−32% / −28…−32%; Small / Large INSERT −17…−26% / −17…−22%; Concurrent execution (16–64 threads) ~flat / −24…−25%. Cache activity and trivial query generation are neutral.

#### Combining (opt-in) — 6.x default → opt-in

| Benchmark | Δ time | Δ memory |
|---|---|---|
| Execute_Insert | +1.7% | +1.5% |
| Render_ComplexParam | +3.6% | +0.2% |
| eager Execute_Hot / Prepare_Cold | ~flat | ~flat |

Opting in costs **1.5–3% locally and shows no time win here — by construction**. The eager benchmark runs against in-memory SQLite, where a "round trip" is an in-process call, so the thing combining actually saves has no price in this environment: eager `Execute_Hot` measures 16.79 / 16.86 / 16.88 ms across master / default / opt-in — indistinguishable. Combining's benefit is the round-trip **count** (N+1 → 1), which is deterministic and asserted by tests rather than timed; the time it converts to scales with per-round-trip latency. So read this table as an upper bound on the overhead, not as evidence about the benefit.

<details><summary>Absolute numbers — Mean (µs) / Allocated (KB), all three configurations</summary>

| Benchmark | master | default | opt-in | master KB | default KB | opt-in KB |
|---|---|---|---|---|---|---|
| Render_ComplexParam | 26,093 | 16,575 | 17,178 | 45,555 | 22,186 | 22,242 |
| Render_ValueDependent | 1,742 | 1,194 | 1,193 | 3,864 | 2,547 | 2,578 |
| Execute_Insert | 929 | 672 | 683 | 2,723 | 2,116 | 2,147 |
| Render_SkipTake | 560 | 548 | 547 | 778 | 777 | 777 |
| Execute_Select | 483 | 479 | 473 | 687.5 | 704.7 | 704.7 |
| Render_Hot | 456 | 437 | 438 | 600 | 598.4 | 598.4 |
| Prepare_Cold | 113,332 | 113,991 | 113,644 | 12,746 | 11,818 | 11,847 |
| eager Execute_Hot (ms / MB) | 16.79 | 16.86 | 16.88 | 4.00 | 4.05 | 4.04 |
| eager Prepare_Cold (ms / MB) | 198.7 | 199.6 | 198.9 | 39.67 | 38.43 | 39.32 |

</details>

One real cost on the simplest path: **`Execute_Select` allocates +17.2 KB/op (+2.5%)** — 704.7 KB against master's 687.5 KB, time flat. Per execution that is ~88 bytes on an operation allocating ~3.5 KB: the structural cost of routing every query through one execution seam instead of two ad-hoc mechanisms. It reproduces to the tenth of a KB across sessions.

An earlier revision of this section also listed `Render_Hot` as +2.0% time. The three-way re-measurement does **not** reproduce it — 456 (master) vs 437 (default) — and master's run there varied by ±17 µs against ±3–5 µs on the branch, so the original figure was noise reported as a finding.

## Verification

- Builds `net462` / `netstandard2.0` / `net8.0`–`net10.0`, Release analyzers + banned-API clean.
- 0 baseline drift (before/after regen on SQLite.MS: 319 rendered-SQL baseline files byte-identical).
- Provider matrix (net10.0): SQLite (concat) + PostgreSQL / SQL Server / MySQL (DbBatch) — eager loading, identity DML, upserts, interceptors.
- Render-pipeline cross-provider gate (net10.0): SQLite.Classic, Firebird 5, DB2, MySqlConnector 8.0 — **38,666 tests, 0 failures, 0 stackhops, 0 direct-vs-remote baseline mismatches**.
- No public (non-`Internal`) API added — the new pipeline types live in `LinqToDB.Internal.*` namespaces (the sanctioned pattern). The regenerated `CompatibilitySuppressions` adds **125 entries — 115 `CP0002` + 10 `CP0006` — every one of them under `LinqToDB.Internal.*`**, covering the members the scenario and render-pipeline reworks removed or re-signatured: `CommandCount` / `BuildCommand` / old-arity `BuildSql` on `ISqlBuilder` + `BasicSqlBuilder`, the `OptimizationContext` and `SqlExpressionConvertVisitor` surface, `ISqlOptimizer.ConvertSkipTake`, `SqlProviderFlags.IsIdentityParameterRequired`, and the rest. Verified by set comparison against `master`: **125 targets added, 0 removed, 0 outside `LinqToDB.Internal`** — no non-`Internal` drift. (The six `LinqToDB.Linq.Translation.*` entries a line diff appears to touch are pre-existing in `master` and only move position when the file is regenerated.)

## Open items

- Combined-engine rollout beyond SQLite / SQL Server / MySQL / PostgreSQL; #1404 (Deferred / Futures) builds on this substrate.
- Keyed-flavor server-side keyset forwarding starts from scratch: its scaffolding (`SqlProviderFlags.IsServerSideVariablesSupported`, `SqlStepParameterBinding` and the `ParameterBindings` plumbing) never got a producer and has been removed as dead code.


sdanyliv:
Review pass over this branch — 16 findings walked, 12 fixed in `49c2cc68c7`.

## Fixed

**Enumeration paths.** `ExpressionQuery.GetForEachUntilAsync` was the one path never reworked alongside its siblings: it passed the bare `Preambles` field that nothing on that path assigns, never opened the read-consistency transaction, and never disposed its enumerator. A compiled query reaching it threw `NullReferenceException` — works on master. `CompileTests.CompiledQuery_ForEachUntilAsync` reproduces it red→green on both the direct and `.LinqService` contexts. `ForEachUntilAsync` had no test coverage anywhere, which is how this shipped.

`ExecuteAsyncEnumerable` disposed the transaction when the method returned, i.e. before the caller enumerated — harvesters ran inside it while the main query streamed outside. New `AsyncEnumerableAsyncWrapper<T>` transfers ownership to the returned sequence.

`EnumeratorWrapper.Dispose` and `AsyncEnumeratorAsyncWrapper.DisposeAsync` released the transaction only when the inner enumerator disposed cleanly; now released in `finally`.

**Combined execution.** `WalkCombinedResultSets` discarded the `NextResult()` result, so a reader running dry produced silently empty eager children and shifted every later step onto the wrong result set — it now throws. `BuildPlan` takes the same `QueryInfo` lock `GetCommand` uses; that monitor is what licenses `GetCommand`'s in-place mutation of the shared statement.

**SAP HANA.** Restored the identity guard that went away with `SapHanaSqlBuilder.BuildCommand`. `CURRENT_IDENTITY_VALUE()` is session-scoped, so `InsertWithIdentity` against a table with no identity column returned a value from an unrelated earlier insert instead of failing. Every sibling provider still carries the guard, and nothing in the suite covered it for any of them. Red→green on a live HANA instance.

**Baselines.** The direct-vs-remote comparison was weakened suite-wide: `Canonicalize` ran for every baseline-capturing test, not only combined ones, and dropped the entire `DECLARE`/`SET` parameter dump before comparing. Removed. The remote leg is now skipped where combining is forced on, so the gate is strict again rather than tolerant.

**Coverage.** `CombinedDmlTests` fans an identity insert and an `InsertOrReplace` across providers under both settings — the DML half of the engine was exercised on SQLite only, while the ten `BuildCommandScenario` overrides each use a different identity mechanism. `CombinedEagerLoad_HonoursDbBatchBackend` covers the `DbBatch` and concat backends. `DataOptions.UseDbBatch` fluent setter added; only `UseCombinedCommands` had one.

## Refuted or not reproduced

- The `Microsoft.Data.Sqlite`-not-deployed concern is void: `linq2db.Benchmarks.csproj` imports `linq2db.Providers.props`, and the built output carries the assembly plus its native binaries.
- `[IncludeDataSources(...)]` already excludes remote contexts — its `params` overload passes `includeLinqService: false` — so the `TruncateTableTests` guards were dead code, not skipped assertions. Guards removed.
- The eager-render vs `GetCommand` race did not reproduce: 1280 concurrent operations against a shared `QueryInfo` with the lock removed, zero failures. The lock is kept as cheap insurance — measured wait median 0.3 µs, p99 381 µs, and on the cold path it prevents 32 threads rendering the same plan.

## Outstanding

`PublicAPI.Unshipped.txt` declared none of this branch's added public surface. Building with `-p:RunApiAnalyzersDuringBuild=true` reports 220 RS0016 and 54 RS0017; master is clean under the identical command. The 220 additions are now declared and verified at zero on net10.0, netstandard2.0 and net462.

The **54 RS0017 remain deliberately undone**. They are removals of public members — `ISqlBuilder.BuildSql` with its ten provider overrides, `BasicSqlBuilder.OptimizationContext`, `IQueryRunner.Parameters`, the `DataProviderParameterContext` constructor. Each `*REMOVED*` entry asserts that breaking that extension point is intentional, so they want a decision rather than a script.

Worth noting that none of this PR's green runs would have caught it: the API-analyzer leg was not among them.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDmlService.cs
```diff
@@ -0,0 +1,102 @@
...
+
+		public override SqlCommandScenario? BuildCommandScenario(SqlStatement statement, SqlProviderFlags flags, ISqlExpressionFactory factory)
+		{
+			switch (statement)
+			{
+				case SqlTruncateTableStatement { ResetIdentity: true } truncate when truncate.Table!.IdentityFields.Count > 0:
+				{
+					// No schema on the reset sequence reference (matches the previous BuildCommand behavior).
+					var seq   = new SqlObjectNameExpression(new SqlObjectName(SequenceName(truncate.Table.TableName.Name)), ConvertType.SequenceName);
+					var reset = new SqlFragmentStatement(factory.Fragment(
```
MaceWindu, line 40: **Major · MAJ001** — The truncate-with-identity-reset block interpolates the sequence name into SQL **string literals** without literal escaping, which the code it replaced applied.

`{0}` is a `SqlObjectNameExpression`, rendered through `BasicSqlBuilder.BuildFormatValues` → `BuildObjectName(…, escape: true)` — that is *identifier* escaping. `OracleSqlBuilderBase.Convert` wraps a non-identifier in `"…"` and never touches apostrophes, so an apostrophe in the name closes the surrounding `'…'` literal.

The removed `OracleSqlBuilderBase.BuildCommand` did escape it: it composed each inner statement into a string and emitted it via `BuildValue(null, <string>)` → `MappingSchema.TryConvertToSql`, which doubles embedded apostrophes. The correct pattern is still in the same file — `AppendExecuteImmediateDrop` (the DROP path) continues to use `BuildValue`.

Rendered offline for a table mapped as `[Table("O'Brien")]`, both blocks from one run:

```sql
-- DROP path (unchanged, still escapes)
BEGIN
	EXECUTE IMMEDIATE 'DROP TRIGGER "TIDENTITY_O''Brien"';
	EXECUTE IMMEDIATE 'DROP SEQUENCE "SIDENTITY_O''Brien"';
	EXECUTE IMMEDIATE 'DROP TABLE "O''Brien"';
END;

-- truncate-with-reset path (this file)
DECLARE
	l_value number;
BEGIN
	EXECUTE IMMEDIATE 'SELECT "SIDENTITY_O'Brien".NEXTVAL FROM dual' INTO l_value;
	EXECUTE IMMEDIATE 'ALTER SEQUENCE "SIDENTITY_O'Brien" INCREMENT BY -' || l_value || ' MINVALUE 0';
	...
```

The literal closes at `O'` and `Brien".NEXTVAL FROM dual` is left to be parsed as PL/SQL. Same provider, same table, same run — so the difference is the migration, not the name. The sequence name is `"SIDENTITY_" + tableName`, and table names reach this from mapping and from `.TableName(...)` at runtime.

Fix: compose the inner statement text (`SELECT <quoted-seq>.NEXTVAL FROM dual`, `ALTER SEQUENCE <quoted-seq> INCREMENT BY 1 MINVALUE 0`) and pass **that string** through `factory.Value(...)` so the mapping schema emits it as a properly escaped literal, leaving only the `INCREMENT BY -' || l_value || '` concatenation assembled in the fragment. Alternatively add a fragment placeholder flavour that renders an object name through the string-literal escaper, or reject an identifier containing `'` at scenario-build time. The same rewrite also restores the four explanatory `--` comment lines the migration dropped.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
```diff
@@ -620,17 +618,460 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
...
+			// executes before every combinable sibling and must only read result slots that have already run - its own
+			// lower-indexed dependencies. Nested eager loads build first, so they get lower indices and materialize earlier; a
+			// parent harvester always reads its child slots after they are populated. A non-combinable harvester that read a
+			// *combinable* sibling's (higher-lifecycle) slot would find it unpopulated and silently produce an empty collection
+			// - no exception. Any new strategy MUST preserve this direction (covered by
+			// MixedCombinableAndNonCombinable_NestedKeyedUnderCombinable).
+			//
+			// Parameter-independent (the partition is fixed and each command's step indices come from the cached template), so
+			// it is built on the cold path and cached with the templates.
+			SqlCommandGroup[] BuildGroups(IReadOnlyList<CombinedCommand> commands)
```
MaceWindu, line 913: **Minor · MIN001** — The ordering invariant documented directly above this method states a precondition that nothing enforces.

The comment's two claims don't compose. It says non-combinable steps run first and "must only read result slots that have already run - its own lower-indexed dependencies", then justifies that with "nested eager loads build first, so they get lower indices and materialize earlier". But `BuildGroups` partitions on **combinability**, not index, so index order is preserved only *within* each partition. A combinable child (low index, built first) under a non-combinable parent (high index) inverts it: the parent runs first, reads an unpopulated slot, and — per the comment's own words — silently produces an empty collection.

Mixing is structurally possible: `CanCombine` on `DetachedHarvester`, `Harvester<TKey,T>` and `CteUnionHarvester` all reduce to `query.GetResultFromReader != null`, which the CteUnion and keyed-buffer overrides leave null, and the per-detail fallback chain (`CteUnion → KeyedQuery → Default`) can land different strategies on different details of one query.

I measured rather than guessed. Instrumenting this partition and running the eager-loading fixtures plus `EagerRenderCacheTests` and `CompileTests` on SQLite gave **86 combined-eager plans built, 1 mixed, 0 inverted**. The single mixed plan is `nc=[0,1,2,3] c=[4]` — `KeysDataSlotHarvester`, `BufferMaterializeHarvester`, `KeyedQueryChildHarvester`, `Harvester`, `Harvester` — i.e. the keyed harvesters at the low indices and the combinable one last, the safe direction that `MixedCombinableAndNonCombinable_NestedKeyedUnderCombinable` already covers.

So this is a robustness finding, not a bug report: the inverted shape is not produced anywhere in the corpus, and the invariant holds today only because combinability happens to correlate with index order in every plan built.

Fix: make the comment true by construction. In `TryGetCombinedEagerEnumerable`, refuse the combined path when `max(_nonCombinableIndexes) > min(_combinableIndexes)` and fall back to the sequential `InitHarvesters` path — roughly three lines, and it turns an unstated precondition into an enforced one. An alternative that also settles reachability permanently is to have `SqlCommandExecutionContext.GetResult` throw when `WasExecuted(step)` is false, turning the silent empty collection into a diagnosable error.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
```diff
@@ -620,17 +618,460 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
...
+			// it, never instead of it.
+			if (!dataContext.Options.LinqOptions.UseCombinedCommands)
+				return null;
+
+			if (dataContext is not DataConnection dataConnection
+				|| !dataConnection.DataProvider.SqlProviderFlags.IsMultiStatementBatchSupported
+				|| !dataConnection.DataProvider.SqlProviderFlags.IsMultipleResultSetsSupported)
+				return null;
+
+			// Query hints (context QueryHints / one-shot NextQueryHints) are applied AND cleared by the sequential
```
MaceWindu, line 1091: **Minor · MIN002** — `DataConnection.ProcessQuery` is silently bypassed by the combined eager path.

`ProcessQuery(SqlStatement, EvaluationContext)` is a shipped `public virtual` extension point whose purpose is to let a `DataConnection` subclass rewrite the statement before render. On the sequential path it runs inside `DataConnection.QueryRunner.GetCommand`, and each eager child goes through its own `GetCommand`.

The combined path never calls `GetCommand`: `EagerResultEnumerable.BuildPlan` feeds `GetStepStatement(...)` — the main query's `QueryInfo.Statement` and each combinable child's — straight into `CombinedReaderBatch`, which renders via `ScenarioCommandRenderer.RenderStatementTemplates` / `RenderCombinedBatchTemplates` with no `ProcessQuery` call anywhere. So with `UseCombinedCommands` enabled, a subclass's rewrite stops applying to the main query and to every combinable child while still applying to the non-combinable ones — a per-harvester split in behaviour with no diagnostic.

This block is the natural place to raise it because it already carries the sibling bail-out for the other piece of `GetCommand`-only state; the enumeration stopped at query hints.

Capped at Minor because the gate is opt-in and off in 6.x and nothing in-tree overrides `ProcessQuery` — the exposure is external consumers, and it grows when the option flips default in v7.

Fix: route every statement the combined path renders through `ProcessQuery` the way `GetCommand` does — in `GetStepStatement`, or once per participant in `CombinedReaderBatch.RenderCold` — with an `EvaluationContext` over this run's values, treating a non-reference-equal result as parameter-dependent so the render is not cached. If that is too invasive for the eager cache, the cheap alternative is a bail-out beside the query-hints one; note it must key on whether `ProcessQuery` is **overridden**, not on `dataConnection.GetType() != typeof(DataConnection)`, which would disable combining for every typed subclass. Either way it belongs in `LinqOptions.UseCombinedCommands`'s documentation, which already enumerates what enabling the option changes.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
```diff
@@ -620,17 +618,460 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
...
+					if (IsCombinable(harvester))
+						return true;
+
+				return false;
+			}
+
+			// Merges every combinable child's and the main query's parameter values into one SqlParameterValues (keyed by
+			// SqlParameter node). Genuinely per-run - the values are this execution's - so it is the only part of the old
+			// PrepareScenario a warm execution still has to do.
+			SqlParameterValues CollectParameterValues()
```
MaceWindu: **Suggestion · SUG001** — The combined path collects parameter values with a **null** `SqlCommandExecutionContext`, which contradicts a stated precondition elsewhere.

Both calls here reach `SetParameters(Query, IQueryExpressions, IDataContext?, object?[]?, SqlParameterValues, SqlCommandExecutionContext? context = null)` with `context` defaulted to null, and it is forwarded straight into `accessor.ClientValueAccessor(expressions, parametersContext, parameters, context)`. The sequential path never does this — `QueryRunnerBase.SetCommand` passes the real `ExecutionContext`, which is how the keyed strategy feeds master keys into a detail query's IN-list.

The two context reads behave differently on null. `ExpressionBuilder.EagerLoadKeyedQuery` line 842 is a direct instance call, so a null context is a `NullReferenceException` — loud. Line 788 routes through a helper that null-guards:

```csharp
// context is null only when the accessor is evaluated without a live execution (e.g. SQL-text generation);
// the value is unused there, so return null.
static TKey[]? GetKeysAt<TKey>(SqlCommandExecutionContext? context, int index)
    => context != null ? (TKey[]?)context.GetResult(index) : null;
```

That comment is a precondition — *context is null only when there is no live execution* — and this call site falsifies it: it passes null **during** a live execution. If a keyed keys-source accessor were ever reached from here it would silently receive `null` parent keys rather than throwing.

No route reaches it today: keyed harvesters are non-combinable so their `AddCombinableParameterValues` is never called, and `SetParameters(_query, …)` runs the main query's accessors, which don't include a child's keys-source. So this is a suggestion, not a defect — but the asymmetry is free to remove, and it guards a path v7 turns on by default.

Fix: thread the enumeration's `SqlCommandExecutionContext` into `AddCombinableParameterValues` and into the main-query `SetParameters` call, so both paths agree.

Source/LinqToDB/Data/DataConnection.QueryRunner.cs
```diff
@@ -358,99 +405,640 @@ static DbParameter CreateParameter(DataConnection dataConnection, DbCommand comm
...
+						throw new InvalidOperationException($"Step kind '{step.Kind}' cannot be executed as a single DML command.");
+				}
+			}
+
+			// Shared combined-command harvest walk (sync): over one open reader, invokes the per-step harvest callback for the
+			// FIRST harvestCount entries of stepIndexes in order, then advances to the next result set for every
+			// result-producing step (a pure non-query step yields no result set, so no advance). A caller that streams the
+			// group's last step itself (the eager main) passes a count one short of the group instead of slicing the list.
+			// DML harvests scalar / rows-affected into the execution context; eager loading plugs its own materializer in.
+			internal static void WalkCombinedResultSets(DbDataReader dr, IReadOnlyList<int> stepIndexes, int harvestCount, IReadOnlyList<ExecutionStep> steps, Action<int, DbDataReader> harvest)
```
MaceWindu, line 667: **Suggestion · SUG002** — The walk's "a pure non-query step yields no result set" premise is a **driver** behaviour that no capability flag expresses.

`NextResult()` is called only after a step whose `Kind != NonQuery`. That holds for SqlClient — a batched `INSERT` emits a DONE token, not a result set — and it is what makes the identity group `INSERT …; SELECT <id>` harvest correctly. It is not universal: Npgsql surfaces one result per statement in a multi-statement command, so the same shape would need an extra advance.

Nothing reaches it today, and the reason is worth recording: PostgreSQL registers no `IDmlService` override, so it inherits `BasicDmlService` → `DmlServiceBase.BuildCommandScenario` → `DefaultScenario` (a single step) for an identity insert because it renders `RETURNING`; and the eager-loading path is all `Reader` steps. So no `NonQuery`-then-`Scalar` group forms on a per-statement-result-set provider.

It becomes live the moment such a provider gains a multi-step DML scenario — which this PR's **Open items** names as the next step. The failure is silent, and the existing guard does not catch it: the `!advanced && k + 1 < harvestCount` check fires on *too few* result sets, whereas this trap is *too many* — the `Scalar` step would read the INSERT's empty result set, `dr.Read()` returns false, and the identity comes back `null` rather than throwing.

Fix: either promote the assumption to a `SqlProviderFlags` capability alongside `IsMultipleResultSetsSupported`, or make the walk tolerant — advance unconditionally and accept a `NextResult()` that returns false.

Tests/Linq/Update/CombinedDmlTests.cs
```diff
@@ -0,0 +1,89 @@
...
+	/// <para>
+	/// Remote contexts are excluded on purpose: combining is a DataConnection-only path, so a remote leg would re-test
+	/// the sequential shape while producing a differently-grouped baseline than the direct one.
+	/// </para>
+	/// </remarks>
+	[TestFixture]
+	public class CombinedDmlTests : TestBase
+	{
+		// Identity insert: two steps (INSERT + identity retrieval) that the combined engine merges into one command.
+		[Test]
```
MaceWindu, line 68: **Minor · MIN003** — Both `[Values]` arms assert identically, so this fixture cannot detect the combined engine ceasing to engage.

`combinedCommands` reaches `o.UseCombinedCommands(combinedCommands)`, but both arms of both tests then assert the same thing — the identity round-trips, the last `InsertOrReplace` wins — and nothing observes whether a merged group was actually sent. If `UseCombinedCommands(true)` stopped taking effect (a planner change that flushes every group, a capability-flag regression, an options-plumbing bug), both arms keep passing on the sequential path.

That is the symmetric half of what the fixture's own `<remarks>` argues: it notes that "a green run with the switch off says nothing about any of them", and the same is true of a green run with the switch on when nothing checks it was on. The eager-loading side does guard this — `EagerLoadingTests.CorrelatedAndDetachedChildren_CombineIntoSingleCommand` counts SELECTs and asserts `ShouldBe(1)`, and `EagerLoadingStrategyKeyedQueryTests` carries an explicit anti-trivial guard.

Note the obvious guard doesn't work here: asserting `1` when `UsesCombinedCommands()` and `2` otherwise is wrong for providers whose identity scenario is already a single step — PostgreSQL registers no `IDmlService` override and uses `RETURNING`, so it is one command either way, and these tests run everywhere via `[DataSources(false, …)]`.

Fix: a self-calibrating guard. Run both configurations inside one method with a `CommandInterceptor` counting `CommandInitialized`, then assert `onCount <= offCount` always and `onCount == 1` when `db.UsesCombinedCommands() && offCount > 1` — using the off arm to discover whether the provider's scenario is multi-step at all. That keeps the expectation derived from the real gate rather than from a provider list.

MaceWindu (commented):
> 🤖 Agentic review. Findings are severity-tagged; line-anchored items are inline comments.

## Review notes

**No completed CI run had validated `d448ac3b`.** Build 23233 was the only run for it and it was **canceled**. Within it, two tests failed on **62 provider legs** — `RemoteContextTests.RemoteRunnerDoesNotLeakClientAcrossTwoQueryPlan` and `RemoteRunnerLeavesAContextOwnedClientAlive`, both on `Assert.That(context.NonQueryExecutions, Is.EqualTo(2))` → `But was: 1`.

That is not a behaviour regression. Deleting `SetNonQueryQuery2` / `NonQueryQuery2` and moving the UPDATE→INSERT reshape into `DmlServiceBase.BuildInsertOrUpdateScenario` means a remote context now issues one `ExecuteNonQuery` and the server rebuilds the gated scenario; all three emulation shapes are preserved, and `PlanScenario` forces every `RunIf` step into its own singleton group. `_client ??= GetClient()` is intact.

What it cost is coverage, and that is worth a decision rather than a number change: those tests reached a second `Execute*` on one runner only through the two-query plan, and every `runner.Execute*` site is now one execute per runner — so relaxing the guard from 2 to 1 leaves `??=` unguarded against a regression to `=`. **Fixed in `fbb29d965`**: the first test is renamed `RemoteRunnerDisposesEveryClientItObtained` and drives two operations through one context (`ClientsDisposed == ClientsCreated`, with a `ClientsCreated >= 2` anti-trivial guard); the second keeps its shared-client assertions with the guard at 1. The comment records what is no longer covered. If you want the `??=` guard back, it needs a test that drives the runner directly.

Remaining failures in that build are environment (`SapHana.Odbc` fails at `CreateDatabase`; Oracle `ORA-01920 c##sequence_schema`) or timeouts (`JoinTests.StackOverflow`, `MergeTests.ComplexProperty_NestedDiscriminator`).

**Also pushed** (each its own commit): `a701edf27` enables `UseCombinedCommands` in `EagerLoadExceptionIntercepted` / `…Async` — they document the `DbBatch` path and `RunBatchReader` as their subject but opened a bare `GetDataConnection`, so with the option off in 6.x no `DbBatch` was ever constructed and both passed on the ordinary `ExecuteReader` path that `ReaderExceptionIntercepted` already covers, leaving `RunBatchReader`'s interceptor `catch` untested on every leg. `2e73e0812` and `bea16b212` correct two XML docs (details below).

**Verification claims.** The `CompatibilitySuppressions` claim is exact — set comparison on `(DiagnosticId, Target, Left, Right)` gives 125 added, 0 removed, 115 `CP0002` + 10 `CP0006`, 0 outside `LinqToDB.Internal`; the raw line diff's 228 added / 103 removed `<Target>` lines are ApiCompat reordering, not entity changes. Two others don't hold:

- *"No public (non-`Internal`) API added"* — five public members are added (`LinqOptions.UseDbBatch`, `LinqOptions.UseCombinedCommands`, `Configuration.UseCombinedCommands`, both `DataOptionsExtensions` overloads), plus the positional constructor and `Deconstruct` at arity 20→22. All intentional and additive, and the *Opting in* section advertises three of them — it is the bookkeeping that is wrong, and this is the line a release-notes author will act on.
- *"0 baseline drift (before/after regen on SQLite.MS: 319 rendered-SQL baseline files byte-identical)"* — 390 SQLite.MS files changed on the branch, 191 of them modified.

Two XML docs were corrected rather than filed as findings:

- `SqlProviderFlags.MaxCombinedCommandLength` said it bounds the combined engine when it "merges scenario steps" and that `0` disables merging. Its only reader is `RenderCombinedBatchTemplates` (the eager-loading render); the DML path is bounded by `MaxStatementsPerCombinedGroup` alone and `RenderScenarioGroups` applies no length test, so `0` still merges there. The code is right for what it must do today — a merged DML group is at most two steps — so this was a doc correction, not a request for a length check.
- `LinqOptions.UseCombinedCommands` listed the implicit read-consistency transaction among the things enabling the option changes, "where sequential loading used none". `ExpressionQuery.StartLoadTransaction` opens it whenever the query has harvesters, with no `UseCombinedCommands` test on that path; `master` does the same gated on `IsAnyPreambles`. The option renames the predicate, not when the transaction opens. The other two observable differences stand.

## Prior-review audit

- **`LinqOptions` positional-record binary compatibility** (open thread) — *partially accurate, not actionable.* The immediately-previous 20-parameter signature is indeed uncovered; the only shims are a 15-parameter constructor and a 15-`out` `Deconstruct`. But `PackageValidationBaselineVersion` is `6.0.0` and the 15-parameter shim **is** the 6.0.0 signature, so ApiCompat reports no break — neither the base nor the head `CompatibilitySuppressions.xml` carries a `LinqOptions` entry. The four options added to this record before this one followed the same pattern without a per-release shim. `ConfigurationID` and the copy constructor were both updated for the new members, so there is no configuration-id collision.
- **`GetForEachUntilAsync` skips harvesters and the implicit transaction** (suppressed) — *inaccurate, and inverted.* At HEAD it calls `StartLoadTransactionAsync`, `GetEagerEnumerableAsync`, and disposes the enumerator under `await using`. The pre-PR version did none of the three — it read the stale `Preambles` field, opened no transaction, and leaked the enumerator on an early `break`. This PR fixes that.
- **NUnit `Assert` instead of Shouldly** in the new `CompileTests` / `Issue2161Tests` / `TruncateTableTests` additions (suppressed) — *not actionable.* Those files are uniformly NUnit-style and the additions match them. Both genuinely new fixtures (`CombinedDmlTests`, `EagerRenderCacheTests`) do use Shouldly.

## Minor

- [ ] **MIN004** — The new null-semantics expansion duplicates a correlated subquery from 2 to 6 occurrences in one `WHERE`
  **Why:** `PredicateTests.Test_ComplexPredicateComparisonWithUnknown` gains an `AND NOT (A IS NULL AND B IS NOT NULL OR A IS NOT NULL AND B IS NULL)` conjunct, narrowing "at least one operand is null" to "both are null" — the CLR `bool?` semantics, and the semantics look right (the test uses `AssertQuery` and is green). The conjunct contains `CASE_B` twice and each carries the correlated `SELECT COUNT(*)` twice, so the subquery goes from 2 to 6 instances evaluated per row. SqlServer cohort grows 9.83×; Access/DB2/Firebird ~15×, PostgreSQL/DuckDB ~14×, Oracle ~10×. This is the PR's own change, not stale drift — baselines master last wrote that file on 2026-08-07, before the branch's 2026-08-21 capture.
  **Fix:** Confirm the correctness-over-cost trade is deliberate, and consider hoisting the duplicated `CASE_B` so the subquery is evaluated once.

## Out-of-scope observations

- **`QueryHelper.Evaluate`: `Equal` / `NotEqual` now honour SQL UNKNOWN while the comparison arm still folds a NULL operand to `false`.** The `default` arm (`Greater`, `GreaterOrEqual`, `Less`, `LessOrEqual` and their `Not…` forms) returns `result = false; return true;` for a non-`IComparable` operand. Callers guard on `value is bool`, so the new `null` correctly declines the fold while `false` proceeds. Pre-existing on `master`; this PR only makes the two halves of one switch disagree. A probe comparing server results against the LINQ-to-Objects oracle passed in both the inlined and parameterized arms, so no divergence is demonstrated — recorded as an inconsistency, not a defect.
- **`InsertWithIdentity` against a table with no identity column silently returns a stale value on unguarded providers.** The new `SapHanaTests.InsertWithIdentity_NoIdentityColumn_Throws` is a good addition and its comment names the wider gap. The seven guard sites are provider-local (DuckDB, Firebird, Oracle, PostgreSQL, SAP HANA, YDB) and `BasicSqlBuilder`'s is only inside `GetSequenceNameAttribute`, a sequence-name helper — so there is no shared guard. Probed on SQLite: no exception, `INSERT …` followed by `SELECT last_insert_rowid()`, returning a connection-scoped value from an unrelated earlier insert. The same applies to `@@IDENTITY` (SqlCe, Access, Sybase) and `LAST_INSERT_ID()` (MySQL). Pre-existing and not caused by this PR — but it means a `[DataSources]` fan-out of the new test is blocked until those providers gain the guard.

## Baselines

Delta: [linq2db.baselines PR #2074](https://github.com/linq2db/linq2db.baselines/pull/2074) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5673)

**The delta is stale for the reviewed HEAD.** Its commits carry author date 2026-08-21, replayed onto a 2026-09-02 baselines master; `d448ac3b` is a 2026-09-02 master merge whose build was canceled, so nothing regenerated. Measured anyway: **8653 added / 5555 modified / 0 deleted**, 388,812 lines added / 30,419 removed, 10,635 net-positive files, 14,121 content-changed, every provider touched.

- **8102 of the 8653 additions** are `EagerLoadingTests` baseline-key renames from the new `[TestFixture(false)] [TestFixture(true)]` parameterization. The old keys remain as orphans, which is why deletions are zero.
- **The ~10,132 added `DECLARE` lines** are dominated by `Test_PredicateWithBoolean(*, False)` across ~67 providers, where a predicate compared against a boolean parameter is now deliberately not folded. That is intended and pinned by this PR's own `Test_PredicateWithBooleanParameterIsValueIndependent`, with the rationale recorded at the site — folding needs the value, and baking it into the memoized structure would make a re-run with the other value reuse it; and `(A = B) = @p` lowers to an inverted comparison on engines with no boolean type, returning wrong rows on Firebird 2.5. Not a finding.
- The rest of the modified set is largely line-rewrap from the new render pipeline. The substantive exception is MIN004 above.

A fresh full run is needed before merge — both to regenerate this delta against the current head and because no completed CI run has covered the code under review.

Source/LinqToDB/Internal/DataProvider/Oracle/OracleDmlService.cs
sdanyliv: **MAJ001 - fixed in c1c2079e34.**

Confirmed as a regression this PR introduced, against master: the removed `OracleSqlBuilderBase.BuildCommand` composed each inner statement into a string and emitted it through `BuildValue(null, <string>)`, which doubles apostrophes via the mapping schema. The fragment rendered `{0}` with identifier escaping only.

`SqlObjectNameExpression` gains `InsideStringLiteral`; when set, the builder escapes the rendered name for literal-body position, and the literal's own delimiters stay in the fragment's format string. Only the Oracle reset block sets it - `CREATE SEQUENCE` and the trigger keep their placeholders outside quotes. Swept every DML service for the same pattern; Oracle was the only hit.

`DmlScenarioRenderTests` renders the scenario offline with no connection and pins both the escaped name and the control case, so unconditional doubling would fail it.

The four dropped `--` comment lines are deliberately not restored: the escaping alone moves no baseline, while the comments would move every Oracle truncate-with-reset baseline. Happy to add them if you would rather have them back.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
sdanyliv: **MIN001 - fixed in b26006c271**, as the enforcement you suggested.

`TryGetCombinedEagerEnumerable` now refuses the combined path when a non-combinable index sits above a combinable one, and the predicate shares `IsCombinable` with the constructor's partition so the two cannot drift. The `BuildGroups` comment now says the invariant is enforced rather than relied on.

There is no red test to write, since nothing in tree produces the inverted shape. The risk the change itself carries is the opposite one - a guard that refuses a layout it should allow - and the existing mixed-shape test cannot catch that: its "more than one round-trip" assertion passes just as happily when combining is refused outright and everything runs sequentially. `MixedCombinableAndNonCombinable_StillCombines` counts the same shape with the switch off and on and requires `on < off`.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
sdanyliv: **MIN002 - fixed in 58cace7389**, as the bail-out rather than the routing.

Measured first: on one eager load, `ProcessQuery` is called twice sequentially and zero times with combining on.

Routing every combined statement through `ProcessQuery` would need this run's values in the render and would disable the render cache for any non-reference-equal result, so the gate goes beside the query-hints one instead. It keys on whether the method is overridden rather than on the context type, exactly as you noted - `TestDataConnection` is a subclass and does not override it, so combining stays live for the whole suite. Cached per type; a missing method reads as overridden, which is the safe direction.

Listed in `UseCombinedCommands`' documentation alongside the query-hints case. Also recorded in `TestExtensions.UsesCombinedEagerLoading` that it does not mirror this gate, the probe being internal with no `InternalsVisibleTo` to reach it.

Source/LinqToDB/Internal/Linq/QueryRunner.cs
sdanyliv: **SUG001 - done in 032a0fe0c9.**

The execution context is already constructed before `BuildPlan`, so it is threaded down through `CollectParameterValues` into both `SetParameters` and `AddCombinableParameterValues`. `IStepMaterializer` gained the parameter and all three implementations forward it. Internal interface, so no public surface change.

Source/LinqToDB/Data/DataConnection.QueryRunner.cs
sdanyliv: **SUG002 - recorded rather than changed, in d1c4e9a299.** All three available fixes break something:

- A `SqlProviderFlags` capability is the honest one, but it adds public, wire-serialized surface for a path nothing reaches and nothing can validate it against. (This branch just spent a merge on a duplicate `DataMember` order in that very type.)
- Advancing unconditionally breaks SqlClient: after a `NonQuery` step that produced no result set it would skip the very result set the next step reads.
- Throwing when a `Scalar` step finds no row breaks the `InsertOrUpdate` emulation - `DmlServiceBase` uses `SqlStepKind.Scalar` for its EXISTS probe, where an empty result set is the legitimate "no row" answer.

So the walk is unchanged and the comment now carries the premise instead: that it is a driver behaviour no flag expresses, why it holds today (PostgreSQL registers no `IDmlService` override, renders `RETURNING`, and becomes a single step; the eager path is all `Reader` steps), how it fails (the `Scalar` step reads the INSERT's empty result set and its value comes back null), that the shortfall check catches the opposite direction, and the three rejected options with their reasons.

If you want the capability flag now rather than at the first multi-step scenario on such a provider, say so and I will add it.

Tests/Linq/Update/CombinedDmlTests.cs
sdanyliv: **MIN003 - fixed in d838ba3d1b**, as the self-calibrating guard.

The `[Values]` parameter is gone; each test runs both configurations itself and counts `CommandInitialized`, asserting `on <= off` always and `on == 1` only when combining is available and `off > 1`.

Verified non-vacuous on SQLite rather than assumed: the identity insert measures 2 -> 1 and takes the exact-one branch, while `InsertOrReplace` measures 1 -> 1 and correctly stands down, SQLite having a native upsert.

Only the third `InsertOrReplace` pass is counted - the earlier ones establish the row, and counting all of them would tie the expected number to how many passes the test happens to make. Noted at the site that registering the counter forces the concatenated backend, since `CanUseDbBatch` requires no command interceptor and a `DbBatch` would not report through `CommandInitialized` anyway.

Source/LinqToDB/LinqOptions.cs
```diff
@@ -221,7 +230,8 @@ public sealed record LinqOptions
 		UpsertEmulationPolicy     UpsertEmulationPolicy       = UpsertEmulationPolicy.Allow,
 		EagerLoadingStrategy      DefaultEagerLoadingStrategy = EagerLoadingStrategy.Default,
 		ImplicitCollectionLoading ImplicitCollectionLoading   = ImplicitCollectionLoading.Allow,
-		bool                      OptimizeForSequentialAccess = false
+		bool                      OptimizeForSequentialAccess = false,
+		bool                      UseDbBatch                  = true
```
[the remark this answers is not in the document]
sdanyliv: Not applicable, on the same evidence as the prior-review audit.

`PackageValidationBaselineVersion` resolves to `BaselineVersion`, which is `6.0.0` (`Directory.Build.props`), and the `[EditorBrowsable(Never)]` 15-parameter constructor and `Deconstruct` are that 6.0.0 signature. ApiCompat therefore compares against it and reports no break - neither the base nor the head `CompatibilitySuppressions.xml` carries a `LinqOptions` entry.

The intermediate 20-parameter signature a shim is asked for here was never a released baseline.

Source/LinqToDB/LinqOptions.cs
```diff
@@ -221,7 +250,9 @@ public sealed record LinqOptions
 		UpsertEmulationPolicy     UpsertEmulationPolicy       = UpsertEmulationPolicy.Allow,
 		EagerLoadingStrategy      DefaultEagerLoadingStrategy = EagerLoadingStrategy.Default,
 		ImplicitCollectionLoading ImplicitCollectionLoading   = ImplicitCollectionLoading.Allow,
-		bool                      OptimizeForSequentialAccess = false
+		bool                      OptimizeForSequentialAccess = false,
+		bool                      UseDbBatch                  = true,
+		bool                      UseCombinedCommands         = false
```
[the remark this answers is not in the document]
sdanyliv: Same as the earlier thread on this file: not applicable.

`PackageValidationBaselineVersion` resolves to `BaselineVersion` = `6.0.0`, and the `[EditorBrowsable(Never)]` 15-parameter constructor and `Deconstruct` are that signature, so ApiCompat compares against 6.0.0 and reports no break. The 20-parameter signature ending at `OptimizeForSequentialAccess` was never released, so nothing was ever compiled against it.

PR #5681, opened 2026-07-07
Add memory-pressure eviction to QueryCache

sdanyliv:
## What

Adds memory-pressure-aware purging to the process-wide LINQ query cache (`QueryCache.Default`): when the runtime reports high memory load, the cache trims its coldest entries (ordered by expiry → last-access → hit-rate) while preserving the hot working set — a safety valve on top of the existing idle-timeout and entry-count eviction.

## Why

`QueryCache` previously evicted only by sliding idle-timeout and an approximate 10k-entry cap; it had no awareness of actual process/host memory, so under pressure it held its full working set until the next timed sweep or the count cap fired.

## How

Three triggers share one decision (`IsUnderMemoryPressure()`) and one primitive (`TrimByFraction` → the existing `TrimGlobalToCapacity`):

- **Gen2 GC callback** — reacts the instant the GC collects under pressure (a faithful port of the runtime's `Gen2GcCallback`, armed on all TFMs).
- **Periodic backstop** — a memory check folded into the existing ~5-minute cache sweep (no new timer).
- **Manual `Compact(double)`** — on-demand trim, available on every target framework.

Trims the coldest 50% by default (overridable via `MemoryPressureTrimFractionOverride`), on by default, opt out via `Configuration.Linq.EnableQueryCacheMemoryPressureEviction`. The pattern mirrors `Microsoft.Extensions.Caching.Memory` (Gen2GcCallback + memory-load check → compact); the trim work is offloaded off the finalizer thread and shares the sweep's single-flight guard.

### Target-framework behavior

- **net8.0+** — automatic detection via `GC.GetGCMemoryInfo()`.
- **net462 / netstandard2.0** — no built-in memory-load API and **no platform-specific code**; automatic eviction happens only when the app supplies `MemoryPressureOverride` (e.g. a cgroup / `GC.GetTotalMemory` probe), otherwise behavior is unchanged. `Compact()` works everywhere. The Gen2 callback's finalizer guards `AppDomain.CurrentDomain.IsFinalizingForUnload()` on .NET Framework.

`MemoryPressureOverride` is invoked on the GC finalizer thread and the maintenance sweep, so it must be fast, non-blocking, and thread-safe; exceptions it throws are caught and treated as "not under pressure".

Scope is `QueryCache` only. The vendored `LinqToDB.Internal.Cache.MemoryCache` (backs the DML/CRUD and raw-SQL caches) has the same gap and is a possible follow-up.

## Public API (`Internal.*` surface)

- `QueryCache.MemoryPressureEvictionEnabled`, `MemoryPressureTrimFractionOverride`, `MemoryPressureOverride`, `Compact(double)`
- `Configuration.Linq.EnableQueryCacheMemoryPressureEviction`

## Tests

`QueryCacheEvictionTests` — 10 new deterministic tests (coldest-fraction trim, hot-entry survival, `Compact`, full clear, predicate-driven sweep, disabled no-op, fraction override, empty-cache, async Gen2-callback path, throwing-probe swallow). Full fixture **19/19 pass**. Release builds green on **net10.0 / net462 / netstandard2.0** (analyzers, public-API baseline, XML docs, cross-TFM fallback).


igor-tkachev (changes requested):
The feature is useful, but the current asynchronous pressure path can lose a detected memory-pressure event.

`OnGen2Gc()` checks `IsUnderMemoryPressure()` and then queues a normal maintenance sweep. The queued worker later calls `IsUnderMemoryPressure()` again inside `SweepGlobal()`. This means a single pressure event requires two successful checks. A stateful or time-sensitive `MemoryPressureOverride` can return `true` in the callback and `false` in the worker, resulting in no trim at all.

There is a second loss scenario: `QueueGlobalMaintenance()` simply returns when `_sweepRunning` is already set. If a Gen2 callback detects pressure while another sweep is running, the pressure request is discarded. The active sweep might have already passed its pressure check or might have started before pressure was detected.

Please preserve the detected pressure intent explicitly, e.g. with a pending flag. The Gen2 callback should set that flag and queue maintenance. The worker should atomically consume it and perform the pressure trim without requiring the signal to be detected again. If the callback fires while a sweep is running, the pending flag must survive and be handled before the worker fully releases the single-flight state or by a follow-up pass.

I also think `MemoryPressureOverride` should not be invoked on the finalizer thread. Calling arbitrary user code there is too risky: even an apparently simple implementation can block on a lock, perform I/O, allocate, log, or otherwise delay process-wide finalization. The Gen2 callback should remain a minimal notification mechanism, while the user-supplied probe runs on the thread-pool maintenance worker.

Additional API concerns:

- `MemoryPressureOverride` is not really an override of an existing value; `MemoryPressureDetector` or similar would be clearer.
- `MemoryPressureTrimFractionOverride` has production semantics despite the existing `*Override` naming being used for test/diagnostic settings.
- `Configuration.Linq.EnableQueryCacheMemoryPressureEviction` proxies the live `QueryCache.Default` singleton, while the other pressure settings are exposed only through `QueryCache.Default`. This creates an inconsistent configuration surface.
- Automatic Gen2 registration is performed only for `QueryCache.Default`, but other `QueryCache` instances default `MemoryPressureEvictionEnabled` to `true`. The documentation should make it explicit that non-default instances do not receive Gen2 callbacks and are checked only during their normal sweeps.

Tests should cover at least:

1. a one-shot/stateful pressure probe that returns `true` only once;
2. a pressure callback arriving while a sweep is already running;
3. multiple coalesced pressure events;
4. verification that the application probe is not executed on the finalizer thread after the architecture is adjusted.

The Gen2 callback implementation itself looks generally reasonable, but please also consider catching callback exceptions inside `Gen2GcCallback` itself rather than relying on every caller to do so, and add the exact runtime source/commit used for the port.

igor-tkachev (commented):
## Overview

Adds memory-pressure-aware eviction to the process-wide `QueryCache.Default`. Three triggers funnel into one gate (`IsUnderMemoryPressure`) and one trim primitive (`TrimByFraction` → the existing `TrimGlobalToCapacity`):

- A ported `Gen2GcCallback` (new file, `Source/LinqToDB/Internal/Cache/Gen2GcCallback.cs`) that fires once per Gen2 GC via a weak-`GCHandle` + self-re-registering finalizer, mirroring `Microsoft.Extensions.Caching.Memory`.
- A periodic check folded into the existing ~5-minute sweep.
- A manual `Compact(double)` API for host-supplied low-memory signals.

Net8.0+ gets automatic detection via `GC.GetGCMemoryInfo()` (new `SUPPORTS_GC_MEMORY_INFO` constant); net462/netstandard2.0 rely on an app-supplied `MemoryPressureOverride` probe. New public surface lives entirely under `LinqToDB.Internal.*` plus one `Configuration.Linq` toggle, and `PublicAPI.Unshipped.txt` is updated accordingly. 10 new tests cover the trim math, hot-entry survival, `Compact`, the predicate path, the disabled no-op, fraction overrides, empty-cache safety, the async Gen2 path, and probe-exception swallowing.

## Correctness / robustness findings

**1. `Gen2GcCallback` doesn't inherit `CriticalFinalizerObject` (Medium)**

`Source/LinqToDB/Internal/Cache/Gen2GcCallback.cs` declares `sealed class Gen2GcCallback` with no base class. The runtime implementation this is explicitly documented as "a faithful port of" (`Microsoft.Extensions.Caching.Memory.Gen2GcCallback`) inherits `System.Runtime.ConstrainedExecution.CriticalFinalizerObject`. That base class guarantees the finalizer runs to completion even during process teardown / AppDomain unload and is prioritized ahead of ordinary finalizers in the finalization queue — the reliability guarantee a perpetually-reregistering finalizer depends on. Without it, this port loses that guarantee, which matters most on the net462 target where this class is most likely to run inside a host with AppDomain recycling (IIS, SQL CLR). `CriticalFinalizerObject` is available on every targeted TFM, so recommend inheriting it to match the "faithful port" claim.

**2. Missing `GC.KeepAlive(this)` after `GCHandle.Alloc` (Low)**

The original `Setup()` calls `GC.KeepAlive(this)` right after `GCHandle.Alloc`. The port's constructor omits it. This guards a narrow but real hazard: in a fully optimized build, the JIT can consider `this` "dead" after its last field-write use, making the object eligible for premature collection/finalization while the constructor is still executing (if `GCHandle.Alloc` itself triggers a GC). Low probability, but it's the documented reason the original includes the call — worth restoring for parity, e.g. add `GC.KeepAlive(this);` as the last statement in the constructor.

Everything else in `Gen2GcCallback` — the deliberate omission of `GC.ReRegisterForFinalize` in the constructor (with the explanatory comment about double-finalization on .NET Framework), the double-finalization guard (`!_handle.IsAllocated`), the `IsFinalizingForUnload()` check gated to `NETFRAMEWORK`, and the `HasShutdownStarted` handling — is correct and matches the canonical pattern.

## `QueryCache` trim logic

- `ComputeFractionTarget` / `TrimByFraction` / the combined cap in `SweepGlobal` (`target = min(capacity cap, fraction target when under pressure)`) are correct: verified by hand against the new tests (e.g. 100 entries, 0.5 fraction → target 50 via `ceil(100*0.5)`; fraction 1.0 → target 0; NaN/`<=0` → no-op via the `!(fraction > 0)` guard).
- `PressureDetected` correctly treats a throwing `MemoryPressureOverride` as "not under pressure" and never lets the exception escape the finalizer thread or the sweep worker.
- The Gen2 callback path stays cheap on the finalizer thread — it only calls `QueueGlobalMaintenance`, which offloads the actual trim (sort + per-bucket lock removal) to the thread pool, sharing the existing single-flight guard (`_sweepRunning`). Good design; avoids the trim's O(n log n) work running on the finalizer thread.
- `Compact()` intentionally bypasses `_sweepRunning`, so it can race a concurrent automatic sweep and over-trim — this is explicitly documented in the XML doc, so it's a conscious trade-off, not an oversight.
- Minor, not worth a fix: `IsUnderMemoryPressure()` eagerly evaluates `NativeMemoryHigh()` (thus always calls `GC.GetGCMemoryInfo()`) before `PressureDetected` gets a chance to short-circuit on a user probe. Negligible cost given the call frequency (once per Gen2 GC / 5-minute sweep).

## Public API / conventions

- `Configuration.Linq.EnableQueryCacheMemoryPressureEviction` deliberately breaks from the `Options with { ... }` immutable-snapshot pattern used by sibling `Configuration.Linq` properties (e.g. `CacheSlidingExpiration`, `GuardGrouping`), delegating straight to the `QueryCache.Default` singleton instead. This is called out explicitly in the XML doc's `<remarks>` — appropriate given the feature is process-global rather than per-`DataOptions`, not an inconsistency to fix.
- `PublicAPI.Unshipped.txt` additions match the new members exactly; placement within the file looks consistent with the existing entries.
- New public members carry XML docs as required; no `CompatibilitySuppressions.xml` edits.
- Scope is correctly limited to `QueryCache`; the PR description flags that the vendored `MemoryCache` (DML/CRUD/raw-SQL caches) has the same gap and defers it as a follow-up rather than scope-creeping this PR.

## Test coverage

The 10 new tests in `QueryCacheEvictionTests.cs` are well-targeted: fraction trim math, hot-entry survival (via hit-rate/deadline ordering), `Compact` as the public entry point, full clear at fraction 1.0, the predicate-driven sweep path, the disabled no-op, the fraction override, empty-cache safety, the async Gen2-callback path (`FireGen2CallbackNow`, polled with a 2s bound), and throwing-probe swallowing. Existing tests were correctly updated to set `MemoryPressureEvictionEnabled = false` where they'd otherwise be flaky under a memory-pressure backstop firing mid-test on a constrained CI runner — a reasonable isolation fix.

One soft spot: `MemoryPressureTrim_Gen2CallbackPath_TrimsAsync` polls up to 2000ms on the real thread pool for the async trim to land. This is inherently less deterministic than the rest of the suite (which mostly uses synchronous test hooks); if it ever flakes in CI, that's the test to look at first.

## Summary

The design is sound and the risky part (thread-pool offload, single-flight sharing, probe exception-swallowing, NaN/negative-fraction guards) is handled carefully and tested. The one thing worth fixing before merge is reliability parity on the ported `Gen2GcCallback`: add `CriticalFinalizerObject` as the base class, and restore the `GC.KeepAlive(this)` call in the constructor. Both are small, mechanical changes that align the implementation with what it claims to port.

---
*Automated review via Claude Code.*

Source/LinqToDB/Internal/Cache/Gen2GcCallback.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
MaceWindu, line 2: **Minor · MIN001** — This new file reinstates a `.NET Foundation` MIT-license header carried over verbatim from the runtime original. linq2db has retired from the .NET Foundation and drops such claims across the codebase, and asserting that "the .NET Foundation licenses this file" is inaccurate for a linq2db port. MIT still requires preserving attribution for the ported portion, so the fix is to reword as adaptation attribution — which also records the exact upstream source the review requested elsewhere.

```suggestion
// Adapted from the .NET runtime's internal Gen2GcCallback (dotnet/runtime), MIT license.
// © .NET Foundation and contributors. https://github.com/dotnet/runtime
```

Source/LinqToDB/Common/Configuration.cs
```diff
@@ -413,6 +413,25 @@ public static TimeSpan CacheSlidingExpiration
...
+			/// target frameworks eviction fires only when a probe is supplied via
+			/// <c>QueryCache.Default.MemoryPressureOverride</c>.
+			/// </summary>
+			/// <remarks>
+			/// Unlike the other <c>Configuration.Linq</c> settings, this is a process-global toggle on the
+			/// <c>QueryCache.Default</c> singleton, not a per-<c>DataOptions</c> option — it is not captured
+			/// in a <c>LinqOptions</c> snapshot. Other <c>QueryCache</c> instances carry their own
+			/// <c>MemoryPressureEvictionEnabled</c> flag (also defaulting to <see langword="true"/>).
+			/// </remarks>
+			public static bool EnableQueryCacheMemoryPressureEviction
```
MaceWindu, line 429: **Minor · MIN002** — `Configuration.Linq.EnableQueryCacheMemoryPressureEviction` is the only non-`Internal.*` public entry point for this feature, and its get/set is a hand-written forwarder to `QueryCache.Default.MemoryPressureEvictionEnabled` — but nothing tests the delegation. The per-instance flag is covered (`MemoryPressureTrim_Disabled_NoOp`); an inverted or mis-targeted forwarder here would be caught by no test (the 10 new tests all use isolated `new QueryCache()` instances and never touch `Configuration.Linq` or `QueryCache.Default`).

Fix: add a test that flips this property and asserts `QueryCache.Default.MemoryPressureEvictionEnabled` tracks it both ways. Since it mutates the process-wide `Default` singleton, mark it `[NonParallelizable]` and restore the original value in a `finally`.

Tests/Linq/Linq/QueryCacheEvictionTests.cs
```diff
@@ -308,6 +332,224 @@ public void MaxEntriesOverride_ZeroDisablesCaching()
...
+			};
+
+			PopulateSpread(cache, db, 100);
+
+			// Fire the real Gen2 callback: OnGen2Gc -> QueueGlobalMaintenance -> thread-pool sweep -> trim.
+			cache.FireGen2CallbackNow();
+
+			// The trim lands on the thread pool; poll until it completes (bounded).
+			var sw = System.Diagnostics.Stopwatch.StartNew();
+			while (sw.ElapsedMilliseconds < 2000 && cache.ApproximateEntryCount > 50L)
```
MaceWindu, line 524: **Minor · MIN004** — `MemoryPressureTrim_Gen2CallbackPath_TrimsAsync` fires the real async path and polls the outcome with a hard 2000 ms deadline. The fixture is `[Parallelizable(ParallelScope.All)]`, so the rest of the assembly contends for the same thread pool; on a saturated CI runner the queued maintenance can be delayed past 2000 ms → spurious failure. Unlike the synchronous `IdleEviction_*` polls, this test can't re-drive the queued work, only wait for it.

Fix: raise the poll deadline so a thread-pool-starved runner still lands the trim. The assertion stays correct; only the flake window shrinks.

```suggestion
			while (sw.ElapsedMilliseconds < 10000 && cache.ApproximateEntryCount > 50L)
```

Tests/Linq/Linq/QueryCacheEvictionTests.cs
MaceWindu, line 1: **Minor · MIN003** — The NaN / non-positive fraction guards are load-bearing: `ComputeFractionTarget`'s `if (!(fraction > 0)) return -1;` and `ResolveTrimFraction`'s NaN→`0.50` fallback are what keep `Compact(NaN)` and a NaN override from *wiping* the cache. No test pins them — `MemoryPressureTrim_EmptyCache_DoesNotThrow` calls `Compact(0.5)` on an empty cache, which never exercises the guard. The `if (!(fraction > 0))` idiom is exactly what a refactor to `fraction <= 0` would silently break for NaN.

Fix: add (a) a populated-cache test calling `Compact(0)`, `Compact(-1)`, `Compact(double.NaN)` asserting the entry count is unchanged; (b) a test setting `MemoryPressureTrimFractionOverride = double.NaN`, firing a pressure check, asserting the default 50% trim rather than a full clear.

MaceWindu (commented):
> [!IMPORTANT]
> **Agentic review — treat with care.** This review was produced by an LLM agent rather than a human reviewer, and is **not** the final word: the PR will also receive a human review, so this complements — rather than replaces — human judgement. Individual findings can be wrong, overconfident, or miss context a human would catch. Feel free to disagree, dismiss, or ask for clarification on any finding.

## Review notes

This pass focuses on the genuinely-open items after the existing reviews: a license-header fix on the new ported file, one public-surface test gap, and two test-quality gaps (all posted inline). The architectural feedback already on the PR — the async pressure-path race, the finalizer-thread probe, and the naming concerns — is audited below rather than re-filed, since the code is byte-identical since those reviews.

## Prior-review audit

Only a `master` merge landed after the two prior reviews, so their points all remain **actual at HEAD**:

**Changes-requested review:**
- **Lost pressure event** (double-check race + signal dropped while a sweep is in flight) — **actual**. Both are real windows, but bounded and self-healing in my read: the double-check only loses the fraction trim when the memory level genuinely cleared between the two `IsUnderMemoryPressure()` calls (the correct outcome for a level probe); the in-flight-sweep drop can lose even a native-signal event, but the next Gen2 GC re-detects sustained pressure. Worth considering the proposed pending-pressure flag consumed by the worker, but not a correctness bug.
- **Probe on the finalizer thread** — **actual**, but a documented, conscious trade-off (the `MemoryPressureOverride` `<remarks>` warns it must be cheap/non-blocking and that a blocking probe stalls finalization). Design opinion, not a defect.
- **`MemoryPressureOverride` naming** — **actual**, concur: it's a detection probe, not a value override, so the `*Override` suffix collides with the value-knob convention of `MaxEntriesOverride` / `IdleTimeoutOverride`. (`MemoryPressureTrimFractionOverride` is fine — it genuinely overrides the `0.50` default.)
- **Config-surface inconsistency, non-default-instance Gen2 documentation gap, catching callback exceptions inside `Gen2GcCallback` itself, and citing the exact runtime source/commit for the port** — all **actual / unaddressed**; reasonable.

**Prior automated review:**
- **Missing `CriticalFinalizerObject` base** — **actual and accurate**: the canonical `Gen2GcCallback` inherits it; this port doesn't. Low practical impact here (idempotent, shutdown-guarded, self-catching), but it does contradict the "faithful port" wording.
- **Missing `GC.KeepAlive(this)`** — **inaccurate**. Verified against dotnet/runtime: the current canonical `Gen2GcCallback` (`System.Private.CoreLib`, mirrored by `Microsoft.Extensions.Caching.Memory`) has **no** `GC.KeepAlive` — `Register` is just `new Gen2GcCallback(...)` and the object is intentionally left unrooted so the finalizer mechanism works. Its omission is not a fidelity gap; that claim cites an older/different variant.

## Out-of-scope observations

Surfaced during review; FYI, not findings on this PR.

- **Enabled by default on net8.0+ (opt-out).** Existing net8+ consumers begin auto-trimming ~50% of cached queries once `MemoryLoadBytes >= HighMemoryLoadThresholdBytes` (default ~90% of the container-aware limit), driven by both the Gen2 callback and the periodic sweep. Intentional, but worth an explicit **release-note** mention — an app steadily near its cgroup limit could see its query cache repeatedly halved on each Gen2 GC, forcing recompilation.
- **Internal-only tuning surface.** The only public (non-`Internal.*`) surface is the enable/disable bool; the fraction and probe knobs live on `Internal.*` `QueryCache`, and `QueryCache.Default` is itself `Internal.*`. So net462 / netstandard2.0 users get no automatic eviction through public API alone. Coherent minimal-surface choice; worth a conscious decision if external pre-net8 use is intended.
- **netstandard2.0 asset would degrade two guards if cross-loaded** (native detection off, AppDomain-unload guard absent) — unreachable in practice via NuGet asset selection, which hands net8+/net462 consumers the matching build.
- **`RunMemoryPressureCheckNow` (`#if BUGCHECK`) doc slightly overclaims** "the exact production gate": it mirrors the pressure gate but skips the `min(ResolveMaxEntries(), fractionTarget)` combination `SweepGlobal` applies (they coincide only because the tests set `MaxEntriesOverride = 10000`). Test-only, no shipping impact.

## Baselines

No baseline changes (no baselines branch for this PR — expected for a change with no SQL surface).

MaceWindu (changes requested):
Let's not rush and instead of moving in different directions design and implement library wide solution. Moved both PRs from release

Issue #2432, opened 2020-08-20
linq2db.SQLite.MS: variety of problems caused by wrapping values in DateTime() function


Edward-Alchikha:
Hello again. 

The first problem is that unfortunately, 3.1.0 doesn't fix the problem I had previously in #2346. When setting up my DataConnection like

```csharp
            var ms = new MappingSchema();

            ms.SetDataType(typeof(DateTime), DataType.Int64);
            ms.SetConvertExpression(
                (long ticks) => new DateTime(ticks, DateTimeKind.Unspecified)
            );
            ms.SetConvertExpression(
                (DateTime dt) => dt.Ticks
            );
            ms.SetValueToSqlConverter(
                typeof(DateTime),
                (sb, dType, v) => sb.Append(((DateTime) v).Ticks.ToString(CultureInfo.InvariantCulture))
            );

            await using var conn = new DataConnection(ProviderName.SQLiteMS, connectionString, ms)
            {
                InlineParameters = true
            };
```

I still end up getting queries like

```sql
SELECT
        [x].[DateTime],
        [x].[Value]
FROM
        [fae5697f79034df99e0ad32292446f5a_testOut] [x]
WHERE
        DateTime([x].[DateTime]) > DateTime(637304378093620000) AND
        DateTime([x].[DateTime]) < DateTime(637304382196380000)
```

produced (which will always return 0 rows), whereas I should be getting

```sql
SELECT
        [x].[DateTime],
        [x].[Value]
FROM
        [fae5697f79034df99e0ad32292446f5a_testOut] [x]
WHERE
        [x].[DateTime] > 637304378093620000 AND
        [x].[DateTime] < 637304382196380000
```

<br />
<br />

The second problem is that even when not trying to store my DateTime values as an Int64, and just using this to set up my DataConnection

```csharp
            await using var conn = new DataConnection(ProviderName.SQLiteMS, connectionString)
            {
                InlineParameters = true
            };
```

wrapping the values in the DateTime() function also causes indexes to not work, which is a huge performance issue for me. It produces queries like

```sql
SELECT
        [x].[DateTime],
        [x].[Value]
FROM
        [d7e9196fe84847faabf1a02518040951_testOut] [x]
WHERE
        DateTime([x].[DateTime]) > DateTime('2020-08-27 19:57:00.035') AND
        DateTime([x].[DateTime]) < DateTime('2020-09-04 01:06:43.64')
```

whereas if it produced queries like

```sql
SELECT
[...]

Issue #3766, opened 2022-09-08
Issue - SQLite.Classic - Using datatimeoffset as primary key


j-dc:
### Describe your issue

Using a DateTimeoffset as a primary key behaves unpredictable when using SQLite.Classic.
When using the MS implementation this is not a problem.

### Steps to reproduce

Create a table with a datetime as Primarykey. 
Do an insert.
Do an InsertOrReplace (with existing key)
=> System.Data.SQLite.SQLiteException: constraint failed
UNIQUE constraint failed: MyTimedTable.Key, MyTimedTable.Timestamp

A full code example can be found on:
https://github.com/j-dc/issues.LinqToDB.DatetimeOffset_PrimaryKey

There are 6 tests:   (I tried a custom valueconverter, but to no use)
- LinqToDB 3.7 (as my project where I ran into this issue is using this version)
  - Using SQLite.classic
      -  With custom ValueConverter (FAIL)
      -  Without custom valueconvert (FAIL)
  - using SQLite microsoft
      -  With custom ValueConverter (FAIL)
      -  Without custom valueconvert  (OK!!!)
- LinqToDB 4.2 (to check if is solved in latest version)
  - Using SQLite.classic
      -  With custom ValueConverter (FAIL)
      -  Without custom valueconvert (FAIL)

`Linq To DB 3.7` `Linq To DB 4.2` 

Database (with version): SQLite

ADO.NET Provider (with version): SQLite.Classic SQLite

Operating system: Windows

.NET Version: .net 6

(I cannot switch to the ms provider because of: https://github.com/linq2db/linq2db/issues/3330 )

I can workaround this by saving the datetimeoffset as string.. but I don't like it..  (but it works)

Issue #1855, opened 2019-08-09
SQLite is not handling DateTimeOffset queries well


to11mtm:
The SQL that is being generated in/around `DateTimeOffset` (Both nullable and non) in SQLite does not work under `Sql.DateAdd` or `!=` comparisons.

### Steps to reproduce

See Below Repro which shows tests passing under Sql Server LocalDB and failing under System.Data.SQLite:

```c#
using System;
using System.Collections.Generic;
using System.Data.SQLite;
using System.Diagnostics;
using System.Linq;
using System.Text;
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;
using Xunit;
using Xunit.Abstractions;

namespace Repro
{
    public class DateTimeOffsetCompatibilityTests
    {
        public class TestTable
        {
            [PrimaryKey]
            [Identity]
            public int Id { get; set; }
            public Guid ConfirmationGuid { get; set; }
            public DateTimeOffset SomeDateTimeOffset { get; set; }
            public DateTimeOffset? SomeNullableDateTimeOffset { get; set; }

        }
        public DateTimeOffsetCompatibilityTests(ITestOutputHelper outputHelper)
        {
            LinqToDB.Data.DataConnection.TurnTraceSwitchOn();
            LinqToDB.Data.DataConnection.OnTrace = info =>
            {
                if (info.TraceInfoStep == TraceInfoStep.BeforeExecute)
                {
                    outputHelper.WriteLine(info.SqlText);
                }
                else if (info.TraceLevel == TraceLevel.Error)
                {
                    var sb = new StringBuilder();

                    for (var ex = info.Exception; ex != null; ex = ex.InnerException)
                    {
                        sb
                            .AppendLine()
                            .AppendLine("/*")
                            .AppendLine($"Exception: {ex.GetType()}")
                            .AppendLine($"Message  : {ex.Message}")
                            .AppendLine(ex.StackTrace)
                            .AppendLine("*/")
                            ;
                    }

                    outputHelper.WriteLine(sb.ToString());
                }
                else if (info.RecordsAffected != null)
                {
                    outputHelper.WriteLine($"-- Execution time: {info.ExecutionTime}. Records affected: {info.RecordsAffected}.\r\n");
                }
                else
                {
                    outputHelper.WriteLine($"-- Execution time: {info.ExecutionTime}\r\n");
                }
            };
        }
[...]

PR #5682, opened 2026-07-07
Fix SQLite datetime rendering losing sub-millisecond ticks

myieye:
SQLite `DateTime`/`DateTimeOffset` parameters and inline literals are rendered with `.fff`, truncating sub-millisecond ticks. Two defects:

1. **Round-trip loss.** Values written through linq2db silently lose sub-millisecond ticks. `Microsoft.Data.Sqlite` (and EF Core on top of it) store full tick precision (`yyyy-MM-dd HH:mm:ss.FFFFFFF`).
2. **Comparisons misclassify rows stored at full precision** (e.g. written by `Microsoft.Data.Sqlite`). Comparison operands are normalized via `strftime('%Y-%m-%d %H:%M:%f', ...)`, and SQLite's date parser *rounds* to the nearest millisecond while `.fff` rendering *truncates*. For a stored value with sub-millisecond part ≥ 0.5 ms (e.g. `2020-02-29 17:54:55.00065`), the stored side normalizes to `.001` but a parameter holding the same instant renders as `.000` — equality misses, and the row compares strictly *greater than* its own value.

Fix: keep the existing 3-digit millisecond rendering (byte-identical output for every millisecond-grain value — no shape change for existing data, baselines, or raw-text comparisons such as unwrapped `IN` list columns) and append the sub-millisecond digits only when present: `.500` stays `.500`, `.1231234` no longer collapses to `.123`. One shared helper (`SQLiteMappingSchema.FormatDateTime`) now feeds both `SetParameter` and literal rendering.

Notes:

- Deliberately *not* `FFFFFFF` (Microsoft.Data.Sqlite's exact shape): trimmed text (`.1` vs `.100`) breaks raw-text comparisons against strftime-normalized operands — `Tests.Linq.TypesTests.DateTimeArray1` catches exactly that.
- `DataType.Date` and `DateOnly` rendering (`yyyy-MM-dd`) unchanged.
- Related to #2432 but independent: value rendering only, no change to comparison translation (the strftime wrapping remains).

Tests: round-trip of sub-millisecond ticks for `DateTime` (parameter + inline) and `DateTimeOffset` (inline, incl. non-zero offset), and comparison agreement against full-precision stored text. All red on master (14/14 cases), green here. Full `Tests/Linq` on `SQLite.MS,SQLite.Classic`: 20448 tests, 0 failures.

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteDataProvider.cs
+++ b/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteDataProvider.cs
@@ -287,7 +287,7 @@ namespace LinqToDB.Internal.DataProvider.SQLite
 
 			if (value is DateTime dt)
 			{
-				value = dt.ToString("yyyy-MM-dd HH:mm:ss.fff", DateTimeFormatInfo.InvariantInfo);
+				value = SQLiteMappingSchema.FormatDateTime(dt);
 				if (string.Equals(Name, ProviderName.SQLiteClassic, StringComparison.Ordinal))
 					dataType = dataType.WithDataType(DataType.VarChar);
 			}
--- a/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteMappingSchema.cs
+++ b/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteMappingSchema.cs
@@ -65,12 +65,14 @@ namespace LinqToDB.Internal.DataProvider.SQLite
 		}
 
 		// Always 3 millisecond digits, plus sub-millisecond digits only when present.
+		// The output is exact: .NET's ".fff" truncates rather than rounds, so the appended
+		// ticks are precisely the digits it dropped.
 		// Not fewer digits (".5"): strftime('%f') emits exactly 3 digits and existing
 		// linq2db-written data stores exactly 3, and raw-text comparisons (e.g. the unwrapped
 		// column side of an IN-list predicate) need those shapes to match.
-		// Not truncated (".123" for ".1236"): comparison operands are normalized via strftime,
-		// which rounds to the nearest millisecond — stored ".1236" normalizes to ".124" while a
-		// truncated ".123" stays ".123", so the same instant compares unequal.
+		// Not plain ".fff" (".123" for ".1236"): SQLite normalizes comparison operands via
+		// strftime, which rounds to the nearest millisecond — stored ".1236" normalizes to
+		// ".124" while a truncated ".123" stays ".123", so the same instant compares unequal.
 		internal static string FormatDateTime(DateTime value)
 		{
 			var result = value.ToString("yyyy-MM-dd HH:mm:ss.fff", DateTimeFormatInfo.InvariantInfo);
--- a/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteMappingSchema.cs
+++ b/Source/LinqToDB/Internal/DataProvider/SQLite/SQLiteMappingSchema.cs
@@ -11,15 +11,11 @@ namespace LinqToDB.Internal.DataProvider.SQLite
 {
 	public sealed class SQLiteMappingSchema : LockedMappingSchema
 	{
-		internal const string DATE_FORMAT_RAW                   = "yyyy-MM-dd";
+		internal const string DATE_FORMAT_RAW = "yyyy-MM-dd";
 #if SUPPORTS_COMPOSITE_FORMAT
-		private static readonly CompositeFormat DATE_FORMAT           = CompositeFormat.Parse("'{0:yyyy-MM-dd}'");
-		private static readonly CompositeFormat DATETIME_FORMAT       = CompositeFormat.Parse("'{0:yyyy-MM-dd HH:mm:ss.fff}'");
-		private static readonly CompositeFormat DATETIMEOFFSET_FORMAT = CompositeFormat.Parse("'{0:yyyy-MM-dd HH:mm:ss.fffzzz}'");
+		private static readonly CompositeFormat DATE_FORMAT = CompositeFormat.Parse("'{0:yyyy-MM-dd}'");
 #else
-		private  const string DATE_FORMAT           = "'{0:yyyy-MM-dd}'";
-		private  const string DATETIME_FORMAT       = "'{0:yyyy-MM-dd HH:mm:ss.fff}'";
-		private  const string DATETIMEOFFSET_FORMAT = "'{0:yyyy-MM-dd HH:mm:ss.fffzzz}'";
+		private const string DATE_FORMAT = "'{0:yyyy-MM-dd}'";
 #endif
 
 		SQLiteMappingSchema() : base(ProviderName.SQLite)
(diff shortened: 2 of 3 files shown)


sdanyliv (approved):

MaceWindu (changes requested):
Problem is that in sqlite DATE precision is exactly 3. You can store strings with any precision and format, but any fractional values less than millisecond will be truncated (not rounded) by sqlite when it converted to DATE.

MaceWindu:
To clarify: because sqlite has "interesting" type system and aproviders like SDS and MDS doesn't make it easier by implementing incompatible with each other custom types conventions, we need a good upfront design that address known issues and doesn't break compatibility for users.

Open issues:
- #1855
- #2107
- #2432
- #3766

igor-tkachev (changes requested):
## ⚠️ Compatibility policy violation — this change cannot be accepted as-is

This PR changes observable provider behavior globally and unconditionally. It changes how SQLite `DateTime` and `DateTimeOffset` values are serialized, stored, compared, and potentially used in keys, joins, `IN` predicates, snapshots, diagnostics, and provider-specific workarounds.

Such changes must not be introduced as mandatory behavior for all users.

### Required compatibility rule

Any feature or fix that changes generated SQL, parameterization, value serialization, provider behavior, or any other externally observable query behavior **must be introduced behind an explicit compatibility option that is disabled by default**.

The rollout policy should be:

1. Introduce the new behavior as opt-in and keep the current behavior as the default.
2. Allow users to enable it explicitly and validate it in real applications, provider combinations, schemas, tests, and production workloads.
3. Keep the fallback option available while reported compatibility problems are addressed.
4. Only after the new behavior has demonstrated sufficient stability across several releases may we consider enabling it by default.
5. Removal of the option should be considered only after the default-on behavior has itself been proven stable for a meaningful period.

This rule applies even when the new behavior appears logically more correct. Logical correctness does not guarantee compatibility with existing data, SQL snapshot tests, query plans, indexes, provider conventions, key semantics, diagnostics, or application-specific workarounds.

Without a compatibility option, any regression leaves users with no practical mitigation until a new linq2db release is produced. With an opt-in option, the implementation can be validated in real-world scenarios without forcing an irreversible behavioral change on the entire user base.

For this PR, the existing millisecond-based SQLite formatting must remain the default. Full sub-millisecond serialization may be added only as an explicitly enabled provider option. The comparison, storage, key, and interoperability semantics must also be tested under that mode before any future decision to enable it by default.

This is a blocking requirement.

igor-tkachev (commented):
## Technical review

The compatibility-policy blocker is covered by my previous `Request changes`. This review focuses on the technical design and test coverage of the proposed implementation.

### 1. Storage precision and comparison precision are inconsistent

The PR starts storing distinct values such as:

```text
2020-02-29 17:54:55.00060
2020-02-29 17:54:55.00065
```

but the current comparison translation still relies on SQLite date functions and `strftime('%f', ...)`, which operate at millisecond resolution. This means linq2db could store and read two distinct .NET values exactly, but compare them as the same value in LINQ predicates.

This is the main semantic issue. A round-trip guarantee is not sufficient if equality and ordering do not preserve the same precision.

### 2. The critical regression case is not tested

The current round-trip test reads rows by integer `Id`, so it proves only serialization/deserialization. The comparison test checks a value against itself, but not two different sub-millisecond values within the same millisecond.

Please add a test with values such as:

```csharp
var first  = wholeSecond.AddTicks(6000);
var second = wholeSecond.AddTicks(6500);
```

Both should be inserted and queried independently using:

- equality;
- `<` and `>`;
- parameters;
- inline values;
- SQLite Classic;
- SQLite Microsoft.

The expected semantics of those operations must be defined before this mode can be considered supported.

### 3. `DateTimeOffset` is changed only for inline literals

`SetParameter()` contains special formatting only for `DateTime`. `DateTimeOffset` parameters continue to use the convention of the underlying ADO.NET provider, while inline values use the new linq2db formatter.

That leaves different serialization behavior for:

- parameterized vs inline queries;
- SQLite Classic vs Microsoft.Data.Sqlite;
- regular insert vs other write paths.

This is especially dangerous for keys, equality predicates, and `InsertOrReplace`. Either the new mode must cover all supported paths consistently, or the PR scope and claims must be reduced accordingly.

### 4. All write paths must use the same convention

The new mode should be verified for at least:

- regular `Insert`;
- fluent insert;
- `Update`;
- `InsertOrReplace`;
- `BulkCopy` modes;
- inline literals;
- parameters;
- `DateTime`;
- `DateTimeOffset`.

SQLite already has known inconsistencies between regular insert, bulk copy, Classic, and Microsoft provider conventions. A new storage format should not add another partially applied convention.

### 5. Temporal key scenarios are missing

Please add tests where `DateTime` and `DateTimeOffset` participate in:

- primary keys;
- unique indexes;
- composite keys;
- `InsertOrReplace`;
- update/delete by key;
- equality joins.

Reading a temporal value by an unrelated integer `Id` does not prove that the representation is safe for key semantics.

### 6. `IN`/`Contains` and raw-text comparison need explicit tests

The formatter comment specifically references raw-text `IN` comparisons as a reason for keeping three mandatory millisecond digits, but the added tests do not directly cover sub-millisecond values in `Contains`/`IN` predicates.

Please add explicit tests for:

```csharp
var values = new[] { first, second };
var query  = table.Where(x => values.Contains(x.DateTime));
```

including multiple values that fall into the same millisecond after SQLite normalization.

### 7. Mixed-format databases need coverage

The database can contain values written by:

- older linq2db versions;
- the new opt-in mode;
- Microsoft.Data.Sqlite;
- System.Data.SQLite;
- EF Core;
- raw SQL;
- values with no fractional part, `Z`, or explicit offsets.

Please verify equality, ordering, grouping, distinct, joins, min/max, and index usage against mixed-format data. The issue is not only whether one representation round-trips, but whether all supported representations interoperate predictably.

### 8. The compatibility option requires an architectural solution for inline literals

`SQLiteMappingSchema` uses a shared static mapping schema, while `SetParameter()` has access to the current `DataConnection` and its options. A compatibility option cannot be implemented only in `SetParameter()`, because inline conversion would still be unconditional.

The option must affect both parameter and inline conversion and must participate in the provider/query configuration identity. This likely requires one of the following:

- option-specific mapping schemas;
- moving literal formatting to a layer with access to `DataOptions`;
- another provider-level mechanism that keeps parameter and inline behavior synchronized.

### 9. The formatter itself can be simplified

The manual tick extraction:

```csharp
var subMillisecondTicks = (int)(value.Ticks % TimeSpan.TicksPerMillisecond);
```

can be replaced by the standard format string:

```csharp
"yyyy-MM-dd HH:mm:ss.fffFFFF"
```

This keeps three mandatory millisecond digits and appends up to four significant sub-millisecond digits without manual arithmetic or `TrimEnd()`.

This is a minor implementation detail compared to the semantic issues above.

### 10. The formatter comment describes provider policy, not local implementation

The long comment above `FormatDateTime()` explains comparison translation, `strftime`, `IN`, and rounding behavior. Most of that belongs in the option documentation, provider documentation, or focused tests. The local code comment should describe only the formatting contract.

## Required direction

The PR can proceed only after the new behavior is introduced as disabled-by-default provider mode and the following are defined and tested consistently:

1. storage format;
2. parameter format;
3. inline-literal format;
4. comparison precision;
5. key semantics;
6. bulk-copy behavior;
7. Classic/Microsoft provider interoperability;
8. mixed old/new data behavior.

Without that contract, this change fixes one self-comparison scenario but introduces a provider mode where values can round-trip exactly while remaining indistinguishable in SQL predicates.

igor-tkachev:
We also need to test and define the upgrade behavior for existing databases.

Existing rows written by linq2db use millisecond precision. After enabling the new format, the same original .NET value will be serialized differently:

```text
old stored value: 2020-02-29 17:54:55.123
new parameter:    2020-02-29 17:54:55.1234567
```

A query that previously matched the old row could stop matching it after the linq2db upgrade, particularly in SQL paths that compare the stored text directly rather than normalizing both operands through SQLite date functions.

This also affects temporal keys, `InsertOrReplace`, updates, deletes, joins, and `IN` predicates. A value written by an older linq2db version could be treated as a different key when the same original .NET value is used by the new version.

The database file itself is not rewritten, so enabling this behavior creates a mixed-format database: legacy millisecond values and new tick-precision values coexist in the same columns. We need an explicit and tested policy for this scenario.

Please add upgrade-compatibility tests that insert legacy-formatted values and query or modify them using the new mode. The expected semantics must be documented. In particular, we need to decide whether a legacy `.123` value represents the exact value `.1230000` or an unknown value truncated to the `.123` millisecond.

This is another reason why the new storage format must remain opt-in and must not be enabled globally for existing databases.

igor-tkachev:
After considering the upgrade and interoperability implications, I think this should not be treated as a temporary compatibility switch with a possible path to default-on behavior.

This is an alternative SQLite temporal serialization convention, and it should remain a permanent opt-in provider option.

The existing `.fff` format must remain the default indefinitely. The new format should only be enabled explicitly by applications that fully control their SQLite schema, existing data, and all data writers.

The option should be documented as changing the physical text representation of `DateTime` and `DateTimeOffset` values, not as providing a general equivalent of a precision-aware server type such as `datetime2(7)`. SQLite does not enforce temporal precision or provide consistent sub-millisecond semantics across storage, date functions, comparisons, keys, indexes, providers, and externally written data.

The documentation must contain an explicit warning that enabling this option:

- does not migrate existing values;
- can create mixed-format columns in existing databases;
- can change equality, ordering, `IN`, join, key, update, delete, and `InsertOrReplace` behavior;
- may be incompatible with previous and future linq2db versions using another convention;
- may be incompatible with `Microsoft.Data.Sqlite`, `System.Data.SQLite`, EF Core, other ORMs, raw SQL, and custom converters;
- does not guarantee exact sub-millisecond comparisons through SQLite date/time functions;
- requires the application developer to ensure that all readers and writers use a compatible convention.

A suitable warning would be along these lines:

> This option changes the physical text representation of temporal values. It is intended only for applications that fully control the SQLite schema, existing data, and all data writers. linq2db does not guarantee compatibility with other linq2db versions, ADO.NET providers, ORMs, raw SQL, custom converters, or SQLite date/time functions when this option is enabled.

Because these risks are intrinsic to SQLite's type system and interoperability model, this option should never be enabled by default, regardless of how stable the implementation appears in linq2db tests.

The PR should also be positioned accordingly: it adds an optional alternative temporal text format, rather than universally fixing SQLite `DateTime` precision.

myieye:
Thanks guys for the thorough review.
There's obviously a lot more to consider than I anticipated.
Considering even this change wouldn't get me all the way to where I want to be, I'll just revert to my workaround.
ef-core respects the ticks in all scenarios, so I'll make sure I don't use linq2db for those queries.

PR #5686, opened 2026-07-08
SqlRawSqlTable: preserve IsScalar in copy ctor (correctness)

MaceWindu:
## What

`SqlRawSqlTable`'s copy ctor `(SqlRawSqlTable, ISqlExpression[])` copied fields, SQL and parameters but **not `IsScalar`**, so a rebuilt instance silently became non-scalar. That ctor is invoked by `QueryElementVisitor.VisitSqlRawSqlTable` (Transform mode) whenever a raw table is rebuilt during a clone/convert pass.

## Why it matters

If a scalar raw table loses `IsScalar`, the SQL builder skips the scalar column-alias list (the `t1(value)` after the derived source), leaving the projection's `t1.value` reference dangling (`column t1.value does not exist`).

## Reachability

**Probably not reachable from a LINQ query on current master** — this is a correctness/invariant fix, not a repro-backed bug fix. Plain / CTE / set-operation / cross-join / correlated-LATERAL scalar `FromSql` all render and execute correctly today, so no query path exercises the loss. The clone ctor should preserve the flag regardless, so a future clone path can't silently regress.

(Distinct from the parallel-only `column t1.value does not exist` failure under investigation elsewhere, which is a scalar-type *misclassification* race, not this copy-ctor loss.)

## Test

`SqlRawSqlTableTests.CopyCtor_PreservesIsScalar` drives the copy ctor directly via a deep `Clone()` and asserts `IsScalar` survives (red before the fix, green after). Provider-independent AST-level test.

## Follow-up commit

Refactored per review: instead of patching the copy ctor, **removed** it and rebuild the table through an explicit `internal` ctor that lists every field (`alias, objectType, fields, sql, isScalar, parameters`). `QueryElementVisitor` now spells out each argument, so a field can't be silently dropped again, and the rebuild fixes `SqlTableType` to `RawSql` (the copy ctor left it `Table`). Removes the public `SqlRawSqlTable(SqlRawSqlTable, ISqlExpression[])` ctor; `CompatibilitySuppressions.xml` regenerated (Internal.* CP0002).

### DIFF
--- a/Source/LinqToDB/CompatibilitySuppressions.xml
+++ b/Source/LinqToDB/CompatibilitySuppressions.xml
@@ -975,6 +975,13 @@
   </Suppression>
   <Suppression>
     <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:LinqToDB.Internal.SqlQuery.SqlRawSqlTable.#ctor(LinqToDB.Internal.SqlQuery.SqlRawSqlTable,LinqToDB.Internal.SqlQuery.ISqlExpression[])</Target>
+    <Left>lib/net10.0/linq2db.dll</Left>
+    <Right>lib/net10.0/linq2db.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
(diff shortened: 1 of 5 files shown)


sdanyliv (dismissed):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `7844582`.</sub>

PR #5689, opened 2026-07-08
Firebird: add scoped FirebirdTools.ClearPool(connection)

MaceWindu:
## What

Adds a **scoped** Firebird pool-clear — `FirebirdTools.ClearPool(DbConnection)` (+ `FirebirdProviderAdapter.ClearPool`) wrapping `FbConnection.ClearPool(FbConnection)`, mirroring the existing YDB `ClearPool`. Switches the test suite's Firebird temp-table teardown to it instead of the process-wide `ClearAllPools()`.

## Background

Firebird serializes DDL against object references: `DROP TABLE` fails with `object TABLE is in use` if any attachment still references it. With `FirebirdSql.Data.FirebirdClient`'s default connection pooling, a closed-but-pooled connection keeps that reference alive, so a subsequent `CREATE`/`DROP` — **even single-threaded and sequential** — hits the lock. The test suite has long worked around this by calling `FirebirdTools.ClearAllPools()` after every Firebird temp-table dispose.

## Problem (parallel execution)

`FbConnection.ClearAllPools()` is process-global across every connection string. Running the Firebird 2.5/3/4/5 suites concurrently (each version is a distinct connection string on its own parallel lane), one version's temp-table teardown calls `ClearAllPools()` and forcibly closes the **other** versions' in-use connections → spurious lock/connection failures. The single-threaded workaround became the parallel failure cause.

## Fix

Scoped `ClearPool(connection)` evicts only the current database's pool — releasing that version's dropped-table reference (keeps the DDL-lock mitigation) while leaving the concurrently-running other-version pools untouched.

- `FirebirdProviderAdapter.ClearPool` + `FirebirdTools.ClearPool(DbConnection)` (public API; `PublicAPI.Unshipped` updated).
- `TestUtils` `FirebirdTempTable.Dispose`/`DisposeAsync` + `ClearDataContext`, and one `JoinTests` cleanup: capture the connection before `Close()`, scoped-clear after.
- `MiniProfilerTests`'s `ClearAllPools()` left as-is — that call deliberately smoke-tests the `ClearAllPools` API resolves.

Real verification is the Firebird CI matrix (all versions run concurrently).

## Follow-up commit

Scoped `ClearPool(connection)` only fired when a live `DataConnection` was available. On the **LinqService (remote)** path `DataContext` isn't a `DataConnection`, so no clear ran and the next DDL hit `lock conflict on no wait transaction` / `object TABLE is in use` — the FB 3/4/5 LinqService failures in build 22015.

All three Firebird teardown sites (`FirebirdTempTable.Dispose`/`DisposeAsync`, `ClearDataContext`) now route through a `ClearFirebirdPool` helper: clear by the live connection when present (direct path), else by connection string — `ClearPool(string)` via `GetConnectionString(config.StripRemote())` — on the remote path. The `ClearAllPools()` fallback is dropped entirely (process-wide; tears down other versions' pools under parallel execution).

Verified red→green locally on `ComplexTests2` + `--provider Firebird.5` (FB5, incl. LinqService): unfixed 3 failed → fixed 0 failed.

### DIFF
--- a/Source/LinqToDB/DataProvider/Firebird/FirebirdTools.cs
+++ b/Source/LinqToDB/DataProvider/Firebird/FirebirdTools.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Data.Common;
 using System.Reflection;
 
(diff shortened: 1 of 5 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

Added `FirebirdTools.ClearPool(DbConnection)` and `FirebirdTools.ClearPool(string connectionString)` to clear the connection pool for a single Firebird database, complementing the existing process-wide `FirebirdTools.ClearAllPools()`. Use it to release one database's pooled connections — e.g. after a `DROP`/`CREATE` that Firebird would otherwise block with *object TABLE is in use* — without evicting every other Firebird pool in the process.


_(none)_


---
<sub>Generated from commit `d5b4ea0`.</sub>

PR #5690, opened 2026-07-08
perf: visitor pooling, drop dead EqualsTo threading + allocation cleanups

MaceWindu:
Performance/cleanup changes extracted from the #5364 (connection-leak) investigation WIP. Unrelated to the leak fix itself — split out so they can be reviewed and merged on their own.

## Changes

- **`EqualsToVisitor`**: drop the `EqualsToInfo` / `dataContext` / `compareConstantValues` context that was threaded through every comparison but never read (constant comparison ignored it). The `SqlQueryDependent` argument comparison now uses `ExpressionsEqual<object?>(null, …)`. No behavioural change — `compareConstantValues` was already dead.
- **Visitor pooling**: `ObjectPool<T>` is now `public` (the EF Core assembly can't see `LinqToDB` internals) and `TransformExpressionVisitor` instances are pooled. Reuse is safe — its `Cleanup()` already resets `DataContext`/`Tracking`/`IgnoreTracking` and is called on both rent and return.
- **`ExpressionVisitorBase`**: cache the `base.Visit` delegate instead of allocating it per `Visit`.
- **`TraceInfo`**: skip `PrintParameters` when the command has no parameters.
- **`CreateLinqToDBContext`**: remove the dead commented-out fallback branch.

## API surface

Additive, `Internal.*` only (per the repo's public-`Internal`-namespace convention), no binary/source break:
- `ObjectPool<T>` and its `RentedElement` become public.

`PublicAPI.Unshipped.txt` updated accordingly.

## Follow-up

The EF Core mapping-schema caching change (`EFCoreMappingSchema : LockedMappingSchema`) was split out to #5695 (WIP). `LockedMappingSchema` keys cache identity on `GetType()`, so all EF models collapsed to one query-cache identity → cross-model contamination that failed the entire EF.Core CI matrix. It needs a per-model identity and a benchmark before it can land.

PR #5695, opened 2026-07-09
perf: cache EF mapping schema by model-cache key (fix query-cache misses)

MaceWindu:
Extracted from #5690. Fixes the EF Core query-cache-miss the mapping-schema change there was meant to address — this time without collapsing distinct models.

## Problem

The linq2db mapping schema built from an EF Core model is created per `DbContext` and cached (`_schemaCache`) keyed on the **`IModel` instance** plus per-context service instances. When EF doesn't share an internal service provider across contexts (pooled contexts, `EnableServiceProviderCaching(false)`, multiple providers), each context gets a **fresh `IModel`** → a fresh `MappingSchema` with a fresh `ConfigurationID`. linq2db's query cache keys on that `ConfigurationID`, so every context recompiled its queries (cache miss on each one).

(#5690's first attempt — `EFCoreMappingSchema : LockedMappingSchema` — over-corrected: `LockedMappingSchema` keys identity on `GetType()`, so *all* EF models collapsed to one identity → cross-model contamination that failed the entire EF.Core CI matrix. Reverted there.)

## Fix

Cache the mapping schema at the `LinqToDBForEFTools` layer keyed on **EF's own model-cache key** (`IModelCacheKeyFactory.Create` — the same key EF uses to cache the `IModel`):

- Same model (same context type / custom `IModelCacheKeyFactory`) → equal key → one shared schema instance → one `ConfigurationID` → **query-cache hits**.
- Different model → different key → different schema → **no cross-model contamination**.
- The no-`accessor` path (`EFCoreMetadataReader`) falls back to the previous behavior.

No public API change (`GetMappingSchema` signature unchanged; new members are private), no `LockedMappingSchema` involvement, no baseline updates.

## Verification

- New regression test `MappingSchemaCacheTests` asserts a stable `ConfigurationID` across two `DbContext` instances of the same model — **red on master** (`15` vs `17`), **green with the fix**.
- EF SQLite suite (EF10, `--provider SQLite.MS`): 178 tests, 0 failed, 161 passed (rest skipped).


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed linq2db's query cache missing on every `DbContext` when EF Core doesn't share its internal service provider across contexts — pooled contexts, `EnableServiceProviderCaching(false)`, or several providers in one app. Each context produced a fresh mapping-schema identity, so every query was recompiled per context instead of being reused. The schema is now shared across contexts of the same model, while different models — including custom `IModelCacheKeyFactory` and multi-tenant setups — still get separate schemas.


Fixed the linq2db query cache missing on every `DbContext` with EF Core pooled contexts or disabled service-provider caching.


---
<sub>Generated from commit `9690105`.</sub>

PR #5696, opened 2026-07-10
tests: SqlCe Flush Interval=1000 to speed up the suite

MaceWindu:
Splits the SqlCe test-suite speedup out of the parallel-test-execution work (#5614) into a focused, independently-mergeable PR.

SQL CE flushes committed transactions to disk every 10s by default. Setting the max `Flush Interval` (1000s) batches those flushes to ~once per run, cutting the full SqlCe suite from ~40m to ~34m (~15%) with no failures. Durability is traded for speed, which is fine for tests (the DB is rebuilt each run).

`Autoshrink Threshold` and `Max Buffer Size` were measured too and had no effect (the bottleneck is the write/flush path, not read cache or shrink).

Only `DataProviders.json` (the committed CI test config) changes; dev `UserDataProviders.json` is unaffected.

## Follow-up commit

Explored a built-in-Windows RAM disk for the SqlCe DB (iSCSI Target `ramdisk:` provider + inbox initiator, no third-party kernel driver; actions/runner-images#7193). It works -- SqlCe reads/writes on the volume -- but yields no measured suite speedup, and the ~4 min `FS-iSCSITarget-Server` install outweighs the I/O gain.

Kept as dormant, opt-in functionality rather than dropped: `Build/Azure/scripts/ramdisk.ps1` mounts a RAM disk and emits the drive letter; the CI step is gated behind the `use_ramdisk` pipeline variable (off by default). Non-obvious bits that made the local iSCSI session connect: `AllowLoopBack=1` + WinTarget restart, and connecting over the agent's private 10.x IP (the target refuses `127.0.0.1`). Set `use_ramdisk=true` to enable for a future I/O-bound use case.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `3f63956`.</sub>

PR #5697, opened 2026-07-10
CI/tests: SQL Server test DBs use RECOVERY SIMPLE + DELAYED_DURABILITY

MaceWindu:
Splits the SQL Server test-DB log-flush tuning out of the parallel-test-execution work (#5614) into a focused, independently-mergeable PR — so master's SQL Server CI benefits now.

The baseline test DBs run FULL recovery + synchronous commit, so the write-heavy suite (CreateData seeding + BulkCopy + Merge) pays a per-commit transaction-log-flush cost. Two settings cut that:

- **`RECOVERY SIMPLE`** — auto-truncates the log; applied on **all** versions (2005+).
- **`DELAYED_DURABILITY = FORCED`** — async commit; applied on **2014+** only (the feature doesn't exist on 2005/2008/2012).

Measured on SQL Server 2022 (bounded `CreateData|BulkCopy|Merge`, 695 tests, net10): **~40% faster** (152.5s → ~92s), no failures. Both are safe for tests — the DBs are ephemeral (rebuilt each run), so relaxed durability / no point-in-time recovery is fine.

Applied to `TestData`/`TestDataMS` and the SA/Contained extras DBs at setup, in the shared `sqlserver.*` container scripts (local + Azure CI). Read-only Northwind is left untouched.

Measured but **not** adopted: `MEMORY_OPTIMIZED TEMPDB_METADATA` (no benefit) and `OPTIMIZE_FOR_AD_HOC_WORKLOADS` (no perf benefit; memory-only).

### DIFF
--- a/Build/Azure/scripts/sqlserver.2005.cmd
+++ b/Build/Azure/scripts/sqlserver.2005.cmd
@@ -16,6 +16,9 @@ docker exec mssql sqlcmd -S localhost -U sa -P Password12! -Q "SELECT @@Version"
 
 docker exec mssql sqlcmd -S localhost -U sa -P Password12! -Q "CREATE DATABASE TestData;"
 docker exec mssql sqlcmd -S localhost -U sa -P Password12! -Q "CREATE DATABASE TestDataMS;"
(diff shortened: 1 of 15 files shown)


MaceWindu:
Not worth it now - 0 improvement. Will be shipped with parallel tests.

PR #5698, opened 2026-07-10
CI/tests: run SQLite + DuckDB tests against in-memory databases

MaceWindu:
Splits the SQLite + DuckDB in-memory test-database work out of the parallel-test-execution PR (#5614) into a focused, independently-mergeable PR.

SQLite and DuckDB on-disk test runs are dominated by per-commit filesystem sync. Pointing them at **shared-cache in-memory** databases removes that: a shared-cache in-memory DB is destroyed once its last connection closes (and linq2db opens/closes a connection per query), so one keep-alive connection per DB is held open for the run (`TestInMemoryDatabases`). ~32% faster on the write-heavy SQLite path; full local in-memory suites pass (SQLite.Classic 10620, DuckDB 9648, 0 in-memory failures).

- **`DataProviders.json`** defaults SQLite + DuckDB to in-memory (dev + CI). The `.Data` on-disk companions and T4/CLI (which build file-based connection strings independently) are unaffected.
- **`Build/Azure/*/{sqlite,duckdb}.json`** point the CI jobs at in-memory (a `Connections` block per config; master's provider lists unchanged).
- The SQLite keep-alive **preloads** each in-memory DB from its committed on-disk file via the SQLite online-backup API, so a filtered run that skips `a_CreateData` still has tables/data; Northwind (no SQL seed script) is only ever loaded this way. DuckDB can't preload (`COPY FROM DATABASE` fails on FK ordering), so raw filtered DuckDB runs rely on the `CreateDatabase` seeding (the `/test` skill injects it).
- **DuckDB identity-reset** helpers (`Person`/`AllTypes`/`SequenceTest3`): DuckDB can't drop a depended-on sequence and lacks `ALTER SEQUENCE RESTART`, so detach the column default, drop+recreate the sequence, then reattach.

## Follow-up commit

Review-driven refinement of the in-memory setup:

- **DuckDB in-memory is now CI-only.** DuckDB can't preload an in-memory DB
  from its committed on-disk file (`COPY FROM DATABASE` fails on FK ordering),
  so a filtered or local run against an in-memory default would hit an empty
  DB. The tracked `DataProviders.json` default stays file-based; in-memory is
  applied only in the CI `duckdb.json` overrides, where the full suite seeds
  it via `a_CreateData`.
- **SQLite in-memory config deduplicated.** SQLite preloads via the backup
  API, so its in-memory default is safe for filtered runs — the in-memory
  strings now live only in `CommonConnectionStrings` (single source), and the
  redundant `Connections` blocks in the CI `sqlite.json` files plus the
  duplicate `LocalConnectionStrings` DuckDB entry were dropped.
- **Tightened SQLite in-memory detection** from `"memory"` to `"mode=memory"`
  to avoid a spurious keep-alive/preload on a stray `memory` in a file path.


MaceWindu:
HUge DuckDB boost, minor sqlite boost

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `4e8764b`.</sub>

PR #5701, opened 2026-07-10
Translate F# captured-lambda join predicates (#1813)

### LINKED ISSUE #1813: LEFT JOIN in F#
I'm trying to form a left join in F# but I'm getting an Exception

```
Exception message: 
variable '_arg1' of type 'Program+Names' referenced from scope '', but it is not defined
Stack trace:
Eccezione non gestita di tipo 'System.InvalidOperationException' in System.Linq.Expressions.dll
```

### Steps to reproduce

The code is quite trivial

```f#
let db = new DataConnection("SqlServer",@"Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=MyDB;Integrated Security=True;Pooling=False")

let query = query {
    for n in db.GetTable<Names>() do
    for a in db.GetTable<Addresses>().Where(fun a1 -> n.Id = a1.Id).DefaultIfEmpty() do
    select (n.Id, n.Name, a)
    }

let names = query |> Seq.toArray
```

I've also tried

```f#
let query = query {
    for n in db.GetTable<Names>() do
    leftOuterJoin a in db.GetTable<Addresses>() 
        on (n.Id = a.Id) into g_a
    select (n.Id, n.Name, g_a)
    }
let names = query |> Seq.toArray
```

but it fails with

```
Exception message: 
Multiple queries are not allowed. Set the 'LinqToDB.Common.Configuration.Linq.AllowMultipleQuery' flag to 'true' to allow multiple queries.
```

and 

```f#
let query = query {
    for n in db.GetTable<Names>() do
    leftOuterJoin a in db.GetTable<Addresses>() 
        on (n.Id = a.Id) into g_a
    for a in g_a do
    select (n.Id, n.Name, a)
    }
```
that gives 

```
Exception message: 
Eccezione non gestita di tipo 'System.NotImplementedException' in linq2db.dll
The method or operation is not implemented.
```

### Environment details
linq2db version: latest
Database Server: Sql Server

MaceWindu:
Fixes #1813.

**Captured-lambda join predicates.** F# emits captured-variable lambdas passed to `.Where` /
`.LeftJoin` as `LeafExpressionConverter.QuotationToLambdaExpression(SubstHelper(...))`, which the
core translator's `UnwrapLambda` cannot cast to a `LambdaExpression` (`InvalidCastException`).
`FSharpRewriteVisitor` now reduces those nodes to the plain `Quote(lambda)` a C# lambda emits —
substituting each free var with a `FreeVarMarker.Get<'T>(i)` call, re-running F#'s own
`QuotationToLambdaExpression` via reflection (version- and TFM-portable), then replacing the marker
calls with the captured outer value-expressions. A marker *call* rather than a placeholder *instance*
keeps the reduction independent of the free var's type - no `string` / interface / abstract /
`Nullable<'T>` free var needs a constructible placeholder. Fixes `Issue1813Test1`, `Test6`, `Test7`.

**Chained group/left joins.** `query{}` chained `groupJoin ... into g; for x in g.DefaultIfEmpty()`
nests each subsequent join inside the previous `SelectMany`'s collection selector (keyed off the
outer range var, as `Enumerable.*` over the group), which linq2db renders as `INNER JOIN LATERAL`
(dropping unmatched rows) or cannot translate without LATERAL (e.g. SQLite). `FSharpRewriteVisitor`
now flattens `X.SelectMany(a => a.g.DefaultIfEmpty().tail, (a,r)=>r)` to the C#-equivalent
`X.SelectMany(a => a.g.DefaultIfEmpty(), (a,x)=>AnonymousObject(a,x)).tail`, re-expressing the
hoisted `Enumerable.*` tail as `Queryable.*`, so it becomes plain LEFT JOINs. Fixes `Issue1813Test4`.

**Test corrections.** `Issue1813Test5/6/7` had never-verified expectations (asserting `Length==4`
and `result[0]` four times); linq2db's output was already correct. Fixed to the actual 6/6/7-row
results, compared order-tolerantly.

All `Issue1813` tests are now un-gated and pass (verified on SQLite and PostgreSQL; full `FSharpTests`
green with no regression; Release build clean across net462 / netstandard2.0 / net8-10).

## Follow-up commit

`b7f41b8e2` addresses the review finding on the free-var substitution.

The original approach built a placeholder *instance* of each `SubstHelper` free variable's type via
`GetUninitializedObject`. That rested on the assumption that those free vars - the outer query range
variables - are always entity reference types. They are not: logging every type passed to
`CreateSentinel` across the `Issue1813` set showed `Nullable<int>` and `String` alongside the entity
and tuple ones.

`Nullable<int>` happens to work - `GetUninitializedObject` returns the boxed *underlying* default on
.NET 10 and on .NET Framework alike. `string` does not: it throws
`ArgumentException: Uninitialized Strings cannot be created`, the reduction's catch-all swallows it,
and the original `UnwrapLambda` `InvalidCastException` comes back.

Free vars are now substituted with `FreeVarMarker.Get<'T>(i)` calls, which survive the
quotation-to-expression conversion as ordinary `MethodCallExpression`s and are replaced with the
captured value-expressions afterwards. Nothing needs an instance of the free variable's type, so the
`GetUninitializedObject` reflection and its `RuntimeHelpers`-vs-`FormatterServices` fork are gone.

New regression tests: `Issue1813Test9` (outer range variable is a `Nullable<int>` projected from an
`IQueryable`) and `Issue1813Test10` (a `string`). `Test10` is red without the fix on net10.0 and
net462, direct and remote. `FSharpTests` green on SQLite.Classic + LinqService: 97/97 net10.0,
93/93 net462; Release build clean on all five TFMs.

## Follow-up commits - review pass

`46d1e9911` drops four step-label comments inside `TryFlattenNestedGroupJoin` that restated the binding they
preceded. The rationale comments in the same function stay.

`7a1798a8b` lifts the encode-and-sort block that `Issue1813Test4/5/6/7` each carried verbatim into a single
`encodeJoins` helper.

`fde80ebd3` fixes the flatten for chains longer than two joins. `TryFlattenNestedGroupJoin` built the
`(outer, x)` pair by re-closing the outer element's own generic definition over two arguments, which only
works while that element is itself a two-argument `AnonymousObject`. F# widens it to the four-argument form
once a third range variable is in scope, so from the third chained `groupJoin` onwards `MakeGenericType`
threw, the catch-all swallowed it, and the un-flattened shape survived - the #1813 defect. A three-way chain
returned 3 of 6 rows on PostgreSQL, silently, and did not translate at all on SQLite; master returns 1 of 6.
The two-argument definition is now resolved from the same assembly regardless of the outer element's arity,
and `Issue1813Test11` pins a three-way chain whose third join deliberately does not match every row.

Green on PostgreSQL.16 and SQLite.Classic, direct and remote (`Issue1813` 22/22 on both); Release build clean
on all five TFMs.

Not fixed here: a chained `groupJoin` whose *inner sequence* is correlated to the outer range variable
(`groupJoin n in table.Where(fun z -> z.Id >= tr.Id) ...`) still fails, since a correlated inner sequence
cannot be expressed as a flat `GroupJoin` at all. Filed as #5790.

## Follow-up commits - review pass 2

`9bd47ed55` routes the reduction's reflection through linq2db's own wrappers. `Eval` used
`LambdaExpression.Compile()` and `Delegate.DynamicInvoke()`, and the reduction used `MethodBase.Invoke`
at three sites - all banned in `Source/BannedSymbols.txt`, and BannedApiAnalyzers does not analyse F#, so
the Release build never flagged them. The compile is the one that matters:
`Compilation.CompileExpression` honours the compiler a consumer installs through the public
`Compilation.SetExpressionCompiler` hook, and a raw `Compile()` bypasses it - which the reduction's
catch-all would turn into a silent no-op rather than a diagnostic. The same commit records why the
query-expression interceptor runs ahead of the query cache.

`d63a8f845` adds `Issue5794Test`, gated on #5794, plus two tidy-ups: `key`/`keyT` move to module level
now that three tests need them, and `Issue1813Test10` moves above `Issue1813Test11` so the F# file and
the C# wrappers read in the same order.

Release build clean on all five TFMs; `FSharpTests` 99/99 on SQLite.Classic.

Also not fixed here: a chained `groupJoin` followed by a plain `join` silently drops rows. The trailing
join puts a further range variable in scope, so F# re-projects the accumulated element field-by-field
instead of passing it through, `TryFlattenNestedGroupJoin` reads that as a non-identity result selector
and declines the shape, and the second `DefaultIfEmpty` is emitted as an INNER JOIN. The repro returns 2
of the 4 rows it should on SQLite with no error, where master raises `no such column` for the same query.
Filed as #5794 with the verified cause.

## Follow-up commits - review pass 3

`d751f5cb7` pins the correlated inner-sequence shape as `Issue5790Test`, gated on #5790 alongside the
existing `Issue5794Test`. The inner sequence is filtered by `n_rec.Id > tr.Id`, which drops trade 2's
N2 match and trade 4's N4 match, so the correlation has to reach the SQL for the expected row set to
hold - a `>=` variant would filter nothing and pass even with the predicate dropped.

`394a550a4` makes `TryFlattenNestedGroupJoin` fail with a named error rather than a cryptic one when it
produces a tree that still references the outer element. Only `ko`/`kr` are remapped off `a`; the group
join's inner sequence and every hoisted tail lambda are re-quoted verbatim, so a correlated one carries
`a` above the lambda that bound it and reaches the translator as `The LINQ expression '_arg1.Item1.Id'
could not be converted to SQL`, naming neither the cause nor this rewrite. The check compares the
flattened tree's occurrence count of `a` against `flatSM`'s rather than against zero, since `flatSM`'s
own collection and result lambdas still bind it - comparing against zero reddens `Issue1813Test4` and
`Issue1813Test11`. Diagnostics only: #5790's shape still does not translate.

`34d00b8a1` moves `toQueryable`'s unsupported operator and `requery`'s unrecognised tail off the
exception path onto `Expression option`. Both used `failwith`, landing in the same handler as every
reflection failure in the surrounding construction - the one that swallowed a `MakeGenericType` throw
in `fde80ebd3` and let PostgreSQL return 3 of 6 rows. The handler stays broad rather than narrowed:
`findGj`'s unguarded `Arguments.[0]` read and the pair-type construction still signal an unsupported
shape by throwing, so tightening it would turn legitimate declines into hard query failures. Also
addresses both suppressed review comments on `toQueryable` / `requery`.

`FSharpTests` 99/99 on SQLite.Classic, direct and remote; Release build clean on all five TFMs.

## Follow-up commits - review pass 4

`441ee5e3d` gives `TryReduceFSharpQuotation` the `FlattenInvariantException` pass-through arm that
`TryFlattenNestedGroupJoin` already carried, so an invariant violation raised by the nested `Visit` stays
loud instead of being swallowed by the blanket decline. The consequence is narrower than it looks: a
decline there leaves the node un-*reduced*, so `UnwrapLambda` throws `InvalidCastException` - the #1813
symptom - rather than producing the un-flattened `INNER JOIN LATERAL` shape, so the fix buys the message
naming #5790, not row-loss prevention. The same commit moves `findGj`'s `Arguments.Count >= 1` test onto
its match arm, so a zero-argument call in the chain declines as an ordinary `None` instead of throwing.
This supersedes the note above: the throw paths the broad handler still covers are `AsLambda`'s cast and
the pair-type construction.

`41ef11b36` corrects the `IsComplexJoinConditionSupported` documentation, which claimed the flag only
moves INNER JOIN conditions to WHERE. `SelectQueryOptimizerVisitor` gates on `JoinType.Inner or
JoinType.Left`, and the left-join branch pushes a subtree-local predicate into a wrapped derived table.
A predicate referencing both join inputs stays in ON regardless of join type.

`f6d1aef3b` replaces the YDB exclusions on `Issue1813Test8` / `Issue1813Test10` with declared
`ThrowsForProvider` refusals, so the limitation is pinned rather than recorded only in a comment.
Verified against a live YDB container on both legs: the direct one raises `Ydb.Sdk.Ado.YdbException`,
the `LinqService` one a `Grpc.Core.RpcException` whose detail carries it, matched by the `Contains`
relaxation `ExpectsFirst` applies to a remote context. `Issue1813Test10`'s stated reason is corrected
alongside the doc above - the null-aware equality cannot be relocated because it references both join
inputs. Expect two added baselines, since the SQL is generated before YDB rejects it.

`Issue1813` 22/22 on SQLite.MS and 8/8 on YDB, direct and remote; Release build clean on all five TFMs.


igor-tkachev (dismissed):
The current sentinel-based substitution does not support all captured-variable types and can silently fall back to the original failing expression tree.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
```diff
@@ -128,15 +142,219 @@ type private FSharpRewriteVisitor(mappingSchema: MappingSchema) =
...
+            match Type.GetType typeName with
+            | null -> null
+            | ty   -> ty.GetMethod("GetUninitializedObject", [| typeof<Type> |])
+        let m =
+            match getMethod "System.Runtime.CompilerServices.RuntimeHelpers" with
+            | null -> getMethod "System.Runtime.Serialization.FormatterServices"
+            | m    -> m
+        (nonNull m).Invoke(null, [| box t |]) |> nonNull
+
+    // Strip the Convert(_, obj) box F# wraps captured values in, then re-Convert to the reduced parameter's
```
igor-tkachev, line 177: This sentinel mechanism doesn't work for all possible free-variable types. In particular, boxing `Nullable<T>` with its default value produces `null`, so `CreateSentinel` fails at `nonNull`; `GetUninitializedObject` also rejects interface and some abstract/special runtime types. The exception is then swallowed and the original `QuotationToLambdaExpression(SubstHelper(...))` node is left unchanged, which means the query still fails with the original translation error.

Could we avoid using an instance of the captured variable's actual type as the substitution marker, or otherwise handle nullable/interface/abstract types explicitly? Please also add tests with a captured `Nullable<T>` and an interface-typed value.

igor-tkachev (commented):
The current sentinel-based substitution does not support all captured-variable types and can silently fall back to the original failing expression tree.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
MaceWindu: Could not reproduce. The sentinel is only ever created for `SubstHelper` free variables, which are the outer query **range variables** — always entity reference types (`GetTable<T>()` elements / tuple deconstructions), which `GetUninitializedObject` handles fine. The `Nullable<T>` / interface / abstract types flagged here can't reach `CreateSentinel`:

- A captured `Nullable<T>` accessed via a member (`filterId.Value`) doesn't compile inside an F# quotation — `FS3155: A quotation may not involve … taking the address of a captured local variable`.
- A captured interface-typed local (e.g. `IList<int>`) is an ordinary closure capture, not a `SubstHelper` free var, and translates correctly — added as `Issue1813Test8` (green on direct + remote).

So this is a robustness gap in `CreateSentinel` for inputs it can never receive, not a live defect; leaving it as-is rather than adding untested branches for unreachable types.

igor-tkachev (commented):
The previous concern is still not addressed by the added test.

The reply assumes that `SubstHelper` free variables are always entity reference types because they represent outer query range variables. That assumption is too strong: an outer range variable can come from any `IQueryable<T>`, including a scalar/value-type projection.

For example, this is a plausible F# query shape:

```fsharp
let ids =
    query {
        for n in db.GetTable<Names>() do
        select (Nullable n.Id)
    }

let query =
    query {
        for id in ids do
        for a in db.GetTable<Addresses>()
            .Where(fun a -> Nullable a.Id = id)
            .DefaultIfEmpty() do
        select (id, a)
    }
```

Here `id` is not a closure local. It is an outer query range variable, so it is exactly the kind of free variable that can be passed through `SubstHelper`; however, its type is `Nullable<int>`, not an entity reference type.

The current implementation then tries to create the sentinel with `GetUninitializedObject(typeof<Nullable<int>>)`. Boxing the default `Nullable<int>` produces `null`, so the following `nonNull` fails. The exception is swallowed by `with _ -> None`, the rewrite is abandoned, and the original `QuotationToLambdaExpression(SubstHelper(...))` node remains in the tree, bringing back the original translation failure.

The newly added `Issue1813Test8` does not exercise this path. It uses an interface-typed closure local:

```fsharp
let ids : IList<int> = List<int>([1])
```

As the test comment itself explains, that value is an ordinary closure capture and never reaches `CreateSentinel`. Therefore the test only proves that a different code path works; it does not prove that interface/abstract/nullable types are unreachable as `SubstHelper` free variables.

Please add a regression test where the problematic type is the outer query range variable itself, preferably `Nullable<T>` from a projected `IQueryable<Nullable<T>>`. If that shape does not produce `SubstHelper`, the generated expression tree should be inspected and the claimed invariant documented based on the actual tree shape rather than inferred from the current entity-based tests. Otherwise `CreateSentinel` needs a strategy that does not depend on constructing a non-null instance of the free variable's runtime type.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
MaceWindu: Correction: my earlier reply on this thread was wrong, and I'm retracting it. `SubstHelper` free vars are *not* always entity reference types, so "the flagged types can't reach `CreateSentinel`" does not hold.

Verified by logging every type passed to `CreateSentinel` while running the whole `Issue1813` set:

| Type reaching `CreateSentinel` | Origin |
|---|---|
| `Issue1813+Names`, `Issue1813+TradeValid` | entity range variables |
| `Tuple<TradeValid,NominationValid>` | tuple range variable |
| `System.Nullable<int>` | the shape you posted - projected `IQueryable<Nullable<int>>` |
| `System.String` | projected `IQueryable<string>` |

Two corrections to the mechanism:

- `Nullable<int>` does reach `CreateSentinel`, but it does **not** fail there. `GetUninitializedObject(typeof<Nullable<int>>)` returns the boxed *underlying* default (a `System.Int32`), not `null` - on `net10.0` and on `net462`'s `FormatterServices` path alike. Your query, added verbatim as `Issue1813Test9`, was already green on both TFMs before any fix.
- The type that actually fails is `string`: `GetUninitializedObject` throws `ArgumentException: Uninitialized Strings cannot be created`, the catch-all swallows it, and the original `UnwrapLambda` `InvalidCastException` comes back. `Issue1813Test10` (outer range variable projected from `IQueryable<string>`) is red pre-fix on net10.0 and net462, direct and remote.

So the defect is real, one type over from the one named.

Fixed in b7f41b8e2 along the lines you asked for - no instance of the free variable's type is constructed at all. Each free var is substituted with a `FreeVarMarker.Get<'T>(i)` call, which survives the quotation-to-expression conversion as an ordinary `MethodCallExpression` and is replaced with the captured value-expression afterwards. `GetUninitializedObject` and its `RuntimeHelpers`-vs-`FormatterServices` fork are gone, so `string` / interface / abstract / `Nullable<'T>` are all handled by construction rather than by luck.

`FSharpTests` on SQLite.Classic + LinqService: 97/97 net10.0, 93/93 net462. Release build clean on all five TFMs. Full CI running.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
```diff
@@ -128,15 +166,237 @@ type private FSharpRewriteVisitor(mappingSchema: MappingSchema) =
...
+                        // expects (it peels Quote/Convert, then casts to LambdaExpression).
+                        Some (Expression.Quote(this.Visit (replaced :?> LambdaExpression) |> nonNull))
+                | bare ->
+                    // No captured variables: convert the quotation directly with F#'s own method.
+                    let q   = this.Eval bare
+                    let lam = (node.Method.InvokeExt(null, [| q |]) |> nonNull) :?> LambdaExpression
+                    Some (Expression.Quote(this.Visit lam |> nonNull))
+            // Fail safe: any unexpected shape/evaluation surprise leaves the node untouched (no worse than
+            // before this rewrite existed), so only genuinely-matching F# quotation predicates are transformed.
+            with _ -> None
```
sdanyliv: **Minor · MIN001** — `FlattenInvariantException` exists so a wrong flatten fails loudly instead of degrading to the un-flattened shape that renders as `INNER JOIN LATERAL` and drops rows — its own doc comment says exactly that, and `TryFlattenNestedGroupJoin` backs it up with an explicit `| :? FlattenInvariantException -> reraise()` arm at `:382`. `TryReduceFSharpQuotation` does not: it calls `this.Visit` on the reduced lambda from inside its own `try` (`:247` for the captured-variable branch, `:252` for the bare one) and catches with a bare `with _ -> None`. An invariant violation raised during that nested visit is therefore swallowed, the `QuotationToLambdaExpression` node is left untouched, and execution continues with exactly the shape the invariant is there to reject.

Reachability is unprobed — it needs a chained-groupJoin `SelectMany` with a correlated inner sequence nested inside a captured-variable predicate lambda, and I could not construct one — so this is a robustness gap rather than a demonstrated bug. The inconsistency itself is introduced here, and it is the one place in the file where the stated fail-loud guarantee does not hold.

Fix: give `TryReduceFSharpQuotation` the same pass-through arm the flatten handler already carries, so the guarantee holds regardless of which rewrite is on the stack when the invariant fires. Everything else still declines exactly as before.

```suggestion
            // Fail safe: any unexpected shape/evaluation surprise leaves the node untouched (no worse than
            // before this rewrite existed), so only genuinely-matching F# quotation predicates are transformed.
            // A flatten-invariant violation from the nested Visit is not such a surprise - it must stay loud.
            with
            | :? FlattenInvariantException -> reraise()
            | _                            -> None
```

— ✓ Fixed in 441ee5e3d

Tests/Linq/Linq/FSharpTests.cs
```diff
@@ -385,38 +384,83 @@ public void Issue1813Test3([DataSources] string context)
...
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/1813")]
 		public void Issue1813Test7([DataSources] string context)
 		{
 			using var db = GetDataContext(context);
 			FSharp.Issue1813.Issue1813Test7(db);
 		}
 
+		// YDB rejects the constant carried into the join ON clause ("each equality predicate argument must depend on exactly one JOIN input").
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/1813")]
+		public void Issue1813Test8([DataSources(TestProvName.AllYdb)] string context)
```
sdanyliv: **Minor · MIN002** — the comment's claim checks out: `Source/LinqToDB/Internal/DataProvider/Ydb/YdbDataProvider.cs:57-59` already records the same YQL error string as the reason for `IsSupportsJoinWithoutCondition = false`, and `a1.Id = ids.[0]` does put an input-independent equality into the ON clause. The concern is the mechanism. `[DataSources(TestProvName.AllYdb)]` removes YDB from the run entirely, so the only record that YDB is expected to fail is a comment — and comments don't run. Nothing detects the day YDB starts accepting the predicate, which is the event that would prompt removing the gate.

The repo declares this class of refusal with `ThrowsForProvider`, including for exactly this kind of raw YDB rejection: `Tests/Linq/Linq/PredicateTests.cs:146` is `[ThrowsForProvider("Ydb.Sdk.Ado.YdbException", TestProvName.AllYdb)]`. That keeps YDB in the matrix and pins the limitation instead of hiding it. It also handles the remote contexts this test runs under — the constructor expands each provider to both the direct and the `LinqService` name, and `ExpectsFirst` relaxes the match from `StartsWith` to `Contains` for the remote one — so `[DataSources]` without the `false` stays correct.

Fix: swap the silent exclusion for a declared refusal. Confirm the exception type on the YDB leg before relying on it — `PredicateTests` uses `Ydb.Sdk.Ado.YdbException` for the same server-side-rejection class, but if linq2db wraps it the type string needs adjusting. If no stable type can be pinned, `[ActiveIssue("<reason>", Configuration = TestProvName.AllYdb)]` — the form `Tests/Linq/Linq/JoinTests.cs:3358` uses for a YDB limitation — is the next-best declared alternative.

```suggestion
		// YDB rejects the constant carried into the join ON clause ("each equality predicate argument must depend on exactly one JOIN input").
		[ThrowsForProvider("Ydb.Sdk.Ado.YdbException", TestProvName.AllYdb)]
		[Test(Description = "https://github.com/linq2db/linq2db/issues/1813")]
		public void Issue1813Test8([DataSources] string context)
```

— ✓ Fixed in f6d1aef3b

Tests/Linq/Linq/FSharpTests.cs
```diff
@@ -385,38 +384,83 @@ public void Issue1813Test3([DataSources] string context)
...
+			using var db = GetDataContext(context);
+			FSharp.Issue1813.Issue1813Test9(db);
+		}
+
+		// The nullable-string join key makes linq2db emit the null-aware equality
+		// (ON a.Text = b.Name OR a.Text IS NULL AND b.Name IS NULL), which YDB rejects ("JOIN ON expression must be
+		// a conjunction of equality predicates"). IsComplexJoinConditionSupported=false only relocates INNER JOIN
+		// conditions to WHERE; an outer join cannot.
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/1813")]
+		public void Issue1813Test10([DataSources(TestProvName.AllYdb)] string context)
```
sdanyliv: **Minor · MIN003** — same mechanism concern as MIN002: `[DataSources(TestProvName.AllYdb)]` drops YDB from the run, so the expectation lives only in prose and nothing fires if YDB later accepts the predicate. The repo declares this class of refusal with `ThrowsForProvider` (`Tests/Linq/Linq/PredicateTests.cs:146` for the same YDB case), which keeps the provider in the matrix. This is genuinely a server-side rejection rather than something a provider flag could carry — the comment establishes that correctly — so declaring it, not adding a flag, is the fix.

Separately, the stated reason overstates the flag's limitation. `IsComplexJoinConditionSupported = false` is not INNER-only: `Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs:2733` gates on `JoinType.Inner or JoinType.Left`, and `:2758-2820` is a dedicated left-join branch. What actually blocks relocation here is that the null-aware equality references *both* join inputs — the code's own comment at `:2764` says a cross-input predicate "must stay in ON", regardless of join type. The flag's XML doc at `SqlProviderFlags.cs:554` does say "INNER JOIN", which is presumably where the reading came from; that doc is stale, noted separately as out of scope.

Fix: declare the refusal instead of excluding the provider, and tighten the reason to the cross-input point. Same caveat as MIN002 about confirming the exception type on the YDB leg.

```suggestion
		// The nullable-string join key makes linq2db emit the null-aware equality
		// (ON a.Text = b.Name OR a.Text IS NULL AND b.Name IS NULL), which YDB rejects ("JOIN ON expression must be
		// a conjunction of equality predicates"). IsComplexJoinConditionSupported=false cannot relocate it
		// either: the predicate references both join inputs, so it has to stay in ON.
		[ThrowsForProvider("Ydb.Sdk.Ado.YdbException", TestProvName.AllYdb)]
		[Test(Description = "https://github.com/linq2db/linq2db/issues/1813")]
		public void Issue1813Test10([DataSources] string context)
```

— ✓ Fixed in f6d1aef3b

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
```diff
@@ -128,15 +166,237 @@ type private FSharpRewriteVisitor(mappingSchema: MappingSchema) =
...
+                            match mc.Arguments.[0] with
+                            | :? MethodCallExpression as dfe when dfe.Method.Name = "DefaultIfEmpty" && dfe.Arguments.Count >= 1 ->
+                                match dfe.Arguments.[0] with
+                                | :? MemberExpression as me -> obj.ReferenceEquals(me.Expression, a)
+                                | _                         -> false
+                            | _ -> false
+                        if   mc.Method.Name = "GroupJoin" && mc.Arguments.Count = 5 && sourcedFromA then Some mc
+                        elif mc.Arguments.Count >= 1                                                then findGj mc.Arguments.[0]
+                        else None
+                    | _ -> None
```
sdanyliv, line 301: **Suggestion · SUG001** — `sourcedFromA` reads `mc.Arguments.[0]` at `:289`, before the `mc.Arguments.Count >= 1` test at `:296`, so a zero-argument call anywhere in the chain throws `ArgumentOutOfRangeException` and is turned into a decline by the broad handler 90 lines below. That is deliberate and documented at `:378-380` — but the same comment notes the handler "also swallows a genuine rewrite bug, whose fallback here is the un-flattened INNER JOIN LATERAL shape - silent row loss", so every throw path that can be expressed as an ordinary `None` narrows the window in which a real bug hides behind a documented one. This continues the direction `34d00b8a1` already took for `toQueryable` and `requery`.

Moving the count test onto the match arm's `when` guard is behaviourally identical for every input — a zero-argument call already ends up declining — and removes the exception round-trip. The handler stays broad, since the pair-type construction still signals an unsupported shape by throwing.

Fix: guard the match arm with `when mc.Arguments.Count >= 1` and drop the now-redundant `elif`; a zero-argument call then falls to the existing `| _ -> None`.

```suggestion
                let rec findGj (e: Expression) : MethodCallExpression option =
                    match e with
                    | :? MethodCallExpression as mc when mc.Arguments.Count >= 1 ->
                        let sourcedFromA =
                            match mc.Arguments.[0] with
                            | :? MethodCallExpression as dfe when dfe.Method.Name = "DefaultIfEmpty" && dfe.Arguments.Count >= 1 ->
                                match dfe.Arguments.[0] with
                                | :? MemberExpression as me -> obj.ReferenceEquals(me.Expression, a)
                                | _                         -> false
                            | _ -> false
                        if mc.Method.Name = "GroupJoin" && mc.Arguments.Count = 5 && sourcedFromA then Some mc
                        else findGj mc.Arguments.[0]
                    | _ -> None
```

— ✓ Fixed in 441ee5e3d

sdanyliv (commented):
> [!IMPORTANT]

## Prior-review audit

- **Fixed** · igor-tkachev — the sentinel substitution cannot build a placeholder for every free-variable type, and the swallowed failure restores the original translation error ([thread](https://github.com/linq2db/linq2db/pull/5701#discussion_r3567268389)). At HEAD no instance of the free variable's type is constructed at all: free vars are substituted with `FreeVarMarker.Get<'T>(i)` calls, and `CreateSentinel` / `GetUninitializedObject` / `FormatterServices` no longer appear in the file.
- **Fixed** · copilot-pull-request-reviewer — `toQueryable` used `failwith` for unsupported tail operators and relied on the surrounding catch-all (suppressed comment on `FSharpQueryExpressionInterceptor.fs:338`). `34d00b8a1` moved it onto `Expression option`; no `failwith` remains in the file.
- **Fixed** · copilot-pull-request-reviewer — same for `requery`'s unrecognised tail shape (suppressed comment on `FSharpQueryExpressionInterceptor.fs:354`).
- Both of the most recent Copilot runs returned "unable to review — quota limit", so the current head has had no bot pass.

## Out-of-scope observations

- **The quotation reduction re-runs — and re-compiles — on every execution of an affected F# query.** `Query<T>.GetQuery` builds the query-cache key from the post-interception tree (`Source/LinqToDB/Internal/Linq/Query{T}.cs:144`, looked up at `:151`), so `FSharpQueryExpressionInterceptor.ProcessExpression` — and with it the quotation reduction — runs ahead of every cache lookup, i.e. on each execution of an F# query carrying a captured-lambda predicate. Each run pays a full `Expression.Lambda(...).CompileExpression()` in `Eval` to materialise the quotation plus its `Var[]`, then `Expr.Substitute`, a reflection call into F#'s `QuotationToLambdaExpression`, a `Transform`, and a re-`Visit`. The compile-free `LinqToDB.Internal.Expressions.ExpressionEvaluator.EvaluateExpression` would not avoid that as written — `IsSimpleEvaluatable` accepts only Constant / Default / MemberAccess / Call, and the `NewArrayInit` this code deliberately builds (to keep `Var` identity within a single execution, which is correct and necessary) falls straight back to the same compile. The lever that would help is memoising the marker-substituted lambda per call site: `q` and `vars` come from the same compiled quotation literal every time, and only the spliced captured value-expressions vary. Nothing regresses — these queries threw outright before this change, and `UseFSharp()` is opt-in — so this is a follow-up worth weighing, not a change needed here.
- **`IsComplexJoinConditionSupported`'s XML documentation is stale.** `Source/LinqToDB/Internal/SqlProvider/SqlProviderFlags.cs:554` describes the flag as moving "all conditions from INNER JOIN ON ... to WHERE", but the implementation is not INNER-only: `Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs:2733` gates on `JoinType.Inner or JoinType.Left`, and `:2758-2820` is a dedicated left-join branch that pushes subtree-local predicates into a wrapped derived table rather than into `WHERE`. Pre-existing and unrelated to this change — surfaced only because `Issue1813Test10`'s new comment reads the doc rather than the code (MIN003).

## Baselines

Delta: [linq2db.baselines#2063](https://github.com/linq2db/linq2db.baselines/pull/2063) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5701)

### New `Issue1813Test1` / `Test4`-`Test11` baselines — 610 added, 0 modified, 0 deleted

68 providers × 9 tests, with `Test8` and `Test10` at 67 each — YDB absent, matching their exclusions. Nothing on `master` moved and no existing baseline changed, so the delta is entirely this branch's own new and un-gated tests. Coverage ties out exactly: 68×7 + 67×2 = 610, with those two the only holes.

- **Chained `DefaultIfEmpty` flattened to LEFT JOIN** — `Test4`, `Test5`, `Test6`, `Test7`, `Test11`. `Test11` is the load-bearing one: it is the three-way chain whose third join deliberately does not match every row, so a degraded join would silently drop two rows. It emits exactly three `LEFT JOIN`s — no `INNER JOIN`, no `LATERAL`, no `CROSS APPLY`/`OUTER APPLY`, no correlated subquery — on every provider sampled, including SQLite, Access, Firebird, Sybase, DuckDB, ClickHouse, Informix, SAP HANA, DB2 and YDB, none of which have LATERAL or APPLY. `Test4` matches across the same set.
- **Captured-lambda predicate reaches the SQL** — `Test1`, `Test8`, `Test9`, `Test10`. The outer range variable renders as a column reference and never as a bound parameter, so the correlation is evaluated server-side. `Test8`'s genuinely captured constant does become a bound parameter, `AND`-ed alongside the correlated column comparison sitting next to it. `Test10`'s nullable-string key produces the null-aware `ON a.Text = b.Name OR a.Text IS NULL AND b.Name IS NULL` form on every provider checked.

No cross-provider anomalies: every divergence observed is routine per-provider syntax (parameter prefix, identifier quoting, paging form).

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
MaceWindu: Applied in `441ee5e3d` - the reduction handler now carries the same `| :? FlattenInvariantException -> reraise()` arm.

Correction to the rationale, though: a decline from `TryReduceFSharpQuotation` does **not** leave the un-flattened `INNER JOIN LATERAL` shape. It leaves the node un-*reduced*, so the tree still carries `QuotationToLambdaExpression(...)` and `UnwrapLambda` throws `InvalidCastException` - the #1813 symptom itself, and how `Issue1813Test10`'s pre-fix red was originally diagnosed. So the swallow costs a bare cast error in place of the message naming #5790, not silent row loss. Worth fixing either way, just not for that reason.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
MaceWindu: Applied in `441ee5e3d`. Guard moved onto the match arm, redundant `elif` dropped.

Behaviour is unchanged, as claimed: `findGj` recurses only on `Arguments.[0]`, so a `None` at any depth propagates to `findGj coll.Body` and declines exactly where the exception route landed.

Also updated the broad handler's own comment, which this change falsified - it cited the zero-argument call as one of the throw paths it still covers. `AsLambda`'s cast and the pair-type construction remain.

Tests/Linq/Linq/FSharpTests.cs
MaceWindu: Applied in `f6d1aef3b`.

Probed against a live YDB container before adopting, because the remote claim had no precedent: all seven existing `ThrowsForProvider` sites naming a raw ADO exception pair it with `[DataSources(false)]`, and every remote-inclusive site uses `typeof(LinqToDBException)`. Both legs match anyway -

- direct: `Ydb.Sdk.Ado.YdbException : Status: GenericError, Issues: [0] (8:64) Error: JOIN: each equality predicate argument must depend on exactly one JOIN input`
- remote: `Grpc.Core.RpcException : Status(StatusCode="Unknown", Detail="Ydb.Sdk.Ado.YdbException (0x80004005): ...` - matched by the `Contains` relaxation `ExpectsFirst` applies to a `.LinqService` context.

8/8 green on `YDB`, `YDB.LinqService`, `SQLite.MS`, `SQLite.MS.LinqService`. One caveat: only the gRPC transport was probed.

Side effect worth expecting in the baselines: re-including YDB adds `Issue1813Test8(YDB).sql` and `Issue1813Test10(YDB).sql`, since the SQL is generated before the server rejects it.

Tests/Linq/Linq/FSharpTests.cs
MaceWindu: Applied in `f6d1aef3b`, both parts - the comment now gives the cross-input reason.

Mechanism confirmed: `SelectQueryOptimizerVisitor.cs:2732` gates on `JoinType.Inner or JoinType.Left` and `:2757-2820` is a dedicated left-join branch, so the INNER-only reading was wrong and the code's own comment at `:2762-2764` states the cross-input rule directly.

The live error is `[0] (7:89) Error: JOIN ON expression must be a conjunction of equality predicates` - that string had no corroboration anywhere in the repo until this run.

Stale flag doc fixed separately in `41ef11b36`.

Source/LinqToDB.FSharp/FSharpQueryExpressionInterceptor.fs
sdanyliv: Correction: your rationale fix is right and my MIN001 reasoning was wrong. `UnwrapLambda()` is `(LambdaExpression?)ex.Unwrap()` (`InternalExtensions.cs:32`) and `Unwrap()` only peels Quote/Convert/ConvertChecked, so a `None` from `TryReduceFSharpQuotation` leaves the node un-reduced and the cast throws `InvalidCastException` — the #1813 symptom — not the un-flattened `INNER JOIN LATERAL` shape I described.

Confirmed empirically on a `master` worktree against PostgreSQL.16, running both shapes side by side:

- correlated (the #5790 shape): `System.InvalidCastException : Unable to cast object of type 'MethodCallExpression1' to type 'LambdaExpression'` — never reaches SQL.
- uncorrelated (`Issue1813Test4`'s shape): runs and returns `1-1-0,1-4-0,4-3-4`, i.e. 3 of the 6 correct rows.

So the silent row loss belongs to the shape this PR *fixes*, not to the one it now refuses. The `reraise()` arm is still right, just for the narrower reason: it preserves the message naming #5790 instead of degrading to a bare cast error.

sdanyliv:
Pushed two commits from a `/verify-review` walk on `f6d1aef3b`. All four prior findings (MIN001–003, SUG001) verified fixed at that head and annotated on their threads.

**`6c95532d69` — Qualify the `IsComplexJoinConditionSupported` documentation**

`41ef11b36` dropped the INNER-only error, but the replacement reads as a universal rule. Three gaps against `SelectQueryOptimizerVisitor`:

- `OptimizeJoinConditions` gates on `join.JoinType is (JoinType.Inner or JoinType.Left) && !join.Condition.IsOr && join.Condition.Predicates.Count != 0` (`:2733`), so an ON clause that is itself a disjunction is skipped whole and emitted verbatim — the case a provider author setting the flag `false` most needs to know about, since it is the one the flag will not rescue them from.
- A LEFT join has three outcomes, not two: `move = !QueryHelper.IsDependsOnSources(...)` (`:2761`) sends an outer-only predicate to WHERE, `!move && !IsDependsOnOuterSources(...)` pushes into the wrapped derived table, and otherwise it stays in ON.
- `move` also fires for a plain equality when either operand is a `SqlValue` (`:2754-2756`), which "not a conjunction of equality predicates" understates.

**`cfb4433b1b` — Pin the flatten invariant refusal with an un-gated test (#5790)**

`Issue5790Test` is `[ActiveIssue(5790)]`-gated, so it is Explicit and never runs — leaving the `FlattenInvariantException` raise at `FSharpQueryExpressionInterceptor.fs:376-378` with no executing coverage. After `441ee5e3d` that raise is the one exception both broad handlers deliberately let through, and every green flatten test (`Issue1813Test4`–`Test7`, `Test11`) exercises only the `countA flattened = countA flatSM` branch. A later edit that turned the invariant back into a silent decline would leave the suite green.

`Issue5790RefusalTest` asserts the refusal itself. Delete it together with the gate when #5790 lands.

Release build of `Source/LinqToDB` clean (0 warnings, 0 errors). `Issue5790RefusalTest` green on `SQLite.MS` / net10.0.

**Baselines caveat.** `baselines/pr_5701` tips at `517545dfa` (2026-08-26T01:10Z), which predates `441ee5e3d` / `41ef11b36` / `f6d1aef3b` (all 18:22Z), and no `test-all` has run since 08-25. `Issue1813Test8` / `Test10` still show `providerCount: 67` with no YDB entries, so the predicted YDB re-inclusion (610 → 612) and `441ee5e3d`'s SQL-inertness are both unverified until a fresh run.

sdanyliv (dismissed):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Join and `Where` predicates that capture an outer range variable now translate. F# compiles such a lambda differently from C#, and LinqToDB failed on it with an `InvalidCastException`:

  ```fsharp
  for a in db.GetTable<Addresses>().Where(fun a1 -> n.Id = a1.Id).DefaultIfEmpty() do
  ...
  .LeftJoin(fun y -> y.ReceiptDealNumber = tr.DealNumber && y.ReceiptParcelID = tr.ParcelID)
  ```

  Chained `groupJoin … into g` followed by `for x in g.DefaultIfEmpty()` now emits real `LEFT JOIN`s as well — including three or more chained blocks — where it previously rendered as `INNER JOIN LATERAL` on providers supporting LATERAL, silently dropping unmatched rows, and did not translate at all elsewhere. Two shapes are still unsupported, and they behave differently: a `groupJoin` whose inner sequence is correlated with the outer row now fails with a clear message naming [#5790](https://github.com/linq2db/linq2db/issues/5790), while a chained `groupJoin` followed by a plain `join` is still emitted as an `INNER JOIN` and silently drops the rows it should keep — avoid that shape until [#5794](https://github.com/linq2db/linq2db/issues/5794) is fixed. On YDB two of the newly-translating shapes still fail, because of its existing restrictions on `JOIN … ON`.


F#: join and `Where` predicates that capture an outer range variable now translate, and chained `groupJoin` / `DefaultIfEmpty` emits real `LEFT JOIN`s.


---
<sub>Generated from commit `a8a0cbe`.</sub>

PR #5703, opened 2026-07-10
linq2db.Analyzers: new analyzer package + LINQ2DB1001 (legacy Sql.Ext window API -> Sql.Window)

MaceWindu:
## What

Adds **`linq2db.Analyzers`** — the first *user-facing* Roslyn analyzer NuGet package (the existing `Source/CodeGenerators` analyzer is internal build-only tooling). Opt-in and development-only (`developmentDependency`, ships no runtime assembly).

First rule **LINQ2DB1001** (Info): flags the legacy `Sql.Ext` analytic / window-function API (superseded by `Sql.Window`) and offers a **trivia-preserving** code fix that rewrites convertible chains to `Sql.Window`.

```csharp
// before
Sql.Ext.RowNumber().Over().PartitionBy(x.Cat).OrderBy(x.Date).ToValue()
// after (code fix)
Sql.Window.RowNumber(f => f.PartitionBy(x.Cat).OrderBy(x.Date))
```

## Package layout

Two assemblies in one package (`analyzers/dotnet/cs`):

- `Source/LinqToDB.Analyzers` — analyzer only, **no `Workspaces` reference** (RS1038; keeps keystroke-time analysis Workspaces-free).
- `Source/LinqToDB.Analyzers.CodeFixes` — code fix + the packable project; packs both DLLs, `readme.md`, `developmentDependency=true`, no `lib/`, no dependency leak.

Roslyn pinned to **4.8.0** — the Roslyn bundled with the lowest supported .NET SDK (.NET 8) — for broad IDE reach, decoupled from the repo's 5.3.0 runtime pin.

## Code-fix coverage

Converts every mechanical `.Over()` case: ranking, aggregates (modifier → `.Distinct()`), statistical, Lead/Lag, value functions (nulls → `IgnoreNulls`/`RespectNulls`, from → `FromFirst`/`FromLast`), KEEP, Median, RatioToReport, windowed PercentileCont/Disc — with PartitionBy/OrderBy/ThenBy(+Desc/+NullsPosition) and Rows/Range frames. Comments and formatting are preserved, and the result stays valid inside expression trees.

Genuinely non-convertible shapes (plain aggregate without `.Over()`, group-form percentile, `ListAgg`, old `Filter`) are reported but not auto-fixed. For the `double?`-returning families (statistical / Median / RatioToReport / percentile) a guard withholds the fix when the target slot wouldn't accept `double?`.

## Tests + CI

`Tests/Tests.Analyzers` (Roslyn testing SDK, NUnit/MTP) — 20 tests: detection (positive/negative + edge), per-family before/after, a trivia battery, and an expression-tree case. DB-free; runs on the build agent via a new `with_analyzer_tests` step (not the provider matrix).

## Follow-ups (post-merge)

- Wiki page for `LINQ2DB1001` (the descriptor's `helpLinkUri` target).
- Add `linq2db.Analyzers` to the release-publish expected-nuget verification list.

## Follow-up commits (review)

Addressed review findings on the code fix:

- **DISTINCT+KEEP bail** (`60852a2`): the KEEP branch parsed a `DISTINCT` aggregate modifier but never emitted `.Distinct()`, so a chain like `Sql.Ext.Sum(x, Sql.AggregateModifier.Distinct).KeepFirst().OrderBy(x).Over().ToValue()` was rewritten to `Sql.Window` with DISTINCT silently dropped. `Sql.Window`'s KEEP builder has no `.Distinct()` state, so the fix now bails (reported, not fixed). Regression test added.
- **MA0154 XML doc** (`db71216`): the rewriter summary wrapped the C# keyword `null` in `<c>...</c>`; switched to the `<see langword>` form. MA0154 is configured as an error and Meziantou runs in Release, so a Release build of `LinqToDB.Analyzers.CodeFixes` failed on it.
- **No-fix coverage** (`01f8d10`): added tests for a windowed `ListAgg` and a legacy `Filter` chain (both reach the unrecognized-root bail and stay diagnostic-only).

37/37 analyzer tests green; Release build of `LinqToDB.Analyzers.CodeFixes` clean.

## Follow-up commit — `L2DB1001` rename + review round 2 (`863d87d2`)

Addressed the remaining review points and renamed the diagnostic id.

- **Rule id `LINQ2DB1001` → `L2DB1001`** (per review): renamed across the descriptor, `helpLinkUri`, `AnalyzerReleases.Unshipped.md`, the readme, the test markup, and the wiki page (now live at [`L2DB1001`](https://github.com/linq2db/linq2db/wiki/L2DB1001) — this supersedes the "Wiki page" post-merge follow-up above).
- **Gate the diagnostic on `.Over()`** (review #1): the rule reported every `AnalyticFunctions` `ToValue()` chain, including plain aggregates and group-form (`WITHIN GROUP`) ordered-set aggregates that have no `Sql.Window` target. It now fires only on genuine window chains; those non-window shapes are no longer reported.
- **Resolve enum modifiers by constant value** (review #3): a `DISTINCT`/`Nulls`/`From` modifier referenced through a const alias was matched by member-name text and silently dropped (compiling, but different SQL). It is now resolved via `GetConstantValue`; a non-compile-time-constant modifier withholds the fix.
- **Line-terminate salvaged single-line comments**: a `//` comment on chain scaffolding was appended without a terminator, swallowing the statement's `;` into the comment (non-compiling). Fixed.
- Added tests for `NthValue` reordered named args and the non-constant-modifier bail; removed an unused `Shouldly` reference.

45 analyzer tests green; Release build of `LinqToDB.Analyzers.CodeFixes` clean.

## Follow-up commits: dogfooding fixes (`2ac84f4a3`, `7e27be30b`)

Dogfooded L2DB1001 over the `Tests/Linq` window-function corpus (~292 sites); two code-fix gaps surfaced, both fixed:

- **Fix-All under-application** (`2ac84f4a3`): `WellKnownFixAllProviders.BatchFixer` computes each fix against the original tree and drops physically-clustered edits (several window columns in one `select new { ... }`), so "Fix all in document" / `dotnet format` converted only a fraction of a file (6/266 in `AnalyticTests.cs`). Replaced with a custom `DocumentBasedFixAllProvider` that rewrites every flagged chain in one `ReplaceNodes` pass. Regression test `FixAllConvertsEveryClusteredChain` (>=3 adjacent occurrences, exercising the SDK Fix-All leg).
- **Opt-in over return-type divergence** (`7e27be30b`): the largest source of un-fixed sites was the return-type-fit gate (e.g. `Sql.Window`'s `double?` into a `double` slot). Added a default-off `.editorconfig` option `linq2db.L2DB1001.apply_fix_on_return_type_mismatch` that applies the migration anyway, leaving the user to reconcile the type (documented in readme + wiki). With it enabled, corpus conversion goes 168 -> 284 / 292; the residual 8 are genuinely non-convertible.

49 analyzer tests green; Release build clean.


sdanyliv (dismissed):

igor-tkachev:
I don't like the `LINQ2DB1001` diagnostic ID prefix. It is unnecessarily long. Please use `L2DB1001` instead and update all related code, tests, release metadata, documentation, and the `helpLinkUri` target accordingly.

igor-tkachev (dismissed):
I found several issues that should be addressed before this analyzer package is published.

1. **LINQ2DB1001 reports plain aggregates that have no `Sql.Window` replacement.**

The analyzer reports every `AnalyticFunctions.IReadyToFunction<T>.ToValue()` call, but it doesn't verify that the chain contains `.Over()`. This means calls such as `Sql.Ext.Count().ToValue()` or `Sql.Ext.Sum(value).ToValue()` are reported with a message telling the user to migrate to `Sql.Window`, while the code fix correctly refuses to rewrite them because there is no equivalent window-function call.

Please either restrict this diagnostic to chains that actually contain `.Over()`, or change the diagnostic semantics/message and provide an appropriate migration target for non-window aggregate calls. I think restricting this rule to genuine window-function chains is the cleaner option.

2. **Root arguments are matched to parameters by position, which is incorrect for named arguments.**

`BuildFromRoot()` uses the syntax argument index to select `rootMethod.Parameters[i]`. C# named arguments can be reordered, so a valid call such as `Lag(expr: ..., offset: ..., @default: ..., nulls: Sql.Nulls.Ignore)` can associate `offset` with the `Nulls` parameter and either suppress a valid fix or construct the wrong intermediate argument list.

Please map syntax arguments through Roslyn's semantic model, e.g. `IArgumentOperation.Parameter`, rather than relying on source order. Tests should cover reordered named arguments for at least `Lag`, `NthValue`, and an aggregate modifier overload.

3. **Enum modifiers are inferred from member names instead of constant values.**

`EnumMemberName()` accepts any member access named `Ignore`, `Respect`, `First`, `Last`, or `Distinct`. It misses valid enum constants expressed through aliases/const fields and can misinterpret properties whose names happen to match an enum member but whose runtime value is not fixed.

Please use semantic constant evaluation (`SemanticModel.GetConstantValue()` or the corresponding operation constant value) and compare the actual enum value. If the argument is not a compile-time constant, the code fix should not be offered.

These are important for the first public release because analyzer diagnostics and code-fix behavior become user-facing compatibility surface immediately.

MaceWindu:
Addressed in 863d87d2 (plus the earlier fcf752e for #2):

1. **Plain aggregates without `.Over()`.** The rule now gates on `.Over()`, so a plain aggregate such as `Sql.Ext.Count(x).ToValue()` is no longer reported. The same gate also stops reporting the ordered-set *group* form (`PercentileCont`/`PercentileDisc` with `WITHIN GROUP` but no `.Over()`), which likewise has no `Sql.Window` target. Only genuine window chains are flagged now.

2. **Root arguments matched by position.** Fixed in fcf752e: each written argument is mapped to its parameter through the semantic model (`IInvocationOperation`/`IArgumentOperation.Parameter`) and value args are re-sorted by parameter ordinal, dropping any stale name-colon. Added a code-fix test for `NthValue` with reordered named value args (its `expr`/`n` value slots interleave with the `From`/`Nulls` modifiers, so it is the most fragile branch of the re-sort).

3. **Enum modifiers inferred from member names.** `EnumMemberName` now resolves the argument's compile-time constant via `SemanticModel.GetConstantValue` and maps it to the canonical enum member, so a const alias (e.g. `K.D` for `AggregateModifier.Distinct`) is recognized instead of silently dropped. When the argument is not a compile-time constant the fix is withheld. Tests cover both the const-alias-converts and the non-constant-bails paths.

Rename `LINQ2DB1001` -> `L2DB1001` applied across the descriptor, `AnalyzerReleases.Unshipped.md`, the readme, the `helpLinkUri`, and the test markup; the wiki page has been renamed to `L2DB1001`.

Also fixed while in here: a salvaged single-line `//` comment on chain scaffolding was appended without a line terminator, swallowing the statement's `;` into the comment and producing non-compiling code; salvaged single-line comments are now line-terminated. 45 analyzer tests green; Release build of `LinqToDB.Analyzers.CodeFixes` clean.

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+
+			// The package carries no linq2db dependency, so it can be installed alongside a linq2db too old to
+			// have the Sql.Window API. WindowFunctionBuilder holds every extension method the rewrite emits; when
+			// it's absent, don't rewrite compiling code into a call to an API that doesn't exist in this compilation.
+			if (model.Compilation.GetTypeByMetadataName("LinqToDB.WindowFunctionBuilder") is null)
+				return null;
+
+			// Collected walking ToValue -> root; ordering lists are reversed to natural order before building.
+			var orderClauses    = new List<OrderClause>();
+			List<OrderClause>? keepOrderClauses       = null;
```
MaceWindu: fix padding

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+			List<OrderClause>? withinGroupOrderClauses = null;
+
+			ArgumentListSyntax? partitionArgs = null;
+
+			var    sawOver     = false;
+			var    keepSeen    = false;
+			var    isKeepFirst = false;
+			var    frameSeen   = false;
+			var    frameIsRange = false;
```
MaceWindu, line 96: fix padding

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+
+						var name = ma.Name.Identifier.Text;
+
+						// KeepFirst/KeepLast and the root analytic function are all declared on the AnalyticFunctions class.
+						if (IsAnalyticFunctionsClass(method.ContainingType))
+						{
+							if (name is "KeepFirst" or "KeepLast")
+							{
+								keepSeen        = true;
+								isKeepFirst     = string.Equals(name, "KeepFirst", StringComparison.Ordinal);
```
MaceWindu: fix padding

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+								break;
+							case "PartitionBy":
+								partitionArgs = inv.ArgumentList;
+								break;
+							case "OrderBy" or "OrderByDesc" or "ThenBy" or "ThenByDesc":
+								orderClauses.Add(new OrderClause { IsDescending = name.EndsWith("Desc", StringComparison.Ordinal), Args = inv.ArgumentList });
+								break;
+							case "ValuePreceding" or "ValueFollowing":
+							{
+								frameSeen = true;
```
MaceWindu: fix padding

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+
+		static void ApplyFrameBoundary(
+			INamedTypeSymbol containingType, string legacyName, ArgumentSyntax? value,
+			ref string? frameStart, ref ArgumentSyntax? frameStartValue,
+			ref string? frameEnd,   ref ArgumentSyntax? frameEndValue)
+		{
+			var mapped = legacyName switch
+			{
+				"UnboundedPreceding" or "UnboundedFollowing" => "Unbounded",
+				"CurrentRow"                                  => "CurrentRow",
```
MaceWindu: fix padding

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+				return null;
+
+			// Split the root call's arguments into positional value args and the special modifier args
+			// (AggregateModifier / Nulls / From / NullsPosition) that become builder calls. Map each written
+			// argument to its parameter through the semantic model rather than by source position: C# named
+			// arguments can be reordered, so a positional match would misclassify a modifier as a value arg (or
+			// vice-versa). Each argument maps to its parameter, so value args are collected with their parameter
+			// ordinal and sorted into declaration order — the order Sql.Window expects.
+			var valueArgsByOrdinal = new List<(int Ordinal, ArgumentSyntax Arg)>();
+			string? distinct = null, nullTreatment = null, fromPosition = null;
```
MaceWindu: never use comosite variable declarations...

Source/LinqToDB.Analyzers.CodeFixes/LegacyWindowChainRewriter.cs
```diff
@@ -0,0 +1,546 @@
...
+			else
+			{
+				// Uniform families: [Distinct][From][Nulls][PartitionBy][order...][frame].
+				if (frameSeen && (frameStart is null || !FrameableFunctions.Contains(functionName)))
+					return null;
+
+				if (distinct      is not null) steps.Add(new Step("Distinct",      SyntaxFactory.ArgumentList()));
+				if (fromPosition  is not null) steps.Add(new Step(fromPosition,    SyntaxFactory.ArgumentList()));
+				if (nullTreatment is not null) steps.Add(new Step(nullTreatment,   SyntaxFactory.ArgumentList()));
+				if (partitionArgs is not null) steps.Add(new Step("PartitionBy",   partitionArgs));
```
MaceWindu: fix padding

Source/LinqToDB.Analyzers/LinqToDB.Analyzers.csproj
```diff
@@ -0,0 +1,45 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
```
MaceWindu, line 3: directory.build.props could contain most of both project files content

Source/LinqToDB.Analyzers/AnalyzerReleases.Unshipped.md
MaceWindu, line 1: ensure release procedure handle those files transition

Tests/Tests.Analyzers/WindowFunctionApiCodeFixTests.cs
```diff
@@ -0,0 +1,1009 @@
...
+				}
+				""");
+
+		[Test]
+		public Task PreservesSingleLineCommentOnChainScaffolding()
+			// A single-line comment on the chain scaffolding travels the same salvage path as the block-comment case,
+			// but runs to end of line — the salvaged comment must be line-terminated so the statement's ';' is not
+			// swallowed into it (which would leave the rewritten code without a terminator and stop compiling).
+			=> Verify.VerifyAsync(
+"""
```
MaceWindu: padding in this test misaligned with other tests

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- New analyzer package `linq2db.Analyzers`, with its first rule `L2DB1001`. It flags the legacy `Sql.Ext` window-function API — `Sql.Ext.<Fn>(...).Over()…ToValue()`, i.e. `LinqToDB.AnalyticFunctions` — as superseded by `Sql.Window`, which is slated for removal in a future major release. Reported at `Info` severity and enabled by default, with a companion code fix ("Convert to Sql.Window API") that rewrites mechanically-convertible chains and supports Fix-All across a document. Aggregates without `.Over()` are not reported — they aren't window functions, so there's no `Sql.Window` target to migrate to. Documented at [L2DB1001](https://github.com/linq2db/linq2db/wiki/L2DB1001). ([#5703](https://github.com/linq2db/linq2db/pull/5703))


New `linq2db.Analyzers` package with rule `L2DB1001` and a code fix, flagging the legacy `Sql.Ext` window API in favour of `Sql.Window`.


---
<sub>Generated from commit `3ea26ac`.</sub>

PR #5707, opened 2026-07-11
F# data-model code generation for the CLI scaffolder (#1553)

### LINKED ISSUE #1553: [T4 template] F# generation recods.
Hi, 

hear that linq2db can generate F# records, is it true? (Also i see some FSharp namespaces in code files). Where can i read about it? 

Thanks)

MaceWindu:
Implements F# data-model generation for the CLI scaffolder: `dotnet linq2db scaffold --target-language f#`. AST + code-generation path only — the T4 template mode stays C#-only.

## What's generated

- **Idiomatic F# records** — entities become `[<Table>]`/`[<Column>]`-attributed records; nullable columns become `'T option` (auto-mapped at runtime via `linq2db.FSharp`'s `UseFSharp()`).
- **Data context** — a `DataConnection`-derived type with a `DataOptions<T>` primary constructor and `GetTable` member accessors.
- `namespace rec` single-file output; additional (non-default) schemas emitted as **modules** (module-per-schema, isolating same-named tables); stored procedures (sync + async `task {}`, `byref<>` out/ref params), table/scalar/aggregate functions, result records, and Find methods.

## CLI

- `--target-language c#|f#` (default `c#`).
- Each option shows a `Supported in:` help line; explicitly setting a C#-only option (`--nrt`, `--partial-entities`, `--add-init-context`, `--add-static-init-context`) under `f#` fails fast rather than being silently ignored.

## Tests

- `Tests/FSharp.Scaffold/` — CLI-generated compile-gate baselines (SQLite, PostgreSQL, SQL Server); a build failure is a generator regression. Regenerate with `Tests/FSharp.Scaffold/generate.ps1`.
- `Tests/FSharp/GeneratedModel.fs` — runtime tests proving linq2db round-trips the generated shape (incl. `option` ↔ SQL `NULL`); pass on SQLite across net462–net10.

## Notes

- Incremental parity: the `IEquatable` comparer, association *extension* methods, and partial-class/`InitDataContext` hooks are gated off for F# for now — records are value-equal already and `namespace rec` covers mutual references.
- VB.NET remains unimplemented (the other half of the `LanguageProviders` TODO).

Closes #1553

## Follow-up commits (review)

- `generate.ps1` resolves provider connection strings from `DataProviders.json` / `UserDataProviders.json` by connection name (no hardcoded values).
- F# nullness: `--nrt` now emits F# 9 `| null` on nullable reference members (nullable scalars stay `'T option`). Both modes are compile-gated: `Tests.FSharp.Scaffold` (`--nrt true`, `<Nullable>enable</Nullable>`) and `Tests.FSharp.Scaffold.NoNrt` (`--nrt false`). F# nullness is project-level, so a file cannot compile in both modes -> two gate projects.


MaceWindu:
This will close last F# ticket

Issue #2953, opened 2021-04-14
Dynamic columns storing not works with `IDataContext.FromQuery(...)` method.


ozzy-ext:
I want to perform custom queries and get results with paging. The application should receive sql-string and produce the page of the queried data. But dynamic columns storing not works with `IDataContext.FromQuery(...)` method.

I use method `FromSql` to query data:
```c#
FromSql<TEntity>(this IDataContext dataContext, RawSqlString sql, params object?[] parameters)
```

My model:
```c#
public class DbEntity
{
  //One of requirements is query should have Id column
  [Column]
  public string Id { get; set; }

  [DynamicColumnsStore]
  public IDictionary<string, object> Properties { get; set; }
}
```

Here I just query all results (without paging):
```c#
IDataContext dc = ....;

var result = dc.FromSql<DbEntity>("select * from main").ToArray();
var found = result.First();

// found.Properties is null
Assert.NotNull(found.Properties);
```

Generated sql query:
```sql
SELECT
	[t1].[Id]
FROM
	(
		select * from main
	) [t1]
```
I think that problem is query wrapping with special selected columns: `[t1].[Id]`.

I tried to use `DataConnection.Query(...)` and it does not wrap input query. But I need `IQueriable` result to use paging. 

How to receive custom columns in `Properties` in my case?

### Environment details
linq2db version: 3.3.0
Database Provider: sqlite
Operating system: windows 10
.NET Framework: .NET 5.0

MaceWindu:
I don't think `FromSql` currently supports columns it isn't aware of through mapping

MaceWindu:
Draft PR #5947 adds `SelectDynamic`, which covers this: the columns you want are named at query-build time, and the wrapper selects them into the `[DynamicColumnsStore]`.

```cs
db.FromSql<DbEntity>("select * from main")
    .SelectDynamic(
        x => new DbEntity { Id = x.Id },
        columnNames,                              // runtime string[]
        (x, name) => Sql.Property<string>(x, name));
```

Against a table whose `Name` column is absent from the mapping, that emits both columns where the wrapper previously emitted only the mapped one:

```sql
SELECT
	[x].[Id],
	[x].[Name]
FROM
	(
		select * from RawRows
	) [x]
```

The column set is materialized to an array and participates in the query cache key, so two different sets never share a compiled query. One call carries one cell type (`TCell`), so columns of different types need one call each.

[...]

PR #5708, opened 2026-07-11
PIVOT / UNPIVOT support (#1475)

### LINKED ISSUE #1475: PIVOT/UNPIVOT support
Tasks:
- [ ] design API
- [ ] Check what databases support it

Documentation:
- SQL Server: https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql?view=sql-server-2017

MaceWindu:
Fixes #1475

Adds three LINQ operators — `SelectDynamic`, `Pivot` and `Unpivot`. All three lower to portable SQL on every provider, in direct and remote (LinqService) contexts. The PR adds no SQL AST node, no `QueryElementType` member, no `QueryElementVisitor` method, no `LinqServiceSerializer` case and no `SqlProviderFlags` member: native `PIVOT`/`UNPIVOT` emission was measured against this lowering on DuckDB, SQL Server 2017 and Oracle 23, won one cell of six, and is not implemented.

## `SelectDynamic`

The underlying operator, and the one thing here that cannot be written today: it materializes **one column per entry of a runtime key set** into a `[DynamicColumnsStore]` member, so the column set is decided when the query is built while every expression stays statically typed.

```cs
sealed class AmountsDto
{
    public int Id { get; set; }

    [DynamicColumnsStore]
    public IDictionary<string, object>? Values { get; set; }
}

var names = new[] { "Usd", "Eur" };         // runtime data, not a compile-time list

db.Amounts.SelectDynamic(
    x => new AmountsDto { Id = x.Id },      // the static part of the projection
    names,                                  // the key set
    (x, n) => Sql.Property<decimal>(x, n)); // one generated column per key

// row.Values["Usd"] -> 10.5m
```

Keys are not restricted to strings. Any key type plus a name factory generates the column names from a pattern:

```cs
static string CurrencyColumn(int id) => $"Currency{id}Amount";

db.Balances.SelectDynamic(
    x => new BalanceDto { Id = x.Id },
    currencyIds,
    (x, id) => Sql.Property<decimal?>(x, CurrencyColumn(id)),
    CurrencyColumn);
```

`SelectDynamic`'s result type has to declare the store itself. `Pivot` is where that requirement is lifted: it can nest its generated columns in a `PivotCells<,>` member of any projection, an anonymous one included (below).

Two implementation points a reviewer will want:

- The per-column expression is instantiated from the template lambda **at the call site**, once per key. Substituting there rather than in a visitor is forced rather than preferred: `ExposeExpressionVisitor` descends into quoted lambdas and rewrites `Sql.Property` while the name argument is still a free parameter, which throws before any builder runs.
- The key set reaches the query as a `SqlQueryDependent` `string[]`, so the column set takes part in the query cache key. A `List<T>` constant is dropped from that key and the dependent comparison is then skipped, which would let two different column sets share one compiled query.

Rejected at the call site, each with a test: a `TResult` with no store member (the generated columns would otherwise be dropped silently), duplicate generated names, names unusable as SQL identifiers, and a generated name that a real member of `TResult` would shadow — member resolution finds the declared member first, so such a column would be accepted and then unreachable.

### Cases it addresses

**[Discussion #4992](https://github.com/linq2db/linq2db/discussions/4992)** — *Selecting multiple dynamic columns into a custom property type.* One column per currency, named by a pattern, with the currency set known only at runtime, selected into a single collection property and then filtered by the same test repeated across that set.

```cs
static string CurrencyColumn(int id) => $"Currency{id}Amount";

// One term per currency. The column name must be a captured local rather than a lambda
// parameter - Sql.Property's name is evaluated before any enclosing lambda is substituted.
static Expression<Func<BalanceDto, bool>> AmountSet(int id)
{
    var name = CurrencyColumn(id);

    return r => Sql.Property<decimal?>(r, name) != null;
}

// Currency1Amount != null || Currency2Amount != null || ... over whatever the set holds
static Expression<Func<BalanceDto, bool>> AnyAmountSet(IEnumerable<int> ids)
    => ids.Select(AmountSet).Aggregate(Or);

static Expression<Func<T, bool>> Or<T>(Expression<Func<T, bool>> left, Expression<Func<T, bool>> right)
    => Expression.Lambda<Func<T, bool>>(
        Expression.OrElse(left.Body, right.Body.Replace(right.Parameters[0], left.Parameters[0])),
        left.Parameters[0]);

var currencyIds = new[] { 1, 2 };

var rows = db.Balances
    .SelectDynamic(
        x => new BalanceDto { Id = x.Id },
        currencyIds,
        (x, id) => Sql.Property<decimal?>(x, CurrencyColumn(id)),
        CurrencyColumn)
    .Where(AnyAmountSet(currencyIds))
    .ToList();

// rows[0].Amounts["Currency1Amount"] -> 10.5m
```

The runtime set drives both halves, and both take part in the query cache key: the same set reuses the compiled query, a different or wider set takes its own entry, and the first set's entry survives the others being compiled in between. The SQL for `{1,3}` projects and filters on `Currency3Amount` and never mentions `Currency2Amount`.

The OR cannot be spelled `currencyIds.Any(id => Sql.Property<decimal?>(r, CurrencyColumn(id)) != null)`. `Sql.Property` is rewritten into a member access before the local-collection `Any` is expanded, so the name is still a bound lambda parameter at that point and the query fails with `InvalidOperationException: variable 'id' … is not defined`. Filed as #5953, together with the diagnostic and the two ways the restriction could be lifted.

Tests: `SelectDynamicTests.SelectsRuntimeCurrencyColumns` on every provider, and `RuntimeCurrencySetDiscriminatesQueryCache` for the cache contract above.

**[Discussion #4248](https://github.com/linq2db/linq2db/discussions/4248)** — *Using Linq2Db with SQLServer Table with unknown extra columns.* A table deployed per customer carries columns that are absent from the shared mapping and are referred to by name, because they are not known upfront.

```cs
// CustomValues maps only Id and CustomerId
var columns = new[] { "WorkLocation", "LastContacted" };

db.GetTable<CustomValues>()
    .SelectDynamic(
        x => new CustomValuesDto { CustomerId = x.CustomerId },
        columns,
        (x, name) => Sql.Property<string>(x, name));

// rows[0].Custom["WorkLocation"] -> "HQ"
```

Test: `SelectDynamicTests.SelectsColumnsMissingFromTheMapping`, on every provider.

**[Issue #2953](https://github.com/linq2db/linq2db/issues/2953)** — *Dynamic columns storing not works with `IDataContext.FromQuery(...)`.* The wrapper a `FromSql` query is built into selects only the mapped columns, so a `[DynamicColumnsStore]` member stays `null`. Naming the wanted columns fills it. This gives that query a working spelling rather than changing what `FromSql` does on its own, so the issue is referenced rather than closed.

```cs
db.FromSql<RawRow>("select * from RawRows")
    .SelectDynamic(
        x => new RawRow { Id = x.Id },
        new[] { "Name" },
        (x, name) => Sql.Property<string>(x, name));

// rows[0].Props["Name"] -> "first"
```

Test: `SelectDynamicTests.SelectsDynamicColumnsOverFromSql`, SQLite only — it embeds literal SQL.

## `Pivot`

Sugar over `GroupBy` plus `SelectDynamic`: the pivoted values are runtime data, while the key, the `FOR` column and every cell expression stay statically typed.

A pivot turns distinct values of one column into columns of their own. Given `Sales`:

| Category | Year | Amount |
|---|---|---|
| A | 2000 | 10 |
| A | 2010 | 20 |
| B | 2000 | 5 |
| B | 2010 | 15 |

grouping by `Category`, pivoting `Year` over `{ 2000, 2010 }` with a `Sum(Amount)` cell gives one row per category and one column per year:

| Key | Y2000 | Y2010 |
|---|---|---|
| A | 10 | 20 |
| B | 5 | 15 |

```cs
var years = await db.Sales.Select(x => x.Year).Distinct().ToListAsync();

db.Sales.Pivot(x => x.Category, x => x.Year, years,
    (g, p) => new { g.Key, Cells = p.Cell(rows => rows.Sum(x => x.Amount), y => $"Y{y}") });
// -> cells reachable as row.Cells["Y2000"] or row.Cells.Get<decimal?>("Y2000")

// ...or into your own type, with static members beside the generated cells
db.Sales.Pivot(x => x.Category, x => x.Year, years,
    g => new SalesDto { Category = g.Key, Total = g.Sum(x => x.Amount) },
    c => c.Cell(rows => rows.Sum(x => x.Amount), y => $"AMT_{y}")
          .Cell(rows => rows.Max(x => x.Note),   y => $"NOTE_{y}"));

// A cell carries any aggregate the provider can translate over the matching rows
db.Sales.Pivot(x => x.Category, x => x.Year, years,
    (g, p) => new { g.Key, Cells = p.Cell(rows => rows.Select(x => x.Amount).Distinct().Count(), y => $"D{y}") });
```

N cell templates crossed with M runtime values give N*M generated columns, so one report can carry several aggregates per pivoted value alongside ordinary static members.

The value set has to be client-side data, which is why the example queries it separately. `forValues` is enumerated at the call site (`forValues as TFor[] ?? forValues.ToArray()`) because the generated column set — how many columns, what they are named, what CLR type each has — is fixed while the expression is built, and those names travel in the query cache key. A value set that is itself a subquery is not known until execution, so there would be nothing to build the projection from. SQL's own `PIVOT` has the same restriction — a literal `IN` list — which is why a dynamic pivot in T-SQL means generating SQL text. Passing an `IQueryable` straight in does compile, since it is also an `IEnumerable`, but it then runs a synchronous query at that point; materializing it yourself keeps the round-trip explicit and lets it be async.

Cells are always built through a lambda over the cell set, and that is what lets them name a source type you cannot write down. C# never infers the type arguments of the type a static member is reached through, so a pivot over a join projected into an anonymous type could not spell `TSource` at all; taking the cells as a lambda over `PivotCells<TSource,TFor>` fixes both from `source` and `forColumn` before the cell lambdas are typed:

```cs
q.Pivot(x => x.e.Id, x => x.e.TheKey, activityIds,
    g => new ListDto { Id = g.Key, CoaName = g.Max(x => x.cm.Name) },
    c => c.Cell(rows => rows.Max(x => x.rir.IdeRub), a => $"IDE_{a}")
          .Cell(rows => rows.Max(x => x.rir.LibRub), a => $"LIB_{a}"));
```

Each cell lowers to `aggregate(g.Where(row => forColumn(row) == value))`, a shape the engine already models: a native `FILTER (WHERE …)` where the provider supports it, `CASE WHEN` emulation everywhere else. An empty cell reads `null` rather than `default(T)` — a cell whose aggregate returns a non-nullable value type has its result lifted.

## `Unpivot`

The inverse: several columns of one row become several rows, each naming the column it came from. Given `Quarterly`:

| Id | Q1 | Q2 | Q3 | Q4 |
|---|---|---|---|---|
| 1 | 10 | 20 | 30 | *NULL* |
| 2 | 5 | *NULL* | 15 | 25 |

unpivoting `Q1`–`Q4` gives one row per non-NULL cell:

| Id | Quarter | Amount |
|---|---|---|
| 1 | Q1 | 10 |
| 1 | Q2 | 20 |
| 1 | Q3 | 30 |
| 2 | Q1 | 5 |
| 2 | Q3 | 15 |
| 2 | Q4 | 25 |

```cs
// By member reference
db.Quarterly.Unpivot(
    (row, quarter, amount) => new { row.Id, Quarter = quarter, Amount = amount },
    x => x.Q1, x => x.Q2, x => x.Q3, x => x.Q4);

// ...or over a column set named at query-build time. With no column selectors, TValue has
// nothing left to be inferred from, so the result selector needs explicit parameter types.
db.Quarterly.Unpivot(
    (Quarterly row, string name, decimal? value) => new { row.Id, Name = name, Value = value },
    quarterNames);

// Several measures per group
db.Monthly.Unpivot(
    (row, quarter, m1, m2, m3) => new { row.Id, Quarter = quarter, M1 = m1, M2 = m2, M3 = m3 },
    ("Q1", x => x.Jan, x => x.Feb, x => x.Mar),
    ("Q2", x => x.Apr, x => x.May, x => x.Jun));
```

The selector overloads exclude NULL cells by default and take `UnpivotNulls` to opt in; the tuple-group overloads keep them. The name column reports the **physical** column name, so a renamed member reads as its column name and the values agree with the database rather than with the CLR member names.

## PostgreSQL `FILTER` gate (#5948)

Filtered aggregates are how pivot cells lower, which surfaced a live defect on master: `PostgreSQLMemberTranslator` reported `IsFilterSupported` unconditionally, and `PostgreSQLDataProvider` hands every version below 9.5 that same translator, so any filtered aggregate on 9.2 or 9.3 emitted SQL the server cannot parse (`42601` on `COUNT(*) FILTER`). Gated at the 9.5 dialect tier — the only boundary the codebase can express, since `PostgreSQLVersion` has no `v94` and the detector maps a real 9.4 server to `v93`. The 34 `AllPostgreSQL93Minus` exclusions annotated `// PostgreSQL 9.4+ (FILTER clause)` are removed and pass.

Three annotated sites keep their restriction — the two in `StringJoinTests` and `Issue1564Tests.CteTest1564` — because they reach `FILTER` through `StringMemberTranslator.TranslateStringJoin`, a sibling nested class this gate cannot reach. That is now tracked as #5952.

Ported to #5949; this branch rebases onto it before merge.

## Tests

`PivotTests`, `PivotDynamicTests` and `SelectDynamicTests`. The result-asserting tests run on bare `[DataSources]` — every configured provider, LinqService included, with no exclusions — so anything a provider genuinely cannot do surfaces as a real failure rather than a gate guessed at in advance. Nine stay narrow for reasons that are not provider support: two assert SQL text, five are call-site guards that throw before any SQL is built, and the rest are query-cache tests plus the one that embeds literal `FromSql` text.

## Review follow-up

`PivotsAvgMinMaxCells` is split into `PivotsMinMaxCells` (bare `[DataSources]`, now asserting both groups) and `PivotsAvgCell`, excluded on `Access.Jet.OleDb`.

The old test pinned a Jet OLE DB failure whose exception type and message are not stable. With byte-identical generated SQL, run [35587574130](https://github.com/linq2db/linq2db/actions/runs/35587574130/job/106297881756) produced `InvalidCastException: The numerical value is too large to fit into a 96 bit decimal` and run [35612267776](https://github.com/linq2db/linq2db/actions/runs/35612267776/job/106385883561) produced `ArgumentException` out of `Decimal.SetBits`, both from `System.Data.OleDb`'s `DbBuffer.ReadNumeric`. The driver defect is filed as #5954; `MIN` and `MAX` stay covered on that provider, since they read the stored column and bind fine.

### Cell API collapsed to one factory member (63dd08269)

The cell surface was largely a verbose alias for the lambda it generates. `PivotCell.Sum/Min/Max/Avg/Count` were wrappers over `Custom`, built by ~50 lines of reflection re-deriving what the compiler already resolves in `rows => rows.Sum(x => x.Amount)`; `Sum` and `Avg` threw `LinqToDBException` at query-build time for a type `Enumerable.Sum` has no overload for, which the lambda form makes a compile error. The `params PivotCell<TSource,TFor>[]` overload pair could not be used with an anonymous `TSource` at all, and forced every other call site to spell its type arguments by hand.

Both are deleted. `PivotCell` is now public with zero public members, `PivotCellFactory.Cell<TCell>` is the whole cell API, and `PublicAPI.Unshipped.txt` drops from 18 lines to 5. The erasure carrier stays: multi-cell is the one thing `Pivot` has over `GroupBy` + `SelectDynamic`, and C# cannot put differently-typed generic lambdas in one `params` array.

The two nullability mechanisms collapse to one, and the pre-existing baselines show why that is safe. `AnEmptyCellOverANonNullableColumnReadsNull` emitted two **identical** `SUM(CASE WHEN ... END)` columns for its wrapper cell and its `Custom` cell, so the mechanisms were never distinguishable in SQL. That test loses the duplicate; every other baseline is byte-identical, including the renamed distinct-count test against its predecessor, which also shows the `Count` cell going from `int` to `int?` is SQL-inert.

Tests convert mechanically and none was a duplicate -- they are provider coverage for the aggregate vocabulary, not tests of the factory members. `PivotsWithACustomAggregate` becomes `PivotsWithADistinctCountCell`. Verified on `SQLite.Classic`, `SQLite.MS`, `DuckDB` and `PostgreSQL.16`: 114 cases green.

### Cells nested in the projection; `PivotRow<TKey>` removed (3a9d54dee)

From @sdanyliv's suggestion to build the cells inside the result selector. Investigating it found the fluency is not what it buys — the placement is. The generated columns were appended to **`TResult`'s own** `DynamicColumnsStore` member, which put the whole contract on the result type: it had to be a declared, mutable class carrying the attribute, so an anonymous projection — the idiomatic LINQ shape — was impossible, `PivotRow<TKey>` existed only to paper over that, and cell names shared a name space with the result's own members, hence the shadowing error.

A second overload takes the cell set as a selector parameter and nests the cells in the member the chain is assigned to, so the result type needs no store of its own:

```cs
db.Sales.Pivot(x => x.Category, x => x.Year, years,
    (g, p) => new
    {
        Category = g.Key,
        Total    = g.Sum(x => x.Amount),
        Cells    = p.Cell(rows => rows.Sum(x => x.Amount), y => $"AMT_{y}")
                    .Cell(rows => rows.Max(x => x.Note),   y => $"NOTE_{y}"),
    });
```

`PivotRow<TKey>` and the no-result-type overload that returned it are deleted — `(g, p) => new { g.Key, Cells = p.Cell(…) }` is the same thing without a library type.

Both placements ship, and they are not redundant: the selector form's chain is read out of an expression tree, so its number of cell **templates** is fixed at compile time, while the `cells` argument of the `DynamicColumnsStore` form is an ordinary `Func` and can build templates in a loop.

`PivotCellFactory` and `PivotCell` collapse into one type. `PivotCells<TSource,TFor>` is the lambda parameter, the chain result and the materialized store, so the `params` array becomes a single chained `Func<PivotCells<,>, PivotCells<,>>` and the whole pivot surface is 9 `PublicAPI.Unshipped.txt` lines across one type and two overloads.

`SelectDynamicBuilder` finds an `Expression.Constant(null, typeof(PivotCells<,>))` placeholder among the constructor's assignments or parameters and splices a nested `SqlGenericConstructorExpression` carrying the `DynamicColumnInfo` assignments into it. With no placeholder the previous top-level path runs unchanged, so `SelectDynamic` is untouched. Identifying the target by the placeholder's type rather than by a member name also reaches a positional record, whose `NewExpression.Members` is null.

A nested cell is still a server-side column. `ComposesAfterAConstantPivot` filters on one after the pivot and the baseline is a bare `GROUP BY` with `HAVING SUM(CASE WHEN [r].[Year] = 2010 …) >= 15`, the unused cell pruned out of the `SELECT` entirely.

**Emitted SQL does not move.** Converting every call site to the new forms leaves `changed: 0` and `removed: 0` across the whole baselines tree, and the two renamed tests hash byte-identical to their `PivotRow` predecessors — the only baseline delta is two file names. `PivotDynamicTests` + `SelectDynamicTests` are 84/84 green on `SQLite.Classic` and `DuckDB`; Release `net10.0`, Release `netstandard2.0`, Tests Release `net10.0` and Tests Release `net462` all build clean.

Four tests go as exact duplicates once the two forms converge: `CellNamedAfterARowMemberThrows` (same mechanism as the `PivotCells` shadowing test), `ComposesOverANestedCell`, `AnEmptyNestedCellReadsNull`, and `SelectDynamicTests.ProjectsIntoBuiltInPivotRow`, which used `PivotRow<int>` as a convenient `SelectDynamic` store type — its one unique assertion, that an ungenerated name reads absent rather than throwing, moves to `PivotsProductionShapeWithoutAResultType`.


jods4:
One sub-feature that I would have use for is supporting dynamic PIVOT values. I currently have to hack some text SQL to achieve this.

In addition to a static anonymous object with `Y2000 = ... `, `Y2001 = ...` (BTW what happens in a native provider if they don't have the same expression?), it'd be really nice if we could target a dictionary property, or if the target class could be have a dynamic property indexer filled by the pivot.

EDIT to clarify a bit: 
By "dynamic" I mean where keys are not know at build time, but are still known at query time.
SQL PIVOT does not support dynamic values based on DB data at runtime, you have to explicitly declare the shape of the SELECT result.

sdanyliv:
Actually, I don’t see any real benefit to this functionality beyond being able to say that we have it. Maybe for LINQPad.

jods4:
@sdanyliv do you mean PIVOT in general or PIVOT on runtime values?

I have to agree that it's a very rare feature, in 20 years of professional dev I must have used it 5 times max.

Sadly, I do maintain one pivot (out of several hundreds requests) in my current project.
And oh boy that query is a monster, that mixes textual search-and-replace, ad hoc `MappingSchema`, many linq2db hacks that tend to break on major updates like v6.
So even though it's only one query, I'd be really happy if I could re-write it natively.

If you're curious, here's the monster in its full glory:
```csharp
public object GetAll(string search)
{
  // We must tell linq2db about all the dynamic pivoted properties that will exist on our model
  var activities = db.IAS_ACTIVITE.Select(a => a.Id).ToList();

  var map = new MappingSchema();

  var entity = new FluentMappingBuilder(map).Entity<ListDto>();
  foreach (var id in activities)
  {
    entity.Property(x => Sql.Property<string>(x, "IDE_" + id));
    entity.Property(x => Sql.Property<string>(x, "LIB_" + id));
  }
  entity.Build();
  db.AddMappingSchema(map);

  // IAS_MOD_E is an unpivoted table.
  // Column COA_MASK_ID (e.Id) identifies one template,
  // but there are repeated lines for each Activity as identified by e.THEKEY.
  // So the business key of IAS_MOD_E is e.COA_MASK_ID (aka Id) + e.THEKEY
  // The template fields (ATI_RUB_SEQ_*, POS_TRI_*, etc.) are identical on all lines.
  // IAS_RUB_SEQ is the pivoted value that is different per Activity (e.THEKEY).
  var query =
    from e in db.IAS_MOD_E
    // This join is here to keep only activities that are configured in IAS_ACTIVITE
    // This is relevant for the uncoded activity filter, we don't want to take into account
    // activities that are not available.
    join act in db.IAS_ACTIVITE on e.THEKEY equals act.Id
    // This is all regular, unpivoted values
    join cm in db.COA_MASK on e.Id equals cm.Id
    from arp in db.ATI_RUB.LeftJoin(arp => e.ATI_RUB_SEQ_POS == arp.Id)
    from arn in db.ATI_RUB.LeftJoin(arn => e.ATI_RUB_SEQ_NEG == arn.Id)
    from arr in db.ATI_RUB.LeftJoin(arr => e.ATI_RUB_SEQ_RES == arr.Id)
    // This is the pivoted field, that provides a different value for each Activity (THEKEY).
    from rir in db.IAS_RUB.LeftJoin(rir => rir.Id == e.IAS_RUB_SEQ)
    group new { e, cm, arp, arn, arr, rir } by e.Id into g
    select new ListDto
    {
      // We have to use Expr to hack the column alias into expression, as linq2db will not produce it in ToSqlQuery() :(
      Id = Sql.Expr<int>($"{g.Key} Id"),
      // These fields are all the same in groups.
      // There are exceptions/corrupt in DB though, which is why we keep the latest change (last ModifiedAt) :(
      COA_NAME = Sql.Expr<string>($"{g.AnyValueKeepLast(x => x.cm.COA_NAME, x => x.e.ModifiedAt)} COA_NAME"),
      ACC_CODE_FORMAT= Sql.Expr<string>($"{g.AnyValueKeepLast(x => x.cm.ACC_CODE_FORMAT, x => x.e.ModifiedAt)} ACC_CODE_FORMAT"),
      ATI_RUB_SEQ_POS = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.e.ATI_RUB_SEQ_POS, x => x.e.ModifiedAt)} ATI_RUB_SEQ_POS"),
      ATI_RUB_SEQ_NEG = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.e.ATI_RUB_SEQ_NEG, x => x.e.ModifiedAt)} ATI_RUB_SEQ_NEG"),
      ATI_RUB_SEQ_RES = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.e.ATI_RUB_SEQ_RES, x => x.e.ModifiedAt)} ATI_RUB_SEQ_RES"),
      POS_TRI_RUB = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.arp.TRI_RUB, x => x.e.ModifiedAt)} POS_TRI_RUB"),
      POS_LIB_RUB = Sql.Expr<string>($"{g.AnyValueKeepLast(x => x.arp.LIB_RUB, x => x.e.ModifiedAt)} POS_LIB_RUB"),
      NEG_TRI_RUB = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.arn.TRI_RUB, x => x.e.ModifiedAt)} NEG_TRI_RUB"),
      NEG_LIB_RUB = Sql.Expr<string>($"{g.AnyValueKeepLast(x => x.arn.LIB_RUB, x => x.e.ModifiedAt)} NEG_LIB_RUB"),
      RES_TRI_RUB = Sql.Expr<int>($"{g.AnyValueKeepLast(x => x.arr.TRI_RUB, x => x.e.ModifiedAt)} RES_TRI_RUB"),
      RES_LIB_RUB = Sql.Expr<string>($"{g.AnyValueKeepLast(x => x.arr.LIB_RUB, x => x.e.ModifiedAt)} RES_LIB_RUB"),
      // This next line is tricky:
      // - We cannot use the same Sql.Expr<bool> trick because linq2db wraps any bool expression in a `CASE WHEN x THEN 1 ELSE 0 END`
      //   I'll fix it with some search-and-replace in sql below :(
      // - The reference to `rir.Id` instead of `e.IAS_RUB_SEQ` is intentional to prevent linq2db from dropping the `rir` left join.
      //   It's unused in this select but we use it in the FromSql below.
      // - I have to wrap in `AsSql` to avoid linq2db performing the `== 0` client-side, which would break our nested SQL request below
      HasUncodedActivity = Sql.AsSql(g.Min(x => Sql.ToNullable(x.rir.Id) ?? 0) == 0),
      // The text /*PIVOT*/ will be replaced by dynamic columns in FromSql below.
      ModifiedAt = Sql.Expr<DateTime>($"/*PIVOT*/{g.Max(x => x.e.ModifiedAt)} ModifiedAt"),
    };

  var sql = query.ToSqlQuery().Sql;

  var pivotIndex = sql.IndexOf("/*PIVOT*/");

  // linq2db doesn't preserve `e` as table alias because of `group by e.Id` and uses `g_1` instead...
  var ias_mod_e = Regex.Match(sql, @"IAS_MOD_E (\w+)").Groups[1].Value;

  var pivotColumns = string.Concat(activities.Select(a =>
    $@"any_value(decode({ias_mod_e}.THEKEY, {a}, rir.IDE_RUB)) IDE_{a},
       any_value(decode({ias_mod_e}.THEKEY, {a}, rir.LIB_RUB)) LIB_{a},
       "));

  sql = sql.Insert(pivotIndex + "/*PIVOT*/".Length, pivotColumns);

  // Add HasUncodedActivity alias
  var previousComma = sql.LastIndexOf(',', pivotIndex);
  sql = sql.Insert(previousComma, " HasUncodedActivity");

  // FromSql is the only way I've found to dynamically select columns.
  // Caveats:
  // - Don't use C# formattable strings, holes are converted to parameters, which doesn't work for subqueries.
  // - Because `query` is turned into a string here, it shouldn't use SQL parameters -- otherwise use `InlineDbParameters()` option.
  var pivoted = db.FromSql<ListDto>(sql);

  return pivoted.SearchWords(search,
    (x, f) => Sql.Collate(x.COA_NAME!, "binary_ai").Contains(f)
           || Sql.Collate(x.ACC_CODE_FORMAT!, "binary_ai").Contains(f));
}
```

sdanyliv:
@jods4, 
That’s exactly my point. For the API to be usable in applications, it needs to be dynamic from the outset. What I see here is complex code that will rarely, if ever, be used.

Source/LinqToDB/Internal/Linq/Builder/SelectDynamicBuilder.cs
```diff
@@ -0,0 +1,71 @@
...
+			var staticBody = SequenceHelper.PrepareBody(staticSelector, sequence);
+
+			var generic = staticBody switch
+			{
+				MemberInitExpression memberInit => new SqlGenericConstructorExpression(memberInit),
+				NewExpression newExpression     => new SqlGenericConstructorExpression(newExpression),
+				_ => throw new LinqToDBException("SelectDynamic static selector must be an object construction expression (new T { ... })."),
+			};
+
+			generic = generic.WithMappingSchema(builder.MappingSchema);
```
sdanyliv: We use sequence.MappingSchema

Source/LinqToDB/Internal/Linq/Builder/UnpivotBuilder.cs
```diff
@@ -0,0 +1,372 @@
...
+			var sourceExpr    = ((MethodCallExpression)buildInfo.Expression).Arguments[0];
+			var sourceContext = builder.BuildSequence(new BuildInfo(buildInfo, sourceExpr, new SelectQuery()));
+			var tableContext  = SequenceHelper.GetTableContext(sourceContext);
+			if (tableContext?.SqlTable is not { } sourceTable)
+				return null;
+
+			var valueCount = groups[0].columns.Length;
+			var node       = new SqlUnpivotTable(sourceTable, includeNulls: false);
+
+			var nameField = new SqlField(builder.MappingSchema.GetDbDataType(typeof(string)), "Name", false) { Table = node };
```
sdanyliv: Wrong MappingSchema in whole file.

sdanyliv (changes requested):
PR Body do not reflect what is inside PR.

Source/LinqToDB/Internal/Linq/Builder/SelectDynamicBuilder.cs
MaceWindu: Applied. `WithMappingSchema` now takes `sequence.MappingSchema`, and the `DynamicColumnsStoreAttribute` guard moved below the sequence build so `resultType`'s descriptor resolves through the same schema that materializes the store - resolving the guard through the context's schema while the store itself is built from the sequence's was the same split.

Not yet pushed; lands in the next commit on this branch.

Source/LinqToDB/Internal/Linq/Builder/UnpivotBuilder.cs
MaceWindu: Applied, and it turned out to be a real defect rather than a convention slip.

The file's single use fed `GetPhysicalColumnName`, and it runs before anything is built: the builder lowers to `Concat(Select(source, ...), ...)` and hands that to `BuildSequence`, so there was no sequence to ask. The source is now built ahead of the lowering purely to read its `MappingSchema`, with its own `SelectQuery` so that throwaway build cannot write into the query the lowered branches share (`new BuildInfo(buildInfo, expr)` would have handed it the parent's).

Measured, not assumed: a table captured before a `UseMappingSchema` scope that renames a column, unpivoted inside the scope, emitted `SELECT [row_1].[Id], [row_1].[Q1] ...` while labelling the value `Q_RENAMED` - a name for a column the query never read. Red on SQLite and DuckDB, direct and LinqService; green after the fix. Pinned by `PivotTests.UnpivotNameColumnUsesTheSourcesSchema`.

The wider audit is out of scope here and tracked in #5951: 35 `builder.MappingSchema` reads under `Internal/Linq/Builder`, twelve of them suspect, the strongest being `AssociationHelper.CreateAssociationQueryLambda`, which ignores the `MappingSchema` parameter its only caller passes (`AssociationHelper.cs:30` against `:401`). That issue also carries the proposal for an internal `LINQ2DB0002` analyzer to enforce the rule. Nothing outside these two builders is changed on this branch.

Not yet pushed; lands in the next commit on this branch.

Source/LinqToDB/Internal/Linq/Builder/SelectDynamicBuilder.cs
MaceWindu: Pushed as `dbe4b5636`.

Source/LinqToDB/Internal/Linq/Builder/UnpivotBuilder.cs
MaceWindu: Pushed as `dbe4b5636`, with `PivotTests.UnpivotNameColumnUsesTheSourcesSchema` as the regression pin.

PR #5710, opened 2026-07-11
tests: DuckDB in-memory preload, SqlServer READPAST fix, and fixture hygiene

MaceWindu:
Directly-unrelated test fixes/refactors split out of the parallel-test-execution work (#5614) so they can merge to master independently.

## DuckDB in-memory testing for local / filtered runs

Make DuckDB's default in-memory (matching SQLite) rather than file-based with a CI-only in-memory override. DuckDB can't preload an in-memory database via `COPY FROM DATABASE` — it inserts child rows before parents and fails the FK constraint — so `TestsInitialization.SetupDuckDBInMemory` seeds the shared in-memory DB by round-tripping the committed on-disk file through `EXPORT`/`IMPORT DATABASE`, which loads schema → data → constraints in phases and preserves sequence values. Filtered / local runs (which skip `a_CreateData`) then still have tables and data; full runs re-seed on top (the create script drops first).

- `DataProviders.json`: `CommonConnectionStrings` DuckDB → in-memory shared cache.
- `Build/Azure/*/duckdb.json`: Providers-only (inherit the Common in-memory string).

Verified on net10: a no-seed filtered run reads back the preloaded data correctly (incl. decimal / blob / hugeint via PARQUET); a full seeded run passes.

## SqlServer READPAST hint isolation fix

`WithReadPastTableTest` / `WithReadPastInScopeTest` (and the `ReadPast` value of `TableHint2005PlusTest`) failed sporadically with *"You can only specify the READPAST lock in the READ COMMITTED or REPEATABLE READ isolation levels"*: with no explicit transaction they inherit the pooled connection's leftover isolation level (SqlClient doesn't reset it on pool return). Pin it with an explicit `READ COMMITTED` transaction, and disable remote/LinqService for that case (the client-side transaction can't set the remote server connection's isolation).

## Test-fixture hygiene refactors

Drop shared / vestigial state from three fixtures:

- **Issue822Tests** — `ID1`/`ID2` were shared fixture instance fields mutated mid-test and captured by the query closures; now method locals, so the `[Values]` cases don't share mutable state (closures still capture + re-read them at execution, the point of #822).
- **Issue1373Tests** — drop the vestigial `Query.ClearCaches()` from all six tests: each uses a distinct local `MappingSchema` (part of the compiled-query cache key), so there's no stale-cache collision to clear.
- **MappingSchemaTests.DoNotUseComplexAttributes** — read the entity descriptor from a local `MappingSchema` instead of the process-global `MappingSchema.Default`.

### DIFF
--- a/Build/Azure/net100/duckdb.json
+++ b/Build/Azure/net100/duckdb.json
@@ -2,11 +2,6 @@
 	"NET100.Azure": {
 		"Providers": [
 			"DuckDB"
-		],
-		"Connections": {
-			// CI runs DuckDB against a shared-cache in-memory DB (no per-commit fsync); one keep-alive
-			// connection is held open in TestsInitialization for the run.
-			"DuckDB": { "Provider": "DuckDB", "ConnectionString": "Data Source=:memory:?cache=shared" }
-		}
+		]
 	}
 }
--- a/Build/Azure/net80/duckdb.json
+++ b/Build/Azure/net80/duckdb.json
@@ -2,11 +2,6 @@
 	"NET80.Azure": {
 		"Providers": [
 			"DuckDB"
-		],
-		"Connections": {
-			// CI runs DuckDB against a shared-cache in-memory DB (no per-commit fsync); one keep-alive
-			// connection is held open in TestsInitialization for the run.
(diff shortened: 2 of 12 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `c9d87b9`.</sub>

Issue #3009, opened 2021-05-25
Limiting query cache size


yallie:
Hi, is sliding expiration setting the only way to tweak the query cache behavior?
Is there a way to limit LINQ query cache size, either in bytes or in entries?
Looks like there is a setting called `MemoryCacheOptions.SizeLimit`, but is it really used?

MaceWindu:
Right now cache management is quite limited. We can provide more flexibility in next release but for that we should understand what and why you want to manage.

yallie:
My nightly CI tests have rough memory measurements, i.e.

```csharp
var memoryBefore = GC.GetTotalMemory(true);

RunTests();

var memoryAfter = GC.GetTotalMemory(true);
```

I noticed that almost each test run increases application's memory footprint:

![image](https://user-images.githubusercontent.com/672878/119481322-3abbec80-bd5b-11eb-81c8-964dd5e825a1.png)

After some test runs memory usage may drop, but most of the time it grows.
I know it's not a leak because memory usage would reduce later as the app stays idle.

To test whether it's related to the query cache, I tried using `Query.ClearCaches()`.
Indeed it seems to help keeping memory usage more or less consistent during CI test runs.
But I'd like to avoid just clearing the caches if there is a way to control the memory footprint.

Issue #4639, opened 2024-09-10
[BUG] Memory leak issue, when using linq2db efcore library 7.7.0


aloksharma1:
Hello Guys,

in the version 7.7.0 of [linq2db.EntityFrameworkCore](https://github.com/linq2db/linq2db.EntityFrameworkCore), main library version 5.4.0 with ef core 7.0.20. I am seeing a sudden surge and retention of objects in linq2db memory cache for much longer than required and it's not disposing correctly.

if I remove linq2db implementation from that part of code it stops leaking memory. I can confirm this issue was not present in my project with previous versions as its an existing project and we just did a library version update and nothing else. 
i did a memory diagnostic too and saw that each of the snapshot linq2db was holding big objects (14-16 mb per request).
its an mvc application, i cant share the code but i think if you can make a complex 3 level table and do merge/complex crud/union on it you will be able to see it.

@MaceWindu @sdanyliv

aloksharma1:
@MaceWindu critical problem, i have updated my project to net 8 and for ef core and linq2db to this package config for net 8

```
<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
	<PackageReference Include="linq2db.EntityFrameworkCore" Version="8.2.0-rc.13787" />
	<PackageReference Include="LinqKit.Microsoft.EntityFrameworkCore" Version="8.1.5" />
	<PackageReference Include="Microsoft.EntityFrameworkCore.Design" Version="8.0.8">
		<PrivateAssets>all</PrivateAssets>
	</PackageReference>
	<PackageReference Include="Microsoft.EntityFrameworkCore.DynamicLinq" Version="8.4.5" />
	<PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="8.0.8" />
	<PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="8.0.8">
		<PrivateAssets>all</PrivateAssets>
	</PackageReference>
</ItemGroup>
```

now another kind of error issue i got in trying to access my already working code on net 7 version
```
[18:02:00 ERR] An error occurred during the normalization of blog posts.
[...]

PR #5711, opened 2026-07-11
Unify caching under a shared registry + primitives; retire the MemoryCache fork; fix #5692

### LINKED ISSUE #5692: Rework MappingSchema attribute caching — unbounded growth (drives NETFX multi-config OOM)
## Summary

Per-`MappingSchema` attribute caching (`MappingAttributesCache`) grows **unbounded** and is the dominant memory-growth component in long, multi-configuration runs. On the NETFX CI legs that run two SQL Server versions' suites in a single process, this is the primary driver of the `OutOfMemoryException` (the query-cache cap added earlier bounds a *different* cache and doesn't touch this one).

## Evidence (heap profiling)

Reproduced the failing CI shape locally: **net462 (x64), `SqlServer.2014` + `SqlServer.2016` in one testhost process**. Live-heap snapshots (ClrMD histogram over a full self-minidump) at three test counts:

| tests | WorkingSet | `MappingAttributesCache` (dict nodes + arrays) | strings / NUnit result tree |
|---|---|---|---|
| 1,156 | 479 MB | ~15 MB | fixed |
| 2,759 | 455 MB | ~23 MB | fixed |
| 5,722 | 675 MB | ~54 MB | fixed |

- The `ConcurrentDictionary` node count climbs ~**60 entries per test** (146K → 327K over 2,759 → 5,722) — roughly linear, no plateau.
- Every other large bucket is **fixed**, not growing: `System.String` (~54 MB), NUnit result XML (`XmlText`/`XmlAttribute`, ~134K each — flat), and the `TestCaseResult`/`TestExecutionContext` tree (pre-built by NUnit for the matched set). So this cache is the component that scales badly across a full suite × 2 providers.

## Mechanism

`MappingAttributesCache` (`Source/LinqToDB/Internal/Mapping/MappingAttributesCache.cs`) is a per-`MappingSchema` instance (`MappingSchema._cache` / `_firstOnlyCache`) of `ConcurrentDictionary<CacheKey/Key, MappingAttribute[]>` populated via `GetOrAdd` with **no eviction, no size cap, no clear**. Keyed by `(attributeType, member, owner)`. On the long-lived schemas (provider base schemas + `MappingSchema.Default`) it retains one entry per distinct key for the process lifetime. Distinct **anonymous / projection types** (one per query shape) appear to mint fresh keys that are never released → unbounded. With two providers, each lookup is cached twice (one entry per mapping-schema config), so a multi-version process pays ~2×.

## Proposed direction (this cache, and caching in general)

1. **This cache:** skip caching anonymous / compiler-generated types (finite entity types are safe to cache; the unbounded keys come from per-projection anon types), and/or add an eviction policy / size bound (mirroring `QueryCache.MaxEntriesOverride`).
2. **In general:** audit the other per-`MappingSchema` / provider `GetOrAdd` caches for the same unbounded-growth shape and settle on a consistent bounded-cache strategy. Related to #2502 (decouple mapping attributes from metadata).

## Notes

- Not a regression from a specific change — long-standing; surfaced now because the restructured NETFX CI legs batch two SQL Server versions into one process.
- Distinct from the non-deterministic parallel crash (stack overflow) also seen on parallel runs — that's a separate problem, not this OOM.

---
🤖 Investigated with [Claude Code](https://claude.com/claude-code) (net462 x64 heap profiling, SqlServer 2014+2016 single-process repro).

MaceWindu:
Unifies linq2db's four ad-hoc caching technologies under one management layer, retires the in-tree `Microsoft.Extensions.Caching.Memory` fork, and fixes the #5692 unbounded-growth OOM.

## What

- **New `LinqToDB.Internal.Cache` framework** — `ILinqToDBCache` + `CacheRegistry` (caches self-register at construction → a *provably-total* clear-all and diagnostics), plus the `BoundedCache<TKey,TValue>` storage primitive (BitFaster-backed LRU / time-LRU / LFU). (A `WeakCache` was prototyped and then dropped — no viable consumer: its intended target, `_combinedSchemasCache`, uses a composite `(ms1, ms2)` key that `ConditionalWeakTable` can't express and only caches stable/lockable schemas anyway, so there is nothing to unpin.)
- **Retired the 18-file in-tree MemoryCache fork**; all ~15 consumers (mapping schema, command-info readers, compiled CRUD/query caches, provider detection, serialization converters, remote services) now use `BoundedCache`.
- **`Tools.ClearAllCaches()` routes through the registry** — closes 6 caches that were previously unreachable from clear-all (remote services, provider-version detection, serialization converters, combined mapping schemas).
- **`QueryCache`** joined the management layer (its tuned bucketed storage is unchanged) and gained `BeginMeasure()` — an `AsyncLocal` flow-scoped hit/miss probe (parallel-safe) — plus process-global `TotalHits` / `TotalMisses`.
- **#5692 fix** — `MappingAttributesCache` no longer caches negative (empty) attribute lookups. Measured ~86–91% of its entries were empty; recompute is cheap (~53 ns on net462, behind the `AttributesExtensions` cache) and infrequent (gated by cached `EntityDescriptor` construction).
- **Diagnostics + config** — public `Tools.GetCacheStatistics()` and opt-in `Configuration.Cache.WorkCacheEntryLimit` (default `null` = unbounded, behavior-preserving).

## Dependency

Adds **`BitFaster.Caching`** (ships `netstandard2.0`, so it works on every TFM incl. net462) — the core library's first functional runtime dependency, replacing the hand-forked cache. (`FastCache.Cached` was evaluated and rejected: net6.0/net8.0-only.)

## Supersedes

- **#5681** (QueryCache memory-pressure via a `Gen2GcCallback` port) — replaced by joining the unified registry; the existing per-bucket + global-entry caps + sweep already bound memory.
- **#5694** (MappingSchema anonymous-type-skip band-aid) — replaced by the root-cause negative-caching fix.

Also relevant: #3009 (configurable cache size) — partially addressed by `Configuration.Cache.WorkCacheEntryLimit` + the new statistics.

## Testing

- Release green on `net10.0` (analyzers) and `netstandard2.0` throughout.
- `CachingTests` 27/27 on SQLite, including the new `QueryCacheMeasurementTracksHitsAndMisses` probe test and the `MappingAttributesCacheDoesNotCacheNegativeLookups` #5692 regression (verified via the public statistics API; post-fix empty-entry count 0).
- #5692 measured red→green: `AttributeReader` cache 84→12 entries / 72→0 negatives; `MappingSchema` cache 132→12 / 120→0.
- `MappingAttributesBenchmark` (BenchmarkDotNet + `MemoryDiagnoser`) added under `Tests/Tests.Benchmarks/` for the attribute-lookup path (empty / attributed / many-distinct) — captures ns/op + allocations; not yet run.
- Full cross-provider CI pending (`/azp run`).

## Follow-ups (not in this PR)

- Migrate the ~280 `Query<T>.CacheMissCount` test assertions to `BeginMeasure()` **solely so those tests can run in parallel** (the global counter is shared mutable state that clobbers across parallel tests; the flow-scoped probe doesn't). `CacheMissCount` and the global `GetMissCount` / `TotalMisses` / `TotalHits` counters are **kept** — `BeginMeasure()` is the additive flow-scoped view, not a replacement.
- Optional diagnostics-only registration of the 5 process-static derived converter/extension-builder caches (left unregistered here to preserve clear-all behavior — see the design-doc comment).

Fixes #5692

Full what / why / how in the design-doc comment below.

## Follow-up commit(s) — review fixes (`6c91846`, `1b3717a`)

Code review across correctness, concurrency/GC/cross-TFM, and API/tests. No correctness defects found in the migration, the #5692 negative-caching fix, or the `CacheRegistry` weak-ref lifecycle; `PublicAPI.Unshipped.txt` matches the reshaped API. Four findings, all fixed:

- **`QueryCache` measurement scoped to the instance** (`6c91846`) — `_activeMeasurement` / `_activeMeasurementCount` were `static` while `BeginMeasure()` is an instance method, so a scope folded in hit/miss events raised by any `QueryCache` instance on the same flow. Made both fields instance and threaded the owner through `CacheMeasurement`; the production gate is now per-instance. Harmless in production (only `Default` exists) — removes a test-isolation trap.
- **`[NonParallelizable]` on the three new process-global cache tests** (`1b3717a`) — `QueryCacheMeasurementTracksHitsAndMisses` asserts on global `TotalHits`/`TotalMisses`/`GetQueryCacheStatistics()` (which a concurrent `ClearAll` resets), `MappingAttributesCacheDoesNotCacheNegativeLookups` calls `Tools.ClearAllCaches()`, and `WorkCacheEntryLimitRejectsValuesBelowMinimum` mutates global config. Serialized so they neither false-fail nor pollute sibling fixtures.

Verified: `CachingTests` 29/29 on SQLite.Classic (net10.0), Release net10.0 core build clean.

Accepted as-is: on net462/netstandard2.0, `BoundedCache.GetOrAdd`'s `#else` fallback allocates a per-call closure (BitFaster's netstandard2.0 assembly lacks the closure-free overload) — unavoidable given the dependency's TFM split.


MaceWindu:
# Design: unifying caching in linq2db

This PR unifies caching. Below is the full rationale — the problem, what was measured, the decisions and *why*, the architecture, and what was deliberately left alone — so the diff reads in context.

## Problem

linq2db had **four** independent caching technologies:

1. An in-tree **fork of `Microsoft.Extensions.Caching.Memory`** (`Internal/Cache/`, ~10 consumers) — every consumer constructed it with empty options, so its size-limit/LRU machinery was dead code; only sliding-TTL ran.
2. The bespoke bucketed **`QueryCache`** (from #5501).
3. Microsoft's real `MemoryCache` (EF Core package).
4. ~40 raw `ConcurrentDictionary` caches — many actually *registration stores*, not evictable caches.

Two consequences:

- **Fragmented clear-all.** `Tools.ClearAllCaches()` reached only three things; six caches (remote services, provider-version detection, serialization converters, combined mapping schemas) were unreachable from it.
- **Unbounded growth → OOM** (#5692 and a long history: #2696, #4639, #2151, #256…).

## What we measured (decisions are data-driven, not guessed)

- **Cache taxonomy.** ~40 of ~50 caches are keyed by `Type`/`MemberInfo` — a keyspace bounded by application metadata. LRU there is useless or harmful (evicting compiled delegates/reflection just forces recompute). Only a short list has an unbounded runtime keyspace.
- **#5692 root cause (corrects the issue's own hypothesis).** Not "anonymous types minted per query" (those are compile-time, bounded). It is **negative-lookup dominance**: linq2db probes attributes across huge numbers of attribute-less `(attributeType × member × owner)` combinations, and every empty result was cached forever. Measured split on a projection workload: **85.7%** of `AttributeReader`'s static cache and **90.9%** of the per-schema cache were empty entries.
- **Is dropping empties safe?** Micro-benchmark of the recompute path (`GetAttributes<MappingAttribute>(false)`, behind the `AttributesExtensions` cache), 5M iters, Release: **net462 52.8 ns/op, net10 21.9 ns/op** — ~2×/1.5× a bare dict lookup, and infrequent (attribute lookups happen during *cached* `EntityDescriptor` construction). ⟹ dropping negative caching is safe.
- **Dependency options.** `FastCache.Cached` is net6.0/net8.0-only → unusable in core. `BitFaster.Caching` ships `netstandard2.0` → usable on every TFM.

## Decisions

1. **Unify management, not storage.** One contract (`ILinqToDBCache`) + a central `CacheRegistry`, so clear-all and diagnostics are provably total. Storage stays a minimal **two-primitive** family; `QueryCache` keeps its bespoke engine (per-query TTL/hotness/entry-cap don't fit a per-key LRU).
2. **Two primitives.** `BoundedCache<TKey,TValue>` (BitFaster `ConcurrentLru`/`ConcurrentTLru`/`ConcurrentLfu`; strong-keyed, size-bounded, optional sliding TTL, closure-free `GetOrAdd`) and `WeakCache<TKey,TValue>` (`ConditionalWeakTable`; GC-driven, for collectible-object keys). Registration stores stay plain dictionaries — explicitly *not* caches.
3. **Adopt BitFaster; retire the fork.** Accepts the core library's first functional runtime dependency in exchange for deleting 18 hand-maintained fork files. (The netstandard2.0 BitFaster assembly lacks the closure-free `GetOrAdd(key,factory,arg)`, so that overload is `#if NET6_0_OR_GREATER`-gated with a closure fallback — caught by the netstandard2.0 build.)
4. **Behavior-preserving migration (Policy A).** Migrated work caches keep sliding-TTL + an "effectively unbounded" default capacity; real bounding is opt-in via `Configuration.Cache.WorkCacheEntryLimit`. The one unavoidable change: TTL is captured at cache construction, and per-context sliding-expiration overrides on process-static caches collapse to the global default.
5. **This work supersedes #5681 and #5694** (per maintainer direction).

## Architecture

- `ILinqToDBCache` — `Name`, `Kind`, `Clear()`, `GetStats()`. Caches self-register at construction: `Register` (strong, process-static) or `RegisterScoped` (weak, per-owner — the registry never extends an owner's lifetime).
- `CacheRegistry` — `ClearAll()` / `Clear(kind)` / `Snapshot()`, pruning dead weak refs. `Tools.ClearAllCaches()` = `CacheRegistry.ClearAll()` + `Query.ClearCaches()` (the latter still drains the legacy `CacheCleaners` queue: `MemberCache`, `IdentifierBuilder`).
- `QueryCache` implements `ILinqToDBCache` explicitly (so its public surface is unchanged) and `.Default` self-registers.
- Public diagnostics: `Tools.GetCacheStatistics()` → `IReadOnlyList<CacheStatistics>`.

## Phasing (commit-grouped)

0 foundation → 1 migrate consumers + retire fork + unify clear-all → 2 QueryCache management + probe + global totals → 3 #5692 negative-caching fix → 4 diagnostics + config + long-tail audit.

## Long-tail audit — why most caches are left unregistered

The old `ClearAllCaches` never cleared any of the ~40 tail stores, and all are metadata-bounded (finite keyspace). So leaving them unregistered is behavior-preserving *and* leak-free:

- **~15 registration-stores** (user `SetConverter`/`MapMember`/fluent mappings, provider config, reserved words, transient-exception registrations, per-schema config) — clearing them is data loss.
- **11 metadata-bounded reflection caches** (all of `AttributesExtensions`, `TypeAccessor._accessors`, `ReflectionExtensions.*`, `ValueComparer`, …) — finite, recomputable, never cleared before. Notably `AttributesExtensions` is the *last* layer before raw reflection, so dropping its negatives would mean µs-scale cold-reflection recompute — deliberately kept.
- **5 process-static derived caches** (`Converter._converters`, `ExprHolder<T>`, `ConvertInfo.Default`, the two extension-builder caches) — optional diagnostics-only registration, left out here to preserve clear-all behavior.

## Verification

Release green on net10.0 + netstandard2.0; `CachingTests` 27/27 on SQLite incl. the new probe + #5692 regression tests; #5692 measured red→green (negatives → 0). Full cross-provider CI pending.

*Prepared with Claude Code.*

MaceWindu:
**Correction to the design-doc comment above** — two refinements landed after it was posted:

- **One storage primitive, not two.** `WeakCache` was prototyped and then dropped — it had no viable consumer. Its intended target, `_combinedSchemasCache`, uses a composite `(ms1, ms2)` key that `ConditionalWeakTable` can't express, and it only caches stable/lockable schema pairs anyway (transient schemas aren't lockable), so there is no schema pinning to fix. The other single-`Type`-key candidates are metadata-bounded (audit verdict: leave). Ship `BoundedCache` alone; reintroduce a weak-keyed cache when a genuine need appears (e.g. collectible `AssemblyLoadContext` support).
- **Added `MappingAttributesBenchmark`** under `Tests/Tests.Benchmarks/` (BenchmarkDotNet + `MemoryDiagnoser`) for the attribute-lookup path — ns/op **and allocations** for the empty/negative, attributed, and many-distinct-empty cases. Not yet run.

MaceWindu:
### `MappingAttributesBenchmark` results — net10.0, `InProcessEmit`, `[MemoryDiagnoser]`

| Benchmark | Mean | Allocated/op |
|---|---:|---:|
| `EmptyLookup` — repeated negative lookup (recomputed, no longer cached) | 159.6 ns | **0 B** |
| `AttributedLookup` — cached hit | 56.9 ns | **0 B** |
| `ManyDistinctEmptyLookups` — 30 distinct cold negatives | 18.6 µs | 1,888 B (~63 B/lookup) |

Confirms the #5692 trade-off: dropping negative caching is **allocation-free** for repeated empty lookups (the shared `Array.Empty<T>` is returned), and allocates only *transiently* on distinct cold lookups (the inheritance-tree `GetInterfaces()` walk) — memory that is recomputed-and-released, versus the old behaviour that cached every empty result and **retained it forever** (the unbounded growth this PR fixes).

_Note: BDN's child-process multi-runtime jobs (net462/8/9) produced no results in this AV-restricted local environment — that's why these cache benchmarks force the in-process toolchain (+ a manual-runner fallback). The `InProcessEmit` net10 job is shown; the full TFM matrix runs on CI._

MaceWindu:
## Review follow-up — findings fixed in `abe68e9d`…`69b605a4`

Reviewed the caching-unification change across code correctness, cross-TFM / concurrency / GC, API + tests, and baselines. CI was green on every provider/TFM leg, baselines were clean (1 added), and the public API matched `PublicAPI.Unshipped.txt`. Five findings, all fixed in the pushed commits:

- **MAJ001** — `Configuration.Cache.WorkCacheEntryLimit` of `0`/`1`/`2` crashed the first work-cache construction with a `TypeInitializationException`: the value fed BitFaster's cache capacity, which requires `>= 3`. Now validated in the setter (throws `ArgumentOutOfRangeException` on `< 3`), documented, and covered by a guard test. (`abe68e9d`)
- **MIN001** — `CacheRegistry._weak` accumulated dead `WeakReference` wrappers unless `ClearAll`/`Snapshot` ran; now pruned opportunistically every 256 scoped registrations. (`a7cd675d`)
- **MIN002** — `CacheStatistics.Count` / `CacheStats` documented a `-1` "size unavailable" sentinel left over from the dropped `WeakCache`; no backing store emits it, so the clause was removed. (`660d7d07`)
- **MIN003** — the new knob had no test; now covered by the MAJ001 guard test.
- **SUG001** — `BoundedCache.Build` silently dropped `expireAfterAccess` when the policy was LFU; now throws `NotSupportedException`. (`69b605a4`)

Verified: Release `net10.0` build clean, guard test green.

**Release-note reminders** (behavior changes, no separate code action):
- Work-cache sliding TTL is now global and captured at type initialization — `Configuration.Linq.CacheSlidingExpiration` and `WorkCacheEntryLimit` must be set at startup, before the first cache is touched.
- `Tools.ClearAllCaches()` now also clears remote-service, provider-version-detection, serialization-converter, and combined-mapping-schema caches (previously unreachable from clear-all).

MaceWindu:
Behavior note (release-notes candidate): the cache-unification migration changes how work caches read the sliding-expiration timeout.

The retired `MemoryCache` path set `entry.SlidingExpiration = dataContext.Options.LinqOptions.CacheSlidingExpirationOrDefault` on **every** insert, so it re-read the value per entry and honored a per-`DataOptions` override. The migrated `BoundedCache` work caches — `QueryRunner` compiled-query caches, `CommandInfo` object/parameter/data-reader readers, `SerializationConverter`, `MappingSchema` `EntityDescriptors`/`CombinedSchemas`, `ProviderDetector` versions, `RemoteDataContext` services — instead capture the global `Configuration.Linq.CacheSlidingExpiration` **once**, at first static access.

Net effect: a per-context `CacheSlidingExpiration` override, or any change to the global value made after these caches are first touched, no longer affects the work caches — the value must be set at application startup. The LINQ plan cache (`QueryCache`) is unaffected: it still reads `CacheSlidingExpirationOrDefault` per entry via `ResolveBaseTimeout`.

Blast radius is small (few users tune this), but it's a user-visible change worth a release-notes line.

igor-tkachev (changes requested):
Adding `BitFaster.Caching` as a mandatory runtime dependency of the core `linq2db` package is not acceptable.

`linq2db` still supports .NET Framework 4.6.2, where dependency resolution, transitive packages, assembly-version conflicts, and binding redirects already cause significant problems for users. We should not expand the dependency graph of the main library for an internal cache implementation detail. The PR already needs TFM-specific fallback code because the `netstandard2.0` package surface differs from the newer assemblies, which demonstrates the additional compatibility burden this dependency introduces.

This is especially difficult to justify because the actual #5692 fix does not require `BitFaster.Caching` at all. The reported unbounded growth is addressed by not caching negative mapping-attribute lookups. That fix should not force every `linq2db` user to consume a new runtime dependency.

Please remove the dependency from the core package.

If a bounded cache is still considered necessary, acceptable alternatives include:

- implementing the required cache primitive inside `linq2db`;
- incorporating only the necessary implementation into the repository with appropriate license attribution;
- introducing a replaceable cache strategy without imposing an external dependency on all users;
- moving the cache implementation and third-party integration into a separate optional `linq2db.caching` package;
- or keeping this PR focused on the `MappingAttributesCache` fix and handling broader cache infrastructure separately.

An optional `linq2db.caching` package could reference `BitFaster.Caching` and provide the corresponding cache strategy while the core `linq2db` package remains free of this dependency.

There are also broader scope and behavior concerns in the current PR:

- `CommandInfo` caches previously used `dataConnection.Options.LinqOptions.CacheSlidingExpirationOrDefault` per connection, but now use the global `Configuration.Linq.CacheSlidingExpiration` fixed at static cache construction time. This silently drops per-`DataOptions` expiration semantics.
- `WorkCacheEntryLimit == null` is described as behavior-preserving/unbounded in the PR description, but the implementation maps it to a hard limit of 100,000 entries. That introduces eviction and potential recompilation where none existed before.
- The PR combines the #5692 bug fix with a complete cache-subsystem rewrite, a new public diagnostics API, query-cache measurement, new eviction semantics, and a new runtime dependency. These should not be treated as one inseparable change.

The dependency must be removed before this PR can be approved.

igor-tkachev:
I think this PR should be split.

The #5692 fix is small and well-defined: stop caching negative mapping-attribute lookups, add the regression test, and keep the change focused.

A broader cache redesign should be handled separately and only after analysing the actual cache scenarios and requirements across linq2db: key cardinality, lifetime, recomputation cost, eviction tolerance, expiration semantics, per-context configuration, and .NET Framework compatibility.

At the moment, a powerful general-purpose external cache implementation is introduced first, and a set of very different caches is then migrated to fit it. That is the wrong direction for this kind of infrastructure change.

We should first determine what each cache actually needs, then choose or implement the smallest appropriate strategy. The result might be a simple internal primitive, several specialized strategies, or an optional `linq2db.caching` package, but this should not be decided implicitly inside the #5692 bug fix.

viceroypenguin:
To @igor-tkachev's point, whatever caching system we use should at the very least, on net8+, support Microsoft.extensions.caching. This allows consumers to configure caching as part of the larger framework and infrastructure, and reuse existing code.

igor-tkachev:
I agree.

Different users and application stacks prefer different caching implementations, so linq2db should support replaceable cache strategies instead of hard-coding one provider into the core library.

The default implementation can be simple and dependency-free, with optional packages for advanced integrations:

- `linq2db.Caching.Microsoft.Extensions`;
- `linq2db.Caching.BitFaster`.

`Microsoft.Extensions.Caching` also requires separate package dependencies for a regular `net8.0` library, so it belongs in an optional integration assembly as well.

This allows users to reuse their existing caching infrastructure without forcing a specific implementation and its dependencies on every linq2db consumer.

PR #5715, opened 2026-07-12
Tests/CI: stop committing partial baselines from failed tests

MaceWindu:
## Problem

CI regenerates SQL baselines into `linq2db.baselines` and opens a baselines PR. On providers with test-task restarts enabled (`retry: true` — Oracle 12–23, Access ACE), a test that dies **between its direct and remote (`.LinqService`) contexts** left a truncated / one-context `.sql` baseline, because the harness flushed the captured SQL unconditionally on teardown regardless of pass/fail. Those partials got committed.

A prior mitigation added `git -C baselines reset --hard` + `clean -fdq` at the head of every retry-enabled test task. It ran at the wrong time: because the test tasks in one job share the same `baselines/` working tree, a later task's `reset --hard` wiped valid baselines earlier tasks had already produced (including per-EF-version `.EF8/.EF9/.EF10` files no other task regenerates).

## Fix

- **Root cause** (`Tests/Base/BaselinesManager.cs`): `Dump` now returns early when the test failed (`TestContext.CurrentContext.Result.FailCount > 0`), so a failed / partially-run test never writes a baseline; a passing (re-)run regenerates it in full. Since direct and remote produce identical SQL, skipping the failing context leaves the passing context's normal file — which matches the existing baseline, so no spurious diff. The intentional direct≠remote divergence check in `BaselinesWriter` (throws *inside* `Dump`, while `FailCount` is still 0) is untouched.
- **CI** (`Build/Azure/pipelines/templates/test-workflow-{windows,linux,macos}.yml`): removed the now-unnecessary retry-task `reset --hard` / `clean` cleanup (14 sites). `retryCountOnTaskFailure` and `$(test_retry_args)` are kept; the final "Commit test baselines" step (gated on `succeeded()`) is unchanged, so a hard job failure still commits nothing.

## Verification

- `dotnet build Tests/Linq/Tests.csproj -c Testing` — clean (0 warnings / 0 errors).
- The three edited pipeline templates validated as YAML.
- End-to-end proof requires a baselines-enabled CI run on an affected provider; Oracle flakiness is non-deterministic, so confirmation is the absence of partial baselines in subsequent baseline PRs.

### DIFF
--- a/Build/Azure/pipelines/templates/test-workflow-linux.yml
+++ b/Build/Azure/pipelines/templates/test-workflow-linux.yml
@@ -125,13 +125,7 @@ steps:
   condition: and(variables.title, eq(variables.net80, 'true'), succeeded(), ne(variables.retry, 'true'), ${{ parameters.full_run }})
   displayName: 'Tests (.NET 8): $(title)'
 
-- script: |
-    if [ -d baselines/.git ]; then
-      echo "Reset baselines working tree so a failed/retried attempt does not leak baselines into the commit"
-      git -C baselines reset --hard
-      git -C baselines clean -fdq
-    fi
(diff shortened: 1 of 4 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `2e54607`.</sub>

PR #5720, opened 2026-07-13
Ship the analyzers with linq2db via the linq2db.Analyzers package

MaceWindu:
Makes the user-facing Roslyn analyzers + code fixes reach every consumer of `linq2db`, by giving `linq2db` a dependency on the `linq2db.Analyzers` package — the shape EF Core uses for `Microsoft.EntityFrameworkCore` → `Microsoft.EntityFrameworkCore.Analyzers`.

`master` has no edge at all from `linq2db` to the analyzer package, so the package ships but nothing pulls it in: the rules reach nobody unless a consumer references `linq2db.Analyzers` by hand. This PR closes that gap.

## How the rules arrive

| Consumer references | Rules delivered |
|---|---|
| `linq2db` | yes — via the `linq2db.Analyzers` dependency |
| only a satellite (`linq2db.Tools`, `Remote.*`, `Extensions`, `Compat`, `Scaffold`, `EntityFrameworkCore`) | yes — the whole chain leaves `analyzers` in the flow |
| `linq2db.Analyzers` alone, against linq2db 6.1–6.3 | yes — the package carries no `linq2db` dependency, and each rule is symbol-gated |

## Changes

- **`LinqToDB.Analyzers.CodeFixes.csproj`** — stays the packable `linq2db.Analyzers` project. Only change vs `master`: the assemblies now pack under **`analyzers/dotnet/roslyn4.8/cs`** instead of the unversioned `analyzers/dotnet/cs`, so hosts below Roslyn 4.8 skip them silently rather than emitting CS9057, which a consumer's `TreatWarningsAsErrors` would turn into an error.
- **`LinqToDB.csproj`** — `<ProjectReference Include="../LinqToDB.Analyzers.CodeFixes/…" PrivateAssets="contentfiles;build" />`. `analyzers` is deliberately left out of `PrivateAssets` — it is the asset being flowed on; the default (`contentfiles;analyzers;build`) asks NuGet *not* to flow it. No version cap, unlike EF Core's: nothing here consumes `linq2db` internals, so a newer analyzer against an older runtime stays a supported combination.
- **Satellite packages** — the same metadata on their own `linq2db` / `Tools` reference. A hop that excludes `analyzers` suppresses the whole `linq2db.Analyzers` subtree for that package's consumers; EF Core keeps the same chain unbroken (`EFCore.SqlServer` → `EFCore.Relational` → `EFCore` → `EFCore.Analyzers` are all `exclude="Build"` in the published nuspecs).
  - Correction: that suppression is what NuGet documents, not what the current SDK does. Measured on SDK 10.0.302 with packages built from this branch, each variant restored into its own isolated `globalPackagesFolder`: with the packed nuspecs patched to `exclude="Build,Analyzers"` on **both** the `linq2db` → `linq2db.Analyzers` and `linq2db.Tools` → `linq2db` edges, the rules still reach a `linq2db.Tools`-only consumer. So this metadata is contract conformance, not a fix for an observed delivery failure.
- **`Source/LinqToDB/readme.md`** — documents where the rules come from, the toolchain floor (.NET SDK 8.0+ / VS 2022 17.8+), the `.editorconfig` severity + `apply_fix_on_return_type_mismatch` options, and the analyzer-against-older-runtime workflow.

## Why not bundle the analyzers inside `linq2db`

Earlier commits on this branch packed both assemblies into the `linq2db` package itself, on the premise that *"NuGet does not flow analyzers through transitive dependencies, so a satellite-only consumer needs a direct `linq2db` reference"*. That premise does not hold — measured on SDK 10.0.302 with a consumer whose only `PackageReference` is `linq2db.Tools`, restored from a folder feed into an isolated `globalPackagesFolder`:

- the analyzers reach the compiler (`csc /analyzer:…`) despite the `exclude="Build,Analyzers"` that `linq2db.Tools.nuspec` carries on its `linq2db` edge;
- identically at `analyzers/dotnet/cs` and `analyzers/dotnet/roslyn4.8/cs`, so it is not an artefact of the version-specific path;
- `ExcludeAssets="analyzers"` on the consumer's own reference does not suppress it either.

So bundling bought nothing that a package dependency doesn't, and cost the two things the analyzer package was designed for: installing it against 6.1–6.3 to plan a migration before upgrading, and shipping a code-fix correction without a runtime release. It also required a custom `_BundleAnalyzersIntoPackage` target hooked to the SDK-private `_GetPackageFiles`, plus a 104-line CI gate to notice when that hook silently stopped firing — 127 of the 203 lines this branch had added. All removed.

Found by @sdanyliv's review; the measurements above were reproduced independently before the rewrite.

## Verified

- Release build green for every project whose reference metadata changed: `LinqToDB`, `Tools`, `Scaffold`, `Extensions`, `Compat`, all five `Remote.*`, `EntityFrameworkCore.EF3`/`EF10` — 0 errors.
- Packed `linq2db`, `linq2db.Analyzers`, `linq2db.Tools`: `linq2db.nuspec` carries `<dependency id="linq2db.Analyzers" … exclude="Build" />`, `linq2db.Tools.nuspec` carries `<dependency id="linq2db" … exclude="Build" />`, the analyzer package holds both DLLs under `analyzers/dotnet/roslyn4.8/cs` and no `lib/`, and no analyzer assembly leaks into any other package's `lib/`.
- Consumer probes against those packages: a project referencing only `linq2db`, and a project referencing only `linq2db.Tools`, both get L2DB1001 reported by `csc` from the analyzer package.
- `Tests.Analyzers`: 51/51 passed (Release).
- `net462` Release leg of `Source/LinqToDB` clean — a `netstandard2.0` Roslyn component referenced from the `net462` TFM produces no warnings.

## Outside this repo

- Wiki [L2DB1001](https://github.com/linq2db/linq2db/wiki/L2DB1001): `Package` row points back at `linq2db.Analyzers`, reverting `a300430`. `Category` row now reads `LinqToDB` per `c76ee6110`, and the severity section documents the category-wide line plus the `EnableLinqToDBAnalyzers` property (`linq2db.wiki` `e6806fc`).
- `linq2db.EntityFrameworkCore`'s own repo uses default metadata on its `linq2db` `PackageReference`. Delivery works today regardless, but adding `PrivateAssets="contentfiles;build"` there would make that hop match the rest of the chain.
- `linq2db.Analyzers` ships for the first time in 6.4.0 — needs a release-notes line and inclusion in the expected-package check at publish time.

## Follow-up commits

Review pass over the reworked branch (`889154d6c..bea5c8c9b`).

- **`ba7e79947`** — adds `Build/Azure/scripts/verify-analyzer-delivery.ps1`, run in the nugets job. Delivery rests on `PrivateAssets="contentfiles;build"` across twelve `ProjectReference` lines and nothing verified it: a satellite added later with default metadata would drop the contract-guaranteed path, and stay invisible because the current SDK delivers analyzers across an excluding edge anyway. The check reads the packed nuspecs — `linq2db` must declare the analyzer dependency in every framework group, no package may exclude `Analyzers` on its `linq2db` / `linq2db.Tools` / `linq2db.Analyzers` edge (`linq2db.cli` and `linq2db.LINQPad` are skip-listed, since nothing compiles against them), and the analyzer package must carry both assemblies under `analyzers/**/cs`. Exercised against synthetic package drops for the clean case and each failure mode, plus a real pack of this branch.
- **`62d1a8611`** — corrects two wrong claims in the `LinqToDB.csproj` comment. `master` delivers no rules because it has no reference to the analyzer project at all, not because of the default `PrivateAssets` value. And the `project.assets.json` evidence cited for the contract argument does not hold: no package has an `analyzers` group in that file (0 occurrences across the 50 assets files in this repo, including `Microsoft.CodeAnalysis.BannedApiAnalyzers` and `Meziantou.Analyzer`, which do deliver), because NuGet's lock-file `targets` schema has no such group. See the measurement noted under *Changes* above.
- **`bea5c8c9b`** — documents that `runtime` staying out of `PrivateAssets` makes the code-fix assembly copy-local downstream, so each of the seven `linq2db.cli.<rid>` sub-packages carries it under all three `tools/<tfm>/<rid>` payloads: 6 files per RID, a few hundred KB against packages that already run 64–164 MB, never loaded. No library package's `lib/` is affected, and `linq2db.LINQPad` is not either — its packaging target packs an explicit file list rather than the build output.

Review pass on @igor-tkachev's second review.

- **`2e85ad95b`** — stops the packaging-only reference entering unrelated build graphs. The SDK's `IncludeTransitiveProjectReferences` injects `@(_TransitiveProjectReferences)` into `@(ProjectReference)` before `AssignProjectConfiguration`, which is what made the code-fix project a compile *and* copy-local item for every downstream project. `Source/Directory.Build.props` now removes only the injected item, leaving LinqToDB's own direct reference — and with it the `linq2db.Analyzers` dependency in `linq2db.nuspec` — untouched. That clears both `MSB3277` warnings on `LinqToDB.EntityFrameworkCore.EF3` (0 warnings on `net462` and `netstandard2.0`) and takes the code-fix assembly out of every `linq2db.cli` RID payload, superseding the *Known cost* section this PR carried. The three metadata alternatives were each measured and rejected: `ReferenceOutputAssembly="false"` drops the nuspec dependency entirely, `Private="false"` covers only the direct hop (`Scaffold` and `CLI` still copy it), and `PrivateAssets` hiding `compile`/`runtime` breaks P2P build ordering.
- **`8b78cf8eb`** — adds the `EnableLinqToDBAnalyzers` MSBuild opt-out, documented in both readmes, implemented by `build/` + `buildTransitive/linq2db.Analyzers.targets` removing this package's analyzers from `@(Analyzer)`. It hooks `ResolveLockFileAnalyzers` — `ResolvePackageAssets` runs before `@(Analyzer)` exists, so hooking there removes nothing. Currently the only opt-out that works at all, since `ExcludeAssets="analyzers"` on a consumer's `linq2db` reference does not suppress the rules. Verified on packed packages across four consumers (`linq2db` and `linq2db.Tools`-only, each with and without the property): it reaches a consumer two hops from the analyzer package, so `exclude="Build"` on the intermediate edge does not block `buildTransitive`.
- **`bffd1fe74`** — `verify-analyzer-delivery.ps1` also asserts both packed opt-out targets folders. Both readmes document the property and only the packed file implements it, so losing either folder would turn it into a no-op with nothing at build time noticing. Verified the check fails when the `buildTransitive` entry is stripped.
- **`c76ee6110`** — moves `L2DB1001` off the `Usage` category, which is shared with Microsoft's analyzers, to `LinqToDB`. Roslyn's only bulk knob is per category, so under `Usage` a consumer could not set the severity of the linq2db rule set without also hitting every CA rule in that category. Free to change before the package's first release, breaking after it. Verified with a consumer whose `.editorconfig` carries only `dotnet_analyzer_diagnostic.category-LinqToDB.severity = error` and no per-rule line.

Also merged `origin/master` into the branch.


igor-tkachev (dismissed):
Looks good overall. Bundling the analyzer and code-fix assemblies into the main package is implemented cleanly, without leaking Roslyn dependencies into the runtime package or running the user-facing analyzers against linq2db itself. The remaining concerns around package-level integration coverage and documentation clarity are non-blocking.

sdanyliv (changes requested):
Requesting changes on one point: the readme paragraph added by `874f7bca5` / `3f0a36298` states something that does not reproduce, and it ships inside the published package and is already mirrored to the wiki. Section 1 below, backed by the measurements.

The finding behind it is wider — analyzers already reach a consumer through transitive package dependencies, **including in this PR's own bundled layout** — which removes the stated reason for bundling. Sections 2–5 set out the one-line alternative and compare the two layouts, including where bundling genuinely wins. That part is a proposal, not a blocker: the packaging layout is a maintainer call, and nothing in it makes the PR incorrect.

## Measurements

Consumer project multi-targeting `net462;netstandard2.0;net10.0`, SDK 10.0.204, local folder feed, single `PackageReference` to **`linq2db.Tools` only** (no `linq2db` reference, no analyzer reference). `.editorconfig` raises L2DB1001 to `error`, so delivery is a build pass/fail signal. Source under test is the analyzer's own test snippet:

```cs
using LinqToDB;

class C
{
	long M(int x) => Sql.Ext.RowNumber().Over().PartitionBy(x).OrderBy(x).ToValue();
}
```

| # | Layout | `linq2db` → `linq2db.Analyzers` | `linq2db.Tools` → `linq2db` | L2DB1001 |
|---|---|---|---|---|
| A | separate package | `PrivateAssets="contentfiles;build"` (`exclude="Build"`) | default (`exclude="Build,Analyzers"`) | **reported** |
| B | separate package | `PrivateAssets="contentfiles;build"` | `PrivateAssets="contentfiles;build"` | **reported** |
| C | **this PR** — bundled in `linq2db` | no dependency | default (`exclude="Build,Analyzers"`) | **reported** |
| D | separate package | default (`exclude="Build,Analyzers"`) | default | **reported** |

Two further probes:

- `ExcludeAssets="analyzers"` on the consumer's own `linq2db.Tools` reference does **not** suppress it either — asset exclusions apply to the assets of the package they name, not to its subtree.
- `ReferenceOutputAssembly="false"` on the `linq2db` → analyzer project reference drops the package dependency from the nuspec entirely (`<dependencies>` empty). Use a plain `ProjectReference`.

Config **C** is this PR as merged. A satellite-only consumer gets the analyzers regardless of the `exclude="Build,Analyzers"` that `linq2db.Tools.nuspec` carries on its `linq2db` edge.

C and D were each run against packages built for all five shipping TFMs and reported on **every** consumer leg — `net462`, `netstandard2.0` and `net10.0` — so this is not a modern-TFM-only effect. A and B were measured on the `net10.0` leg.

Config D was additionally measured in the PR's primary scenario — a plain consumer whose only reference is `linq2db`, no satellite involved. `linq2db.Analyzers` enters the restored graph unprompted:

```
linq2db/6.4.0-local.1
linq2db.Analyzers/6.4.0-local.1
...
```

and L2DB1001 is reported on all three consumer legs.

## 1. The shipped readme states the opposite, and it is wrong

`874f7bca5` / `3f0a36298` added to `Source/LinqToDB/readme.md` — i.e. into the published package, and mirrored to the wiki:

> NuGet does not flow analyzers through transitive dependencies, so a project that references only a satellite package (`linq2db.EntityFrameworkCore`, the Tools or Remote packages) needs a direct `linq2db` reference to get them.

Config C above is exactly that scenario and the analyzer fires. This paragraph tells users to add a reference they do not need, and it is the stated justification for the whole change. It needs removing regardless of which layout wins.

## 2. The goal is reachable in one line

Master has no edge from `linq2db` to the analyzer package at all, so the delivery gap is real — but closing it is a single `ProjectReference` with default metadata (config D, measured):

```xml
<ProjectReference Include="../LinqToDB.Analyzers.CodeFixes/LinqToDB.Analyzers.CodeFixes.csproj" />
```

That yields `<dependency id="linq2db.Analyzers" version="6.4.0" exclude="Build,Analyzers" />` in `linq2db.nuspec`, and a consumer of `linq2db.Tools` alone gets L2DB1001. No `PrivateAssets` tuning on any satellite project, no packaging target, no CI gate. Same observable outcome as this PR.

(EF Core reaches the same place from the other direction: `Microsoft.EntityFrameworkCore.Analyzers` stays a separate package and `EFCore.csproj` references it with `PrivateAssets="contentfiles;build"`, so the published nuspec carries `exclude="Build"` on that one dependency and `exclude="Build,Analyzers"` on its siblings.)

## 3. What bundling costs once its stated benefit is gone

- **`_BundleAnalyzersIntoPackage` hooks `_GetPackageFiles`**, an SDK-private target with no compatibility contract. If it stops firing on an SDK bump, `pack` succeeds and produces an analyzer-free package. The documented hook for contributing package content from a target is `TargetsForTfmSpecificContentInPackage`; the supported route for shipping analyzers is a packable project, which this PR removes.
- **`verify-analyzer-packaging.ps1` (+104), `build-job.yml` (+13), `nuget-job.yml` (+10) = 127 of 203 added lines** exist only to notice that the private hook broke. `ad2de30d2` documents the two silent modes. A packable analyzer project needs none of it: a project that fails to produce its assemblies fails `pack` loudly.
- **L2DB1001 becomes unavailable on the versions it migrates from.** It flags legacy `Sql.Ext` window API; its audience is on 6.2/6.3. Bundling makes 6.4 a prerequisite for the diagnostic that helps prepare the 6.4 upgrade. Listed under *Trade-offs accepted*, but with the delivery argument gone it is no longer paying for anything.
- **Code-fix servicing follows the runtime release.** #5703 landed a long tail of code-fix corrections (named-argument mapping, `DISTINCT`+`KEEP` bail, return-type-fit gate, Fix-All rewrite). Each future one would need a full linq2db release.

## 4. Side by side

Measured rows are marked; the rest are properties of the two layouts.

| | Bundled (this PR) | Separate package + dependency |
|---|---|---|
| Direct `linq2db` consumer gets analyzers | yes | **yes** (measured) |
| Satellite-only consumer gets analyzers | **yes** (measured — the readme says no) | **yes** (measured) |
| Consumer csproj changes needed | none | none |
| Packaging mechanism | custom target on the SDK-private `_GetPackageFiles` | stock packable project + one `ProjectReference` |
| A broken packaging change is caught by | a dedicated 104-line CI gate | `pack` failing |
| Analyzer/runtime version skew | impossible | floor dependency; see below |
| Installable on 6.2 / 6.3 to plan the migration | no | yes |
| Analyzer fix without a runtime release | no | yes |
| Package ids on nuget.org | 1 | 2 |
| Consumer's dependency tree | no extra node | `linq2db.Analyzers` appears |
| Opt-out | `.editorconfig` severity | `.editorconfig` severity (`ExcludeAssets` works in neither — measured) |

On skew: a `ProjectReference` packs `version="6.4.0"`, which is a *minimum*, not a pin — NuGet would accept a higher `linq2db.Analyzers` if something else in the graph asked for one. Nothing else does, and the analyzer is symbol-gated through `GetTypeByMetadataName` (`LinqToDB.WindowFunctionBuilder`), so a mismatched pair goes quiet rather than misbehaving. If exact lockstep is wanted anyway, the dependency can be pinned to `[$(Version)]`.

Bundling's honest wins are the last three rows on its side: one id, no extra node in the tree, and no skew at all. The proposal trades those for a supported packaging path, a CI gate that can be deleted, and the ability to ship the migration analyzer to the people who are still on the API it flags. On the id count — `Source/Directory.Build.props` sets `IsPackable=true` repo-wide and linq2db already publishes ~15 packages, so the 16th is not the expensive part of this decision.

## 5. The 6.2 / 6.3 row is a designed capability, not a side effect

That row reads as a nice-to-have in the table above, but the capability is engineered on purpose in #5703, by two mechanisms this PR removes.

**The analyzer package deliberately carries no `linq2db` dependency.** `LinqToDB.Analyzers.CodeFixes.csproj` on master:

```xml
<SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
```

A package built from master confirms it: the produced `linq2db.Analyzers.nuspec` has no `<dependencies>` element at all, so the package composes with any linq2db version. This PR drops the property — the csproj at `7c7f88d9` is down to `IsPackable=false`.

**The code fix has a capability gate written for exactly this case.** `LegacyWindowChainRewriter.cs:79-83`:

```cs
// The package carries no linq2db dependency, so it can be installed alongside a linq2db too old to
// have the Sql.Window API. WindowFunctionBuilder holds every extension method the rewrite emits; when
// it's absent, don't rewrite compiling code into a call to an API that doesn't exist in this compilation.
if (model.Compilation.GetTypeByMetadataName("LinqToDB.WindowFunctionBuilder") is null)
    return null;
```

Under bundling that branch is unreachable: linq2db is always present, always at the matching version.

**How far back it actually works.** `WindowFunctionBuilder` is present in `Source/LinqToDB/Sql/Sql.Window.cs` at `v6.1.0`, `v6.2.0` and `v6.3.0`, so on all three the gate is satisfied — a user on a shipped release gets the diagnostic *and* a working Fix-All against their existing runtime:

```xml
<PackageReference Include="linq2db"           Version="6.3.0" />
<PackageReference Include="linq2db.Analyzers" Version="6.4.0" PrivateAssets="all" />
```

That is the natural workflow for a migration rule: size the change, apply the fix, then upgrade — rather than upgrade first to find out what the rule would have told you.

So the row is not a trade-off being accepted; it is a designed capability being withdrawn, with the code implementing it left as dead weight.

## Requested change

Drop the bundling — `_BundleAnalyzersIntoPackage`, the `IsPackable=false` flip on `LinqToDB.Analyzers.CodeFixes`, `verify-analyzer-packaging.ps1` and its two pipeline steps — and replace it with the plain `ProjectReference` above. Rewrite the readme's *Analyzers* section: drop the transitive-flow claim, and say the analyzers arrive with `linq2db` (directly or transitively).

## Keep

- `analyzers/dotnet/roslyn4.8/cs` instead of master's unversioned `analyzers/dotnet/cs`, in `LinqToDB.Analyzers.CodeFixes.csproj` — correct independently of this decision; older hosts skip silently rather than warning under `TreatWarningsAsErrors`.
- The readme's toolchain floor wording (.NET SDK 8.0+ / VS 2022 17.8+) and the `.editorconfig` severity + `apply_fix_on_return_type_mismatch` documentation.
- `a7a65444c` (`verify-nuget-sizes.ps1` `[switch] $NoAzdoLogs`) and `2df081783` (listing it in `linq2db.slnx`) — unrelated to the packaging question, both valid.
- The wiki `L2DB1001` page fix, with the `linq2db.Analyzers` package row restored.

## Timing

`linq2db.Analyzers` has not been published — `https://api.nuget.org/v3-flatcontainer/linq2db.analyzers/index.json` returns 404, and 6.3.0 (2026-05-17) predates #5703 (2026-07-12). Either layout is still free to choose. After 6.4.0 ships, one of them is a dead package id on nuget.org.

## Repro / scope of the measurement

`dotnet pack -p:EnablePackageValidation=false` of `LinqToDB.Analyzers.CodeFixes`, `LinqToDB` and `LinqToDB.Tools` into a folder feed (`-c Debug` for the five-TFM runs, `-c Testing` for the `net10.0`-only ones); consumer restored from that feed with the extracted packages cleared from `%USERPROFILE%\.nuget\packages` between configurations, and `obj`/`bin` removed before each `--no-incremental` build. The satellite runs reference `linq2db.Tools`; the direct run references `linq2db`. All runs on SDK 10.0.204 — behaviour on older SDKs was not measured, though the `net462` and `netstandard2.0` consumer legs above make a modern-only restore rule unlikely.

MaceWindu:
Confirmed and applied — the bundling is gone, `linq2db` depends on `linq2db.Analyzers` again. Pushed in `732f2ecca..889154d6c`; title and body rewritten, wiki row reverted (`linq2db.wiki` `57e9064`).

Reproduced the central measurement independently before rewriting anything: SDK 10.0.302, packages built from this branch into a folder feed, consumer restored with an isolated `globalPackagesFolder`, single `PackageReference` to `linq2db.Tools`. `csc` gets both analyzer DLLs and L2DB1001 fires, despite the `exclude="Build,Analyzers"` on the Tools → linq2db edge. Two additions to your data:

- It is not specific to the `roslyn4.8` path — repacked at `analyzers/dotnet/cs` and it flows identically. So the readme sentence was never true, not even for master's layout.
- `project.assets.json` shows NuGet *did* apply the exclusion: the target entry for the analyzer-carrying library has no `analyzers` group at all. The SDK hands the assemblies to `csc` anyway. Since `analyzers` is in the default `PrivateAssets` set, delivery through an excluded edge contradicts the documented contract — it works on 10.0.204 and 10.0.302, but it isn't a guarantee either layout should rest on.

That last point is why the implementation deviates from the requested change in two places:

1. **`PrivateAssets="contentfiles;build"`, not a plain `ProjectReference`.** Your configs A/B, not D. The plain default writes `exclude="Build,Analyzers"` — it delivers only because the exclusion is disregarded. The EF Core precedent you cite is the A/B shape: `Microsoft.EntityFrameworkCore`'s published nuspec carries `<dependency id="Microsoft.EntityFrameworkCore.Analyzers" version="9.0.0" exclude="Build" />`.
2. **Every satellite hop carries it too.** An intermediate edge that excludes `analyzers` suppresses the whole subtree, so `Tools` / `Remote.*` / `Extensions` / `Compat` / `Scaffold` / `EntityFrameworkCore` would still have depended on the bypass. EF keeps the same chain unbroken end to end — `EFCore.SqlServer → EFCore.Relational → EFCore → EFCore.Analyzers` are all `exclude="Build"` in the published nuspecs.

No version cap, unlike EF's `[$(Version), $(NextMajorVersion))`: nothing in the analyzer consumes `linq2db` internals, and an exact or capped range would break the newer-analyzer-against-older-runtime case (NU1107) that section 5 of your review is about.

Verified after the change: `linq2db.nuspec` → `exclude="Build"` on the analyzer dependency, `linq2db.Tools.nuspec` → `exclude="Build"` on linq2db; consumers referencing only `linq2db` and only `linq2db.Tools` both get the rule; Release build green on all 11 affected projects; `Tests.Analyzers` 51/51.

`ReferenceOutputAssembly="false"` — confirmed, empties the dependency group entirely. Worth recording what the supported shape costs instead: the real reference puts the Roslyn-4.8-built code-fix assembly into consuming projects' reference closure, which gives `LinqToDB.EntityFrameworkCore.EF3` two `MSB3277` warnings on its `net462` / `netstandard2.0` legs (`System.Collections.Immutable` 7.0.0 vs the 1.7.0 EF Core 3.x pins). A `PrivateAssets` that also hides `compile`/`runtime` fixes the warning but breaks build ordering — clean build fails with CS0006 on the code-fix ref assembly. Taking the warning; EF3 never loads the assembly.

All four **Keep** items are kept: the `roslyn4.8` path, the toolchain-floor wording, `a7a65444c`, `2df081783`.

One correction: the readme paragraph was not mirrored to the wiki. The L2DB1001 page only carried the `Package` row from `a300430`, which is now reverted; the false transitive-flow claim shipped in the package readme only.

Follow-up outside this repo: `linq2db.EntityFrameworkCore`'s own `linq2db` `PackageReference` still uses default metadata, so that hop is the one left riding on the bypass.

MaceWindu:
`Correction:` the `project.assets.json` argument in my earlier comment does not hold.

I claimed the assets file showed NuGet applying the `analyzers` exclusion — no `analyzers` group on the analyzer-carrying library's target entry — while the SDK handed the assemblies to `csc` anyway, and used that to justify deviating from the plain `ProjectReference` you asked for towards `PrivateAssets="contentfiles;build"` at every hop.

`project.assets.json` has no `analyzers` group for **any** package: 0 occurrences across the 50 assets files in this repo, including `Microsoft.CodeAnalysis.BannedApiAnalyzers` and `Meziantou.Analyzer` (`[type, build]` only), which do deliver their rules. NuGet's lock-file `targets` schema has no such group at all — the SDK resolves analyzers from the package folder. So its absence indicates nothing about exclusions, and the lock file cannot distinguish the two layouts.

Direct A/B instead. SDK 10.0.302, all three packages built from this branch into a folder feed, each variant restored into its own isolated `globalPackagesFolder`, `.editorconfig` raising L2DB1001 to `error`:

| Variant | `linq2db` → `linq2db.Analyzers` | `linq2db.Tools` → `linq2db` | L2DB1001, `linq2db` only | L2DB1001, `Tools` only |
|---|---|---|---|---|
| this PR | `exclude="Build"` | `exclude="Build"` | reported | reported |
| packed nuspecs patched to the default | `exclude="Build,Analyzers"` | `exclude="Build,Analyzers"` | reported | reported |

`csc` receives both analyzer assemblies in every case. So the `PrivateAssets` metadata on the twelve edges has no observable effect on the current SDK, and your config D measurement was the whole picture — there is no contract-versus-bypass distinction to be had from measurement, only from reading the documented `exclude` semantics.

The metadata stays, because that reading is still the right side to be on: `exclude="Build"` is what NuGet documents as required for the subtree to flow, and it is what EF Core's published nuspecs carry. But it is contract conformance, not a fix for an observed delivery failure, and the `LinqToDB.csproj` comment said otherwise. Corrected in `62d1a8611`, together with a second wrong claim in the same paragraph — `master` delivers no rules because it has no reference to the analyzer project at all, not because of the default `PrivateAssets` value.

Two additions from the same pass:

- `ba7e79947` adds `Build/Azure/scripts/verify-analyzer-delivery.ps1` to the nugets job. Since the metadata's effect is invisible today, a satellite added later with default metadata would drop the contract-guaranteed path with nothing noticing. The check asserts it on the packed nuspecs instead of the mechanism: `linq2db` must declare the analyzer dependency in every framework group, no package may exclude `Analyzers` on its `linq2db` / `linq2db.Tools` / `linq2db.Analyzers` edge (`linq2db.cli` and `linq2db.LINQPad` skip-listed — nothing compiles against them), and the analyzer package must carry both assemblies under `analyzers/**/cs`.
- `bea5c8c9b` documents a cost that was measured but undocumented: leaving `runtime` public makes the code-fix assembly copy-local downstream, so each of the seven `linq2db.cli.<rid>` sub-packages now carries it under all three `tools/<tfm>/<rid>` payloads — 6 files per RID, a few hundred KB against packages that already run 64–164 MB, never loaded. No library package's `lib/` is affected, and `linq2db.LINQPad` is not either.

igor-tkachev:
Overall, the separate analyzer package approach looks reasonable. I have two implementation suggestions and one note regarding EF3.

### 1. Consider excluding the analyzer packaging reference from unrelated build graphs

The reference to `LinqToDB.Analyzers.CodeFixes` is used only to produce the NuGet dependency. `LinqToDB` itself does not consume analyzer or code-fix types at compile time or runtime.

It would be useful to check whether this packaging-only reference can be excluded from project graphs where the linq2db package is not being produced, especially `linq2db.cli`. This could prevent analyzer assemblies from participating in reference resolution and from being copied into CLI tool packages.

Several options could be tested.

The reference could be marked as not providing a compile-time assembly:

```xml
<ProjectReference
	Include="../LinqToDB.Analyzers.CodeFixes/LinqToDB.Analyzers.CodeFixes.csproj"
	ReferenceOutputAssembly="false"
	PrivateAssets="contentfiles;build" />
```

Please verify that `dotnet pack` still generates the required dependency on `linq2db.Analyzers`.

Another option is a conditional reference controlled by an MSBuild property:

```xml
<ProjectReference
	Condition="'$(IncludeLinqToDBAnalyzerPackage)' != 'false'"
	Include="../LinqToDB.Analyzers.CodeFixes/LinqToDB.Analyzers.CodeFixes.csproj"
	ReferenceOutputAssembly="false"
	PrivateAssets="contentfiles;build" />
```

A project that does not need this packaging reference could disable it:

```xml
<PropertyGroup>
	<IncludeLinqToDBAnalyzerPackage>false</IncludeLinqToDBAnalyzerPackage>
</PropertyGroup>
```

The property could also be passed explicitly through a project reference:

```xml
<ProjectReference
	Include="..\LinqToDB.Scaffold\LinqToDB.Scaffold.csproj"
	AdditionalProperties="IncludeLinqToDBAnalyzerPackage=false" />
```

This needs to be checked across the complete project graph, because the analyzer project could otherwise arrive through another linq2db project-reference path.

It is also worth checking whether the packaging reference can be enabled only during `Pack`, for example through an explicit property set by the package build:

```text
dotnet pack -p:IncludeLinqToDBAnalyzerPackage=true
```

The goal is not to require a specific implementation, but to see whether this packaging-only dependency can be kept out of unrelated source-build graphs without breaking NuGet dependency generation.

### 2. Provide an MSBuild opt-out for analyzers

Since analyzers will now be enabled automatically for linq2db consumers, please provide and document an MSBuild property that allows consumers to disable them.

For example:

```xml
<PropertyGroup>
	<EnableLinqToDBAnalyzers>false</EnableLinqToDBAnalyzers>
</PropertyGroup>
```

The exact implementation depends on the chosen analyzer delivery mechanism. The important requirement is that the opt-out is implemented by the published package and works for both direct and transitive consumption.

### 3. EF3

EF3 is legacy and should not drive the overall analyzer packaging design.

First, check whether one of the approaches from section 1 can keep the analyzer packaging dependency out of the EF3 build graph.

Otherwise, the EF3 package can use the opt-out mechanism from section 2 to disable analyzers for its consumers. If the `MSB3277` warning still remains, it can be suppressed by package-level MSBuild assets shipped with the EF3 package.

I would not add significant complexity to the primary implementation solely to preserve analyzer support for EF3.

### Testing expectations

The package verification should cover:

- direct consumption through `linq2db`;
- transitive analyzer delivery through representative satellite packages;
- opt-out through the documented MSBuild property;
- absence of analyzer and code-fix assemblies from application runtime output;
- absence of analyzer and code-fix assemblies from CLI RID packages;
- EF3 consumer builds with the selected package-level workaround;
- clean restore using an isolated NuGet cache.

MaceWindu:
Implemented in `2e85ad95b` (1 and 3), `8b78cf8eb` (2), `bffd1fe74` (packaging gate). Branch synced with master.

### 1. Packaging-only reference in unrelated build graphs

Real, and now fixed — but none of the reference-metadata options work. Measured on SDK 10.0.302:

- `ReferenceOutputAssembly="false"` drops the `linq2db.Analyzers` dependency from `linq2db.nuspec` entirely (empty `<group/>`), so the rules reach nobody. That applies to both snippets above.
- `Private="false"` removes the code-fix assembly from `bin/LinqToDB/<tfm>/` only. `LinqToDB.Scaffold` and `LinqToDB.CLI` still copy it, so the tool payload is unchanged — MSBuild's `Private` has no bearing on the downstream project-reference closure.
- `PrivateAssets` that also hides `compile` / `runtime` breaks P2P build ordering: clean build fails with CS0006 on the code-fix reference assembly.

The mechanism is the SDK's `IncludeTransitiveProjectReferences`, which injects `@(_TransitiveProjectReferences)` into `@(ProjectReference)` before `AssignProjectConfiguration` — that is what turns a packaging-only reference into a compile *and* copy-local item for every downstream project. `Source/Directory.Build.props` now removes only the injected item:

```xml
<Target Name="_DropTransitiveAnalyzerCodeFixesReference" AfterTargets="IncludeTransitiveProjectReferences">
	<ItemGroup>
		<ProjectReference Remove="@(_TransitiveProjectReferences->WithMetadataValue('Filename','LinqToDB.Analyzers.CodeFixes'))" />
	</ItemGroup>
</Target>
```

LinqToDB's own direct reference is untouched, so `dotnet pack` still emits the dependency — `verify-analyzer-delivery.ps1` passes on the packed nuspecs. No conditional reference and no pack-only property, so the dependency shape can't differ between a CI pack and a local one.

Result: no `LinqToDB.Analyzers*` in `LinqToDB.CLI` output (`dotnet linq2db help` runs), and every `Source/*` project builds on every TFM in Release.

### 2. MSBuild opt-out

`EnableLinqToDBAnalyzers=false`, documented in both readmes, implemented by `build/` + `buildTransitive/linq2db.Analyzers.targets` removing `@(Analyzer->WithMetadataValue('NuGetPackageId','linq2db.Analyzers'))`.

Worth recording why it is more than a convenience: it is currently the *only* opt-out that works. `ExcludeAssets="analyzers"` on a consumer's `linq2db` reference does not suppress the rules, because the SDK hands analyzers to csc even across a dependency edge that excludes them. (Same measurement is why the `PrivateAssets` metadata on this PR's dozen project references is contract conformance rather than an observed behaviour difference.)

Verified against packed packages on four consumers — `linq2db` and `linq2db.Tools`-only, each with and without the property. The rule fires by default on both reference paths and goes silent with the property on both, so the opt-out reaches a consumer two hops from the analyzer package: `exclude="Build"` on the intermediate edge does not block `buildTransitive`.

### 3. EF3

No EF3-specific handling needed — the `MSB3277` was a symptom of 1, not something EF3 has to work around. It came from RAR unifying the Roslyn-4.8 `System.Collections.Immutable` 7.0.0 in the reference closure against EF Core 3.x's 1.7.0. With the transitive reference gone, `LinqToDB.EntityFrameworkCore.EF3` builds with 0 warnings on `net462` and `netstandard2.0`.

On the two fallbacks: the opt-out property would not have cleared it, since `MSB3277` is reference resolution rather than analyzer loading, and package-level assets would not either, since the warning only ever occurred building EF3 in this repo. A consumer of the EF3 package never sees the code-fix assembly at all — `linq2db.Analyzers` ships `analyzers/` only, no `lib/`.

### Testing

Covered: direct consumption, transitive delivery through a satellite package, opt-out on both paths, absence of analyzer and code-fix assemblies from consumer runtime output, clean restore against an isolated `globalPackagesFolder`. CLI RID packages are covered structurally — the assembly is out of the build output that `PackAsTool` packs. `bffd1fe74` adds a check that both packed targets folders are present, so the documented property cannot silently stop working.

igor-tkachev (approved):

sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- The analyzer rules ship with the library: `linq2db` now depends on `linq2db.Analyzers`, so they arrive with a direct `linq2db` reference or through any satellite package (`linq2db.Tools`, `linq2db.Remote.*`, `linq2db.Extensions`, `linq2db.Compat`, `linq2db.Scaffold`, `linq2db.EntityFrameworkCore`) with no extra reference — on by default. Requires .NET SDK 8.0+ or Visual Studio 2022 17.8+; older hosts skip the rules silently. Severity is configurable per rule, or for every linq2db rule at once via `dotnet_analyzer_diagnostic.category-LinqToDB.severity`, and `<EnableLinqToDBAnalyzers>false</EnableLinqToDBAnalyzers>` disables them entirely. Reference `linq2db.Analyzers` directly only to run the migration rule against an older linq2db (6.1–6.3) when planning an upgrade. ([#5720](https://github.com/linq2db/linq2db/pull/5720))


_(none)_


---
<sub>Generated from commit `c76ee61`.</sub>

Issue #2874, opened 2021-03-05
Support composite properties in select queries


MaceWindu:
https://rsdn.org/forum/prj.rfd/7965340.1

```cs
   [Table(Schema = "dbo", Name = "Test")]
    [Column("Coord.X", MemberName = "Coord.X", CanBeNull = false)]
    [Column("Coord.Y", MemberName = "Coord.Y", CanBeNull = false)]
    [Column("Coord.Z.Z1", MemberName = "Coord.Z.Z1", CanBeNull = false)]
    [Column("Coord.Z.Z2", MemberName = "Coord.Z.Z2", CanBeNull = false)]
    public partial class Test
    {
        [Column, PrimaryKey, Identity] 
        public long ID;// { get; set; } 
        public Coord Coord;// { get; set; }
    }
```

Recognize composite property in select queries, e.g.
```cs
db.Tests.Select(t => t.Coord).Distinct();
```

MaceWindu:
Implemented in `6.0.0-preview.1`

MaceWindu:
Note that in preview 1 it has issue with selecting also parent columns. Will be fixed in next release

MaceWindu:
My bad, stuill not fixed

MaceWindu:
Actually tests show it is fixed in v6, but I get strage test results from 5.4.1 where same query pass in tests and fail in standalone app.

PR #5724, opened 2026-07-22
Tests: composite property projection coverage (#4568)

### LINKED ISSUE #4568: Don't select all entity columns when only composite property selected
6.0 preview.1 fixed issue with exception on composite property selection (https://github.com/linq2db/linq2db/issues/2874) but it currently selects also main entity columns.

sdanyliv:
Tests only, no product code changes.

## Why

#4568 reports that selecting a composite (complex-mapped) property in a final projection emits all entity columns. It does not reproduce.

The behaviour was fixed by b7de32bf46 (#4676, "Refactored SQL generation"), which also added `SelectCompositePropertyMapped_Class_OnlyRequiredColumns` / `_Struct_OnlyRequiredColumns` - ungated, and green. First release containing the fix is `v6.0.0-preview.2`; re-verified on the released `v6.3.0` tag, where the same variants also produce minimal column lists. So the issue is stale rather than open.

Confirmed by bisect across that commit - same worktree, same toolchain, same provider. At these commits `User` has no `Id`, so its full column set is exactly 4 (`user_name`, `city`, `street`, `building_number`):

| Commit | `Select(u => u.Residence)` | `+ .Distinct()` | `new { u.Residence }` |
|---|---|---|---|
| `b7de32bf46^` (f9000ca39d, 2024-12-13) | 4 (all columns) | 4 | 4 |
| `b7de32bf46` (2024-12-20) | 3 | 3 | 3 |

Both sides were run with sentinel expected values, so each figure is a reported actual rather than an inferred pass.

## What

The two existing tests only cover the bare `Select(u => u.Residence)` shape. This adds the untested variants:

| Test | Query | Expected columns |
|---|---|---|
| `SelectCompositePropertyMapped_Class_Distinct_OnlyRequiredColumns` | `Select(u => u.Residence).Distinct()` | 3 |
| `SelectCompositePropertyMapped_Class_Projected_OnlyRequiredColumns` | `Select(u => new { u.Residence })` | 3 |
| `SelectNestedCompositeProperty_OnlyRequiredColumns` | `Select(t => t.Coord)` / `Select(t => t.Coord!.Z)` | 4 / 2 |
| `SelectNestedCompositeProperty` | materialization round-trip | n/a |

The nested cases add a `NestedCoordEntity` model mirroring the two-level `Coord.X` / `Coord.Y` / `Coord.Z.Z1` / `Coord.Z.Z2` mapping from #2874's original example, which had no coverage at all. `SelectNestedCompositeProperty` round-trips values through `CreateLocalTable`, since a column-count assertion alone can pass while returning wrong data.

`Distinct()` is the variant worth pinning: extra columns there change result semantics, not just payload size.

All six #4568 tests - the four added here plus the two pre-existing `_OnlyRequiredColumns` ones - are pinned to `[IncludeDataSources(true, TestProvName.AllSQLite)]`. The assertions are provider-agnostic (SQL AST column counts, plus one materialization round-trip), so running them across the full matrix cost CI time without adding signal. Switching the two pre-existing tests off `[DataSources]` is the only change here to code the PR didn't introduce.

## Verification

Assertions confirmed non-vacuous by re-running them with sentinel expected counts and reading back the actual values - 3 / 3 / 4 / 2 on both `master` and `v6.3.0`.

Run green on SQLite.Classic and SQLite.MS, including the LinqService remote variant of each. The SQLite pin was verified to actually bind rather than merely pass: adding `SqlServer.2019.MS` to the provider set takes the run from 32 tests to 20, i.e. the six restricted tests decline it while the two unrestricted #2874 tests pick it up.

Before the tests were pinned to SQLite they also ran green on SqlServer.2019.MS, Oracle.11.Managed, Firebird.2.5, DB2 and Informix.DB2, so the expected column counts are not SQLite-specific.

Closes #4568

### DIFF
--- a/Tests/Linq/Linq/ComplexTests.cs
+++ b/Tests/Linq/Linq/ComplexTests.cs
@@ -704,7 +704,7 @@ namespace Tests.Linq
 		}
 
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/4568")]
-		public void SelectCompositePropertyMapped_Class_OnlyRequiredColumns([DataSources] string context)
+		public void SelectCompositePropertyMapped_Class_OnlyRequiredColumns([IncludeDataSources(true, TestProvName.AllSQLite)] string context)
 		{
 			using var db    = GetDataContext(context);
 
@@ -714,7 +714,7 @@ namespace Tests.Linq
 		}
 
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/4568")]
-		public void SelectCompositePropertyMapped_Struct_OnlyRequiredColumns([DataSources] string context)
+		public void SelectCompositePropertyMapped_Struct_OnlyRequiredColumns([IncludeDataSources(true, TestProvName.AllSQLite)] string context)
 		{
 			var ms = new MappingSchema();
 			ms.SetScalarType(typeof(AddressStruct), false);
@@ -727,7 +727,7 @@ namespace Tests.Linq
 		}
 
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/4568")]
-		public void SelectCompositePropertyMapped_Class_Distinct_OnlyRequiredColumns([DataSources] string context)
+		public void SelectCompositePropertyMapped_Class_Distinct_OnlyRequiredColumns([IncludeDataSources(true, TestProvName.AllSQLite)] string context)
 		{
 			using var db = GetDataContext(context);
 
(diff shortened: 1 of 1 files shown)


MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `094422a`.</sub>

PR #5726, opened 2026-07-23
Tests: fix progress heartbeat over-counting expected-throw cases

sdanyliv:
## Problem

The `--test-progress` heartbeat over-reports failures: every `[ThrowsForProvider]` (and other `ThrowsWhen`-family) case where the provider-capability exception fires is counted as a failure, even though the run's verdict is a pass. On a 5-provider sweep this showed as `failed 466` in the heartbeat vs `failed 0` in the MTP summary.

## Cause

`ThrowsWhenCommand` (an `IWrapSetUpTearDown` wrapper) runs the inner test, then rewrites the expected throw `Failed -> Success`. The `TestProgressReporter` heartbeat is an `ITestAction` that fires *inside* that wrapper, so its `AfterTest` samples the result before the rewrite and counts it `Failed` (with the raw exception in `recentFailures`).

## Fix

The unit is **deferred** rather than counted-then-corrected. `ThrowsWhenCommand.Execute` brackets the inner command with `TestProgressTracker.BeginDeferred(fullName)` / `CommitDeferred(fullName, result)`; while a deferral is in flight the reporter action only *samples* the outcome, and the wrapper books the unit once, with its final verdict.

No provisional result is ever counted, added to `recentFailures`, or published. The bounded failure list therefore only ever holds real failures — a provisional entry can no longer occupy the last slot and cause a concurrent real failure to be dropped — and there is exactly one publish per unit. An expected throw that fires books `Passed`, which is not a forced write at all, so the common case costs *fewer* writes than before.

Details worth knowing:

- **Deferrals nest by depth, and only the outermost wrapper books.** Several `ThrowsWhen`-family attributes on one method are common — `[ThrowsRequiresCorrelatedSubquery]` + `[ThrowsForProvider(…)]` pairs across `Tests/Linq/Update/` and `Tests/Linq/Linq/` — and their wrappers nest. The inner one typically has `expectsException == false` for the running provider and returns the result untouched while the outer one rewrites it, so the outermost level is the one that sees the final verdict.
- **No dependence on NUnit's command ordering.** If the reporter action ever ends up *outside* the wrapper, `CompleteTest` finds no deferral in flight, books immediately, and `CommitDeferred` no-ops.
- **Nothing drops out of the tally.** A unit still held back at run end (an exception escaping between the sample and the commit) is flushed by the root-suite teardown.

The accounting itself moves out of the static holder into `TestProgressState` (`Tests/Base/TestProgressState.cs`): the counters, the bounded recent-failure list and the deferral protocol, with no file IO, no statics and no command-line dependency, so it can be exercised directly. `TestProgressTracker` holds one instance and keeps `Write` / `EnsureInit`.

## Verification

`Tests/Linq/Infrastructure/TestProgressStateTests.cs` — 10 tests covering both expected-throw directions, the wrong-exception (`Failed -> Failed`) message rewrite, the `MaxRecentFailures` boundary with a provisional in flight, nested wrappers, the reporter-outside-wrapper ordering, the run-end straggler flush, message truncation, the done-latch, and a 500-unit `Parallel.For` consistency check.

End-to-end, same slice (`DistinctTests` against `SQLite.Classic`, `--test-progress`); MTP reports `failed: 0` in both runs:

- pre-fix build — heartbeat `"passed":136,"failed":2` plus two phantom `recentFailures` entries
- with the fix — `"passed":176,"failed":0,"recentFailures":[]` (176-case run, `DistinctTests` + `ElementOperationTests`)

Test-infra only: `TestProgressState` lives in the `linq2db.Tests.Base` test-support assembly, which is not packed — no public-API or baseline impact.

### DIFF
--- a/Tests/Base/TestProgressReporter.cs
+++ b/Tests/Base/TestProgressReporter.cs
@@ -147,6 +147,55 @@ namespace Tests
 			}
 		}
 
+		/// <summary>
+		/// Corrects the running tally when a result the heartbeat already sampled is subsequently rewritten by an
+		/// outer command. The <see cref="TestProgressReporterAttribute"/> action runs <em>inside</em> the
+		/// <c>IWrapSetUpTearDown</c> layer, so for an expected-throw case (<see cref="ThrowsWhenAttribute"/>) it
+		/// counts the pre-rewrite (Failed) status; <c>ThrowsWhenCommand</c> then rewrites the result to Success and
+		/// calls this to move the unit between buckets (and prune the now-stale <c>recentFailures</c> entry). Handles
+		/// the symmetric case too — an expected throw that never fired is counted Passed here, then reclassified Failed.
+		/// A no-op when the status is unchanged or the heartbeat is disabled.
+		/// </summary>
+		internal static void Reclassify(TestStatus before, TestStatus after, string? test, string? message)
+		{
+			if (before == after)
+				return;
+
+			lock (_sync)
+			{
+				if (!_enabled)
+					return;
+
+				switch (before)
+				{
+					case TestStatus.Passed : if (_passed  > 0) _passed--;  break;
+					case TestStatus.Skipped: if (_skipped > 0) _skipped--; break;
+					case TestStatus.Failed :
+						if (_failed > 0)
+							_failed--;
+						if (test != null)
+							_failures.RemoveAll(f => f.Test == test);
+						break;
+					// Inconclusive / Warning were never counted into a pass/fail bucket — nothing to undo.
+					default: break;
+				}
+
+				switch (after)
+				{
+					case TestStatus.Passed : _passed++;  break;
+					case TestStatus.Skipped: _skipped++; break;
+					case TestStatus.Failed :
+						_failed++;
+						if (test != null && _failures.Count < MaxRecentFailures)
+							_failures.Add((test, Trim(message)));
+						break;
+					default: break;
+				}
+
+				Write(force: true);
+			}
+		}
+
 		static bool EnsureInit()
 		{
 			if (_initialized)
--- a/Tests/Base/TestProgressReporter.cs
+++ b/Tests/Base/TestProgressReporter.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
@@ -47,26 +46,17 @@ namespace Tests
 		const string Tfm = "unknown";
 #endif
 
-		const int MaxRecentFailures = 20;
-		const int WriteThrottleMs   = 1000;  // at most ~1 file write/sec for the common (passing) case
+		const int WriteThrottleMs = 1000;  // at most ~1 file write/sec for the common (passing) case
 
-		static readonly Lock                                _sync     = new();
-		static readonly Stopwatch                           _clock    = new();
-		static readonly List<(string Test, string Message)> _failures = new();
+		static readonly Lock              _sync  = new();
+		static readonly Stopwatch         _clock = new();
+		static readonly TestProgressState _state = new(Write);
 
 		static bool    _initialized;
 		static bool    _enabled;
 		static string? _file;
 		static int     _pid;
 
-		static long     _total;
-		static long     _started;
-		static long     _completed;
-		static long     _passed;
-		static long     _failed;
-		static long     _skipped;
-		static string?  _current;
-		static bool     _done;
 		static DateTime _startedUtc;
 		static long     _lastWriteMs = -WriteThrottleMs;
 
@@ -81,14 +71,11 @@ namespace Tests
 				{
 					// The assembly (root) suite has the largest case count → that is the run total.
 					// Note: under a --filter this is the discovered (pre-filter) count, so it over-counts.
-					if (test.TestCaseCount > _total)
-						_total = test.TestCaseCount;
+					_state.SetTotal(test.TestCaseCount);
 					return;
 				}
 
-				_started++;
-				_current = test.FullName;
-				Write(force: false);
+				_state.StartTest(test.FullName);
 			}
 		}
 
@@ -102,97 +89,48 @@ namespace Tests
 				if (test.IsSuite)
 				{
 					// Suite teardown fires bottom-up; the root suite (no parent) finishing means the run is done.
-					if (!_done && test.Parent == null)
-					{
-						_done    = true;
-						_current = null;
-						Write(force: true);
-					}
+					if (test.Parent == null)
+						_state.MarkDone();
 
 					return;
 				}
 
-				_completed++;
-				// Deliberately do NOT clear _current here: with throttled writes, nulling between every test
-				// makes the on-disk snapshot almost always catch the gap (showing no current test). Keep the
-				// most-recently-started test as "current" until the run completes — during a run that is the
-				// in-flight (or just-finished) test, which is what a watcher wants to see.
+				var result = TestContext.CurrentContext.Result;
 
-				var force = false;
-
-				switch (TestContext.CurrentContext.Result.Outcome.Status)
-				{
-					case TestStatus.Passed : _passed++;  break;
-					case TestStatus.Skipped: _skipped++; break;
-					case TestStatus.Failed :
-						_failed++;
-						force = true;
-						if (_failures.Count < MaxRecentFailures)
-							_failures.Add((test.FullName, Trim(TestContext.CurrentContext.Result.Message)));
-						break;
-					// Inconclusive / Warning — count as completed but neither pass nor fail bucket.
-					default: break;
-				}
-
-				// For unfiltered runs total is exact, so completed reaching it marks the run done even if the
-				// root-suite teardown ordering surprises us.
-				if (!_done && _total > 0 && _completed >= _total)
-				{
-					_done    = true;
-					_current = null;
-					force    = true;
-				}
-
-				Write(force);
+				_state.CompleteTest(test.FullName, result.Outcome.Status, result.Message);
 			}
 		}
 
 		/// <summary>
-		/// Corrects the running tally when a result the heartbeat already sampled is subsequently rewritten by an
-		/// outer command. The <see cref="TestProgressReporterAttribute"/> action runs <em>inside</em> the
-		/// <c>IWrapSetUpTearDown</c> layer, so for an expected-throw case (<see cref="ThrowsWhenAttribute"/>) it
-		/// counts the pre-rewrite (Failed) status; <c>ThrowsWhenCommand</c> then rewrites the result to Success and
-		/// calls this to move the unit between buckets (and prune the now-stale <c>recentFailures</c> entry). Handles
-		/// the symmetric case too — an expected throw that never fired is counted Passed here, then reclassified Failed.
-		/// A no-op when the status is unchanged or the heartbeat is disabled.
+		/// Opens a deferral for a test whose result an outer <c>IWrapSetUpTearDown</c> wrapper is going to rewrite
+		/// (<see cref="ThrowsWhenAttribute"/>). NUnit nests this reporter's action <em>inside</em> such a wrapper, so
+		/// the outcome it samples is the pre-rewrite one; deferring holds the unit back until
+		/// <see cref="CommitDeferred"/> supplies the final verdict, so no provisional outcome is ever counted, added
+		/// to <c>recentFailures</c>, or published. Protocol details: <see cref="TestProgressState"/>.
 		/// </summary>
-		internal static void Reclassify(TestStatus before, TestStatus after, string? test, string? message)
+		public static void BeginDeferred(string fullName)
 		{
-			if (before == after)
-				return;
-
 			lock (_sync)
 			{
-				if (!_enabled)
+				if (!EnsureInit())
 					return;
 
-				switch (before)
-				{
-					case TestStatus.Passed : if (_passed  > 0) _passed--;  break;
-					case TestStatus.Skipped: if (_skipped > 0) _skipped--; break;
-					case TestStatus.Failed :
-						if (_failed > 0)
-							_failed--;
-						if (test != null)
-							_failures.RemoveAll(f => f.Test == test);
-						break;
-					// Inconclusive / Warning were never counted into a pass/fail bucket — nothing to undo.
-					default: break;
-				}
+				_state.BeginDeferred(fullName);
+			}
+		}
 
-				switch (after)
-				{
-					case TestStatus.Passed : _passed++;  break;
-					case TestStatus.Skipped: _skipped++; break;
-					case TestStatus.Failed :
-						_failed++;
-						if (test != null && _failures.Count < MaxRecentFailures)
-							_failures.Add((test, Trim(message)));
-						break;
-					default: break;
-				}
+		/// <summary>
+		/// Closes a deferral opened by <see cref="BeginDeferred"/>, booking the unit with the wrapper's final verdict.
+		/// <paramref name="result"/> is <see langword="null"/> when the wrapper failed before producing one.
+		/// </summary>
+		public static void CommitDeferred(string fullName, ITestResult? result)
+		{
+			lock (_sync)
+			{
+				if (!EnsureInit())
+					return;
 
-				Write(force: true);
+				_state.CommitDeferred(fullName, result?.ResultState.Status ?? TestStatus.Failed, result?.Message);
 			}
 		}
 
@@ -306,9 +244,14 @@ namespace Tests
 
 			_lastWriteMs = nowMs;
 
+			// Read under the state's lock (Write is invoked from inside it), so the snapshot is coherent.
+			var total      = _state.Total;
+			var completed  = _state.Completed;
+			var failures   = _state.RecentFailures;
+
 			var elapsedSec = _clock.Elapsed.TotalSeconds;
-			var rate       = elapsedSec > 0 ? _completed / elapsedSec : 0d;
-			var eta        = rate > 0 && _total > _completed ? (_total - _completed) / rate : (double?)null;
+			var rate       = elapsedSec > 0 ? completed / elapsedSec : 0d;
+			var eta        = rate > 0 && total > completed ? (total - completed) / rate : (double?)null;
 
 			var sb = new StringBuilder(512);
 
@@ -317,26 +260,26 @@ namespace Tests
 			sb.Append("\"pid\":")        .Append(Int(_pid))                             .Append(',');
 			sb.Append("\"startedUtc\":") .Append(JsonString(Iso(_startedUtc)))          .Append(',');
 			sb.Append("\"updatedUtc\":") .Append(JsonString(Iso(DateTime.UtcNow)))      .Append(',');
-			sb.Append("\"done\":")       .Append(_done ? "true" : "false")              .Append(',');
-			sb.Append("\"total\":")      .Append(Long(_total))                          .Append(',');
-			sb.Append("\"completed\":")  .Append(Long(_completed))                      .Append(',');
-			sb.Append("\"started\":")    .Append(Long(_started))                        .Append(',');
-			sb.Append("\"passed\":")     .Append(Long(_passed))                         .Append(',');
-			sb.Append("\"failed\":")     .Append(Long(_failed))                         .Append(',');
-			sb.Append("\"skipped\":")    .Append(Long(_skipped))                        .Append(',');
-			sb.Append("\"currentTest\":").Append(JsonString(_current))                  .Append(',');
+			sb.Append("\"done\":")       .Append(_state.Done ? "true" : "false")        .Append(',');
+			sb.Append("\"total\":")      .Append(Long(total))                           .Append(',');
+			sb.Append("\"completed\":")  .Append(Long(completed))                       .Append(',');
+			sb.Append("\"started\":")    .Append(Long(_state.Started))                  .Append(',');
+			sb.Append("\"passed\":")     .Append(Long(_state.Passed))                   .Append(',');
+			sb.Append("\"failed\":")     .Append(Long(_state.Failed))                   .Append(',');
+			sb.Append("\"skipped\":")    .Append(Long(_state.Skipped))                  .Append(',');
+			sb.Append("\"currentTest\":").Append(JsonString(_state.Current))            .Append(',');
 			sb.Append("\"elapsedSec\":") .Append(Num(elapsedSec))                       .Append(',');
 			sb.Append("\"testsPerSec\":").Append(Num(rate))                             .Append(',');
 			sb.Append("\"etaSec\":")     .Append(eta.HasValue ? Num(eta.Value) : "null").Append(',');
 			sb.Append("\"recentFailures\":[");
 
-			for (var i = 0; i < _failures.Count; i++)
+			for (var i = 0; i < failures.Count; i++)
 			{
 				if (i > 0)
 					sb.Append(',');
 
-				sb.Append("{\"test\":").Append(JsonString(_failures[i].Test))
-				  .Append(",\"message\":").Append(JsonString(_failures[i].Message)).Append('}');
+				sb.Append("{\"test\":").Append(JsonString(failures[i].Test))
+				  .Append(",\"message\":").Append(JsonString(failures[i].Message)).Append('}');
 			}
 
 			sb.Append("]}");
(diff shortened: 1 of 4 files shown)


Tests/Base/TestProgressReporter.cs
```diff
@@ -147,6 +147,55 @@ public static void OnAfter(ITest test)
...
+				{
+					case TestStatus.Passed : if (_passed  > 0) _passed--;  break;
+					case TestStatus.Skipped: if (_skipped > 0) _skipped--; break;
+					case TestStatus.Failed :
+						if (_failed > 0)
+							_failed--;
+						if (test != null)
+							_failures.RemoveAll(f => f.Test == test);
+						break;
+					// Inconclusive / Warning were never counted into a pass/fail bucket — nothing to undo.
```
igor-tkachev: This cannot reliably restore `recentFailures` under parallel execution. A provisional expected failure can occupy the last slot in the bounded list. While it is present, a concurrent real failure is skipped because `_failures.Count == MaxRecentFailures`; when this entry is removed later, the real failure is not backfilled. The counters become correct, but `recentFailures` can be incomplete despite having free capacity. We should avoid inserting provisional failures into the bounded final-failure list, or retain enough information to rebuild it correctly.

Tests/Base/TestProgressReporter.cs
```diff
@@ -147,6 +147,55 @@ public static void OnAfter(ITest test)
...
+						_failed++;
+						if (test != null && _failures.Count < MaxRecentFailures)
+							_failures.Add((test, Trim(message)));
+						break;
+					default: break;
+				}
+
+				Write(force: true);
+			}
+		}
```
igor-tkachev, line 135: This produces a second forced file write for every expected-throw case. `OnAfter()` has already forced a write for the provisional `Failed` result, then this immediately forces another write for the corrected result. With hundreds of such tests this bypasses `WriteThrottleMs`, creates substantial synchronous I/O, and still allows an external watcher to observe the false failure between the two writes. The provisional result should ideally not be published at all.

igor-tkachev (changes requested):
The goal of this PR is valid: the heartbeat should report the final NUnit verdict rather than the intermediate result observed inside `ThrowsWhenCommand`. However, the current compensation approach introduces correctness and performance issues of its own.

The main problem is that the reporter first records and publishes a result that is known to be provisional, and then `ThrowsWhenCommand` tries to undo the side effects. That is not equivalent to accounting for the final result directly:

1. `recentFailures` cannot be repaired reliably under parallel execution. A provisional expected failure can occupy the last available slot, causing a concurrent real failure to be discarded due to `MaxRecentFailures`. When the provisional entry is later removed, the real failure cannot be restored, leaving free capacity but an incomplete failure list.

2. Every successfully expected exception causes two forced heartbeat writes: one for the provisional failure and another for the corrected success. This bypasses the existing write throttle and still exposes the invalid intermediate state to external readers.

Please change the design so that provisional results are not committed/published as final accounting, or otherwise make the accounting transactional enough that counters and the bounded failure list remain correct. Regression coverage should include both expected-throw directions and a parallel scenario involving the `MaxRecentFailures` boundary.

Tests/Base/TestProgressReporter.cs
```diff
@@ -147,6 +147,55 @@ public static void OnAfter(ITest test)
...
+		/// <c>IWrapSetUpTearDown</c> layer, so for an expected-throw case (<see cref="ThrowsWhenAttribute"/>) it
+		/// counts the pre-rewrite (Failed) status; <c>ThrowsWhenCommand</c> then rewrites the result to Success and
+		/// calls this to move the unit between buckets (and prune the now-stale <c>recentFailures</c> entry). Handles
+		/// the symmetric case too — an expected throw that never fired is counted Passed here, then reclassified Failed.
+		/// A no-op when the status is unchanged or the heartbeat is disabled.
+		/// </summary>
+		internal static void Reclassify(TestStatus before, TestStatus after, string? test, string? message)
+		{
+			if (before == after)
+				return;
```
MaceWindu: **Minor · MIN001** — the `before == after` short-circuit reconciles the *status* but silently skips the *message*, and the one case where it fires is a real failure mode.

When an expected-throw case throws the **wrong** exception type — or the right one with a non-matching `ErrorMessage` — `ThrowsWhenCommand` rewrites the result Failed → Failed with a new message (`Expected a <X> to be thrown, but found: '<original>'`). The status is unchanged, so `Reclassify` returns immediately and the stale pre-rewrite entry stays in `recentFailures` carrying only the raw exception text. A watcher then sees an ordinary exception with no indication that what actually failed is a stale `ThrowsForProvider` expectation — precisely the context the finalized message adds.

Failed → Failed is the only reachable equal-status case: a passing inner result has a null message and always takes the `SetResult(Failure, …)` branch, and Skipped / Inconclusive always end at Failure or Success.

Fix: let the equal-status case fall through when the final verdict is `Failed`, so the `before` arm prunes the stale entry and the `after` arm re-adds it with the finalized message; counters net out unchanged. This costs one extra forced write per wrong-exception case, but those are genuine failures and therefore rare, so it does not materially add to the throttling concern already raised on this PR. An alternative that avoids the extra write entirely is to keep the early return but update the matching `recentFailures` entry's message in place before returning. Either way, whichever redesign the throttling feedback leads to should carry this case explicitly, or it will be missed again.

```suggestion
			if (before == after && after != TestStatus.Failed)
				return;
```

Tests/Base/TestProgressReporter.cs
```diff
@@ -147,6 +147,55 @@ public static void OnAfter(ITest test)
...
+
+			lock (_sync)
+			{
+				if (!_enabled)
+					return;
```
MaceWindu: **Suggestion · SUG001** — `OnBefore` and `OnAfter` both gate on `EnsureInit()`; `Reclassify` is the only entry point that reads the `_enabled` field directly.

Today that is safe — the reporter declares `ActionTargets.Test | ActionTargets.Suite`, so `OnBefore` runs for the assembly suite at run start and again for the test itself before the wrapper can reach `Reclassify`, meaning `_initialized` is always latched by then. But the invariant is implicit and lives in a different file from the code that relies on it; a future caller not nested inside the reporter's per-test action would turn this into a silent no-op that quietly drops a correction rather than failing visibly.

Fix: gate on `EnsureInit()` like the other two entry points. It returns `_enabled`, is idempotent, and is already only ever called under `_sync` — which `Reclassify` holds at this point — so the change is behaviour-preserving today and self-initialising if the call order ever changes.

```suggestion
			if (!EnsureInit())
				return;
```

Tests/Base/TestProgressReporter.cs
```diff
@@ -147,6 +147,55 @@ public static void OnAfter(ITest test)
...
+
+				switch (before)
+				{
+					case TestStatus.Passed : if (_passed  > 0) _passed--;  break;
+					case TestStatus.Skipped: if (_skipped > 0) _skipped--; break;
+					case TestStatus.Failed :
+						if (_failed > 0)
+							_failed--;
+						if (test != null)
+							_failures.RemoveAll(f => f.Test == test);
```
MaceWindu: **Suggestion · SUG002** — the correction is for exactly one unit of accounting, but `RemoveAll` deletes *every* entry whose test name matches.

The list is keyed by `ITest.FullName` only, and a single full name can legitimately appear more than once: outer repeat/retry wrappers sit at the outermost position of NUnit's command chain, so they re-enter the reporter's per-test action on every iteration and each failing iteration appends its own entry under the same name (`Tests/Linq/Mapping/MappingSchemaTests.cs:388`, `Tests/Linq/Extensions/QueryHintsTests.cs:58` and `Tests/Linq/UserTests/Issue548Tests.cs:55` all use `[Repeat(100)]`). No test currently combines a repeat wrapper with an expected-throw attribute, so this is latent rather than live — but the pruning is broader than the correction it compensates for, and narrowing it costs one line.

Fix: remove only the most recently added matching entry, which is the one this test's own provisional record created. `List<T>.FindLastIndex` is available on every target framework here.

```suggestion
					if (test != null)
					{
						var idx = _failures.FindLastIndex(f => f.Test == test);

						if (idx >= 0)
							_failures.RemoveAt(idx);
					}
```

Tests/Base/TestProgressReporter.cs
MaceWindu, line 1: **Major · MAJ001** — `Reclassify` is new accounting logic with five distinct paths (Failed → Passed, Passed → Failed, Skipped → Failed, Inconclusive → Failed, and the `before == after` short-circuit) plus a bounded-list mutation, and the PR adds no test for any of them.

It is not merely untested — it is currently **untestable**:

- `TestProgressTracker` is `internal` in `Tests/Base/Tests.Base.csproj` (assembly `linq2db.Tests.Base`);
- that project opts out of being a test container (`<ProjectCapability Remove="TestContainer" />`, `Tests/Base/Tests.Base.csproj:9`), so no test can be discovered inside it;
- there is no `InternalsVisibleTo` anywhere under `Tests/`, so nothing in `linq2db.Tests` can reach the type;
- every counter is `static`, `_initialized` latches on first use, and `_enabled` derives from the process command line — so even with visibility, a test could neither enable nor reset the tracker without corrupting the heartbeat of the very run executing it.

The blast radius is not small: `ThrowsForProviderAttribute` derives from `ThrowsWhenAttribute` and is applied 515 times across 83 test files, so this path executes on a large fraction of a full run.

Fix: extract the pure accounting out of the static holder so it can be exercised directly. Minimal seam — move the counters, the bounded `_failures` list, `MaxRecentFailures`, the `OnAfter` bucket switch and `Reclassify` into a small `public sealed class TestProgressState` in `Tests/Base` with no file IO and no command-line dependency; have `TestProgressTracker` hold one static instance and delegate to it, keeping `Write` / `EnsureInit` where they are. A fixture in `Tests/Linq` can then construct its own instance and cover:

1. Failed → Passed removes the provisional entry and moves the unit to `passed`;
2. Passed → Failed adds an entry and moves the unit to `failed`;
3. the `MaxRecentFailures` boundary — a provisional failure occupying the last slot while a real failure arrives, then reclassifying, asserting the resulting list contents;
4. concurrent record / reclassify from multiple threads leaving `passed + failed + skipped` consistent with the number of recorded units.

Cases 3 and 4 are the coverage asked for in the open review. That refactor is also the natural vehicle for making the accounting transactional, since the state becomes an object you can reason about and assert on in isolation. If a seam is genuinely unwanted, the fallback is `[assembly: InternalsVisibleTo("linq2db.Tests")]` on `Tests.Base` — but the extracted instance class is preferable, because it also removes the process-global-state hazard.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes

- [x] No baseline changes — CI produced no `baselines/pr_5726` branch, consistent with the test-infra-only scope.
- [x] No public API surface change — `Reclassify` is `internal` to the test-support assembly.
- [x] CI green at `9f2faec` (`build`, `test-sqlite`).

## Prior-review audit

- **Still actual** · igor-tkachev — provisional expected-failure entries can occupy the last `MaxRecentFailures` slot, so a concurrent real failure is discarded permanently. Confirmed at HEAD: `OnAfter` (`Tests/Base/TestProgressReporter.cs:130`) adds only while `_failures.Count < MaxRecentFailures`, and `Reclassify`'s `RemoveAll` frees the slot without backfilling. Two refinements: the loss is not parallelism-specific — single-threaded, a provisional entry in slot 20 also drops the next real failure arriving before the correction lands; and the impact is bounded, because the list is effectively "first 20", not "recent 20" — once 20 real failures are banked, everything later is dropped anyway.
- **Still actual** · igor-tkachev — every successfully-expected exception costs two forced writes, bypassing `WriteThrottleMs`. Confirmed at HEAD: `OnAfter` sets `force = true` on `Failed` (`Tests/Base/TestProgressReporter.cs:129`) and `Reclassify` ends with `Write(force: true)`. The volume is larger than "hundreds": `ThrowsForProviderAttribute` derives from `ThrowsWhenAttribute` and is applied 515 times across 83 test files, each fanning out per `[DataSources]` provider case, so a full multi-provider run pays thousands of extra unthrottled write + `File.Replace` cycles. The 466 phantom failures reported in the PR description are a lower bound on that.
- **Still actual** · igor-tkachev — regression coverage should include both expected-throw directions plus a parallel `MaxRecentFailures` scenario. No test is added by this PR. Carried into the findings below as **MAJ001**, which also covers why such a test cannot be written today without a new seam.
- **No verifiable claims** · copilot-pull-request-reviewer[bot] — summary-only review; its description of `Reclassify` and the `ThrowsWhenCommand` call site matches HEAD.

### Suggestion (SUG)

- [ ] **SUG003** — Concrete shape for the deferred-accounting redesign

  Why: the redesign requested in the open review — do not commit/publish provisional results as final accounting — is implementable with no new plumbing, because `ThrowsWhenAttribute.ApplyToTest` already stamps the test (`test.Properties.Add("ThrowsWhen", this)`) and `TestProgressTracker.OnAfter(ITest test)` receives that same `ITest`. Offered as implementation support for the change already asked for, not as a separate objection.

  Fix:
  - In `OnAfter`, when `test.Properties.ContainsKey("ThrowsWhen")`, increment `_completed` and update `_current` as today, but skip the pass/fail/skip bucket, skip the `_failures` append, and leave `force` false. Track the unit as pending (a `HashSet<string>` of full names under `_sync`).
  - Replace `Reclassify(before, after, …)` with a commit entry point taking only the final `(status, message)` — no `before`. It performs one bucket increment, at most one `_failures` add, and one write.
  - Call it unconditionally at the end of `ThrowsWhenCommand.Execute` — **not** only inside `if (expectsException)` — since `OnAfter` deferred regardless of whether a rewrite happened.
  - `AllowMultiple = true` on `ThrowsWhenAttribute` means a test can carry several attributes and therefore several nested `ThrowsWhenCommand`s: make the commit idempotent (clear the pending marker on first commit).
  - Flush any still-pending unit at run end (root-suite `OnAfter`) so an exception escaping the wrapper cannot silently drop a unit from the tally.

  This addresses both objections in the open review at once — no provisional entry ever enters the bounded list, and there is exactly one forced write per expected-throw case. It also drops the `before` parameter, so the accounting stops depending on the caller having correctly guessed what the heartbeat recorded. That guess is currently correct — in NUnit 4.6 `SimpleWorkItem.MakeTestCommand` applies upstream (fixture + assembly) actions in reverse order, so the assembly-level reporter ends up the outermost of them and nothing runs between its `AfterTest` and `ThrowsWhenCommand` regaining control — but nothing in this repo pins that ordering, and the two unguarded bucket switches would corrupt counters rather than throw if it ever changed.

  Verified: `ThrowsForProviderAttribute` overrides neither `ApplyToTest` nor `Wrap`, so all 515 usages are covered by the one property stamp and the one call site.

## Out-of-scope observations

- **`ThrowsWhenCommand` converts a runtime skip into a hard failure** — pre-existing on `master`. A `ThrowsWhen` / `ThrowsForProvider` test that calls `Assert.Ignore(...)` or `Assert.Inconclusive(...)` for a provider on the expected-throw list produces a non-null `testResult.Message` (it holds the skip reason), so `ThrowsWhenCommand` takes the `SetResult(ResultState.Failure, …)` branch and reports `Expected a <X> to be thrown, but found: '<skip reason>'`. The new accounting reconciles this correctly (`Skipped → Failed` and `Inconclusive → Failed` both land on the final verdict), so it is not caused by this PR — but an early-out for non-Passed/non-Failed inner results would stop a deliberate skip being converted into a failure.
- **Outer repeat/retry wrappers inflate the heartbeat tally** — pre-existing on `master`. Repeat/retry wrappers sit at the outermost position of NUnit's command chain, so each iteration re-enters the reporter's per-test action and increments `started`/`completed` plus a pass/fail bucket, while `total` counts the test once (`[Repeat]` is command-level, so `TestCaseCount` stays 1). `Tests/Linq/Mapping/MappingSchemaTests.cs:388`, `Tests/Linq/Extensions/QueryHintsTests.cs:58` and `Tests/Linq/UserTests/Issue548Tests.cs:55` each use `[Repeat(100)]`, contributing ~297 phantom units on a full run. Beyond the cosmetic miscount, the inflated `completed` trips the done-latch at `Tests/Base/TestProgressReporter.cs:139` early, so an unfiltered run publishes `done: true` / `currentTest: null` well before it finishes and never un-latches. Filtered runs are unaffected, since `total` over-counts pre-filter there. If the accounting is being reworked anyway this is cheap to fold in — and it is also the mechanism that would make the over-broad `recentFailures` pruning actually bite.

## Baselines

No baseline changes — CI produced no `baselines/pr_5726` branch and no tracking pull request on `linq2db/linq2db.baselines`.

sdanyliv:
Review feedback addressed in 2cfb1656ee. `Reclassify` is gone — the reporter no longer publishes a provisional result and then compensates. Instead the unit is **deferred**: `ThrowsWhenCommand.Execute` brackets the inner command with `TestProgressTracker.BeginDeferred(fullName)` / `CommitDeferred(fullName, result)` (in a `finally`, unconditionally). While a deferral is in flight the reporter action only *samples* the outcome; the wrapper books the unit once, with the final verdict.

### Per finding

| Finding | Resolution |
| --- | --- |
| @igor-tkachev — provisional entry occupies the last `recentFailures` slot, real failure dropped and never backfilled | No provisional entry ever enters the list. Only booked (final) failures are added, so the bound is never consumed by something that will be withdrawn. |
| @igor-tkachev — two forced writes per expected-throw case, bypassing `WriteThrottleMs` | One publish per unit. An expected throw that fires books `Passed`, which is not a forced write at all — so the common case now costs *fewer* writes than on `master`, not more. |
| @igor-tkachev — regression coverage, both directions plus a parallel `MaxRecentFailures` scenario | `Tests/Linq/Infrastructure/TestProgressStateTests.cs`, 10 tests. Covers both directions, the boundary with a provisional in flight, and a 500-unit `Parallel.For` consistency check. |
| MIN001 — `before == after` skips the finalized message on Failed → Failed | Structurally gone: the commit always carries the wrapper's final message, there is no status comparison. Pinned by `WrongExceptionKeepsTheFinalizedMessage`. |
| SUG001 — `Reclassify` read `_enabled` instead of `EnsureInit()` | Both new entry points gate on `EnsureInit()`, like `OnBefore` / `OnAfter`. |
| SUG002 — `RemoveAll` prunes every entry with a matching name | No removal path exists any more. |
| MAJ001 — the accounting is untestable (internal, static, command-line-derived) | Extracted to `public sealed class TestProgressState` (`Tests/Base/TestProgressState.cs`): counters, the bounded list and the deferral protocol, with no file IO, no statics and no command-line dependency. `TestProgressTracker` holds one instance and keeps `Write` / `EnsureInit`. A test constructs its own instance and passes a publish callback that records every snapshot the heartbeat *would* have written. |

### Two deviations from SUG003 worth calling out

**Deferrals nest by depth, and only the outermost wrapper books.** SUG003 proposed making the commit idempotent with the first commit winning. That is wrong here: several `ThrowsWhen`-family attributes on one method are common — `[ThrowsRequiresCorrelatedSubquery]` + `[ThrowsForProvider(…)]` pairs appear across `Tests/Linq/Update/` and `Tests/Linq/Linq/` — and their wrappers nest. The *inner* one usually has `expectsException == false` for the running provider and returns the result untouched, while the *outer* one rewrites it. First-commit-wins would have booked the inner wrapper's pre-rewrite status. `BeginDeferred` / `CommitDeferred` therefore count depth, and only the level that closes last books.

**No property lookup, and no dependence on NUnit's command ordering.** The deferral is established by the wrapper itself rather than by inspecting `test.Properties["ThrowsWhen"]` in `OnAfter`, which also makes the protocol self-correcting: if the reporter action ever ends up *outside* the wrapper, `CompleteTest` finds no deferral in flight, books immediately, and `CommitDeferred` no-ops. Pinned by `ReporterOutsideTheWrapperBooksImmediately`. A unit still held back at run end (an exception escaping between the sample and the commit) is flushed by the root-suite teardown, so nothing silently drops out of the tally.

### Verification

Red → green on the same slice, `DistinctTests` against `SQLite.Classic` with `--test-progress`; MTP reports `failed: 0` in both runs:

- pre-fix build — heartbeat `"passed":136,"failed":2` plus two phantom `recentFailures` entries
- with the fix — `"passed":176,"failed":0,"recentFailures":[]` (176-case run, `DistinctTests` + `ElementOperationTests`)

`Tests.Base` builds clean in `Release`/`net462`; the new fixture is green on `net10.0`.

Out-of-scope items from the review (`ThrowsWhenCommand` converting a runtime skip into a hard failure, and `[Repeat]` wrappers inflating the tally and tripping the done-latch early) are both pre-existing on `master` and left alone here.

igor-tkachev (dismissed):
The previous blocking concerns are addressed by the deferred-commit design: provisional results are no longer published, the bounded failure list remains correct, and the accounting is now covered by dedicated tests, including parallel and boundary scenarios.

This change is limited to test infrastructure and does not affect linq2db production code.

Please update the PR description before merge: it still describes the old `Reclassify(...)` implementation, while the current code uses `BeginDeferred` / `CommitDeferred` with `TestProgressState`.

Tests/Base/TestProgressReporter.cs
MaceWindu: Verified met at `2cfb165`, and structurally rather than by repair: while a deferral is in flight `CompleteTest` returns before `Book` (`Tests/Base/TestProgressState.cs:131-138`), so a provisional outcome can no longer reach the bounded list at all. The "provisional entry takes the last slot, concurrent real failure discarded and never backfilled" sequence is therefore unreachable, not compensated.

Covered by `ProvisionalFailureDoesNotConsumeARecentFailureSlot` single-threaded, and by `ConcurrentUnitsKeepTheTallyConsistent` under concurrency — the latter saturates all 20 slots with real failures while ~167 deferred units are in flight and asserts every retained entry carries the real failure's message, which is exactly the "no provisional consumed a slot" property.

One caveat on the evidence, not the fix: the end-to-end verification on this PR is a single-TFM SQLite slice of ~176 cases, so the parallel behaviour is argued from the state machine rather than observed on a real multi-provider sweep.

Tests/Base/TestProgressReporter.cs
MaceWindu: Verified met at `2cfb165`. There is one booking per unit now, and an expected throw that fires books `Passed`, which does not set `force` — so the common case costs *zero* throttle-bypassing writes, against one on `master` and two in `9f2faec`. Pinned by the `Writes.Count(w => w.Force).ShouldBe(0)` assertion in `ExpectedThrowIsNeverPublishedAsFailure`, and the symmetric expected-but-didn't-throw case asserts exactly one forced write.

The "external watcher observes the false failure between the two writes" half is also gone, since nothing provisional is published — `ExpectedThrowIsNeverPublishedAsFailure` asserts every recorded snapshot has `failed == 0`.

MaceWindu:
Pushed three follow-up commits to this branch (`2cfb165..094fabae3`). Two are cosmetic; the third is a separate accounting defect I found while reviewing, measured rather than reasoned.

| Commit | Change |
| --- | --- |
| `19c400f55` | The class doc said the deferral gives "a single write per unit". Every unit publishes twice (`StartTest` and `Book`), both throttled — what the deferral guarantees is a single *booking*, and so at most one forced write. Wording only. |
| `3e2865784` | Comment on `_deferred` recording why keying on `ITest.FullName` is sound: test names are required to be unique project-wide (the baseline files are keyed by the same fully-qualified name), so a colliding name is a test-authoring bug to fix rather than a case this protocol has to handle. |
| `094fabae3` | `Book` counted a unit per invocation, so a `[Repeat]` / `[Retry]` test booked one unit **per iteration**. |

### On `094fabae3`

`[Repeat]` / `[Retry]` are command-level wrappers that sit *outside* the reporter's per-test action, so the action runs once per iteration for a single test case, while the run total counts that case once. Measured on a throwaway Playground probe with `--test-progress`:

| Probe | runner | heartbeat before | heartbeat after |
| --- | --- | --- | --- |
| plain test | 1 test | `started 1, completed 1` | `started 1, completed 1` |
| `[Repeat(5)]` | 1 test | one booking per iteration | `started 1, completed 1` |
| both together | 2 tests | `completed 3` published, 6 internally | `started 2, completed 2, passed 2` |

`Tests/Linq` has three `[Repeat(100)]` sites — `Mapping/MappingSchemaTests.cs:388`, `Extensions/QueryHintsTests.cs:58`, `UserTests/Issue548Tests.cs:55` — so an unfiltered run accumulated ~297 phantom units. That pushed `completed` past `total` and tripped the done-latch in `Book` early; after that, remaining passes are not forced writes and `MarkDone` returns before its `Publish` when `_done` is already set, so the heartbeat file stayed frozen at `done: true` for the rest of the run. Filtered runs were unaffected, since `total` is the pre-filter count there and the latch never fires.

`TestProgressState` now records the verdict booked per case and, on re-entry, withdraws the previous one instead of counting a new unit — keeping the latest verdict, which is what the runner's summary reports. Cost is one dictionary entry per test case for the run.

`RepeatIterationsBookOneUnitWithTheLatestVerdict` pins it, and also that a withdrawn failure drops its `recentFailures` entry. Verified red with the fix reverted, green with it; the full fixture is 11/11 on net10.0.

Happy to drop any of these or move the third to its own PR if you'd rather keep this branch scoped to the deferral work — it is separable.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `094faba`.</sub>

PR #5727, opened 2026-07-23
Fix #4624: eager-load an association off a GroupBy entity key

### LINKED ISSUE #4624: Regression in L2DB 3.0.0-rc0 + linq2db.EntityFrameworkCore 3.2.0 - Null reference exception
I'm facing regression on following code path, I receive crashes with null reference exception. The code worked fine with L2DB 2.xx + EF Core provider 2.xx

I'm attaching a sample repro for the problem
[CrashRep.zip](https://github.com/linq2db/linq2db.EntityFrameworkCore/files/4780072/CrashRep.zip)

Sample code goes like this:
```cs
var query = db.ACL_ItemTicketDate.ToLinqToDB()
                                 .Where(p => p.Id < 1000)
                                 .GroupBy(p => p.ACL_Item)
                                 .Select(p => new 
                                 {
                                     AclItemId = p.Key.Id,
                                     AclItemName = p.Key.Name,
                                     TotalEntryCount = p.Key.ACL_Entries.Sum(ec => ec.EntriesCount),
                                     TotalEntryAllowed = p.Key.ACL_ItemTicketDates.Select(at => new 
                                     {
                                         EntryCount = at.EntryCount
                                     }).ToList()
                                 });

```

And the crash stack trace like this:
```cs
at LinqToDB.EntityFrameworkCore.EFCoreMetadataReader.CompareProperty(MemberInfo property, MemberInfo memberInfo)
at LinqToDB.EntityFrameworkCore.EFCoreMetadataReader.CompareProperty(IProperty property, MemberInfo memberInfo)
at LinqToDB.EntityFrameworkCore.EFCoreMetadataReader.<>c__DisplayClass8_0`1.<GetAttributes>b__1(IProperty p)
at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
at LinqToDB.EntityFrameworkCore.EFCoreMetadataReader.GetAttributes[T](Type type, MemberInfo memberInfo, Boolean inherit)
at LinqToDB.Mapping.MappingSchema.<>c__DisplayClass51_0`1.<GetAttributes>b__0(IMetadataReader mr)
at System.Linq.Enumerable.<SelectManyIterator>d__180`3.MoveNext()
at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
at LinqToDB.Mapping.MappingSchema.GetAttributes[T](Type type, MemberInfo memberInfo, Boolean inherit)
at LinqToDB.Mapping.MappingSchema.GetAttributes[T](Type type, MemberInfo memberInfo, Func`2 configGetter, Boolean inherit, Boolean exactForConfiguration)
at LinqToDB.Mapping.EntityDescriptor.Init()
at LinqToDB.Mapping.EntityDescriptor..ctor(MappingSchema mappingSchema, Type type)
at LinqToDB.Mapping.MappingSchema.<>c__DisplayClass91_0.<GetEntityDescriptor>b__0(ICacheEntry o)
at LinqToDB.Common.Internal.Cache.CacheExtensions.GetOrCreate[TItem](IMemoryCache cache, Object key, Func`2 factory)
at LinqToDB.Mapping.MappingSchema.GetEntityDescriptor(Type type)
at LinqToDB.SqlQuery.SqlTable..ctor(MappingSchema mappingSchema, Type objectType, String physicalName)
at LinqToDB.Linq.Builder.TableBuilder.TableContext..ctor(ExpressionBuilder builder, BuildInfo buildInfo, Type originalType)
at LinqToDB.Linq.Builder.TableBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
at LinqToDB.Linq.Builder.WhereBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
at LinqToDB.Linq.Builder.GroupByBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
at LinqToDB.Linq.Builder.SelectBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
at LinqToDB.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
at LinqToDB.Linq.Builder.ExpressionBuilder.Build[T]()
at LinqToDB.Linq.Query`1.CreateQuery(IDataContext dataContext, Expression expr)
at LinqToDB.Linq.Query`1.GetQuery(IDataContext dataContext, Expression& expr)
at LinqToDB.Linq.ExpressionQuery`1.GetQuery(Expression& expression, Boolean cache)
at LinqToDB.Linq.ExpressionQuery`1.System.Collections.Generic.IEnumerable<T>.GetEnumerator()
at CrashRepro.Program.Main(String[] args) in C:\Users\Bruno\Desktop\CrashRep\CrashRepro\Program.cs:line 35
```

sdanyliv:
Fixes #4624.

Eager-loading a collection association off a **GroupBy entity key** — e.g. `GroupBy(x => x.Parent).Select(g => new { g.Key.Id, g.Key.Children.Select(...).ToList() })` — failed. A GroupBy key entity is materialized as a value-construct of its grouped columns with no associations, so navigating `g.Key.<association>` off it couldn't resolve (`NullReferenceException` historically, `Table not found` / `could not convert to SQL` latterly).

## Fix (2 core changes)

- **`EntityConstructorBase.BuildFullEntityExpressionInternal`** — when building a key entity during association expansion (`IsKeys() && IsExpand()`), include the entity's associations in the construct, so `g.Key.<association>` resolves for navigation.
- **`GroupByBuilder` `KeyContext.MakeExpression`** — preserve the `Expand` purpose when resolving an *entity* key (it was hardcoded to `Sql`, dropping the flag). Scalar keys stay `Sql` (guarded by `!IsScalarType`) so their eager-load correlation isn't disturbed — under `Expand` a bare scalar comes back as an unresolved column reference, whereas entity/composite keys resolve via construction.

Correlation / key-extraction runs with `Sql` + `ForKeys` (no associations → no recursion); only `Expand`-driven navigation gets the associations.

## Tests

`TestGroupByEntityKeyEagerLoad` (entity key → association) and `TestGroupByCompositeKeyEagerLoad` (anonymous composite key) in `EagerLoadingTests.cs`; EF `Issue4624Test` un-gated.

## Validation

- SQLite full suite: 10,355 pass / 0 fail
- Both tests on 14 providers (incl. YDB, ClickHouse, Oracle, DB2, Informix, SAP HANA, Sybase, Firebird): 56 / 0 fail
- Release `net10.0` (analyzers) + `netstandard2.0` clean

No `[ThrowsForProvider]` gating needed — the eager-load compiles to a keyed batch re-query (`SELECT DISTINCT keys … INNER JOIN detail`), which every tested provider supports.


MaceWindu (approved):
*Agentic review — produced with AI assistance and verified against the code at HEAD.*

## Review notes

- CI was green on `b1a38c71`, and build 22968 was triggered after that commit landed, so the run genuinely covered the reviewed head.
- Baselines: 195 added (`TestGroupByEntityKeyEagerLoad` x68, `TestGroupByCompositeKeyEagerLoad` x68, `Issue4624Test` x59), 1 modified, 0 deleted. The single modification is `Issue1174Tests.TestConcurrentSelect(ClickHouse.Octonica)` dropping one of two identical dumps — a concurrency-test capture artifact, unrelated to this change. No pre-existing baseline moved, which bounds the observed blast radius of the `GroupByBuilder` change to zero.
- Generated SQL for all three tests is a proper keyed batch (`SELECT DISTINCT keys` joined to detail) with no N+1 — which was the original complaint in #4624. The shape is identical across SQL Server, YDB, Firebird 2.5 and the EF configurations.

## Prior-review audit

All four earlier Copilot claims are fixed at HEAD:

| Claim | Verdict |
|---|---|
| `g.Key` dereferenced without null-forgiving | **Fixed** — `g.Key!` at both sites. `d.Master` is `MasterClass?`, so the `!` is required to compile, and `GenerateData` produces no null-key group. |
| Comment said "only for entity keys" while the code gated on non-scalar | **Fixed** — reworded. |
| Method-based associations would throw in `Expression.MakeMemberAccess` | **Fixed** — guard added. The premise was wrong, though: `EntityDescriptor._associations` is populated only from `TypeAccessor.Members`, so a method-based `[Association]` never appears in that list. |
| Rebuilding via the constructor drops `Constructor` / `ConstructorMethod` metadata | **Fixed correctly** — `ReplaceAssignments` routes through the copy constructor, which preserves `ObjectType`, `ConstructType`, `Constructor`, `ConstructorMethod`, `Parameters`, `ConstructionRoot` and `MappingSchema`. |

## Follow-up commits pushed to this branch

Three commits were pushed on top of `b1a38c71`, plus a `master` merge.

### 1. `c77ada28a` — remove the unreachable key-entity association block

The 43-line block added to `EntityConstructorBase.BuildFullEntityExpressionInternal` was gated on `flags.IsKeys() && flags.IsExpand()`, which can never be true, so the block never executed. The fix for #4624 is entirely the `GroupByBuilder` key-purpose gate.

Three independent lines of evidence:

**Static.** `ProjectFlags.Keys` is `1 << 5` and `ProjectFlags.Expand` is `1 << 11` — distinct bits. Across all of `Source/`, `ExpressionBuildVisitor.GetProjectFlags` is the sole producer of the `ProjectFlags` value that reaches `IBuildContext.MakeExpression`. It switches on a single `_buildPurpose` and adds `Keys` only in the `Sql`, `Expression` and `Extract` arms, `Expand` only in the `Expand` arm — and that arm ignores `BuildFlags.ForKeys` entirely. Nothing ORs the two bits onto an existing value, and every `BuildFullEntityExpression` call site either hardcodes `ProjectFlags.SQL` / `ProjectFlags.Expression` or forwards `flags` straight from `MakeExpression`.

**Empirical.** Instrumenting the call site and running the whole `EagerLoadingTests` fixture on SQLite recorded 1621 calls in six distinct flag combinations — `SQL`, `Expression`, `SQL+Keys`, `SQL+MemberRoot`, `Expression+ForSetProjection`, `SQL+Keys+MemberRoot`. `Expand` never appeared at all, so it is not merely that `Keys` and `Expand` cannot co-occur — `Expand` never reaches this method under any combination.

**Behavioural.** All 162 `EagerLoadingTests` pass on SQLite with the block removed, including both new tests. The file is restored to its pre-PR state exactly (blob `79994b66a`, identical to `master`).

The PR carries its own proof of this: the comment added at `GroupByBuilder.cs:382-384` states that "under Expand a ForKeys build loses the Keys flag in GetProjectFlags" — which is exactly what makes `IsKeys() && IsExpand()` unsatisfiable. The two new comments contradict each other.

**One thing left for the author:** the PR description still lists this as fix (1) — "include the entity's associations in the construct, so `g.Key.<association>` resolves for navigation". That paragraph describes a mechanism that never ran and is worth correcting, since it will otherwise become the merge commit message.

### 2. `3846808cf` — correct `TestGroupByCompositeKeyEagerLoad`'s description

The description read "composite group key must not take the entity-key Expand path", which is the opposite of what the test does. `new { d.MasterId }` is an anonymous type, so `MappingSchema.IsScalarType` reports it non-scalar, `!IsScalarType(Body.Type)` is true, and — together with `IsExpand()` and the `MemberRoot` flag that `g.Key.MasterId` carries — the test takes the new `Expand` branch, not the `Sql` one.

Worth noting that `MemberRoot` is set for *any* member access off the key, not just an association: `ExpressionBuildVisitor.HandleRoot` (line 1583) wraps `Visit(node.Expression!)` in `CombineBuildFlags(BuildFlags.ForMemberRoot)` for every `MemberExpression`, so a plain column read such as `g.Key.Id1` takes the new branch too. Only a bare `g.Key` with no member access on top stays `Sql`. Nothing currently pins that scalar-key path — the nearest existing coverage, `TestWhenMasterConnectedViaExpression`, is SQLite-only and has no assertion — but since that branch keeps its pre-PR behaviour it carries no regression risk from this change.

### 3. `1fba644b7` — assert results in EF `Issue4624Test`

The un-gated test ended at `var result = query.ToList();` with no assertion, and `Issue4624ItemTicketDates` is never seeded anywhere in `Tests/EntityFrameworkCore` — the only references to the model are the `DbSet` and `modelBuilder` registrations in `Models/IssueModel/IssueContextBase.cs:26,155`. It therefore ran against zero rows and could not distinguish a working eager load from `TotalEntryAllowed` coming back empty; removing `[ActiveIssue]` on that basis only established that the query translates.

It now seeds one item with two entries and two ticket dates and asserts the grouped projection — the correlated sum and the contents of the eager-loaded collection. Both fail if the eager load returns nothing. Verified green on `SQLite.MS` / EF9.

## Out-of-scope observations

- **The `IsExpand() && !IsKeys()` guards in `TableContext` / `SelectContext` are vestigial for the same reason.** `TableBuilder.TableContext.cs:215` and `SelectContext.cs:89` both test `flags.IsExpand() && !flags.IsKeys()`, which — given `GetProjectFlags` can never set both bits — is just `flags.IsExpand()`. This is pre-existing on `master` and not caused by this PR, but it is what makes an `IsKeys() && IsExpand()` gate look idiomatic: the codebase reads as though `Keys | Expand` were a live combination. Filed as #5814, which proposes an internal analyzer over `Source/LinqToDB` to reject unsatisfiable `ProjectFlags` conjunctions in both directions (always-false conjunctions and redundant negated clauses), so this class of dead code fails the build rather than shipping green.

## Baselines

Delta: [linq2db.baselines PR #2075](https://github.com/linq2db/linq2db.baselines/pull/2075) (OPEN)

195 added / 1 modified / 0 deleted, confined entirely to this PR's own three tests. Structural scan: `DECLARE` signal delta 0, identifier-rename mismatches 0, all 195 added files net-positive. Provider coverage is uniform — 2 files per non-EF provider configuration and 1 per EF configuration — with no provider missing from the expected set, so there is no subset-coverage failure fingerprint.

Note the three baseline files for the two new `EagerLoadingTests` and the EF test will need regenerating after the follow-up commits above, since the EF test now issues inserts before its query.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Eager-loading an association off a `GroupBy` entity key did not work:

  ```csharp
  from d in db.Detail.LoadWith(x => x.Master.Details)
  group d by d.Master into g
  select new { g.Key.Id, Details = g.Key.Details.Select(x => x.DetailId).ToList() }
  ```

  This applies to a whole-entity group key, in plain LinqToDB and through the EF Core integration. A scalar group key follows a different path and is unchanged.


Eager-loading an association off a `GroupBy` entity key now works.


---
<sub>Generated from commit `8187f14`.</sub>

PR #5732, opened 2026-07-29
Fix DistinctBy losing preceding query state ("Table not found")

sdanyliv:
`DistinctBy` after a state-carrying operator threw at SQL-build time:

    LinqToDBException : Table not found for 't9[Id:6].Country => t6[5].Country := t7.Country??'

Reproduces on plain master with an ordinary mapped entity — e.g.
`table.Where(c => c.CustomerId.StartsWith("DST")).OrderBy(c => c.CustomerId).DistinctBy(c => c.Country)`.
Not provider-specific (SQL Server behaves identically); PostgreSQL/DuckDB are unaffected
only because they take the `DISTINCT ON` branch. Nullability is irrelevant — a
non-nullable key fails the same way, and `Select().Distinct().OrderBy().DistinctBy()`
fails with no `Where` at all.

## Cause

`DistinctByBuilder`'s ROW_NUMBER rewrite re-enters `TryBuildSequence` with a bare
`ContextRefExpression` over the already-built sequence — the only site in the codebase
that does so. `ExpandToRoot` resolves it under `ProjectFlags.Subquery`, and
`SubQueryContext` answers with a reference to its *own inner context*, repeatedly, until
the reference names the innermost table context. Everything the peeled wrappers carried
is lost — the `Where` predicate **silently disappears from the SQL** — and the OVER-clause
placeholders captured against those wrappers dangle.

Fix: anchor the sequence at that one re-entry site. `SourceAnchorContext` is a
`PassThroughContext` that answers a self-referencing `Subquery` path with the path itself,
so the reference keeps naming the sequence that was actually built.

## Why not fix it in `SubQueryContext`

That was tried first — making `SubQueryContext` answer a self-referencing `Subquery` path
with the path, mirroring its existing `Root` case. It fixes the bug, but the descend is
load-bearing: it also decides **which query an aggregate is attached to**. Suppressing it
moves a grouping aggregate into a correlated subquery while its arguments stay bound to the
outer grouped query, which is invalid for an ordered-set aggregate:

```sql
SELECT g_1."CategoryId",
       (SELECT PERCENTILE_DISC(0.5) WITHIN GROUP (ORDER BY g_1."IntValue")
        FROM "WindowFunctionTestEntity" t
        WHERE g_1."CategoryId" = t."CategoryId" FETCH NEXT 1 ROWS ONLY)
FROM "WindowFunctionTestEntity" g_1
GROUP BY g_1."CategoryId"           -- ORA-00979: "IntValue" must appear in the GROUP BY
```

That produced 70 Oracle failures across Oracle 12/18/19/21/23 in `test-all` 22512.
Confining the change to the one re-entry site avoids touching aggregate placement at all.

## Verification

- `DistinctByMethodTests` + `GroupByTests` + `SelectManyTests` + `WindowFunctionsTests` on
  SQLite ×2, Oracle 23 and Access: 2610/2610
- `WindowFunctionsTests` on Oracle 23 (where the `SubQueryContext` variant failed): 311/311
- Release / netstandard2.0 / net462: 0 warnings

Adds 15 tests: 14 `DistinctBy` shapes (two filters, projection, `AsSubQuery`, `Distinct`,
composite key, join, filters on both sides, nested `DistinctBy`, `Take`, …) and one covering
an aggregate reached through nested `let` transparent identifiers. Ten of the 14 reproduce
the failure on pre-fix source; the rest are breadth coverage and are marked as such in the
file.

Also drops a dead `TestProvName.AllSybase` entry from `Issue672Test`'s `ThrowsForProvider`:
`DataSources` takes an exclusion list, so Sybase never runs that test.

### DIFF
--- a/Source/LinqToDB/CompatibilitySuppressions.xml
+++ b/Source/LinqToDB/CompatibilitySuppressions.xml
@@ -345,6 +345,13 @@
   </Suppression>
   <Suppression>
     <DiagnosticId>CP0002</DiagnosticId>
+    <Target>F:LinqToDB.Internal.SqlProvider.BasicSqlBuilder.Step.TypedExpression</Target>
+    <Left>lib/net10.0/linq2db.dll</Left>
+    <Right>lib/net10.0/linq2db.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
     <Target>F:LinqToDB.Internal.SqlQuery.CteClause.CteIDCounter</Target>
     <Left>lib/net10.0/linq2db.dll</Left>
     <Right>lib/net10.0/linq2db.dll</Right>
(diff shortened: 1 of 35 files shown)


MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Fixed `DistinctBy` failing with `Table not found for '...'` when operators preceded it in the query — most commonly a `Where`. On the `ROW_NUMBER` emulation path the preceding query state was dropped, taking the filter with it, and query building then failed. ([#5732](https://github.com/linq2db/linq2db/pull/5732))


_(none)_


---
<sub>Generated from commit `9b387b5`.</sub>

PR #5670, opened 2026-06-30
Unify agent instructions under .agents/ + major curation tooling (release, review, analyzers)

MaceWindu:
> Continues #5521, which GitHub auto-closed when its head branch was renamed (infra/claude-curation -> infra/agents-curation). Review history lives on #5521.

---

## 1. Agent-instruction unification (`.agents/`)

Consolidates the formerly Claude-only `.claude/` corpus, the Copilot review file, and (newly) Codex into one shared contributor source under `.agents/`, with thin per-agent entry points:

- New root **`AGENTS.md`** — canonical agent-agnostic contributor core; read natively by Codex, `@import`ed by Claude, pointed to by Copilot.
- `.claude/{docs,skills,agents,hooks,scripts,knowledge-base}` → `.agents/` (git renames); **`.claude` → `.agents` symlink** (Claude Code hardcodes skill/hook/settings discovery there).
- `CLAUDE.md` thinned to an entry point + Claude overlay; `.github/copilot-instructions.md` repointed at `AGENTS.md`.
- Renames: KB area `CLAUDE-INFRA` → `AGENTS-INFRA`; skill `audit-claude` → `audit-agents`.
- Agent-agnostic guidance split out of the Claude-only docs into `maintaining-the-corpus.md` + `windows-dev-gotchas.md` (+ the untrusted-fetched-config rule into `AGENTS.md` Security); Claude-tool specifics remain in the overlay.

**Reviewer note — Windows symlink:** `.claude` is a git-tracked symlink to `.agents/`; on Windows it materializes only with `core.symlinks=true` + Developer Mode, otherwise a clone gets a broken `.claude` and Claude Code discovery fails. Needs a CONTRIBUTING note.

## 2. Major curation changes (rolling `infra/claude-curation` work since master)

- **Release-prep flow** — `/release` orchestrator + sub-skills (`/release-deps`, `/release-publicapi`, `/release-milestone-check`, `/release-test-matrix`, `/release-notes-validate`, `/release-verify`, `/release-publish`, `/release-postpublish`) with the single-build verification model.
- **`/review-pr` + `/verify-review` rework** — resolver-aware PR-context + thread-reply scripts, mode-choice gate.
- **`/profile-analyzers`** — Roslyn analyzer build-time profiling.
- **Baselines regression detection** — size-outlier + regression-candidate scan in `baselines-diff.ps1`.
[...]

PR #5735, opened 2026-07-31
Move agent instructions to a linq2db/agents submodule at .claude/

MaceWindu:
Moves the AI-agent instruction corpus out of this repo into [linq2db/agents](https://github.com/linq2db/agents), mounted here as a git submodule at `.claude/`.

## Why

- **Instruction churn had nowhere to go.** The corpus accumulated on a long-lived `infra/agents-curation` branch that was rarely merged, because merging produced a large diff of files no product change depended on. It now has its own repo, its own history and its own review surface, and this repo carries none of it.
- **`.claude` was a symlink to `.agents/`.** Claude Desktop can't follow it; it needed `core.symlinks` + Developer Mode on Windows; and it broke tool `Glob` for any pattern using the symlink as a path component (measured: `.claude/skills/*/SKILL.md` → 0 files, `.agents/skills/*/SKILL.md` → 35, while `Read`/`Grep` resolved it either way — which is why the failures looked arbitrary). `.claude/` is now a real directory and `.agents/` is gone entirely; no symlink in either direction.

## The 1211 deletions are the move, not a shrink

Every deleted `.agents/**` file is in the agents repo, with history across **both** path eras — the corpus lived at `.claude/` until #5670 renamed it to `.agents/`, and both prefixes were rewritten to that repo's root, so `git log`/`git blame` cross the rename without `--follow`. Its root commit is still `Initial setup for Claude code (#5452)`.

## What this repo keeps

| Path | Role |
|-|-|
| `.gitmodules` | `path = .claude`, `branch = master`, `update = merge`, `ignore = all` |
| `.claude` | gitlink — a **bootstrap pointer, not a version pin** |
| `AGENTS.md`, `CLAUDE.md` | one-screen trampolines into the corpus |
| `.githooks/` | corpus refresh + two commit guards (opt in: `git config core.hooksPath .githooks`) |

`branch = master` + `update = merge` let `git submodule update --remote --merge` follow the corpus tip — `merge` is a no-op when the submodule is already ahead, where `checkout` mode would rewind it. `ignore = all` is required rather than `ignore = dirty`: `dirty` still reports committed differences from the gitlink, which would leave this repo permanently dirty. It also means `git add .` can't stage a gitlink bump.

`.githooks/` refreshes `.claude/` on checkout / merge / rebase (skipping any tree with uncommitted corpus edits, and never failing the triggering git command), and `pre-commit` refuses a gitlink bump and edits to either trampoline — both overridable with `--no-verify`.

Verified before opening, in a throwaway clone:

| Check | Result |
|-|-|
| `pre-commit` on a trampoline edit | refused, exit 1 |
| `pre-commit` on a forced gitlink bump | refused, no commit created |
| plain `git add .claude` | won't stage the bump |
| `git status` with corpus ahead **and** dirty | clean |
| `post-checkout` refresh | corpus rolled back to an older commit → branch switch → back at tip, still on `master` |
| hook files | `i/lf w/lf`, mode `100755` |

## Also in here

- `linq2db.slnx` loses the `/.agents/*` virtual-folder block (227 lines, 43 folders) along with the skill that maintained it — listing submodule content would put corpus churn back into this repo.
- `.github/copilot-instructions.md` becomes **self-contained**: GitHub-side Copilot review has no submodule checkout, so review-relevant invariants are stated inline instead of linked.
- `CONTRIBUTING.md` documents the submodule clone (`--recurse-submodules`, `core.hooksPath`) instead of the symlink workaround. Note a plain `git clone` leaves `.claude/` empty *and* `git status` clean — an unpopulated submodule isn't a deletion — so the only symptom is an agent with no instructions.
- `.gitattributes` pins `.githooks/**` to LF (those files have no `.sh` extension, so `text=auto` would have shipped them with a CRLF shebang).
- Four `Tests/` doc-comment path fixes (`.agents/docs/testing.md` → `.claude/docs/testing.md`). `.build/.agents/` — the gitignored scratch dir referenced from `TestProgressReporter` — deliberately keeps its name; it was never part of the corpus.

No `Source/` changes; nothing about building or testing linq2db depends on the submodule.


CONTRIBUTING.md
```diff
@@ -4,18 +4,22 @@ uid: contributing
...
+Nothing about building or testing linq2db depends on the submodule — clone it only if you use an AI agent on this codebase:
 
-- **Developer Mode** (*Settings → Privacy & security → For developers*), or run git elevated, so the OS permits symlink creation.
-- `git config --global core.symlinks true` (or pass `-c core.symlinks=true` to `git clone`).
+```
+git clone --recurse-submodules https://github.com/linq2db/linq2db.git
+git config core.hooksPath .githooks
+```
+
+Already cloned? `git submodule update --init`, then `git -C .claude switch master` (init checks out a detached HEAD, and a corpus commit made there would go nowhere useful), then the same `core.hooksPath` line. Note that a plain `git clone` leaves `.claude/` as an **empty directory** and `git status` still reads clean — an unpopulated submodule is not a deletion — so the only symptom is that your agent loads no instructions.
```
sdanyliv, line 18: **Major (MAJ001)** — populating the submodule does not fix an agent session that is already running, and this paragraph reads as though it does.

Claude Code resolves its skill/subagent registry and the `CLAUDE.md` import set **once, at session start**. Running `git submodule update --init` mid-session puts every corpus file on disk and changes nothing about that session — so the closing sentence here, "the only symptom is that your agent loads no instructions", describes the symptom that *persists after* the documented fix. The natural read of that is "the submodule move is broken", not "restart the agent".

Reproduced on this branch: after a clean `git submodule update --init` — all 34 `.claude/skills/*/SKILL.md` present, each with valid `name:`/`description:` frontmatter — the running session still resolved none of the project's skills (`kb-status` reported as an unknown skill), and the root instructions it had loaded were still the pre-migration `.agents/` copies.

This is the dominant migration path for anyone who already has a clone, so it is the first thing most people will hit.

```suggestion
Already cloned? `git submodule update --init`, then `git -C .claude switch master` (init checks out a detached HEAD, and a corpus commit made there would go nowhere useful), then the same `core.hooksPath` line. Note that a plain `git clone` leaves `.claude/` as an **empty directory** and `git status` still reads clean — an unpopulated submodule is not a deletion — so the only symptom is that your agent loads no instructions.

**Restart your agent after populating the submodule.** Claude Code reads skills, subagents and the `CLAUDE.md` import set once at session start, so a session that was already open while `.claude/` was empty stays empty-handed even after `git submodule update --init` succeeds.
```

The same sentence is missing from the corpus's own setup doc (`docs/claude-setup.md` in the agents repo, under the fresh-clone note) — separate commit there.

MaceWindu: Confirmed and fixed in `2ebb4416`.

I couldn't re-run your repro (restarting the session mid-check isn't available to me), but the mechanism corroborates from the inside: this session's always-loaded set is still the pre-migration `.agents/`-era `CLAUDE.md`/`AGENTS.md` even though the corpus it describes was rebuilt and pushed during the session. Load-once, exactly as you describe.

Applied your suggestion to `CONTRIBUTING.md`, and extended it to two sites the finding also covers — **both root trampolines**. Those matter more than the contributing guide here: `AGENTS.md` and `CLAUDE.md` are what an agent reads *when it has loaded nothing*, so "restart the session" has to be there or the reader follows the init steps, sees no change, and concludes the submodule move is broken. The `CLAUDE.md` comment was also missing `git -C .claude switch master` (init leaves a detached HEAD).

Corpus side is a separate commit as you asked: [linq2db/agents@4e7bbce](https://github.com/linq2db/agents/commit/4e7bbce) adds the same warning to `docs/claude-setup.md`'s fresh-clone note, with the load-once rationale and a pointer back to this review.

MaceWindu: **Correction:** to my earlier reply — I adopted your mechanism wholesale, and half of it doesn't hold. The **advice** (restart) is right; the **reason** I wrote into the docs was wrong about skills.

Measured just now, switching the primary clone onto this layout: immediately after `git submodule update --init`, the harness re-issued its skill list from the newly-populated corpus **mid-session** — a skill deleted in the migration vanished from it, and reworded descriptions updated — while the always-loaded instruction set stayed on the pre-migration copy. So skill discovery *can* refresh without a restart; the `CLAUDE.md` import set does not.

That doesn't weaken the finding, it sharpens it: the import set is the load-once half, and it's the half that leaves an agent with no project rules — which is exactly the symptom you flagged. And skill refresh isn't dependable in either direction, since your repro had all 34 skills on disk with `kb-status` still unknown.

Fixed in #5736 (`CONTRIBUTING.md` + both trampolines) and [linq2db/agents@aa4c00a](https://github.com/linq2db/agents/commit/aa4c00a).

One more measured result while I was in there, since it touches the worktree bootstrap this PR documents: a directory **junction** to the primary clone's `.claude/` is not viable. `Glob` doesn't traverse a reparse point used as a path component — `.claude/skills/*/SKILL.md` matched 0 files through the junction against 34 in the primary clone, the same defect the old `.claude` → `.agents` symlink caused. Git-level access fails too (`not a git repository: .claude/../.git/modules/.claude`). A control `Glob` for a real directory in the same worktree worked, so it's the junction, not the worktree. The clone-with-`--reference` recipe is verified working and stays the documented one.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `2ebb441`.</sub>

PR #5736, opened 2026-07-31
Correct the corpus reload semantics in the submodule setup docs

MaceWindu:
Follow-up to #5735, correcting a mechanism claim I got half wrong there.

#5735 added a "restart your agent after populating the submodule" warning in response to [sdanyliv's review](https://github.com/linq2db/linq2db/pull/5735#discussion_r3691843764). The **advice is right**; the **reason given for it wasn't**. It said skills, subagents *and* the `CLAUDE.md` import set are all resolved once at session start.

Measured while switching the primary clone onto this layout: immediately after `git submodule update --init`, the harness re-issued its skill list from the newly-populated corpus **mid-session** — a deleted skill vanished, reworded descriptions updated — while the always-loaded instruction set stayed on the pre-migration copy.

So:

- **The `CLAUDE.md` import set is the load-once half**, and the one that matters: a session that started against an empty `.claude/` keeps running with no project rules however many files land on disk afterwards.
- **Skill discovery can refresh, but isn't dependable** — sdanyliv's own repro had a freshly-initialized submodule with all 34 skills present still reporting `kb-status` as unknown.

Restart remains the only way to be sure the session matches disk; the docs now say that for the right reason. Three files, +6/−5: `CONTRIBUTING.md` and both root trampolines.

The corpus side is [linq2db/agents@aa4c00a](https://github.com/linq2db/agents/commit/aa4c00a), which also records a second measured result: a directory **junction** to the primary clone's `.claude/` is ruled out as a worktree bootstrap. It half works — files readable, `rev-parse` resolves — but `Glob` doesn't traverse a reparse point used as a path component, so `.claude/skills/*/SKILL.md` matched 0 files in the worktree against 34 in the primary clone. That's the same defect the old `.claude` → `.agents` symlink caused and that #5735 existed to remove. A control `Glob` for a real directory in the same worktree worked, so the junction is the cause.

Committed with `--no-verify`: `.githooks/pre-commit` (added in #5735) refuses staged edits to the trampolines. It fired here unprompted, which is correct default behaviour — this is the documented override for a deliberate wording fix in them.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `56ef67d`.</sub>

Issue #4306, opened 2023-10-13
SQLSever Type for TimeSpan is "Time"


jogibear9988:
### Describe your issue

I'm currently looking to work on issue #3993 

There I now recognized that the default type for "TimeSpan" if the tests create the database is "Time". Is this really a good idea?
I would think the default should be "bigint", cause bigint values also could be used in DateAdd, and a TimeSpan is not a Time, cause Time means "TimeOfDay".

What do you think?

MaceWindu:
That will be a breaking change to change database defaults and we probably will not do it.

Regarding int64 support:
- user should specify data type for TimeSpan explicitly using
  - `DataType.Int64` on column mapping
  - `mappingSchema.AddScalarType(typeof(TimeSpan), DataType.Int64)` on context (mapping schema level)
- there should be required converters set between database and client values

Sql Server provider already has all required mappings to map TimeSpan to int64 ticks (as I was requesting it like 6 years or more ago and still use it in my projects). Other providers could require additional configuration to provide such mappings out-of-box

jogibear9988:
all infos tracked in #4308

Issue #3993, opened 2023-02-23
Problem with Calculated DateTime and Where Filter 


jogibear9988:
### Describe your issue

see also issue #3840 

The fields here do not work when used in a where filter

```
Exception message:

 System.ArgumentException : Property 'Int32 Milliseconds' is not defined for type 'System.Nullable`1[System.TimeSpan]' (Parameter 'property')

or 

System.Data.SqlClient.SqlException : Operandentypkollision: datetime2 ist inkompatibel mit bigint

or

System.Data.SqlClient.SqlException : Der Operanddatentyp datetime2 ist für den subtract-Operator ungültig.
   


see my pull req for the test

jogibear9988:
@MaceWindu sorry, forgot to add filter tests also in the issue #3840 
They still do not work

jogibear9988:
Is there a workaround at the moment?

MaceWindu:
Closing as duplicate of #4308, which tracks the same bug — TimeSpan sub-property access (`TotalMilliseconds`/`TotalSeconds`/`Milliseconds`) dropped or mistranslated in generated SQL — with a cleaner minimal repro. The WHERE-filter error variants and the #3840 reference have been consolidated there. (PR #3994, which targeted both, was closed unmerged, so the bug is still open.) Any further info, please add to #4308.

PR #5739, opened 2026-08-03
Add native TimeSpan translation support

### LINKED ISSUE #2950: PostgreSQL Test Error when get Hours on TimeSpan
model
```c#
	[Table(Schema="Member", Name="Table")]
	public partial class Table
	{
		[PrimaryKey, NotNull    ] public int       Id   { get; set; } // integer
		[Column,        Nullable] public DateTime? Date { get; set; } // date
		[Column,        Nullable] public TimeSpan? Time { get; set; } // time without time zone
	}
```
test
```c#
var query = (from x in db.Table
                         select new
                         {
                             Output = x.Time.Value.Hours,
                         });
            query.ToArray();
```
Error: LinqToDB.Common.LinqToDBConvertException: 'Cannot convert value '12:34:56' to type 'System.Int32''
### Environment details
linq2db version: 3.3.0
Database: PostgreSQL 12.1
.NET Framework: 4.7.2

### LINKED ISSUE #4308: Propertyaccess get's completly removed from "TimeSpan" in SQL, so wrong query created
I've a object like this:

		public class Test
		{
			public virtual TimeSpan? PreNotification  { get; set; }
		}

with a maaping like this:

              new FluentMappingBuilder()
				.Entity<Test>()
					.HasTableName("Common_Topology_Locations")
					.Property(e => e.PreNotification)
						.HasDataType(LinqToDB.DataType.Int64)
				.Build()
				.MappingSchema);

and following query:

			var qry2 =
				(from t in db.GetTable<Test>()
				 select new
				 {
					 t1 = t.PreNotification!.Value.TotalMilliseconds,
					 t2 = t.PreNotification!.Value.TotalSeconds
				 }).Where(x => x.t2 < x.t1).ToList();

the resulting SQL is like this:

            SELECT
	            [x].[PreNotification]
            FROM
	            [Common_Topology_Locations] [x]
            WHERE
	            [x].[PreNotification] < [x].[PreNotification]

so I think we need a way for each Dataprovider to extract the TimeSpan Properties, in the way how a timespan is stored on this Provider. In SQLServer the default type is "Time" wich does not make much sense with the property access, so I think we don't need to support this.

Maybe someone could lead me in the direction where to implement this?

jogibear9988:
## Summary

Adds built-in SQL translation for `TimeSpan` members and date/time interval operations, including native database intervals and numeric duration columns with explicitly declared storage units.

This covers the scenarios from #3993, #4308, #2950, and #3840:

- mapped `TimeSpan` columns
- intervals produced by subtracting database date/time values
- positive, negative, nullable, subsecond, and long-running intervals
- native interval columns
- numeric duration columns stored as ticks, seconds, milliseconds, or other units
- arithmetic and comparisons between differently stored durations
- parameters, projections, subqueries, CTEs, and set operations

The implementation uses linq2db's existing translation registration and provider conversion pipelines. It does not introduce `Sql.TimeSpan.cs` or require a global `AddTimeSpanMappings()` opt-in.

## Duration mapping

A duration column can declare its storage semantics using an attribute:

    [Column(DataType = DataType.Int64)]
    [Duration(DurationUnit.Second)]
    public TimeSpan Timeout { get; set; }

Or fluent mapping:

    mappingBuilder
        .Entity<Event>()
        .Property(e => e.Timeout)
            .HasDataType(DataType.Int64)
            .HasDuration(DurationUnit.Second);

Supported units are:

- `Native`
- `Tick`
- `Nanosecond`
- `Microsecond`
- `Millisecond`
- `Second`
- `Minute`
- `Hour`
- `Day`

The declaration does not change the database type, value converter, or existing default `TimeSpan` mapping.

`Native` requires a native interval/time database type. Explicit units require numeric storage. Incompatible mappings fail during descriptor construction.

An undeclared numeric `TimeSpan` column is not assumed to contain ticks or seconds. Duration operations on such a column remain unsupported instead of silently producing incorrect results.

## Duration normalization

Duration operations use CLR ticks as their canonical calculation representation.

This allows, for example:

- comparing a native interval column with a `BIGINT` seconds column
- adding tick and second duration columns
- comparing a stored duration with `endDate - startDate`
- using native and numeric durations in date addition
- accessing members after projections, CTEs, and set operations

Native interval columns are not globally converted to ticks. Direct materialization continues to use the provider's native interval mapping. Conversion is generated only when an operation needs canonical duration arithmetic.

A semantic `SqlDurationExpression` preserves the unit and duration origin through:

- query optimization
- projection nesting
- joins
- nullability processing
- CTEs
- `UNION`, `INTERSECT`, and `EXCEPT`
- LinqService serialization

Every set-operation arm is normalized independently, so a tick column and a seconds column cannot accidentally inherit one another's unit.

Native interval lowering occurs in provider conversion after the query shape has been finalized.

## Supported `TimeSpan` members

The following members translate server-side:

- `Days`
- `Hours`
- `Minutes`
- `Seconds`
- `Milliseconds`
- `Microseconds`
- `Nanoseconds`
- `Ticks`
- `TotalDays`
- `TotalHours`
- `TotalMinutes`
- `TotalSeconds`
- `TotalMilliseconds`
- `TotalMicroseconds`
- `TotalNanoseconds`

`Total*` members return the same total-duration meaning as CLR. They are not implemented using date-part component extraction.

Negative component extraction uses explicit truncate-toward-zero arithmetic and does not depend on provider-specific division or modulo behavior.

Nanosecond components are calculated from the tick remainder, avoiding overflow from multiplying the complete tick value by 100.

## Supported operations

The following nullable and non-nullable operations translate:

- `TimeSpan + TimeSpan`
- `TimeSpan - TimeSpan`
- unary `-TimeSpan`
- duration equality and ordering comparisons
- `DateTime ± TimeSpan`
- `DateTimeOffset ± TimeSpan`
- `DateTime - DateTime`
- `DateTimeOffset - DateTimeOffset`

Automatic date subtraction uses elapsed-time semantics matching CLR. It is separate from the boundary-counting contract of `Sql.DateDiffLong`.

Arithmetic respects the mapped database type's actual precision. Translation is rejected when the stored representation cannot preserve the requested semantics rather than silently discarding a remainder.

## DateTimeOffset safety

`DateTimeOffset` arithmetic is enabled only when the provider and mapped SQL type preserve the represented instant.

Supported configurations include:

- SQL Server 2008+ `datetimeoffset`
- PostgreSQL timestamp-with-time-zone
- Oracle timestamp-with-time-zone
- DuckDB `TIMESTAMPTZ`
- SQLite offset-aware storage
- Firebird 4+ timestamp-with-time-zone
- ClickHouse mappings normalized to UTC

Offset-losing mappings remain unsupported, including Access, DB2, Informix, SAP HANA, Sybase, MySQL/MariaDB, YDB, SQL CE, and older Firebird configurations.

This check uses the actual mapped operand types, so explicitly mapping one operand to an offset-losing type is also rejected.

## Legacy rewrite removal

The hardcoded `(DateTime - DateTime).Total*` rewrite was removed from `ExposeExpressionVisitor`.

That rewrite:

- supported only a subset of `Total*` members
- intercepted expressions before provider member translation
- converted elapsed-time operations to `Sql.DateDiff`
- could therefore apply date-boundary semantics where CLR requires elapsed-duration semantics

Database-produced `TimeSpan` values now use the same registered translation path as mapped duration columns.

## Long date APIs

Adds:

- `Sql.DatePartLong`
- `Sql.DateDiffLong`
- `Sql.DateParts.Microsecond`
- `Sql.DateParts.Nanosecond`
- `Sql.DateParts.Tick`

`DatePartLong` and `DateDiffLong` use dedicated builders. Existing `Sql.DatePart`, `Sql.DateAdd`, and `Sql.DateDiff` retain their existing builders and compatibility behavior.

The new component contracts are:

- `Microsecond`: `0..999999` within the current second
- `Nanosecond`: `0..999999900`, in 100-nanosecond increments
- `Tick`: `0..9999999` within the current second
- one tick is 100 nanoseconds

Nanosecond addition truncates toward zero when converted to ticks:

- `99` nanoseconds → `0` ticks
- `100` nanoseconds → `1` tick
- `-99` nanoseconds → `0` ticks
- `-101` nanoseconds → `-1` tick

## `DateDiffLong` contract

`Sql.DateDiffLong` provides provider-independent boundary counting:

- calendar boundaries for year, quarter, month, week, and day
- a fixed Sunday boundary for weeks
- timeline boundaries for hour through tick
- UTC-normalized operands for `DateTimeOffset`

Nanosecond overflow is checked and fails instead of wrapping.

The existing 32-bit `Sql.DateDiff` APIs retain their previous SQL behavior and use checked narrowing during client-side evaluation.

## Provider handling

Provider-specific duration lowering is implemented for:

- Access
- ClickHouse
- DB2
- DuckDB
- Firebird
- Informix
- MariaDB
- MySQL
- Oracle
- PostgreSQL
- SAP HANA
- SQLite
- SQL CE
- SQL Server
- Sybase
- YDB

Notable handling includes:

- bounded SQL Server day/sub-day composition
- `DATEDIFF_BIG` where available and exact composed fallbacks where required
- separate SQL Server handling for `smalldatetime`, `datetime`, `datetime2`, and `datetimeoffset`
- UTC normalization for SQL Server and Oracle offset values
- native PostgreSQL, Oracle, Informix, DuckDB, Firebird, MySQL, and YDB interval lowering
- MySQL/MariaDB `TIME` range handling
- integer timestamp decomposition for SQLite
- Access OLE Automation day/sub-day arithmetic
- widened 64-bit arithmetic for large intervals
- projection-scoped Access and Informix result conversion

Provider-specific native interval conversion was moved out of the early member translators. Direct native interval projection remains unchanged; conversion to canonical ticks is applied only when required by duration operations.

## Compatibility and existing-test audit

| Existing test | Previous expectation | Why it changed | Breaking? | New contract | Providers and replacement coverage |
|---|---|---|---|---|---|
| `DateTimeOffsetAddTimeSpan` | YDB participated despite losing the stored offset | YDB cannot preserve offset-sensitive CLR semantics | No; unsupported is safer than an incorrect instant | Offset arithmetic requires an instant-preserving mapping | YDB is excluded; explicit rejection is covered by `DateTimeOffsetArithmeticIsRejectedForOffsetLosingMappedType` |
| `NullableDateTimeSubtractionProjectionTest` | Accepted a broad `1.9..2.1` hour result and asserted subtraction was absent from SQL | Date subtraction is now translated as elapsed server-side arithmetic | Execution location changes, result contract does not | Exact nullable elapsed duration at mapped precision | Access ODBC and ClickHouse; strengthened by the signed, nullable, precision, and range matrices |
| `NullableDateTimeSubtractionProjectionSqlTest` | Expected translation failure | These providers now support server-side nullable subtraction | Additive support | The SQL projection returns the exact duration and preserves nulls | Access ODBC and ClickHouse; the test now performs readback assertions |
| `NullableDateTimeOffsetSubtractionProjectionTest` | Used a broad tolerance and required client-side evaluation | Instant-preserving providers now support UTC-correct server-side subtraction | Execution location changes, result contract does not | Exact elapsed duration for supported mappings; offset-losing mappings remain unsupported | Supported `DateTimeOffset` providers; separate matrices cover equal instants with different offsets and equal local clocks with different offsets |

All other additions are new regression coverage. Existing legacy date APIs are not routed through the new long builders.

## Tests

The complete `Issue3993Tests` fixture from #5334 was ported without its global mapping opt-in and expanded with readback assertions.

Coverage includes:

- every component and `Total*` member
- positive, negative, zero, and nullable durations
- signed tick boundary values
- 25-, 30-, and 365-day ranges
- multi-decade ranges where supported
- tick, microsecond, and nanosecond precision
- constants, parameters, columns, captured values, and mixed expressions
- all nullable operand combinations
- database-produced durations
- direct projections and comparisons
- nested projections, subqueries, CTEs, and set operations
- native interval and numeric-unit composition
- undeclared numeric duration rejection
- invalid native/numeric mapping combinations
- unchanged default `TimeSpan` mapping
- equal `DateTimeOffset` instants with different offsets
- equal local clocks with different offsets
- `DateDiffLong` boundaries, large ranges, nullability, result types, and overflow
- integer-tick, decimal-tick, and native-interval bulk copy with readback
- provider-specific SQL shape and 64-bit intermediate assertions

Expected automatic arithmetic results are derived from values read back through the mapped database type. Precision comparisons use at most one mapped storage quantum rather than broad second/hour tolerances.

## Local validation

- Release `net10.0` build: passed with zero warnings/errors
- Release `netstandard2.0` build: passed with zero warnings/errors
- `Issue4308Tests` on SQLite and DuckDB: 34/34 passed
- `Test3993Tests` on SQLite and DuckDB: 29/29 passed
- `git diff --check`: passed

The complete Azure provider and alternate-driver matrix remains the final runtime acceptance gate.

Fixes #2950  
Fixes #4308

Also covers #3993 and #3840.

Related to #4306 and #5334.


jogibear9988:
Open Questions:

- Should we change SQL Server’s default TimeSpan mapping from time to bigint? (issue #4306) (mapping a TimeSpan to a Time is clearly wrong)

igor-tkachev (changes requested):
## Review

The overall architectural direction is much better than the previous implementation from #5334.

Using the existing member, unary, and binary translation registration pipeline is the correct approach. In particular, the PR avoids introducing a parallel translation infrastructure and avoids process-wide mutable mappings such as `AddTimeSpanMappings()`. Moving provider-independent normalization into `DateFunctionsTranslatorBase` and provider-specific lowering into individual translators is also the right design.

Removing the special-case rewrite from `ExposeExpressionVisitor` is directionally correct as well. The old rewrite of:

```csharp
(end - start).TotalHours
```

into:

```csharp
Sql.DateDiff(Sql.DateParts.Hour, start, end)
```

did not preserve C# semantics because `DateDiff` can count crossed date-part boundaries rather than elapsed duration.

For example:

```csharp
var start = new DateTime(2026, 1, 1, 10, 59, 0);
var end   = new DateTime(2026, 1, 1, 11, 1, 0);
```

C# produces:

```csharp
(end - start).TotalHours == 2.0 / 60.0
```

while a boundary-based `DateDiff(hour, ...)` can return `1`.

However, the replacement implementation must fully preserve C# semantics. It is not sufficient to replace one approximation with another.

There are several fundamental issues that must be addressed before this can be merged.

---

# Fundamental requirements

## MAJ01 — C# semantics must remain unchanged

This is one of the primary principles of linq2db.

When a user writes:

```csharp
left - right
date + interval
date - interval
timeSpan.Hours
timeSpan.TotalHours
-timeSpan
```

the generated SQL must be semantically equivalent to the corresponding C# operation, within the precision and range actually supported by the mapped database type.

The complexity or size of the generated SQL is secondary. If preserving C# semantics requires a large composed SQL expression, linq2db should generate that expression.

The translator must not silently:

* reduce the valid range;
* lose precision that the mapped database type can represent;
* change truncation or modulo rules;
* replace elapsed-duration semantics with date-part boundary semantics;
* use local wall-clock subtraction for `DateTimeOffset`;
* return an approximate result where C# defines an exact one.

If a provider cannot implement the operation correctly, the translator should return `null` and leave the operation unsupported. It must not generate SQL with silently different semantics.

### SQL Server before 2016: artificial range limitation

The current implementation produces:

```sql
CAST(DATEDIFF(millisecond, right, left) AS BIGINT) * 10000
```

The cast does not make the calculation 64-bit.

`DATEDIFF(millisecond, ...)` is evaluated first and returns `int`. It overflows for an interval of approximately 24 days and 20 hours.

Consequently, this valid C# expression:

```csharp
var duration = endDate - startDate;
```

will fail for an interval of one month on SQL Server 2005–2014.

This is a direct violation of C# semantics and a compatibility regression. A query that could previously be materialized and evaluated in .NET can now fail on the server.

The implementation must use a composed calculation that keeps every intermediate `DATEDIFF` within a safe range.

For example, the calculation can be decomposed into:

```text
full days
+ hours within the remaining day
+ minutes within the remaining hour
+ seconds within the remaining minute
+ the finest fractional component supported by the mapped type
```

Another possible scheme is:

```sql
days = DATEDIFF(day, right, left)

anchor = DATEADD(day, days, right)

days =
    CASE
        WHEN left >= right AND anchor > left THEN days - 1
        WHEN left < right AND anchor < left THEN days + 1
        ELSE days
    END

anchor = DATEADD(day, days, right)

remaining =
      CAST(DATEDIFF(second, anchor, left) AS bigint) * 10000000
    + fractional_ticks(left)
    - fractional_ticks(anchor)

result =
    CAST(days AS bigint) * 864000000000
    + remaining
```

The exact SQL can differ, but the result must preserve C# semantics and must not introduce an artificial 24.8-day limit.

### SQL Server before 2016: translator-induced precision loss

The same expression:

```sql
CAST(DATEDIFF(millisecond, right, left) AS BIGINT) * 10000
```

drops the entire sub-millisecond part.

For example:

```csharp
var start  = new DateTime(2026, 1, 1);
var result = start.AddTicks(9999) - start;
```

C# returns:

```csharp
result.Ticks == 9999
```

The generated SQL returns zero.

For `datetime2`, this is not a limitation of the database type. `datetime2` can represent 100-nanosecond units. The precision is being lost by the translator.

The translation should preserve all precision available in the actual mapped database type:

* `datetime2` and `datetimeoffset`: up to 100 ns;
* `datetime`: its native approximately 1/300-second precision;
* `smalldatetime`: minute precision.

The translator must not impose a coarser precision than the storage type.

### `DateTimeOffset` subtraction

C# subtracts `DateTimeOffset` values as absolute instants.

These two values represent the same instant:

```csharp
var left  = new DateTimeOffset(
    2026, 1, 1, 12, 0, 0,
    TimeSpan.FromHours(2));

var right = new DateTimeOffset(
    2026, 1, 1, 10, 0, 0,
    TimeSpan.Zero);
```

Therefore:

```csharp
left - right == TimeSpan.Zero
```

The implementation must not subtract local wall-clock components and return two hours.

A second test is also required:

```csharp
var left  = new DateTimeOffset(
    2026, 1, 1, 12, 0, 0,
    TimeSpan.FromHours(2));

var right = new DateTimeOffset(
    2026, 1, 1, 12, 0, 0,
    TimeSpan.Zero);
```

These values have the same local time but represent different instants, so their difference must not be zero.

Every provider that claims `DateTimeOffset` subtraction support must pass both cases.

Where native database arithmetic does not preserve offset semantics, operands must be normalized to UTC before subtraction.

### Negative `TimeSpan` components

C# uses truncation toward zero for integer division.

For example:

```csharp
var value = TimeSpan.FromHours(-25);

value.Days  == -1;
value.Hours == -1;
```

Database division, `MOD`, `%`, `FLOOR`, and remainder rules differ between providers.

The translator cannot assume that:

```sql
value / divisor
value % divisor
```

have C# semantics for negative values.

The portable semantic definition should be:

```text
quotient  = truncate_toward_zero(value / divisor)
remainder = value - quotient * divisor
```

For a provider where division does not truncate toward zero, this can be expressed as:

```sql
CASE
    WHEN value >= 0
        THEN FLOOR(value / divisor)
    ELSE CEILING(value / divisor)
END
```

and:

```sql
value - quotient * divisor
```

The complete negative-component matrix must be tested for every implementation strategy.

### `DateTime ± TimeSpan` fallback

The base fallback currently attempts progressively coarser units:

```text
Tick
Microsecond
Millisecond
Second
```

If a provider does not support a finer unit, the remaining fraction can be silently discarded.

That is only acceptable when the mapped database type itself cannot represent the finer value.

The decision must be based on the actual mapped date type and its precision, not merely on whether a particular translator method returned `null`.

If the mapped type can store milliseconds, the translator cannot fall back to seconds and discard the fraction.

If correct translation at the required precision is impossible, the operation must remain unsupported.

---

## MAJ02 — `Sql.*` methods must have identical semantics across all providers

Explicit `Sql.*` methods are a separate API contract from automatic translation of C# expressions.

For example:

```csharp
Sql.DateDiff(...)
Sql.DateDiffLong(...)
Sql.DatePart(...)
Sql.DatePartLong(...)
Sql.DateAdd(...)
```

must each have a clearly defined semantic contract, and that contract must be the same for every provider without exception.

A provider implementation cannot interpret `DateDiff` as boundary counting while another interprets it as elapsed duration.

The documentation describes `DateDiff`/`DateDiffLong` as counting requested boundaries.

However, some provider implementations appear to use elapsed-duration expressions such as:

```sql
EXTRACT(EPOCH FROM end - start)
```

Boundary-counting and elapsed-duration semantics are different.

For example:

```text
10:59 -> 11:01
```

has:

```text
elapsed duration: 2 minutes
hour boundaries crossed: 1
```

If the intended contract is boundary counting, every provider must implement boundary counting.

If the intended contract is elapsed duration, the documentation and existing behavior must be reconciled separately.

Provider-specific interpretations are not acceptable.

If a method is documented as returning a 64-bit result, its provider implementation must actually perform a 64-bit calculation. Merely assigning a `long` SQL metadata type to an expression that internally calls a 32-bit database function does not satisfy the contract.

---

## MAJ03 — Existing `Sql.*` methods must preserve their old behavior

Compatibility is a hard requirement.

Existing public methods must preserve their previous semantics, SQL translation, result types, range, precision, and error behavior.

The intended structure should be:

```text
Sql.DatePart     — keep unchanged
Sql.DateAdd      — keep unchanged
Sql.DateDiff     — keep unchanged
Sql.DatePartLong — add separately
Sql.DateDiffLong — add separately
```

There is no reason to rewrite:

```csharp
Sql.DateDiff(...)
```

through:

```csharp
Sql.DateDiffLong(...)
```

and then narrow the result.

The PR currently uses an implementation equivalent to:

```csharp
public static int? DateDiff(...)
{
    var result = DateDiffLong(...);

    return result == null
        ? null
        : checked((int)result.Value);
}
```

This is unnecessary and changes the implementation and potentially the behavior of an existing public method.

The existing local implementation, SQL mappings, provider builders, result types, overflow behavior, and tests for `Sql.DateDiff` should remain unchanged.

The new method can duplicate the small switch. Compatibility is more important than eliminating a few repeated lines.

The same principle applies to:

```csharp
Sql.DatePart(...)
Sql.DatePartLong(...)
```

There is no need to implement the existing method by calling the new method and narrowing the result.

These are public APIs with different return types and potentially different range contracts.

Separate builders are preferable:

```text
DateDiffBuilder
DateDiffLongBuilder
```

They may share internal helpers for date-part names and safe SQL fragments, but they should have separate public contracts and result-type handling.

Do not infer the contract from:

```csharp
builder.Member.Name
```

inside one mixed builder when two explicit builders make the behavior clearer and safer.

If compatibility requires duplicate implementations for `DateDiff` and `DateDiffLong`, duplicate implementations should be used. In this case duplication is also simpler because `DateDiff` can remain completely untouched.

---

## MAJ04 — Changes to existing tests must be explicitly explained

An existing test defines existing behavior.

Any test that is:

* removed;
* weakened;
* changed from runtime execution to SQL generation;
* changed from exact comparison to a broad tolerance;
* changed by removing an assertion;
* changed by removing `[ActiveIssue]`;
* changed by removing `[ThrowsCannotBeConverted]`;
* changed by expanding or reducing provider lists;

must be explicitly explained in the PR description.

The explanation must state:

1. what behavior the test previously protected;
2. why that behavior is being changed;
3. whether the change is breaking;
4. what the new contract is;
5. which providers are affected;
6. which new tests define and verify the replacement behavior.

A test should not simply be modified to match the new implementation.

Several existing tests are changed in a way that alters or removes previously asserted behavior.

For example, assertions were removed that explicitly verified that date subtraction did not appear in SQL because it was evaluated client-side.

Changing this execution path may be correct, but it is an observable behavior change and must be documented.

The PR description should explain:

* why server-side execution now preserves C# semantics;
* how every provider implementation has been validated;
* what precision differences are expected from mapped types;
* why this does not introduce new runtime failures;
* which old client-side scenarios are now server-side.

Likewise, removing `[ActiveIssue]`, removing `[ThrowsCannotBeConverted]`, changing column data types, changing provider lists, or replacing exact assertions with broad ranges must be individually justified.

---

# `DateDiffLong` implementation

## MAJ05 — `DateDiffLong` is not consistently implemented as a 64-bit API

For SQL Server 2005–2014, the new method still maps to ordinary:

```sql
DATEDIFF(...)
```

not `DATEDIFF_BIG`.

Changing the result metadata to `long` does not change the return type or range of the SQL function.

For example:

```csharp
Sql.DateDiffLong(
    Sql.DateParts.Millisecond,
    start,
    end)
```

still overflows after approximately 24.8 days on those versions.

This violates the name and documentation of the new API.

The implementation must either:

* use a composed 64-bit expression;
* leave the method unsupported for those providers;
* or define a different explicit contract.

It cannot present a 32-bit calculation as a 64-bit method.

### Hardcoded `int` result types

Some provider branches still construct expressions using:

```csharp
GetDbDataType(typeof(int))
```

even for `DateDiffLong`.

For example, the PostgreSQL millisecond branch appears to retain an `int` result type while other newly added branches use the dynamic result type.

All provider branches must be audited.

`DateDiffLong` must produce a true `long` SQL expression from the first intermediate operation through the final result.

### Nanosecond overflow

The local implementation uses:

```csharp
difference.Ticks * 100
```

A `long` nanosecond result covers a smaller range than `TimeSpan`.

The multiplication must be checked:

```csharp
checked(difference.Ticks * 100)
```

The supported range must be documented and tested.

A method introduced to avoid overflow cannot silently wrap on overflow.

---

# Access-specific implementation

## MAJ06 — Access translation contains incorrect mappings and global side effects

### Incorrect millisecond mapping

Access `DatePart("s", ...)` returns seconds, not milliseconds.

Mapping:

```csharp
Sql.DateParts.Millisecond -> "s"
```

silently returns the wrong value.

An unsupported translation is preferable to an incorrect translation.

The same mapping must be checked in both the date-part and date-add paths.

### Millisecond date addition path

There is logic that adjusts a millisecond increment by dividing by 1000, but the part-selection switch does not appear to return a value for `Millisecond`.

That makes the branch unreachable.

In addition, division using the original integer type can turn:

```text
500 ms
```

into:

```text
0 seconds
```

through integer division.

This must be corrected and covered by runtime tests.

### Global mapping-schema conversions

The PR adds Access mapping-schema conversions such as:

```text
decimal -> TimeSpan
double  -> TimeSpan
double  -> DateTime
```

In particular:

```csharp
(double value) => DateTime.FromOADate(value)
```

is a broad mapping-schema rule, not a conversion scoped to one generated expression.

This can affect existing Access projections and conversion selection outside this feature.

These mappings require regression tests demonstrating that existing `double`, `decimal`, `DateTime`, and `TimeSpan` materialization behavior remains unchanged for unrelated queries.

A feature-specific result conversion should not introduce a global conversion side effect.

---

# Test coverage

## MAJ07 — Current tests do not prove runtime semantic correctness

The test suite is broad in terms of provider lists and scenario count, but it is not sufficient to prove semantic correctness.

Many tests verify only that SQL can be generated:

```csharp
query.ToSqlQuery()
```

This proves that the translator returned an expression and the SQL builder did not throw.

It does not prove:

* correct sign;
* correct operand order;
* correct units;
* correct range;
* correct precision;
* correct rounding;
* correct negative modulo behavior;
* correct nullable semantics;
* correct result materialization;
* correct provider-version support;
* correct `DateTimeOffset` instant semantics.

A provider should not be considered covered merely because it appears in an `IncludeDataSources` list for a SQL-generation smoke test.

The following runtime matrices are needed.

### Boundary matrix

The following values should be tested:

```text
TimeSpan.Zero
1 tick
-1 tick
9 ticks
-9 ticks
10 ticks
-10 ticks
9999 ticks
-9999 ticks
10000 ticks
-10000 ticks
TicksPerSecond - 1
-(TicksPerSecond - 1)
TicksPerDay - 1
-(TicksPerDay - 1)
TicksPerDay
-TicksPerDay
TicksPerDay + 1
-(TicksPerDay + 1)
```

These cases expose errors in:

* truncation;
* modulo;
* sign handling;
* unit conversion;
* boundary handling.

### Long-range matrix

The following intervals must execute on the server:

```text
25 days
30 days
365 days
multiple decades where supported
```

These cases are required to detect `DATEDIFF(millisecond)` and similar intermediate overflow.

A SQL-shape test is not sufficient.

### Precision matrix

Use exact fractional values:

```text
1 tick
9 ticks
10 ticks
9999 ticks
1234567 ticks
1.234567 seconds
```

Assertions should use the precision of the actual mapped database type.

A broad tolerance such as:

```csharp
expected ± 1 second
```

is too weak. It can hide:

* lost milliseconds;
* incorrect rounding;
* wrong coefficients;
* truncation errors.

### Negative-component matrix

For positive and negative values, test every member:

```text
Days
Hours
Minutes
Seconds
Milliseconds
Microseconds
Nanoseconds
Ticks
TotalDays
TotalHours
TotalMinutes
TotalSeconds
TotalMilliseconds
TotalMicroseconds
TotalNanoseconds
```

A representative negative value should include multiple components, for example:

```csharp
new TimeSpan(-1, -2, -3, -4, -5)
    .Add(TimeSpan.FromTicks(-67))
```

The SQL result must match the C# result member by member.

### Nullable operator matrix

All registered combinations should be executed:

```text
DateTime     + TimeSpan
DateTime     + TimeSpan?
DateTime?    + TimeSpan
DateTime?    + TimeSpan?

DateTime     - TimeSpan
DateTime     - TimeSpan?
DateTime?    - TimeSpan
DateTime?    - TimeSpan?

DateTime     - DateTime
DateTime     - DateTime?
DateTime?    - DateTime
DateTime?    - DateTime?
```

The same is required for `DateTimeOffset`.

Each combination must include both non-null and null operand cases.

Registration coverage is not runtime semantic coverage.

### Expression-shape matrix

There is special handling for SQL parameters, so these forms must be tested:

```text
column + captured parameter
column + inline constant
column - captured parameter
column - inline constant
column - column
captured date + column interval
```

Constants can become `SqlValue` or `SqlParameter` depending on options and query context. Both paths must be verified.

### `DateTimeOffset` matrix

At minimum:

```text
same UTC instant, different offsets -> zero
same local time, different offsets -> non-zero
positive interval
negative interval
nullable operands
```

### `DateDiffLong` runtime matrix

For every supported provider, execute and compare:

```csharp
Sql.DateDiffLong(DateParts.Microsecond, ...)
Sql.DateDiffLong(DateParts.Nanosecond, ...)
Sql.DateDiffLong(DateParts.Tick, ...)
```

Test:

```text
positive difference
negative difference
zero
cross-second boundary
cross-day boundary
large interval
overflow boundary
```

Also verify the actual SQL result type and materialized CLR type.

### Bulk copy round-trip

The current bulk-copy test mainly verifies that bulk copy does not throw.

It should:

1. write a `TimeSpan`;
2. read it back;
3. compare it with the original value;
4. use its members in a server-side query;
5. verify nullable and non-nullable columns.

This should cover each storage strategy:

```text
Int64 ticks
Decimal ticks for Access
native PostgreSQL interval
native Oracle interval
native Informix interval
MySQL/MariaDB representation
```

---

# Public API contracts

## MIN01 — `DateParts.Tick` needs an exact public contract

The new value is used for several different operations:

```csharp
Sql.DatePart(DateParts.Tick, date)
Sql.DateAdd(DateParts.Tick, value, date)
Sql.DateDiff(DateParts.Tick, start, end)
```

The intended semantics appear to be:

* `DatePart(Tick)` — 100-nanosecond component within the current second;
* `DateAdd(Tick)` — add 100-nanosecond units;
* `DateDiff(Tick)` — difference expressed in 100-nanosecond units.

These semantics must be explicitly documented for every overload.

Boundary tests around one second are required, including negative values.

The name is also potentially confusing because `TimeSpan.Ticks` represents the full interval, while `DatePart(Tick)` returns a component within a second.

The public name should be reviewed before release because it cannot be changed later without another compatibility issue.

## MIN02 — Nanosecond addition must document truncation to `DateTime` precision

The local implementation effectively performs:

```csharp
date.AddTicks(number / 100)
```

Therefore:

```csharp
AddNanoseconds(99)  // adds zero ticks
AddNanoseconds(100) // adds one tick
```

This is consistent with `DateTime` precision, but it must be documented and tested, especially for negative values:

```csharp
-99  / 100 == 0
-101 / 100 == -1
```

---

# Compatibility assessment

The PR is mostly source- and binary-compatible at the signature level:

* existing methods are not removed;
* new enum values are appended;
* new methods are additive.

However, behavioral compatibility is not currently preserved.

Potential or confirmed behavioral changes include:

* automatic C# date subtraction moving from client-side to server-side;
* changed precision;
* changed rounding;
* new server-side overflow;
* different `DateTimeOffset` handling;
* changed overflow behavior of existing `Sql.DateDiff`;
* changed SQL and execution plans;
* broad Access conversion rules affecting unrelated queries.

Some behavioral changes may be justified bug fixes, especially replacing boundary-counting semantics for C# `TimeSpan.Total*` expressions.

But every such change must be explicit, deliberate, documented, and covered by tests.

---

# Recommended implementation structure

Keep these methods and their existing provider mappings untouched:

```csharp
Sql.DatePart
Sql.DateAdd
Sql.DateDiff
```

Add separately:

```csharp
Sql.DatePartLong
Sql.DateDiffLong
```

For C# expression translation and `DateDiffLong`, use arbitrarily complex SQL where necessary.

The priority order should be:

1. C# or explicit `Sql.*` semantics;
2. correct range;
3. precision of the mapped database type;
4. provider and version compatibility;
5. SQL compactness.

Compact SQL is not a correctness requirement.

---

# Conclusion

The architectural direction is good and should be preserved.

However, this PR currently combines:

* automatic C# `TimeSpan` translation;
* date/time subtraction;
* provider-native interval handling;
* new `Sql.DatePartLong`;
* new `Sql.DateDiffLong`;
* new public date parts;
* changes to existing `Sql.DatePart` and `Sql.DateDiff`;
* changes to existing test contracts.

That makes correctness and compatibility difficult to verify.

At minimum:

1. existing `Sql.DatePart`, `Sql.DateAdd`, and `Sql.DateDiff` behavior and tests should be restored unchanged;
2. the new long-valued APIs should be implemented through separate code paths;
3. automatic translation must preserve C# semantics within the precision of the mapped database type;
4. explicit `Sql.*` semantics must be identical across all providers;
5. runtime semantic tests must cover every provider that claims support;
6. every removed, weakened, or modified existing test must be explained.

Until those requirements are met, I do not think the PR is safe to merge.

jogibear9988:
@igor-tkachev 

following is now changed:

| Review point | Resolution |
|---|---|
| **MAJ01: C# semantics** | Automatic subtraction now produces elapsed time, not boundary counts. Translation uses the operands’ actual mapped type, storage precision, interval range, and offset-preservation capability. Unsupported exact translations are rejected instead of losing precision. |
| **Large SQL Server intervals** | Pre-2016 translations use bounded day/sub-day decomposition instead of an overflowing `DATEDIFF(ms)`. Newer versions use `DATEDIFF_BIG` where appropriate. All intermediates are 64-bit. |
| **Mapped precision** | `smalldatetime`, `datetime`, `datetime2(p)`, `datetimeoffset(p)`, native intervals, and tick-based mappings are handled according to their actual precision. |
| **`DateTimeOffset`** | Operands are normalized to UTC on offset-preserving mappings. Equal instants with different offsets produce zero. Offset-sensitive arithmetic is rejected for offset-losing mappings. |
| **Negative `TimeSpan` members** | Components and totals use explicit truncate-toward-zero quotient/remainder formulas, independent of provider-specific negative division/modulo behavior. |
| **MAJ02: Explicit SQL APIs** | Automatic subtraction remains elapsed arithmetic. `Sql.DateDiffLong` has a provider-independent boundary-counting contract with calendar boundaries for year through day and timeline boundaries for hour through tick. |
| **MAJ03: Legacy compatibility** | Existing `Sql.DatePart`, `Sql.DateAdd`, and `Sql.DateDiff` behavior was restored. The long APIs have dedicated builders and no longer select behavior through `builder.Member.Name`. |
| **MAJ04: Test changes** | All removed or weakened existing tests were restored. The final test diff against `master` is additive: the complete `Issue3993Tests`, a new provider/runtime matrix, and additional date-function tests. No broad second/hour tolerances remain. |
| **MAJ05: `DateDiffLong`** | Uses 64-bit boundary calculations, including on old SQL Server versions. Nanosecond overflow is checked and fails instead of wrapping. |
| **MAJ06: Access** | Removed the incorrect millisecond-to-second mapping and broad numeric mapping conversions. Access now uses OADate day/sub-day decomposition with projection-scoped result conversion, so unrelated numeric materialization is unaffected. |
| **MAJ07: Runtime tests** | Added runtime coverage for tick boundaries, negative components, subsecond precision, long intervals, nullable operands, expression shapes, offsets, `DateDiffLong`, and bulk-copy readback. Expected values are based on values read back at the mapped storage precision. |
| **MIN01: `Tick` contract** | Documented as the component within the current second: `0..9999999`, with one tick equal to 100 ns. Microsecond and nanosecond component contracts are documented too. |
| **MIN02: Nanosecond addition** | Conversion truncates toward zero: `99 → 0`, `100 → 1`, `-99 → 0`, and `-101 → -1` ticks. |
| **Unsupported cases** | Translation returns “not translated” where exact semantics cannot be preserved, especially offset-sensitive operations on offset-losing mappings. |
| **`ExposeExpressionVisitor`** | The old special case remains removed. It incorrectly mapped elapsed `Total*` members to date-part/boundary operations. The replacement translates the complete interval first and then applies CLR-compatible `TimeSpan` member semantics. |
| **`Sql.TimeSpan.cs`** | Not needed: automatic operators and members belong to the existing translation registration system and do not require another public SQL façade. |

The provider matrix covers Access, ClickHouse, DB2, DuckDB, Firebird, Informix, MySQL/MariaDB, Oracle, PostgreSQL, SQLite, SAP HANA, SQL CE, SQL Server, Sybase, and YDB. Provider/type combinations that cannot preserve the required semantics are explicitly rejected rather than returning an approximate result.

Also open question:

Should we change SQL Server’s default TimeSpan mapping from time to bigint? (issue https://github.com/linq2db/linq2db/issues/4306) (mapping a TimeSpan to a Time is clearly wrong)

sdanyliv:
The main issue is that I’m waiting for V7, which should provide proper `TypeMapping` and handle this scenario natively, without workarounds such as `Sql.DatePartLong(...)` and `Sql.DateDiffLong(...)`.

In my opinion, `TimeSpan` support should be treated as a first-class feature, with its own AST node and proper optimization support. It should be designed carefully from the outset rather than implemented through a brute-force solution.

jogibear9988:
> The main issue is that I’m waiting for V7, which should provide proper `TypeMapping` and handle this scenario natively, without workarounds such as `Sql.DatePartLong(...)` and `Sql.DateDiffLong(...)`.
> 
> In my opinion, `TimeSpan` support should be treated as a first-class feature, with its own AST node and proper optimization support. It should be designed carefully from the outset rather than implemented through a brute-force solution.

@igor-tkachev @sdanyliv 

what does this mean for me now? I really need correct timespan support in our software, and i always have to maintain a fork, rewrite the implementation the whole time.

Can we get this into a v6.x sometime?
Or do I need to wait for a V7? (are there estimates when we will have this? will it then contain timespan support?)

At the moment we support some parts wrongly... I don't see where when this would be used make it worse? The proper(er?) implementation in V7 then still could follow.

sdanyliv:
I’m reviewing the PR now and will propose a broader plan separately. For the moment, please answer one simple question: **how do you intend to define the mapping between `TimeSpan` and a database column?**

Different users may store a duration in different representations - for example, as ticks, seconds, minutes, or hours in a numeric column. How will the mapping specify which unit and representation the column uses?

jogibear9988:
> I’m reviewing the PR now and will propose a broader plan separately. For the moment, please answer one simple question: **how do you intend to define the mapping between `TimeSpan` and a database column?**
> 
> Different users may store a duration in different representations - for example, as ticks, seconds, minutes, or hours in a numeric column. How will the mapping specify which unit and representation the column uses?

I can not configure this at all (at the moment). At the moment we use database interval type if it has one, or if mapped to long or the database has no interval type we use ticks.  Yeah this is something that could be added, but I think it is a additional function.

sdanyliv:
> I can not configure this at all (at the moment). […] Yeah this is something that could be added, but I think it is a additional function.

That's the crux, and I don't think it can be an additional function — the unit decides the shape of everything underneath it.

Right now it's a constant in the code: `if (expressionType.DataType != DataType.Int64) return TranslateNative…`, three times in `DateFunctionsTranslatorBase`. "Int64 means ticks" is therefore baked into all 16 provider implementations. If someone stores a duration as **seconds** in a `bigint` — a very common choice — every translation is silently wrong by a factor of 10,000,000, with no error. Adding the unit afterwards doesn't extend that code, it invalidates it: every provider lowering has to be rewritten to take the unit as an input.

Two more consequences of the same root:

- With the shipped defaults nothing translates — 9 of the 16 providers have no native override, and SQL Server maps `TimeSpan` to `time`. To switch the feature on you have to remap `TimeSpan` → `bigint` for the whole schema, which reinterprets every existing column. The `AddTimeSpanMappings()` opt-in didn't disappear, it moved into `AddScalarType`.
- It's also why the "should SQL Server's default become `bigint`?" question keeps resurfacing. That's not a loose end — the design forces it.

So we'll do this as a separate, clean implementation on a new branch, targeting 6.x. To be clear: **not** waiting for V7 — I checked, nothing here is actually blocked by it.

- a declarative duration mapping (`[Duration(DurationUnit.Second)]` / `.HasDuration(…)`) stating the unit per column, so precision and range become computable instead of assumed;
- interval AST nodes that keep the intent (`end - start` is elapsed, not a boundary count) through optimization and are lowered per provider afterwards — the same mechanism `SqlCastExpression` already uses;
- no default-mapping flips: a `TimeSpan` column not declared as a duration keeps exactly its current meaning.

This supersedes #5739 as an implementation, but not as work. We're taking from it the test scenarios, the provider-specific knowledge (which databases have native intervals, precision quanta, where `DATEDIFF` overflows), and @igor-tkachev's review as the acceptance criteria. That research is the expensive part and it stays.

Separately: `Sql.DatePartLong` / `Sql.DateDiffLong` and the new date parts are an independent feature with their own contract. Split out on their own they're reviewable and could land in 6.x much sooner.

jogibear9988:
> That's the crux, and I don't think it can be an additional function — the unit decides the shape of everything underneath it.

Why? We don't support it now. We also don't support it when this is merged. We then support mapping TimeSpan to and Int64 as Ticks. If we add another mode, it will not break anything. I don't get why we need this immediately.

But for me it is also okay, if we add this.
But I think we should convert every duration inside of the database still to Ticks, so calculated durations can be compared to the ones in columns. 

> no default-mapping flips: a TimeSpan column not declared as a duration keeps exactly its current meaning.

=> what is the current meaning? it is mapped to time, wich is wrong so I would say we should break this. I think default should be our default timespan type, wich should be Interval on supporting databases and ticks on other ones.

> Separately: Sql.DatePartLong / Sql.DateDiffLong and the new date parts are an independent feature with their own contract. Split out on their own they're reviewable and could land in 6.x much sooner.

this does not help me, I need the Tiemspan Column calculation support, cause we use it in our software

sdanyliv:
Fair push-back, and let me correct my own wording: adding units later would not *break users*. What it costs is the provider layer — the base tick arithmetic plus the seven native-interval overrides are all written against "Int64 means ticks", so a unit dimension added afterwards means going through all of them again. That's a rework argument, not a correctness one, and I should have said it that way.

The correctness argument is narrower but stands: with the unit implicit, a `bigint` column holding **seconds** translates to silently wrong SQL — not an error, just wrong numbers. I'd rather that case be "not supported" than "supported and wrong".

> I think we should convert every duration inside of the database still to Ticks, so calculated durations can be compared to the ones in columns

Agreed, and that is exactly what a declared unit buys you: normalize to one canonical representation at the boundary and everything composes afterwards. Without the declaration there is nothing to normalize *from*.

> what is the current meaning? it is mapped to time, wich is wrong so I would say we should break this

That is a real proposal, and it needs @igor-tkachev's call, because it is a breaking change: models that use `TimeSpan` as a time of day against SQL Server `time` work today and would silently change meaning. My position is that the default flip belongs in v7, where breaking changes are allowed — while the mapping and the translation can ship in 6.x behind an explicit declaration. That way you get the feature without waiting for a major release, and the default question stays open instead of being decided by this PR.

> this does not help me, I need the Timespan Column calculation support

Understood — then forget the split, we will take `Sql.DatePartLong` / `Sql.DateDiffLong` and the new date parts into the same work. You do not need to carve anything out of this PR. The interval column arithmetic is the part you actually need, and it is what is being built now.

sdanyliv:
Good move, but actually I've already added the at least four new AST nodes for supporting interval arithmetic. It is too raw to create PR right now. Testing and improving.

igor-tkachev (commented):
This PR has exposed a real architectural limitation in linq2db's type system.

The current type information is not always sufficient to describe a mapped or computed value. The same combination of CLR type and database type can represent values with different semantics. That additional information must remain available through projections, CTEs, set operations, query optimization, serialization, and provider lowering.

This is a general type-system problem, not an interval-specific one.

The current implementation addresses it locally for `TimeSpan` by introducing a parallel duration-specific mechanism:

- `DurationAttribute`;
- `DurationUnit`;
- `ColumnDescriptor.Duration`;
- `SqlDurationExpression`;
- duration-specific handling in visitors, serialization, set operations, and provider conversion.

After some refinement, this may solve the immediate interval scenarios. However, it does not address the underlying architectural limitation. A similar problem can occur whenever the CLR type and database type are insufficient to describe the actual semantics or representation of a value.

As a minor naming point, `Duration` also appears too narrow for the mapping metadata. It describes the storage representation of a temporal value, including native temporal types and numeric representations. Something like `TimeStorage` might express that role more accurately. This does not necessarily apply to computed interval expressions, which are a separate concern.

The main concern is that once this duration-specific solution becomes part of the public API and internal query infrastructure, it becomes an architectural constraint.

A future redesign would then have to either preserve this model and adapt the general type system around it, or replace it through substantial internal changes and public breaking changes. The former accumulates another architectural compromise; the latter postpones the problem until there are more users, dependencies, and compatibility requirements.

We already know how difficult it is to correct years of accumulated architectural problems in one large release. This PR could become another locally reasonable implementation that later has to be preserved, worked around, or removed when the general type system is finally addressed.

Before committing to the current public API, I think we should first agree on the long-term model:

- what constitutes the complete type information of a mapped or computed SQL expression;
- how logical semantics and physical representation are described;
- where this information belongs;
- how it propagates through the SQL tree;
- how set operations reconcile different representations;
- how mapped columns and computed expressions participate in the same type system;
- when provider-specific lowering occurs.

There are two possible ways forward.

The first (recommended) is to design the general type-system extension first and use interval support as its initial validation case.

The second is to continue the current implementation, but design its public surface as the first part of that general extension rather than as an API specific to `TimeSpan`. In that case, we should first define how additional type metadata is declared, propagated, compared, transformed, serialized, and consumed by providers. Time storage units would then be the first application of that mechanism.

Interval support is a strong validation case because it covers native intervals, numeric storage representations, computed date differences, arithmetic, projections, CTEs, and set operations with different representations.

The provider-specific research and translation work in this PR is valuable and should be retained. The architectural question is whether the new metadata and propagation infrastructure should remain specific to intervals or become part of a general type model. We should make that decision before fixing the duration-specific API as a permanent contract.

What do you think?

igor-tkachev:
I think the existing default mapping from `TimeSpan` to SQL Server `time` should remain unchanged for compatibility.

The concrete `TimeSpan` → `bigint` scenario does not have to wait for the general type-system redesign. It can be implemented separately using a dedicated mapping attribute based on the existing `ValueConverterAttribute` mechanism.

For example:

```csharp
[Column(DataType = DataType.Int64)]
[TimeSpanStorage(TimeSpanStorageUnit.Tick)]
public TimeSpan Duration { get; set; }
```

A minimal implementation for the current `bigint` case could look like this:

```csharp
public enum TimeSpanStorageUnit
{
	Tick,
	Millisecond,
	Second
}

[AttributeUsage(
	AttributeTargets.Field | AttributeTargets.Property,
	Inherited = true,
	AllowMultiple = true)]
public sealed class TimeSpanStorageAttribute : ValueConverterAttribute
{
	public TimeSpanStorageAttribute(TimeSpanStorageUnit unit = TimeSpanStorageUnit.Tick)
	{
		Unit            = unit;
		ValueConverter = unit switch
		{
			TimeSpanStorageUnit.Tick        => TimeSpanTicksConverter.Instance,
			TimeSpanStorageUnit.Millisecond => TimeSpanMillisecondsConverter.Instance,
			TimeSpanStorageUnit.Second      => TimeSpanSecondsConverter.Instance,
			_ => throw new ArgumentOutOfRangeException(nameof(unit))
		};
	}

	public TimeSpanStorageUnit Unit { get; }

	public override string GetObjectID()
		=> $".{Configuration}.{Unit}.";
}

sealed class TimeSpanTicksConverter : IValueConverter
{
	public static readonly TimeSpanTicksConverter Instance = new();

	static readonly Expression<Func<TimeSpan, long>> ToProvider =
		static value => value.Ticks;

	static readonly Expression<Func<long, TimeSpan>> FromProvider =
		static value => TimeSpan.FromTicks(value);

	private TimeSpanTicksConverter()
	{
	}

	public bool HandlesNulls => false;

	public LambdaExpression ToProviderExpression   => ToProvider;
	public LambdaExpression FromProviderExpression => FromProvider;
}
```

The millisecond and second variants can use the same pattern with the corresponding conversions. Ticks can be the default because they preserve the complete `TimeSpan` value and require no additional scale convention.

The attribute can later be extended with storage-specific options without changing the default `TimeSpan` mapping or introducing duration-specific metadata into the SQL tree.

This is a local mapping solution with a narrow and stable responsibility. It can be implemented in a separate PR now and close the specific `TimeSpan` → `bigint` issue independently of the broader architectural discussion.

igor-tkachev:
@sdanyliv's work on dedicated AST nodes gives us a concrete architectural direction for interval arithmetic.

The provider-specific research and implementations from this PR are still valuable and can be reused when implementing provider support for those AST nodes.

The remaining work can then be separated into three independent parts:

- `TimeSpan` column storage mappings, such as `bigint` ticks, can be implemented now through a dedicated `ValueConverterAttribute`. This is a local mapping concern and does not depend on interval arithmetic in the SQL AST.
- `DatePartLong`, `DateDiffLong`, the new date parts, provider fixes, and their tests can be merged independently where they do not depend on the interval AST.
- General interval arithmetic should be completed on top of @sdanyliv's AST model, reusing the provider-specific work from this PR.

Together, these three workstreams cover the complete scope of this PR without introducing a second, competing representation for intervals. The remaining issues are provider-specific correctness, precision, range, and testing rather than an unresolved architectural problem.

jogibear9988:
@igor-tkachev 

I don't need SqlDurationExpression or the extra Attributes, ... I only added them cause of the dicussions here. For me it would be enough if we could merge my fixed TimeSpan support and only support mapping TimeSpans to Interval and Long (with Ticks as default) and if we map to Time we keep the old support. This will not introduce new API wich we need to support later, and also only add functionality wich should work later.
I'd rather like we do that, instead of delaying this further and further. My first pull request to the whole problem is now over 3 years old, and I'd rather like to get a not perfect solution instead of waiting for more years.

So I think if we switch back to the commit before the AST, it should not any new public API or Attributes (only in SQL.Date... but I think this are ok). So we can later change the implentation how we like, only the calculations and the Tick support need to stay, but I think they also should stay if we have the advanced system.

jogibear9988:
replaced by #5750

PR #5754, opened 2026-08-10
DB2 setup: disable AUTO_STMT_STATS (real-time statistics) — 3.7x faster tests

igor-tkachev:
## Problem

`DB2` is by far the slowest provider in our local/CI test matrix, despite the DB2 container itself being nearly idle (~3% CPU) while tests run — a classic "client and server both idle, but slow" symptom.

Root cause: `testdb`'s `AUTO_STMT_STATS` (real-time statistics) is on by default. Our test suite constantly creates/drops small tables and immediately runs a query against them (schema setup, `WindowFunctionsTests`, etc.), so almost every query compiles against a table with missing/stale statistics. With `AUTO_STMT_STATS` on, the optimizer synchronously profiles the table right there during query compilation — adding latency to nearly every statement, without showing up as CPU load.

Confirmed this is DB2-side, not the client or the network: `Informix.DB2` goes through the exact same DB2 CLI driver client, with a near-identical test volume, and runs **14.6x faster**. Same client code, wildly different behavior depending on which server it talks to.

## Before / after

Controlled A/B on the same test matrix (`--provider DB2`, `--no-linq-service` to keep the comparison to the direct-connection variant only) — identical test counts and pass/fail either way, only `AUTO_STMT_STATS` differed:

| | `AUTO_STMT_STATS=ON` (before) | `AUTO_STMT_STATS=OFF` (after) |
|---|---|---|
| Total / passed / failed | 5235 / 5120 / 0 | 5235 / 5120 / 0 |
| Duration | 42:52.8 | **11:35** |
| Throughput | 2.00 tests/s | **7.42 tests/s** |

**3.71x faster**, with no change in test outcomes.

## Change

Adds one `db2 UPDATE DB CFG FOR testdb USING AUTO_STMT_STATS OFF` call, run against the `db2inst1` OS user's local (password-less, OS-trusted) connection right after the container finishes its own setup:

- `Build/Azure/scripts/db2.sh` (CI, Linux runner)
- `Build/Azure/scripts/mac.db2.sh` (CI, macOS runner)
- `Data/Setup Scripts/db2.cmd` (local dev image)

`db2.provider.sh`/`mac.db2.provider.sh` are untouched — those only swap in the client driver binaries and are shared with the `Informix.DB2` job, which doesn't need this (different server, no `AUTO_STMT_STATS` mechanism).

## Also included: deterministic AllTypes lookup

Investigating this PR's own CI run surfaced a second, unrelated bug: `DataProviderTestBase.TestType`'s `PassNullSql`/`PassValueSql` look up a just-inserted row by value with no `ORDER BY`, assuming the result is always the well-known canonical ID (1/2). `AllTypes` is a shared fixture other tests also insert into without deleting afterward (by design - see `ResetAllTypesIdentity`'s own comment), so a leftover row whose value collides with another test's magic number can make that lookup return a different, wrong ID. This DB2 config change did not create such a row - it just changed which query plan the optimizer picked for the ambiguous lookup, which is what made an existing leftover row visible as a failure.

Fixed by adding `ORDER BY ID` (or the provider's quoted equivalent) to the base implementation and every override that does not already delegate to it (SapHana, PostgreSQL, Informix's `PassValueSql`, Access's ODBC branch) - deterministically picks the lowest-ID (canonical) row. Verified locally against DB2: deliberately duplicated the canonical `AllTypes` row without recreating the schema, reproduced the failure, confirmed it is fixed.

### DIFF
--- a/Build/Azure/scripts/db2.sh
+++ b/Build/Azure/scripts/db2.sh
@@ -17,3 +17,9 @@ until docker logs db2 | grep -q 'Setup has completed'; do
 done
 
 docker logs db2
+
+# AUTO_STMT_STATS (real-time statistics) makes the optimizer synchronously profile a table's stats
+# during query compilation whenever they're missing/stale - our tests constantly create/drop small
+# tables and immediately query them, so this fires on nearly every fresh table. Measured 3.7x slower
+# with it on (confirmed via a controlled run: identical test counts, only this setting differed).
+docker exec -u db2inst1 db2 bash -lc "db2 connect to testdb && db2 UPDATE DB CFG FOR testdb USING AUTO_STMT_STATS OFF"
--- a/Build/Azure/scripts/mac.db2.sh
+++ b/Build/Azure/scripts/mac.db2.sh
@@ -17,3 +17,9 @@ until docker logs db2 | grep -q 'Setup has completed'; do
 done
 
 docker logs db2
+
+# AUTO_STMT_STATS (real-time statistics) makes the optimizer synchronously profile a table's stats
+# during query compilation whenever they're missing/stale - our tests constantly create/drop small
+# tables and immediately query them, so this fires on nearly every fresh table. Measured 3.7x slower
+# with it on (confirmed via a controlled run: identical test counts, only this setting differed).
+docker exec -u db2inst1 db2 bash -lc "db2 connect to testdb && db2 UPDATE DB CFG FOR testdb USING AUTO_STMT_STATS OFF"
--- a/Data/Setup Scripts/db2.cmd
+++ b/Data/Setup Scripts/db2.cmd
@@ -30,3 +30,9 @@ REM   https://community.ibm.com/community/user/discussion/121-container-communit
 REM ---------------------------------------------------------------------------
 
 docker exec db2 sed -i "/^if ! create_instance; then$/i [ -f /database/config/db2inst1/sqllib/adm/fencedid ] && chown root:db2iadm1 /database/config/db2inst1/sqllib/adm/fencedid" /var/db2_setup/lib/setup_db2_instance.sh
+
+REM AUTO_STMT_STATS (real-time statistics) makes the optimizer synchronously profile a table's stats
+REM during query compilation whenever they're missing/stale - our tests constantly create/drop small
+REM tables and immediately query them, so this fires on nearly every fresh table. Measured 3.7x slower
+REM with it on (confirmed via a controlled run: identical test counts, only this setting differed).
+docker exec -u db2inst1 db2 bash -lc "db2 connect to testdb && db2 UPDATE DB CFG FOR testdb USING AUTO_STMT_STATS OFF"
--- a/Tests/Base/DataProvider/DataProviderTestBase.cs
+++ b/Tests/Base/DataProvider/DataProviderTestBase.cs
@@ -19,9 +19,14 @@ namespace Tests.DataProvider
 		{
 			// number of parameters to create for providers with unnamed parameters
 			paramCount = 1;
-			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p";
+			// ORDER BY ID: AllTypes is a shared fixture other tests also insert into (deliberately never
+			// cleaned up here - see TestBase.Identity.cs's ResetAllTypesIdentity - so a leftover row with a
+			// duplicate value is a real, if rare, possibility). Only the lookup's chosen row matters (Execute
+			// reads a single scalar), so ordering by ID makes it deterministically the lowest-ID (canonical)
+			// match instead of whichever row the engine happens to return first.
+			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p ORDER BY ID";
 		}
-		protected virtual string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = @p";
+		protected virtual string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = @p ORDER BY ID";
 
 		protected T TestType<T>(DataConnection conn, string fieldName,
 			DataType dataType          = DataType.Undefined,
--- a/Tests/Linq/DataProvider/AccessTests.cs
+++ b/Tests/Linq/DataProvider/AccessTests.cs
@@ -46,12 +46,12 @@ namespace Tests.DataProvider
 		{
 			paramCount = dc.DataProvider.Name.IsAnyOf(TestProvName.AllAccessOdbc) ? 3 : 1;
 			return dc.DataProvider.Name.IsAnyOf(TestProvName.AllAccessOdbc)
-				? "SELECT ID FROM {1} WHERE ? IS NULL AND {0} IS NULL OR ? IS NOT NULL AND {0} = ?"
+				? "SELECT ID FROM {1} WHERE ? IS NULL AND {0} IS NULL OR ? IS NOT NULL AND {0} = ? ORDER BY ID"
 				: base.PassNullSql(dc, out paramCount);
 		}
 		protected override string PassValueSql(DataConnection dc) =>
 			dc.DataProvider.Name.IsAnyOf(TestProvName.AllAccessOdbc)
-				? "SELECT ID FROM {1} WHERE {0} = ?"
+				? "SELECT ID FROM {1} WHERE {0} = ? ORDER BY ID"
 				: base.PassValueSql(dc);
 
 		[Test]
--- a/Tests/Linq/DataProvider/InformixTests.cs
+++ b/Tests/Linq/DataProvider/InformixTests.cs
@@ -30,7 +30,7 @@ namespace Tests.DataProvider
 			paramCount = 1;
 			return null;
 		}
-		protected override string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = ?";
+		protected override string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = ? ORDER BY ID";
 
 		[Test]
 		public void TestDataTypes([IncludeDataSources(CurrentProvider)] string context)
--- a/Tests/Linq/DataProvider/PostgreSQLTests.cs
+++ b/Tests/Linq/DataProvider/PostgreSQLTests.cs
@@ -51,10 +51,10 @@ namespace Tests.DataProvider
 		protected override string? PassNullSql(DataConnection dc, out int paramCount)
 		{
 			paramCount = 1;
-			return "SELECT \"ID\" FROM \"AllTypes\" WHERE :p IS NULL AND \"{0}\" IS NULL OR :p IS NOT NULL AND \"{0}\" = :p";
+			return "SELECT \"ID\" FROM \"AllTypes\" WHERE :p IS NULL AND \"{0}\" IS NULL OR :p IS NOT NULL AND \"{0}\" = :p ORDER BY \"ID\"";
 		}
 		protected override string  GetNullSql  (DataConnection dc) => "SELECT \"{0}\" FROM {1} WHERE \"ID\" = 1";
-		protected override string  PassValueSql(DataConnection dc) => "SELECT \"ID\" FROM \"AllTypes\" WHERE \"{0}\" = :p";
+		protected override string  PassValueSql(DataConnection dc) => "SELECT \"ID\" FROM \"AllTypes\" WHERE \"{0}\" = :p ORDER BY \"ID\"";
 		protected override string  GetValueSql (DataConnection dc) => "SELECT \"{0}\" FROM {1} WHERE \"ID\" = 2";
 
 		[Test]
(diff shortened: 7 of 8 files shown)


MaceWindu:
> DB2 is by far the slowest provider in our local/CI test matrix

That's a strange statement as DB2 was one of fastest providers and CI times prove it

MaceWindu:
And this PR run displays same time as without it...

MaceWindu:
Also PR changes baselines for all providers but don't regenerate them

MaceWindu:
> [!IMPORTANT]

Post-merge note on the `ORDER BY ID` half of this PR.

The CI run that motivated it — [build 22698](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22698), commit 5b6fcaaf9, which carried only the `AUTO_STMT_STATS` change — failed `TestDataTypes("DB2")` with eight assertions:

```
Assert.That(id, Is.EqualTo(2))   Expected: 2   But was: 183642
Assert.That(id, Is.EqualTo(2))   Expected: 2   But was:  83641
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was:  50773   (x4)
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was: 182126
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was: 216310
```

That is not one leftover row colliding with a magic number. At least five rows survived at IDs up to 216310 — `AllTypes` on DB2 had accumulated roughly 216k inserts' worth of uncleaned rows.

The PR body cites `ResetAllTypesIdentity` as evidence that leftovers are tolerated by design. `Tests/Base/TestBase.Identity.cs:12-13` states the opposite: extra records are deliberately *not* removed "to allow us detect tests that forget to cleanup and leave junk records in those tables". The detector is the identity collision after the reseed — and that reseed is to `lastValue + 1 = 3` (`TestBase.Identity.cs:146`), so junk sitting at ID 50773 never collides with anything. The `TestType` assertion was the only thing surfacing this class of pollution, `ORDER BY ID` switches it off, and the rows are still there.

`ORDER BY ID` is defensible in isolation — the lookup does mean "the canonical row", not "assert exactly one match exists". The gap is the other half: the polluting test was never identified, and nothing replaces the signal that found it.

Suggested follow-up: an explicit `COUNT(*)` / `MAX(ID)` assertion on `AllTypes` in fixture setup, so the next occurrence fails deliberately rather than depending on which plan the optimizer picks, plus tracking down the source of the rows.

MaceWindu:
On the `AUTO_STMT_STATS` half: the 3.71x speedup does not reproduce in CI.

`Tests (.NET 10): DB2 LUW 11.5` step, recent `test-all` runs:

| Build | Date | Tests | Duration | `AUTO_STMT_STATS` |
|---|---|---|---|---|
| [22622](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22622) | 08-04 | 9141 | **6m 35s** | ON |
| [22693](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22693) | 08-09 | 9301 | 7m 12s | ON |
| [22710](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22710) | 08-10 | 9301 | 7m 10s | ON |
| [22726](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22726) | 08-11 | 9142 | **6m 41s** | OFF |

Build 22622 is the apples-to-apples comparison: same test count as the post-merge run (9141 vs 9142), and 6 seconds *faster* without the change. Job-level step durations across the last ten pre-change runs span 6.66-7.24 min; the single post-change run sits at 6.73 min, mid-band.

The command does apply — build 22726's setup log ends with `DB20000I  The UPDATE DATABASE CONFIGURATION command completed successfully.` — so this is not a silent failure. It simply has no effect on the CI DB2 job.

On the "by far the slowest provider" framing: in build 22710, `Tests: Linux DB2` was 13.7 min, ranking **35th of 54** jobs — behind `Windows SqlServer2019Extras` (116.7 min) and ten SQL Server legs at 44-65 min. CI times agree with @MaceWindu.

The local measurement may well be real for that environment, but note the scale: the "after" figure quoted in the PR body is 11m35s for 5235 tests, while CI runs ~9100 tests in under 7 minutes with the setting **on**. That points at the local Docker host as the bottleneck rather than `AUTO_STMT_STATS`.

Two things I could not check without DB2 access, either of which would explain a null result: whether the parameter is inert in this image because `AUTO_MAINT` / `AUTO_TBL_MAINT` / `AUTO_RUNSTATS` gate it, and whether the update took effect `IMMEDIATE` or `DEFERRED`. A single `db2 GET DB CFG FOR testdb` would settle both.

MaceWindu:
Smaller point on the same scripts: the new command is now the sole gate on the DB2 setup task.

`Build/Azure/scripts/db2.sh` has no `set -e` and no explicit `exit`, so the script's exit code is its last command's. The `CmdLine@2` step that runs it (`Build/Azure/pipelines/templates/test-workflow-linux.yml:102-107`, `displayName: Setup tests`) fails the job on a non-zero exit. Before this PR the last command was `docker logs db2`, which returns 0 whenever the container exists — the wait loop above already proved that — so the script was effectively always-succeeds.

Now the optional perf tweak is the only command that can fail the DB2 leg, while `docker run` and the readiness loop's non-`exit 1` paths still don't gate it. The DB2 CLP also returns non-zero for warnings, and `db2 connect to testdb` now runs at the earliest instant after the `Setup has completed` log line rather than minutes later when the tests connect — a tighter readiness window than the script previously exercised. No failure observed so far; this is robustness, not a demonstrated break.

Making it explicitly best-effort restores the prior contract and keeps a skipped optimisation diagnosable in the log:

```suggestion
docker exec -u db2inst1 db2 bash -lc "db2 connect to testdb && db2 UPDATE DB CFG FOR testdb USING AUTO_STMT_STATS OFF" || echo "WARNING: could not disable AUTO_STMT_STATS - DB2 test run will be slower than expected"
```

`Build/Azure/scripts/mac.db2.sh:25` is byte-identical and needs the same edit. `Data/Setup Scripts/db2.cmd:38` has the inverse problem — batch with no `errorlevel` checks, so a failure there is swallowed and a local dev just gets the slow behaviour with no signal.

(Moot if the two CI-side hunks get reverted per the timing data above.)

MaceWindu:
Correction to my earlier comment: I wrote that `ORDER BY ID` is "defensible in isolation". Per @MaceWindu, leftover rows in `AllTypes` are not a tolerated state at all — a leftover row is a test bug. On that reading the change is not defensible: it turns a test that was correctly reporting a bug into a test that passes while the bug remains.

So the `ORDER BY ID` hunks — `DataProviderTestBase.PassNullSql` / `PassValueSql` plus the AccessTests / InformixTests / PostgreSQLTests / SapHanaTests overrides — should be reverted, and the actual work is finding the test that leaves rows behind in DB2's `AllTypes`. The evidence for what to look for is in the failed run above: surviving rows at IDs 50773, 83641, 182126, 183642 and 216310.

The rest of my earlier comment stands: the `ResetAllTypesIdentity` citation in the PR description is inverted relative to what that code says, and the identity-collision detector does not cover junk at IDs far above the reseed point.

igor-tkachev:
Checked `TestBase.Identity.cs` — confirmed. DB2's `ResetAllTypesIdentity` branch only reseeds the identity counter (`RESTART WITH 3`); it never deletes rows, unlike the Informix branch (which has an explicit `DELETE FROM AllTypes WHERE ID > lastValue`). The class comment does say leftovers are meant to surface as a failure so the offending test gets found — you're right that `ORDER BY ID` removes that signal.

That said, the junk accumulation itself isn't something this PR introduced — it was already possible before `AUTO_STMT_STATS` was touched here, `ResetAllTypesIdentity` has looked like this independently of anything in this PR. `ORDER BY ID` is a separate, legitimate fix in its own right: `PassNullSql`/`PassValueSql` selecting an unspecified row among ties is nondeterministic regardless of *why* ties exist (junk rows are one cause, but not the only possible one), and that nondeterminism predates this PR too. So I don't think reverting it is the right move — it wasn't broken by this change, and un-reverting it would just bring back flaky row selection for everyone, DB2 or not.

Filed the actual leak as its own issue so it gets tracked and investigated properly, rather than living only in this thread: #5765. Included the build 22698 failure data and the `ResetAllTypesIdentity` read from this thread.

Also looking into the `AUTO_STMT_STATS` timing discrepancy and the baselines point separately.

igor-tkachev:
On "DB2 was one of fastest providers" — that doesn't match what I saw locally. This is from the full, mass parallel run (`--no-linq-service`, the whole provider matrix running concurrently, not DB2 in isolation) that led to this PR: `Informix.DB2` (Informix accessed through the same DB2 CLI client) finished **first** among all providers — 5120 tests in 2:52 (29.1 tests/s) — while `DB2` itself finished **last** — 5235 tests (2% more) in 42:53 (2.0 tests/s). Same underlying client code, same concurrent-run conditions, 14.6x apart, purely by which server it's talking to. That rules out the client and the general "everything's slower under load" explanation — Informix ran under the exact same contention and wasn't slow.

I'm re-checking the CI-timing side of your comment separately — currently re-running DB2 locally with `AUTO_STMT_STATS` toggled to compare against the current codebase. One caveat that already showed up: run in isolation (just DB2, no other providers competing for the host), the current `AUTO_STMT_STATS=ON` leg finished in under 12 minutes — nowhere near the original 42:53. So the mass-run number and an isolated-run number aren't directly comparable to each other; I'll report the isolated ON-vs-OFF comparison once both sides are done, with that caveat attached.

igor-tkachev:
> > DB2 is by far the slowest provider in our local/CI test matrix
> 
> That's a strange statement as DB2 was one of fastest providers and CI times prove it

Local run:

```
  waiting for 38 run(s) across 31 resource group(s) (max 20 concurrent)...
PASSED DuckDB:net10.0           exit=0  providers: DuckDB  (01:43)
PASSED SQLite.MS:net10.0        exit=0  providers: SQLite.MS  (01:45)
PASSED SQLite.Classic:net10.0   exit=0  providers: SQLite.Classic  (01:46)
PASSED Access.Jet.Odbc:net462   exit=0  providers: Access.Jet.Odbc  (01:52)
PASSED Access.Ace.Odbc:net462   exit=0  providers: Access.Ace.Odbc  (01:59)
PASSED Informix.DB2:net10.0     exit=0  providers: Informix.DB2  (02:55)
PASSED PostgreSQL.19:net462     exit=0  providers: PostgreSQL.19  (04:25)
PASSED Access.Jet.OleDb:net462  exit=0  providers: Access.Jet.OleDb  (05:09)
PASSED cli:net10.0              exit=0  (00:19)
PASSED SQLite.MS:ef:net10.0     exit=0  providers: SQLite.MS  (00:17)
PASSED Access.Ace.OleDb:net462  exit=0  providers: Access.Ace.OleDb  (09:18)
PASSED Sybase.Managed:net10.0   exit=0  providers: Sybase.Managed  (05:41)
PASSED PostgreSQL.14:net10.0    exit=0  providers: PostgreSQL.14  (07:34)
PASSED ClickHouse.MySql:net10.0 exit=0  providers: ClickHouse.MySql  (13:20)
PASSED Oracle.11.Managed:net10.0 exit=0  providers: Oracle.11.Managed  (11:36)
PASSED ClickHouse.Octonica:net10.0 exit=0  providers: ClickHouse.Octonica  (13:32)
PASSED PostgreSQL.14:ef:net10.0 exit=0  providers: PostgreSQL.14  (02:03)
PASSED ClickHouse.Driver:net10.0 exit=0  providers: ClickHouse.Driver  (14:30)
PASSED PostgreSQL.19:net10.0    exit=0  providers: PostgreSQL.19  (04:55)
PASSED MariaDB.11:net10.0       exit=0  providers: MariaDB.11  (15:22)
PASSED Oracle.23.Managed:net10.0 exit=0  providers: Oracle.23.Managed  (15:35)
PASSED PostgreSQL.19:ef:net10.0 exit=0  providers: PostgreSQL.19  (01:23)
PASSED SapHana.Native:net10.0   exit=0  providers: SapHana.Native  (10:03)
PASSED SqlServer.SA:net10.0     exit=0  providers: SqlServer.SA  (19:57)
PASSED SqlServer.2017:net10.0   exit=0  providers: SqlServer.2017  (19:59)
PASSED SqlServer.2025.MS:net462 exit=0  providers: SqlServer.2025.MS  (20:28)
PASSED SqlServer.2025:net462    exit=0  providers: SqlServer.2025  (20:32)
PASSED SqlServer.SA.MS:net10.0  exit=0  providers: SqlServer.SA.MS  (19:31)
PASSED MySql.8.0:net10.0        exit=0  providers: MySql.8.0  (20:13)
PASSED SqlServer.SA.MS:ef:net10.0 exit=0  providers: SqlServer.SA.MS  (01:02)
PASSED MySql.5.7:net10.0        exit=0  providers: MySql.5.7  (20:56)
PASSED SqlServer.2025.MS:net10.0 exit=0  providers: SqlServer.2025.MS  (14:10)
PASSED SqlServer.2025:net10.0   exit=0  providers: SqlServer.2025  (14:07)
PASSED SqlServer.2025.MS:ef:net10.0 exit=0  providers: SqlServer.2025.MS  (01:03)
PASSED Firebird.2.5:net10.0     exit=0  providers: Firebird.2.5  (38:13)
PASSED YDB:net10.0              exit=0  providers: YDB  (41:48)
PASSED Firebird.5:net10.0       exit=0  providers: Firebird.5  (42:19)
PASSED DB2:net10.0              exit=0  providers: DB2  (42:52)
```

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `a52e9d9`.</sub>

PR #5755, opened 2026-08-10
Tests: make DataProviderTestBase.TestType's ID lookup deterministic (ORDER BY)

igor-tkachev:
## Problem

`DataProviderTestBase.TestType` (the shared helper behind every provider's
`TestDataTypes`) inserts a row, then looks it up by *value* to confirm the
returned identity matches the well-known canonical ID (1 for the NULL row,
2 for the value row):

```csharp
protected virtual string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = @p";
...
id = conn.Execute<int?>(sql, ...);
Assert.That(id, Is.EqualTo(2));
```

No `ORDER BY`. `AllTypes` is a shared fixture many other tests also insert
into. By design, `ResetAllTypesIdentity` only resets the identity *counter*
and deliberately leaves behind any row a test forgot to clean up (its own
comment: *"we intentionally don't remove extra records and allow it to fail
in that case to allow us to detect tests that forget to cleanup"*) - so a
leftover row whose value happens to collide with another test's magic
number is a real, if rare, possibility. Without `ORDER BY`, which of several
matching rows comes back is whatever order the engine/query plan happens to
pick - usually the low-ID canonical row by luck, but never guaranteed.

## How this surfaced

CI on #5754 (an unrelated DB2 performance change - `AUTO_STMT_STATS OFF`)
hit `DB2Tests.TestDataTypes` failing with `id` values in the tens/hundreds
of thousands instead of 1/2. The DB2 config change doesn't touch which rows
exist in a table - it only changes which query plan the optimizer picks for
an ambiguous `WHERE` - so it didn't *cause* the extra rows, it just changed
which one of several already-matching rows got returned first, surfacing a
pre-existing determinism gap.

Reproduced locally against DB2 without any config change: duplicated the
canonical `AllTypes` row (same values, new ID) without recreating the
schema, then re-ran `TestDataTypes` - failed the same way pre-fix, passed
post-fix.

## Fix

Adds `ORDER BY ID` (or the provider's quoted equivalent) to
`PassNullSql`/`PassValueSql` in the base class and every override that
doesn't already delegate to it (SapHana, PostgreSQL, Informix's
`PassValueSql`, Access's ODBC branch) - deterministically picks the
lowest-ID (canonical) match instead of whatever the engine returns first.
`GetNullSql`/`GetValueSql` (`WHERE ID = 1`/`WHERE ID = 2`) are untouched -
they already filter by the unique PK and were never ambiguous.

### DIFF
--- a/Tests/Base/DataProvider/DataProviderTestBase.cs
+++ b/Tests/Base/DataProvider/DataProviderTestBase.cs
@@ -19,9 +19,14 @@ namespace Tests.DataProvider
 		{
 			// number of parameters to create for providers with unnamed parameters
 			paramCount = 1;
-			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p";
+			// ORDER BY ID: AllTypes is a shared fixture other tests also insert into (deliberately never
+			// cleaned up here - see TestBase.Identity.cs's ResetAllTypesIdentity - so a leftover row with a
+			// duplicate value is a real, if rare, possibility). Only the lookup's chosen row matters (Execute
+			// reads a single scalar), so ordering by ID makes it deterministically the lowest-ID (canonical)
+			// match instead of whichever row the engine happens to return first.
+			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p ORDER BY ID";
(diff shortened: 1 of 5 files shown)


igor-tkachev:
Folded into #5754 instead — same underlying investigation, no reason to keep separate.

PR #5757, opened 2026-08-10
EF PostgreSQL: recover from Postgres 'invalid database' state

igor-tkachev:
## Problem

`EnsureDeleted()`'s own `Exists()`-then-DROP path doesn't recover from Postgres's "invalid database" state: a connection attempt to an invalid database fails with `55000: cannot connect to invalid database`, which `Exists()` doesn't treat as "doesn't exist" the way it does the ordinary `3D000: database does not exist` — so the invalid state persists and every subsequent test against that provider fails identically (confirmed: 193 of 201 EF PostgreSQL.19 tests failing with this exact cascade in one run). Root cause of what marks the database invalid in the first place wasn't isolated (reproduces instantly with zero host contention; Postgres server logging with `log_statement=all`/`log_min_messages=debug1` shows nothing before the final rejection, meaning it happens too early in backend startup — `postinit.c` — for statement logging to capture). Confirmed not a parallelism race: reproduces on a single isolated run with zero load on the machine.

## Fix

Postgres's own documented fix for an invalid database is exactly `DROP DATABASE` (works on both invalid and normal existing databases) via a connection to the `postgres` maintenance database. `Tests/EntityFrameworkCore/ContextTestBase.cs`'s `InitializeDatabase` now catches specifically `Npgsql.PostgresException` with `SqlState=55000` and an "invalid database" message, runs that DROP-and-retry recovery, then retries `EnsureCreated()` once — every other call proceeds exactly as before, no `DROP DATABASE` issued unless the invalid state is actually hit. Also raises Npgsql's `CommandTimeout` to 120s: under real host contention `EnsureCreated`'s `CREATE DATABASE` can itself run past Npgsql's 30s default (confirmed: a real run's first attempt timed out at 30.6s), which is a *different* exception than the one the recovery catch matches, so it propagates uncaught otherwise — this buys the normal path enough headroom that the timeout, and therefore the interrupted-DDL state the recovery path exists for, is far less likely to happen at all.

This PR keeps the three commits from the original investigation separate rather than squashing, since the middle one documents a real regression caught and fixed before merge: the first version ran `DROP DATABASE ... WITH (FORCE)` unconditionally before every `EnsureDeleted`/`EnsureCreated` call — since `InitializeDatabase` reruns whenever the cached `(connectionString, TContext type)` pair changes, not just once per suite, the unconditional `FORCE` drop was killing other still-pooled connections to the same database on every one of those calls (`57P01: terminating connection due to administrator command`), a regression worse than the original problem. The second commit scopes the drop to only fire inside the specific `catch`.

## Verification

Reproduced the original cascade 3 times in isolation with zero host contention, then passed cleanly 3 times in a row after the fix. A full EF PostgreSQL.19 suite run afterward: 223 total, 0 failed, 196 passed, 27 skipped — no trace of the invalid-database cascade. Locally re-verified on this branch against a local PostgreSQL.19 container: `TestToList`/`TestInheritance`/`Issue5296Test` — 9 passed, 0 failed, 4 skipped.

### DIFF
--- a/Tests/EntityFrameworkCore/ContextTestBase.cs
+++ b/Tests/EntityFrameworkCore/ContextTestBase.cs
@@ -54,6 +54,9 @@ namespace LinqToDB.EntityFrameworkCore.Tests
 		{
 			using var _ = new DisableBaseline("create db");
 
+			if (provider.IsAnyOf(TestProvName.AllPostgreSQL))
+				DropPostgresDatabaseIfExists(connectionString);
+
 			context.Database.EnsureDeleted();
 			context.Database.EnsureCreated();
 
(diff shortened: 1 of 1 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `97648e1`.</sub>

PR #5761, opened 2026-08-11
Stabilise the x86 Access test legs: 4GB address space, and pin System.Data.OleDb

igor-tkachev:
## Problem

Two independent causes of the same symptom — the 32-bit Access legs falling over instead of reporting test results.

**1. The 2GB address-space cap.** An x86 test host is capped at 2GB of address space no matter how much RAM the machine has, and NUnit3's classic engine spends part of that budget on a second AppDomain per assembly. #5753 hit exactly this — 14648 tests reaching ~1.7GB of reserved address space, leaving no contiguous block for NUnit's result document — and worked around it by splitting the ACE leg in two, noting there was no lever inside the test harness. That's correct as far as it goes: the lever is just outside the harness, in the PE header.

**2. `System.Data.OleDb` 10.0.0+.** A confirmed upstream regression ([dotnet/runtime#125221](https://github.com/dotnet/runtime/issues/125221), fixed by PR #130906): COM error-wrapper objects (`IErrorInfo`/`IErrorRecords`) skip `GC.SuppressFinalize()` after `Marshal.Release()`, so the finalizer later double-releases an already-freed pointer. The result is an `AccessViolationException`/`NullReferenceException` raised from the finalizer thread, where nothing can catch it and the whole process dies. Only OLE DB providers reach that code, which in this repo means Access — and the tests deliberately provoke driver errors, so the wrappers accumulate. Timing depends on the GC, so it presents as an intermittent crash. The upstream fix ships only in .NET 11.0.0.

## Change

`IMAGE_FILE_LARGE_ADDRESS_AWARE` in the built exe's COFF header raises the cap to 4GB under WOW64, set via the `LargeAddressAware` package and gated on `$(RuntimeIdentifier) == 'win-x86'` — which is what `dotnet publish -a x86` in `build-job.yml` resolves to, so it covers exactly the legs that build 32-bit and nothing else. `$(PlatformTarget)` isn't resolved early enough to gate on instead.

This does **not** undo #5753's leg split. The two are orthogonal — one halves the tests per process, the other doubles the space each process has — and together they leave real headroom rather than sitting just under the limit. Any future 32-bit leg (`Access_MDB`, the 32-bit DB2/Informix clients) gets the same benefit without having to be split.

`System.Data.OleDb` is pinned to `9.0.10`, the newest release confirmed without the regression, with a comment recording when to revert.

## Verification

Built exactly as CI does — `-c Azure -a x86`, the configuration `build-vars.yml` sets:

| Check | Result |
|---|---|
| net462 x86 host COFF characteristics | `0x0122` — `LARGE_ADDRESS_AWARE` set |
| net8.0 x86 publish, shipped OleDb | `9.0.10` |
| net462 output | no `System.Data.OleDb.dll` — Framework-native, unaffected either way |

**What is not verified:** the finalizer crash itself was not reproduced locally. The case for that half rests on the upstream issue matching the CI stack traces (`ComObject.Finalize` → `Marshal.Release`) and on the affected version range, not on a red-to-green run here.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -33,6 +33,19 @@
 		<!--used by VersionOverride in project files-->
 		<OracleManagedLinqPadVersion>23.26.300</OracleManagedLinqPadVersion>
 
+		<!--
+		used by VersionOverride in project files (tests only - linq2db.cli and the LINQPad driver stay on $(Net10Latest))
+		10.0.0+ has a confirmed regression where COM error-wrapper objects (IErrorInfo/IErrorRecords) skip
+		GC.SuppressFinalize() after Marshal.Release(), so the finalizer later double-releases an already-freed
+		pointer - AccessViolationException/NullReferenceException in ComObject.Finalize(), intermittent (depends on
+		GC timing relative to accumulated OLEDB errors). Only OLE DB providers hit it, which here means Access: the
+		tests deliberately provoke driver errors, so the wrappers accumulate. net462 is unaffected (it uses the
+		Framework-native OleDb, not this package). Fixed upstream (dotnet/runtime#125221, PR #130906) but only ships
+		in .NET 11.0.0; 9.0.10 is the newest version confirmed NOT to have the regression. Drop the override once
+		either .NET 11 ships or a patched 10.x is released.
+		-->
+		<OleDbTestsVersion>9.0.10</OleDbTestsVersion>
+
 		<!--bundled version for legacy SignalR packages (net462 + SignalR.Client !net8+ + SignalR.Core)-->
 		<SignalRLegacyVersion>1.2.11</SignalRLegacyVersion>
 		<!--bundled version for legacy AspNetCore [net462] packages (Microsoft.AspNetCore + Kestrel.Core)-->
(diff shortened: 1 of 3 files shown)


MaceWindu:
Two commits on this branch reshape it.

**Dropped the address-space half.** `LARGEADDRESSAWARE` works around memory consumption in linq2db and the test suite rather than addressing it, and the OOM it targeted is already mitigated by #5753's per-provider ACE leg split.

It also had no effect on the legs it named. `LargeAddressAware.targets` patches `@(IntermediateAssembly)`, which is the exe only on `net462`. On `net8.0`/`net9.0`/`net10.0` that is the managed `linq2db.Tests.dll`, while the process image is the SDK apphost — whose `win-x86` template ships with the flag clear in 8.0.0, 9.0.10 and 10.0.0 alike. Measured on a publish built the same way:

| | `net462` x86 | `net10.0` x86 |
|---|---|---|
| flag on launched `.exe` | `0x0122` — set | `0x0102` — not set |
| `lpMaximumApplicationAddress` | `0xFFFEFFFF` (4GB) | `0x7FFEFFFF` (2GB) |

#5753's failing leg was `Tests (.NET 10 x86)`, and the verification table in the description covers `net462` only.

**Scoped the `System.Data.OleDb` pin to tests.** As a global `PackageVersion` it also lowered the version bundled into `linq2db.cli` and the LINQPad driver — the case this file's own header comment warns about. It now applies via `VersionOverride` on the test-side reference in `Tests/linq2db.Providers.props`, and everything else stays on the latest 10.x. Restore confirms `Tests/Linq` → 9.0.10, `Source/LinqToDB.CLI` → 10.0.10.

Title and description still lead with the address-space change and need updating.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `8d084b9`.</sub>

PR #5762, opened 2026-08-11
Remote: stop leaking a client per two-query plan, and guard double dispose

igor-tkachev:
## Problem

`RemoteDataContextBase`'s `QueryRunner` assigned `_client = _dataContext.GetClient()` unconditionally at the top of every execute:

```csharp
public override async Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
{
    _client = _dataContext.GetClient();   // every call, not just the first
    ...
}
```

`QueryRunner.NonQueryQuery2` / `QueryQuery2` execute **two** queries on the *same* `IQueryRunner` — that's the UPDATE-then-INSERT plan `MakeAlternativeInsertOrUpdate` builds for `InsertOrReplace` on any provider without a native upsert (Access, and everything else that reports `IsInsertOrUpdateSupported == false`) — while the runner is disposed once, at the end. So the second execute overwrote the field and the first client was simply dropped: never disposed, and for a real transport its connection with it. One leaked client per such call.

Separately, `Dispose`/`DisposeAsync` released `_client` without clearing the field, so a second dispose could release the same client twice.

## Fix

`_client ??= _dataContext.GetClient()` reuses the client the runner already holds, and `Interlocked.Exchange` makes the release single-shot:

```csharp
var client = Interlocked.Exchange(ref _client, null);

if (client != null)
    DisposeClient(client);
```

## Test

The regression test drives the real path in-process — an actual `LinqService` over a real `DataConnection`, not a mock of the protocol — and counts clients obtained against clients disposed.

Getting it to exercise the bug at all took one non-obvious step worth flagging for review: SQLite compiles `InsertOrReplace` into a single native upsert statement, which never re-executes the runner, so the straightforward version of this test passes even with the bug present. The test therefore turns `IsInsertOrUpdateSupported` off on a SQLite provider instance, which selects the same two-query plan providers without native upsert use for real.

Verified red before green: **3 clients created, 2 disposed** without the fix; equal after it. The rest of `RemoteContextTests` still passes (32 passed / 30 skipped — the skips are the non-remote and other-transport combinations that fixture always skips).

Builds clean in Release for `net10.0` and `netstandard2.0`.

## Notes

Found while chasing connection churn under a heavy parallel test run, but it isn't test-only or parallelism-only — any consumer of a remote context (gRPC / WCF / SignalR / HTTP) calling `InsertOrReplace` against a provider without native upsert support leaks a client per call.


MaceWindu (commented):
Reviewed, synced with master (`8c5d240da`) and pushed three follow-up commits. The fix is correct as written — `_client ??=` covers both two-query plans (`NonQueryQuery2`'s UPDATE-then-INSERT and `QueryQuery2`'s scalar-then-nonquery), and the sync entry points route through the same async bodies. No public API removed or changed, so no `PublicAPI` or `CompatibilitySuppressions` work.

Two corrections to the description, and three changes that grew out of them.

## The impact paragraph overstates which transports leaked

The body says gRPC / WCF / SignalR / HTTP. Two of those never leaked, because they do not hand out a fresh client:

| transport | `GetClient()` returns | client disposal | leaked per two-query call |
|---|---|---|---|
| gRPC | `new GrpcLinqServiceClient(GrpcChannel.ForAddress(…))` | `IDisposable` | yes, a channel and its HTTP/2 connection |
| WCF | `new WcfLinqServiceClient(…)` | `ClientBase<T>` → `IDisposable` | yes, a channel |
| HTTP | the shared `Client` field | not disposable | no |
| SignalR | the shared `_client` field | `DisposeAsync() => default` | no |

For HTTP and SignalR the old `_client = GetClient()` overwrote the field with the same instance, so there was nothing to drop. The leak is real for gRPC and WCF.

## That sharing was also an unguarded contract, and it hid two larger leaks

The runner disposed whatever `GetClient()` returned, including those two shared instances. It was harmless only by accident — one client is not disposable at all and the other's disposal does nothing. `RemoteDataContextBase` now carries `protected virtual bool OwnsClient => true`, and the three release sites (configuration preload, `CommitBatchAsync`, the runner) honour it. gRPC and WCF keep the default.

Making that explicit exposed why those two clients cannot release anything, which is that both contexts leak the transport they build themselves:

- `SignalRDataContext(HubConnection)` sets a `_dispose` flag and disposes the client on teardown, but `SignalRLinqServiceClient.DisposeAsync()` is `=> default`, so the `HubConnection` and its WebSocket were never disposed.
- `HttpClientDataContext(Uri, string)` constructs an `HttpClient` behind a `CA2000` suppression and never disposes it.

Both are fixed the same way: the context holds the instance it created and disposes it in `Dispose` and `DisposeAsync`. Constructors that receive a client or an `HttpClient` leave it to its owner, so no existing behaviour changes. `SignalRLinqServiceClient.DisposeAsync()` deliberately stays a no-op and now says so — the context hands that one instance to every query, so releasing the connection there would kill the context after its first query.

`SignalRDataContext` also only overrode `DisposeAsync`, so a synchronous `using` leaked regardless. It overrides `Dispose` too now. Disposal order changed to `base` first, then the transport: `base.DisposeAsync()` runs `IDataContextInterceptor.OnClosing`/`OnClosed`, and an interceptor that issues a query needs the transport alive.

## Smaller items

`Interlocked.Exchange` in the runner's disposal is now a plain null-out. The double dispose it guarded is sequential — `NonQueryQuery2` uses `using` and its async twin `await using`, so exactly one of `Dispose`/`DisposeAsync` runs per runner — while the `_client ??=` acquisition is not atomic, so the pair implied a concurrency guarantee only half of it kept.

The regression test's cleanup called `SqliteConnection.ClearAllPools()`, which empties every SQLite pool in the process. Since #5614 provider lanes run in parallel, and a test with no provider parameter is classified `GatedInline`, so it runs alongside the SQLite lane. It now uses the scoped `SqliteConnection.ClearPool`. This mirrors `FirebirdTools`, which documents `ClearPool` as the scoped alternative and is used that way in `TestUtils`.

## Verification

- Red first: without the SignalR fix the two ownership tests fail with `HttpRequestException` (the connection is alive and dials the dead port) instead of `ObjectDisposedException`; the caller-supplied-client control stays green.
- `RemoteContextTests` on `SQLite.MS` + `SQLite.MS.LinqService`, all four transports: 35 passed, 0 failed, 30 skipped.
- Release across `net462`, `netstandard2.0`, `net8.0`, `net9.0`, `net10.0` for both remote clients and `LinqToDB`: clean. `CA2000` does not fire once the object is actually disposed, so the suppression is gone. Release also caught an `IDE0200` that only appears on `net462`/`netstandard2.0`, where `HubConnection.DisposeAsync()` returns `Task` rather than `ValueTask`.

CI has not run against this head. The last green `test-all` was 2026-08-11 against `e128fe2de`, before the master merge brought in parallel test execution (#5614) and before these commits.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A remote data context leaked one client — and the connection it holds — per `InsertOrReplace`, `InsertOrUpdate` or fluent `Upsert` that LinqToDB emulates as two statements. That happens on providers without native upsert support, and also on SAP HANA whenever the insert and update branches differ, which they routinely do for an entity with `SkipOnInsert` or `SkipOnUpdate` columns. Materially visible on the gRPC and WCF transports, which build a channel per call; the HTTP and SignalR clients are context-shared and were unaffected.
- `SignalRDataContext(HubConnection, …)` now disposes the `HubConnection` you passed in when the context is disposed — it was documented as owning it but never released it. Note that the synchronous `Dispose()` now waits for the connection to tear down, where previously it released nothing; prefer `DisposeAsync()`. If you reuse that connection after disposing the context, wrap it instead: `new SignalRDataContext(new SignalRLinqServiceClient(hubConnection))`. Likewise, the `HttpClientDataContext(Uri, string)` constructor now disposes the `HttpClient` it created for you.
- If you subclass `RemoteDataContextBase` and hand out a client you own — a shared `ILinqService` — override the new `OwnsClient` property to return `false`. It defaults to `true`, which is what makes the disposal above happen, so without the override your shared client is disposed after every query.


Remote data contexts no longer leak a client per emulated upsert, and now dispose a `HubConnection` or `HttpClient` they were documented as owning.


---
<sub>Generated from commit `0e70df4`.</sub>

PR #5763, opened 2026-08-11
YdbSchemaProvider: retry schema reads that race a just-committed DDL

igor-tkachev:
## Problem

`GetSchema("Tables")` / `GetSchema("Columns")` are served by `DescribeTable` against YDB's scheme service, and that service can lag a just-committed `CREATE`/`DROP TABLE` by up to roughly a second. This is visible in the server's own logs: `SCHEME_BOARD_REPLICA` / `SCHEME_BOARD_SUBSCRIBER` propagation keeps going for about that long *after* the DDL's own commit and audit-log entry. A schema read issued right after creating a table can land inside that window and come back with `SchemeError` for a table that, from every other connection's perspective, already exists.

Observed three times on different tests, always with the same signature — `Ydb.Sdk.Ado.YdbException`, `Status: SchemeError`, immediately after `DROP TABLE IF EXISTS` + `CREATE TABLE`:

- `SchemaProviderTests.IncludeExcludeCatalogTest`
- `SchemaProviderTests.IncludeExcludeSchemaTest`
- `YdbTests.SchemaProvider_ReturnsCreatedTable` / `_ReturnsCorrectColumnMetadata`

Different tests each time, identical failure — which is what points at a propagation race rather than at any one test.

## Fix

`GetSchemaWithRetry` wraps both reads: at most five attempts, 150ms apart, retrying **only** `SchemeError`.

Scoped deliberately to this provider's own schema reads rather than to `YdbRetryPolicy`. That policy leaves `SchemeError` out of its retryable set on purpose — blanket-retrying "no such object" would paper over real errors, and this change doesn't touch it. Nothing is swallowed here either: once the attempts run out the exception propagates unchanged, so a genuine scheme failure is delayed by a few hundred milliseconds rather than hidden. `YdbTransientExceptionDetector`'s existing reflection-based lookup is reused, so this adds no dependency on `Ydb.Sdk` types.

## Verification — and its limits

Worth being explicit, since this is the kind of fix that invites a red→green demand: **the race is timing-dependent and I could not reproduce it deterministically.** It only ever appeared under a full parallel run's load, never in isolation. So there is no red-first test here, and I did not add a synthetic one: `Ydb.Sdk.Ado.YdbException` has only internal constructors, so faking the failure would mean reflecting into both the SDK's internals and this file's private method — a test that would prove "the loop loops" while breaking the moment the SDK changes shape.

What is verified: all 11 YDB schema-provider tests pass against a real YDB container, three runs in a row; `Source/LinqToDB` builds clean in Release for `net10.0`, `netstandard2.0` and `net462`. The happy path is unchanged — the retry only engages on a caught `SchemeError`.

## Follow-up

Filed upstream, since this turned out to be a driver-level bug rather than something a consumer should have to work around: [ydb-platform/ydb-dotnet-sdk#687](https://github.com/ydb-platform/ydb-dotnet-sdk/issues/687).

A minimal `Ydb.Sdk`-only repro (no linq2db) reproduces it deterministically under concurrent DDL: `YdbSchema.GetColumns` lists tables and then calls `DescribeTable` per name as two separate, non-atomic steps, so a concurrent `DROP`/`CREATE` between the two steps races regardless of anything done on our side.

Keeping the retry here for now — a local run against real concurrency fails without it, and there's no telling how long the upstream fix will take. Added a `TODO` in `GetSchemaWithRetry` linking the issue so we remember to re-check once it's fixed and released upstream.

### DIFF
--- a/Source/LinqToDB/Internal/DataProvider/Ydb/YdbSchemaProvider.cs
+++ b/Source/LinqToDB/Internal/DataProvider/Ydb/YdbSchemaProvider.cs
@@ -4,6 +4,7 @@ using System.Data;
 using System.Data.Common;
 using System.Globalization;
 using System.Linq;
+using System.Threading;
 
 using LinqToDB.Data;
 using LinqToDB.Internal.SchemaProvider;
@@ -25,7 +26,7 @@ namespace LinqToDB.Internal.DataProvider.Ydb
 
 		protected override List<TableInfo> GetTables(DataConnection dataConnection, GetSchemaOptions options)
 		{
-			var dt     = dataConnection.OpenDbConnection().GetSchema("Tables");
+			var dt     = GetSchemaWithRetry(dataConnection.OpenDbConnection(), "Tables");
 			var tables = new List<TableInfo>();
 
 			foreach (DataRow row in dt.Rows)
@@ -55,7 +56,7 @@ namespace LinqToDB.Internal.DataProvider.Ydb
 
 		protected override List<ColumnInfo> GetColumns(DataConnection dataConnection, GetSchemaOptions options)
 		{
-			var dt      = dataConnection.OpenDbConnection().GetSchema("Columns");
+			var dt      = GetSchemaWithRetry(dataConnection.OpenDbConnection(), "Columns");
 			var columns = new List<ColumnInfo>();
 
 			foreach (DataRow row in dt.Rows)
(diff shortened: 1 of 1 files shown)


MaceWindu:
This is provider bug and should be fixed in Ydb.Sdk, not here

igor-tkachev:
Agreed, this is a driver-level bug — the retry here is a workaround, not a fix. I confirmed it with a minimal repro against `Ydb.Sdk` directly (no linq2db involved): `YdbSchema.GetColumns` lists tables and then calls `DescribeTable` per name as two separate, non-atomic steps, so concurrent DDL between the two races regardless of anything we do on our side. Filed upstream: https://github.com/ydb-platform/ydb-dotnet-sdk/issues/687

I'm keeping the workaround here for now, since a local run against real concurrency fails without it, and there's no telling how long the upstream fix will take. Added a `TODO` in `GetSchemaWithRetry` linking the issue, so we remember to re-check and drop the retry once it's fixed and released upstream.

PR #5773, opened 2026-08-12
Add native Oracle 12 identity DDL

### LINKED ISSUE #1879: Oracle 12+ dialect
It would be nice to be able to opt-in a "Oracle 12" dialect when using the Oracle provider.

Some improvements that I can think of:

- when creating tables, AS GENERATED ALWAYS WITH IDENTITY can be used instead of the sequence + trigger combo. It's much cleaner.

- Take / Skip is now supported in ORDER BY OFFSET / FETCH, which results in much cleaner SQL queries. There initially was poor query plan evaluation when using this syntax but it was later fixed by Oracle.

- JSON support! We can query column containing JSON with the `JSON_*` family of functions.

- There is a new aggregate function APPROX_COUNT_DISTINCT

Pre-Oracle 12 stuff I'd love to see added as well:

- `REGEXP_*` family of functions (#698)

jogibear9988:
Generate Oracle 12+ identity columns with GENERATED BY DEFAULT AS IDENTITY instead of auxiliary sequence and trigger objects. Preserve explicit identity inserts for bulk-copy KeepIdentity scenarios while leaving Oracle 11 on the legacy sequence/trigger implementation.

Make drop generation skip legacy identity objects for the native dialect and reset identity-backed tables after truncate with ALTER TABLE ... START WITH 1.

Update the existing Oracle native-identity bulk-copy fixture and add SQL-builder coverage for Oracle 12 create, drop, and truncate behavior plus Oracle 11 compatibility.

Related to #1879


MaceWindu:
Synced with master and reviewed. The PR was red on every Oracle version — `test-oracle` build 22842 failed on 12c/18c/19c/21c/23c with ~48 failures each — so I've pushed 15 commits, one per concern. Verified locally against Oracle 11 and Oracle 21: 436 tests, 0 failures. No CI test run has covered this HEAD yet.

## Defects

**`INSERT ALL` is incompatible with a native identity — this is most of the CI failures.** Oracle's defaults are `BulkCopyType.MultipleRows` + `AlternativeBulkCopy.InsertAll`, which emits `INSERT ALL … INTO t (…) VALUES (…) ×N SELECT * FROM dual`. In a multitable insert the identity generator advances once per row of the *driving query*, and `dual` returns one row, so every `INTO` branch received the same value. A three-row `BulkCopy` stored ids `[1, 1, 1]`; tables with a unique constraint on the identity column raised `ORA-00001` instead (`PK_Position`, `PK_Categories`, `PK_SubData2`, `PK_Issue4702Table`). The trigger this replaces fired per row, so it never met the restriction — and `TestNativeIdentityBulkCopy` already carried an `Assert.Inconclusive` for exactly `MultipleRows + InsertAll + KeepIdentity=false`, a combination this change makes the default for every identity table on 12+. Fixed by falling back to the single-table `INSERT … SELECT … UNION ALL` form when the server generates the value; `INSERT ALL` is kept when `KeepIdentity` is set or the table opted out.

**`OracleTests.ItentityColumnTable` spelled the identity clause in `DbType`.** With the builder emitting one as well the column came out carrying it twice, so `CREATE TABLE` stopped parsing — `ORA-00907` on 12/18/19/21, `ORA-03076` on 23. `DbType` carries the column type only, so the fixture now relies on `IsIdentity`.

**`TruncateTableTests.Issue2847Test` drops `SIDENTITY_TestIdTrun` by hand.** Under a native identity that sequence does not exist, so the `DROP SEQUENCE` itself raised `ORA-02289` and both cases failed before reaching the truncate. Restricted to Oracle 11, with a new Oracle 12+ case built on an opted-out table.

**The truncate reset emitted `GENERATED AS IDENTITY`** — unqualified, which in `CREATE TABLE` grammar means `ALWAYS` — while `CREATE TABLE` emits `GENERATED BY DEFAULT`. Oracle keeps the existing generation type when identity options are changed through `ALTER TABLE … MODIFY`, so the behaviour was correct; verified by round-tripping an explicit value after a reset. The two sites reading differently invited the opposite conclusion, so they now match.

## The old behaviour is still available

Native identity is now the Oracle 12+ *default* rather than the only option. A `SequenceName` on the identity member keeps the auxiliary sequence and trigger for that table: three create commands, sequence and trigger drop, `ALTER SEQUENCE` reset on truncate. That attribute already means "this identity is fed by this named sequence" — `ColumnDescriptor` infers `IsIdentity` from it — so this is its literal reading rather than a new switch, and it gives schemas created by an earlier version a per-table upgrade path. The sequence name and schema now come from the attribute, so the DDL, the trigger body and `GetIdentityExpression` name the same object; previously `CREATE` emitted `SIDENTITY_<table>` while `INSERT` read the attribute's sequence.

Unifying them surfaced a defect: `QueryRunner.Insert` and `QueryRunner.InsertWithIdentity` both put `GetIdentityExpression` into the statement for an identity field, so the INSERT already carried `SEQ.nextval` and the trigger drew a second value from the same sequence — two inserts stored ids 2 and 4 instead of 1 and 2 on Oracle 21. The generated trigger now assigns only when the row supplied no value, which also makes `KeepIdentity` work on that path.

## Behaviour changes worth recording

- `BulkCopyOptions.KeepIdentity = true` is now honoured on Oracle 12+. It never was on a linq2db-created table: the trigger assigned unconditionally and discarded the caller's value. Oracle 11 still does — pinned by `TestLegacyIdentityIgnoresKeepIdentity`.
- As with any native identity, an explicitly supplied value does not advance the generator, so a `KeepIdentity` copy followed by a generated insert can collide — pinned by `TestNativeIdentityKeepIdentityLeavesGeneratorBehind`.

## Coverage

Runtime, on `AllOracle12Plus`: default bulk copy produces distinct ids; truncate-reset restarts at 1 and leaves the generation mode alone; a `SequenceName` table round-trips through the legacy path; `InsertWithIdentity` consumes one value per row; a `Nullable<T>` identity member still yields a `NOT NULL` column. Builder-level: create, drop, drop-if-exists and truncate-reset for Oracle 12 native, Oracle 12 opted-out and Oracle 11.

Two findings did not survive checking and are recorded here so they are not re-raised. A nullable identity column cannot emit a `NULL` constraint conflicting with the implicit `NOT NULL`, because `ColumnDescriptor.AnalyzeCanBeNull` returns `false` for any identity member whatever the CLR type. And `ALTER TABLE … MODIFY (… GENERATED AS IDENTITY …)` does not switch an existing `BY DEFAULT` column to `ALWAYS`, so no truncate ever changed the generation mode.

<sub>Review and changes produced with agent assistance; every claim above is backed by a local run or a cited line, and the failures quoted come from build 22842.</sub>

Issue #5265, opened 2025-12-18
Stack overflow


livedys-debug:
### Description

Perhaps a duplicate of the previous one, but:

### Reproduction Steps

var rawOutputQuery = db.GetTable<SomeType>();
var outputItems = rawOutputQuery.ToArrayAsync();

SomeType is complex type with many associations && calculated properties

[StackTrace.txt](https://github.com/user-attachments/files/24237281/StackTrace.txt)

### Linq To DB version

6.1

### .NET version

8.0

### Database provider

Postgre 15

sdanyliv:
Any chance to see the LINQ query?

livedys-debug:
> Any chance to see the LINQ query?
`
var rawOutputQuery = db.GetTable<TSomeTypeWithManyAssociationsAndExpressionProperties>();
var outputItems = rawOutputQuery.ToArrayAsync();
`
Или что ты имеешь ввиду?
В версии 5.4.1.9 и более ранних работало без проблем

sdanyliv:
Well, then we need model classes with associations. Query that you have shown tells nothing.

MaceWindu:
Reopening, we plan to implement additional fixes

MaceWindu:
As no test example provided, we just implemented some improvements in this area

PR #5774, opened 2026-08-12
Avoid compiling a lambda per `default(T)` during query build (~11x on deep `LoadWith` chains)

MaceWindu:
Two independent query-build optimisations, plus the benchmark that found them. No linked issue - #5265 (the stack overflow on this shape) is closed; this is a separate performance concern on the same shape.

## 1. `default(T)` no longer compiles a lambda

`IsSimpleEvaluatable` admitted the BCL's `DefaultExpression` (`NodeType == Default`) but not linq2db's own `DefaultValueExpression`, whose `NodeType` is `Extension`. Every `default(T)` therefore fell through to `Expression.Lambda(expr).CompileExpression().DynamicInvokeExt()`.

Building a 13-level `LoadWith` chain compiled **38460** lambdas, 38220 of them for the same `Default(ChainNode)` node. A CPU profile put 81% of the building thread inside `BuildConstant` -> `EvaluateExpression`, with a second core saturated in `DynamicResolver+DestroyScout.Finalize` destroying the resulting `DynamicMethod`s.

The fix reduces the node instead. `Reduce()` returns the same constant the compiled lambda produced - including a mapping schema's custom default - so it is exact, not an approximation.

## 2. Weak-join dependency scan probes the enclosing query first

`ResolveWeakJoins` asked whether a removable join's sources are referenced anywhere in the statement, walking from the statement root once per candidate. The reference that keeps a join is almost always in the join's own query, so the walk paid a full descent to reach it.

This is exact, not heuristic: `selectQuery` is reachable from `_rootElement` without passing through the join, and reference dedup never drops a node from a closure, so a local hit is necessarily a root hit; a local miss falls through to the original scan. Instrumentation found 6961 of 7030 scans hit locally, and **zero** cases of a local hit without a root hit across 7117 scans.

**This reduces work, not wall-clock time** - weak-join scanning is under 1% of query build for these shapes, so no timing improvement is claimed for it.

## Measurements

| | before | after |
|---|---|---|
| lambda compiles per build (deep chain) | 38460 | 0 |
| deep-chain query build | ~4470 ms | ~410 ms |
| weak-join scan work, whole suite | - | -58% |
| weak-join scan work, deep chain | - | -71% |

## Verification

- Full suite, SQLite.MS, net10.0: **10373 passed, 0 failed, 30 skipped** - identical to the pre-change reference.
- `StackUseTests.EagerLoadProjection` canary on PostgreSQL.16: passes (200 KB stack ceiling holds).
- Release builds clean on net10.0, netstandard2.0, net462.
- Generated SQL byte-identical for both benchmark shapes.

### DIFF
--- a/Source/LinqToDB/Internal/Expressions/ExpressionEvaluator.cs
+++ b/Source/LinqToDB/Internal/Expressions/ExpressionEvaluator.cs
@@ -29,6 +29,9 @@ namespace LinqToDB.Internal.Expressions
 				{ NodeType: ExpressionType.Default } => true,
 				{ NodeType: ExpressionType.Constant } => true,
 
+				// our own default-value node reduces to a constant, so it needs no compiled lambda
+				DefaultValueExpression => true,
+
 				MemberExpression { NodeType: ExpressionType.MemberAccess } member =>
 					member.Member.MemberType is MemberTypes.Field or MemberTypes.Property
 					&& IsSimpleEvaluatable(member.Expression),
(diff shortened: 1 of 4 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Queries with deep `LoadWith` association chains build substantially faster on first execution.


_(none)_


---
<sub>Generated from commit `8441103`.</sub>

PR #5783, opened 2026-08-14
Mark Sql.Window functions as server-side only

### LINKED ISSUE #5782: New Window API's COUNT(*) translation breaks when used with another window function
### Description

New Window API's COUNT(*) translation breaks when used with another window function 

### Reproduction Steps

`
        var sql = query1
            .Select(x => new
            {
                Count = Sql.Window.Sum(x.Column, w => w) == 4 && Sql.Window.Count(w => w) == 3
            })
            .ToSqlQuery();
` 
breaks while
`
        var sql = query1
            .Select(x => new
            {
                Count = Sql.Window.Sum(x.Column, w => w) == 4 && Sql.Window.Count(x.Column, w => w) == 3
            })
            .ToSqlQuery();
`  
works fine.
Also `Sql.Window.Count(1, w => w)` breaks, but `Sql.Window.Count(Sql.AsSql(1), w => w)` is working

### Exception details

Unhandled exception. LinqToDB.ServerSideOnlyException: 'Count' is server-side API.
   at LinqToDB.WindowFunctionBuilder.Count(IWindowFunction window, Object argument, Func`2 func) in /_/Source/LinqToDB/Sql/Sql.Window.cs:line 991

### Regression?

No, newly introduced API

### Linq To DB version

6.4.0

### .NET version

10.0.300

### Database

_No response_

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
`Sql.Window.Count(w => w) == 3` inside a boolean expression threw `ServerSideOnlyException: 'Count' is server-side API` at execution time.

The window API is translated by member translators only and carried no `[ServerSideOnly]` marker, so `CanBeEvaluatedOnClient` returned `true` for any call that referenced no table column. The enclosing comparison was folded into a query parameter and the throwing stub ran at execution time — the translator was never asked.

Not COUNT-specific: `RowNumber(w => w.OrderBy(1))`, `Rank`, `NTile(4, …)`, `Lead(1, …)` and `Sum(1, w => w)` failed the same way.

Fix: `[ServerSideOnly]` on all 104 public methods of `WindowFunctionBuilder`.

Test: `WindowFunctionsTests.FunctionWithoutColumnReference` — red before, green after, on both the direct and LinqService paths.

Fixes #5782


igor-tkachev (dismissed):
Looks good overall. Minor test coverage note: the fix applies `ServerSideOnly` to the whole Window API, but the regression test effectively verifies the no-column case only for `Count` (`Sum` still references `t.IntValue`). It would be useful to cover a couple of other no-column paths such as `Sum(1, ...)` and `RowNumber(...OrderBy(1))`.

Also, consider a small reflection test that ensures all public `WindowFunctionBuilder` server-side stubs have `[ServerSideOnly]`. That would protect this invariant for newly added window methods without adding functional tests for every overload.

MaceWindu (dismissed):

MaceWindu:
Two follow-up commits pushed, plus a merge of current `master`.

**`26a5c24bb` — give `FunctionWithoutColumnReference`'s assertions a failure mode.**

`AllRows` compared `COUNT(*) OVER ()` against `data.Length`. An array-length access on a captured array is evaluated client-side from the already-projected `COUNT(*) OVER ()` column, so the emitted SQL carried no `COUNT(*) OVER () = ...` term at all, and the assertion held regardless of how that comparison translated — for all 53 providers the third column was just `SUM(...) OVER () > 0`. It now compares against a local `int`, which parameterizes and keeps the comparison server-side (`COUNT(*) OVER () = @rowCount AND SUM([t].[IntValue]) OVER () > 0`, the shape reported in #5782), and a `WrongCount` member whose count conjunct is false gives the assertion something to fail on.

**`3e64b23dc` — cover the column-free `Sum` overload.**

`Sum(1, w => w)` is listed in the description as failing the same way as `Count(w => w)`, and goes through a different translator path, but nothing exercised it. Supplying no `OrderBy` or frame keeps it clear of the aggregate-window gate, so it needs no `[ThrowsForProvider]`.

Both verified on `SQLite.MS`, `DuckDB`, `PostgreSQL.17` and `SqlServer.2022.MS`, direct and LinqService.

**`RowNumber(w => w.OrderBy(1))` was tried and left out.** It emits `ROW_NUMBER() OVER (ORDER BY 1)`, which SQL Server rejects with error 5308 — *"Windowed functions, aggregates and NEXT VALUE FOR functions do not support integer indices as ORDER BY clause expressions"* — so including it here would need `[ThrowsForProvider]` for every SQL Server version. PostgreSQL, DuckDB and SQLite accept it, but treat the literal as a constant rather than an ordinal (`ORDER BY 2` against a one-column projection also succeeds there), so the window ends up with no effective ordering. Filed separately as #5806.

On igor-tkachev's two review points:

- **Coverage of other no-column paths** — `Sum(1, ...)` addressed by `3e64b23dc`; `RowNumber(...OrderBy(1))` is the #5806 case above.
- **Reflection test pinning `[ServerSideOnly]` across `WindowFunctionBuilder`** — not added. #5788 covers the same invariant as an analyzer instead: both directions, and for user-defined extension methods as well as linq2db's own source. That also catches `Sql.GroupBy`'s `Rollup` / `Cube` / `GroupingSets`, which throw `ServerSideOnlyException` today carrying neither form of the marker, unlike `Sql.Grouping` beside them.

Baselines: the test's SQL changed, so its 53 baseline files need regenerating — linq2db/linq2db.baselines#2057 is stale until the next run.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A `Sql.Window` function that references no table column threw `ServerSideOnlyException: 'Count' is server-side API` at execution instead of translating — `Sql.Window.Count(w => w) == 3`, and likewise `Count(1, …)`, `Sum(1, …)`, `RowNumber`, `Rank`, `NTile` and `Lead`. All window functions are now recognised as server-side, so they translate as written. ([#5783](https://github.com/linq2db/linq2db/pull/5783))


`Sql.Window` functions that reference no column — `Sql.Window.Count(w => w) == 3` — now translate instead of throwing `ServerSideOnlyException`.


---
<sub>Generated from commit `4f393c0`.</sub>

PR #5784, opened 2026-08-15
Fix DB2 tests leaving ~90k rows in AllTypes, revert the ORDER BY workaround

### LINKED ISSUE #5765: DB2: AllTypes table accumulates junk rows across test runs (ResetAllTypesIdentity never deletes on DB2)
## Problem

`AllTypes` on DB2 accumulates rows that are never cleaned up. `ResetAllTypesIdentity`'s DB2 branch (`Tests/Base/TestBase.Identity.cs`) only reseeds the identity counter (`RESTART WITH 3`) — it never deletes rows, unlike the Informix branch, which has an explicit `DELETE FROM AllTypes WHERE ID > lastValue`. The class-level comment documents this as intentional:

> We intentionally don't remove extra records and allow it to fail in that cases to allow us detect tests that forget to cleanup and leave junk records in those tables

So a leftover row is meant to surface as a test failure (typically via an identity/row-selection mismatch), pointing at whichever test forgot to clean up after itself.

## Evidence

CI build [22698](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=22698) (commit 5b6fcaaf9) failed `DataProviderTests.TestDataTypes("DB2")`'s `TestType` assertions with rows at IDs far above the expected 1/2:

```
Assert.That(id, Is.EqualTo(2))   Expected: 2   But was: 183642
Assert.That(id, Is.EqualTo(2))   Expected: 2   But was:  83641
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was:  50773   (x4)
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was: 182126
Assert.That(id, Is.EqualTo(1))   Expected: 1   But was: 216310
```

At least five distinct surviving IDs up to 216310 — `AllTypes` on DB2 had accumulated roughly 216k inserts' worth of rows never cleaned up. This is a real leak somewhere in the DB2 test path, not one isolated leftover row colliding with a magic number.

(Surfaced via [#5754](https://github.com/linq2db/linq2db/pull/5754), which added `ORDER BY ID` to `DataProviderTestBase.PassNullSql`/`PassValueSql` — see that PR's discussion for the review thread this issue was split out of. The `ORDER BY` change itself is a separate, independent fix for non-deterministic row selection and is not reverted by this issue; it's tracked here because silencing the symptom removes the only signal that was catching this leak.)

## What's needed

1. Identify the test (or tests) that insert into the shared `AllTypes` table on DB2 without cleaning up — the accumulation is large enough (~216k rows) that it's plausibly a single high-iteration or repeatedly-run test/fixture rather than a one-off leftover.
2. Consider adding an explicit `COUNT(*)` / `MAX(ID)` sanity assertion on `AllTypes` in fixture setup (suggested in the #5754 review thread), so a future recurrence fails deterministically and immediately, rather than depending on which row the optimizer happens to pick without `ORDER BY`.
3. Once the leaking test is fixed, DB2's `ResetAllTypesIdentity` could optionally gain the same explicit `DELETE FROM AllTypes WHERE ID > lastValue` the Informix branch already has, if leaving no ambiguity is preferred over the current detect-via-failure design.

## Environment

- Provider: DB2 LUW 11.5
- Surfaced under both CI (`test-all`) and local parallel test runs; not yet confirmed whether reproduction requires parallel execution against the shared DB2 container or also occurs sequentially.

MaceWindu:
Fixes #5765.

`DataProviderTestBase.TestType` resolves `SELECT ID FROM AllTypes WHERE <field> IS NULL / = @p` and takes the first row, so any leftover row displaces the seeded rows 1/2 and `DB2Tests.TestDataTypes` fails with ids like 50773 / 182126 / 216310 in place of 1 / 2.

## Who leaves the rows

A local DB2 run instrumented to report every change in the leftover row count named the writers exactly:

```
grew   0->1000      MiniProfilerTests.TestDB2("DB2",Raw)
grew   1000->1997   MiniProfilerTests.TestDB2("DB2",MiniProfiler)
grew   1997->46462  DB2Tests.BulkCopyProviderSpecific("DB2")
grew   46462->90927 DB2Tests.BulkCopyProviderSpecificAsync("DB2")
shrank 90927->0     TransactionTests.AutoRollbackTransaction("DB2")
```

~90 900 rows per DB2 run, which is what drives the identity burn behind the ids seen in CI.

Both cleanups look correct and delete nothing:

- `TestDB2` bulk-copies rows as `new ALLTYPE { ID = 2000 + n }` and deletes `ID >= 2000`, but `AllTypes.ID` is `GENERATED ALWAYS AS IDENTITY` — DB2 discards the supplied ids and assigns its own low values, so the predicate matches no row.
- `BulkCopyTest` inserts 100 001 rows and deletes `SMALLINTDATATYPE >= 5000`, where the column is set to `(short)(5000 + n)`. That cast wraps negative once `n` passes 27 767, so every row beyond the wrap escapes the filter.

Both now capture the id high-water mark before the copy and delete `ID > maxId` in the `finally`, which is immune to either trap.

The rows were invisible after the fact because `RestoreBaseTables.Dispose` deletes `ID > 2`, so whichever test uses it next wipes them — which is why a post-run check finds a clean table and this looked like it did not reproduce locally.

## Reverts the ORDER BY workaround

Also reverts the `DataProviderTestBase` `ORDER BY ID` changes from #5754 (test half only; the DB2 `AUTO_STMT_STATS` setting it shipped alongside is kept). That change made the lookup pick the lowest-id match so a leftover row could not win, which treated the symptom — with the writers fixed there is nothing to disambiguate, and the test SQL goes back to expressing what it actually asserts.

## Scope

Only the DB2 sites are changed. `MiniProfilerTests` uses the same `delete ID >= 2000` cleanup for several other providers, but those allow identity insert, so the supplied ids land and the cleanup works; no leftovers were measured for them.

## Testing

Test-only change, verified by build. The measurement above came from a full local DB2 leg (9141 tests); the acceptance check on CI is `DB2Tests.TestDataTypes` staying green with no leftover rows reported.

### DIFF
--- a/Tests/Base/DataProvider/DataProviderTestBase.cs
+++ b/Tests/Base/DataProvider/DataProviderTestBase.cs
@@ -19,14 +19,9 @@ namespace Tests.DataProvider
 		{
 			// number of parameters to create for providers with unnamed parameters
 			paramCount = 1;
-			// ORDER BY ID: AllTypes is a shared fixture other tests also insert into (deliberately never
-			// cleaned up here - see TestBase.Identity.cs's ResetAllTypesIdentity - so a leftover row with a
-			// duplicate value is a real, if rare, possibility). Only the lookup's chosen row matters (Execute
-			// reads a single scalar), so ordering by ID makes it deterministically the lowest-ID (canonical)
-			// match instead of whichever row the engine happens to return first.
-			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p ORDER BY ID";
+			return "SELECT ID FROM {1} WHERE @p IS NULL AND {0} IS NULL OR @p IS NOT NULL AND {0} = @p";
 		}
-		protected virtual string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = @p ORDER BY ID";
+		protected virtual string  PassValueSql(DataConnection dc) => "SELECT ID FROM {1} WHERE {0} = @p";
 
 		protected T TestType<T>(DataConnection conn, string fieldName,
 			DataType dataType          = DataType.Undefined,
--- a/Tests/Linq/Data/MiniProfilerTests.cs
+++ b/Tests/Linq/Data/MiniProfilerTests.cs
@@ -646,6 +646,10 @@ namespace Tests.Data
 			}
 
 			// bulk copy
+			// AllTypes.ID is GENERATED ALWAYS, so the IDs set below are discarded and the server assigns
+			// its own - clean up by the high-water mark rather than by a value we think we inserted.
+			var maxId = db.GetTable<ALLTYPE>().Select(_ => _.ID).Max();
+
 			try
 			{
 				db.BulkCopy(
@@ -656,7 +660,7 @@ namespace Tests.Data
 			}
 			finally
 			{
-				db.GetTable<ALLTYPE>().Delete(p => p.ID >= 2000);
+				db.GetTable<ALLTYPE>().Delete(p => p.ID > maxId);
 			}
 
 			// just check schema (no api used)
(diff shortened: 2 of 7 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `ca03a0b`.</sub>

PR #5785, opened 2026-08-15
Add [QueryCacheTest] and mark the query-cache-dependent tests

MaceWindu:
Split out of #5614 so it can be reviewed and CI'd on its own, and so that PR's diff stops carrying 31 test files.

145 tests across 31 files depend on process-global query-cache state — they assert exact `GetCacheMissCount` deltas, compare query-object identity across two executions, or clear the cache themselves. Nothing said so at the call site, so the dependency was invisible until one of them failed.

## The attribute

`[QueryCacheTest]` declares it, and carries the two guarantees such a test needs:

- derives from `ParallelizableAttribute(ParallelScope.None)`, so the test never runs concurrently with another — `Query<T>` reaches `QueryCache.Default` statically, and a concurrent test's compilation perturbs the counters;
- lifts `QueryCache.MaxEntriesOverride` for the duration of the test and restores the configured value afterwards, so nothing can be evicted from under it while it runs.

It deliberately does **not** clear the cache. NUnit fires `ITestAction` per test *case*, and many of these tests are parameterised by iteration (`[Values(1, 2)]`) where the first case warms the cache and the second asserts it was hit — a clear lands between them and fails them by construction. The configured cap is captured once, before the first test lifts it, so a mispaired `Before`/`After` can never leave a whole run uncapped.

## Behaviour on master

Both effects are inert here: nothing configures `MaxEntriesOverride` on master, and these tests do not currently run in parallel. The value today is that the dependency is declared where the tests live. It becomes load-bearing in #5614, which runs the suite in parallel and caps the cache at 100 entries on netfx / 32-bit legs — the configuration under which these tests fail without it.

## Selection

Scripted, not hand-picked: a test qualifies when its body reads `GetCacheMissCount`, `CacheMissCount`, `Query.ClearCaches` or `Query<T>.GetQuery`. Heaviest files: `EnumerableSourceTests` (26), `Issue3148Tests` (20), `ParameterTests` (11), `PostgreSQLArrayTests` (10), `FromSqlTests` (10).

Two groups are worth calling out because they don't read as cache tests at a glance:

- `FromSqlTests.TestQueryCaching_*` assert query-object identity across two executions rather than a miss count, so they depend on the entry surviving between them;
- `MappingTests.MappingTypingByConstant_FromEnumerable_*` call `Query.ClearCaches` themselves, which perturbs any concurrently running cache-counter test.

## Testing

Build-verified. In #5614, where the attribute is load-bearing, the equivalent state runs the full SQLite suite (4 providers, cache capped at 100) at 40135 tests / 0 failures, against 279 failures with plain `[NonParallelizable]` and 256 with a clearing variant.

### DIFF
--- a/Tests/Base/Attributes/QueryCacheTestAttribute.cs
+++ b/Tests/Base/Attributes/QueryCacheTestAttribute.cs
@@ -1,0 +1,69 @@
+using System;
+
+using LinqToDB.Internal.Linq;
+
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+namespace Tests
+{
+	/// <summary>
+	/// Marks a test that depends on process-global query-cache state - typically asserting exact
+	/// <c>GetCacheMissCount()</c> deltas, query-object identity across two executions, or clearing the
+	/// cache itself. Such a test needs two things that ordinary tests do not:
+	/// <list type="number">
+	/// <item>no other test may run concurrently, because <c>Query&lt;T&gt;</c> reads the process-wide
+	/// <see cref="QueryCache.Default"/> and a concurrent test's compilation perturbs the counters;</item>
+	/// <item>nothing may be evicted from the cache while it runs, because netfx / 32-bit runs cap the
+	/// cache at 100 entries (see TestsInitialization) and a suite that has already filled it evicts the
+	/// entry the test expects to hit - which reads as one extra compilation.</item>
+	/// </list>
+	/// </summary>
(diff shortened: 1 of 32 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `6b7796d`.</sub>

PR #5795, opened 2026-08-18
Test-infrastructure and CI cleanups split out of #5614

MaceWindu:
Split out of #5614 (parallel test execution). Everything here stands on its own without the parallel dispatcher, so it can be reviewed and CI'd independently.

## CI

- Remove the `test-metrics` job: the `SqlServer2022_Metrics` matrix entry, the `[metrics]` db_filter (and the not-metrics guard it forced onto the plain SqlServer2022 entry), the pipeline mapping, the four per-TFM configs and their solution-file entries. The harness side (`TestConfiguration.StoreMetrics` / `BaselinesWriter.WriteMetrics`) is untouched.
- Generate each workflow template's per-TFM block (extract → configure → install SDK → setup → test → cleanup) from a `tfms` list instead of three or four copies. On Windows the x86/x64 duplication collapses too: `test-jobs.yml` resolves the leg's `x86` flag into `$(arch)`, so each step exists once. **Side effect worth noting:** master's net10 Windows section was missing the x64 `script_local` setup step every other TFM has — the loop restores it.
- Probe the heaviest resident processes and free the image's idle desktop/telemetry ones before the test run (637 MB measured on the Windows image); probe only on Linux. `WindowsTerminal` is deliberately excluded and the reason is documented in-file — the agent runs inside it, so killing it abandons every job in the build. No services are stopped (`Stop-Service -Force` has no `-NoWait` and hung jobs inside the step).
- Shorter job display names (`Win`/`Lin`/`Mac`/`CLI`); 300-minute timeout for `full_run` legs, which run every TFM's main suite rather than net10 alone.

## Firebird

Firebird 3/4/5 move from the third-party `jacobalberty/firebird` image to the project's own `firebirdsql/firebird`, which carries current point releases. The env-var contract differs (`FIREBIRD_ROOT_PASSWORD` / `FIREBIRD_USE_LEGACY_AUTH`) and the created database lands under `/var/lib/firebird/data`, so the connection strings move with it. Covers the CI scripts, the local `Data/Setup Scripts` helpers and `DataProviders.json`. Firebird 2.5 stays on `jacobalberty` — out of support, no official image.

## Tests

- PostgreSQL version exclusions for four tests plus one read-matrix row, each with the server-side reason in the commit message (`unknown to text` before PG 10; RANGE value offsets need PG 11; the `line` type has no I/O functions before 9.4).
- Query cache capped at 100 entries on x86 and netfx legs — the cap `QueryCacheTestAttribute`'s own documentation already refers to. `L2DB_TEST_QUERYCACHE` overrides it per leg.
- `LINQ2DB_ASSERT_STATE=1` now enables the shared-table consistency check, which was hard-wired off behind a `false` literal.
- Five more cache-sensitive tests marked `[QueryCacheTest]`.
- `TestIQueryableParameterEvaluationMultiThreaded`: per-invocation state instead of a shared `_params` instance field, and awaiting instead of blocking. The remote contexts route sync calls through `SafeAwaiter.Run`, so 30 blocking workers held 60 pool threads while the in-process host serving them needed its own — the test sat for 11 minutes on one CI leg and was hangdump-killed on another.
- `SQLiteTests.CreateDatabase` uses a per-context database file instead of one shared `TestDatabase.sqlite` across both SQLite providers.
- The OrmBattle Northwind cache is published only once fully wired (`ComplexAllTest` compares navigation properties by reference).
- Entity-descriptor callbacks scoped to their own work: `LowerCaseMappingTest` passes its callback to `GetEntityDescriptor(type, callback)` and no longer mutates the process-wide slot; `OnEntityDescriptorCreatedTest` keeps the global slot (that is what it tests) but records only descriptors built by its own context's mapping schema, and serializes its per-provider cases with a static lock so one case cannot replace or clear another's handler.
- Both environment variables documented in `CONTRIBUTING.md`.
- DuckDB identity-reset comments moved out of the statement arrays; `MaxCpuCount=1` dropped from `.runsettings` (a VSTest setting, inert under Microsoft.Testing.Platform).

## Revert

Reverts #5757 (EF PostgreSQL invalid-database recovery + `CommandTimeout(120)`). Both halves were tuned for the host contention a full parallel run produces, and neither addressed the interruption that produced the invalid-database state; if it recurs the fix belongs at the cause.

## Verification

`LowerCaseMappingTest` and `OnEntityDescriptorCreatedTest` run green on SQLite.Classic + SQLite.MS (net10.0). The rest — pipeline templates, provider setup scripts, the query-cache cap's effect on 32-bit legs — is CI-verified only.

## Not taken from #5614

`Tests/EntityFrameworkCore/TestBase.cs`'s `[SetUp]` hook calls `CustomTestContext.Begin(bool, string?)`, which exists only on the feature branch, so it stays there.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `a714461`.</sub>

PR #5800, opened 2026-08-19
CI: test PostgreSQL 9.2-12

MaceWindu:
CI covered PostgreSQL 13-19 only, so the six older versions linq2db still declares providers for — `ProviderName.PostgreSQL92`/`93`/`95` and `TestProvName.PostgreSQL10`/`11`/`12` — were never exercised and have no SQL baselines.

Each gets its own job, cloned from the existing `pgsql13` pattern:

- `Build/Azure/scripts/pgsql{92,93,95,10,11,12}.sh` + `mac.` variants — `postgres:<tag>` on port 5432
- `Build/Azure/net{80,90,100}/pgsql{92,93,95,10,11,12}.json`
- six `test-matrix.yml` entries under `[postgresql.all]`, linux + macos, net80/net90/net100
- six `AzureConnectionStrings` entries in `DataProviders.json`
- the 30 new files registered in `linq2db.slnx`

`test-postgresql` goes from 7 jobs to 13.

The tests these versions cannot pass are already excluded — the `unknown`→`text` coalesce cases, the `RANGE` value-offset frames and the `line`-type read-matrix row were gated in #5795.

`Build/Azure/README.md`: the job table already claimed 11 and 12 were covered, which was never true; it now matches what runs. The per-provider table's "not tested by CI (EOL)" notes are down to the EOL marker.

### Why now

#5614 consolidates these six into a single `pgsql1` job. Landing the coverage first means their baselines are generated against master, so #5614's own baselines PR carries only the differences its changes actually cause instead of ~6 versions' worth of first-time captures.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `79dde7f`.</sub>

PR #5801, opened 2026-08-19
Fix LoadWith filter closing over an optional value being cached across closure states

### LINKED ISSUE #5793: Query cache ignores closure state in a `LoadWith` filter lambda, so the first call's eager-load SQL is reused
## Summary

When a `LoadWith` filter closes over an **optional** value (`set == null || set.Contains(...)`), the compiled-query
cache does not key on that closure state. Whichever call compiles first wins, and later calls with different
closure state silently get its eager-load SQL. Both directions produce wrong data:

- unfiltered first → the **filtered** call returns unfiltered rows;
- filtered first → the **unfiltered** call gets `WHERE 1 = 0` and returns nothing.

The same idiom in a plain top-level `Where` is cached correctly, and `DisableQueryCache = true` makes both correct
— so the translation is fine and only the cache lookup is wrong.

**Regression:** introduced in **6.2.0**. Correct in 6.0.0 and 6.1.0; broken in 6.2.0, 6.2.1, 6.3.0 and 6.4.0.

## Repro

```csharp
[Table("Headers")]
public class Header
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string No { get; set; } = null!;

    [Association(ThisKey = nameof(No), OtherKey = nameof(Line.HeaderNo))]
    public List<Line>? Lines { get; set; }
}

[Table("Lines")]
public class Line
{
    [Column] public string HeaderNo { get; set; } = null!;
    [Column] public string Ledger { get; set; } = null!;
}
```

```csharp
using var db = new DataConnection(new DataOptions().UseSQLite(cs));
db.CreateTable<Header>();
db.CreateTable<Line>();
db.Insert(new Header { No = "H1" });
db.Insert(new Line { Id = 1, HeaderNo = "H1", Ledger = "L1" });
db.Insert(new Line { Id = 2, HeaderNo = "H1", Ledger = "L2" });

// The captured set is optional: null means "no ledger filter".
IQueryable<Header> Headers(HashSet<string>? ledgers) =>
    db.GetTable<Header>()
        .LoadWith(h => h.Lines, qb => qb.Where(l => ledgers == null || ledgers.Contains(l.Ledger)))
        .Where(h => h.Lines!.Any());

Headers(null).ToList();      // expect lines L1, L2
Headers(["L1"]).ToList();    // expect line  L1
```

Run the two calls in either order in the same process — the second one is wrong.

## Expected

`Headers(null)` → `L1, L2` and `Headers(["L1"])` → `L1`, each with its own eager-load SQL. This is 6.2.0+ with
`DisableQueryCache = true`:

```sql
-- Headers(null)
SELECT [m_1].[No], [d].[Id], [d].[HeaderNo], [d].[Ledger]
FROM [Headers] [m_1] INNER JOIN [Lines] [d] ON [m_1].[No] = [d].[HeaderNo]
WHERE EXISTS( SELECT * FROM [Lines] [l] WHERE [m_1].[No] = [l].[HeaderNo] )

-- Headers(["L1"])
... WHERE EXISTS( SELECT * FROM [Lines] [l] WHERE [m_1].[No] = [l].[HeaderNo] AND [l].[Ledger] IN ('L1') )
      AND [d].[Ledger] IN ('L1')
```

## Actual (6.2.0 – 6.4.0)

**Unfiltered first** — `Headers(["L1"])` returns `L1, L2`; it re-executes the first call's SQL and `IN ('L1')`
never appears.

**Filtered first** — `Headers(null)` returns nothing. The `ledgers == null` branch was folded to a constant when
the plan was built with a non-null set, so reuse with `null` gives an unsatisfiable predicate:

```sql
SELECT [m_1].[No], [d].[Id], [d].[HeaderNo], [d].[Ledger]
FROM [Headers] [m_1] INNER JOIN [Lines] [d] ON [m_1].[No] = [d].[HeaderNo]
WHERE 1 = 0
```

## Possibly relevant: 6.1.0 solved this differently

6.1.0 does not specialise the SQL per nullness — it emits a marker column and keeps one shape covering both cases:

```sql
-- 6.1.0, Headers(["L1"])
FROM ( SELECT [h].[No] as [No_1], 0 as [cond] FROM [Headers] [h] WHERE EXISTS(... AND [l].[Ledger] IN ('L1')) ) [m_1]
    INNER JOIN [Lines] [d] ON [m_1].[No_1] = [d].[HeaderNo] AND ([m_1].[cond] OR [d].[Ledger] IN ('L1'))
```

From 6.2.0 the eager-load SQL is specialised per nullness instead — a fine strategy, but the cache key appears not
to have been updated to match, so the two specialisations share one entry.

## Notes

- Order-dependent within a process, so a single call or an isolated test is always correct — easy to miss, and it
  will mis-report regression windows for any harness that runs several similar queries in one run.
- `DisableQueryCache = true` is a workaround, but recompiles every query.

### Reproduction Steps

Full sample:

```csharp
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;

namespace Repro;

[Table("Headers")]
public class Header
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string No { get; set; } = null!;

    [Association(ThisKey = nameof(No), OtherKey = nameof(Line.HeaderNo))]
    public List<Line>? Lines { get; set; }
}

[Table("Lines")]
public class Line
{
    [Column] public string HeaderNo { get; set; } = null!;
    [Column] public string Ledger { get; set; } = null!;
}

public static class Program
{
    // arg: "filtered-first" to run the calls in the other order
    public static void Main(string[] args)
    {
        var filteredFirst = args.Length > 0 && args[0] == "filtered-first";

        var cs = "Data Source=min2;Mode=Memory;Cache=Shared";
        using var keep = new Microsoft.Data.Sqlite.SqliteConnection(cs);
        keep.Open();

        using var db = new DataConnection(new DataOptions().UseSQLite(cs));
        db.CreateTable<Header>();
        db.CreateTable<Line>();
        db.Insert(new Header { No = "H1" });
        db.Insert(new Line { Id = 1, HeaderNo = "H1", Ledger = "L1" });
        db.Insert(new Line { Id = 2, HeaderNo = "H1", Ledger = "L2" });

        // The captured set is optional: null means "no ledger filter".
        IQueryable<Header> Headers(HashSet<string>? ledgers) =>
            db.GetTable<Header>()
                .LoadWith(h => h.Lines, qb => qb.Where(l => ledgers == null || ledgers.Contains(l.Ledger)))
                .Where(h => h.Lines!.Any());

        void Show(string label, HashSet<string>? f, string expected)
        {
            var got = Headers(f).ToList().SelectMany(h => h.Lines ?? []).Select(l => l.Ledger).OrderBy(x => x);
            Console.WriteLine($"  {label,-20} expect [{expected}] got [{string.Join(",", got)}]");
        }

        Console.WriteLine($"linq2db {typeof(DataConnection).Assembly.GetName().Version}, {args.FirstOrDefault() ?? "unfiltered-first"}");
        if (filteredFirst)
        {
            Show("Headers([\"L1\"])", ["L1"], "L1");
            Show("Headers(null)", null, "L1,L2");
        }
        else
        {
            Show("Headers(null)", null, "L1,L2");
            Show("Headers([\"L1\"])", ["L1"], "L1");
        }
    }
}
```

### Exception details

_No response_

### Regression?

It worked in 6.1.0, which we are now stuck at, and has been broken on all the newer versions

### Linq To DB version

6.4.0

### .NET version

10.0.0

### Database

SQL Server 2019

### Database provider

Microsoft.Data.Sqlite 9.0.5

### Operating system

macOS and Ubuntu CI machines

### Configuration

_No response_

sdanyliv:
Fixes #5793

## Symptom

A filter on an association that closes over an optional value — `values == null || values.Contains(...)` — was not keyed in the query cache. The first call to compile the query won, and every later call with different closure state silently re-executed that plan. Order-dependent, and wrong in both directions:

- unfiltered first → the filtered call returns unfiltered rows, the filter is missing entirely;
- filtered first → the unfiltered call gets `WHERE 1 = 0` and returns nothing.

Both the eager-load command and the main query's `EXISTS` are affected.

The defect was not confined to the overload the issue reports. All three ways to filter an association were broken:

| Form | Path |
|---|---|
| `LoadWith(sel, q => q.Where(...))` | `LoadWithMember.FilterFunc` |
| `ThenLoad(sel, q => q.Where(...))` | `LoadWithMember.FilterFunc` |
| `LoadWith(m => m.Items.Where(...))` | `LoadWithMember.FilterExpression` |

## Cause

Bisected to `0778057c10` ("Rewrite Expression Optimizer", #5262), first released in 6.2.0 — matching the reported window. It changed the tail of `AssociationHelper.CreateAssociationQueryLambda`:

```diff
- definedQueryMethod = (LambdaExpression)definedQueryMethod.OptimizeExpression(builder.MappingSchema)!;
+ definedQueryMethod = (LambdaExpression)definedQueryMethod.OptimizeExpression(builder.CanBeEvaluatedOnClient)!;
```

The old `IsEvaluable` carried an explicit closure guard:

```csharp
// don't return true for closure classes
ExpressionType.Constant => expression is ConstantExpression c
    && (c.Value == null || c.Value is string || c.Value.GetType().IsValueType),
```

`CanBeEvaluatedOnClient` has no such guard, so `values == null` began folding at build time — baking the first caller's state into the compiled plan. This call site runs after the `loadWith` block has inlined the user's filter, which is why it is the filter that gets baked.

Nothing downstream can catch it. The closure constant is replaced by a `ConstantPlaceholderExpression` in the cache key, and a decision that was folded away leaves no surviving `SqlValue`, so the by-value guard in `ExpressionCacheManager` — gated on `sqlValues.Contains(value)` — never arms. No dynamic-expression accessor is registered either: that gate opens only for associations carrying an expression predicate or entities carrying query filters, neither of which describes a plain key association with a `LoadWith` filter.

## Fix

1. Drop the `OptimizeExpression` call. Shape decisions stay in the expression and are resolved per execution at render time, the way a plain top-level `Where` already behaves — one cached plan serves both closure states.
2. Its structural half moves to `ExpressionBuildVisitor`, which translates per build and has no say in the cache key: collapsing `Convert(Convert(x, T), T?)`, where the outer conversion only adds the nullable wrapper over the type the inner one already produced. Comparison operands arrive in that shape from two directions — the C# compiler emits it for a lifted implicit conversion, and `ExpressionBuilder.Equal` builds it when it reconciles the unwrapped types and the nullable wrapper in two independent steps.

The relocated rule is deliberately narrower than the original in both reach and shape:

- it runs on the two operands of a comparison, not on every `Convert` in the tree. Applied tree-wide it also caught the compiler's lifted conversions, which added a widening `CAST` to INSERT projections on every provider;
- it fires only when the outer conversion does nothing but add or drop the nullable wrapper, and never on a conversion carrying a `Method`. A chain through a different intermediate type has no direct conversion to collapse to, and a user-defined operator must not have its call dropped. The original guarded neither.

It also has to run *after* `RestoreCompare`, which pattern-matches on the nested shape itself for its lifted `char` cases and its own double-conversion branch.

## Verification

- Full `Tests/Linq` on SQLite.MS: 10687 tests, 0 failed.
- Release builds clean on `net10.0` and `netstandard2.0`.
- Association and eager-loading suites green on DB2, SQL Server 2022, Oracle 23 and Informix.DB2 — the type-strict engines SQLite cannot stand in for.
- Each new test confirmed red before the fix and green after.

## Baselines

Generated SQL is unchanged. The only baseline additions are the new tests' own files.


MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A `LoadWith` filter that closes over an optional value returned **wrong data** once both closure states had run in the same process:

  ```csharp
  db.GetTable<MainItem>()
    .LoadWith(m => m.SubItems, q => q.Where(s => values == null || values.Contains(s.Value)))
  ```

  Whichever call ran first won: a later filtered call came back with unfiltered rows, or a later unfiltered call came back empty. No exception, and a single call in isolation was always correct. Applies to all three ways of filtering an association — `LoadWith(sel, q => …)`, `ThenLoad(sel, q => …)` and `LoadWith(m => m.Items.Where(…))` — and to the parent query's `EXISTS` as well as the eager-load query.


A `LoadWith` filter closing over an optional value no longer returns the previous call's rows.


---
<sub>Generated from commit `1c1eb22`.</sub>

PR #5802, opened 2026-08-19
Fix JoinsOptimizer crash on left join to a grouping with a computed key

### LINKED ISSUE #5799: JoinsOptimizer throws "Cannot get field for SqlFunction" when removing a left join onto a grouping with a computed key
## Summary

`Count()` over a query that LEFT JOINs a `GroupBy` whose **key is a computed expression** (not a plain column)
throws from inside the SQL optimizer. Materialising the same query with `ToList()` is fine — the crash only happens
once left-join removal engages, which `Count()` triggers because the joined columns become unused.

**Regression:** introduced in **6.2.0**. Correct in 6.0.0 and 6.1.0; broken in 6.2.0, 6.2.1, 6.3.0 and 6.4.0.

## Repro

```csharp
[Table("Items")]
public class Item
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string No { get; set; } = null!;
}

[Table("Stock")]
public class Stock
{
    [Column] public string Code { get; set; } = null!;
    [Column] public int Quantity { get; set; }
}
```

```csharp
using var db = new DataConnection(new DataOptions().UseSQLite(cs));
db.CreateTable<Item>();
db.CreateTable<Stock>();
db.Insert(new Item { No = "ITM" });
db.Insert(new Stock { Id = 1, Code = "ITM-A", Quantity = 5 });

// LEFT JOIN a grouping whose key is a computed expression (a SqlFunction), not a plain column.
var query =
    from item in db.GetTable<Item>()
    from g in db.GetTable<Stock>().GroupBy(s => s.Code.Substring(0, 3)).LeftJoin(g => g.Key == item.No)
    select new { item.No, Qty = g.Sum(s => s.Quantity) };

query.ToList();   // ok
query.Count();    // throws
```

Grouping by a plain column (`GroupBy(s => s.Code)`) does **not** trigger it — the key has to be an expression.

## Expected

`query.Count()` returns `1`, as it does on 6.1.0.

## Actual (6.2.0 – 6.4.0)

```
System.InvalidOperationException: Cannot get field for LinqToDB.Internal.SqlQuery.SqlFunction
   at JoinsOptimizer.GetTableKeysInternal            (JoinsOptimizer.cs:381)
   at JoinsOptimizer.GetTableKeys                    (JoinsOptimizer.cs:359)
   at JoinsOptimizer.IsLeftJoinCardinalityPreserved  (JoinsOptimizer.cs:260)
   at JoinsOptimizer.CanRemoveLeftJoin               (JoinsOptimizer.cs:236)
   at JoinsOptimizer.RemoveUnusedLeftJoins           (JoinsOptimizer.cs:30)
   at BasicSqlOptimizer.Finalize                     (BasicSqlOptimizer.cs:68)
   at ExpressionBuilder.BuildQuery
```

No SQL is produced — it fails during `Finalize`, before generation.

`GetTableKeys` appears to assume a table source's key elements are `SqlField`s; for a grouping keyed on an
expression the element is a `SqlFunction`, and it throws rather than treating the join as non-removable.

## Impact

This is an unhandled exception on an ordinary paging path: a repository that returns a page does
`query.Skip/Take` for the data plus `query.Count()` for the total, and only the `Count()` half crashes — so the
query works everywhere until someone asks for a total.

### Reproduction Steps

Repro:
```csharp
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;

namespace Repro;

[Table("Items")]
public class Item
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string No { get; set; } = null!;
}

[Table("Stock")]
public class Stock
{
    [Column] public string Code { get; set; } = null!;
    [Column] public int Quantity { get; set; }
}

public static class Program
{
    public static void Main()
    {
        var cs = "Data Source=bug3;Mode=Memory;Cache=Shared";
        using var keepAlive = new Microsoft.Data.Sqlite.SqliteConnection(cs);
        keepAlive.Open();

        using var db = new DataConnection(new DataOptions().UseSQLite(cs));
        db.CreateTable<Item>();
        db.CreateTable<Stock>();
        db.Insert(new Item { No = "ITM" });
        db.Insert(new Stock { Id = 1, Code = "ITM-A", Quantity = 5 });

        // LEFT JOIN a grouping whose key is a computed expression (a SqlFunction), not a plain column.
        var query =
            from item in db.GetTable<Item>()
            from g in db.GetTable<Stock>().GroupBy(s => s.Code.Substring(0, 3)).LeftJoin(g => g.Key == item.No)
            select new { item.No, Qty = g.Sum(s => s.Quantity) };

        Console.WriteLine($"linq2db {typeof(DataConnection).Assembly.GetName().Version}");
        Try("ToList()", () => query.ToList().Count);
        Try("Count() ", () => query.Count());
    }

    private static void Try(string what, Func<int> run)
    {
        try { Console.WriteLine($"  {what} -> ok ({run()})"); }
        catch (Exception e) { Console.WriteLine($"  {what} -> THROWS: {e.Message}"); }
    }
}
```

### Exception details

```
  Full stack from the minimal repro on 6.4.0 — every frame is linq2db except the last two (the repro's own Count() call):

  System.InvalidOperationException: Cannot get field for LinqToDB.Internal.SqlQuery.SqlFunction
     at JoinsOptimizer.<GetTableKeys>g__GetTableKeysInternal|11_0(SqlTableSource)   JoinsOptimizer.cs:381
     at JoinsOptimizer.GetTableKeys(SqlTableSource)                                 JoinsOptimizer.cs:359
     at JoinsOptimizer.IsLeftJoinCardinalityPreserved(SqlJoinedTable, EvaluationContext)  JoinsOptimizer.cs:260
     at JoinsOptimizer.CanRemoveLeftJoin(SqlStatement, SqlJoinedTable, EvaluationContext) JoinsOptimizer.cs:236
     at JoinsOptimizer.<>c.<RemoveUnusedLeftJoins>b__3_0(ValueTuple`3, IQueryElement)     JoinsOptimizer.cs:37
     at SqlQueryActionVisitor`1.Visit(IQueryElement)                                SqlQueryActionVisitor{TContext}.cs:50
     at QueryElementVisitor.<VisitElements>g__VisitElementsReadOnly|106_0[T](List`1) QueryElementVisitor.cs:3789
     at QueryElementVisitor.VisitElements[T](List`1, VisitMode)                      QueryElementVisitor.cs:3769
     at QueryElementVisitor.VisitSqlFromClause(SqlFromClause)                        QueryElementVisitor.cs:1884
     at SqlFromClause.Accept(QueryElementVisitor)                                    SqlFromClause.cs:203
     at QueryElementVisitor.Visit(IQueryElement)                                     QueryElementVisitor.cs:80
     at SqlQueryActionVisitor`1.Visit(IQueryElement)                                 SqlQueryActionVisitor{TContext}.cs:50
     at QueryElementVisitor.<VisitSqlQuery>g__VisitSelectQuery|64_1(SelectQuery, VisitMode)  QueryElementVisitor.cs:2447
     at QueryElementVisitor.VisitSqlQuery(SelectQuery)                               QueryElementVisitor.cs:2305
     at SelectQuery.Accept(QueryElementVisitor)                                      SelectQuery.cs:327
     at QueryElementVisitor.Visit(IQueryElement)                                     QueryElementVisitor.cs:80
     at SqlQueryActionVisitor`1.Visit(IQueryElement)                                 SqlQueryActionVisitor{TContext}.cs:50
     at QueryElementVisitor.VisitSqlSelectStatement(SqlSelectStatement)              QueryElementVisitor.cs:1334
     at SqlSelectStatement.Accept(QueryElementVisitor)                               SqlSelectStatement.cs:35
     at QueryElementVisitor.Visit(IQueryElement)                                     QueryElementVisitor.cs:80
     at SqlQueryActionVisitor`1.Visit(IQueryElement)                                 SqlQueryActionVisitor{TContext}.cs:50
     at SqlQueryActionVisitor`1.Visit(TContext, IQueryElement, Boolean, Action`2)    SqlQueryActionVisitor{TContext}.cs:26
     at QueryVisitorExtensions.Visit[TContext](IQueryElement, TContext, Action`2)    QueryVisitorExtensions.cs:33
     at JoinsOptimizer.RemoveUnusedLeftJoins(SqlStatement, EvaluationContext)        JoinsOptimizer.cs:30
     at JoinsOptimizer.Optimize(SqlStatement, EvaluationContext)                     JoinsOptimizer.cs:18
     at BasicSqlOptimizer.Finalize(MappingSchema, SqlStatement, DataOptions)         BasicSqlOptimizer.cs:68
     at ExpressionBuilder.BuildQuery[T](...)                                         ExpressionBuilder.cs:278
     at ExpressionBuilder.Build[T](IQueryExpressions&)                               ExpressionBuilder.cs:225
     at Query`1.CreateQuery(...)                                                     Query{T}.cs:220
     at Query`1.GetQuery(...)                                                        Query{T}.cs:197
     at ExpressionQuery`1.GetQuery(...)                                              ExpressionQuery.cs:91
     at ExpressionQuery`1.System.Linq.IQueryProvider.Execute[TResult](Expression)    ExpressionQuery.cs:312
     at Repro.Program...  // query.Count()
```

### Regression?

It worked in 6.1.0, which we are now stuck at

### Linq To DB version

6.4.0

### .NET version

10.0.0

### Database

SQL Server 2019

### Database provider

Microsoft.Data.Sqlite 9.0.5

### Operating system

macOS and Ubuntu CI machines

### Configuration

_No response_

sdanyliv:
Fixes #5799.

`JoinsOptimizer.GetTableKeysInternal` assumed every element of a unique keyset reduces to a `SqlField`/`SqlColumn`, and threw otherwise:

    fields[j] = GetUnderlyingFieldOrColumn(keyset[j]) ?? throw new InvalidOperationException($"Cannot get field for {keyset[j]}");

That assumption is wrong. `QueryHelper.CollectUniqueKeys` contributes `Select.GroupBy.Items` — and, for `DISTINCT`, the projected column expressions — verbatim, so a grouping keyed on an expression (`GroupBy(s => s.Code.Substring(0, 3))`) puts a `SqlFunction` in the keyset and the reducer correctly returns `null`.

`Count()` is what exposes it: dropping the aggregate from the projection leaves the joined subquery's columns unreferenced, so `CanRemoveLeftJoin` reaches the cardinality check — the only caller that walks the keyset. `ToList()` never gets there. Every other `GetUnderlyingFieldOrColumn` call site already handles `null`; this was the sole one treating it as impossible.

Fix: keep a non-reducible keyset element as is instead of throwing.

This is not only crash avoidance. `IsLeftJoinCardinalityPreserved` already compares against an unwrapped set built as `column -> column.Expression`, so carrying the raw expression through lets a computed grouping key match the column that projects it. The join is then correctly recognised as cardinality-preserving and removed — the `Count()` query renders with no `LEFT JOIN` instead of crashing.

The `throw` itself predates the regression window (it only moved in the namespace reorg 639d598a37); what 6.2.0 changed is that this AST shape now reaches it.

### Tests

`JoinOptimizeTests.LeftJoinToGroupingWithComputedKey` pins both halves: `ToList()` keeps the join and returns the expected row, `Count()` returns 1 and drops the join.

Verified locally: full `Tests/Linq` on SQLite.MS / net10.0 — 10680 total, 0 failed, 277 pre-existing skips. Release build with analyzers — 0 warnings. Baselines were not regenerated locally.


MaceWindu (approved):
> [!IMPORTANT]

## Review notes

- [ ] **The description understates the fix, and its "only caller" reasoning doesn't hold.** Two corrections, both worth making because this body becomes the merge commit message.

  *"the only caller that walks the keyset"* — `GetTableKeys` has a second caller, `JoinsOptimizer.RemoveDuplicates.cs:96` (`TryMergeSources`), feeding `TryMergeSources2`'s `uniqueKeys`. For the GROUP BY / DISTINCT route the conclusion survives, but not for the stated reason: `TryMergeSources` first requires both sources be exactly `typeof(SqlTable)` (`QueryHelper.cs:913-918`), which a `SelectQuery`-backed grouping never is. That gate — not caller count — is what keeps those keysets away from the duplicate-join pass.

  Second, the fix closes **three** distinct crash routes, not one. All three throw the identical `InvalidOperationException: Cannot get field for LinqToDB.Internal.SqlQuery.SqlFunction` on the pre-fix optimizer and pass at this head, verified by swapping only `JoinsOptimizer.cs` back to `master` with everything else at HEAD:

  - **GROUP BY items** — the reported case, covered by the new test.
  - **DISTINCT projections** — `QueryHelper.CollectUniqueKeys` contributes a non-`DISTINCT ON` query's projected column expressions verbatim (`QueryHelper.cs:1577-1578`), so `Select(s => s.Code.Substring(0, 3)).Distinct()` left-joined and then `Count()`ed reaches the same line.
  - **`HasUniqueKey`** — `LinqExtensions.HasUniqueKey` (shipped public API) writes the raw selector straight into `SelectQuery.UniqueKeys` (`HasUniqueKeyBuilder.cs:30`). This one needs no subquery at all: the emitted join is a plain `LEFT JOIN [StockEntry] [s] ON [s].[Code] = [item_1].[No]`, `ToList()` succeeds and `Count()` throws. It is also the only route that genuinely reaches the `RemoveDuplicates.cs:96` consumer — harmlessly, since every `RightField` there came out of `GetUnderlyingFieldOrColumn` and is always a field or column, so a `SqlFunction` can never match and the keyset is simply skipped.

- [x] CI: `build` and the full `test-all` matrix both pass, and the run postdates the branch's only commit, so it covers the reviewed head.

## Out-of-scope observations

- **The newly-enabled removal matches by instance identity, not structurally.** `IsLeftJoinCardinalityPreserved` looks the kept expression up in `HashSet<ISqlExpression>`s built with the default comparer. `Equals` resolves structurally via `SqlExpressionBase`, but `HashSet` gates on `GetHashCode()` first — and `QueryElement` declares only an abstract `GetElementHashCode()`, never overriding `object.GetHashCode()`. `SqlFunction` and `SqlColumn` add no override; `SqlField` returns `RuntimeHelpers.GetHashCode(this)`. So for exactly the node types this fix newly admits, the lookup degenerates to reference identity, and the removal works because `Select.GroupBy.Items[0]` and the projecting column's `Expression` are the same instance. A later pass that clones the grouping item would silently stop removing the join — a safe degradation, and one the new test's `ShouldNotContain("LEFT JOIN")` would catch. `ISqlExpressionEqualityComparer.Instance` already exists and keys on `GetElementHashCode()`. Pre-existing and repo-wide, not caused by this PR.

## Baselines

Delta: [linq2db.baselines#2071](https://github.com/linq2db/linq2db.baselines/pull/2071) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5802)

**4 added, 0 modified, 0 deleted**, over `origin/master...origin/baselines/pr_5802`. All four are the new test's own baselines, across exactly the `AllSQLite` set.

- [`LeftJoinToGroupingWithComputedKey`](https://github.com/linq2db/linq2db.baselines/blob/baselines/pr_5802/SQLite.MS/Tests/Linq/JoinOptimizeTests/Tests.Linq.JoinOptimizeTests.LeftJoinToGroupingWithComputedKey%28SQLite.MS%29.sql) — added (4 providers: SQLite.Classic, SQLite.Classic.MPM, SQLite.Classic.MPU, SQLite.MS), byte-identical modulo the header comment. The captured pair confirms both halves: `ToList()` keeps the `LEFT JOIN` onto the `GROUP BY [g_2].[Key_1]` subquery, `Count()` collapses to `SELECT COUNT(*) FROM [InventoryItem] [t1]`.

**No pre-existing baseline moved anywhere in the corpus** — the branch carries only two commits, both from the Windows SQLite legs, so every other leg of the green matrix produced nothing to commit. That is a hard bound on blast radius and the expected fingerprint: the `??` fallback fires only where the old code threw, so no query that previously succeeded can render differently.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Calling `Count()` on a query that left-joins a grouping with a computed key crashed with `InvalidOperationException: Cannot get field for …`:

  ```csharp
  from item in t1
  from g in t2.GroupBy(s => s.Code.Substring(0, 3)).LeftJoin(g => g.Key == item.No)
  select new { item.No, Qty = g.Sum(s => s.Quantity) }
  ```

  `ToList()` on the same query worked, so a paging query failed only on the total. A grouping by a plain column was unaffected — the key had to be an expression. The same crash was reachable through `Distinct()` over a computed projection and through `HasUniqueKey`.


`Count()` over a left-joined grouping with a computed key no longer crashes.


---
<sub>Generated from commit `2496737`.</sub>

PR #5027, opened 2025-07-20
[Cleanup] Internal APIs cleanup

MaceWindu:
- change namespaces for some internal apis
- seal some classes
- remove public visibility from some internal apis

PR #5804, opened 2026-08-20
Make IInterceptable public

### LINKED ISSUE #5803: Public IInterceptable?
### Description

As discussed in #4365, I must use IInterceptable<> to work around interceptors not working with custom IDataContext implementations. But this interface is no longer public in 6.x.

I'll create a PR to restore public on this interface.

### Reproduction Steps

See #4365 and existing tests (TestDataCustomConnection)

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.4.0

### .NET version

8

### Database

_No response_

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

cal-tlabwest:
Resolves #5803. Required to work around interceptors not working with custom IDataContext.

### DIFF
--- a/Source/LinqToDB/Internal/Interceptors/IInterceptable.cs
+++ b/Source/LinqToDB/Internal/Interceptors/IInterceptable.cs
@@ -2,11 +2,11 @@
 
 namespace LinqToDB.Internal.Interceptors
 {
-	interface IInterceptable
+	public interface IInterceptable
 	{
 	}
 
-	interface IInterceptable<T> : IInterceptable
+	public interface IInterceptable<T> : IInterceptable
 		where T : IInterceptor
 	{
 		T? Interceptor { get; set; }


MaceWindu (approved):

MaceWindu:
Superseded by #5813, which carries your commit as-is (authorship preserved) plus tests and the engine fix.

Reason for moving it rather than extending this branch: the fork is owned by the `TLabWest` org, and GitHub does not offer "allow edits by maintainers" for org-owned forks — `maintainerCanModify` is `false` and cannot be turned on — so the master sync and the follow-up commits could not be pushed here.

Confirmed your reasoning about `IInterceptable<>` being required, and it is broader than the compiled-query case: interceptor dispatch is driven entirely by `dataContext is IInterceptable<T>` type tests, while `IDataContext` exposes only `AddInterceptor` / `RemoveInterceptor`. So with the interface `internal`, a custom `IDataContext` can accept an interceptor and nothing will ever call it. The visibility was dropped in #5027, a bulk "remove public visibility from some internal apis" sweep — its criterion was "not used outside the assembly", which by construction cannot see external implementers.

Two things surfaced while testing it that are worth knowing for your own context:

1. A custom `IDataContext` must also implement `IInfrastructure<IServiceProvider>` (public, `LinqToDB.Internal.Infrastructure`) or every query build throws `InvalidCastException` — it is hard-cast in `ExposeExpressionVisitor` and `ExpressionBuilder` but not declared on the interface. Filed as #5812 for 7.0.0.
2. The `.ToList()` variant was failing for an unrelated reason that also affects plain `DataConnection`, fixed in #5813.

Thanks for the report and the patch.

PR #5807, opened 2026-08-21
CI: do not fail the build when the Azure Artifacts feed is over quota

MaceWindu:
## Problem

`default (Nugets Generation)` fails on every `master` build once the Azure Artifacts feed fills up. The feed answers a push with `HTTP 402 Payment Required` ("Artifact cannot be uploaded because max quantity has been exceeded or the payment instrument is invalid"), and `NuGetCommand@2` treats any non-zero push exit as a build failure — so the build goes red with every other step green.

Most recent occurrence: [build 22947](https://dev.azure.com/linq2db/0dcc414b-ea54-451e-a54f-d63f05367c4b/_build/results?buildId=22947) on af5eda3d8, where the only failing step was `Publish to Azure Artifacts feed`.

## Change

The Azure Artifacts push moves from `NuGetCommand@2` to `NuGetAuthenticate@1` + `Build/Azure/scripts/publish-azure-artifacts.ps1`.

- Feed over quota (`402`) → warning + `SucceededWithIssues`; the build is not failed.
- Any other push failure → the step fails, as before.

The Azure Artifacts feed is a convenience mirror. Releases publish to nuget.org through a separate step, which this PR does not touch.

## Why a script rather than `continueOnError: true`

`continueOnError` on the task would tolerate *every* push failure, including genuine publish breakage. The script distinguishes the quota case from the rest.

The numeric match is deliberately anchored to a status-code context rather than matching a bare `402`: the same message carries a DevOps activity GUID that can itself contain `402`, and matching that would silently swallow an unrelated push failure.

## Verification

- Pipeline YAML parses; step order unchanged apart from the inserted auth task.
- Script parses clean.
- Matcher checked against the verbatim 402 body from build 22947, plus negatives: `401 Unauthorized`, `500 Internal Server Error`, and a GUID containing `402`.

## Reviewer note

Two things this PR's own CI cannot validate, because the publish step is `condition: eq(variables['Build.SourceBranchName'], variables['master_branch'])` and is skipped on any PR branch:

1. The feed URL `https://pkgs.dev.azure.com/linq2db/0dcc414b-.../_packaging/967a4107-.../nuget/v3/index.json` is derived from the GUIDs in the previous `publishVstsFeed` value.
2. Authentication now comes from `NuGetAuthenticate@1` + `dotnet nuget push` instead of the task's built-in internal-feed handling.

Both only execute after merge to `master`, so they are worth an explicit look.

### DIFF
--- a/Build/Azure/pipelines/templates/nuget-job.yml
+++ b/Build/Azure/pipelines/templates/nuget-job.yml
@@ -28,13 +28,18 @@ jobs:
       filePath: '$(Build.SourcesDirectory)/Build/Azure/scripts/verify-analyzer-delivery.ps1'
       arguments: '-PackagesDir "$(Build.SourcesDirectory)/.build/nugets"'
 
-  - task: NuGetCommand@2
-    inputs:
-      command: 'push'
-      packagesToPush: '$(Build.SourcesDirectory)/.build/nugets/*.nupkg'
-      nuGetFeedType: 'internal'
-      publishVstsFeed: '0dcc414b-ea54-451e-a54f-d63f05367c4b/967a4107-9788-41a4-9f6d-a2318aab1410'
+  # Pushed through a script rather than NuGetCommand@2 so that the feed being over its storage
+  # quota (HTTP 402) warns instead of failing the build, while any other push failure still fails.
+  - task: NuGetAuthenticate@1
+    displayName: Authenticate to Azure Artifacts feed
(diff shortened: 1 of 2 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `e4ce3ca`.</sub>

Issue #4404, opened 2024-01-30
ToListAsync() thread problems with LoadWith() queries


cal-tlabwest:
`ToListAsync()` has a thread problem when called with a non `ExpressionQuery<>` source such as `LoadWithQueryable<>`.
`db.MyTable.Where(...).LoadWith(...).ToListAsync()` can crash as the DB connection is used in a thread pool task.
`db.MyTable.LoadWith(...).Where(...).ToListAsync()` works however as the source is a `ExpressionQuery<>` in that case.

```
You can get various thread problem related crashes, such as:

Cannot access a disposed object.
Object name: 'SQLiteCommand'.
at System.Data.SQLite.SQLiteCommand.set_Transaction(SQLiteTransaction value)
   at LinqToDB.Data.DataConnection.<>c.<BeginTransaction>b__181_1(DataConnection dataConnection, IsolationLevel isolationLevel)
   at LinqToDB.Data.DataConnection.TraceAction[TContext,TResult](DataConnection dataConnection, TraceOperation traceOperation, Func`2 commandText, TContext context, Func`3 action)
   at LinqToDB.Data.DataConnection.BeginTransaction(IsolationLevel isolationLevel)
   at LinqToDB.Linq.ExpressionQuery`1.System.Collections.Generic.IEnumerable<T>.GetEnumerator()
   at LinqToDB.LinqExtensions.LoadWithQueryable`2.GetEnumerator()
   at System.Linq.Enumerable.TakeWhileIterator[TSource](IEnumerable`1 source, Func`2 predicate)+MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at LinqToDB.AsyncExtensions.<>c__DisplayClass8_0`1.<ToListAsync>b__0()
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
[...]

PR #5809, opened 2026-08-22
Fix async materialization for LoadWith and other query wrappers

### LINKED ISSUE #5805: LoadWith() forces ToListAsync() to use thread pool
### Description

LoadWith() returns a LoadWithQueryable<,> that ToListAsync() etc does not recognize as an ExpressionQuery<>:

```
public static async Task<List<TSource>> ToListAsync<TSource>(
	this IQueryable<TSource> source,
	CancellationToken token = default)
{
	if (source is ExpressionQuery<TSource> query)
	{
		var list = new List<TSource>();
		await query.GetForEachAsync(list.Add, token).ConfigureAwait(false);
		return list;
	}

	if (LinqExtensions.ExtensionsAdapter != null)
		return await LinqExtensions.ExtensionsAdapter.ToListAsync(source, token).ConfigureAwait(false);

	return await Task.Run(() => source.AsEnumerable().ToListToken(token), token).ConfigureAwait(false);
}
```

Instead it will fallback to using Task.Run(). It would be better and more efficient if this could be avoided. Doing query,LoadWith(...).Select(o => o).ToListAsync() avoids the problem. FirstAsync() etc works as expected though.

### Reproduction Steps

query,LoadWith(...).ToListAsync();

### Exception details

_No response_

### Regression?

_No response_

### Linq To DB version

6.4.0

### .NET version

8

### Database

_No response_

### Database provider

_No response_

### Operating system

_No response_

### Configuration

_No response_

sdanyliv:
Fixes #5805

## Problem

`LoadWith()` returns a wrapper over the linq2db query, and `AsyncExtensions` detected a linq2db query with a direct `source is ExpressionQuery<TSource>` check. A wrapper fails that check, so `ToListAsync()` fell through to `Task.Run(() => source.AsEnumerable()...)` — a synchronous read on a thread-pool thread. `FirstAsync()` and friends were unaffected because the generated overloads go through `source.Provider as IQueryProviderAsync`, which wrappers forward.

The same defect affects every query wrapper, not only `LoadWith`:

| Wrapper | Reachable via |
|---|---|
| `LoadWithQueryableBase<T>` | `LoadWith` / `ThenLoad` |
| `TempTable<T>` | `CreateTempTable` / `CreateLocalTable` |
| `DatabaseSpecificQueryable<T>` | `query.AsSqlServer()`, `AsOracle()`, ... |
| `DatabaseSpecificTable<T>` | `table.AsSQLite()`, `AsSqlServer()`, ... |

`PersistentTable<T>` also wraps a query, but it is a builder-internal association artifact that never reaches user materialization (its `DataContext` is `null!`), so it is left alone.

## Fix

Wrappers implement a new internal `IQueryableWrapper<T>` marker exposing the wrapped query. `LinqInternalExtensions.UnwrapQueryable` walks that chain, and both query lookups use it:

- `AsLinqToDBQuery` (new) — returns `ExpressionQuery<T>` or `null`, used by the 11 detection sites in `AsyncExtensions`: `ToListAsync`, `ToArrayAsync`, `ToDictionaryAsync` (x4), `ToLookupAsync` (x2), `ForEachAsync`, `ForEachUntilAsync`, `AsAsyncEnumerable`.
- `GetLinqToDBSource` — unchanged contract, one `.UnwrapQueryable()` inserted into the existing chain.

`AsLinqToDBQuery` deliberately does not apply `ProcessIQueryable`: the EF hook throws `LinqToDBForEFToolsException` for a plain non-EF `IQueryable`, which would turn today's graceful fallback into an exception.

### Second family, same root cause

`GetLinqToDBSource` rejected a `LoadWith` query outright, so `ElementAt` / `ElementAtOrDefault` (and their `Async` overloads), `Insert`, `Delete`, `Update`, `Merge`, `MultiInsert` and the `AnalyticFunctions` overloads threw `LinqToDB method 'X' called on non-LinqToDB IQueryable`. `ToSqlQuery` escaped only because it carried a hand-rolled `LoadWithQueryableBase` unwrap — removed here as redundant.

## Tests

SQLite-only, in the fixtures that own the behaviour:

- `AsyncTests.LoadWithQueryMaterializedAsynchronously`, `TempTableMaterializedAsynchronously`, `DatabaseSpecificTableMaterializedAsynchronously` — a `CommandInterceptor` counts `ExecuteReader` vs `ExecuteReaderAsync`, so the assertion is on the ADO.NET API actually used rather than on timing.
- `LoadWithTests.LoadWithQueryIsLinqToDbSource` — `ToSqlQuery` plus `ElementAtOrDefault`.

All verified red before the fix (`AsyncCalls == 0` for the async tests, `LinqToDBException` for `ElementAtOrDefault`).

Not covered: `DatabaseSpecificQueryable` (the `query.AsSqlServer()` shape) has no SQLite entry point, so only its `DatabaseSpecificTable` sibling is exercised by a test.

## Verification

- Full `Tests/Linq` suite on `SQLite.MS`: 10912 tests, 0 failed, 10634 passed, 278 skipped.
- Release build across `net462`, `netstandard2.0`, `net8.0`, `net9.0`, `net10.0` with analyzers: 0 warnings, 0 errors.

### DIFF
--- a/Source/LinqToDB/Async/AsyncExtensions.cs
+++ b/Source/LinqToDB/Async/AsyncExtensions.cs
@@ -168,6 +168,9 @@ namespace LinqToDB.Async
 			if (source is IAsyncEnumerable<TSource> asyncQuery)
 				return asyncQuery;
 
+			if (source.AsLinqToDBQuery() is { } query)
+				return query;
+
 			if (LinqExtensions.ExtensionsAdapter != null)
 				return LinqExtensions.ExtensionsAdapter.AsAsyncEnumerable(source);
 
@@ -212,7 +215,7 @@ namespace LinqToDB.Async
 			this IQueryable<TSource> source, Action<TSource> action,
 			CancellationToken token = default)
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 				return query.GetForEachAsync(action, token);
 
 			if (LinqExtensions.ExtensionsAdapter != null)
@@ -245,7 +248,7 @@ namespace LinqToDB.Async
 			this IQueryable<TSource> source, Func<TSource,bool> func,
 			CancellationToken token = default)
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 				return query.GetForEachUntilAsync(func, token);
 
 			return Task.Run(
@@ -278,7 +281,7 @@ namespace LinqToDB.Async
 			this IQueryable<TSource> source,
 			CancellationToken token = default)
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var list = new List<TSource>();
 				await query.GetForEachAsync(list.Add, token).ConfigureAwait(false);
@@ -306,7 +309,7 @@ namespace LinqToDB.Async
 			this IQueryable<TSource> source,
 			CancellationToken token = default)
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var list = new List<TSource>();
 				await query.GetForEachAsync(list.Add, token).ConfigureAwait(false);
@@ -338,7 +341,7 @@ namespace LinqToDB.Async
 			CancellationToken        token   = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var dic = new Dictionary<TKey,TSource>();
 				await query.GetForEachAsync(item => dic.Add(keySelector(item), item), token).ConfigureAwait(false);
@@ -368,7 +371,7 @@ namespace LinqToDB.Async
 			CancellationToken        token = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var dic = new Dictionary<TKey,TSource>(comparer);
 				await query.GetForEachAsync(item => dic.Add(keySelector(item), item), token).ConfigureAwait(false);
@@ -399,7 +402,7 @@ namespace LinqToDB.Async
 			CancellationToken        token = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var dic = new Dictionary<TKey,TElement>();
 				await query.GetForEachAsync(item => dic.Add(keySelector(item), elementSelector(item)), token).ConfigureAwait(false);
@@ -432,7 +435,7 @@ namespace LinqToDB.Async
 			CancellationToken        token = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var dic = new Dictionary<TKey,TElement>(comparer);
 				await query.GetForEachAsync(item => dic.Add(keySelector(item), elementSelector(item)), token).ConfigureAwait(false);
@@ -492,7 +495,7 @@ namespace LinqToDB.Async
 			CancellationToken        token = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var lookup = new Lookup<TKey,TSource>(comparer);
 				await query.GetForEachAsync(item => lookup.AddItem(keySelector(item), item), token).ConfigureAwait(false);
@@ -551,7 +554,7 @@ namespace LinqToDB.Async
 			CancellationToken        token = default)
 			where TKey : notnull
 		{
-			if (source is ExpressionQuery<TSource> query)
+			if (source.AsLinqToDBQuery() is { } query)
 			{
 				var lookup = new Lookup<TKey,TElement>(comparer);
 				await query.GetForEachAsync(item => lookup.AddItem(keySelector(item), elementSelector(item)), token).ConfigureAwait(false);
--- a/Source/LinqToDB/Internal/DataProvider/DatabaseSpecificQueryable.cs
+++ b/Source/LinqToDB/Internal/DataProvider/DatabaseSpecificQueryable.cs
@@ -11,8 +11,10 @@ using LinqToDB.Internal.Linq;
 
 namespace LinqToDB.Internal.DataProvider
 {
-	public abstract class DatabaseSpecificQueryable<TSource>(IExpressionQuery<TSource> query) : IExpressionQuery<TSource>
+	public abstract class DatabaseSpecificQueryable<TSource>(IExpressionQuery<TSource> query) : IExpressionQuery<TSource>, IQueryableWrapper<TSource>
 	{
+		IQueryable<TSource> IQueryableWrapper<TSource>.WrappedQuery => query;
+
 		public IEnumerator<TSource> GetEnumerator() 
 			=> query.GetEnumerator();
 
--- a/Source/LinqToDB/Internal/DataProvider/DatabaseSpecificTable.cs
+++ b/Source/LinqToDB/Internal/DataProvider/DatabaseSpecificTable.cs
@@ -11,7 +11,7 @@ using LinqToDB.Internal.Linq;
 
 namespace LinqToDB.Internal.DataProvider
 {
-	public abstract class DatabaseSpecificTable<TSource> : ITable<TSource>
+	public abstract class DatabaseSpecificTable<TSource> : ITable<TSource>, IQueryableWrapper<TSource>
 		where TSource : notnull
 	{
 		protected DatabaseSpecificTable(ITable<TSource> table)
@@ -20,6 +20,8 @@ namespace LinqToDB.Internal.DataProvider
 		}
 
 		readonly ITable<TSource> _table;
+
+		IQueryable<TSource> IQueryableWrapper<TSource>.WrappedQuery => _table;
 
 		public IEnumerator<TSource> GetEnumerator()
 		{
--- a/Source/LinqToDB/Internal/Linq/IQueryableWrapper.cs
+++ b/Source/LinqToDB/Internal/Linq/IQueryableWrapper.cs
@@ -1,0 +1,18 @@
+﻿using System.Linq;
+
+namespace LinqToDB.Internal.Linq
+{
+	/// <summary>
+	/// Implemented by <see cref="IQueryable{T}"/> implementations, which wrap another query to add extra
+	/// query-building functionality (e.g. queries, returned by <c>LoadWith</c> or database-specific hint methods).
+	/// Provides access to wrapped query, so query consumers could detect linq2db query behind the wrapper.
+	/// </summary>
+	/// <typeparam name="TSource">Query element type.</typeparam>
+	interface IQueryableWrapper<out TSource>
+	{
+		/// <summary>
+		/// Gets wrapped query.
+		/// </summary>
+		IQueryable<TSource> WrappedQuery { get; }
+	}
+}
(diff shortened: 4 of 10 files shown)


MaceWindu (commented):
> [!IMPORTANT]

## Review notes

- [x] **No correctness findings.** The mechanism was checked from three independent angles — unwrap semantics, the changed `GetLinqToDBSource` contract, and wrapper-inventory completeness — and cleared on all three. Details below.
- [x] **Two test gaps found and filled in `4af12d8fc`** (pushed to this branch). (1) `AsAsyncEnumerable` is the only detection site this PR *adds* a branch to rather than substituting an existing predicate, and it sits behind the pre-existing `if (source is IAsyncEnumerable<TSource>)` check. Of the four wrappers only `LoadWithQueryable<,>` implements `IAsyncEnumerable<T>`, so `TempTable<T>`, `DatabaseSpecificTable<TSource>` and `DatabaseSpecificQueryable<TSource>` depend entirely on the new branch; without it they fall through to `AsyncEnumerableAdapter<T>` and enumerate synchronously behind an async facade — no exception, no baseline change, no red test. `TempTableMaterializedAsynchronously` now also drains `AsAsyncEnumerable`. (2) `DatabaseSpecificQueryable<TSource>` had no test, and is the only marked wrapper whose async path was genuinely broken pre-PR, since `AsLinqToDBQuery` casts to the concrete `ExpressionQuery<T>` and the `*SpecificQueryable` types are not one. Covered on DuckDB, which has the `IQueryable`-taking `AsDuckDB` overload that SQLite lacks. Both were confirmed red before the fix and green after — the `AsAsyncEnumerable` case with the new branch removed *in isolation*, so the failure is attributable to that branch and not to `ToListAsync`.
- [x] **Unwrapping is semantically inert — checked at each construction site, not assumed.** Every wrapper is handed an inner query whose `Expression` already carries the wrapper's own call node: `LoadWith` wraps `Provider.CreateQuery(Expression.Call(LoadWith, …))` (`LinqExtensions.LoadWith.cs:196-203`), each `As<Provider>()` / `*Hints` method wraps a *fresh* `Table<T>` or `ExpressionQuery<T>` built around its call node, and `TempTable<T>._table` is always the `Table<T>` returned by `db.CreateTable<T>()`. The new baselines confirm it end-to-end — see the Baselines section.
- [x] **`GetLinqToDBSource` returning the inner query rather than the wrapper is safe.** Across the call sites in `LinqExtensions.{Update,Insert,Delete,Merge,Upsert,EntityInsert,EntityUpdate}.cs`, `MultiInsertExtensions.cs` and `AnalyticFunctions.cs`, the only casts applied to the result are `(IExpressionQuery)` and `(IQueryable<TSource>)` — both satisfied by the inner query. None casts to `ITable<T>`, `ITableMutable`, `IDisposable` or a provider-specific interface, and none reads `TableName`, `TableOptions` or `DataContext` off it. `MethodHelper.GetMethodInfo(...)` is passed the original `source` everywhere, so the generated method infos don't move.
- [x] **The wrapper inventory is complete.** Swept `Source/` independently for queryable types that delegate `Expression` / `Provider` / `GetEnumerator` to an inner query — that delegation is the wrapper signature. Exactly seven exist: the four marked here, `PersistentTable<T>`, and `LinqToDBForEFQueryProvider<T>` in the EF assembly. The 21 provider-specific `*SpecificTable` / `*SpecificQueryable` classes all derive from the two marked abstract bases, so marking the bases covers every provider; `LinqToDB.Tools`, `LinqToDB.Remote.*`, `LinqToDB.LINQPad`, `LinqToDB.FSharp`, `LinqToDB.Scaffold` and `LinqToDB.CLI` implement neither `IQueryable` nor `ITable`.
- [x] **`ProcessIQueryable().UnwrapQueryable()` is the right order.** The EF hook short-circuits on `queryable.Provider is IQueryProviderAsync` (`LinqToDBForEFTools.cs:56-77`), and every wrapper forwards `Provider` to its inner query — so for the only reachable shape the hook is a no-op and the composition reduces to a plain unwrap. A wrapper over an EF query is not constructible while the hook is installed, because `LoadWith` and `As<Provider>()` call `ProcessIQueryable()` first, which forces the `IQueryProviderAsync` arm.
- [x] **Wrappers do not nest over a linq2db source**, so `UnwrapQueryable`'s `while` is defensive rather than covering a reachable case: every hint step wraps a fresh `Table<T>` (e.g. `SQLiteSpecificExtensions.cs:19-26`), and the one nesting path — `LoadWith` / `ThenLoad`'s `Provider is not IQueryProviderAsync` arm — bottoms out in a query that is not linq2db anyway. Recorded so the loop is not mistaken for coverage of a live case.
- [x] **No public-API surface change, and no baseline work needed.** `IQueryableWrapper<T>` is internal and all four implementations are explicit, so no new public or protected member appears on `TempTable<T>`, `DatabaseSpecificQueryable<TSource>` or `DatabaseSpecificTable<TSource>`. `PublicAPI.*.txt` records types and members rather than interface lists, and ApiCompat does not see an internal interface — so `Source/**/CompatibilitySuppressions.xml` needs no regeneration.
- [x] **XML doc references all resolve** — `IQueryable{T}`, `ProcessIQueryable{T}`, `AsLinqToDBQuery{T}` and the cross-namespace `LinqExtensions.ExtensionsAdapter` — so there is no CS1574 exposure on the Release leg.
- [ ] **One inaccuracy in the PR description.** It states that `PersistentTable<T>` "never reaches user materialization". It does reach user code: `ExpressionBuilder.Associations.cs:277` constructs one whenever a materialized association member is declared as `ITable<TChild>`, and that instance is the property value handed back — confirmed at runtime (`parent.Children.GetType().Name` is `PersistentTable`1`, and the association loads correctly). Leaving it unmarked is still the right call, but for a different reason than stated: line 271 wraps the already-materialized sequence in `Queryable.AsQueryable` first, so the inner query is an in-memory `EnumerableQuery` and `AsLinqToDBQuery` returns `null` whether or not the marker is present.
- [ ] **Prior report worth linking.** This defect was filed as #4404 in January 2024 and closed as `wontfix` in favour of #4405. That thread diverged onto ADO.NET thread-safety, even though a commenter had already identified this exact mechanism — `ToListAsync` type-testing `ExpressionQuery<>` where `CountAsync` tests `IQueryProviderAsync`. Referencing #4404 from this PR, or reopening it and closing it as a duplicate of #5805, would make the history findable.

## Out-of-scope observations

- **`IQueryableWrapper<T>` is internal, so the repo's own EF wrapper cannot participate.** `LinqToDBForEFQueryProvider<T>` (`Source/LinqToDB.EntityFrameworkCore/Internal/LinqToDBForEFQueryProvider.cs:30`) is a genuine query wrapper — it holds an `ExpressionQuery<T>` and delegates `Expression`, `Provider` and `DataContext` to it — but it lives in another assembly and so cannot implement an internal marker. For every method with an `IExtensionsAdapter` fallback this costs nothing: they route through EF's own async extensions, which find the same `IAsyncEnumerable<T>`. `ForEachUntilAsync` and both `ToLookupAsync` overloads are the exception — `IExtensionsAdapter` declares neither, so `AsyncExtensions.cs:251` and `:498` fall straight to `Task.Run` over synchronous enumeration for a `ToLinqToDB()` query. Unchanged from `master`, so not caused by this PR. Widening the marker to public would close it, at the cost of a new public API entry under `LinqToDB.Internal.Linq`; keeping it internal has the merit of making the unwrap loop cycle-free by construction. Flagged only because this PR is what establishes the marker, so the accessibility choice is being made now rather than inherited.
- **`AsAsyncEnumerable` throws for a `LoadWith` wrapper over a non-linq2db source, while `ToListAsync` on the same query succeeds.** Reproduced on `master`: `Enumerable.Empty<Child>().AsQueryable().LoadWith(c => c.Parent)` takes `LoadWith`'s `Provider is not IQueryProviderAsync` arm, so the wrapper holds the plain in-memory query. `ToListAsync` returns normally (count 0); `AsAsyncEnumerable` is caught by the pre-existing `source is IAsyncEnumerable<TSource>` branch and enumerating reaches `LinqExtensions.LoadWith.cs:111-114`, giving `LinqToDBException: Async enumeration is not supported for LoadWith source 'System.Linq.EnumerableQuery`1[Child]'.` `LoadWith`'s own `<remarks>` document the non-linq2db case as a supported passthrough mirroring EF Core `Include`, so the throw reads as unintended. Pre-existing; calling `UnwrapQueryable()` before the `IAsyncEnumerable` test would make the two consistent.
- **`ITable<T>`-mutating extensions have the same wrapper blind spot on `DatabaseSpecificTable<TSource>`** — tracked as #5815 · not caused by this PR. Nine sites hard-cast `((ITableMutable<T>)table)`; `TempTable<T>` delegates the interface but `DatabaseSpecificTable<TSource>` does not implement it, so `db.Child.AsSQLite().TableName("Child2")` and `.TableOptions(...)` throw `InvalidCastException` while the reverse order works. Outside the new marker's reach regardless, since `WrappedQuery` is typed `IQueryable<T>` rather than `ITable<T>`.
- **`ToSqlQuery` on an eager-loaded `ITable<T>` association throws** — tracked as #5816 · not caused by this PR. `PersistentTable<T>` satisfies `IQueryProviderAsync` through `ITable<T>`, so `GetLinqToDBSource` accepts it, and `ToSqlQuery` then indexes `[0]` into the `Array.Empty<QuerySql>()` returned at `PersistentTable{T}.cs:36`, giving `ArgumentOutOfRangeException`. Marking `PersistentTable<T>` with the new interface would not fix it — the unwrap yields an `EnumerableQuery`, still not an `ExpressionQuery<T>`.

## Baselines

Delta: [linq2db.baselines#2076](https://github.com/linq2db/linq2db.baselines/pull/2076) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5809)

16 files added, 0 modified, 0 deleted (+136 / −0) — exactly the four new tests across the four SQLite configs (`SQLite.Classic`, `SQLite.Classic.MPM`, `SQLite.Classic.MPU`, `SQLite.MS`). No existing baseline moved, and the whole diff shows zero `DECLARE` movement and zero cast movement, consistent with a change that touches no SQL generation. Note this predates `4af12d8fc`, so the two tests added there are not yet represented.

The captured SQL corroborates the mechanism rather than merely recording it:

- `AsyncTests.LoadWithQueryMaterializedAsynchronously` — `SELECT [t1].[ParentID], [t1].[ChildID], [a_Parent].[ParentID], [a_Parent].[Value1] FROM [Child] [t1] LEFT JOIN [Parent] [a_Parent] ON [t1].[ParentID] = [a_Parent].[ParentID]`. Unwrapping the `LoadWith` wrapper preserved the eager-load directive.
- `LoadWithTests.LoadWithQueryIsLinqToDbSource` — the same join plus `LIMIT 1 OFFSET 0`, confirming `ElementAtOrDefault(() => 0)` bound to the linq2db overload and translated server-side rather than falling back to the BCL one.
- `AsyncTests.DatabaseSpecificTableMaterializedAsynchronously` — plain `SELECT [t1].[ParentID], [t1].[ChildID] FROM [Child] [t1]`, as expected for `AsSQLite()`, whose `Sql.QueryExtension` scope is `None` and emits no hint.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

**Improved**

- Async materialization now really uses async ADO.NET for queries carrying a wrapper. `await query.LoadWith(…).ToListAsync()` fell back to a synchronous read on a thread-pool thread — blocking it for the whole read, and only observing the `CancellationToken` between rows — and `ToArrayAsync`, `ToDictionaryAsync`, `ToLookupAsync`, `ForEachAsync` and `ForEachUntilAsync` behaved the same way. `AsAsyncEnumerable` was affected too, though it read synchronously on the calling thread rather than a pooled one; that improvement applies to `CreateTempTable` results and to queries carrying a provider hint such as `AsSqlServer()` / `AsSQLite()`.

**Fixed**

- LinqToDB-only operations threw *"LinqToDB method 'X' called on non-LinqToDB IQueryable"* when applied to a query carrying `LoadWith` — `Insert`, `Update`, `Delete`, `Merge`, `MultiInsert`, `ElementAt` / `ElementAtOrDefault` and the analytic-function overloads.


Async materialization now really uses async ADO.NET for queries carrying `LoadWith`, a temp table or a provider hint.


---
<sub>Generated from commit `4af12d8`.</sub>

Issue #5812, opened 2026-08-24
Declare IInfrastructure<IServiceProvider> on IDataContext


MaceWindu:
`IDataContext` is public and implementable, but the query pipeline unconditionally casts it to `LinqToDB.Internal.Infrastructure.IInfrastructure<IServiceProvider>`, which the interface doesn't declare:

- `Source/LinqToDB/Internal/Linq/Builder/Visitors/ExposeExpressionVisitor.cs:56` — `GetRequiredService<IMemberConverter>()`
- `Source/LinqToDB/Internal/Linq/Builder/ExpressionBuilder.cs:168` — `GetRequiredService<IMemberTranslator>()`
- `Source/LinqToDB/Internal/Linq/QueryRunner.DropTable.cs:108`

Any `IDataContext` implementation that isn't `DataConnection` / `DataContext` / `RemoteDataContextBase` therefore throws on **every** query build:

```
System.InvalidCastException : Unable to cast object of type 'MyContext' to type
'LinqToDB.Internal.Infrastructure.IInfrastructure`1[System.IServiceProvider]'.
   at LinqToDB.Internal.Linq.Builder.Visitors.ExposeExpressionVisitor.ExposeExpression(...)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.ExposeExpression(...)
```

The requirement is real and there is no alternative route — `IDataContext` exposes no `DataProvider`, so this interface is the only seam to the provider's service container. Today it is undeclared and undiscoverable, and surfaces as a runtime cast failure rather than a compile error.

### Proposal

Add it to the interface's base list:

```csharp
public interface IDataContext : IConfigurationID, IInfrastructure<IServiceProvider>, IDisposable, IAsyncDisposable
```

`IDataContext` already inherits `IConfigurationID` from `LinqToDB.Internal.Common`, so an `Internal.*` base is precedented.

Breaking for existing implementers, hence 7.0.0 — though they are already broken at runtime, so this converts a runtime crash into a compile error.

Introduced by `9cc8fae38` (#5168).

Found while fixing #4365, where `Tests/Model/TestDataCustomConnection.cs` had to implement the interface by hand before a custom-context regression test could run at all.

PR #4754, opened 2024-12-23
Reorganize Namespaces

viceroypenguin:

Issue #3266, opened 2021-10-15
CompiledQuery  with Async Update doesn't work


MaceWindu:
CompiledQuery work fine with sync calls and increases speed significantly.
However when using it with async calls I get exceptions on some calls.
For example: 
`CompiledQuery.Compile((Database d, string q) =>  d.Positions.Where(p => p.Code== q).ToArrayAsync(default)); ` 
works well, 
whereas
`CompiledQuery.Compile((Database d, string q) =>  d.Positions.Where(p => p.Code == q).Set(p=>p.Value,0).UpdateAsync(default))` doesn't work, `await Invoke(...)` throws
LinqToDB.Linq.LinqException: 'Sequence ... cannot be converted to SQL.'

_Originally posted by @alexarchen in https://github.com/linq2db/linq2db/discussions/3262_

MaceWindu:
Then it's probably related to fixed issue with bad caching of queries with `Set` function

MaceWindu:
Issue confirmed. For some of APIs we don't build async versions directly due to expression "rewrite".
So in theory it also affects users who build queryable expression manually. Prepare fix soon

sdanyliv:
Heads-up for whoever picks this up: PR #5844 changes this issue's failure mode, so the message recorded above will no longer be what you see.

`CompileTestsAsync.Issue3266Test` compiles `db.Person.Where(...).Set(...).UpdateAsync(default)`. `LinqExtensions` is on `MethodCallExpression.IsQueryable`'s declaring-type allowlist and the call is typed `Task<int>`, for which `IEnumerable<>.GetGenericType` returns null — so `elementType == null` and it folds as `MethodType.Element`.

- **Before #5844:** `CompiledTable<T>.GetInfo` ran `ReplaceAsyncWithSync` on it, which found no candidates and threw `InvalidOperationException("Cannot convert async method call to sync.")`. That is the exception this test currently fails with.
- **After #5844:** de-asyncing runs only in `CompileQuery`'s `[ElementAsync]` branch, which this call does not take. The `UpdateAsync` node reaches the expression builder in async form instead, so the failure is different.

Both are broken, and `[ActiveIssue]` marks the test Explicit, so CI shows nothing either way.

[...]

PR #5813, opened 2026-08-24
Support interceptors on custom IDataContext, fix compiled query over ToList()

### LINKED ISSUE #4365: CompiledQuery do not call IEntityServiceInterceptor.EntityCreated()?
### Describe your issue

IEntityServiceInterceptor.EntityCreated() is not called when using a CompiledQuery. It works fine using a normal query.

### Steps to reproduce

using var db = new MyContext(); // ctor calls DataConnection.ctor(new DataOptions().UseInterceptor(this))

var query = CompiledQuery.Compile<MyContext, CancellationToken, Task<List<MyDTO>>>(static (db,  ct) => db.MyTable.Where(...).ToListAsync(ct));
var result = await query(db, default);  // This will not call MyContext.EntityCreated();

However, this will call EntityCreated():
var query = await db.MyTable.Where(...).ToListAsync();  // This works

### Environment details

`Linq To DB` version: 5.3.2

Database (with version): SQLite 3.42.0

ADO.NET Provider (with version): System.Data.SQLite 1.0.118.0
Operating system: Win11 x64

.NET Version: 8.0

### LINKED ISSUE #5803: Public IInterceptable?

MaceWindu:
Supersedes #5804 (carries @cal-tlabwest's commit — the fork is org-owned, so "allow edits by maintainers" cannot be enabled and the branch could not be extended in place).

Fixes #4365
Fixes #5803

## Interceptors on a custom `IDataContext`

Interceptor dispatch is driven entirely by `dataContext is IInterceptable<T>` type tests (~20 sites, e.g. `EntityConstructorBase.cs:855`, `QueryFlagsHelper.cs:16`). `IDataContext` exposes only the write side — `AddInterceptor` / `RemoveInterceptor` — so while `IInterceptable` is `internal`, a custom context can accept an interceptor and nothing will ever call it.

Both interfaces were `public` in `LinqToDB.Interceptors` through v5.4.1.9 and remained `public` in `Internal.Interceptors` after #4754; the visibility was dropped in #5027, a 374-file "remove public visibility from some internal apis" sweep whose criterion — "not used outside the assembly" — cannot see external implementers.

`TestDataCustomConnection` now implements `IInterceptable<IEntityServiceInterceptor>` and `IInfrastructure<IServiceProvider>`, making it an actually-complete custom context. The second one is undeclared on `IDataContext` but hard-cast by the query pipeline; declaring it is breaking, so it is filed separately as #5812 for 7.0.0.

## Compiled query ending in a client-side materializer

This turned out not to be custom-context-specific: `IDataContext_CompiledQueryTest_AsList` uses the plain `TestDataConnection` and had been `[ActiveIssue]` since #4540.

`Enumerable.ToList` satisfies `MethodCallExpression.IsQueryable`, so `CompiledQuery` treated `.ToList()` as the table boundary. `List<T>` is not `IQueryable`, so the `IEnumerable<>` branch matched, `qtype` came back non-null as `IEnumerable<Person>`, `MethodType.Queryable` was selected with `T = Person`, and the resulting `Query<Person>` was built over an expression still ending in `.ToList()` — which `BuildSequence` cannot translate.

The async counterpart works because `AsyncExtensions` is not in the `IsQueryable` list at all, so `ToListAsync` is left in place and only the *inner* queryable is compiled. The fix makes the sync path behave the same: use the queryable form only when the call's static type can hold the `Table<T>` that `CompiledTable<T>.Create` substitutes.

That predicate also means no previously-working path shifts — if the old code worked, the substituted `Table<T>` was assignable, so the guard is true and behaviour is identical.

## Verification

Local, SQLite.Classic / net10.0:

- both `CompileTests` fixtures — 148/148
- widened to every fixture that uses `CompiledQuery` — 1119/1119
- Release builds of `Source/LinqToDB` on `net10.0` (analyzers) and `netstandard2.0` (portable API surface) — both clean
- red→green confirmed: before the `CompiledQuery.cs` change both `_AsList` tests fail with the builder error; before the `TestDataCustomConnection` change both `CustomContext_` tests fail with `Expected: same as <Person>`

Each commit is independently green, so the branch is bisect-clean.

## Not fixed here

#3266 (async `Update` in a compiled query) was probed with its gate removed and is unaffected — `UpdateAsync` is typed `Task<int>`, which the `IEnumerable<>` match does not decompose, so it takes the same branch before and after. Its gate is left in place. Its current failure is `InvalidOperationException : Cannot convert async method call to sync.` from `ReplaceAsyncWithSync`, which looks for a sync `Update` counterpart on `Task<int>.DeclaringType` (null) and on `Queryable` (no `Update`) but never on `LinqExtensions`, where `Update` actually lives.

### DIFF
--- a/Source/LinqToDB/CompiledQuery.cs
+++ b/Source/LinqToDB/CompiledQuery.cs
@@ -119,11 +119,20 @@ namespace LinqToDB
 									typeof(IQueryable<>) :
 									typeof(IEnumerable<>);
 
-							var qtype  = type.GetGenericType(expr.Type);
-							var helper = ActivatorExt.CreateInstance<ITableHelper>(
-								typeof(TableHelper<>).MakeGenericType(qtype == null ? expr.Type : qtype.GetGenericArguments()[0]));
+							var qtype       = type.GetGenericType(expr.Type);
+							var elementType = qtype?.GetGenericArguments()[0];
 
-							return helper.CallTable(context.query, expr, context.ps, context.preambles, qtype != null ? MethodType.Queryable : MethodType.Element);
+							// The Queryable form substitutes a Table<T>, so it applies only when the call's static type
+							// can hold one. A client-side materializer such as ToList() is typed List<T> and cannot:
+							// leave the call in place so the inner queryable becomes the compiled table instead, which
+							// is what already happens for its async counterpart.
+							if (elementType == null || expr.Type.IsAssignableFrom(typeof(Table<>).MakeGenericType(elementType)))
+							{
+								var helper = ActivatorExt.CreateInstance<ITableHelper>(
+									typeof(TableHelper<>).MakeGenericType(elementType ?? expr.Type));
+
+								return helper.CallTable(context.query, expr, context.ps, context.preambles, elementType != null ? MethodType.Queryable : MethodType.Element);
+							}
 						}
 
 						if (string.Equals(expr.Method.Name, "GetTable", StringComparison.Ordinal) && expr.Method.DeclaringType == typeof(DataExtensions))
--- a/Source/LinqToDB/Internal/Interceptors/IInterceptable.cs
+++ b/Source/LinqToDB/Internal/Interceptors/IInterceptable.cs
@@ -2,11 +2,11 @@
 
 namespace LinqToDB.Internal.Interceptors
 {
-	interface IInterceptable
+	public interface IInterceptable
 	{
 	}
 
-	interface IInterceptable<T> : IInterceptable
+	public interface IInterceptable<T> : IInterceptable
 		where T : IInterceptor
 	{
 		T? Interceptor { get; set; }
--- a/Tests/Linq/Linq/CompileTests.cs
+++ b/Tests/Linq/Linq/CompileTests.cs
@@ -575,7 +575,6 @@ namespace Tests.Linq
 			}
 		}
 
-		[ActiveIssue]
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/4365")]
 		public void IDataContext_CompiledQueryTest_AsList([DataSources(false)] string context)
 		{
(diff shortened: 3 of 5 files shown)


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

**Added**

- `IInterceptable` / `IInterceptable<T>` are public again, in the `LinqToDB.Internal.Interceptors` namespace, so a hand-written `IDataContext` can receive interceptors. Implement `IInterceptable<IEntityServiceInterceptor>` (and one per other interceptor kind you want) and store the interceptor; you will likely also need `IInfrastructure<IServiceProvider>` from `LinqToDB.Internal.Infrastructure`, which the reference implementation in this PR adds alongside it. Without them a context could accept an interceptor through `AddInterceptor` and never have it called. Thanks to [@cal-tlabwest](https://github.com/cal-tlabwest).

**Fixed**

- A compiled query whose body ends in a client-side materializer failed on first invocation:

  ```csharp
  CompiledQuery.Compile((IDataContext db, int id) =>
      db.GetTable<Person>().Where(p => p.Id == id).ToList());
  ```

  `ToArray`, `ToDictionary`, `ToLookup` and `ToHashSet` are covered too; the async forms already worked.


`IInterceptable` is public again so a hand-written `IDataContext` can receive interceptors, and a compiled query ending in `ToList()` / `ToArray()` now works.


---
<sub>Generated from commit `55d990a`.</sub>

Issue #2078, opened 2020-02-05
how remove SqlFunctions in T4 models


MajidSafari:
how  remove SqlFunctions in T4 models 

like this

```
{
public static partial class SqlFunctions
	{
		[Sql.Function(Name=dbo.CountString, ServerSideOnly=true)]
		public static int? CountString(string @pSearchString, string @pInput)
		{
			throw new InvalidOperationException();
		}
}}
```

MaceWindu:
https://linq2db.github.io/articles/T4.html#configuring-schema-load-process
```cs
// Enables loading of functions and procedures information
GetSchemaOptions.GetProcedures         = true;
```

MajidSafari:
hi
i want just functions  ignore and Procedures create model

sdanyliv:
Everything in documentation:
```cs
GetSchemaOptions.GetProcedures     = true;
GetSchemaOptions.LoadProcedure     = (ProcedureSchema p) => !p.IsFunction;
```

MajidSafari:
tanx 

i do this 

```
GetSchemaOptions.GetProcedures     = true;
GetSchemaOptions.LoadProcedure     = (ProcedureSchema p) => !p.IsFunction;
```

  but

i want

block scaler value func

	public static partial class SqlFunctions
	{
		#region CountString

		[Sql.Function(Name="dbo.CountString", ServerSideOnly=true)]
		public static int? CountString(string @pSearchString, string @pInput)
		{
			throw new InvalidOperationException();
		}

![Capture](https://user-images.githubusercontent.com/5099564/74081887-038fd200-4a69-11ea-9caf-523bcbdde6fb.PNG)

sdanyliv:
So you have to write appropriate filter. Find result parameter in Parameters property.
https://github.com/linq2db/linq2db/blob/5e26fa1693208b92b0791dff445ab7b42e79fb1b/Source/LinqToDB/SchemaProvider/ProcedureSchema.cs#L81

MaceWindu:
Check ProcedureSchema class properties. If you want to block only scalar functions, then you need to update load procedure condition to allow table functions (and aggregates):
```cs
!p.IsFunction || p.IsTableFunction || p.IsAggregateFunction
```

MajidSafari:
> Check ProcedureSchema class properties. If you want to block only scalar functions, then you need to update load procedure condition to allow table functions (and aggregates):
> 
> ```cs
[...]

PR #5818, opened 2026-08-25
Fix set-operation projection rejecting an untranslatable concat operand

MaceWindu:
## Problem

A projection member that mixes string concatenation with an expression that has no SQL translation — e.g. `"p_" + x.Id.ToString("N")` — throws when that projection is an **operand of a set operation**:

```
LinqToDB.LinqToDBException: The LINQ expression 'x.Id.ToString("N")' could not be converted to SQL.
   at LinqToDB.Internal.Linq.Query`1.CreateQuery(...)
   at LinqToDB.Internal.Linq.Query`1.GetQuery(...)
```

The same projection in a terminal `Select` has always worked — the untranslatable part is deferred to the materializer. Provider-independent.

Regression from #5504, first shipped in **6.3.0**. On `d4559979a` (its parent) the query builds and returns correct rows.

## Cause

A set operand's projection is built with `BuildFlags.ForSetProjection`, which makes `GetTranslationFlags()` report `TranslationFlags.Sql` — read by translators as "the caller explicitly asked for SQL emission". #5504 routed binary string `+` into `StringMemberTranslatorBase.TranslateConcat`, which under that flag returns an **error expression** for an operand it cannot translate rather than declining with `null`. The error escapes `MergeProjectionHelper.BuildProjectionExpression` → `SetOperationBuilder.InitializeProjections` and kills the whole set operation.

The flag conflates *prefer* SQL with *require* SQL: a set operand's projection carries a client-side computation perfectly well, as long as the operands select the columns it reads.

## Fix

`ExpressionBuildVisitor.DeclinedForSetProjection` — while building a set projection, treat a non-critical `SqlErrorExpression` from `TranslateMember` as "not translated", so the two call sites (`VisitBinary`, `VisitMethodCall`) fall through to the client-side path exactly as they do for a terminal `Select`.

Generated SQL is identical to the pre-#5504 baseline — each operand selects the constant and column placeholders the projection reads:

```sql
SELECT CAST('p_' AS NVarChar(255)), [p].[Id], CAST('N' AS NVarChar(255)), [p].[Name] FROM [ProbeParent] [p]
UNION ALL
SELECT CAST('t_' AS NVarChar(255)), [t].[Id], CAST('N' AS NVarChar(255)), [t].[Name] FROM [ProbeTask]   [t]
```

Two notes on the shape of the fix:

- `RegisterTranslatedSql` only caches `SqlPlaceholderExpression`, so a declined error is never cached and cannot poison a later build of the same expression.
- Rejected alternative: retrying the projection build *without* `ForSetProjection`. It fixes `Concat` but breaks `Union` with `Could not decide which construction type to use` — dropping the flag also stops constants becoming SQL value columns, and those are what `GetDifferencePredicateConstants` uses as the operand discriminator.

## Tests

`Tests/Linq/Linq/StringConcatTests.cs`, all `[DataSources]` + `AssertQuery` so values are verified, not just the absence of an exception:

- `Concat_SetOperationOperand_UntranslatableConcatOperand`
- `Union_SetOperationOperand_UntranslatableConcatOperand`
- `Concat_SetOperationOperand_UntranslatableConcatOperand_DifferentShapes` — the member holding the untranslatable concat exists on one operand only, NULL-padded on the other

Red→green: 6/6 fail on master with the reported message, 6/6 pass with the fix.

Local verification: full `Tests/Linq` on SQLite.MS — 10675 total, 0 failed, 277 skipped. `ConcatUnionTests` + `StringConcatTests` + `StringJoinTests` — 375 passed / 0 failed. Release `net10.0` and `netstandard2.0` builds clean, 0 warnings.

## Out of scope

While reproducing this I found a separate, **long-standing** defect (not a regression — identical on `d4559979a`): a set operation over two `(object)`-cast, differently-shaped anonymous types silently materializes the wrong object per row — each row comes back as a `String` holding one member's value — instead of either constructing the projected type or throwing. Not touched here.

## Follow-up commits

Response to @sdanyliv's review. Two findings reproduced and fixed, three claims did not hold.

**`bb0207605` — narrow the fall-through to declinable refusals.** `{ IsCritical: false }` excluded exactly one error in `Source/` — the association handler's, which `TranslateMember` never raises — so the guard swallowed refusals translators raise *regardless* of the flags. And because `ForSetProjection` is exempt from the client-side preference, the fall-through emitted SQL rather than deferring to the materializer. Measured on SQLite: a declared duration compared with an undeclared one answered `False` where the CLR answers `True`, and the same pairing *added* answered `01:30:00.0005400` for three hours — silent wrong data from `SELECT [r].[InTicks] + [r].[UndeclaredSeconds]`, where a terminal `Select` refuses both by name.

The guard now re-asks the translator with `TranslationFlags.Expression` and declines only when that returns `null`, which separates "would have declined had SQL not been demanded" from "refuses this construct outright". `TranslateConcat` declines under `Expression`, so the case this PR exists to fix still falls through; `TranslateIntervalComparison` and `TranslateIntervalArithmetic` error whatever they are asked, so they no longer get swallowed. The second ask runs only on a path that has already failed.

The same guard went on the `HandleMember` call site, where a `TimeSpan` member refused only under strict SQL failed the whole set operation although a terminal `Select` answers it in .NET. That one is pre-existing on `master` — the guard was only ever on two of the seven `TranslateMember` call sites.

**`7bd3d7b98` — test quality.** `_DifferentShapes` renamed to `_UntranslatableConcatInDifferentMembers` with a corrected comment: `Concat` forces both operands to project the same type, so no member is absent from one side and the NULL padding is a placeholder-layout detail rather than a difference in projected shape. The `Union` test's operands were disjoint, so it answered exactly what the `Concat` test answers — they now overlap with a shared prefix so the `UNION` has a duplicate to collapse. Added a `string.Concat` case, which enters at `VisitMethodCall`; the other three all enter at `VisitBinary`.

Regression tests: `ARefusedPairingStaysRefusedInsideASetOperation` (both refused pairings inside a `Concat`) and `ADifferenceMemberReadableInAProjectionIsReadableInsideASetOperation` (asserted against the terminal projection, so it holds both where the member translates and where it falls back).

Emitted SQL for the three original tests is unchanged — still byte-identical to `baselines/pr_5818` on SQLite.MS.

**`2888553fc` — merged `master`** (through #5832; brings in #5701, #5764, #5766, #5762).

**`634dd9854` / `2f9a463d6` — second review round.** Three findings:

- **NIT001** taken — the guard summary named two users of `BuildFlags.ForSetProjection`; `EnumerableContextDynamic` builds in-memory-sequence row projections through `MergeProjectionHelper` too, so there are three.
- **SUG002** declined — the four `ExpressionBuilder.EagerLoadUnion` sites pass a `ContextRefExpression`, not a user lambda body, so an untranslatable member never reaches the fall-through there; forcing `EagerLoadingStrategy.CteUnion` over the shape still throws. Noted beside the tests. Filed #5823 for an unrelated `CteUnion` defect found while probing.
- **SUG003** declined — confining the guard to `BuildPurpose.Expression` reintroduces the original failure on all three binary-`+` tests. `StringMemberTranslatorBase.TranslateBinaryStringConcat` re-enters the visitor via `TranslationContext.Translate`, which sets `BuildPurpose.Sql` while `CombineBuildFlags` keeps `ForSetProjection` — so that state is where the fix runs, not an edge case.

Verified after both commits: `StringConcatTests` + `IntervalTranslationTests` on SQLite.MS net10.0 — 264 passed, 0 failed.


Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
```diff
@@ -2906,6 +2906,19 @@ static bool IsPrimitiveConstant(Expression expression)
...
+		/// A set-operation projection is built with <see cref="BuildFlags.ForSetProjection"/>, which makes
+		/// <see cref="GetTranslationFlags"/> ask translators for strict SQL — so a member they cannot translate
+		/// (e.g. <c>Guid.ToString("N")</c>) comes back as an error instead of being declined. Such a member is
+		/// still projectable through a set operation: each operand selects the columns it reads and the
+		/// materializer computes the value, exactly as for a terminal <c>Select</c>. Treat the error as
+		/// "not translated" so the caller falls through to the client-side path.
+		/// </summary>
+		bool DeclinedForSetProjection(Expression translated)
+		{
+			return _buildFlags.HasFlag(BuildFlags.ForSetProjection) && translated is SqlErrorExpression { IsCritical: false };
```
sdanyliv: **Blocker · BLK001** — `{ IsCritical: false }` narrows nothing, so this swallows refusals that translators raise *unconditionally* — and one of them then returns silently wrong data.

`isCritical: true` is passed at exactly one place in `Source/` — `ExpressionBuildVisitor.cs:1797`, the unsupported-subquery association error — and that one is raised by the association handler, never by `TranslateMember`. At both patched call sites the predicate therefore means "any translator error".

Some translator errors are deliberately raised whatever the flags say, precisely because there is no correct fall-through. `DateFunctionsTranslatorBase.TranslateIntervalComparison` returns `Error_Interval_UndeclaredOperand` unconditionally, and the comment directly above it states the reason and the consequence:

> Refused whether or not SQL is demanded, as the arithmetic is and for the same reason: the error is the pairing itself, so there is no half of it left for a projection to read the two sides through.

> Left to the generic handling the two stored numbers are compared as they stand - 1800 seconds against 18000000000 ticks for the same ninety minutes, which answers no rows where the CLR answers one.

It is registered for all six comparison operators over `TimeSpan` / `TimeSpan?` via `RegisterBinaryInternal`, so it reaches the `VisitBinary` site this PR patches. `PreferClientCalculation` is false under `ForSetProjection`, so that branch is always entered.

**Reproduced at this head on SQLite**, using the interval fixture's own `DurationRow` (`InSeconds` stores 5400, `Undeclared` stores 54000000000 — both ninety minutes, so the CLR answer is `True`):

```csharp
// throws Error_Interval_UndeclaredOperand — as ADeclaredDurationRefusesComparisonWithAnUndeclaredOne pins
t.Select(r => r.InSeconds == r.Undeclared).ToArray();

// same projection, now a set operand: no throw, returns [False]
t.Where(r => r.Id >  0).Select(r => r.InSeconds == r.Undeclared)
 .Concat(t.Where(r => r.Id <= 0).Select(r => r.InSeconds == r.Undeclared))
 .ToArray();
```

Both run in the same test, so the only difference is `ForSetProjection`. The emitted operand is the raw numeric comparison the translator exists to prevent:

```sql
SELECT [r_1].[InSeconds] = [r_1].[Undeclared] FROM [DurationRow] [r_1] WHERE [r_1].[Id] <= 0
```

`Error_Interval_Difference` in the same block is swallowed identically. `WindowFunctionsMemberTranslator`'s provider-capability refusals are also flag-independent — there the cost is a worse diagnostic rather than wrong data, since `Sql.Window.*` is `[ServerSideOnly]`.

Fix: separate "the translator would have declined had it not been asked for strict SQL" from "the translator refuses this construct outright" — only the first is safe to swallow. Re-asking the translator for the same node with `TranslationFlags.Expression` and falling through only when that returns `null` reproduces exactly the condition this fix targets and costs nothing on the success path; it needs the original node, so the helper takes it and both call sites pass it. A regression test wrapping `ADeclaredDurationRefusesComparisonWithAnUndeclaredOne`'s projection in a `Concat` would pin it either way.

— ✓ Fixed in `bb0207605`. Verified at HEAD `548ee23e`: the `TranslationFlags.Expression` re-ask discriminates as intended. `TranslateConcat` returns `null` under `Expression` (`StringMemberTranslatorBase.cs:241-242`), so the concat case still falls through; `Error_Interval_UndeclaredOperand` (`DateFunctionsTranslatorBase.cs:381`, `:1175`) and `Error_Interval_Difference` (`:1161`, `:1333`) carry no flag test and still error, so they are no longer swallowed. `IsUndeclaredStorage` (`:489`) still answers `true` under `Expression` flags, so the discrimination does not collapse on its own path. `ARefusedPairingStaysRefusedInsideASetOperation` pins both shapes.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
```diff
@@ -2906,6 +2906,19 @@ static bool IsPrimitiveConstant(Expression expression)
...
 		}
 
+		/// <summary>
+		/// A set-operation projection is built with <see cref="BuildFlags.ForSetProjection"/>, which makes
+		/// <see cref="GetTranslationFlags"/> ask translators for strict SQL — so a member they cannot translate
+		/// (e.g. <c>Guid.ToString("N")</c>) comes back as an error instead of being declined. Such a member is
+		/// still projectable through a set operation: each operand selects the columns it reads and the
+		/// materializer computes the value, exactly as for a terminal <c>Select</c>. Treat the error as
+		/// "not translated" so the caller falls through to the client-side path.
+		/// </summary>
```
sdanyliv, line 2934: **Minor · MIN001** — The summary describes the helper as if `BuildFlags.ForSetProjection` were only ever a set-operation thing, so a reader concludes the change is scoped to `Concat` / `Union` / `Except` / `Intersect`. `MergeProjectionHelper.cs:48` is one of five sites that set the flag — the other four are in `ExpressionBuilder.EagerLoadUnion.cs` (lines 203, 313, 471, 904), which build the parent and nested-detail projections of an eager load's CTE-union rewrite. The new guard fires there too, so a `LoadWith` detail projection containing an untranslatable operand changes from throwing to computing client-side.

That change looks desirable — a non-union eager load never sets the flag, so its detail projection already falls back to .NET, and this brings the CTE-union variant into line — but it is a behaviour change on a second feature that neither the doc nor the description mentions, and no test covers it.

Fix: name both users of the flag so the reach is visible where a reader will look for it.

```suggestion
		/// <summary>
		/// A projection built with <see cref="BuildFlags.ForSetProjection"/> — a set-operation operand, and the
		/// CTE-union branches an eager load builds in <c>ExpressionBuilder.EagerLoadUnion.cs</c> — makes
		/// <see cref="GetTranslationFlags"/> ask translators for strict SQL, so a member they cannot translate
		/// (e.g. <c>Guid.ToString("N")</c>) comes back as an error instead of being declined. Such a member is
		/// still projectable: each operand selects the columns it reads and the materializer computes the value,
		/// exactly as for a terminal <c>Select</c>. Treat the error as "not translated" so the caller falls
		/// through to the client-side path.
		/// </summary>
```

— ✓ Fixed in `bb0207605`. The summary now names the eager-load CTE-union branches alongside the set-operation operand. One user is still unnamed — `EnumerableContextDynamic.cs:105` / `:194` also construct a `MergeProjectionHelper` — raised separately as NIT001 rather than reopening this.

Tests/Linq/Linq/StringConcatTests.cs
```diff
@@ -839,5 +839,73 @@ public void TestStringConcatenation([DataSources(TestProvName.AllInformix)] stri
...
+			using var table = db.CreateLocalTable(ConcatSetOpData);
+
+			var query = table.Where(e => e.ParentId == null).Select(e => new { id = "p_" + e.Id.ToString("N"), name = e.Name })
+				.Union(table.Where(e => e.ParentId != null).Select(e => new { id = "c_" + e.Id.ToString("N"), name = e.Name }));
+
+			AssertQuery(query);
+		}
+
+		// Same defect with operands of different shape: the member holding the untranslatable concat
+		// exists on one side only, so it is NULL-padded on the other.
```
sdanyliv: **Minor · MIN002** — The comment describes something the test cannot do, and the method name repeats it. `Concat` is `IQueryable<T>.Concat(IEnumerable<T>)`, so both operands must project the *same* `T` — the compiler enforces it. Both here project `new { string id, string? name, string? parent }`; no member "exists on one side only", and the `null` in the first operand is an explicit `(string?)null` the author wrote, not padding the set operation applied.

The NULL-padding the name gestures at is real but lives one level down, in the *placeholder* layout: the two operands read different columns, so each selects NULLs in the other's slots. That is not a difference in projected shape.

The variation the test does add is worth keeping — the untranslatable concat sits in `id` on the first operand and in `parent` on the second, while `id` on the second is a bare `ToString("N")` with no concat around it, which is a different path through the projection merge than the first two tests. Consider renaming to something like `..._UntranslatableConcatInDifferentMembers`. If NULL-padding of a genuinely absent member was the intent, that needs a separate test whose operands project different anonymous types.

Fix: correct the comment to describe the variation actually exercised.

```suggestion
		// Same defect where the untranslatable concat sits in a different member on each side: the first operand
		// concatenates into `id` and passes a null literal for `parent`, the second leaves `id` a bare ToString
		// and concatenates into `parent`. Both operands still project the same anonymous type.
```

— ✓ Fixed in `7bd3d7b98`. Renamed to `Concat_SetOperationOperand_UntranslatableConcatInDifferentMembers`, and the comment now describes the placeholder layout rather than a difference in projected shape. Leaving the different-anonymous-types case out is consistent with this finding's own text, which described that shape as inexpressible through `Concat`.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
sdanyliv, line 1: **Major · MAJ001** — The guard is applied at 2 of the 7 `TranslateMember` call sites in this file, and the identical failure is still reachable through `HandleMember` at line 1628 (`VisitMember`), which is not guarded.

The concat translator's decline-vs-error shape is not unique to it. `DateFunctionsTranslatorBase.TranslateTimeSpanMember` has exactly the same shape twice — `if (!translationFlags.HasFlag(TranslationFlags.Sql)) return null;` followed by `Error_Interval_ComponentBelowResolution`, and `translationFlags.HasFlag(TranslationFlags.Sql) ? CreateErrorExpression(…, Error_Interval_Member) : null` — with the surrounding comment stating the intent outright: *"An error only propagates where SQL is required, so a projection still falls back to .NET and answers exactly."* These are `TimeSpan` members registered via `RegisterMember`, so they arrive at `VisitMember` → `HandleMember` → `TranslateMember` at line 1628 — the unpatched path.

Because `ForSetProjection` makes `GetTranslationFlags()` report strict `TranslationFlags.Sql`, a set-operation operand takes the error branch where a terminal `Select` takes the `null`/client-side branch. On Access (`IntervalResolution => Second`, `CanMeasureDifferenceInTicks => false`), using the `EventRow` model from `IntervalTranslationTests.Difference.cs`:

```csharp
// works today: falls back to .NET and answers exactly
t.Select(r => new { m = (r.FinishedOn - r.StartedOn).Milliseconds });

// throws Error_Interval_ComponentBelowResolution
t.Where(r => r.Id >  0).Select(r => new { m = (r.FinishedOn - r.StartedOn).Milliseconds })
 .Concat(t.Where(r => r.Id <= 0).Select(r => new { m = (r.FinishedOn - r.StartedOn).Milliseconds }));
```

SQLite, Sybase, SqlCe and Firebird 4/5 (`IntervalResolution => Millisecond`) hit the same wall one unit lower, with `.Microseconds`.

This one is **pre-existing on `master`** — the PR does not introduce it — but it is the same defect class the PR exists to fix, at a call site in the file the PR already edits, so flagging it makes the choice of where to apply the guard a deliberate one rather than an accident of which two sites the repro happened to hit.

Fix: either guard line 1628 the same way, or — better — move the decision into `TranslateMember` itself so every caller gets it, since the root cause is that `ForSetProjection` tells translators SQL is mandatory when the projection can in fact fall back to .NET. Do not apply the mechanical one-liner on its own: without BLK001's narrowing it widens the set of unconditional refusals that get swallowed. Lines 947 (`VisitNew`), 2255 / 2278 (`VisitUnary`) and 3553 (`TryConvertPredicate`) are worth the same look — I found no flag-gated translator that errors on those node kinds today, so they are latent rather than reachable.

— ✓ Fixed in `bb0207605`. Verified at HEAD `548ee23e`: the guard is in the `if` condition at `ExpressionBuildVisitor.cs:1628-1629`, ahead of `RegisterTranslatedSql` and the `_translationCache` write. Both refusals landing there are genuinely flag-gated (`DateFunctionsTranslatorBase.cs:1363`, `:1390`), so the member falls back to .NET; and `RegisterTranslatedSql` (`:611-614`) returns early for non-placeholders, so no error expression can be cached by either path. The four sites left unguarded (`:947`, `:2256`, `:2279`, `:3569`) are recorded as an out-of-scope observation in the verification review — latent, no reaching input constructible.

Tests/Linq/Linq/StringConcatTests.cs
sdanyliv, line 1: **Suggestion · SUG001** — The three new tests are non-vacuous (each throws on `master`, so each genuinely pins the fix) but they cover one corner of the changed surface. Gaps, in rough order of value:

- **Eager-load CTE unions.** Four of the five `BuildFlags.ForSetProjection` sites are in `ExpressionBuilder.EagerLoadUnion.cs` (see MIN001) and none is exercised. A `LoadWith` whose detail projection contains `"p_" + e.Id.ToString("N")` now changes behaviour, untested.
- **The `VisitMethodCall` half of the patch.** All three tests use binary `+`, which enters at `VisitBinary`; the `VisitMethodCall` site is reached only indirectly, when the concat translator re-enters the visitor with the rewritten `string.Concat` call. An explicit `string.Concat(a, b.ToString("N"))` or `string.Join` inside a set-operation operand would cover it directly.
- **Other set operators.** Only `Concat` and `Union` — no `UnionAll`, `Except` or `Intersect`, no chain of three or more operands, no nested set operation.

Also worth noting: with only two rows partitioned disjointly by `ParentId == null` / `ParentId != null`, the `Union` test cannot distinguish `Union` from `Concat` — nothing is ever deduplicated. Overlapping operands would let that test carry its own weight.

Fix: add at least the eager-load case and one direct `string.Concat(...)` case; extend operator coverage opportunistically.

sdanyliv (dismissed):
_Agentic review. Findings are machine-assisted and verified against source at PR HEAD; treat them as claims to check, not verdicts._

The diagnosis in the description is correct and I verified every link of it. The two structural findings are about the **breadth** of the guard, not the cause analysis. One of them reproduces as silent wrong data and is marked accordingly.

## Review notes

- [x] **CI is green on this head.** `test-all` build 22995 passed every provider leg. The blocker below is not something CI could have caught — no existing test wraps an unconditional translator refusal in a set operation.
- [x] **BLK001 was reproduced at runtime**, on SQLite at this head, using the interval fixture's own model. Details and emitted SQL are in the inline comment.
- [x] **Baselines: 204 added / 0 modified / 0 deleted** — exactly the 3 new tests x 68 providers. Nothing else in the suite moved, which confirms the change only converts a throw into a fall-through and cannot alter SQL that already worked.
- [x] **No public API surface change.** `DeclinedForSetProjection` is a private method on an internal type; no `PublicAPI.Unshipped.txt` or `CompatibilitySuppressions.xml` obligation.
- [x] **The `RegisterTranslatedSql` claim in the description holds** for the overload `TranslateMember` actually calls — `ExpressionBuildVisitor.cs:611` returns before touching the cache unless the node is a `SqlPlaceholderExpression`, so a declined error is never cached.
- [x] **The cause analysis is correct.** `MergeProjectionHelper.cs:48` sets `ForSetProjection`; `GetTranslationFlags()` escalates to `TranslationFlags.Sql`; translators that decline with `null` under `Expression` raise an error instead. `PreferClientCalculation` is false under the same flag, so `VisitBinary` reliably enters the `TranslateMember` branch.

## Out-of-scope observations

- **`Sql.AsSql` inside a set-operation operand no longer forces server-side evaluation for these shapes.** `DeclinedForSetProjection` keys only on `_buildFlags`, and `_buildFlags` survives a nested `BuildPurpose.Sql` build (`ExpressionBuildVisitor.cs:298-306` changes the purpose without touching the flags). That is load-bearing — the concat translator re-enters the visitor at `BuildPurpose.Sql`, so the guard must fire there for the fix to work at all. The side effect is that `Sql.AsSql(...)` in a set-operand projection now falls back client-side instead of producing the refusal it exists to produce. Narrowing BLK001 may resolve this too, depending on the mechanism chosen.
- **`HandleMember` caches error expressions directly**, bypassing `RegisterTranslatedSql`'s placeholder-only rule. `ExpressionBuildVisitor.cs:1632` writes `_translationCache[cacheKey] = translatedLocal` unconditionally, errors included, and `GetAlreadyTranslated` only discards a cached error under `BuildPurpose.Expression`. Harmless while line 1628 is unguarded; it needs a look if MAJ001 is addressed there.
- **Constant leaves are not deduped across set-operation operands.** In the `DifferentShapes` baseline the literal `'N'` is materialized as three separate server columns (slots 3, 5, 10) though it is identical in both operands, and `'p_'` as two. `'p_'`/`'c_'` genuinely must be columns — they differ per operand and the materializer cannot otherwise tell which branch a row came from — but `'N'` need not be. On `Access.Ace.Odbc` this costs three dead `?` parameters bound to all-zero Guids inside `IIF(False, ?, NULL)`, an expression that is constant `NULL`, plus a 3-line `DECLARE`/`SET` preamble. Not a regression (pre-#5504 this path threw) and not a correctness problem — a possible follow-up.

## Baselines

Delta: [linq2db.baselines PR #2078](https://github.com/linq2db/linq2db.baselines/pull/2078) (OPEN) · [compare view](https://github.com/linq2db/linq2db.baselines/compare/master...baselines/pr_5818)

**204 added / 0 modified / 0 deleted** — the entire delta is the 3 new tests across 68 providers. No metrics baselines changed.

### New test baselines — correct on every provider sampled

- **Same-shape operands (4 columns/operand)** — `Concat_SetOperationOperand_UntranslatableConcatOperand` and `Union_SetOperationOperand_UntranslatableConcatOperand`, 29 added lines uniformly on all 68 providers. Each operand selects `'p_'`/`'c_'`, `Id`, `'N'`, `Name`; column counts match on both sides. Verified on SqlServer.2022, PostgreSQL.15, SQLite.Classic, MySql.8.0, DB2, Oracle.19, Firebird.5, ClickHouse.Driver, Informix, YDB, SqlCe, SapHana, Sybase and all four Access variants.
- **Different-shape operands (10 columns/operand)** — `..._DifferentShapes`, 41 lines on 66 providers. Slot map verified: 1-3 left `id`, 4-5 right `id`, 6 shared `Name`, 7 left `parent` null literal, 8-10 right `parent`. NULL pads present on both sides throughout.
- **Access ODBC is the only shape deviation** (48 lines) — Access has no `CAST`, so a typed NULL becomes `IIF(False, <typed>, NULL)`, and ODBC cannot inline a Guid so it becomes a `?` parameter. The exact `IIF(False, ?, NULL)` string already appears in three master `Access.Ace.Odbc` baselines — pre-existing idiom, not introduced here.

### Cross-provider anomalies

- **MySQL/MariaDB NULL-pad casting is asymmetric within one query** (`MySql.8.0/..._DifferentShapes(MySql.8.0).sql`): the left operand casts all pads, the right leaves the three string pads bare (`NULL`) and casts only the uuid pad. Every other provider is internally consistent — PostgreSQL/DB2/Informix/ClickHouse/YDB cast all, SQLite/Oracle/Firebird/Sybase/SqlCe/SapHana leave all bare. **Not attributable to this PR**: bare `NULL,` select-list entries already exist in five master MySQL baselines, and neither guard clause can reach NULL-pad rendering. Harmless — the first UNION branch pins the column type. Cosmetic only.

### One claim I could not verify

The description states the generated SQL is *"identical to the pre-#5504 baseline"*. That is not checkable from the baselines repo — these are brand-new tests, so no pre-#5504 artifact exists to diff against. The empty modified set is consistent with it and the mechanism argument supports it, but this is inference rather than measurement.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
MaceWindu: Fixed in bb0207605. Reproduced first, and it is broader than the comparison path: `TranslateIntervalArithmetic` raises `Error_Interval_UndeclaredOperand` (line 381) and `Error_Interval_Operation` (397) unconditionally on the same `VisitBinary` route. Measured on SQLite at the old head:

- `InSeconds == Undeclared` inside a `Concat` returned `False` where the CLR answers `True`
- `InTicks + UndeclaredSeconds` inside a `Concat` returned `01:30:00.0005400` where the CLR answers `03:00:00`, from `SELECT [r].[InTicks] + [r].[UndeclaredSeconds]`

The second is a wrong *duration* rather than a wrong bool.

Mechanism confirmed as you described it: once the guard declines, `VisitBinary` reaches `TryConvertToSql` (2977), and 2210 exempts `ForSetProjection` from the client-side preference, so the node becomes SQL instead of falling back.

Took your suggested remedy. It discriminates exactly: `TranslateConcat` returns `null` under `TranslationFlags.Expression` (`StringMemberTranslatorBase` 241-242), so the concat case this PR exists to fix is still swallowed, while both interval functions error whatever they are asked. `ARefusedPairingStaysRefusedInsideASetOperation` pins both shapes.

Also checked the window-function half of your argument: `WindowFunctionsMemberTranslator.cs` has zero `translationFlags.HasFlag(TranslationFlags.Sql)` checks across its 64 `CreateErrorExpression` sites, so all of them were swallowable too.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
MaceWindu: Fixed in bb0207605, guarding the `HandleMember` site only. Reproduced on SQLite: `(FinishedOn - StartedOn).Microseconds` works in a terminal `Select` and threw `Error_Interval_ComponentBelowResolution` inside a `Concat`. Confirmed pre-existing on master, as you said.

Left 947 / 2255 / 2278 / 3553 alone rather than moving the decision into `TranslateMember`. 3553 (`TryConvertPredicate`) is the one I would not fold in blind: a declined error there falls through to generic SQL rather than to a client-side answer, which is the BLK001 failure mode again, and I have no repro either way.

`ADifferenceMemberReadableInAProjectionIsReadableInsideASetOperation` asserts the set operand agrees with the terminal projection rather than with a computed expectation, so it holds both where the member translates server-side and where both sides fall back to .NET.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
MaceWindu: Doc fix taken in bb0207605, with the wording adjusted. The census is right - 4 of the 5 flag sites are in `ExpressionBuilder.EagerLoadUnion.cs`.

The behaviour claim does not hold, though. All four of those sites build a `ContextRefExpression` (or a field of one), and with `EagerLoadingStrategy.CteUnion` forced at the old head, an untranslatable concat in an eager-loaded detail projection **still throws** ``The LINQ expression 'c.Id.ToString("N")' could not be converted to SQL``. So the summary now names the flag's other users without asserting a change in what eager loads do.

Unrelated find while probing that, for the record: with `CteUnion` forced, *any* computed member in the parent projection breaks the strategy - a fully translatable `tag = e.Name + "_x"` throws `CteUnion SetupCteUnionQuery failed: ArgumentException: must be reducible node`, while `tag = e.Name` works. Not from this PR - the guard only fires on a `SqlErrorExpression`, which a translatable concat never produces.

Tests/Linq/Linq/StringConcatTests.cs
MaceWindu: Applied in 7bd3d7b98 - comment corrected, renamed to `Concat_SetOperationOperand_UntranslatableConcatInDifferentMembers`. You are right on every point: `Concat` forces both operands to project the same `T`, the `null` is an author-written `(string?)null`, and the padding lives in the placeholder layout.

Did not add the "operands project different anonymous types" test. That shape is only writable through `(object)` casts, which hits the separate long-standing defect this PR's description lists as out of scope, so the test would pin that bug rather than this fix.

Tests/Linq/Linq/StringConcatTests.cs
MaceWindu: Two of four taken in 7bd3d7b98.

- **`string.Concat` case added** (`Concat_SetOperationOperand_UntranslatableStringConcatCall`). The other three do all enter through `VisitBinary` - the error text confirms it, `'"p_" + /* String.Concat(...) */ e.Id.ToString("N")'`.
- **`Union` operands now overlap** with a shared prefix, so the `UNION` has a duplicate to collapse. No ClickHouse gate needed: `ClickHouseSqlBuilder.BuildSetOperation` emits `UNION DISTINCT` (line 497), and the `isDistinct` handling in `SetOperatorTests` concerns duplicates *within* one operand under `EXCEPT`/`INTERSECT`.
- **Eager load**: see the MIN001 thread - not reproducible.
- **`Except`/`Intersect`**: probed rather than committed. Both pass under `AssertQuery`, which was the interesting question, since the operands select placeholder columns - so the database applies set semantics to the raw tuples while LINQ-to-Objects applies them to the computed strings, and here that mapping is injective. Not committed because the fall-through is operator-agnostic, so they pin nothing `Concat`/`Union` do not, and they would need the `isDistinct` + `[ThrowsRequiresCorrelatedSubquery]` gating this fixture has none of.

MaceWindu:
On the three out-of-scope observations, all three verified.

**`Sql.AsSql` no longer forcing server-side** - could not reproduce. The mechanism half is right (the flags do survive the purpose change at 298-306), but the consequence does not follow. At the old head on SQLite, `Sql.AsSql("p_" + e.Id.ToString("N"))` throws both in a terminal `Select` and inside a `Concat` operand - only the message widens to name the whole concat rather than the inner call. The guard now re-asks the translator anyway, so the narrowing this anticipated is moot.

**`HandleMember` caching error expressions** - accurate, and it stopped being hypothetical the moment the guard went on 1628, so it got the look you asked for. Handled: the guard sits in the `if` condition, so a declined error short-circuits past both `RegisterTranslatedSql` and the cache write, and `TranslateMember` cannot have cached it either - its own `RegisterTranslatedSql` returns early for anything that is not a `SqlPlaceholderExpression`. A later strict-SQL build re-translates and throws, which is correct. One detail: the unconditional write is 1633-1634, guarded by the `ContainsKey` check at 1632.

**Constant leaves not deduped** - confirmed exactly, `'N'` in slots 3/5/10 and `'p_'` in 1/8. Leaving it as a possible follow-up: deduping identical constant leaves across operands is a change to shared projection-merge machinery and wants its own driver rather than tidiness here. Minor correction - the Access ACE ODBC preamble is six lines, three `DECLARE`/`SET` pairs.

**On the claim you could not verify**: right that #2078 cannot show it - the three tests are new, so there is no pre-#5504 artifact to diff. It was measured locally against `d4559979a` when the fix was written. Re-checked after these two commits: the emitted SQL for all three original tests on SQLite.MS still matches `baselines/pr_5818` byte for byte, so the narrowing moved no SQL.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
```diff
@@ -2906,6 +2907,34 @@ static bool IsPrimitiveConstant(Expression expression)
...
+		{
+			if (!_buildFlags.HasFlag(BuildFlags.ForSetProjection) || translated is not SqlErrorExpression { IsCritical: false })
+				return false;
+
+			using var translationContext = _translationContexts.Allocate();
+
+			translationContext.Value.Init(this, BuildContext, Alias);
+
+			return Builder._memberTranslator.Translate(translationContext.Value, node, TranslationFlags.Expression) == null;
+		}
```
sdanyliv, line 2953: **Suggestion · SUG003** — The guard keys on `_buildFlags` alone, but the justification in its own doc comment — *"each operand selects the columns it reads and the materializer computes the value"* — holds only while `_buildPurpose` is `BuildPurpose.Expression`.

That is what all five current `BuildFlags.ForSetProjection` sites pass (`MergeProjectionHelper.cs:48` reached from `SetOperationBuilder.cs:420` / `:609` and `EnumerableContextDynamic.cs:105` / `:194`; and `ExpressionBuilder.EagerLoadUnion.cs` 203 / 313 / 471 / 904, all four explicitly `BuildPurpose.Expression, BuildFlags.ForSetProjection`). But the two pieces of state are saved and restored independently — `UsingBuildPurpose` at line 209, `UsingBuildFlags` at 214 — so a nested build can change one without the other. `BuildSearchCondition` (line 3539) does exactly that:

```csharp
using (UsingBuildContext(context ?? BuildContext))
using (UsingBuildPurpose(BuildPurpose.Sql))
```

In that state `GetTranslationFlags()` reports `Sql` because of the *purpose*, not the flag; the caller genuinely requires a `SqlPlaceholderExpression` and there is no client-side path for the fall-through to reach — yet `DeclinedForSetProjection` would still fire and swallow the refusal.

I did **not** probe whether a set-operand projection can actually drive `BuildSearchCondition` with the flag still set, so this is a tightening rather than a demonstrated defect. `VisitConditional` (line 1969) — the obvious candidate — visits its test under `CombineBuildFlags(ForceOuter)` without switching purpose, so it is not one.

Fix: confine the helper to exactly the state its doc comment describes. One extra condition, and it removes the need to reason about which purpose the visitor happens to be in when the flag is set. The suggestion also braces the now-multiline condition, per the file's convention.

```suggestion
		bool DeclinedForSetProjection(Expression node, Expression translated)
		{
			if (_buildPurpose is not BuildPurpose.Expression
				|| !_buildFlags.HasFlag(BuildFlags.ForSetProjection)
				|| translated is not SqlErrorExpression { IsCritical: false })
			{
				return false;
			}

			using var translationContext = _translationContexts.Allocate();

			translationContext.Value.Init(this, BuildContext, Alias);

			return Builder._memberTranslator.Translate(translationContext.Value, node, TranslationFlags.Expression) == null;
		}
```

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
```diff
@@ -2906,6 +2907,34 @@ static bool IsPrimitiveConstant(Expression expression)
 			return expression.NodeType == ExpressionType.Constant && (expression.Type == typeof(int) || expression.Type == typeof(bool));
 		}
 
+		/// <summary>
+		/// A projection built with <see cref="BuildFlags.ForSetProjection"/> — a set-operation operand, or one of
+		/// the CTE-union branches an eager load builds — makes <see cref="GetTranslationFlags"/> ask translators
```
sdanyliv: **Nit · NIT001** — The summary enumerates two users of the flag; there are three.

Besides `SetOperationBuilder.cs:420` / `:609` and the four `ExpressionBuilder.EagerLoadUnion.cs` sites, `EnumerableContextDynamic.cs:105` and `:194` construct a `MergeProjectionHelper` too — so the row projections of an in-memory sequence are also built with `BuildFlags.ForSetProjection` (via `MergeProjectionHelper.cs:48`) and reach this helper. A reader tracing the flag from this doc will miss that path.

```suggestion
		/// A projection built with <see cref="BuildFlags.ForSetProjection"/> — a set-operation operand, a row of an
		/// in-memory sequence, or one of the CTE-union branches an eager load builds — makes
		/// <see cref="GetTranslationFlags"/> ask translators
```

Tests/Linq/Linq/StringConcatTests.cs
sdanyliv, line 1: **Suggestion · SUG002** — Follow-up to SUG001(a): the eager-load half of `BuildFlags.ForSetProjection` is still untested, and the two paths are not obviously equivalent.

The flag reaches the visitor from three distinct places. `MergeProjectionHelper.cs:48` is called from `SetOperationBuilder.cs:420` / `:609` (the covered path) and from `EnumerableContextDynamic.cs:105` / `:194`; separately `ExpressionBuilder.EagerLoadUnion.cs` sets the flag directly at lines 203, 313, 471 and 904 — four of the five direct uses in `Source/`. All four new tests exercise the set-operation path only, while the helper's own summary now names the eager-load CTE-union branches as a user of the flag, so the change is documented as affecting a path nothing pins.

What makes the gap worth closing rather than waving through: `MergeProjectionHelper.BuildProjectionExpression` calls `SequenceHelper.FindError(projected)` at line 50 and fails the whole merge on any error, whereas the eager-load sites feed the built expression straight into `CollectDistinctPlaceholders` (lines 204, 318) with **no** error check. So on the covered path a swallowed error turns a throw into a result — visible either way — while on the eager-load path a member that previously errored and now falls through client-side would change behaviour **silently**. That asymmetry is exactly what a test would pin.

Your "not reproducible" reply on the MIN001 thread may well be right, but it is not settleable by reading — running it is what decides.

Fix: either add one eager-load regression test — a parent with a collection navigation whose detail projection contains the same untranslatable concat (`"p_" + d.Id.ToString("N")`), loaded via `LoadWith` so the CTE-union eager-load path builds it — or, if the shape genuinely cannot be constructed, record why in a one-line comment beside the existing set-operation tests, so the gap is a documented decision rather than an accident of which call site the repro happened to hit.

`UnionAll` / `Except` / `Intersect` and multi-operand chains (SUG001(c)) are worth much less and can be skipped: they all route through the same `MergeProjectionHelper`.

sdanyliv: **✓ Partial fix in `7bd3d7b98`.** Two of four confirmed at HEAD:

- **(b) the `VisitMethodCall` half** — `Concat_SetOperationOperand_UntranslatableStringConcatCall` uses `string.Concat("p_", e.Id.ToString("N"))`, a real `MethodCallExpression` that reaches the guarded call site at `ExpressionBuildVisitor.cs:883` directly. `ForSetProjection` makes `PreferClientCalculation` return false (line 2192), so the `TranslateMember` branch is entered rather than skipped.
- **(d) the `Union` test carrying its own weight** — operands now overlap (`e.Name != null` vs `e.ParentId == null`, both with the `p_` prefix), so the `UNION` has a duplicate to collapse and no longer answers exactly what the `Concat` test answers.

Residual — **(a) the eager-load CTE-union sites**, four of the five `BuildFlags.ForSetProjection` uses, remain uncovered. Raised as **SUG002** on `Tests/Linq/Linq/StringConcatTests.cs` with the reason it is worth more than the operator coverage: `MergeProjectionHelper.BuildProjectionExpression` checks `SequenceHelper.FindError` and fails the merge, while the eager-load sites feed `CollectDistinctPlaceholders` unchecked — so the same swallow is loud on one path and silent on the other.

**(c)** `UnionAll` / `Except` / `Intersect` and multi-operand chains — agreed as low value, and your reasoning (operator-agnostic fall-through, all routed through the same helper) holds. Not carried forward.

sdanyliv (commented):
> [!IMPORTANT]

## Verification update — 2026-08-27, against HEAD `548ee23e`

Re-verified the five findings from the 2026-08-26 review against `bb0207605` + `7bd3d7b98`. Four are fixed; one is partially addressed. Three new findings, all low severity.

## Review notes

- [ ] **CI has not run on this head.** Azure build 23064 is still pending on `548ee23e`. The green `test-all` cited in the previous review (build 22995) was against `7ff8dbcb3`, before both follow-up commits — so neither the narrowing nor the new tests have any CI evidence yet.
- [ ] **The baselines branch is stale** — see the Baselines section. There is no baseline evidence for either follow-up commit, and in particular `modified: 0` there is not confirmation that the narrowing moved no SQL.
- [x] **Still no public API surface change.** `DeclinedForSetProjection` remains a private method on an internal type; no `PublicAPI.Unshipped.txt` or `CompatibilitySuppressions.xml` obligation.
- [x] **No bot review coverage on this PR.** Both Copilot review attempts returned *"unable to review … quota limit"*; neither carries a claim to audit.

### Fixed

- [x] **BLK001** — flag-independent translator refusals were being swallowed, one of them returning silently wrong data. Fixed in `bb0207605`.

  The `TranslationFlags.Expression` re-ask discriminates as intended, verified against both translator families: `StringMemberTranslatorBase.TranslateConcat` returns `null` under `Expression` (`StringMemberTranslatorBase.cs:241-242`) and an error otherwise (`:244`), so the concat case this PR exists to fix still falls through; `Error_Interval_UndeclaredOperand` (`DateFunctionsTranslatorBase.cs:381`, `:1175`) and `Error_Interval_Difference` (`:1161`, `:1333`) carry no flag test at all, so the re-ask still errors and the guard does not decline. `IsUndeclaredStorage` (`:489`) keeps answering `true` under `Expression` flags — a column still translates to a placeholder — so the discrimination does not quietly collapse on the very path it was added for. `WindowFunctionsMemberTranslator.cs` has no `HasFlag(TranslationFlags.Sql)` test across its `CreateErrorExpression` sites, so its capability refusals stay errors too. `ARefusedPairingStaysRefusedInsideASetOperation` pins both interval shapes.

- [x] **MAJ001** — the identical failure remained reachable through the unguarded `HandleMember` call site. Fixed in `bb0207605`.

  The guard is now in the `if` condition at `ExpressionBuildVisitor.cs:1628-1629`, ahead of `RegisterTranslatedSql` and the `_translationCache` write. Both refusals that land there are genuinely flag-gated (`DateFunctionsTranslatorBase.cs:1363` and `:1390`), so the re-ask returns `null` and the member falls back to .NET. The placement also means the declined path registers nothing, and `RegisterTranslatedSql` (`ExpressionBuildVisitor.cs:611-614`) returns early for anything that is not a `SqlPlaceholderExpression`, so no error expression can be cached by either path.

- [x] **MIN001** — the helper's summary read as if `BuildFlags.ForSetProjection` were only a set-operation concern. Fixed in `bb0207605`. (One user is still unnamed — raised separately as **NIT001**, not as a residual on this finding.)

- [x] **MIN002** — the `_DifferentShapes` test's name and comment described a shape `Concat` cannot express. Fixed in `7bd3d7b98`: renamed to `Concat_SetOperationOperand_UntranslatableConcatInDifferentMembers`, with the comment corrected to describe the placeholder layout rather than a difference in projected shape.

### Partially addressed

- [~] **SUG001** — test-coverage gaps. Two of four taken in `7bd3d7b98`: the explicit `string.Concat` case now reaches the `VisitMethodCall` site directly, and the `Union` test's operands overlap so the query has a duplicate to collapse. The eager-load half — four of the five `BuildFlags.ForSetProjection` call sites — and the other set operators remain uncovered. Follow-up: **SUG002**.

### New findings

All three are attached as comments on the relevant lines/files:

- **SUG002** (`Tests/Linq/Linq/StringConcatTests.cs`) — the eager-load CTE-union half of the flag is untested, and the two paths handle a residual error differently.
- **SUG003** (`Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs:2926`) — the guard keys on `_buildFlags` alone, though its justification holds only while `_buildPurpose` is `Expression`.
- **NIT001** (`Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs:2911`) — the new summary names two of the three users of the flag.

## Out-of-scope observations

Surfaced during verification but not findings on this PR — included as FYI.

- **Four `TranslateMember` call sites still escalate a flag-gated refusal under `ForSetProjection`.** The fix guards three of seven. The rest keep the pre-fix behaviour: `VisitNew` (`:947`), the two `VisitUnary` sites (`:2256`, `:2279`, both gated on `node.Method != null`, so only a user-defined conversion operator reaches them), and `TryConvertPredicate` (`:3569`). No reaching input could be constructed for any of the four, so this is latent rather than live — worth recording because the next translator to add a flag-gated refusal on a constructor or conversion-operator path reopens it silently, and because centralising the decision inside `TranslateMember` would close all seven at once.
- **The declined path now costs two translator invocations.** `DeclinedForSetProjection` re-runs the translator over the same node, and translators do real work before refusing — `TranslateConcat` rewrites every argument through `ConvertOperandToString` and calls `TranslateToSqlExpression` on each until one fails. It is bounded to a path that has already failed and most sub-translations hit the cache, so it is unlikely to matter; but note that when the guard declines at `HandleMember` nothing is written to `_translationCache`, so a member visited repeatedly across build passes pays both asks each time. Worth a look only if query-build time regresses on projections leaning on client-side concat.

## Baselines

**The baselines branch is stale — this pass is not authoritative for either follow-up commit.**

- `baselines/pr_5818` tip is `c614a8dd97`, committed **2026-08-25T17:49Z** — before `bb0207605` and `7bd3d7b98` (both 2026-08-26T18:00Z).
- Baselines PR [linq2db.baselines#2078](https://github.com/linq2db/linq2db.baselines/pull/2078) was last updated 2026-08-25T19:07Z. No CI run has fired against the new commits.
- The branch still reports **204 added / 0 modified / 0 deleted** for the same three test names as the previous review, including the *pre-rename* `..._DifferentShapes`. There are no baselines for the renamed test, the overlapped `Union`, `Concat_SetOperationOperand_UntranslatableStringConcatCall`, or the two new interval tests.

Consequences for what can and cannot be concluded:

- `modified: 0` here predates the commit that would have moved SQL, so it is **not** evidence that the narrowing left the existing suite untouched. The claim that the three original tests are byte-identical on SQLite.MS is currently unverified from the baselines side; the pending build 23064 is what will settle it.
- The MySQL/MariaDB NULL-pad cast asymmetry in `..._DifferentShapes` is unchanged in this snapshot and the previous review's "not attributable to this PR / cosmetic" reading still looks right — the narrowing changes *when* the client-side fallback triggers, not the shape once it does — but that is inference, not measurement, until CI regenerates.

Re-check the baselines once build 23064 lands; that is the first run capable of showing rename orphans, the new tests' SQL, and any diff on the three original tests.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
MaceWindu: Declined — measured.

Applied the suggestion verbatim and ran `StringConcatTests` + `IntervalTranslationTests` on `SQLite.MS` (net10.0, 264 tests): **6 failed**, all six the binary-`+` set-operation cases — `Concat_SetOperationOperand_UntranslatableConcatOperand`, `Concat_SetOperationOperand_UntranslatableConcatInDifferentMembers`, `Union_SetOperationOperand_UntranslatableConcatOperand`, each on `SQLite.MS` and `SQLite.MS.LinqService` — with the pre-fix error:

```
LinqToDBException : The LINQ expression 'e.Id.ToString("N")' could not be converted to SQL.
```

`Concat_SetOperationOperand_UntranslatableStringConcatCall` (the `VisitMethodCall` site) and both interval tests stayed green.

So `BuildPurpose.Sql` with the flag still set is not an edge state to exclude — it is the state the fix runs in, and the translator callback is what creates it. Instrumenting the guard to dump `_buildPurpose` plus a stack per call, over `Concat_SetOperationOperand_UntranslatableConcatOperand` alone, gives 28 hits — per operand, three under `Sql` followed by four under `Expression`. The `Sql` ones look like this:

```
purpose=Sql nodeType=Call node=Concat("p_", Ref(...ConcatSetOpEntity).Id.ToString("N"))
  ExpressionBuildVisitor.DeclinedForSetProjection
  ExpressionBuildVisitor.VisitMethodCall                    ExpressionBuildVisitor.cs:883
  ExpressionBuildVisitor.BuildExpression                    ExpressionBuildVisitor.cs:265
  ExpressionBuilder.BuildSqlExpression                      ExpressionBuilder.Expressions.cs:21
  ExpressionBuildVisitor.TranslationContext.Translate       ExpressionBuildVisitor.cs:5394
  StringMemberTranslatorBase.TranslateBinaryStringConcat    StringMemberTranslatorBase.cs:118
  StringMemberTranslatorBase.TranslateOverrideHandler       StringMemberTranslatorBase.cs:103
  MemberTranslatorBase.Translate                            MemberTranslatorBase.cs:91
  ...
  ExpressionBuildVisitor.TranslateMember                    ExpressionBuildVisitor.cs:5607
```

`TranslationContext.Translate` (`ExpressionBuildVisitor.cs:5388`) derives the purpose from the *translation flags* — `GetBuildPurpose(TranslationFlags.Sql)` → `BuildPurpose.Sql` — and passes `BuildFlags.None`, which `BuildExpression` (`:257-263`) merges through `CombineBuildFlags`, so `ForSetProjection` survives while the purpose flips. That is exactly the divergence the finding describes, produced here by `TranslateBinaryStringConcat` re-entering the visitor to translate the concat it was handed. Confining the guard to `BuildPurpose.Expression` blocks that inner decline, the outer translator errors again, and the regression returns.

Worth noting for the `BuildSearchCondition` half of the argument: the purpose-only switch is not specific to it. Any `BuildSqlExpression` overload routes through `BuildExpression(…, BuildPurpose.Sql)` and leaves `_buildFlags` alone, so the state is ordinary on this path rather than exotic.

Guard left keyed on `_buildFlags` alone.

Tests/Linq/Linq/StringConcatTests.cs
MaceWindu: Declined — it was run, before this review.

**Structure.** All four `ExpressionBuilder.EagerLoadUnion` sites (203 / 313 / 471 / 904) pass a `ContextRefExpression` — `mainRef`, `cteDetailRef`, `nestedCteDetailRef`, `parentCteRef` — an entity/detail expansion, never the user's lambda body. Re-checked at HEAD.

**Behaviour.** Probed with

```csharp
t.Select(e => new { e.Id, det = t.Where(c => c.ParentId == e.Id).Select(c => "p_" + c.Id.ToString("N")).ToArray() })
```

on the default strategy and with `EagerLoadingStrategy.CteUnion` forced. Default took the two-query path and answered correctly; `CteUnion` **still threw** `The LINQ expression 'c.Id.ToString("N")' could not be converted to SQL`. The guard does not fire there, so the `CollectDistinctPlaceholders` sites never see an error it could have swallowed — the loud-vs-silent asymmetry the finding rests on is not reachable, and a test written for it would pass on `master` too.

Recorded beside the set-operation tests instead, per your second option — `2f9a463d6`.

Unrelated defect found while probing, filed as #5823: under `EagerLoadingStrategy.CteUnion`, *any* computed member in the parent projection — including a fully translatable `tag = e.Name + "_x"` — throws `CteUnion SetupCteUnionQuery failed: ArgumentException: must be reducible node`, while `tag = e.Name` works and the same query succeeds under `Default`. Reproduces on master `a180e8c06`, so not guard-related.

Source/LinqToDB/Internal/Linq/Builder/ExpressionBuildVisitor.cs
MaceWindu: Taken — `EnumerableContextDynamic.cs:105` / `:194` confirmed. The summary now names all three users, with your wording (rewrapped to the block width): `634dd9854`.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A projected member inside a set-operation branch that mixes translatable parts with a part that has no SQL translation failed the whole query with *"The LINQ expression … could not be converted to SQL"*, where the same projection outside a set operation falls back to .NET:

  ```csharp
  t.Where(e => e.ParentId == null).Select(e => new { id = "p_" + e.Id.ToString("N"), e.Name })
   .Concat(t.Where(e => e.ParentId != null).Select(e => new { id = "c_" + e.Id.ToString("N"), e.Name }))
  ```

  This applies to every set operation — `Concat` / `UnionAll`, `Union`, `Except`, `ExceptAll`, `Intersect` and `IntersectAll` — and to a projection over a row of an in-memory sequence. Regression in 6.3.0. A member the provider refuses outright — an interval it cannot measure, for instance — still fails rather than silently degrading.


A projection mixing translatable and client-side parts inside a set-operation branch no longer fails the whole query (regression in 6.3.0).


---
<sub>Generated from commit `3ba0410`.</sub>

PR #5819, opened 2026-08-25
CI: cut test-all's prefix, tail and per-leg overhead

sdanyliv:
Five changes to `test-all`, measured against real runs rather than modelled. Baseline is build 22995 on master: 28 legs, 1199 agent-minutes, 10 parallel slots, 264.2 min end to end.

| | build prefix | test phase | end to end |
|---|---|---|---|
| 22995, master | 28.3 | 157.3 | **264.2** |
| 23024, sparse baselines + parallel win-x86 | 23.9 | 157.0 | 175.2 |
| 23028, + lettered keys and the solution filter | **16.1** | **140.6** | **156.8** |
| 23050, + `maxParallel: 6` | 18.9 | 183.4 | 202.3 |
| 23059, + the matrix sorted by key | 19.6 | 144.8 | 164.4 |
| **23063, + the cap dropped again** | **17.0** | **131.4** | **148.4** |

Per-leg agent time went 1199 to 1130 min over the same 28 legs. `Checkout test baselines` fell from 37.0 to 5.5 agent-minutes and artifact extraction from 25.7 to 18.3.

### Check out the baselines index without its tree

Every leg cloned all 343541 baseline files; on 22995 a windows leg spent 109 of its 135 checkout seconds laying them out. None are read — `BaselinesWriter` creates its own directories and always writes whole files, and its only `ReadAllText` is against a path the same run already wrote. `--sparse` leaves the cone at the two root files and `--depth 1 --single-branch` drops ~420 Mb of history to ~35 Mb.

`git add --sparse` is the other half of the pairing, not a precaution: with a root-only cone a plain `git add -A` refuses the written paths, exits 1 and stages nothing. With it, the add stages exactly what the leg wrote and no deletions for the ~120 configurations it never touches.

### Build the win-x86 test artifacts in a parallel job

`-a x86` rebuilds the whole graph for that RID, so the eight win-x86 publishes cost 1.10 to 1.37 min each against 0.18 to 0.37 for their x64 counterparts, which reuse `Build Solution for Tests` — 6.8 of the prefix's 28.3 minutes. `build_x86_job` starts alongside `build_job`, in a window where nine of the ten slots sit idle anyway. The margin is thinner than it reads: it finished 36 s ahead of `build_job` on 23050 and 0.6 min on 23028, and the windows legs are kept clear of it only because `create_baselines_branch` then takes another two to three minutes.

Those binaries also rode along inside every per-TFM artifact, so all 15 linux legs downloaded win-x86 binaries they cannot execute. They ship as their own artifact now and a windows leg fetches only the architecture it runs.

### Build only the test graph

`Build Solution for Tests` compiled all 51 projects in `linq2db.slnx`, of which the five that `Build Test Artifacts` publishes pull in 28. The other 23 are what made the step long at both ends: `LinqToDB.CLI` owns the slowest restore in the run at 1.4 min and finished one last, and `Tests.T4` was the last project to compile, at +16.63 when everything the legs need was ready at +15.4. The step now builds `linq2db.Testing.slnf`, which already described the closure of `Tests.csproj` and only needed the EF.Core chain added.

Nothing stops being compiled on a PR: the `build` pipeline runs `Build Solution for Nuget` over all of `linq2db.slnx` on the same commit, and `Tests.LinqToDB.CLI` has the `test-cli` pipeline.

### Order the legs

Two separate problems with one symptom: the longest legs were starting last and finishing alone. Two rules decide the order, and which one is visible depends on whether the job caps its parallelism.

Without a cap, all of a job's legs are released at once and they take agents in case-insensitive alphabetical order of the key. Builds 22995 and 23009 had different declaration orders and both matched the alphabet 13/13 on their windows legs, and probe run 23039 matched 13/13 exactly. So the keys carry an `a_ .. y_` prefix in descending measured duration. 23024 and 23028 differ only in these prefixes: the test phase went 157.0 to 140.6 with leg work unchanged at 1153 against 1148 agent-minutes, so the whole gain is packing — average parallelism 7.34 to 8.17.

That orders legs inside a block but cannot get the block itself a slot. Both test jobs become eligible in the same second, and linux takes all ten within 0.3 min; windows then waited 44.4 min on 23028 and 54.8 on 23024. `maxParallel` on the linux job was tried against exactly that, and it worked as far as it went — on 23045 the first windows leg started at +0.1 instead of +44.4 — but it is not what this branch ships, for the reason two paragraphs below.

The cap is what brings the second rule into play, and 23050 is where it showed: `maxParallel` decides *which* legs are released next, and takes them in declaration order. The six linux legs it started were the first six ubuntu-enabled entries in the file — `s p i c l n` — followed by `m`, `t` and `j` as slots freed, again in file order, where the alphabet would have given `b c h i j k`. The file was grouped by provider family, which left ClickHouse, the second-longest leg in the run, last: it started 106 minutes in, and its 76.8 minutes became the tail on their own. 23050 finished in 202.3 min against 156.8 for the uncapped 23028.

So the file is sorted by key. Both rules then give the same order, and there is one thing to maintain instead of two. The first six ubuntu entries are now the six longest linux legs and the first four windows entries the four longest windows legs, checked against 23050's measured durations rather than assumed — no reletter was needed.

Sorting recovered most of that: 23059, the same cap over the sorted file, came in at 164.4 against 202.3. It did not recover all of it, and the cap is gone as a result — 23063, sorted and uncapped, ran **148.4 min**, 16 min under 23059 and 8.3 under the uncapped 23028.

The cap loses on the floor rather than the packing. Uncapped, both jobs draw on all ten slots and the floor is `total_work/10` — 115 min on 23059's 1151 agent-minutes. Capped at six, windows is confined to four slots and its 501 agent-minutes need 125 min there, so the floor is 10 min worse before a leg runs. Every build landed 20 to 25 min above its own floor, so the cap had no packing advantage to trade back. Windows waiting 44 min for its first slot reads badly on a timeline but does not cost wall clock — 23028 and 23063 both ran windows last and both finished ahead of the capped runs.

`linux_max_parallel` stays as a parameter, defaulting to 0, and the property is emitted only when it is above zero, so the default leaves it out rather than resting on how Azure reads `maxParallel: 0`. The matrix stays sorted: with no cap the file-order rule is dormant, but sorted is what the file should be either way, and a cap can then be switched on without reopening the trap.

### Create the baselines branch lazily, and retire create_baselines_pr

`create_baselines_pr` ran after every test leg and needed a fresh agent for 0.3 min of work, which it got only when the pool had a slot to spare: **78.1 min on 23009 and 77.5 on 22995**, in 23009's case behind two other `test-all` runs holding all ten slots. That single wait was larger than everything else here saves.

Its duties move to where an agent is already held. A leg clones baselines master rather than the run branch, and its push is what creates `baselines/pr_N`; whichever leg commits first wins the ref, and the rest are rejected, fetch it and rebase onto it. The winner also opens the baselines PR and comments on the source PR, with concurrent legs either finding it already open or losing the create, which GitHub answers with "A pull request already exists". A run that changes no baselines never creates a branch, so the empty-branch deletion the job existed for has no subject.

Verified on 23045, which carried a deliberate baseline change: one PR, one comment, and several legs going through push rejected, fetch, rebase, push, then "Baselines PR already exists" concurrently.

The same build also caught a regression of mine: nine windows legs failed on a push that had succeeded. Windows runs the task under PowerShell 5.1, where `2>&1` on a native command turns its stderr into error records that the task's default `$ErrorActionPreference` of Stop makes terminating, and `git push` writes its status to stderr even when it works. The stderr capture is not decoration — `gh` answers a lost create race on stderr — so the fix is `pwsh: true`, which the neighbouring tasks in these templates already use.

23050 confirmed that over the same path rather than by reasoning: the nine windows legs that had failed committed their baselines, and the branch diff came out as exactly the 58 `Any1` files the test change touches, all modified, none added or deleted. That test change has since been reverted — see **Before merge**.

### Tried and reverted

Capping the linux job with `maxParallel: 6` was carried for two builds and dropped. It fixed the symptom it was aimed at and still lost the run, for the floor reason given above; 23050 and 23059 are what it cost. `linux_max_parallel` survives it as a parameter defaulting to 0, so a pipeline whose work splits differently can switch a cap on with a measurement behind it.

Reordering the matrix entries appeared to do nothing: 23009 dispatched in the same order 22995 had on the old layout. That reading was correct for an uncapped job and wrong the moment one was capped, which cost 23050 about 45 minutes before the mechanism was found. The entries are sorted now, and `test-matrix.yml` carries both rules with the evidence for each, so the next person does not have to rediscover them.

Numbering the keys `01_ .. 25_` did nothing either; those names break the documented rule that a matrix configuration name must start with a letter, and the build that tested them also shared the pool, so it settled neither question. Reverted.

Hiding the prefixes behind a templated `displayName` is not possible: probe build 23041 rendered one as the literal `Win $(title) a_SqlServer2014` — the key is appended to the job name and a variable is not expanded there.

A stub pipeline mirroring the matrix with three-minute legs answered ordering questions in about a quarter of an hour against three, and is what settled the alphabetical rule on 23039 and the `displayName` limit on 23041. It is not in the branch: it was never registered in Azure DevOps, and it carried its own copy of the 28 keys with nothing to keep them in step with the real matrix.

### Before merge

One thing is left: linq2db.baselines#2080 must be **closed rather than merged**.

It exists because the commit `Rename the lambda parameters in AllAnyTests.Any1` moved 240 baseline files, which is what let the rewritten baselines path run with 28 legs pushing at once — something no YAML-only run could exercise. That commit is reverted, so the run now generates the same SQL as master and nothing will overwrite the 58 files still sitting on that branch.

### Not addressed

`Build Solution for Tests` is still serial at ~14 min after the filter; splitting it duplicates work across jobs for a much worse ratio.

Per-leg setup is the largest remaining pool of agent time after the tests themselves — 94 agent-minutes across `Setup tests` and `Execute Global Setup Script`, of which `Lin Oracle1819` alone is 18.5. That is Oracle creating its databases on first container start: 18c XE takes ~9.4 min and the community 19c EE image ~14.0, against ~0.8 for `free:23.2.0.0` and 3.3 or less for `express:21.3.0-xe` from the same registry. An image-version question, not a pipeline one.

Merging the two test jobs into a single 28-leg matrix is the largest remaining lever, and after this branch it is the only structural one left: the tail of 23063 is `Win f_SqlServer2016`, and windows does not get its first slot until t+48.7. One matrix means one queue, so the ten longest legs of either OS start together. Greedy packing over 23063's own durations puts that at 120 to 123 min of test phase against the 131.4 measured, though a real run would give some of it back to the pool latency that cost 23063 33.8 slot-minutes and that merging does not fix.

`test_cli_job` shows the mechanism works — `pool: vmImage: $(imageName)` takes the image from a matrix variable. The cost is that matrix variables are runtime, so OS branching cannot use `${{ if }}` and every OS-specific step needs a `condition:` instead — one steps list of about thirty steps, half of them skipped on each leg, which is the opposite of what this branch did with `arch`.


MaceWindu:
> [!IMPORTANT]
> **Agentic review — treat with care.** This was produced by an LLM agent rather than a human reviewer, so individual points can be wrong, overconfident, or miss context. Push back on anything that reads off.

Reviewed the branch and pushed two commits to it (`08e1efd37..147b00673`).

## `5dfc880a0` — replay only the leg's own commit when rebasing onto the baselines branch

`git rebase -X theirs FETCH_HEAD` replays everything back to the merge base, and the leg's clone is `--depth 1`, so its only commit object is the master tip and `.git/shallow` cuts the ancestry. While the branch is still based on that same master commit the merge base *is* the shallow root, exactly one commit replays, and `-X theirs` does what the comment says — that is the case build 23050 exercised.

If baselines master advances **after** `create_baselines_branch` rebased the branch — one unrelated baselines PR merging during a 2.5-hour run is enough — the shallow root is no longer reachable from the branch tip. Git reads it as an unrelated root commit, replays its whole tree onto the branch, and `-X theirs` lets that tree win: every baseline another leg or an earlier run had pushed is reset to master's content. The leg stays green.

Reproduced on git 2.55 against a scratch origin, with this step's exact command sequence and disjoint files:

```
run 1 captured    SqlServer.2014/Tests/Linq/Test1.sql
master advanced   SqlServer.2014/Tests/Linq/Test2.sql   (unrelated baselines merge)
this leg captured SQLite.Classic/Tests/Linq/Test1.sql

branch after the leg's push:
  [Linux / SQLite] baselines
  M2 (unrelated baselines merge, mid-run)   <- master's own commit, replayed
  [Windows / SQL Server 2014] baselines
  M1

SqlServer.2014/Tests/Linq/Test1.sql = 'MASTER-original ...'   <- run 1's capture lost
SQLite.Classic/Tests/Linq/Test1.sql = 'LEG-CAPTURE'
```

The resulting `master...branch` diff then omits the lost file and gains `SqlServer.2014/Tests/Linq/Test2.sql`, which is master's own change — so the baselines PR under-reports and over-reports at the same time, silently.

`--onto FETCH_HEAD HEAD~1` replays this leg's commit and nothing else. Re-running the harness over two consecutive rejection rounds keeps run 1's capture and both legs' captures and drops the phantom entry. `HEAD~1` is exact because the step makes exactly one commit, which the added comment records so a later edit adding a second one does not break the anchor silently.

Not a blocker on the design: it needs the mid-run master advance, the loss is a review artifact rather than product data, and a later run of the same PR re-captures the file. But nothing surfaces it, and the pre-PR flow could not hit it — the leg used to clone the run branch itself at full depth, so its merge base was always real.

## `147b00673` — ask GitHub whether the branch needs a rebase, instead of cloning to find out

`if ($branchHash -eq $masterHash)` compares the two head hashes, and a branch carrying even one baselines commit is by construction ahead of master — so the guard can never fire, and the `else` ran on every re-run of a baseline-carrying PR: a clone of the whole baselines repository (343541 files, ~420 Mb, no `--sparse`, no `--depth`) plus a rebase and an unconditional `git push -f`, even when master has not moved and the rebase is a no-op.

`create_baselines_branch` is what all three test jobs `dependsOn`, so this is serial prefix. The two builds that bracket it:

| build | `baselines/pr_5819` | `Create Test Baselines Branch` |
|---|---|---|
| 23050 | exists, 24 commits | 2.30 min |
| 23068 | pruned (#2080 closed the previous evening) | 0.25 min |

`GET /repos/linq2db/linq2db.baselines/compare/{master}...{branch}` answers the actual question in one request — `ahead` means there is nothing to rebase onto. Verified against the live API, including a head ref containing a slash. Dropping the master-head lookup left `Get-RemoteHash`'s non-`--heads` branch unreachable, so the switch went with it, and the head hash the final log line prints is now read from the rebased clone rather than assumed to equal master.

Riskier than the first commit, since it changes *when* the rebase runs and there is no live baselines branch to exercise it against — worth a look before the next baseline-carrying run.

## Checked, no change needed

- **The `test-matrix.yml` reorder is content-preserving.** 25 entries before and after, identical set, zero per-entry field differences — every `+ enable_os_windows: true` / `+ enable_fw_netfx: false` line in that diff is a realignment artifact.
- **"Nothing stops being compiled on a PR" holds.** All 23 projects dropped from `Build Solution for Tests` (the 15 `NuGet/*`, `LinqToDB.CLI`, `LinqToDB.Compat`, both `LinqToDB.LINQPad*`, `Tests.LinqToDB.CLI`, `Tests.Analyzers`, `linq2db.Benchmarks`, `Tests.T4`) build in `build`'s `Build Solution for Nuget` on the same commit; `Tests.LinqToDB.CLI` and `Tests.Analyzers` additionally still build in the `Azure` configuration via `test-cli.yml` and `Run Analyzer Tests`. `#if AZURE` / `#if BUGCHECK` code exists only under `Tests/Linq` and `Tests/Base`, both still in the filter, so no configuration-specific code loses compile coverage.
- **The `--sparse` premise is exact.** `BaselinesWriter` has one `File.ReadAllText`, reachable only for a path the same process already wrote; `BaselinesManager` reads nothing. The baselines root holds only `.gitattributes` (`* text diff eol=lf`) and an empty `.gitignore`, both inside the root-only cone, and it is the tree's only `.gitattributes` — so line-ending handling is unchanged. `git clone --sparse` + `git add --sparse -A` reproduced locally on git 2.55: stages an out-of-cone modification, a new out-of-cone file **and** a new out-of-cone directory, and stages no deletions for the untouched tree. Plain `git add -A` does exit 1 as described. The new-file and new-directory cases are the ones 23050 could not cover, since it only moved baselines that already existed.
- **The three commit scripts are byte-identical** across the windows / linux / macos templates apart from the `[Windows|Linux|MacOS]` label, before and after the change.
- **`build-job.yml`'s x86 margin figures are right** — 23050 gives 16.02 − 15.42 = 0.60 min = 36 s, 23028 gives 15.22 − 14.64 = 0.58 min. They are two builds that happen to agree, not one measurement counted twice.

## Two notes rather than findings

- **The x86 margin has flipped on the current head.** On 23068, `Build (win-x86)` ended at +16.51 min against `Build` at +16.47 — the x86 job finished 2.4 s *later*, not ahead. Absorbed by `create_baselines_branch` sitting after `build_job`, and the comment already flags exactly this risk; raising it only because the 36 s / 0.6 min / 1.9 min figures in it no longer describe the current run.
- **`test-matrix.yml`'s ClickHouse start offset.** The comment says it started "90 minutes in" on 23050. The timeline says t+125.5 from the first job, or +106.6 from the first test leg — the latter matching the 106 in the PR description. Left alone: the conclusion (it started very late and defined the tail alone, 76.8 min, ending exactly at the build's 202.3) is unaffected, and every neighbouring claim checks out exactly — the six legs the cap released really were `{s p i c l n}` (p, c, n, l, s, i all start at t+18.9) and `m`, `t`, `j` really did follow at +88.5 / +88.6 / +90.7.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `147b006`.</sub>

PR #5821, opened 2026-08-26
Fix the .claude corpus bootstrap: detached HEAD, swallowed errors, fresh-clone claim

MaceWindu:
Follow-up to #5820, from reviewing it after merge. Three fixes to `.githooks/_refresh-corpus.sh` plus one falsified sentence in `CONTRIBUTING.md`.

## The corpus was bootstrapped onto a detached HEAD

`git submodule update --init` checks out the recorded gitlink detached, and the `--remote --merge` refresh below merges into *that detached HEAD* rather than attaching to a branch. So every bootstrapped worktree got a permanently detached corpus, which breaks the write half of the workflow the bootstrap exists to enable — `CONTRIBUTING.md` defines corpus editing as `git -C .claude commit` + `git -C .claude push`:

```
git -C <worktree>/.claude symbolic-ref -q HEAD   ->  (empty)
git -C <worktree>/.claude push origin
  fatal: You are not currently on a branch.
```

The silent half is worse than the failed push: a commit made while detached is *accepted*, and git mentions the loss only in passing when you later attach — `Warning: you are leaving 1 commit behind, not connected to any of your branches`.

`CONTRIBUTING.md` already prescribed the missing step for the manual path, with the reason spelled out — *"then `git -C .claude switch master` (init checks out a detached HEAD, and a corpus commit made there would go nowhere useful)"* — and the manual worktree recipe #5820 replaced used `git clone … --branch master`, i.e. deliberately on a branch. The automation dropped the step the documentation it automates calls out.

`refs/heads/master` already exists locally at that point (the clone creates it), so the added `switch` is a no-op fast-forward and `--remote --merge` then advances the branch instead of a detached HEAD.

## The real git error was discarded on failure

`>/dev/null 2>&1` threw away git's message and the hint attributed every failure to "offline or clone failed". Two conditions that misreports, both hit while exercising the script: a blocked transport, and a leftover non-empty `.claude/` directory — `fatal: destination path … already exists and is not an empty directory`, a user-fixable state the message actively misdirects away from, and one where the suggested manual command fails identically.

The exit status of an assignment whose value is a command substitution is the command's status in POSIX sh, so the `if !` guard is unchanged. Checked under `sh`.

## "fresh clone" was not covered

`core.hooksPath .githooks` is per-clone local config that `CONTRIBUTING.md` tells you to set *after* cloning — it is not repo-shared and not global. During a `git clone` the new repository therefore has no `hooksPath`, git looks for `.git/hooks/post-checkout`, and this script is never invoked. What the change does cover is an already-configured clone whose submodule was never initialised, at its next branch checkout. Comment corrected; the claim also appears in #5820's own body and commit message, which are frozen.

## `CONTRIBUTING.md`

The sentence describing `post-checkout` / `post-merge` / `post-rewrite` as *fast-forwarding* `.claude/` predates #5820's bootstrap path. The "Already cloned?" instructions above it are deliberately left alone — they remain correct for a clone whose hooks are not enabled yet, which is exactly the case above.

## Verification

`sh -n` clean. Everything below is measured through a real `git worktree add` running the fixed hook against this repository, not a fixture:

| | before | after |
|---|---|---|
| `.claude/` populated | yes | yes |
| `symbolic-ref HEAD` | *(empty — detached)* | `refs/heads/master` |
| upstream | — | `origin/master` |
| `push --dry-run origin` | `fatal: You are not currently on a branch.` | `Everything up-to-date` |
| primary clone's `.claude` | untouched | untouched |
| failure path output | `bootstrap skipped (offline or clone failed)` | git's actual `fatal:` message, then the manual hint |

Two notes for reviewers, both found while testing and neither a change here:

- **The hook that runs during `git worktree add` is the *primary clone's* copy.** `core.hooksPath` is relative and resolves against the invoking clone, so a primary clone sitting on a branch that predates #5820 still runs the old no-op script and the new worktree comes up empty — silently, since the old early return printed nothing. This cost me a wrong conclusion mid-review before I spotted it.
- **`git submodule update --init` inside a linked worktree is safe on git 2.55**, contrary to the guidance in the agents corpus that #5820 implicitly overrode. Measured against a purpose-built superproject: the linked worktree gets its own git dir at `.git/worktrees/<id>/modules/.claude`, the shared `.git/modules/.claude/config` `core.worktree` is byte-identical before and after, and the primary clone's submodule stays on its branch with a clean status. The corpus guidance has been corrected separately. The git version at which the per-worktree `modules/` layout arrived was not established, so an older git may still behave as the old warning described.


sdanyliv (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `ba02965`.</sub>

PR #5831, opened 2026-08-29
Fix #5792: keep set-operation branch in nullability scope

### LINKED ISSUE #5792: Entity query filter causes a left-join null-check to be dropped in the second operand of a set operation
## Summary

When an entity has a `HasQueryFilter` and is projected and `LeftJoin`-ed, a `joined != null ? joined.X : fallback`
conditional is discarded in every operand of a `Concat`/`Union` **except the first**. Only the `.X` branch survives,
so a non-matching left join yields `NULL` instead of the fallback.

Remove the query filter and the same query is translated correctly. It is positional — swapping the operands moves
the breakage to the other side.

**Regression:** introduced in **6.4.0**. Correct in 6.0.0, 6.1.0, 6.2.0, 6.2.1 and 6.3.0.

## Repro

```csharp
[Table("Orders")]
public class Order
{
    [Column] public string CustomerCode { get; set; } = null!;
}

[Table("Customers")]
public class Customer
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string Code { get; set; } = null!;
    [Column] public string Name { get; set; } = null!;
}
```

```csharp
var ms = new MappingSchema();
var b = new FluentMappingBuilder(ms);
var off = false;                       // not a compile-time constant, so the filter is not folded away
b.Entity<Customer>().HasQueryFilter((e, ctx) => off || e.Name != "");
b.Build();

using var db = new DataConnection(new DataOptions().UseSQLite(cs).UseMappingSchema(ms));
db.CreateTable<Order>();
db.CreateTable<Customer>();
db.Insert(new Customer { Code = "C1", Name = "Acme" });
db.Insert(new Order { Id = 1, CustomerCode = "C1" });        // resolves
db.Insert(new Order { Id = 2, CustomerCode = "MISSING" });   // does not

IQueryable<string> Names(int id) =>
    from o in db.GetTable<Order>().Where(o => o.Id == id)
    from c in db.GetTable<Customer>()
        .Select(x => new { x.Code, Name = x.Name + "!" })
        .LeftJoin(x => x.Code == o.CustomerCode)
    select c != null ? c.Name : "Unknown";

Names(1).Concat(Names(2)).ToList();
```

Three ingredients are each necessary — drop any one and it does not reproduce:

- the `HasQueryFilter` on the joined entity;
- a filter predicate that is not a compile-time constant (`(e, ctx) => true` is elided);
- a **computed** projected column (`x.Name + "!"`); with a plain column the join is flattened.

## Expected

`["Acme!", "Unknown"]`, and both operands keep the `CASE` (this is 6.3.0, and 6.4.0 with the filter removed):

```sql
SELECT CASE WHEN [c_2].[Code] IS NOT NULL THEN [c_2].[Name] ELSE 'Unknown' END
FROM [Orders] [o]
    LEFT JOIN ( SELECT [c_1].[Code], Coalesce([c_1].[Name], '') || '!' as [Name] FROM [Customers] [c_1]
                WHERE [c_1].[Name] <> '' OR [c_1].[Name] IS NULL ) [c_2] ON [c_2].[Code] = [o].[CustomerCode]
WHERE [o].[Id] = @id
UNION ALL
SELECT CASE WHEN [c_4].[Code] IS NOT NULL THEN [c_4].[Name] ELSE 'Unknown' END
FROM [Orders] [o_1]
    LEFT JOIN ( ... ) [c_4] ON [c_4].[Code] = [o_1].[CustomerCode]
WHERE [o_1].[Id] = @id_1
```

## Actual (6.4.0)

`["Acme!", NULL]` — the second operand's projection is reduced to the bare column:

```sql
UNION ALL
SELECT
	[c_4].[Name]          -- CASE ... ELSE 'Unknown' dropped
FROM [Orders] [o_1]
    LEFT JOIN ( ... ) [c_4] ON [c_4].[Code] = [o_1].[CustomerCode]
WHERE [o_1].[Id] = @id_1
```

## Notes

- With a longer fallback chain (`a != null ? a.X : b != null ? b.X : "Unknown"`) the whole chain collapses to
  `a.X`; `b` stays in the `FROM` clause but is never referenced in the `SELECT`.
- The result is a `NULL` in a column the projection type declares non-nullable, so it surfaces as bad data rather
  than an exception.

### Reproduction Steps

Full sample:

```csharp
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.Mapping;

namespace Repro;

[Table("Orders")]
public class Order
{
    [Column] public string CustomerCode { get; set; } = null!;
}

[Table("Customers")]
public class Customer
{
    [Column(CanBeNull = false, IsPrimaryKey = true)] public string Code { get; set; } = null!;
    [Column] public string Name { get; set; } = null!;
}

public static class Program
{
    public static void Main(string[] args)
    {
        var withFilter = args.Length == 0 || args[0] != "nofilter";

        var cs = "Data Source=min1;Mode=Memory;Cache=Shared";
        using var keep = new Microsoft.Data.Sqlite.SqliteConnection(cs);
        keep.Open();

        var ms = new MappingSchema();
        if (withFilter)
        {
            var b = new FluentMappingBuilder(ms);
            var off = false;   // not a compile-time constant, so the filter is not folded away
            b.Entity<Customer>().HasQueryFilter((e, ctx) => off || e.Name != "");
            b.Build();
        }

        using var db = new DataConnection(new DataOptions().UseSQLite(cs).UseMappingSchema(ms));
        db.CreateTable<Order>();
        db.CreateTable<Customer>();
        db.Insert(new Customer { Code = "C1", Name = "Acme" });
        db.Insert(new Order { Id = 1, CustomerCode = "C1" });        // resolves
        db.Insert(new Order { Id = 2, CustomerCode = "MISSING" });   // does not

        IQueryable<string> Names(int id) =>
            from o in db.GetTable<Order>().Where(o => o.Id == id)
            from c in db.GetTable<Customer>()
                .Select(x => new { x.Code, Name = x.Name + "!" })
                .LeftJoin(x => x.Code == o.CustomerCode)
            select c != null ? c.Name : "Unknown";

        Console.WriteLine($"linq2db {typeof(DataConnection).Assembly.GetName().Version} filter={withFilter}");
        var q = Names(1).Concat(Names(2));
        Console.WriteLine("  got: " + string.Join(", ", q.ToList().Select(x => x ?? "NULL")));
        Console.WriteLine("  expected: Acme!, Unknown");
        Console.WriteLine(q.ToSqlQuery().Sql.Trim());
    }
}

```

### Exception details

_No response_

### Regression?

It worked in 6.1.0, which we are now stuck at

### Linq To DB version

6.4.0

### .NET version

10.0.0

### Database

SQL Server 2019

### Database provider

Microsoft.Data.Sqlite 9.0.5

### Operating system

macOS and Ubuntu CI machines

### Configuration

_No response_

sdanyliv:
Fixes #5792

## Problem

With a query filter on a left-joined entity, a `joined != null ? joined.X : fallback` conditional was dropped in every set-operation branch except the first, so a non-matching left join yielded `NULL` instead of the fallback. Regression in 6.4.0.

## Cause

`SqlExpressionConvertVisitor.VisitSqlQuery` descends into a query and optimizes its column expressions, but only set `JoinSource` — it never added the query to `NullabilityContext.Queries`.

Nullability was therefore resolved against an outer scope. `IsNullableSourceCalculator` walks `From`/`Joins`, which never enters `SetOperators`, so a source inside a set-operation branch came back `null` (unreachable). `CanBeNull` contributes nothing for `null`, and since the joined key is a non-nullable PK the join was the only source of nullability — so `IS NOT NULL` folded to `true` and `VisitSqlConditionExpression` reduced the `IIF` to its true-branch.

This is positional for a structural reason: the first branch *is* the outer query, so its tables are reachable from `From`/`Joins`; the second branch is reachable only through `SetOperators`.

`SqlExpressionOptimizerVisitor.VisitSqlQuery` already registers the query via `WithQuery`. The two visitors performed the same descent under different contracts; this restores the symmetry.

## Fix

```csharp
NullabilityContext = NullabilityContext.WithQuery(selectQuery).WithJoinSource(selectQuery);
```

## Verification

Local full runs, paired (same command, same binary path, DLL freshness asserted in-run):

| provider | without fix | with fix |
|---|---|---|
| SQLite.MS | 10927 total, 1 failed — exactly the new test | 10927 total, 0 failed |
| SQLite.Classic | — | 10938 total, 0 failed |
| SqlServer.2019.MS | 11749 total, 10 failed | 11749 total, same 10 |

Exactly one test changed state on SQLite.MS; no other test moved. The 10 SQL Server failures are pre-existing in the no-fix baseline — `CURRENT_TIMESTAMP`/`GETDATE` off by exactly 180 minutes, a container-vs-host timezone offset, unrelated to this change.

## Follow-up commits

- `eb31446b79` — the set-operation assertion is order-insensitive. `Concat` renders as UNION ALL with no ORDER BY, so branch order is not guaranteed. Each branch yields one row and the two values are distinct, so the test keeps its full discriminating power.
- `df41dd6001` — coverage for two facets of #5792 the first test did not pin: a three-operand `Concat`, so the null check is asserted in every branch after the first rather than only the second, and `FilteredLeftJoinFallbackChainSurvivesSetOperation` for the issue's second manifestation, where a two-level fallback collapsed to the first join's column and left the second join in `FROM` but unreferenced in the projection.
- `50798d58a5` — the same root cause at the consumer side. `NullabilityContext.CanBeNull` resolved a set-operation column by evaluating each branch's column with the current context, which does not contain the branch query. Since `IsNullableSourceCalculator` never enters `SetOperators`, the branch's own tables came back unreachable and a value nullable only through a join inside the branch was reported as not nullable to the enclosing query — the null check folded away and the NULL surfaced as a default value. Registering the branch via `WithQuery` fixes it. The visited-column cycle guard is carried over explicitly because `WithQuery` allocates a fresh context.

### Verification of the follow-ups

`QueryFilterTests` 42/42 on SQLite.MS. Full SQLite.MS run: 10929 total, 7 failed — all on the `SQLite.MS.LinqService` context and all flaky rather than caused by the change: a run with `50798d58a5` reverted failed a **disjoint** set of 3 (`Distinct6` ×3), and all 10 pass when re-run. Both tests added in `df41dd6001` are confirmed red at `8a80301ee5` and green after. The test added in `50798d58a5` is confirmed red at `8a80301ee5`, where the emitted SQL drops the null check entirely and the left join's NULL materialises as `0`.

`50798d58a5` touches core `NullabilityContext`, so its blast radius is wider than the original one-line scope fix — the CI matrix and the baselines diff it generates are the meaningful check.

## Notes

- The regression test is SQLite-only; the defect is provider-agnostic by trace (the fold happens in the shared convert/optimize pipeline, no provider-specific code on the path), not verified on other providers.
- SQL baselines are not evaluated locally here — CI is the check for generated-SQL shifts elsewhere.

### DIFF
--- a/Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
+++ b/Source/LinqToDB/Internal/SqlProvider/SqlExpressionConvertVisitor.cs
@@ -184,7 +184,7 @@ namespace LinqToDB.Internal.SqlProvider
 				return selectQuery;
 
 			var saveNullabilityContext = NullabilityContext;
-			NullabilityContext = NullabilityContext.WithJoinSource(selectQuery);
+			NullabilityContext = NullabilityContext.WithQuery(selectQuery).WithJoinSource(selectQuery);
 
 			var newQuery = base.VisitSqlQuery(selectQuery);
 
--- a/Source/LinqToDB/Internal/SqlQuery/NullabilityContext.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/NullabilityContext.cs
@@ -156,7 +156,17 @@ namespace LinqToDB.Internal.SqlQuery
 							if (index >= set.SelectQuery.Select.Columns.Count)
 								return true;
 
-							if (set.SelectQuery.Select.Columns[index].CanBeNullable(this))
+							// the branch query is reachable only through SetOperators, which the source walk does not enter.
+							// Without registering it the branch's own tables resolve as unreachable, so a value that is nullable
+							// only through a join inside the branch is reported as not nullable to the enclosing query.
+							var setNullability = WithQuery(set.SelectQuery);
+
+							// carry the cycle guard over - WithQuery allocates a fresh context, which would otherwise start
+							// with an empty visited set and lose the protection against self-referencing column chains
+							if (!ReferenceEquals(setNullability, this))
+								setNullability._visitedColumns = _visitedColumns;
+
+							if (set.SelectQuery.Select.Columns[index].CanBeNullable(setNullability))
 								return true;
 						}
 					}
--- a/Source/LinqToDB/Internal/SqlQuery/NullabilityContext.cs
+++ b/Source/LinqToDB/Internal/SqlQuery/NullabilityContext.cs
@@ -153,6 +153,11 @@ namespace LinqToDB.Internal.SqlQuery
 
 						foreach (var set in column.Parent.SetOperators)
 						{
+							// EXCEPT / INTERSECT narrow the accumulated left side and contribute no values of their
+							// own, so a branch that is nullable only in itself cannot put a NULL into this column.
+							if (set.Operation is not (SetOperation.Union or SetOperation.UnionAll))
+								continue;
+
 							if (index >= set.SelectQuery.Select.Columns.Count)
 								return true;
 
--- a/Tests/Linq/Linq/QueryFilterTests.cs
+++ b/Tests/Linq/Linq/QueryFilterTests.cs
@@ -844,6 +844,43 @@ namespace Tests.Linq
 		}
 
 		[Test]
+		public void FilteredLeftJoinNullCheckSurvivesSetOperation([IncludeDataSources(false, TestProvName.AllSQLite)] string context)
+		{
+			// Regression: with a query filter on the joined entity, the `joined != null ? joined.X : fallback`
+			// conditional was discarded in every set-operation branch except the first, so a non-matching left
+			// join produced NULL instead of the fallback. Two ingredients keep the shape intact: the filter must
+			// not fold to a constant, and the projected column must be computed — with a plain column the join
+			// is flattened and the null check never has to survive on its own.
+
+			var builder = new FluentMappingBuilder(new MappingSchema());
+
+			builder.Entity<MasterClass>().HasQueryFilter<MyDataContext>((e, dc) => !dc.IsSoftDeleteFilterEnabled || !e.IsDeleted);
+
+			builder.Build();
+
+			var masters = new[] { new MasterClass { Id = 1, Value = "Master" } };
+			var details = new[]
+			{
+				new DetailClass { Id = 1, MasterId = 1   }, // resolves
+				new DetailClass { Id = 2, MasterId = 100 }  // does not
+			};
+
+			using var db = new MyDataContext(context, builder.MappingSchema);
+			using (db.CreateLocalTable(masters))
+			using (db.CreateLocalTable(details))
+			{
+				IQueryable<string> Values(int id) =>
+					from d in db.GetTable<DetailClass>().Where(d => d.Id == id)
+					from m in db.GetTable<MasterClass>()
+						.Select(x => new { x.Id, Value = x.Value + "!" })
+						.LeftJoin(x => x.Id == d.MasterId)
+					select m != null ? m.Value : "Unknown";
+
+				Values(1).Concat(Values(2)).ToList().ShouldBe(["Master!", "Unknown"]);
+			}
+		}
+
+		[Test]
 		public void NamedFilter_InterfaceTypedLambda_FastPath([IncludeDataSources(false, TestProvName.AllSQLite)] string context)
 		{
 			// Regression: a filter lambda typed against an interface (Func<ISoftDelete, ...>) applied to a concrete
--- a/Tests/Linq/Linq/QueryFilterTests.cs
+++ b/Tests/Linq/Linq/QueryFilterTests.cs
@@ -927,10 +927,10 @@ namespace Tests.Linq
 		public void SetOperationBranchNullabilityReachesEnclosingQuery([IncludeDataSources(false, TestProvName.AllSQLite)] string context)
 		{
 			// The enclosing query resolves a set-operation column by fanning out over SetOperators. Branch 1 is
-			// deliberately non-nullable so that fan-out is actually reached, and branch 2 is nullable only through
-			// a LEFT JOIN living inside the branch - which the source walk cannot see unless the branch query is
-			// registered in the nullability scope. Without that the null check folds away and the NULL surfaces
-			// as a default value.
+			// deliberately non-nullable so that the fan-out is the only thing that can report the column nullable,
+			// and branch 2 is nullable only through a LEFT JOIN living inside the branch - which the source walk
+			// cannot see unless the branch query is registered in the nullability scope. Without that the null
+			// check folds away and the NULL surfaces as a default value.
 
 			var builder = new FluentMappingBuilder(new MappingSchema());
 
--- a/Tests/Linq/Linq/QueryFilterTests.cs
+++ b/Tests/Linq/Linq/QueryFilterTests.cs
@@ -876,7 +876,7 @@ namespace Tests.Linq
 						.LeftJoin(x => x.Id == d.MasterId)
 					select m != null ? m.Value : "Unknown";
 
-				Values(1).Concat(Values(2)).ToList().ShouldBe(["Master!", "Unknown"]);
+				Values(1).Concat(Values(2)).ToList().ShouldBe(["Master!", "Unknown"], ignoreOrder: true);
 			}
 		}
 
--- a/Tests/Linq/Linq/QueryFilterTests.cs
+++ b/Tests/Linq/Linq/QueryFilterTests.cs
@@ -968,6 +968,51 @@ namespace Tests.Linq
 		}
 
 		[Test]
+		public void SetOperationNullabilityIgnoresNarrowingBranches([IncludeDataSources(false, TestProvName.AllSQLite)] string context)
+		{
+			// EXCEPT / INTERSECT narrow the accumulated left side and contribute no values of their own, so a
+			// branch that is nullable only through a join inside itself must not make the enclosing column
+			// nullable. The UNION ALL case shares the shape and must keep the check - that is the #5792 fix.
+
+			var builder = new FluentMappingBuilder(new MappingSchema());
+
+			builder.Entity<MasterClass>().HasQueryFilter<MyDataContext>((e, dc) => !dc.IsSoftDeleteFilterEnabled || !e.IsDeleted);
+
+			builder.Build();
+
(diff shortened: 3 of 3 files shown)


MaceWindu (approved):
*This review was produced with AI assistance.*

Reviewed `50798d58a`. Two follow-up commits pushed onto this branch (`42f4a33f4`, `a25d6f044`) and `test-all` re-triggered — details below.

## Review notes

- The scope fix is sound and the mechanism in the description matches the code. `IsNullableSourceCalculator` pushes only `query.From.Tables`, never `SetOperators`, so a branch's own tables were genuinely unreachable from the enclosing scope; registering the branch query is the right correction. The `_visitedColumns` carry-over is load-bearing rather than defensive — without it each branch context starts with an empty set and a chain re-entering the same fan-out would recurse instead of being caught.
- `test-all` was green on `50798d58a` (build 23113, triggered after the head commit, so it covered the reviewed code). The new run covers the two follow-up commits.

## Follow-up commit `42f4a33f4` — test comment correction

The comment on `SetOperationBranchNullabilityReachesEnclosingQuery` said branch 1 is non-nullable *"so that fan-out is actually reached"*. The fan-out runs unconditionally once `column.Parent.HasSetOperators` is true, and the head branch's own expression is evaluated only afterwards — nothing about branch 1 gates it. What branch 1's non-nullability actually buys is discrimination: were it nullable, the head-expression check would report the column nullable on master too and the test would pass without the fix. Reworded to say that.

## Follow-up commit `a25d6f044` — EXCEPT / INTERSECT branches skipped in the fan-out

`CanBeNull` iterates `column.Parent.SetOperators` without looking at `SqlSetOperator.Operation`, so an `EXCEPT` / `INTERSECT` branch can report the enclosing column nullable. Those operations narrow the accumulated left side and contribute no values of their own, so a branch nullable purely through a join inside itself cannot put a NULL into the column.

Latent before this PR — such a branch resolved as unreachable and answered `false` — and reachable after it. Measured on SQLite with one query shape across all four operations, where the head branch is non-nullable and the second is nullable only via a `LEFT JOIN` inside it:

| operation | master | this PR | correct |
|---|---|---|---|
| `UNION ALL` | no null check | `CASE WHEN … IS NOT NULL` | this PR |
| `UNION` | no null check | `CASE WHEN … IS NOT NULL` | this PR |
| `EXCEPT` | no null check | `CASE WHEN … IS NOT NULL` | master |
| `INTERSECT` | no null check | `CASE WHEN … IS NOT NULL` | master |

The guard restores master's output for the bottom two while leaving the top two as this PR makes them, so it should move no baselines. `SetOperationNullabilityIgnoresNarrowingBranches` pins all three shapes through `LastQuery`. Local verification is SQLite-only: `QueryFilterTests` 43/43, plus 453/453 across `ConcatUnionTests`, `SetOperator*`, `ExceptByMethodTests`, `UnionByMethodTests`, `IndexOutOfRangeInUnions` and `EagerLoadingStrategyUnionTests`. The matrix is the real check.

## Note — the convert-visitor registration is not set-operation-specific

`NullabilityContext.WithQuery(selectQuery)` in `SqlExpressionConvertVisitor.VisitSqlQuery` changes nullability resolution for every query the convert visitor descends into, not only set-operation branches. Pre-PR the root query was the only entry in `Queries` for that whole descent (`BasicSqlBuilder` seeds it, and `VisitSqlQuery` reset `JoinSource` alone), so a source inside a subquery nested two or more levels under an outer `LEFT JOIN` now resolves non-nullable where it previously inherited the outer join's nullability — a tightening in the direction that removes null handling.

No evidence it changes anything. A constructed nested-subquery-under-`LEFT JOIN` shape emits byte-identical SQL before and after, because `SqlExpressionOptimizerVisitor.VisitSqlQuery` has always carried `WithQuery` and folds it first; zero existing baselines moved across the provider matrix. So this looks inert, and it also closes a real divergence — `SqlExpressionConvertVisitor.Optimize` passes its own context into the optimizer visitor, so an expression optimized *through* the convert visitor previously got a less precise scope than the same expression optimized standalone. Raised only because one constructed shape is not the whole input domain, and the description frames the change as set-operation-specific.

## Baselines

12 added, 53 modified, 0 deleted.

- **Added** — this PR's three new tests across the four SQLite configs. All three confirm the fix: every branch of the three-operand `UNION ALL` keeps its `CASE WHEN … IS NOT NULL`, and the two-level fallback keeps both `WHEN` arms.
- **Modified** — none attributable to this change. Every one is `TestDataTypes` losing `ORDER BY ID`, or the DB2 `BulkCopy*` cleanup shape. Both trace to baselines master having been written by two recently-merged baselines PRs whose branches predate #5784; this branch contains #5784, so its run emits the current SQL. Merging this PR's baselines PR corrects that drift rather than introducing any.

The useful number: **a core `NullabilityContext` change that moved zero existing baselines across the whole provider matrix.**

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A `LEFT JOIN` fallback inside a set-operation branch lost its null check, so it returned **wrong values** with no error. Every branch after the first was affected:

  ```csharp
  // the joined entity must carry a query filter whose predicate is not a compile-time constant
  builder.Entity<Master>().HasQueryFilter<MyDataContext>((q, dc) => q.Where(m => m.TenantId == dc.TenantId));

  IQueryable<string> Values(int id) =>
      from d in db.Details.Where(d => d.Id == id)
      from m in db.Masters.Select(x => new { x.Id, Value = x.Value + "!" }).LeftJoin(x => x.Id == d.MasterId)
      select m != null ? m.Value : "Unknown";

  Values(1).Concat(Values(2)).Concat(Values(3));   // "Unknown" came back as null
  ```

  A non-matching join produced `NULL` instead of the fallback, and where the fallback was a value type it materialised as the default (`0` for `int`). A two-level fallback chain collapsed to the first join's column entirely. Three things had to coincide: a query filter on the joined entity, a filter predicate that is not a compile-time constant, and a computed projected column. Regression in 6.4.0. Related: `EXCEPT` and `INTERSECT` branches no longer contribute nullability, so generated SQL for those operators may carry fewer null checks than before.


A `LEFT JOIN` fallback inside a set-operation branch lost its null check and returned wrong values (regression in 6.4.0).


---
<sub>Generated from commit `a25d6f0`.</sub>

PR #5832, opened 2026-08-29
Document how to configure provider-specific options in EF Core integration

### LINKED ISSUE #5830: Documentation Improvement: Improve Docs around Setting EF Core Provider specific options
### Background:

While Trying to Set up Oracle Bulk Copy via EFCore shim to use Array Insert (i.e. `AlternativeBulkCopy.InsertInto`) I ran into a small snag getting things set up;

I tried to use [the readme.md documentation](https://github.com/linq2db/linq2db/tree/master/Source/LinqToDB.EntityFrameworkCore#how-to-use) to set things up, i.e. `.builder.AddCustomOptions(o => o.UseOracle(/*etc*/));` but when we tried to test, we wound up getting an exception stating 'Connection String Not Found' from the options builder.

We were able to Fix it via changing the call to be `builder.AddCustomOptions(o => o.WithOptions<OracleOptions>()` instead; Github Copilot using Opus 5 reported that the `.UseOracle` resulted in any configuration mapped from the EFCore builder options was being cleared out upon calls to `.UseOracle`. (Did not get a chance to check if other providers would have the same behavior or not.) Using `.WithOptions` was confirmed as successful in providing the options to the created data provider.

### Proposed Solution:

It would probably be good to update the https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB.EntityFrameworkCore/README.md file in that section, to give more correct instructions for how to set up DB specific options without clearing out the already-provided Connection string (and/or other things such as mappings)

Alternatively, if this is in fact an oracle-specific problem, it would be nice to align the behavior with other providers wire-ups.

### New API

_No response_

sdanyliv:
Fixes #5830

## Problem

`AddCustomOptions` hands the callback an **empty** `DataOptions`: the EF Core integration resolves the connection string, connection and data provider later, when a linq2db context is created for the `DbContext`.

Two consequences were undocumented:

* `UseXxx` overloads that leave the dialect at `AutoDetect` — including short forms like `UseOracle(optionSetter)` — need to read the server version, so they fail with `Connection string is not provided.` (`ProviderDetectorBase.CreateOptions`).
* Even when a `UseXxx` call succeeds, it pins a concrete `IDataProvider` on the options, and `LinqToDBForEFToolsImplDefault.GetDataProvider` returns that instance before any EF-based detection runs.

The README's "How to use" example steered users straight into this: it registered `UseSqlite()` and then configured `UseSqlServer(...)`.

## Change

README only, no code changes.

* Made the example self-consistent (a single Oracle scenario) and switched it to `WithOptions<OracleOptions>`, which is the shape that actually works for provider options.
* Added a "Configuring provider-specific options" subsection covering the three cases: `WithOptions<TOptions>` for provider behavior, `UseXxx` for deliberately overriding provider/dialect detection, and why the `AutoDetect` overloads throw here.

### DIFF
--- a/Source/LinqToDB.EntityFrameworkCore/README.md
+++ b/Source/LinqToDB.EntityFrameworkCore/README.md
@@ -41,17 +41,52 @@ example:
 
 ```cs
 var optionsBuilder = new DbContextOptionsBuilder<MyDbContext>();
-optionsBuilder.UseSqlite();
+optionsBuilder.UseOracle(connectionString);
 optionsBuilder.UseLinqToDB(builder =>
 {
     // add custom command interceptor
     builder.AddInterceptor(new MyCommandInterceptor());
     // add additional mappings
     builder.AddMappingSchema(myCustomMappings);
-    // configure SQL Server dialect explicitly
-    builder.AddCustomOptions(o => o.UseSqlServer(SqlServerVersion.v2022));
+    // configure provider-specific options
+    builder.AddCustomOptions(o => o.WithOptions<OracleOptions>(oo => oo with
+    {
+        AlternativeBulkCopy = AlternativeBulkCopy.InsertInto
+    }));
 });
 ```
+
+### Configuring provider-specific options
+
+`AddCustomOptions` receives an **empty** `DataOptions` instance. Connection string, connection and data provider are
+not part of it: they are taken from the EF Core context later, when a `LINQ To DB` context is created for it. This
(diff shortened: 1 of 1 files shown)


MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `43b97f9`.</sub>

PR #5680, opened 2026-07-06
Fix #5457: recursive CTE drops columns with object-typed projection member

sdanyliv:
Fixes #5457.

A recursive CTE (`db.GetCte<T>(anchor.Concat(recursive))`) whose projection type has an `object`-typed member silently dropped CTE columns — the header collapsed to one column and references failed (`Invalid column name` on SQL Server / `no such column` on SQLite). A `string`-typed member worked; the `object` member was the trigger.

## Root cause
The merged set-operation projection is a conditional (`set-id == left ? anchorProjection : recursiveProjection`). `BuildProxyBase.ProcessTranslated` had no `ConditionalExpression` case, so it fell through to the generic visitor, which dropped the per-member tracking path. The recursive-CTE fields then reached `CteContext.RegisterField` with a null path, never resolved (their `SqlCteField.Column` stayed null), and the column-usage optimizer pruned them.

## Fix
`BuildProxyBase.ProcessTranslated` now handles the projection-merge `ConditionalExpression`, propagating the member path into both branches so union-merged members keep their tracking and resolve against the CTE fields. Guarded to constructor branches so scalar / eager-load conditionals are unaffected.

## Tests
New `CteTests.Issue5457` (recursive CTE, object-typed member). Verified: full SQLite suite, SQL Server + PostgreSQL CTE/set-operation suites.

PR #5833, opened 2026-08-29
Fix set operation branches projecting differing-but-assignable types

### LINKED ISSUE #5683: Recursive CTE drops columns when recursive term projects a differing-but-assignable type
### Summary

A recursive CTE built with `GetCte<T>(anchor.Concat(recursive))` drops columns when the
recursive term's projection constructs a different-but-assignable type (a polymorphic /
base-cast projection) than the anchor term. The dropped columns are the ones used only in
the recursive self-join and the outer join, so the emitted SQL is malformed (SQLite:
`circular reference: partHierarchy`; other providers: invalid/missing column). This is the
same class of column-tracking loss as #5457, for a query shape that #5680's fix does not cover.

### Background

Set-operation projection merging represents the two operand projections as a conditional
`test ? proj1 : proj2`. When the two projections have differing-but-assignable static types,
`SetOperationBuilder.MakeCondition` wraps one branch in `Expression.Convert` to unify the type.

### Current behavior

- `Source/LinqToDB/Internal/Linq/Builder/SetOperationBuilder.cs:244` — `MakeCondition`
  Convert-wraps a branch when `ifTrue.Type != ifFalse.Type`, yielding
  `test ? BaseCtor : Convert(DerivedCtor)`.
- `Source/LinqToDB/Internal/Linq/Builder/BuildProxyBase{TOwner}.cs:238` (master) —
  `ProcessTranslated` has no `ConditionalExpression` case, so the merged conditional falls
  through to the generic `BuildProxyVisitor`, losing the per-member tracking path; the
  recursive-CTE fields register with a null path and the unresolved columns are pruned.
  (PR #5680 adds a `ConditionalExpression` case — on its branch, unmerged — but it matches
  only when *both* branches are bare `SqlGenericConstructorExpression`, so a Convert-wrapped
  branch still falls through.)

The emitted SQL declares the CTE with only a subset of columns while the recursive branch
references a pruned one — SQLite rejects it with `SQLite Error 1: 'circular reference: partHierarchy'`.

### Expected behavior

Columns used only in the recursive/outer joins must survive regardless of whether the two
set-operation branches share the exact same projected type. A recursive CTE whose recursive
term projects a derived type (cast to the CTE type) should generate the same well-formed SQL
as one whose terms share the exact type.

### Notes

- Related: #5457 (same column-drop class, same-type case) and its fix PR #5680. This gap was
  surfaced while reviewing #5680. Verified pre-existing: the repro fails **identically** on
  `master` and on #5680's branch, so it is **not** introduced by #5680.
- Suggested fix (unverified): in the `ConditionalExpression` case #5680 adds to
  `ProcessTranslated`, unwrap a `Convert`/`ConvertChecked` around each branch before matching,
  then re-wrap after propagating the member path. Needs its own regression tests.
- Repro (fails on SQLite; recursive-CTE providers generally):

```csharp
sealed class Part      { [PrimaryKey] public int Id { get; set; } public string? Name { get; set; } }
sealed class Reference { [PrimaryKey] public int Id { get; set; } public int ParentId { get; set; } public int ReferenceId { get; set; } }
class        CteBase   { public object? RootPartSortField { get; set; } public int RootPartId { get; set; } public int PartId { get; set; } public int HierarchyLevel { get; set; } }
sealed class CteDerived : CteBase { }

var anchor = parts
    .Select(x => new CteBase { RootPartSortField = x.Name, RootPartId = x.Id, HierarchyLevel = 0, PartId = x.Id })
    .OrderBy(x => x.RootPartSortField).Skip(0).Take(20);

var partCte = db.GetCte<CteBase>(h =>
    anchor.Concat(
        h.InnerJoin(refs, (cte, r) => r.ParentId == cte.PartId,
            // recursive term projects a DERIVED type cast to the CTE type -> Convert-wrapped branch
            (cte, r) => (CteBase)new CteDerived { RootPartSortField = cte.RootPartSortField, RootPartId = cte.RootPartId, PartId = r.ReferenceId, HierarchyLevel = cte.HierarchyLevel + 1 })));

var result = parts.InnerJoin(partCte, (me, cte) => me.Id == cte.PartId, (me, id) => new { id.RootPartId, me }).ToList();
// throws: SQLite Error 1: 'circular reference: partHierarchy'
```

sdanyliv:
Fixes #5683

### Cause

Not the one the issue proposes. The root is the key a projection's member path is built from: [`ExpressionPathVisitor`](https://github.com/linq2db/linq2db/blob/master/Source/LinqToDB/Internal/Linq/Builder/MergeProjectionHelper.cs) pushes the **type being constructed** onto the path stack, and `SetOperationBuilder.InitializeProjections` pairs the two branches by exactly that array.

When one branch projects a derived type cast to the set operation's element type, the two branches name different types for the same place:

```
$Path$ -> "construct" -> CteBase    -> RootPartId    // anchor
$Path$ -> "construct" -> CteDerived -> RootPartId    // recursive term
```

Nothing pairs up. Every member of the first branch gets a NULL placeholder for the second, and every member of the second registers under a place of its own — so each member is emitted twice, once filled and once NULL-padded. The CTE body that came out of the issue's repro:

```
(42) SELECT t38[21].RootPartSortField, t27[10].RootPartId, t27[10].RootPartId
UNION ALL
(40) SELECT t13.RootPartSortField, NULL, NULL
```

Hence SQLite's `circular reference: partHierarchy` on the original repro, and `Table not found` here — `t38` is a query that is nowhere in the statement.

### Fix

- **`SequenceHelper.IsConstructorBranch`** — a branch is a constructor seen through whatever conversion states the type it is read as. Two overloads, one handing back the constructor.
- **`MergeProjectionHelper.ExpressionPathVisitor.VisitUnary`** — a constructor wrapped in a conversion to an assignable type is keyed in the path by **the type the projection is read as**, not the one constructed. Through a chain of conversions the outermost one wins; anything constructed inside the assignments names itself.
- **`BuildProxyBase{TOwner}.ProcessTranslated`** — the `ConditionalExpression` case added by #5680 matched only bare constructor branches, so a Convert-wrapped one fell through to the generic visitor and lost its tracking path.

Both parts are needed: with the path fix alone the repro fails with `ArgumentOutOfRangeException`.

### Tests

`Tests/Linq/UserTests/Issue5683Tests.cs`:

- `RecursiveCteWithDerivedRecursiveTerm` — the issue's repro.
- `ConcatBranchProjectingDerivedType` — the same defect without a CTE: `AssertQuery` plus a column-count assertion.

`AssertQuery` alone does not catch this — the values are right, only the columns are surplus (4 where 3 are due) — so the second test asserts the emitted column count as well. Verified negatively: with the path fix disabled, both tests fail across all three configurations.

Full SQLite suite green (10955 tests, 0 failed).

---

### `as`-spelled projections — `9d185a1009`

The same defect survives in the other spelling of the cast. `select new Derived { … } as Base` never reached the pairing above, for two independent reasons, both in `ExpressionBuilder.Project`:

- **A `MemberInit` asked for as a whole threw** `NotImplementedException: Projecting '' is not supported yet`. The sibling `New` case answers the same question with its own body — `MemberInit` was the asymmetric one. Pre-existing on master and unrelated to set operations: a plain `select … as Base` with no `Concat` throws identically on `master`.
- **The `TypeAs` case stripped the conversion**, handing back the bare operand. The branch then keyed by the type it *constructs* rather than the type it is *read as* — the same defect this PR fixes for `(Base)`. With only the first fix the query builds and returns correct data, but emits four columns where three are due:

```sql
SELECT [p].[ParentID], [p].[Value1], NULL, NULL FROM [Parent] [p] WHERE [p].[ParentID] > 3
UNION ALL
...
```

Type preservation routes through `SequenceHelper.EnsureType`, which already special-cased `ContextRefExpression` via `WithType`; it gained the same treatment for `SqlErrorExpression`, so an error keeps its kind instead of being wrapped in a conversion.

New tests: `SelectConstructedObjectThroughAs` (the bare shape that throws on master) and `ConcatBranchProjectingDerivedTypeThroughAs` (asserting the same three-column result as its cast-spelled twin).

### Dead visitor copies removed — `0eaaddf56e`

`SetOperationBuilder` carried never-instantiated `ExpressionOptimizerVisitor` and `ExpressionPathVisitor` — 335 lines whose `VisitSqlGenericConstructorExpression` pushed `node.Type`, the exact line this PR changes in the live copy in `MergeProjectionHelper`. The only instantiations are in `MergeProjectionHelper`, so a later fix could land in the dead copy and appear to do nothing.

### Test strengthening

`ConcatBranchProjectingDerivedType` now asserts the per-branch runtime type — the discriminator column's stated purpose was unobserved, so a regression materialising both branches as `Projection` would have passed. `RecursiveCteWithDerivedRecursiveTerm` now pins `HierarchyLevel`, the one member written in both branches and never read back.

### Verification

21071 tests, 0 failed (`SQLite.MS` + `SQLite.Classic`). Release `net10.0` and `netstandard2.0` builds clean.


MaceWindu:
Pushed `7b7a4b41e` to fix the `test-all` failure from build 23346, then merged `master` (`93b3241f1`, CI-only commits).

### What was failing

`InheritanceTests.TypeCastAsTest11` — four cases, on `Lin s_SQLite` and `Lin l_SqlServer2017_2019`:

```
System.InvalidOperationException : No coercion operator is defined between types 'System.String' and 'Tests.Model.Northwind+Product'.
   at LinqToDB.Internal.Linq.Builder.SequenceHelper.EnsureType(...) in SequenceHelper.cs:269
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.Project(...) in ExpressionBuilder.SqlBuilder.cs:1816
```

Not SQLite-specific — it fires wherever a Northwind data source is configured, which is exactly those two jobs.

### Cause

`9d185a1009`'s `TypeAs` change converts `truePath` unconditionally. But `Project` is re-entered with `nextIndex` **unchanged**, so a remaining member path resolves *through* the operand and `truePath` is that member's value, not the cast object. The test's query is

```csharp
.Select(p => new { p = p as Northwind.Product })
.Select(p => p.p == null ? "NULL" : p.p.ProductName)
```

so the path still carries `.ProductName`, `truePath` comes back a `string` placeholder, and `EnsureType(string, Product)` falls through to `Expression.Convert`.

### Fix

Convert only when the whole object was asked for:

```csharp
return next == null ? SequenceHelper.EnsureType(truePath, unary.Type) : truePath;
```

`next == null` is reachable only when `path` is a non-member root, which is how `SelectContext.MakeExpression` and `SetOperationBuilder` request a whole projection. Every `Project` call site passes `nextPath: null`, so the dichotomy is clean — `path == null && next != null` (member path) or `path != null && next == null` (whole object). Same guard this PR already added for the `MemberInit` case.

### Verification

Worktree at `18d3ad248`, `Tests/Linq` net10.0 Debug:

| Arm | Filter / providers | Result |
|---|---|---|
| Red (unmodified head) | `TypeCastAsTest11` · `Northwind.SQLite` | `total: 3 · failed: 1`, identical exception |
| Green | `TypeCastAsTest11` · `Northwind.SQLite`, `Northwind.SQLite.MS` | `total: 4 · failed: 0` |
| Green | `Issue5683Tests` · `SQLite.MS`, `SQLite.Classic` | `total: 10 · failed: 0` |
| Green | `InheritanceTests` + `QueryInheritanceTests` + `Issue5683Tests` · 4 SQLite providers | `total: 357 · failed: 0` |

`test-all` build 23372 on `7b7a4b41e`: 42 checks green, both previously-red legs included.

### Checked, not a finding

The same `TypeAs` case also hands back `truePath.Type` rather than `unary.Type` in its other two arms — the constant-`false` branch and the non-constant-predicate conditional. Both pre-date this PR. Probed all three whole-object shapes on Northwind:

- statically-false test (`(Product)p as ActiveProduct` from `DiscontinuedProduct`) — 8 rows, all null. Correct.
- discriminator predicate (`db.Product.Select(p => p as DiscontinuedProduct)`) — 77 rows, 8 non-null, runtime type `DiscontinuedProduct`. Correct: the declared node type is cosmetic, materialization follows the inheritance mapping.
- the same under `Concat` — 11 columns (`Product`'s 10 plus the conditional's null test), branches paired, no padding.

No defect, nothing to change there.

MaceWindu:
Pushed `4dc056a93`, `a5ac7baef`, `117987295` after reviewing the branch. `test-all` triggered on the new head.

### Pair branches by the element type — `4dc056a93`

The fix keyed a projection's place by the type the branch *constructs*, corrected only where the branch physically carried a conversion node. C# doesn't require one: `Queryable.Concat` binds `TSource` from the first operand and takes the second through `IEnumerable<T>` covariance, so `baseQuery.Concat(derivedQuery)` compiles with no conversion anywhere. Such a branch reached the merge as a bare constructor typed after the derived class and paired with nothing.

The issue's own recursive-CTE repro, spelled without the `(Hierarchy)` cast, still failed with `Table not found` on this branch before the change, and a plain `Concat` emitted four columns where three are due — both identical on `master`, so this was never fixed rather than newly broken.

`InitializeProjections` already knows the type both branches are read as, so it now passes `ElementType` down and `ExpressionPathVisitor` is seeded with it, re-applied before each of `ProcessExpression`'s two passes. Confined to the set-operation path — `EnumerableContextDynamic`, the other `BuildProjectionExpression` caller, keeps its behaviour through the parameter default.

### Doc correction — `a5ac7baef`

`IsConstructorBranch`'s summary said "whatever conversion states the type it is read as", which reads as covering `TypeAs`. `UnwrapConvert` handles `Convert` / `ConvertChecked` carrying no conversion operator, plus `SqlAdjustTypeExpression`. `TypeAs` is rewritten into a conversion earlier, which is the only reason the `as`-spelled tests pass. The predicate now also gates `BuildProxyBase.ProcessTranslated`'s conditional case, so the wrong assumption could reopen the issue from a site where nothing fails.

### Test coverage — `117987295`

- `ConcatBranchProjectingDerivedTypeWithoutCast` and `RecursiveCteWithDerivedRecursiveTermWithoutCast` pin the covariance spelling.
- `UnionBranchProjectingDerivedTypeIsRefused` and `ExceptBranchProjectingDerivedType` pin what the distinct set operations do with differing branch types. Pairing the branches removes the NULL padding that used to double as the branch discriminator, and the anchor field is created only for `UNION ALL`, so `Union` and `Intersect` now refuse instead of comparing rows the database cannot compare. On 6.4.0 all three were silently wrong — `Union` returned 9 rows where 7 were due, `Intersect` none where 2 were due, `Except` removed nothing. `Except` needs no discriminator and is now correct.
- `ConcatBranchProjectingDerivedTypeThroughAs` gained the per-branch construction assertions its cast-spelled twin already carried. Without them an inverted discriminator passed, since `AssertQuery` compares member-wise and the column count is symmetric. Verified by inverting the discriminator and re-running: the test stayed green as authored and went red once the assertions were added, while both sibling tests caught it.
- The three tests that excluded remote contexts now run on them as well.

### Verification

34/34 green on `SQLite.MS` + `SQLite.Classic` and their `LinqService` legs.

That is deliberately not enough on its own. `4dc056a93` seeds every set operation, not only the differing-type case — for same-type branches the seed equals the type already being keyed, so it should be inert, but that is exactly the claim the full provider matrix exists to test. The earlier baselines run for this PR showed 0 modified baselines; this change could move some. Worth a look at the baselines diff from the run in flight before merging.

### Unrelated finding

Filed #5881 for a limitation found while probing: a chain of three or more set operations with a differing-type branch pads every member apart, emitting 7 columns where 2 suffice and dropping the reader into slow mapping mode. Verified pre-existing and unaffected by this PR — `master` produces the identical SQL — so it is out of scope here.

MaceWindu (dismissed):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- ⚠ `Union`, `Intersect` and `IntersectAll` over branches that project different but assignable types now report a clear error instead of returning rows. The rows they returned were wrong — unlike `UnionAll`, a distinct set operation has no way to tell the branches apart — so a silent data defect became a loud one. Add an explicit `Select(x => new Projection { ... })` to each branch to say which shape to build.
- `Concat` and `UnionAll` where one branch projects a derived type read as the set operation's element type now generate valid SQL, including inside a recursive CTE. Previously every member was emitted twice — once filled, once padded with `NULL` — and the server rejected the statement.
- `Except` and `ExceptAll` over such branches now actually remove rows. Previously the branches had no column in common, so the operation removed nothing and returned everything.
- Projecting a whole constructed object through a base type — `select new Derived { ... } as Base` — no longer throws. This applies with or without a set operation involved.


Set operations over branches projecting different but assignable types are fixed; `Union` / `Intersect` over such branches now report an error instead of returning wrong rows.


---
<sub>Generated from commit `7214ff0`.</sub>

PR #5834, opened 2026-08-31
CI: drop macOS test jobs and trim non-netfx Windows TFMs

MaceWindu:
First step of moving the test matrix to GitHub Actions. Both changes stand on their own and shrink the Azure DevOps run today, independently of anything that follows.

## Drop macOS

The macOS legs have been configured but switched off for a long time — `mac_enabled` defaults to false in every caller — and there are no macOS-only test scenarios, so the configuration was dead weight spread over four pipeline files, 25 matrix entries and five setup scripts.

Removes `test_macos_job` and `test-workflow-macos.yml`, the `mac_enabled` parameter, the `config_macos` / `script_macos_global` / `script_macos_local` / `install_docker_macos` / `enable_os_macos` fields, and the five mac-only setup scripts. Setup scripts that several entries shared with Linux (`clickhouse.sh`, `firebird.sh`, `oracle*.sh`, `sqlserver.*.sh`, `ydbsybase.sh`) are kept. Also removes the deleted files from `linq2db.slnx` and a comment in `db2informixduckdb.sh` about raising the Informix wait bound if the macOS legs ever came back.

## Trim non-netfx Windows TFMs

The Windows legs ran the EF.Core suite on every enabled TFM unconditionally, and on release runs the main suite too. For a leg that also has a Linux job, net8/9/10 there repeat what the Linux leg already ran — so the Windows matrix now enables net8/9/10 only for a leg with no Linux job. `netfx` is untouched: it runs nowhere but Windows.

The rule is derived from `enable_os_ubuntu` rather than restated per entry, because every leg that must keep its non-netfx TFMs on Windows (Access, SqlCE, SQL Server 2005–2016) has no Linux job. `win_tfms_always` opts an entry out, for a Windows leg testing something the Linux leg cannot reach on the same TFM — `w_DB2InformixDuckDB`, whose Windows leg runs DuckDB alone against the windows-native binary. Without the opt-out that leg would have had no TFM enabled at all and failed on a missing `.trx`, since DuckDB has no netfx tests.

Coverage-neutral for the four affected legs: SQLite, SqlServer2022_2025 and SqlServer2019Extras use one config file for both OSes, and `l_SqlServer2017_2019`'s Linux config is a superset of its Windows one — `sqlserver.fts.2017_2019.json` adds the Northwind FTS providers.

| Windows leg | TFMs after |
|---|---|
| SQL Server 2005 / 2008 / 2012 / 2014 / 2016, SqlCE, Access MDB + ACE ODBC + ACE OLEDB | netfx + net8/9/10 |
| SQL Server 2017_2019, 2022_2025, Extras, SQLite | netfx |
| DB2 + Informix + DuckDB (release runs only) | net8/9/10 |

`Build/Azure/README.md` gains the `:arrow_right:` legend entry and marks the nine affected cells. It also corrects the SQL Server 2017 row's Linux cell, which claimed no Linux leg — that leg exists, and marking the Windows cell Linux-covered would otherwise have made the row contradict itself.

## Verification

Every changed pipeline YAML re-parses. The per-entry Windows TFM set was enumerated mechanically from the matrix rather than read off, confirming the table above and that no leg is left with zero TFMs enabled.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `6d92e4e`.</sub>

PR #5836, opened 2026-08-31
CI: run the database-free build and checks on GitHub Actions

MaceWindu:
Second step of moving CI to GitHub Actions. Ports the database-free half of the Azure Pipelines `build` definition, which is everything that needs no database, no container and no secret — so it can run automatically on every PR.

The provider test matrix is **not** part of this change; it stays on Azure Pipelines and moves in a later step.

## Jobs

| Job | Runner | Ported from |
|---|---|---|
| `build` | windows-2025 | Release build + `dotnet pack`, then `verify-nuget-sizes.ps1` and `verify-analyzer-delivery.ps1`; uploads the `nugets` and `linq2db_linqpad_lpx` artifacts |
| `examples` | windows-2025 | `Build Examples (verify)` |
| `analyzer-tests` | ubuntu-24.04 | `Run Analyzer Tests` |
| `singlefile-smoke` | windows-2025 | `PublishSingleFile Smoke Test` (the #5488 regression guard) |
| `cli-tests` | ubuntu-24.04 + windows-2025 | `templates/test-cli.yml` |

## Two constraints that drive the layout

**Anything under `Tests/` builds `-c Azure`.** `Tests/Directory.Build.props:17-21` defines the `AZURE` symbol from that configuration name, and `TestConfiguration.cs:62-65` uses it to append `.Azure` to the config key — which is what selects the CI connection strings and `BaselinesPath` from `DataProviders.json`. Building `-c Release` instead is silently wrong rather than broken: the run picks up local connection strings and captures no baselines, with nothing red anywhere. The configuration name is kept verbatim for the same reason; renaming it is a much larger change than it looks.

**Every job that touches `net462` runs on Windows.** That is the Release build of `linq2db.slnx` and the `Examples` solution, which has `net462` projects of its own (`net462;net9.0;` and a bare `net462`). The repo carries no `Microsoft.NETFramework.ReferenceAssemblies` reference, so a Linux build of either fails.

## Reuse

`verify-nuget-sizes.ps1` and `verify-analyzer-delivery.ps1` are used **unmodified** — both already carry a `-NoAzdoLogs` switch and signal through their exit code, so only the switch was needed. Note these currently run in `nuget-job.yml`, which never runs on a `master`-targeting PR, so this is the first time a PR gets the nupkg size and analyzer-delivery gates at all.

Actions are GitHub-owned only and pinned by SHA, matching the repository's `allowed_actions: selected` + `github_owned_allowed` policy.

## Also here

- `.github/CODEOWNERS` gains `/Build/` and `/.github/workflows/` entries. Neither path had an owner, so CI definitions have been landing with no required-owner review — which matters more once workflow files exist, since a workflow executes with repository credentials the moment it merges.
- `.gitattributes` gains an explicit `*.yml eol=lf` rule beside the existing `*.sh` one. This is consistency, not a fix: all 25 tracked `.yml` files are already LF in the index under `text=auto`, so nothing is renormalized and no stored content changes.

`.github/` needs no `linq2db.slnx` edit — #5752 replaced the per-file listing with a `NoTargets` glue project that globs the directory.

## Not in this change

The Azure `build` definition still runs on every PR. Turning it off (`pr: none`) is deliberately sequenced **after** this workflow is proven green and after the required-check list in branch protection is updated — flipping it first would stall every PR if that check is currently required.

## Verification

`pull_request` runs execute from the merge commit, so this PR's own run is the verification. `if-no-files-found: error` on both artifact uploads is deliberate: a silently missing `.lpx` is how that artifact goes stale without anyone noticing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,5 +4,6 @@
 *.slnx     text      eol=crlf
 *.csproj   text      eol=crlf
 *.sh       text      eol=lf
+*.yml      text      eol=lf
 .githooks/** text    eol=lf
 "/Data/Setup Scripts/ase.sql" text eol=lf
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,7 @@
 /Source/**/*.cs  @igor-tkachev @sdanyliv @MaceWindu @viceroypenguin @jods4 @Shane32
 /Source/**/*.ttinclude  @igor-tkachev @sdanyliv @MaceWindu @viceroypenguin @jods4 @Shane32
+
+# CI definitions. A workflow file executes with repository credentials the moment it merges, so it
+# wants the same review bar as source. Neither path had an owner before.
+/Build/  @MaceWindu
+/.github/workflows/  @MaceWindu
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,23 +30,12 @@ concurrency:
   group: build-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-# Every actions/setup-dotnet step below carries `DOTNET_INSTALL_DIR: ${{ runner.temp }}/dotnet`, which
-# installs the SDKs into a clean directory instead of alongside the image's preinstalled ones.
-#
-# That is not tidiness - the build does not compile without it. CodeGenerators is a source generator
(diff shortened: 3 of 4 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `8853cb2`.</sub>

PR #5839, opened 2026-09-01
THROWAWAY probe: Linux containers and native DB servers on Windows agents

MaceWindu:
**Do not merge. Delete the branch when the answers are recorded.**

Exists only to trigger the Azure `build` definition, which is bound to `Build/Azure/pipelines/build.yml` and runs on every PR — so that file is *replaced* on this branch with a probe. The GitHub half is `.github/workflows/probe-win-db.yml` and runs on push.

## The question

We don't run EF.Core netfx tests against PostgreSQL or MySQL. The standing reason is that netfx runs only on Windows, those servers run in Linux containers, and a Windows agent can't run Linux containers.

That may be the wrong framing. `images/windows/Windows2025-Readme.md` lists **PostgreSQL 17.11 as a real service** on the image — `postgresql-x64-17`, Stopped and Disabled, but present, with a documented `postgres`/`root` credential. A native server needs no container at all. MySQL 8.0.46 appears under *Database tools* rather than under *Databases* with a ServiceName, which suggests client-only — but that's an inference from a heading, not a fact.

## What both halves ask

1. Can the agent run a Linux container? (Expected no; the value is in *how* it fails — daemon in Windows mode versus no nested virtualisation.)
2. Does the native PostgreSQL service start and accept a connection?
3. Is there a MySQL server, or only client tools?

Both on `windows-2025` and `windows-2022`.

## Why both CIs

Azure and GitHub run the **same** `actions/runner-images` artifact. If the answers differ, the difference is the host or the agent rather than what's installed. That decides whether migrating CI is what unblocks this coverage — or whether it was available on Azure all along and the container framing was simply never revisited.

This PR also triggers the newly merged `build` workflow (#5836). That's incidental, and a first real-world exercise of it.

### DIFF
--- a/.github/workflows/probe-win-db.yml
+++ b/.github/workflows/probe-win-db.yml
@@ -117,7 +117,12 @@ jobs:
           export DEBIAN_FRONTEND=noninteractive
           apt-get update -qq
           apt-get install -y -qq docker.io iptables >/dev/null
+          # Before dockerd, not after. The previous run set this once the daemon was already up, so
+          # dockerd had built its chains under nftables and the legacy binary could not see them:
+          # "Unable to enable DNAT rule ... iptables: No chain/target/match by that name". Containers
+          # ran fine, only -p publishing broke - which is the one thing this probe needs.
           update-alternatives --set iptables /usr/sbin/iptables-legacy || true
+          update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy || true
           nohup dockerd >/tmp/dockerd.log 2>&1 &
           for i in $(seq 1 40); do docker info >/dev/null 2>&1 && break; sleep 2; done
           echo "--- docker info"
@@ -125,7 +130,7 @@ jobs:
           echo "--- hello-world"
           docker run --rm hello-world || { echo "--- dockerd.log"; tail -40 /tmp/dockerd.log; exit 11; }
           echo "--- postgres with a published port"
-          docker run -d -p 15432:5432 -e POSTGRES_PASSWORD=probe postgres:17-alpine
+          docker run -d -p 15432:5432 -e POSTGRES_PASSWORD=probe postgres:17-alpine || { echo "--- dockerd.log"; tail -40 /tmp/dockerd.log; exit 13; }
           for i in $(seq 1 40); do (echo > /dev/tcp/127.0.0.1/15432) >/dev/null 2>&1 && break; sleep 2; done
           ss -lntp | grep 15432 || { echo "port never opened inside wsl"; exit 12; }
           '@ -replace "`r`n", "`n"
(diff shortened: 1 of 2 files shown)


MaceWindu:
Closing — questions answered, recorded here since the branch goes away.

## Settled

| | GitHub windows-2025 | Azure windows-2025 |
|---|---|---|
| Linux image via the agent's docker | no — exit 125, `no matching manifest for linux/amd64` | identical |
| Native **PostgreSQL 17.11** | **works** — service starts from Stopped/Disabled, `select version()` answers, `CREATE DATABASE` succeeds | identical |
| Native **MySQL 8.0** server | `mysqld`/`mysql`/`mysqladmin` present, no registered service | identical |
| WSL2 distro boots | yes — `Linux 6.18.33.2-microsoft-standard-WSL2` | yes (when `wsl --install` completes) |
| dockerd inside WSL2 | yes — `server=29.1.3`, ran a real amd64 container | unproven |
| Publishing a port from WSL2 (`-p`) | **no** — DNAT fails, `iptables: No chain/target/match by that name` | unproven |

**The headline: EF Core netfx coverage for PostgreSQL and MySQL was never blocked by "Azure can't run Linux containers".** A native PostgreSQL 17 server ships on the Windows image of *both* CIs — stopped and disabled, three lines to start. MySQL needs `mysqld` initialising rather than just starting, but the server is installed. Both CIs use the same `actions/runner-images` artifact, so this has been available on Azure all along.

## Not settled, and why we stopped

Port publishing out of WSL2 fails on the `nat` table. Likely mechanism: since kernel 6.6.y, WSL2 ships `NETFILTER_XT_NAT` / `NFT_COMPAT` / `IP_NF_NAT` as modules in a separate VHD while `NF_NAT`/`NFT_NAT` are built in — which matches chain creation succeeding and only the DNAT *target* failing.

Two further runs never reached the diagnostics: the Docker service wasn't running on a GitHub runner (`npipe:////./pipe/docker_engine` not found), and `wsl --install` didn't complete on Azure (22 min when it worked, unfinished when it didn't).

## Worth carrying forward

- **The Docker service on GitHub Windows runners is not reliably up at job start.** Any future docker-dependent step there needs `Start-Service docker` plus a readiness wait.
- **`wsl --install` is not CI-stable on hosted agents.** A cached `wsl --export`/`--import` distro is a precondition for this approach, not an optimisation.
- **Per database:** Oracle has a native Windows server (23ai Free, silent install) and never needed WSL. SAP HANA needs 8 GB minimum / 12 recommended — impossible on the Azure 2 vCPU / 7 GB agent and marginal on GitHub's 4/16 with a 14 GB disk. That leaves **Informix and Sybase ASE** as the only two genuinely needing the WSL route.

`Build/Azure/pipelines/build.yml` was *replaced* on this branch to hijack the `build` definition's PR trigger — which is why this must never merge.

Issue #5838, opened 2026-09-01
InsertWithOutput/UpdateWithOutput on an inheritance-mapped table read the key as the discriminator


MaceWindu:
### Summary

On a table mapped with `[InheritanceMapping]`, `InsertWithOutput` and `UpdateWithOutput` write the row correctly and then fail materializing the row they return:

```
LinqToDBException: Inheritance mapping is not defined for discriminator value '30' in the 'Base' hierarchy.
```

The reported discriminator value is always the row's **primary key**, not its discriminator column — `'30'` for the row with `ID = 30`, `'70'` for `ID = 70`, `'72'` for `ID = 72`. The returned-row projection appears to read the wrong column for the discriminator.

### Background

This is currently **masked** by #5729. On `master`, both calls fail earlier with `ArgumentException: Property 'X' is not defined for type 'Base'` out of `Expression.MakeMemberAccess`, so the materialization step is never reached. The defect below is only observable once #5729's fix is applied, and it is not caused by it.

The reason an output clause reaches this at all: the default output projection over an inheritance root builds a full entity, and for a TPH root `EntityDescriptor.InitInheritanceMapping` has merged every subtype's columns into the base descriptor (`Source/LinqToDB/Mapping/EntityDescriptor.cs:518-524`). So the projection legitimately carries subtype columns even when the *insert setter* mentions only base members.

### Current behavior

Measured on SQLite with a base `Base` (`ID` primary key, `Code` discriminator) and a derived `Child1` adding a `Name` column, with #5729's fix applied:

| Call | Result |
|---|---|
| `GetTable<Base>().InsertWithOutput(() => new Base { ID = 70, Code = Code.Child2 })` | throws, discriminator value `'70'` |
| `GetTable<Base>().InsertWithOutput(() => new Child1 { ID = 72, ... })` | throws, discriminator value `'72'` |
| `GetTable<Base>().Where(...).UpdateWithOutput(t => new Base { ... })` | throws, discriminator value = the matched row's `ID` |
| `GetTable<Child1>().InsertWithOutput(() => new Child1 { ... })` | throws `NotSupportedException: Specified method is not supported.` |
| `GetTable<Base>().Where(...).DeleteWithOutput()` | no failure observed |

Note the first row: the initializer mentions only base members, so this is not specific to writing a derived type.

The rows themselves are written correctly — reading the table back with raw SQL after the failure shows the expected values, including the discriminator:

```
ID / Code / Name
70 / 1    / <null>
[...]

Issue #5837, opened 2026-09-01
Derived-entity writes through a base-mapped table fail in entity-builder, Upsert and Merge


MaceWindu:
### Summary

Writing an entity of a derived type through a table mapped to its `[InheritanceMapping]` base fails with `InvalidOperationException: Member 'X' not found in type 'Base'` on the entity-builder spec API, the `Upsert` API, and Merge's parameterless clauses.

#5729 reports the same user-visible situation on the setter-lambda path (`Insert(() => new Derived { ... })`), where it fails by a different mechanism — `ArgumentException` out of `Expression.MakeMemberAccess`. That path is being fixed separately; this one needs a different fix and is filed on its own.

### Background

For a TPH root, `EntityDescriptor.InitInheritanceMapping` merges each subtype's `ColumnDescriptor` verbatim into the base descriptor's `Columns` (`Source/LinqToDB/Mapping/EntityDescriptor.cs:518-524`). A base descriptor's column can therefore carry a `MemberInfo.DeclaringType` that is a *derived* type. Any code that walks those columns against a base-typed instance has to cope with that, and these paths do not.

### Current behavior

- `Source/LinqToDB/Internal/Mapping/ColumnDescriptorExtensions.cs:67-72` — the leaf branch resolves a column with `instance.Type.GetMemberEx(descriptor.MemberInfo)`, a by-name lookup on the instance type, and throws when it returns null. For a derived-only column against a base-typed instance it always returns null.
- `Source/LinqToDB/Internal/Linq/Builder/EntitySetterBuilder.cs:47` — iterates every merged column of every branch; `:68` builds each value through the helper above.
- `Source/LinqToDB/Internal/Linq/Builder/UpsertBuilder.cs:117`, `:126` — the same shape against a base-typed `entityParameter`.
- `Source/LinqToDB/Internal/Linq/Builder/MergeBuilder.UpdateWhenMatched.cs:51-66` — the no-setter branch walks every updatable field with `targetRef` typed as the table's object type; `MergeBuilder.UpdateWhenMatchedThenDelete.cs:61-62` is the sibling.

Measured against `639d6aaf8` on SQLite, with a base `Base` and a derived `Child1` carrying a `Name` column declared only on `Child1`:

| Call | Result |
|---|---|
| `GetTable<Base>().Insert(child1, spec => spec)` | throws |
| `GetTable<Base>().Upsert(child1)` | throws |
[...]

PR #5840, opened 2026-09-01
Fix writing a derived-type entity through a base-mapped table

### LINKED ISSUE #5729: Issue with InheritanceMapping
### Description

Hello,
I am trying to upgrade to version 6.3.0 and have encountered an issue with InheritanceMapping.
The code below throws a System.ArgumentException: 'Property 'System.String Name' is not defined for type 'ConsoleApp2.Test2+Base' (Parameter 'property')', although it works perfectly fine in version 2.6.4.

### Reproduction Steps

```c#
using LinqToDB;
using LinqToDB.Data;
using LinqToDB.DataProvider.SQLite;
using LinqToDB.Mapping;

internal class Program
{
    public enum Code
    {
        Child1,
        Child2,
    }

    [Table("MyTable")]
    [InheritanceMapping(Code = Code.Child1, Type = typeof(Chil1))]
    [InheritanceMapping(Code = Code.Child2, Type = typeof(Chil2))]
    public class Base
    {
        [Column(IsPrimaryKey = true)]
        public int ID { get; set; }
        [Column(IsDiscriminator = true)]
        public Code Code { get; set; }
    }
    public class Chil1 : Base
    {
        [Column]
        public string? Name { get; set; }
    }
    public class Chil2 : Base { }

    private static void Main(string[] args)
    {
#if OLD_LINQ2DB
        using var db = new DataConnection(new SQLiteDataProvider(ProviderName.SQLiteClassic), "Data Source=:memory:");
#else
        using var db = new DataConnection(new DataOptions().UseConnectionString("Data Source=:memory:").UseSQLite());
#endif
        db.Execute("""
                        CREATE TABLE `MyTable` (
            	            `ID`	INTEGER NOT NULL,
            	            `Code`	INTEGER NOT NULL,
            	            `Name`	TEXT,
            	            PRIMARY KEY(ID)
                        )
            """);
        db.GetTable<Base>().Insert(() => new Chil1 { Code = Code.Child1, Name = "Jack" });
        Console.WriteLine(db.LastQuery);
    }
}
```

csproj
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <!--<DefineConstants>$(DefineConstants);OLD_LINQ2DB</DefineConstants> -->
  </PropertyGroup>
	
  <PropertyGroup Condition=" !$(DefineConstants.Contains('OLD_LINQ2DB')) ">
  	<LinqToDB_Version>6.3.0</LinqToDB_Version>
  </PropertyGroup>
  <PropertyGroup Condition=" $(DefineConstants.Contains('OLD_LINQ2DB')) ">
  	<LinqToDB_Version>2.6.4</LinqToDB_Version>
  </PropertyGroup>

	<ItemGroup>
		<PackageReference Include="linq2db" Version="$(LinqToDB_Version)" />		
		<PackageReference Include="linq2db.SQLite" Version="$(LinqToDB_Version)" />
		<PackageReference Include="System.Data.SQLite.Core" Version="1.0.115.5" />		
	</ItemGroup>

</Project>
```

### Exception details
```
System.ArgumentException
  HResult=0x80070057
  Message=Property 'System.String Name' is not defined for type 'Program+Base' (Parameter 'property')
  Source=System.Linq.Expressions
  StackTrace:
   at System.Linq.Expressions.Expression.Property(Expression expression, PropertyInfo property)
   at LinqToDB.Internal.Linq.Builder.UpdateBuilder.ParseSetter(ExpressionBuilder builder, ContextRefExpression targetRef, ContextRefExpression sourceRef, Expression setterExpression, List`1 envelopes)
   at LinqToDB.Internal.Linq.Builder.InsertBuilder.BuildMethodCall(ExpressionBuilder builder, MethodCallExpression methodCall, BuildInfo buildInfo)
   at LinqToDB.Internal.Linq.Builder.MethodCallBuilder.BuildSequence(ExpressionBuilder builder, BuildInfo buildInfo)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.TryBuildSequence(BuildInfo buildInfo)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.BuildSequence(BuildInfo buildInfo)
   at LinqToDB.Internal.Linq.Builder.ExpressionBuilder.Build[T](IQueryExpressions& expressions)
   at LinqToDB.Internal.Linq.Query`1.CreateQuery(ExpressionTreeOptimizationContext optimizationContext, ParametersContext parametersContext, IDataContext dataContext, IQueryExpressions& expressions)
   at LinqToDB.Internal.Linq.Query`1.GetQuery(IDataContext dataContext, IQueryExpressions& expressions, Boolean& dependsOnParameters)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.GetQuery(IQueryExpressions& expression, Boolean cache, Boolean& dependsOnParameters)
   at LinqToDB.Internal.Linq.ExpressionQuery`1.System.Linq.IQueryProvider.Execute[TResult](Expression expression)
   at LinqToDB.LinqExtensions.Insert[T](ITable`1 target, Expression`1 setter)
   at Program.Main(String[] args)
```

### Regression?

_No response_

### Linq To DB version

6.3.0

### .NET version

8.0

### Database

SQLite

### Database provider

System.Data.SQLite.Core 1.0.115.5

### Operating system

Windows

### Configuration

_No response_

MaceWindu:
Fixes #5729.

`GetTable<Base>().Insert(() => new Derived { ... })` on a table mapped with `[InheritanceMapping]` threw:

```
System.ArgumentException: Property 'Name' is not defined for type 'Base' (Parameter 'property')
```

## Cause

`UpdateBuilder.ParseSetter` builds the field side of a setter with `Expression.MakeMemberAccess` against a target that carries the **table's** entity type, while the member being assigned is declared on the **subtype**. The BCL factory rejects that pairing. The message is not ours — it comes straight from `System.Linq.Expressions`.

The same failure reaches `Update` (both overloads), `InsertOrUpdate` (all three of its lambdas share one context ref), and the output-clause paths.

## Fix

Retype the target to the member's declaring type before building the access, through `SequenceHelper.EnsureType` — `ContextRefExpression.WithType` when the target is a context ref, `Expression.Convert` otherwise.

Retyping alone is sufficient: `EntityDescriptor.InitInheritanceMapping` merges each subtype's `ColumnDescriptor` verbatim into the base descriptor's `Columns`, so the derived column is a real `SqlField` on the base `SqlTable`, and `TableBuilder.TableContext.GetField` matches on **member identity** rather than on the target's type.

Three sites share one file-private helper:

- `ParseSetter`'s assignments arm — the reported shape.
- `ParseSetter`'s parameters arm — constructor-parameter (positional/record) subtypes.
- `ParseSet`'s nested-generic recursion — what the output-clause paths reach. A row-returning output projection over an inheritance root carries the subtypes' merged columns, so it fails there **even when the setter mentions only base members**.

The retype idiom already exists in the tree at four `SequenceHelper` sites and at `ExpressionBuildVisitor.cs:1682`; this adds no new pattern.

## Tests

Six regression tests in `InheritanceTests.cs`, one per edit-point at minimum:

| Test | Covers |
|---|---|
| `InsertDerivedThroughBaseTable` | the reported shape; asserts the derived column reaches the SQL |
| `UpdateDerivedThroughBaseTable_Setter` / `_ExplicitTarget` | both `Update` overloads (two independent `ParseSetter` call sites) |
| `InsertOrUpdateDerivedThroughBaseTable` | all three `InsertOrUpdate` lambdas |
| `InsertPositionalDerivedThroughBaseTable` | the parameters arm — needs a new positional model, since the existing one is property-only |
| `UpdateWithOutputDefaultProjectionOverInheritanceRoot` | the `ParseSet` recursion |

All six fail on `master` with `ArgumentException`; all pass with the fix.

## Verification

Verified locally on **SQLite.MS only** — 13/13 passing. SQL Server was not reachable on this machine (`sql2016` is a named instance, not a container), so the SQL Server leg is unverified locally and relies on CI.

## Known remaining, deliberately not fixed here

- **#5838** — the row-returning output paths now get *past* expression building and fail materializing the returned row, reading the primary key as the discriminator. Previously masked by this bug. `UpdateWithOutputDefaultProjectionOverInheritanceRoot` therefore asserts only that the failure is no longer `ArgumentException`, rather than asserting success.
- **#5837** — the entity-builder spec API (`Insert(item, spec)`), the `Upsert` API, and Merge's parameterless clauses fail the same user-visible situation by a *different* mechanism, in `ColumnDescriptorExtensions.GetMemberAccessExpression`. Fixing that one in the same helper would emit cross-branch downcasts and turn a build-time error into a runtime `InvalidCastException`, so it needs a different design and its own change.

## Follow-up commits (review)

- `bdf0080db` — corrected the `EnsureDeclaringType` remarks. Both halves of the justifying sentence were overstated: a subtype's `ColumnDescriptor` is merged into the base descriptor only when its member name has not already been seen (a name-colliding sibling is diverted to the inheritance sibling columns), and the field lookup compares members by name plus a same-or-parent declaring-type relation via `ReflectionExtensions.EqualsTo` rather than by member identity. The relation is load-bearing in both directions — identity alone would be too narrow to resolve the derived column at all, while requiring a same-or-parent relation is what keeps two sibling-declared members of the same name from matching each other.
- `d7db0c91a` — added the missing space after the comma in two of the three new `EnsureDeclaringType` calls; the third was already correct.
- `5a34f1c22` — removed `ParseSet`'s `targetPath` parameter, which was **write-only**. It had exactly two occurrences: its declaration, and a single `MakeMemberAccess` at the head of the nested-generic recursion whose result reached only the next recursion's `targetPath`, while the terminal branch envelopes `correctedField`. The declaring-type guard added for that site was therefore protecting a value nobody reads, so the site and one of the three guard calls are gone with it. The two guards in `ParseSetter` stay, because there the member access is also the field expression and does reach the envelope.

  This supersedes the third edit-point rather than keeping it: a discarded computation no longer needs a guard. Verified rather than assumed — with the parameter deleted, a temporary probe at the recursion recorded 48 entries across 16 distinct members, including all eight members of the `BaseClass` hierarchy (both child branches and all four grandchildren) and three unrelated nested-generic groups (`InsertWithOutputTests.ExternalId`, `Model.FullName`, and the entity Insert/Update nested names), while the Insert, Update, InsertOrUpdate, Merge, MultiInsert, output-clause and Inheritance suites ran 755/755 green on SQLite.MS. A second run with the probe removed was 756/756. The branch is genuinely exercised, so the green run is evidence rather than an untouched code path.

### DIFF
--- a/Tests/Linq/Linq/InheritanceTests.cs
+++ b/Tests/Linq/Linq/InheritanceTests.cs
@@ -1536,9 +1536,10 @@ namespace Tests.Linq
 			Assert.That(result.Value, Is.EqualTo(42));
 		}
 
-		// Matches FeatureUpdateOutputWithoutOldSingle in UpdateWithOutputTests: UpdateWithOutput returning
-		// a single set built from INSERTED only.
-		const string FeatureUpdateOutput = $"{TestProvName.AllSqlServer},{TestProvName.AllFirebirdLess5},{TestProvName.AllPostgreSQL},{TestProvName.AllSQLite},{TestProvName.AllYdb},{TestProvName.AllDuckDB}";
+		// FeatureUpdateOutputWithoutOldSingle from UpdateWithOutputTests, minus YDB: the default projection
+		// emits Deleted and Inserted, so every column appears twice in RETURNING and YDB rejects that with
+		// "Duplicated member".
+		const string FeatureUpdateOutput = $"{TestProvName.AllSqlServer},{TestProvName.AllFirebirdLess5},{TestProvName.AllPostgreSQL},{TestProvName.AllSQLite},{TestProvName.AllDuckDB}";
 
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/5729")]
 		public void UpdateWithOutputDefaultProjectionOverInheritanceRoot([IncludeDataSources(true, FeatureUpdateOutput)] string context)
@@ -1546,28 +1547,23 @@ namespace Tests.Linq
 			using var db = GetDataContext(context);
 			using var _  = db.CreateLocalTable(BaseClass.Data);
 
-			// BaseClass is abstract, so the setter must construct a concrete leaf type, but it assigns only
-			// Id/Code - both declared on BaseClass itself. The pre-fix ArgumentException does not come from
-			// the setter: UpdateWithOutput<T> without an explicit outputExpression defaults to projecting
-			// Deleted/Inserted as full BaseClass instances, and that default projection flattens in every
-			// mapped subtype's columns (Child1Field, Child2Field, ...) against a BaseClass-typed target -
-			// the same EnsureDeclaringType mismatch as the setter case, just reached through the default
-			// OUTPUT projection instead of the setter. Once fixed, the query still fails, now with a
-			// LinqToDBException about a discriminator value - a separate, still-open bug:
-			// https://github.com/linq2db/linq2db/issues/5838. So we assert only that the failure is no
-			// longer ArgumentException.
-			// Deliberately not Assert.Catch: that would require a throw, so this test would start
-			// failing the day #5838 is fixed and the query succeeds. Only the regression is asserted.
-			try
+			// The setter names only Code, declared on BaseClass itself, so it reaches the guard nowhere. The
+			// projection does: UpdateWithOutput without an explicit output expression builds full entities
+			// for Deleted and Inserted, and over an inheritance root those carry every mapped subtype's
+			// merged columns against a BaseClass-typed target. Pre-fix that threw ArgumentException before
+			// any SQL was emitted. Asserting on Inserted only - the providers here do not supply the old
+			// row, per FeatureUpdateOutputWithoutOldSingle.
+			var output = db.GetTable<BaseClass>()
+				.Where(t => t.Id == 1)
+				.UpdateWithOutput(t => new Child1 { Code = t.Code })
+				.ToArray();
+
+			Assert.That(output, Has.Length.EqualTo(1));
+			using (Assert.EnterMultipleScope())
 			{
-				db.GetTable<BaseClass>()
-					.Where(t => t.Id == 1)
-					.UpdateWithOutput(t => new Child1 { Code = t.Code })
-					.ToArray();
-			}
-			catch (Exception ex)
-			{
-				Assert.That(ex, Is.Not.InstanceOf<ArgumentException>());
+				Assert.That(output[0].Inserted, Is.InstanceOf<Child1>());
+				Assert.That(output[0].Inserted.Id, Is.EqualTo(1));
+				Assert.That(((Child1)output[0].Inserted).Child1Field, Is.EqualTo(11));
 			}
 		}
 
--- a/Tests/Linq/Linq/InheritanceTests.cs
+++ b/Tests/Linq/Linq/InheritanceTests.cs
@@ -1557,6 +1557,26 @@ namespace Tests.Linq
 			}
 		}
 
+		// Derives from BaseClass but carries no [InheritanceMapping] entry, so UnmappedField maps to no
+		// column on the base entity descriptor. EnsureDeclaringType retypes the target for it just the
+		// same, so the guard must not turn the failure into a silently omitted column.
+		class UnmappedChild : BaseClass
+		{
+			[Column(CanBeNull = true)] public int UnmappedField { get; set; }
+		}
+
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/5729")]
+		public void InsertUnmappedDerivedThroughBaseTable([IncludeDataSources(true, TestProvName.AllSQLite, TestProvName.AllSqlServer2016)] string context)
+		{
+			using var db = GetDataContext(context);
+			using var _  = db.CreateLocalTable<BaseClass>();
+
+			Assert.Throws<LinqToDBException>(
+				() => db.GetTable<BaseClass>().Insert(() => new UnmappedChild { Id = 1, Code = 1, UnmappedField = 5 }));
+
+			Assert.That(db.GetTable<BaseClass>().Count(), Is.Zero);
+		}
+
 		#endregion
 	}
 }
--- a/Tests/Linq/Linq/InheritanceTests.cs
+++ b/Tests/Linq/Linq/InheritanceTests.cs
@@ -1485,8 +1485,18 @@ namespace Tests.Linq
 				_ => new Child1 { Child1Field = 66 },
 				() => new Child1 { Id = 1 });
 
-			var result = db.GetTable<BaseClass>().OfType<Child1>().Single();
-			Assert.That(result.Child1Field, Is.EqualTo(55));
+			var inserted = db.GetTable<BaseClass>().OfType<Child1>().Single();
+			Assert.That(inserted.Child1Field, Is.EqualTo(55));
+
+			// The row now exists, so the second call takes the update branch and the derived column is
+			// exercised through the update setter as well as the insert one.
+			db.GetTable<BaseClass>().InsertOrUpdate(
+				() => new Child1 { Id = 1, Code = 1, Child1Field = 55 },
+				_ => new Child1 { Child1Field = 66 },
+				() => new Child1 { Id = 1 });
+
+			var updated = db.GetTable<BaseClass>().OfType<Child1>().Single();
+			Assert.That(updated.Child1Field, Is.EqualTo(66));
 		}
 
 		[Table("InheritanceFilterPositional")]
--- a/Tests/Linq/Linq/InheritanceTests.cs
+++ b/Tests/Linq/Linq/InheritanceTests.cs
@@ -1431,7 +1431,7 @@ namespace Tests.Linq
 		}
 
 		[Test(Description = "https://github.com/linq2db/linq2db/issues/5729")]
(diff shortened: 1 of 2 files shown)


Tests/Linq/Linq/InheritanceTests.cs
```diff
@@ -1430,6 +1430,133 @@ public void TestInsertIssue2([DataSources] string context)
 			}
 		}
 
+		[Test(Description = "https://github.com/linq2db/linq2db/issues/5729")]
+		public void Issue5729_InsertDerivedThroughBaseTable([IncludeDataSources(true, TestProvName.AllSQLite, TestProvName.AllSqlServer2016)] string context)
```
sdanyliv: I would like to do not see `Issue5729_` prefixes in new common tests.

MaceWindu: Done in `fef600f91` — the six tests are now `InsertDerivedThroughBaseTable`, `UpdateDerivedThroughBaseTable_Setter`, `UpdateDerivedThroughBaseTable_ExplicitTarget`, `InsertOrUpdateDerivedThroughBaseTable`, `InsertPositionalDerivedThroughBaseTable` and `UpdateWithOutputDefaultProjectionOverInheritanceRoot`. They keep their `[Test(Description = "…/issues/5729")]` links, which is how this fixture already ties a test to an issue.

Renamed straight away rather than at the end of review, because baselines key on the fully-qualified test name and never auto-prune — every CI run under the old names would leave orphans behind.

sdanyliv:
Reviewed and pushed seven commits (`5a34f1c222..6ebffb20e5`), all test-side except one comment fix. Two claims in the PR description are no longer accurate and are worth correcting there, since I can't edit the body.

**"Covers: all three `InsertOrUpdate` lambdas"** — two of three. The key selector is `() => new Child1 { Id = 1 }`, and `Id` is declared on `BaseClass`, so `EnsureDeclaringType` no-ops for it. It also can't carry a derived member: a key selector has to name key columns. The insert and update setters do reach the guard.

**"#5838 … `UpdateWithOutputDefaultProjectionOverInheritanceRoot` therefore asserts only that the failure is no longer `ArgumentException`"** — the premise doesn't hold at this head. That query raises nothing at all on SQLite.Classic, SQLite.MS, SqlServer.2019, SqlServer.2019.MS, PostgreSQL.16 and DuckDB, LinqService variants included, so the `catch` never fired and the test asserted nothing on any provider it ran on. The original setter shape was checked as a control in case dropping the redundant `Id = t.Id` had changed the outcome — it hadn't: with `Id = t.Id, Code = t.Code` on SQLite.MS and SqlServer.2019 the call returns one row whose `Deleted` and `Inserted` both materialize as `Child1` with `Id = 1`. This says nothing about #5838 itself, whose repro uses a different model and `InsertWithOutput`; it says this shape doesn't reach it. The test now asserts the row it actually gets.

Also from the same pass, recorded in the commits rather than here:

- `OUTPUT … INTO` had no test. It builds its target ref from the output table's object type, a route no other test takes, and it materializes no rows — so unlike the row-returning form it can assert success. That was SC-4's anchor after `U-2` narrowed.
- The provider gates were `AllSQLite, AllSqlServer2016`, and `AllSqlServer2016` is exactly 2016, not 2016-plus. Widened to `[DataSources]` / `[InsertOrUpdateDataSources]` / a `Feature*` constant. Widening surfaced that the update setters assigned the primary key to its own value, which YDB rejects outright; dropping those assignments keeps YDB in the matrix and leaves each setter naming only the member the fix is about.
- A control test for a subtype carrying no `[InheritanceMapping]` entry: the guard retypes for it too, and the failure stays loud — `LinqToDBException`, no INSERT emitted — rather than dropping the column.
- A `new`-shadowed mapped member now writes to the *shadowed* member's column silently, where `master` refuses the query with `ArgumentException`. Root cause is shared and outside this change; tracked as #5852, with a gated regression test added here.

122 cases green across Access.Ace.Odbc, SQLite.Classic, SQLite.MS, SqlServer.2019, SqlServer.2019.MS, PostgreSQL.16, Oracle.23.Managed, DuckDB and YDB. CI not re-triggered.

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- Writing a derived-type entity through a base-mapped table threw `ArgumentException: Property 'X' is not defined for type 'Base'`:

  ```csharp
  db.GetTable<BaseClass>().Insert(() => new Child1 { Id = 1, Code = 1, Child1Field = 11 });
  ```

  The derived member's column now reaches the SQL. This covers `Insert`, `Update(setter)`, `InsertOrUpdate` and constructor-based (record-style) projections. A subtype with no `[InheritanceMapping]` entry still fails, but loudly rather than dropping the column.
- Separately, `UpdateWithOutput` and `UpdateWithOutputInto` against an inheritance-mapped table now work without an explicit output expression. Previously any such call threw, whether or not a derived type was involved, because the default output projection builds whole `Deleted` / `Inserted` entities carrying every subtype's columns.
- Known limitation: a subtype that shadows an inherited member's name — `public new int Value` — is still written incorrectly, and does so silently rather than throwing. Tracked as [#5852](https://github.com/linq2db/linq2db/issues/5852).


Writing a derived-type entity through a base-mapped table now includes the derived member's column, and `UpdateWithOutput` works against an inheritance-mapped table.


---
<sub>Generated from commit `09153ed`.</sub>

PR #5845, opened 2026-09-01
CI: drop ClickHouse.Octonica from the ClickHouse test job

MaceWindu:
`clickhouse/clickhouse-server:latest` (`Build/Azure/scripts/clickhouse.sh`) moved to **26.8.2.7** and broke `Octonica.ClickHouseClient` 4.1.4: the native-protocol session (`:9000`) dies mid-test and every following command on that connection fails with

```
Octonica.ClickHouseClient.Exceptions.ClickHouseException : The connection is closed.
ErrorCode: 3
   at Octonica.ClickHouseClient.ClickHouseCommand.OpenSession(...)
```

A different test fails on each run, and the two automatic task retries do not clear it.

| build | finished | CH server | ClickHouse leg |
|---|---|---|---|
| [23125](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23125) | 2026-08-30 | 26.7.5.10 | green |
| [23133](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23133) | 2026-09-01 | 26.8.2.7 | 3 failures, all `ClickHouse.Octonica` |
| [23144](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23144) | 2026-09-01 | 26.8.2.7 | 1 failure, `ClickHouse.Octonica` |
| [23146](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23146) | 2026-09-01 | 26.8.2.7 | 1 failure, `ClickHouse.Octonica` |

`ClickHouse.Driver` (HTTP `:8123`) and `ClickHouse.MySql` (`:9004`) run as parallel lanes against the same server and stay green, so this drops only the Octonica lane from `Build/Azure/net{80,90,100}/clickhouse.json`.

Hotfix to unblock CI; root cause still under investigation. 4.1.4 is the newest `Octonica.ClickHouseClient` release (repo last pushed 2026-07-14), and its [#109](https://github.com/Octonica/ClickHouseClient/issues/109) is a prior instance of the client lagging a ClickHouse native-protocol change.

PR #5846, opened 2026-09-01
DIAGNOSTIC: isolate ClickHouse.Octonica against ClickHouse 26.8 (do not merge)

MaceWindu:
> **Diagnostic branch — do not merge.** Both changes exist only to get a signal out of CI; the fix, if any, goes on a separate PR.

Follow-up to #5845, which disabled `ClickHouse.Octonica` because ClickHouse 26.8.2.7 kills its native TCP session mid-run.

**What this branch changes**

1. `Build/Azure/net{80,90,100}/clickhouse.json` — `ClickHouse.Octonica` becomes the *only* provider. Driver and MySql are dropped, so the run is short and a three-lane contention effect can be distinguished from an Octonica-only defect.
2. `test-workflow-linux.yml` — after the ClickHouse leg, dump the server's view: `docker logs clickhouse`, `system.errors`, `system.query_log` exceptions, `version()`. No current step captures any of this, which is why the three red builds only ever showed the client side.

**What it is testing**

The client-side error (`ClickHouseException : The connection is closed.`, ErrorCode **3**) is a *secondary* symptom. In `Octonica.ClickHouseClient` 4.1.4, code 3 is only thrown when `ClickHouseConnection`'s `TcpClient` is already `null` (`ClickHouseConnection.cs:1041-1044`), which happens after a *previous* operation called `SetFailed`. The candidate path swallows its own exception:

- `ClickHouseDataReader.Close(disposing: true)` sends `Cancel` and then drains server messages to `EndOfStream`.
- The drain `switch` (`ClickHouseDataReader.cs:1240-1277`) handles only `Data`/`Totals`/`Extremes`/`ProfileEvents`/`Error`/`Progress`/`EndOfStream`/`ProfileInfo`; every other code falls to `default: throw new NotSupportedException`.
- `catch (Exception ex) { State = Broken; await _session.SetFailed(ex, false, async); if (disposing) return; }` — the exception is discarded.
- `SetFailed` → `ReleaseOnFailure` → `TryChangeConnectionState(..., Broken, null, ...)` nulls `TcpClient`, so the next command reports "The connection is closed." with no trace of the original failure.

Consistent with the CI evidence: in build 23146 the last query before the break logged `Rows Count: 0` (reader abandoned with data in flight — exactly the condition that runs cancel+drain), and every earlier query logged `Completed`.

Ruled out already: `Protocol.h` is byte-identical between v26.7.5.10 and v26.8.2.7 (no new server packet code), and the negotiated protocol revision is **54483 on both** servers (Octonica declares `CurrentRevision = 54483`, below every new gate in 26.8), so a correctly-gated protocol change cannot explain the difference. 26.8 does contain a large batch of "cancellation now interrupts promptly" fixes, which would change *when* packets arrive after a `Cancel` without changing the protocol.

The server-side dump is the discriminator: an abrupt client disconnect in `docker logs` points at Octonica's drain path; a server-side exception or resource error points the other way.


MaceWindu:
Closing — this branch did its job and must never merge (it strips the Driver and MySql lanes and patches the shared Linux workflow).

What it established:

- The failure is not contention between the three ClickHouse lanes: it still fails with `ClickHouse.Octonica` as the only provider.
- The server logs exactly one `NETWORK_ERROR (210) Broken pipe, while writing to socket` — the client hung up mid-result, the server did not drop anything.
- The real exception, caught by a `DOTNET_STARTUP_HOOKS` first-chance listener attached to the test host: `ClickHouseClient received 3 out-of-order bucket(s) in the server reply. This feature is not supported on the client side.` Octonica advertises protocol revision 54483, above its own `MinRevisionWithOutOfOrderBucketsInAggregation` (54480), then rejects the feature. Reproduces on ClickHouse 26.5.1 as readily as 26.8.2.7.

Fix and re-enable: #5860. Upstream report against Octonica/ClickHouseClient to follow.

PR #5847, opened 2026-09-01
PROBE: trace VisitConditional exit paths for #5737 (do not merge)

MaceWindu:
**Not for merge.** Temporary instrumentation to locate the mechanism behind a baseline change on #5737. The branch is deleted and this PR closed once the trace is captured.

## Why

On #5737 all 47 provider baselines for `IntervalTranslationTests.ConcatSurroundsADifferenceWithColumns` gain a duplicated literal column in every `UNION ALL` branch, and the set-operation discriminator is rewritten to reference it:

```diff
-			[t1].[Source] = 1 as [c1],
+			[t1].[Source] = [t1].[c1] as [c1],
 				SELECT
 					CAST(1 AS INTEGER) as [Source],
+					CAST(1 AS INTEGER) as [c1],
 				UNION ALL
 				SELECT
 					CAST(2 AS INTEGER) as [Source],
+					CAST(1 AS INTEGER) as [c1],
```

Results are unaffected (the literal is the same in every branch), but it is added SQL on every provider, and the in-code comment introduced by that PR names this exact test as the case its conditional re-visit is meant to prevent.

The shape does not reproduce locally under any configuration tried, each verified by byte comparison against both `linq2db.baselines` `master` and `baselines/pr_5737`:

| axis | tested | result |
| --- | --- | --- |
| provider set | 1, 4, and the full 7-provider SQLite lane | clean |
| cross-provider concurrency | 4 providers in one process | clean |
| compile configuration | `Debug`, `Azure` | clean |
| settings node | `NET100`, `NET100.Azure` | clean |
| suite scope | 1 test to a full `Tests.Linq` run (5316 baselines captured) | clean |
| TFM | net10.0, net8.0 | clean |

The `[Linux / SQLite (all providers)]` CI job is the only environment that has produced it, so the probe runs there.

## What this adds

One `Console.WriteLine` per `VisitConditional` exit, filtered to conditionals whose text mentions both `Source` and `Duration`, recording the exit taken, `_buildPurpose`, `_buildFlags`, and the runtime shapes of `test` / `ifTrue` / `ifFalse`. Lines are prefixed `L2DBPROBE|` for grepping out of the job log. Nothing else is changed.

For reference, the local (clean) trace for this query takes `client-side` with `revisitChanged=False` every time, never `deferred-tryconvert`, with the literal already inline in the test placeholder:

```
exit=after-children purpose=Sql flags=ForceDefaultIfEmpty, ForSetProjection, ForceOuter
  test=SqlPlaceholderExpression ifTrue=SqlGenericConstructorExpression ifFalse=SqlGenericConstructorExpression
  node=IIF((SQL(S:13): {t11.Source} == 1), construct ...)
```


MaceWindu:
Probe complete, closing. Result: **the PR's code does not emit the SQL the baselines PR shows.**

Two findings from the instrumented Linux SQLite leg ([build 23190](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23190)):

1. **The trace is identical to a local run.** 496 `L2DBPROBE` lines, only two exits reached (`after-children` and `client-side`, 248 each), no whole-expression conversion, and the discriminator's literal already inline in the test placeholder — `IIF((SQL(S:807453): {t807451.Source} == 1), construct …)`. `VisitConditional` is not where the duplicated column originates.

2. **This branch's own baselines don't contain the change.** `baselines/pr_5847` vs baselines master is 16 files / 48 insertions / 48 deletions across `JoinTests.LeftJoin5`, `QueryFilterTests.FilteredLeftJoinFallbackChainSurvivesSetOperation`, `SelectTests.TestConditionalInProjection` and `WhereTests.CheckField6` — all net-zero column reorders. `IntervalTranslationTests.ConcatSurroundsADifferenceWithColumns` is absent, and the test did run (its conditionals produced the probe lines above).

The likely provenance of the stale content: the Linux SQLite job now runs **.NET 10 only** after #5834 trimmed the matrix, while the 2026-08-31 run that wrote those baselines also ran .NET 8 / .NET 9 on Linux. net8.0, net9.0 and net10.0 are all clean on Windows, so it would be a Linux-only, now-retired leg.

Consequence for #5737: [linq2db.baselines#2047](https://github.com/linq2db/linq2db.baselines/pull/2047) carries 47 `ConcatSurroundsADifferenceWithColumns` files whose SQL the current code no longer produces. Re-running CI on #5737 should regenerate them away.

PR #5858, opened 2026-09-03
CI: build the generator before the solution, working around the CS2012 double-build race

MaceWindu:
The GitHub `build` workflow's **Build and pack** job fails intermittently on a clean build:

```
CSC : error CS2012: Cannot open '.build\obj\CodeGenerators\Release\CodeGenerators.dll' for writing
-- The process cannot access the file ... because it is being used by another process.
```

This is **dotnet/roslyn#77538** — reported from this repo in 2024, still open.

## Cause

`CodeGenerators` is built **twice concurrently**: once plain, and once as an inner build carrying an explicit `TargetFramework` global property, which is how the analyzer `ProjectReference` from multi-targeting `LinqToDB` resolves it —

```xml
<ProjectReference Include="../CodeGenerators/CodeGenerators.csproj"
                  OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```

Because the project is single-TFM, its obj path under `ArtifactsPath` carries **no target-framework segment** — `.build/obj/CodeGenerators/Release/CodeGenerators.dll` — so both instances write the identical file and one loses. (`LinqToDB` multi-targets, hence its `/net8.0/` segment; that asymmetry is visible in the error paths.)

The issue shows the SDK printing both instances:

```
CodeGenerators netstandard2.0 failed with 1 error(s) (6.8s)
  CSC : error CS2012: Cannot open '...\CodeGenerators.dll' for writing
CodeGenerators succeeded (5.2s) → .build\bin\CodeGenerators\Debug\CodeGenerators.dll
```

.NET triage put the cause in how this repo declares its generator dependency and moved it to Roslyn for guidance. Nothing has landed.

It is scheduling-dependent, so it fails on some runs and not others. Azure hits it far less often — 2 vCPU against a GitHub runner's 4 leaves much less parallelism for the two instances to overlap in — which is why it surfaced only once the DB-free checks moved to GitHub in #5836.

## Fix

Build `Source/CodeGenerators` on its own before the solution. Both instances then find it current and skip compiling, so no second write exists to race.

Deterministic where a retry would be probabilistic, and a few seconds against a possible second full solution build. It is really a targeted form of the workaround the issue already records — *"second build will succeed (if you don't remove .build)"* — since the reason a second build works is that the outputs are already there.

One constraint: the pre-build has to carry the **same properties** as the solution build, `VERSION_ARGS` in particular. `VersionSuffix` feeds the assembly version, and a mismatch would make the solution build recompile the generator, restoring the race.

Net change against master is a single added step.

## Rejected alternatives

All three were tried on this branch and are recorded in the step comment so they are not re-attempted:

- **`--no-build` on pack** — breaks it. `LinqToDB.CLI` packs a .NET tool with per-RID publishes that a plain build never produces, so pack failed with `MSB3030: Could not copy ... osx-x64\dotnet-linq2db.runtimeconfig.json`.
- **`-p:UseSharedCompilation=false`** — did not help. It only changed which racer lost: the locking process went from `VBCSCompiler` to `csc` and the build still failed, on `linq2db.dll` instead of `CodeGenerators.dll`.
- **Retrying the build once** — would work, since a second build succeeds, but it is probabilistic, hides the problem, and cannot be validated: on a run where the first build passes the retry never fires.

The underlying double-build is still worth fixing at the source, by changing how `Source/LinqToDB` references `Source/CodeGenerators`. That affects every consumer and analyzer delivery, so it belongs in its own change rather than in a CI workaround.

## Verification

The `build` workflow runs on `pull_request`, so the failing job runs here. Worth reading on the run: **Build code generators** succeeds, **Build solution** and **Pack solution** succeed, and **Verify nupkg sizes** / **Verify analyzer delivery metadata** still find packages under `.build/package/release`.

Because the failure is intermittent, one green run is not proof on its own — but unlike a retry, this change is falsifiable: if CS2012 appears again on `Build solution`, the up-to-date check is not holding and the reasoning above is wrong.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,36 +73,12 @@ jobs:
             'VERSION_ARGS=-p:VersionSuffix=-dev.${{ github.run_number }}' | Out-File -FilePath $env:GITHUB_ENV -Append
           }
 
-      # Build the Roslyn components on their own first, or a clean solution build intermittently
-      # loses this race:
-      #
-      #   CSC : error CS2012: Cannot open '.build\obj\CodeGenerators\Release\CodeGenerators.dll' for
-      #   writing -- ... because it is being used by another process.
-      #
-      # Each one gets built twice concurrently: once plain, and once as an inner build carrying an
-      # explicit TargetFramework global property, which is how a ProjectReference from
-      # multi-targeting LinqToDB resolves it. Being single-TFM, their obj path under ArtifactsPath
-      # has no target-framework segment - .build/obj/<name>/Release/<name>.dll - so both instances
-      # write the very same file and one of them loses. dotnet/roslyn#77538 shows the SDK printing
-      # both, one failed and one succeeded; .NET triage put the cause in how this repo declares the
-      # dependency, and it has been open since 2024.
-      #
-      # Building them up front means both instances find them current and skip compiling, so there
-      # is no second write left to race.
-      #
-      # This is the whole class, not one project: every Roslyn component here is single-TFM and
-      # reached from multi-TFM LinqToDB. CodeGenerators sets TargetFramework itself; the other two
-      # inherit it from Source/Analyzers.Common.props, which pins netstandard2.0 and clears the
-      # inherited TargetFrameworks list. A new Roslyn component belongs in this list - fixing only
-      # CodeGenerators just moved the failure onto LinqToDB.Analyzers.
-      #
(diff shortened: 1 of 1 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `2909661`.</sub>

PR #5859, opened 2026-09-03
CI: authenticate every baselines git read so the download throttle cannot kill a leg

MaceWindu:
## Problem

Three Linux legs on [build 23238](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23238) ([#5841](https://github.com/linq2db/linq2db/pull/5841)) failed in `Checkout test baselines`:

```
fatal: remote error: GitHub is temporarily limiting some unauthenticated downloads
to protect the stability of the platform. Please retry later or authenticate.
##[error]Bash exited with code '128'.
```

The clone already carried `BASELINES_GH_PAT`, and the token was inert. Git sends HTTP credentials only **after a 401 challenge**, and GitHub never challenges an anonymous read of a *public* repository — so every baselines read has been counting against the anonymous quota, and the `PublishTestResults` failures on those legs are downstream of the dead checkout.

## Fix

Two things are needed together at every baselines read site:

- `-c http.proactiveAuth=basic` — send the credential unprompted (git 2.46+; both Azure agent pools report git 2.55)
- a **complete** credential — `x-access-token:<token>@`, not `<token>@`, or git asks for the password it is missing

| file | sites |
|---|---|
| `test-workflow-linux.yml` | baselines clone, `$repoUrl`, `GITHUB_TOKEN` mapped onto *Setup tests* |
| `test-workflow-windows.yml` | baselines clone, `$repoUrl` |
| `ensure-baselines-branch.ps1` | url, `ls-remote` (runs on every build), rebase-path clone |
| `hana2.sh` | `linq2db.ci` clone — carried no token at all |

`git clone -c` writes the setting into the clone's config, so the `git fetch origin` in the push-retry loop authenticates too. `hana2.sh` keeps an anonymous fallback when `GITHUB_TOKEN` is unset, so it still runs outside the pipeline; it is the only setup script that clones, which is why the token is mapped on the Linux *Setup tests* step only.

Pushes were never affected — `git-receive-pack` does issue a 401, so the credential was already being used there.

## Verification

Probed against the real repository:

| command | result |
|---|---|
| `git ls-remote https://<bogus-token>@github.com/linq2db/linq2db.baselines.git` | **succeeds** — credentials never sent |
| `git -c http.proactiveAuth=basic ls-remote https://x-access-token:<bogus-token>@…` | `remote: Invalid username or token` |

`git clone -c http.proactiveAuth=basic` was also confirmed to persist the setting into the new repository's config.

With `proactiveAuth` a read cannot silently fall back to anonymous, so a **successful** `Checkout test baselines` step is itself evidence the token is being sent and accepted — a wrong credential form fails loudly with `Invalid username or token` instead of quietly going anonymous.

Both templates parse as YAML, `bash -n` is clean on `hana2.sh`, and the PowerShell parser is clean on `ensure-baselines-branch.ps1`. A pipeline template cannot be exercised locally, so CI on this PR is the gate.

### DIFF
--- a/Build/Azure/pipelines/templates/test-workflow-linux.yml
+++ b/Build/Azure/pipelines/templates/test-workflow-linux.yml
@@ -68,7 +68,14 @@ steps:
     # two root files instead of 343541, and --depth 1 --single-branch skips ~420 Mb of history for a
     # tree the leg only appends to. Master is cloned rather than the run branch, which may not
     # exist yet - see the push below. The commit step pairs this with git add --sparse.
-    script: 'git clone --depth 1 --single-branch --sparse --branch $(baselines_master) https://$(BASELINES_GH_PAT)@github.com/linq2db/linq2db.baselines.git baselines'
+    #
+    # http.proactiveAuth is what makes the token in the url count: git sends credentials only after a
+    # 401 challenge, and github never challenges an anonymous read of a public repository - so without
+    # it this clone is unauthenticated and github's unauthenticated-download throttle kills the leg
+    # (three legs on build 23238). The credential also has to be complete - user:token, not token
+    # alone, or git asks for the password it is missing. -c is written to the clone's config, so the
+    # fetch in the push retry below authenticates too.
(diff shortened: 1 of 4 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `6570899`.</sub>

Issue #5862, opened 2026-09-03
Source the SAP HANA provider from SAP's NuGet packages and shrink Redist/SapHana


MaceWindu:
### Summary

SAP now publishes the HANA .NET provider on nuget.org as `Sap.Data.Hana.Net.v6.0`, `Sap.Data.Hana.Net.v8.0` and `Sap.Data.Hana.Net.v10.0`, including the native `libadonetHDB` libraries for Windows, Linux and macOS. The .NET-side binaries currently checked into `Redist/SapHana` can be restored from those packages instead. The .NET Framework `Sap.Data.Hana.v4.5.dll` has no package equivalent and is tracked separately.

### Background

`Sap.Data.Hana.Net.v8.0` 2.29.27 (author `SAP`) contains `lib/net8.0/Sap.Data.Hana.Net.v8.0.dll`, `runtimes/{win-x64,win-x86,linux-x64,linux-arm64,linux-musl-x64,osx-x64,osx-arm64}/native/libadonetHDB.*` and `licenses/developer-license-3_2.txt`. `v6.0` and `v10.0` exist at the same version; there is no `v9.0` package, so a net9.0 target has to consume the `v8.0` one — the same per-TFM conditioning `Directory.Packages.props` already applies to `Net.IBM.Data.Db2`. Neither a `v4.5` nor an `hdbsql` package is published.

`Redist/SapHana` holds roughly 60 MB of checked-in binaries, most of it the natives and their PDBs.

### Current behavior

- `NuGet/NuGet.csproj:47` — `HintPath` to `Redist/SapHana/v4.5/Sap.Data.Hana.v4.5.dll`; packed into `linq2db.SapHana` and `linq2db.t4models`.
- `Tests/Linq/Tests.csproj:16,29,55,58` and `Tests/Tests.Playground/Tests.Playground.csproj:20,26,31,34` — reference both `v4.5` and `v8.0`, and copy `libadonetHDB.dll` / `libadonetHDB32.dll` to output.
- `Data/Setup Scripts/saphana2.cmd:17` — runs `Redist/SapHana/hdbsql.exe`.

### Expected behavior

[...]

Issue #5861, opened 2026-09-03
Source SQL Server Compact from its NuGet package and retire Redist/SqlCe


MaceWindu:
### Summary

The SQL Server Compact binaries redistributed by `linq2db.SqlCe` and `linq2db.t4models` are hand-copied into `Redist/SqlCe` rather than restored from a package. They are byte-identical to the contents of the `Microsoft.SqlServer.Compact` NuGet package, which also carries the licence terms the hand-copy left behind. Referencing the package instead removes the checked-in binaries and lets the licence and version travel with them.

### Background

Verified while inventorying redistributed binaries for #5731. [`Microsoft.SqlServer.Compact`](https://www.nuget.org/packages/Microsoft.SqlServer.Compact) 4.0.8876.1 contains `lib/net40/System.Data.SqlServerCe.dll` plus `NativeBinaries/{amd64,x86}/**` — the same layout as `Redist/SqlCe` — and `SQLCE_EULA_ENU.rtf`, with `requireLicenseAcceptance: true`.

SHA-256 comparison of five files against the package, all identical:

| File | Bytes |
|---|---|
| `System.Data.SqlServerCe.dll` | 470240 |
| `amd64/sqlceca40.dll` | 463584 |
| `amd64/sqlcese40.dll` | 561888 |
| `amd64/Microsoft.VC90.CRT/msvcr90.dll` | 624448 |
| `x86/sqlceca40.dll` | 349920 |

### Current behavior

- `NuGet/NuGet.csproj:31-44` — references `Redist/SqlCe/System.Data.SqlServerCe.dll` via `HintPath` and copies the `amd64/` and `x86/` native folders to the build output.
- `NuGet/SqlCe/linq2db.SqlCe.csproj:36-37` and `NuGet/t4models/linq2db.t4models.csproj:98,127` — pack those files into `tools/`.
- `Redist/Readme.md` — describes the folder; the SqlCe entry is one of only two whose contents are actually redistributed.

### Expected behavior

`NuGet/NuGet.csproj` restores `Microsoft.SqlServer.Compact` and packs the files from the package, `Redist/SqlCe` is deleted, and the licence text in `Build/licenses/texts/` is sourced from the package rather than converted by hand. `Build/licenses/components.json` should then carry the component's version automatically through `-Action harvest`, as it does for every other package-backed component.

### Notes

[...]

PR #5863, opened 2026-09-03
Add third-party license notices to the packages that bundle third-party binaries

### LINKED ISSUE #5731: Add third-party license notices to the linq2db CLI tool package
### Summary

Audit the third-party assemblies bundled into the `linq2db.cli` tool package and include the required license texts, notices, and attributions in the published package. The package should clearly distinguish the linq2db MIT license from the terms that apply to redistributed dependencies.

### Current behavior

- `Source/LinqToDB.CLI/LinqToDB.CLI.csproj:22` publishes the CLI as a .NET tool.
- `Source/LinqToDB.CLI/LinqToDB.CLI.csproj:42` declares `PackageLicenseExpression` as `MIT`.
- `Source/LinqToDB.CLI/LinqToDB.CLI.csproj:62-84` references the runtime, MCP SDK, and database-provider packages whose assemblies are bundled into the tool output.
- `ModelContextProtocol` 1.4.0 declares `Apache-2.0` in its NuGet package metadata. Other bundled provider packages also use licenses other than MIT.

No consolidated third-party notices file is currently included in the tool package. The package license expression describes linq2db itself but does not expose the additional terms attached to redistributed binaries.

### Expected behavior

The published `linq2db.cli` package should include a generated or maintained third-party notices artifact covering every bundled dependency that requires attribution, notice preservation, or license-text redistribution.

The implementation should:

- inventory the assemblies physically shipped for every supported tool RID;
- map each bundled assembly to its package and license metadata;
- include required license and NOTICE content in the packed tool;
- document that the MIT expression applies to linq2db code while bundled third-party components remain under their respective licenses;
- add packaging validation that detects newly bundled dependencies missing from the notices inventory.

### Notes

This is an existing package-wide concern rather than an MCP-only issue. PR #5678 adds `ModelContextProtocol`, but the same audit must cover the already bundled database providers and runtime dependencies so the result is complete and maintainable.

MaceWindu:
Fixes #5731

## What this is

Three artifact families physically redistribute binaries linq2db does not own, under a package license expression that covers only linq2db's own code:

| Artifact | How the foreign binaries get in |
|---|---|
| `linq2db.cli` (pointer + 7 RID sub-packages) | `PackAsTool` publishes the whole dependency closure into `tools/<tfm>/<rid>/` — **158 assemblies from 90 NuGet packages** per RID, against the 22 `PackageReference` lines in the project |
| the 14 T4 packages | each `NuGet/*/linq2db.*.csproj` copies named provider assemblies into `tools/`, including SQL CE and the SAP HANA client out of `Redist/`, and IBM's 223-file `clidriver` tree |
| `linq2db.LINQPad.lpx` | `Pack.cmd` archives the entire `net472` output directory — 172 binaries, including `linq2db4iSeries` |

Everything else was clean and is now *asserted* to be clean rather than assumed: `linq2db`, `.Tools`, `.Scaffold`, `.Compat`, `.Extensions`, `.FSharp`, the six `.Remote.*`, the four EF ids, `linq2db.Analyzers` and the `linq2db.LINQPad` nupkg must contain no third-party binary, and CI fails if one acquires it.

## How it works

`Build/licenses/components.json` inventories **118 components**; `Build/licenses/texts/` holds 23 license bodies; `Build/licenses/generated/` holds the **16 notices files** that ship — tracked in git, so the exact text that goes out is reviewable in the PR that changes it, the same reasoning that keeps `CompatibilitySuppressions.xml` and `PublicAPI.*.txt` in the tree.

`Build/Azure/scripts/third-party-notices.ps1` has four actions:

- `generate` renders the notices from the manifest.
- `check` re-renders and byte-compares, catching a hand-edit or an un-regenerated manifest, plus an orphaned or missing license text.
- `verify` opens the produced `.nupkg` / `.lpx` / publish output and asserts every shipped binary maps to a component scoped to that artifact, that the notices file is present, and that the excluded packages bundle nothing.
- `harvest` reads restore graphs and proposes what a dependency bump changed. It never writes the manifest.

`verify` reads the **produced artifact**, never the csproj. A csproj cannot tell you what a transitive dependency dragged in, and that is the case the gate exists for.

`Build/licenses/README.md` documents the manifest, the field traps, and how to add a component.

## Where it runs

Both pipelines pack on every PR, so both carry `check` + `verify`: `.github/workflows/build.yml` (which already runs the two sibling verifiers) and `Build/Azure/pipelines/templates/build-job.yml` (the only `dotnet pack` in the Azure tree, and the only job where the `.lpx` exists).

Release prep gets two hooks, as corpus changes rather than code: `/release-deps` grows a step 5b that restores the three graph-owning projects and harvests **after** the batch apply — before it, the ~75 transitive versions do not exist yet — and `/release-verify` step 4a runs `check` + `verify` beside the two existing packaging guards.

## Vendor terms

The audit found eleven components whose terms needed establishing. Ten resolved to `redistribution: permitted`, each on a clause quoted from the license the package itself ships, or on advice obtained for IBM (the ILAN grant is conditional — no standalone distribution, bundled in a value-added application, object code only — and linq2db meets all three).

Two are worth calling out:

- **`Microsoft.SqlServer.Types` 170.1000.7** ships the *SQL Server vNext Community Preview* pre-release terms: term ended 09/30/2022, and no `DISTRIBUTABLE CODE` section or distribution grant of any kind. `160.1000.6` of the same package carries ordinary SQL Server 2022 terms with an explicit third-party distribution grant, so this is a packaging error on a stable release. Reported to Microsoft. The notices reproduce the 160.1000.6 terms with the substitution stated in the component's `notes`, and the entry carries a `revisit` marker that `harvest` surfaces every release.
- **SAP HANA** stays `redistribution: unresolved`. SAP's own packages carry Developer License Agreement 3.2, whose §2(a) says the software may not be made available to any third party; our netfx copy came from a client install whose terms were never captured. That text now ships with its provenance stated. Decision deferred to #5862.

Two follow-up tasks came out of the audit: #5861 (source SQL Server Compact from its NuGet package — the `Redist/SqlCe` binaries are byte-identical to it — and retire the folder) and #5862 (the same for SAP HANA's .NET-side binaries, plus the netfx redistribution decision).

**No package's shipped binaries change in this PR.** The only additions to any artifact are the notices text files.

## Verification

Run against real artifacts — 14 T4 nupkgs, the CLI pointer and a packed `win-x64` RID sub-package, the built `.lpx`, and two publish outputs:

- `check` and `verify` exit 0 across all 17 inspectable artifacts, **0 unmapped binaries**.
- Notices placement confirmed by reading the archives: package root and `tools/` for the T4 set; root plus all three payloads (`tools/net8.0|net9.0|net10.0/win-x64/`) for the CLI RID package; root for the pointer; archive root for the `.lpx`.
- The per-TFM version check reads each payload's own `deps.json`: 89 packages on net10.0, 97 on net8.0 and net9.0.

Each gate was also proven able to fail:

| Control | Result |
|---|---|
| Add a `PackageReference` nothing else pulls in | `verify` exits 1 on both the publish output and the packed nupkg, naming the file in each of the three payloads; exits 0 after revert |
| Hand-edit a generated notices file | `check` exits 1 naming the file and both byte counts; exits 0 after regeneration |
| Set one TFM's version wrong | fails **only** that payload, across the 11 components sharing it; net8.0 and net10.0 stay clean |
| Drop an unreferenced license text in | `check` exits 1 naming it |
| Point at an artifact set it cannot enumerate | exits **2**, not 0 — so a silent no-op cannot pass |

No unit tests: this is packaging validation, and the repo's precedent for that is a CI script (`verify-nuget-sizes.ps1`, `verify-analyzer-delivery.ps1`). The controls above are the substitute.

`.gitattributes` pins `Build/licenses/generated/**` and `texts/**` to `eol=lf`; without it `* text=auto` checks them out CRLF on Windows and LF elsewhere and the byte comparison would disagree across platforms.

CI has not been triggered on this branch yet.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- `THIRD-PARTY-NOTICES.txt` now ships in every package that carries a third-party binary — the 14 T4 packages, `linq2db.cli` (the pointer package and all seven RID packages) and the LINQPad 5 `.lpx`. No shipped binary changed. If you redistribute these packages, two entries are worth reading: the SAP HANA provider's redistribution terms are unresolved ([#5862](https://github.com/linq2db/linq2db/issues/5862)), and the bundled `Microsoft.SqlServer.Types` 170.1000.7 carries expired pre-release licence text, so the notices reproduce 160.1000.6's terms instead. The core `linq2db` package and the `linq2db.LINQPad` nuget bundle no third-party binaries and carry no notices file.


`THIRD-PARTY-NOTICES.txt` now ships in every package that carries a third-party binary.


---
<sub>Generated from commit `7648c8d`.</sub>

PR #5864, opened 2026-09-03
Fix the three build 23263 failures: docker daemon retry, merge-source order, CharTest isolation

MaceWindu:
All three job failures in [build 23263](https://dev.azure.com/linq2db/linq2db/_build/results?buildId=23263) (on [#5737](https://github.com/linq2db/linq2db/pull/5737)) are pre-existing infrastructure/test defects, not regressions in that PR. Evidence came from the previous attempt's timeline, since the failed jobs were restarted.

#5737 needs no code change — only a master merge and a CI re-run once this lands.

## 1. `Win d_SqlServer2012` — the agent's docker daemon was down

Every call in `sqlserver.2012.cmd` answered *"failed to connect to the docker API … check if the path is correct and if the daemon is running"*, so `docker run` created nothing and the `max=100` / `sleep 1` wait loop then spent 106 s asking a container that did not exist for a row. Every other Windows docker leg in that build was green, so it was a per-agent flake.

All 8 `sqlserver.*.cmd` carried the same hand-rolled loop, which could not tell apart the three ways the setup fails and recovered from none of them:

- the daemon is unreachable;
- the container died during startup (`firebird.sh` already detects this and fails fast with the logs);
- the container was never created, or came up wedged.

New `Build/Azure/scripts/mssql-container.ps1` probes the daemon first (and gives a stopped `docker` service one push), then retries create-and-wait up to 3 times, with `docker rm -f` ahead of each attempt so a retry cannot collide on the container name, and a `State.Running` check so a crashed startup is not waited out for the full budget. All of a leg's containers still start before any of them is waited on, so the merged 2017+2019 and 2022+2025 legs keep initialising both servers concurrently.

Worst-case step budget is ~7 min (120 s daemon wait + 3 × 100 s readiness) against ~110 s today; those three numbers are documented in the script header. No pipeline file changes — `script_win_global` still points at the same `.cmd`, and the scripts artifact is assembled by a wholesale `xcopy`, so the new `.ps1` ships automatically.

## 2. `Lin k_Firebird` — unordered materialized merge source

`MergeTests.DeleteReservedAndCaseNamesFromList("Firebird.4.LinqService")` failed in TearDown with *"Baselines for remote context doesn't match direct access baselines"*.

The merge source is `GetSource2(db).ToList()` — unordered. `TestMerge2` is read with no `ORDER BY` and that row order flows straight into the generated `MERGE … USING (SELECT … UNION ALL …)`. `PrepareData` deletes and re-inserts before every case, so the direct and the remote run of one provider can materialize different orders, and `BaselinesWriter` compares the second capture against the first byte-for-byte.

`7c3adc075` ("fix unstable tests/baselines") already added `.OrderBy` after `.ToList()` to **seven** sibling merge tests. **Six** sites were missed; this adds the same call to all of them, so every materialized merge source in the suite is now ordered.

## 3. `Lin s_SQLite` — `CharTest` compared two different databases

`CharTest11("SQLite.Classic.MPU")` failed with `Expected: property Count equal to 4 / But was: 3`.

The three `CharTest` methods built their expected list from `new DataConnection()` — the default configuration, which `DataProviders.json` pins to `SQLite.MS` on every leg — and compared it against the context under test. The captured trace shows the two sides hitting different databases:

| side | traced context | rows |
|---|---|---|
| expected | `-- SQLite.MS SQLite`, `SELECT … FROM [Person]` | 5 (4 male) |
| actual | `-- SQLite.Classic.MPU …`, `WHERE [Gender] = 'M'` | 3 |

Those are separate files (`TestData.MS.sqlite` and `TestData.MiniProfiler.Unmapped.sqlite`) and the pristine seed is 4 rows / 3 male, so the default database was carrying a transient extra row written by a concurrent lane. It only bites on the SQLite leg — the one leg where the default provider is itself under test and therefore being mutated.

These now use `TestBase.Person`, the cached reference list every other reference-data comparison in the suite already goes through, which removes the cross-database read and the per-test round trip with it.

## Verification

**Item 1** — the `win-mssql-*` images can't run off a Windows agent, so the helper was exercised against a local SQL Server image on all four paths: unreachable daemon (bogus `DOCKER_HOST`), a container that never answers (3 attempts, `rm -f` + re-create each time, logs dumped, exit 1), the happy path, and a re-run over an already-existing container with two comma-separated specs — the last driven through a real `cmd /c` so the `cmd → pwsh -File` hand-off is proven rather than assumed.

**Item 2** — deterministic red→green on Firebird.4:

| step | state | result |
|---|---|---|
| red | ordering removed, plus a probe reversing the source for the remote case only | 1 of 4 failed, `.sql.other` emitted |
| — | diff `.sql` vs `.sql.other` | 248 lines each, differing **only** in the `UNION ALL` operand order |
| green | `.OrderBy` restored, probe still active | 4/4, no `.other` |
| clean | probe removed, whole `MergeTests` fixture | 357/357 |

The `.other` diff is what settles the mechanism: trace misrouting would have shown missing or extra statements, not a pure reordering of one clause.

Of the 165 regenerated Firebird.4 baselines, **164 are byte-identical** to the committed ones and 0 are new. The one change is `UpdateWithConditionDeleteWithConditionUpdate`, where the `Concat`-appended synthetic `Id = 1` row moves from last to first — same rows, same columns.

**Item 3** — 27/27 passed on `SQLite.Classic`, `SQLite.MS` and `SQLite.Classic.MPU`; the discovered set confirms `CharTest11("SQLite.Classic.MPU")`, the exact CI failure, was included.

**Not verified locally:** 4 of the 7 merge sites need Firebird 5 (`NOT MATCHED BY SOURCE`) or Oracle, so they're covered only by CI here. They are mechanically identical one-line edits to the three that were exercised — and the riskiest of the seven *is* in the covered set, since `TestParametersInSourceEnumerableFirebird` asserts on generated SQL via `LastQuery` and still passes.

A Release build of `Tests/Linq` reported no diagnostics in any file this PR touches.

### DIFF
--- a/Build/Azure/scripts/mssql-container.ps1
+++ b/Build/Azure/scripts/mssql-container.ps1
@@ -1,0 +1,202 @@
+<#
+mssql-container.ps1 - bring up the SQL Server container(s) a Windows test leg needs, and keep
+retrying until they answer or the budget runs out. Called by every sqlserver.*.cmd, which then
+runs its own version-specific `docker exec ... sqlcmd -Q "..."` setup statements.
+
+Replaces the hand-rolled `docker run` + `set max=100` / `sleep 1` / `docker exec ... SELECT 1`
+loop those scripts each carried. That loop could not distinguish the three ways the setup fails,
+and recovered from none of them:
+
+  - the agent's docker daemon is not reachable at all. On build 23263 the Win d_SqlServer2012 leg
+    got "failed to connect to the docker API ... check if the path is correct and if the daemon is
+    running" from every single call, so `docker run` never created anything and the wait loop then
+    spent 106s asking a container that did not exist for a row. Every other Windows docker leg in
+    that build was green, so it is a per-agent flake, not a broken image.
+  - the container was created but died during startup. firebird.sh already checks for this
+    (docker inspect .State.Running) and fails fast with the logs; the Windows scripts waited out
+    the full budget instead.
+  - the container was never created, or came up wedged. Nothing re-created it, so a single lost
+    `docker run` was terminal for the leg.
+
+So: probe the daemon first (and give a stopped docker service one push), then treat "container up
+and answering" as the thing being retried rather than assuming one `docker run` is enough.
+`docker rm -f` before each attempt is what makes a retry idempotent - without it a second attempt
+dies on "name is already in use" whenever the first one did create the container.
+
+All of a leg's containers are started before any of them is waited on, so the merged 2017+2019
+and 2022+2025 legs keep initialising both servers concurrently the way they do today.
+
+Invoked from a .cmd as:
+
+  pwsh -NoProfile -ExecutionPolicy Bypass -File "%~dp0mssql-container.ps1" -Container "mssql2012|1412|linq2db/linq2db:win-mssql-2012"
+  if %errorlevel% NEQ 0 exit /b 1
+
+Parameters:
+  -Container          one or more `name|hostPort|image|sqlcmdExtraArgs` specs. The last field is
+                      optional and is appended to every readiness probe (the 2025 image ships
+                      mssql-tools18, whose sqlcmd needs -C to accept the self-signed cert).
+                      Several specs may be passed as separate array elements or comma-separated
+                      inside one string - `pwsh -File` hands array syntax through as a literal
+                      string, so the comma form is what a .cmd can actually rely on.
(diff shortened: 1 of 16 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `1497369`.</sub>

PR #5878, opened 2026-09-05
CI: base a test leg's baselines clone on the run branch, not master

MaceWindu:
A test leg clones baselines **master**, so the commit `push-baselines.ps1` creates is a delta against master. A baseline this PR changed in an earlier run and has since changed back is an *empty* delta there: it enters no commit, the earlier run's version stays on `baselines/pr_N`, and the baselines PR keeps showing SQL the branch no longer emits. No later run clears it — each one produces the same empty delta — so the entry survives until the release-time baselines reset.

Introduced in [#5819](https://github.com/linq2db/linq2db/pull/5819), which switched the leg from checking out the run branch to cloning master so that the branch would no longer have to exist up front (retiring `create_baselines_pr`). Before that the leg checked out the run branch and the revert was staged.

```diff
- git clone ... --branch $(baselines_branch) origin/$(baselines_branch)   # pre-#5819: delta vs the branch
+ git clone --depth 1 --single-branch --sparse ... --branch master        # today: delta vs master
```

## Observed on #5725

`GroupByTests.Max11/Max12` for `ClickHouse.Driver` / `ClickHouse.MySql` still carry the CASE-folded `MAX` written by the run that preceded the fold's removal, three ClickHouse legs later:

```sql
maxOrNull(CASE WHEN t1.ChildID > 20 THEN 1 ELSE 0 END)   -- on baselines/pr_5725
maxOrNull(t1.ChildID > 20)                               -- what the branch actually emits, = master
```

Those legs ran the full suite — build 23338's ClickHouse leg reports `total: 26975, failed: 0` and the main step filters on nothing but `TestCategory != SkipCI`. `AggregationTests.MinMaxOverBooleanExpression`, added by the same PR and exercising the same non-grouped `Max(bool)`, *was* corrected on the first run after the fix: a file master does not have is never an empty delta.

## Fix

`Build/CI/checkout-baselines.ps1` clones the run branch when it exists and master when it does not — the case `push-baselines.ps1`'s push resolves by creating the branch, and a branch that does not exist carries nothing to revert, so the fallback loses nothing. Every #5819 property of the clone is kept: `--depth 1 --single-branch --sparse -c http.proactiveAuth=basic`.

The three inline copies of the clone — both Azure templates and the GitHub Actions leg — call it instead, which also puts the reasoning in one place rather than three. `Build/CI` is `xcopy`'d wholesale into the test-scripts artifact, so no packaging change is needed, and every leg already extracts that artifact before this step.

`ensure-baselines-branch.ps1` matters more now, not less: it rebases a run branch that has fallen behind master *before* any leg starts, so a branch-based clone is never stale. Its header and two comments in `push-baselines.ps1` that asserted the master base are corrected.

## Verification

Against the real repositories, writing master's content for the `Max11` baseline into a leg-shaped clone and staging it the way `push-baselines.ps1` does:

| clone base | staged |
|---|---|
| `baselines/pr_5725` (branch exists) | `M …/Tests.Linq.GroupByTests.Max11(ClickHouse.Driver).sql` |
| `master` (branch absent — today's behaviour) | *nothing* |

The second row reproduces the bug; the first is the fix. The fallback path is exercised by the same probe, so both branches of the base selection are covered. CI is the real gate — a `test-all` run on this PR exercises the changed step on every leg.

## Not covered

The GitHub Actions path (`tests-comment.yml`) resolves the branch name but has no `ensure-baselines-branch` equivalent, so a GH leg can base its clone on a branch that is behind master. That introduces nothing — only files the leg writes are staged, and the branch was already behind — but it is a gap worth closing when that migration lands.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -266,23 +266,17 @@ jobs:
         run: chmod +x scripts/*.sh
 
       # Cloned to <leg root>/baselines because AzureConnectionStrings sets
-      # BaselinesPath: ./../../../baselines, resolved from the test assembly. Sparse and shallow: the
-      # leg only ever appends whole files, and the full tree is 343541 files. Master, not the run
-      # branch - the push is what creates that.
+      # BaselinesPath: ./../../../baselines, resolved from the test assembly. The script picks the
+      # base - the run branch when it exists, master otherwise - and says why it matters.
       - name: Checkout test baselines
         if: inputs.baselines_branch != ''
+        shell: pwsh
         env:
-          TOKEN: ${{ secrets.BASELINES_GH_PAT }}
+          GITHUB_TOKEN: ${{ secrets.BASELINES_GH_PAT }}
+          BRANCH: ${{ inputs.baselines_branch }}
         run: |
-          # Named, because an empty token clones as x-access-token:@... and git then fails on the
-          # password it is missing rather than on the thing that is actually wrong.
-          if [ -z "$TOKEN" ]; then
-            echo "::error::BASELINES_GH_PAT is not set, but baselines_branch was supplied - nothing could be pushed"
-            exit 1
-          fi
-          git clone --depth 1 --single-branch --sparse -c http.proactiveAuth=basic \
-            --branch "$BASELINES_MASTER" \
-            "https://x-access-token:$TOKEN@github.com/linq2db/linq2db.baselines.git" baselines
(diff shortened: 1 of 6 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `ea9b6cb`.</sub>

PR #5880, opened 2026-09-05
CI: bring the GitHub Linux legs to parity with Azure

MaceWindu:
Working branch for closing the gap between the GitHub Linux legs and Azure, and for splitting the legs between the two CIs. Started as an **empty commit on master**, so the first run compared both CIs on identical code — any difference is the environment, which a run on a real PR cannot establish.

## Commits

| | |
|---|---|
| `67f0352fb` | empty commit — **drop before merge** |
| `25c2bbd9c` | raise `open_cursors` on the Oracle containers; stop hiding why `DROP USER` failed |
| `7e293078f` | pin ClickHouse `max_threads=2` |
| `a98566bf7` | split the Linux legs — 11 to GitHub, 4 stay on Azure |

## What the runs established

The empty-commit baseline reproduced the three Oracle failures on **unmodified master**, so they are environment issues, not caused by any PR's changes. ClickHouse was cancelled mid-run in the baseline, so it has no baseline datapoint.

Root cause of the Oracle failures, from the log:

```
failed ContractLinqInsert("Oracle.19.Managed") (121ms)
  ORA-01920: user name 'c##sequence_schema' conflicts with another user or role name
  ORA-00604: error occurred at recursive SQL level 1
  ORA-01000: maximum open cursors exceeded
```

`DROP USER` failed with **ORA-01000**, `AK107Tests.CreateUser`'s bare `catch { }` swallowed it, and `CREATE USER` then failed with **ORA-01920**. The visible error named neither the cause nor the resource, and the same ORA-01000 is the headline failure on `t_Oracle1112` — one root cause across all three legs.

After both fixes: `q_Oracle2123` **passes**; `j_Oracle1819` still exhausts cursors **at 3000**, and ClickHouse's replicated-serialization error is gone but a dropped connection replaced it.

**The open question:** Azure passes at the default **300** cursors while GitHub exhausts **3000**. A 10x gap is not explained by 2x the cores, so "GitHub drives more throughput" is insufficient and something more structural differs.

## The split

Eleven legs move to GitHub; ClickHouse and the three Oracle legs stay on Azure until the above is understood — **ClickHouse has since moved too**, see *Follow-up commits* below. Both CIs read the same `gh_linux` marker in `test-matrix.yml`, so the partition cannot drift — verified mechanically that no leg is in both and none is dropped.

Two design points:

- **The split is a `testing.yml`-only parameter.** `default.yml` — the release and master run — does not pass it, so it still runs every leg on Azure. A flat per-entry flag would have silently shrunk release coverage the moment a leg moved.
- **The empty-selection guard had to be split.** It exists because build 23192 went green having run nothing, but `[oracle.all]` now legitimately selects only Azure-retained legs. The selector counts matches *before* the `gh_linux` filter and fails only when a surface matches no leg at all.

Verified by dispatching the branch directly, since `issue_comment` runs the default branch's copy:

| surface | outcome |
|---|---|
| `[oracle.all]` | `Select legs` **success** + notice "every leg for this surface runs on Azure" |
| `[nonexistent.x]` | `Select legs` **failure**, build skipped |

## Follow-up commits

`5c9df08e4` moves the ClickHouse linux leg to GitHub. `f7edd43b8` makes an empty-matrix job say so, after `4043579ee` reverted a first attempt that a probe refuted.

### What the 15-leg run established

Run [34038663621](https://github.com/linq2db/linq2db/actions/runs/34038663621) ran **every** linux leg on GitHub over this branch's code — `issue_comment` runs master's `tests.yml`, which has no `gh_linux` filter — so it measures the container fixes, not the split. 13 green, 2 red.

| leg | GitHub | detail |
|---|---|---|
| `b_ClickHouse` | pass, 67.0 min | no failures, no retry |
| `t_Oracle1112` | pass, 33m16s | the `open_cursors` fix holds |
| `q_Oracle2123` | fail | `NonNullableAverage("Oracle.21.Managed")` — expected `InvalidOperationException`, got **ORA-00942**. It passed at 10:24, so it is flaky, not fixed |
| `j_Oracle1819` | fail | `ContractLinqInsert("Oracle.18.Managed")` — **ORA-01000** at 3000 cursors; `TestDateTimeSQL("Oracle.19.Managed",False)` — **ORA-00942** |

### ClickHouse moves; the Oracle legs stay

Retaining ClickHouse rested on it failing here, and it does not. Nor is the failure that justified retaining it GitHub-specific: after the `max_threads` pin what remained was `Issue1813Test11("ClickHouse.Octonica")` dropping its connection, and **Azure build 23422 hit exactly that**, on this branch's own code. Azure is not a safe harbour for this leg, and GitHub runs it in 67.0 min against 72.1 / 72.8 / 74.6 / 77.7 / 73.8 / 74.1 over six Azure builds. The split is now **12 GitHub / 3 Azure**, Azure keeping only the Oracle legs.

**ORA-00942 is a new signature and it re-aims the follow-up.** A table vanishing mid-test is isolation damage, not cursor exhaustion, and it now appears in two different legs. That is what the unmeasured `nunitWorkers` difference — 4 on a GitHub runner against 2 on an Azure agent — would produce, so the `.runsettings` experiment is pointed at the right variable.

### An empty matrix reported a green job — and still does on master

Azure cannot build a matrix at runtime, so a job whose matrix selects nothing is emitted anyway, runs once with `$(title)` unexpanded, skips every step on its `and(variables.title, …)` guard, and succeeds. Build **23319** (a `test-clickhouse` run) carries a job named literally `Win`, result **succeeded**, every test step **skipped** — `b_ClickHouse` has no windows leg, so that matrix was empty. It took an agent slot and three SDK installs to report nothing.

This predates the branch. What the split adds is the linux side of it: every surface but `[all]` and `[oracle.all]` now has all its linux legs on GitHub.

**The first fix was wrong, and the probe caught it.** Requiring `ne(variables['title'], '')` in each job's own `condition` was gated on `test-sqlite`, chosen because it exercises both arms — populated windows matrix must run, empty linux matrix must skip. Build **23453** came back green in 52m53s with **no `Win` or `Lin` job record at all**, while `Build (win-x86)` and `Create Test Baselines Branch` both succeeded, so every dependency and `enabled` held. The populated windows matrix never ran `s_SQLite`.

So `variables['title']` is **empty when a job-level condition is evaluated** — it resolves at step level only, which is where the ~20 existing uses in `test-workflow-*.yml` sit. Applied to every test job, that condition turns any `test-all` into a green run that executes nothing: the very shape it was written to prevent. Reverted in `4043579ee`.

`f7edd43b8` fixes it at the level where the variable exists: one step per OS template, unguarded on `title`, `condition: and(eq(variables.title, ''), succeeded())`, so the timeline carries a named step explaining the state instead of a wall of skips. A job with nothing to do still pays for the *unguarded* setup steps — `Cleanup Junk` on linux most of all; guarding those is a separate change.

Re-run on build **23475**, the same two-armed probe. It discriminates rather than passing everywhere:

| arm | notice step | test steps |
|---|---|---|
| `Lin` — matrix empty | **succeeded** | all 30 skipped, 1m6s |
| `Win s_SQLite` — matrix populated | skipped | **ran** — `EF.Core Tests (NETFX): SQLite (all providers)`, baselines committed, results published |

The title expanded in the populated arm rather than rendering the literal `$(title)` that 23319 showed, and the notice step was skipped there — so the guard is genuinely selective, and `4043579ee` did restore the leg the first attempt had suppressed.

### Both halves of the split, measured

`test-all` on this head — Azure build **23500** and GitHub run **34149486258** — confirms the partition in both directions:

| | legs | result |
|---|---|---|
| Azure | **3**, exactly `j_Oracle1819` / `q_Oracle2123` / `t_Oracle1112`, down from 15 | all pass (47m38s / 40m0s / 37m46s), plus 13 windows legs and both CLI jobs |
| GitHub | 15 selected — see below — of which the **12 `gh_linux` legs all pass** | only `j_Oracle1819` and `t_Oracle1112` fail, both Azure-retained |

3 + 12 = 15: no leg runs twice, none is dropped.

`t_Oracle1112` failed here after passing the previous run, while `q_Oracle2123` did the reverse — so all three Oracle legs are independently flaky on GitHub with the same ORA-00942 signature, which is stronger evidence for retaining them than the single-sample reading this PR started from.

**The `tests [all]` check is expected red until this merges, and cannot be made green here.** `issue_comment` runs `tests-comment.yml` from the *default branch*, and that workflow calls `tests.yml` at the same ref — where the `gh_linux` filter does not exist yet (`git grep gh_linux origin/master -- .github/workflows/tests.yml` returns nothing). So GitHub currently selects all 15 legs, including the two this PR exists to move off it, and those two are the only failures. After merge the filter is in master, GitHub selects 12, and those 12 are green above. Note `tests [all]` is **not** a required context — master's protection requires only `build`, which passes.

## Still to do
- ~~**Drop the empty commit** before merging.~~ Not needed — linq2db squash-merges, so `67f0352fb` and its "Do not merge" message do not reach master.
- **Oracle** returns as follow-up. Next experiment: pin NUnit parallelism via `.runsettings` so GitHub matches Azure's `nunitWorkers=2`. Note `MaxParallelLanes` is **not** it — effective lanes are `min(maxLanes, distinct providers)`, which is 2 on both CIs for these legs.
- **Release runs stay on Azure** by decision; GitHub stays out of release entirely for now.


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `f8a9cde`.</sub>

PR #5884, opened 2026-09-06
Fix nested window functions when a projected window value is reused

### LINKED ISSUE #5867: ClickHouse window function regression in 6.4.0
### Description

In 6.4.0, some queries using multiple window functions generate invalid ClickHouse SQL. This worked correctly in 6.3.0.

The example below generates the following (invalid) SQL:

```
INSERT INTO Projection(I1) VALUES ...

SELECT
	toInt64(ROW_NUMBER() OVER (PARTITION BY toInt64(ROW_NUMBER() OVER ())))
FROM
	Projection x
```

### Reproduction Steps

```
[Test]
public async Task Test()
{
        .UseClickHouse(CONNECTION_STRING, ClickHouseProvider.Octonica);

    await using var connection = new DataConnection(options);

    await using var tempTable = await connection.CreateTempTableAsync<Projection>();

    var toInsert1 = new[]
    {
        new Projection(1),
        new Projection(2),
        new Projection(3)
    };

    await tempTable.BulkCopyAsync(toInsert1);

    var intermediateQuery = tempTable
        .Select(x => new
        {
            Value = x.I1,
            RowNumber1 = Sql.Ext.RowNumber().Over().ToValue()
        });

    var finalQuery = intermediateQuery
        .Select(x => new
        {
            RowNumber2 = Sql.Ext.RowNumber().Over()
                .PartitionBy(x.RowNumber1)
                .ToValue()
        });

    var results = await finalQuery.ToListAsync();
}

private sealed record Projection(int I1);
```

### Exception details

```
DB::Exception: Not found column ROW_NUMBER() OVER () in block __table1.I1 Int32 Int32(size = 0)

at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteDbDataReader(CommandBehavior behavior, Boolean ignoreProfileEvents, Boolean async, CancellationToken cancellationToken)
at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteDbDataReader(CommandBehavior behavior, Boolean ignoreProfileEvents, Boolean async, CancellationToken cancellationToken)
at Octonica.ClickHouseClient.ClickHouseCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
at LinqToDB.Data.DataConnection.ExecuteReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken) in /_/Source/LinqToDB/Data/DataConnection.Async.cs:line 696
at LinqToDB.Data.DataConnection.ExecuteReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken) in /_/Source/LinqToDB/Data/DataConnection.Async.cs:line 696
at LinqToDB.Data.DataConnection.ExecuteDataReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken) in /_/Source/LinqToDB/Data/DataConnection.Async.cs:line 745
at LinqToDB.Data.DataConnection.ExecuteDataReaderAsync(CommandBehavior commandBehavior, CancellationToken cancellationToken) in /_/Source/LinqToDB/Data/DataConnection.Async.cs:line 745
```

### Regression?

Regression in 6.4.0, worked in 6.3.0

### Linq To DB version

6.4.0

### .NET version

10.0.400

### Database

ClickHouse 26.4.1.2212

### Database provider

Octonica.ClickHouseClient 3.1.9

### Operating system

Windows

### Configuration

_No response_

sdanyliv:
Fixes #5867

`SubQueryContext.MakeExpression` handed a member's defining expression to the caller under `Expand`, and the `UpdateNesting` that follows can only lift placeholders. A window function projected by an inner `Select` therefore reached the parent as a bare expression and was rebuilt inside whatever the parent computed:

```sql
SELECT ROW_NUMBER() OVER (PARTITION BY ROW_NUMBER() OVER ())
FROM Projection x
```

which no provider accepts. Regression in 6.4.0 — before #5468 the legacy `Sql.Ext` window functions went through `ConvertExtension`, where the member reference survived and became a subquery column.

Such a member is now built as SQL and lifted into a column of the subquery. What has no SQL value of its own — a window definition among them — keeps expanding, so `UseWindow` still resolves it.

Tests go into the shared window fixtures: `WindowFunctionsTests.Combinations` for the `Sql.Window` API (fluent chain and `DefineWindow`/`UseWindow`), `AnalyticTests` for the legacy `Sql.Ext` spellings, including the reported empty `OVER ()`.


Source/LinqToDB/Internal/Linq/Builder/SubQueryContext.cs
```diff
@@ -84,6 +84,16 @@ public override Expression MakeExpression(Expression path, ProjectFlags flags)
 
 			var corrected = SequenceHelper.CorrectExpression(path, this, SubQuery);
 
+			// Expand hands a member's defining expression to the caller, and a value bound to this subquery's row set
+			// is then evaluated at the caller's level - a window function referenced from the parent lands inside
+			// whatever the parent computes. Materialize it as a column instead. What has no SQL value of its own, a
+			// window definition among them, keeps expanding below (#5867).
+			if (flags.IsExpand()
```
MaceWindu, line 92: **Major · MAJ001** — this branch turns off the KeyedQuery eager-load strategy for a query that requested it by name, on every captured provider, invisibly to CI.

`EagerLoadingStrategyKeyedQueryTests.Select_KeyedQuery_NestedChildProjectsGrandparentFieldFallback` calls `.WithKeyedLoadStrategy()`. Its innermost eager load on `master`:

```sql
FROM (VALUES (101,'Company1',1), (102,'Company1',1), ... ) k_1("Item1","Item2","Item3")
     INNER JOIN "Employee" d ON d."DepartmentId" = k_1."Item1"
```

and on this branch:

```sql
FROM (SELECT DISTINCT d."Id", c_1."Name", c_1."Id" as "Id_1"
      FROM "Company" c_1 INNER JOIN "Department" d ON d."CompanyId" = c_1."Id") m_1
     INNER JOIN "Employee" d_1 ON d_1."DepartmentId" = m_1."Id"
```

— the Default strategy, across all 43 captured providers, plus a `SELECT DISTINCT` re-scan of `Company ⋈ Department` the keyed form existed to avoid. The test asserts data rather than SQL, so it stays green and CI cannot see this.

**Measured mechanism.** I instrumented `HasOnlySimpleFilterDependencies`' verdict and the dependency set it sees, and ran two arms off one build differing only in whether this branch is compiled in (SQLite.Classic, 4/4 green in both):

| KeyedQuery attempt | branch absent | branch present |
|---|---|---|
| outer level (`previousKeys` empty) | declines — `Ref(…::Company).Id`, `Ref(…::Company).Name` | declines |
| nested level (`previousKeys` = 2) | **accepts** — `Ref(…::Department).Id`, `Ref(…::Company).Name` | **declines** — `SQL[Int32]{t109[40].Id}`, `SQL[String]{t100[35].Name}` |

`ProcessEagerLoadingKeyedQuery` calls `ExpandContexts` (`ExpressionBuilder.EagerLoadKeyedQuery.cs:58`), which reaches this method under `BuildPurpose.Expand`. The dependency now arrives as a bare `SqlPlaceholderExpression` instead of a `PropertyExpression` member chain, so `CollectDependencies` collects placeholders, `HasOnlySimpleFilterDependencies` — which matches a dependency appearing as an operand of a simple comparison in the child's filter predicates — no longer recognises it, `ComplexParentReference` is recorded, and the dispatcher steps down to Default silently.

Note the outer level already declined on `master`, so only the nested level changes. That is also the only keyed test whose baselines moved: the two-level sibling `Select_KeyedQuery_ChildProjectsParentFieldFallback` was already on the query strategy and is untouched.

**Is this intended?** Arguably it aligns the nested case with its flat sibling, which is what the test's own name suggests should happen. But it is unmentioned, it silently discards a strategy the caller opted into, and the step-down costs an extra parent re-scan. Worth a decision either way.

Fix: narrow the gate so it fires only where inlining is actually invalid — i.e. when the built SQL is or contains a window function — leaving other `Expand` paths on the pre-PR shape. If the fallback is acceptable instead, say so in the PR body and add an assertion to that test pinning the strategy actually used, so the next change to this seam is not silent.

Two things I checked so you don't have to. The branch also skips `SequenceHelper.CorrectTrackingPath`, `SequenceHelper.CorrectSelectQuery` and the `ReplaceContext(result, SubQuery, this)` remap that the fall-through tail applies and that `Expand` always reached before — `GetProjectFlags` sets one purpose flag per arm, so `Expand` never co-occurs with the flags guarding those. Applying all three is green (10/10 over this test plus the three new window tests) but does **not** restore the keyed strategy: the dependency is still a placeholder, because `CorrectTrackingPath` rewrites the tracking path, not the node type. So that is not a route to a fix.

Source/LinqToDB/Internal/Linq/Builder/SubQueryContext.cs
```diff
@@ -84,6 +84,16 @@ public override Expression MakeExpression(Expression path, ProjectFlags flags)
...
+				&& Builder.BuildSqlExpression(SubQuery, corrected) is SqlPlaceholderExpression sqlValue)
```
MaceWindu: **Minor · MIN002** — the speculative build runs on every member resolved under `Expand` through a `SubQueryContext` and is discarded whenever it is not a placeholder, after which line 97 builds the same expression again.

`BuildSqlExpression(context, expr)` is `BuildExpression(context, expr, BuildPurpose.Sql)` plus `UpdateNesting`, and inside the visitor `RegisterTranslatedSql` inserts the produced placeholder into `_translationCache`. So the discarded work leaves cache state behind.

**Measured** on `Issue2596Tests` / SQLite.Classic (4/4 green), tagging each entry into this branch by whether its result was used:

| | entries | `_translationCache` insertions |
|---|---|---|
| result used (branch taken) | 2 | 0 |
| result discarded (fell through) | 2 | **132** |

All four are un-nested, so half the entries into this code throw their result away, and every one of the 132 insertions happens under a discarded one. A separate arm instrumenting `GetAlreadyTranslated` found that **no** primed key is ever read from outside a speculative build — so those 132 entries serve nothing on the real build path.

Separately observed, route untraced: on YDB — the only provider that renders top-level column aliases — six `WindowFunctionsTests.*WithMultiplePartitionsWithDefineWindow` baselines go from `x."Value" as Value_1` / `x."Timestamp" as Timestamp_1` to `as wnd2_1` / `as wnd1_1`, named after the `let wnd1` / `let wnd2` window-definition variables, and `Issue2596Tests.TestLoadWithInfiniteLoop` goes from `a_Invoice.Id as Id_5` to `as Lines_1`, named after the eager-loaded collection member. In both the branch cannot have been taken — a window definition and a collection are not scalars — so the only new code that ran is this discarded call. I could not trace the path from the cache state to the alias, and I am not claiming one; the drift is cosmetic here, and I confirmed across all 134 modified baselines that no discarded build leaked an extra column anywhere.

Fix: pre-test `corrected` cheaply before probing — the scope here is a scalar window value, so restricting it to members whose type the mapping schema can map (and which are not eager-load or association members) would keep the probe off the paths whose result is always discarded, drop the cache insertions, and halve the added build work. If it must stay unconditional, the comment above is worth extending to say that a full SQL build runs on every expand, which it currently does not mention.

MaceWindu (commented):
> [!IMPORTANT]

## Review notes

- [x] The four red legs on the GitHub Actions `tests [all]` run are environment failures, not PR failures. `Lin j_Oracle1819` and `Lin q_Oracle2123` died on `ORA-01920: user name 'c##sequence_schema' conflicts with another user or role name`; `Lin t_Oracle1112` on `ORA-00604 / ORA-01000: maximum open cursors exceeded`; `Lin w_DB2InformixDuckDB` reported `total: 27278, failed: 0` and then `Test application process didn't exit gracefully, exit code is '139'`.
- [x] Pushed `b08dd44ee` to this branch — a test-only assertion so `LegacyAnalytic_WindowFunctionOverWindowFunctionColumn` cannot pass on an empty result set. Its sibling `WindowFunctionOverWindowFunctionColumn` already pins its row count; this matches it. Revert if unwanted.
- [ ] No completed full CI run has validated the code under review. The Azure `test-all` (build 23456) was still running at review time, and `b08dd44ee` supersedes it.
- [ ] Because the Oracle / DB2 / Informix / DuckDB legs failed, no baselines were captured for those providers this run — the recorded delta covers 40 providers, not the full matrix.

## Out-of-scope observations

- **`LambdaResolveVisitor`'s expand-then-build fallback may now have a dead second step.** `VisitMember` (`LambdaResolveVisitor.cs:31-53`) and `VisitMethodCall` (`:55-77`) already implement the same two-step shape this PR adds: `BuildExpandExpression(node)` first, and only if expansion changed nothing, `BuildSqlExpression(_context, node, _buildPurpose, BuildFlags.None)` keeping the result when it is a `SqlPlaceholderExpression`. For a member reached through a `SubQueryContext` under `Expand`, expansion now returns a placeholder itself, so `!Equals(expanded, node)` holds and control takes `return Visit(expanded)` — the second step no longer runs for those. The two differ in build purpose: those sites pass the caller's `_buildPurpose`, the new branch always builds under `Sql`. I could not construct a shape where that changes an outcome, so this is a note rather than a finding — but if the second step is dead for the common case, the duplicated logic is worth collapsing rather than leaving two copies to drift.

## Baselines

Pass **outstanding**, not declined. `baselines/pr_5884` currently holds 120 added / 134 modified / 0 deleted, from a single run (17:14–18:09 UTC) that postdates the reviewed commit, with 0 files inherited from unrelated merges. Of the modified set, `IntervalTranslationTests.DifferenceSurvivesAsASubqueryColumn` (44), `EagerLoadingTests.TestJoin` / `TestCardinality*` (10) and `OrmBattleTests.ComplexGroupingTest` (2) are column reorders and alias renumbering — inert; `AnalyticTests.TestFunctionsInSubquery` (28) is a pure column reorder inside the innermost subquery projection. The 43 `EagerLoadingStrategyKeyedQueryTests` files are MAJ001. Oracle / DB2 / Informix / DuckDB are missing per the note above, and the delta predates `b08dd44ee`, so this wants re-reading once a complete run lands.

Source/LinqToDB/Internal/Linq/Builder/SubQueryContext.cs
sdanyliv: Confirmed and fixed in b794849462.

I reproduced your measurement with the same instrumentation on the strategy verdict, `EagerLoadingStrategyKeyedQueryTests` / SQLite.Classic, two arms off one build differing only in the predicate:

| KeyedQuery attempt | without predicate | with predicate |
|---|---|---|
| `previousKeys=0 deps=1` (55x) | accepted | accepted |
| `previousKeys=0 deps=2` (12x) | declined | declined |
| `previousKeys=0 deps=2` (6x) | accepted | accepted |
| `previousKeys=0 deps=1` (2x) | declined | declined |
| **`previousKeys=2 deps=2` (2x)** | **declined** | **accepted** |

Both arms 79/79 green, which is your point about the test being unable to see this.

The fix is the narrowing you suggested: the gate now also requires `QueryHelper.ContainsAggregationOrWindowFunction(sqlValue.Sql)`, so it fires only where inlining is actually invalid and every other `Expand` path keeps its pre-PR shape. That helper is the same predicate `SelectQueryOptimizerVisitor` already uses to refuse merging a subquery whose column carries a window or aggregate, so it is not a new notion.

Worth recording that the predicate was in this branch earlier and was dropped precisely because the full suite stayed green without it. Your finding is why that was the wrong conclusion.

Gates on the fixed commit: full SQLite.Classic suite 11160/0, and `EagerLoadingStrategy*` + `AnalyticTests` + `WindowFunctions*` + `DynamicWindow*` on SQLite and DuckDB 1317/0.

Source/LinqToDB/Internal/Linq/Builder/SubQueryContext.cs
sdanyliv: Stands, and the MAJ001 fix makes it worse rather than better.

b794849462 adds `ContainsAggregationOrWindowFunction` to the gate, which narrows only the *use* of the result, not the probe. The speculative `BuildSqlExpression` still runs on every member resolved under `Expand` through a `SubQueryContext`, and now a member that produces a placeholder but no window or aggregate is discarded too. So your 2-used / 2-discarded split on `Issue2596Tests` can only move further toward discarded, and the 132 cache insertions under discarded builds are unaffected.

I have not implemented the pre-test. Two things I would want settled first, and I would rather ask than guess:

1. Your suggested shape restricts the probe to members whose type the mapping schema can map, excluding eager-load and association members. That is a second predicate guarding the first, and after MAJ001 I am wary of narrowing this seam on reasoning alone. If you want it, I would rather measure it the same way MAJ001 was measured than reason about it.

2. The alias drift you observed but did not trace is the part that bothers me more than the wasted work. `as wnd2_1` and `as Lines_1` on YDB are named after a window-definition variable and an eager-loaded collection, neither of which can have taken the branch. If the discarded build is what primes those names, a pre-test would silently change them again, and I would want to know the route before touching it rather than after.

Happy to take either: fix it in this PR with a measured arm, or file it so the seam is addressed with its own gates. Which do you prefer?

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

- A window function projected by an inner query and reused by the outer projection — typically as another window's `PARTITION BY` key — is now read as a column of the inner query instead of being nested inside the outer window function. Previously this produced SQL no database accepts. Regression in 6.4.0; both `Sql.Window` and the older `Sql.Ext` surface are affected.
- [Known limitation] When a query combines a keyed eager load with a window value projected by an inner query and reused by the parent, the keyed eager-loading strategy falls back to the default one. The rows returned are unaffected; the query plan and its performance are.


A window function projected by an inner query and reused by the outer projection no longer generates invalid SQL (regression in 6.4.0).


---
<sub>Generated from commit `6c75757`.</sub>

PR #5894, opened 2026-09-08
Prepare CLI 6.5.0 for MCP Registry publication

igor-tkachev:
Prepare linq2db.cli 6.5.0 for publication as io.github.linq2db/linq2db.cli in the official MCP Registry.

Add the ownership marker to the packaged README, a NuGet/dnx STDIO manifest with required --config and optional --profile and --enable-execute-tool arguments, and publication instructions. Registry publication follows the NuGet release containing the marker.

Validation: server.json passes the official 2025-12-11 JSON Schema; git diff --check passes. Metadata and documentation only; no runtime changes.

### DIFF
--- a/.github/workflows/publish-mcp.yml
+++ b/.github/workflows/publish-mcp.yml
@@ -16,6 +16,15 @@ name: publish-mcp
 on:
   release:
     types: [published]
+  # Recovery path. If the automatic run fails - nuget.org lag beyond the poll budget, a transient
+  # registry error - re-run it here rather than falling back to the interactive Owner device flow.
+  # The tag supplies the version; nothing else is read from the tag's tree, so dispatching from
+  # the default branch is correct.
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Release tag to publish, e.g. v6.5.0.
+        required: true
 
 permissions:
   contents: read
@@ -25,8 +34,8 @@ jobs:
   publish:
     name: Publish to MCP Registry
     runs-on: ubuntu-24.04
-    # A prerelease must not claim the registry version.
-    if: ${{ !github.event.release.prerelease }}
+    # A prerelease must not claim the registry version. A manual dispatch is always deliberate.
+    if: ${{ github.event_name == 'workflow_dispatch' || !github.event.release.prerelease }}
     timeout-minutes: 40
 
     env:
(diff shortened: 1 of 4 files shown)


MaceWindu:
Pushed four commits to this branch. Flagging them here rather than editing the description, which still reads "Metadata and documentation only" — the branch now carries a CI workflow.

**Why.** `server.json` has to pin a literal version (the 2025-12-11 schema rejects `"latest"` via `not: {const: "latest"}`), and nothing updated it or triggered registry publication. Checked rather than assumed: no reference to `server.json`, `mcp-publisher`, `mcp-name` or `io.github.linq2db` existed anywhere outside this PR, the version-bump procedure touches only `Directory.Build.props`, and the post-release procedure ended at closing the milestone. So the manifest would have gone stale at 6.6.0 and publication would have depended on someone remembering a runbook nothing links.

- `10bf85b8f` — adds `.github/workflows/publish-mcp.yml`, triggered on `release: published`. Derives both version fields from the tag, polls the `ReadmeUriTemplate/6.13.0` endpoint the registry's own validator reads until nuget.org serves the README carrying the marker, then publishes via `mcp-publisher login github-oidc`. `MCP-REGISTRY.md` rewritten to match.
- `371ab4898` — records `repository.id` (`1916228`), which the schema documents for repository-resurrection detection. No validator consumes it yet.
- `260fb31b4` — commits both version fields as `0.0.0`. The workflow populates them, so a real version in the committed file only sets a trap for the manual path: `0.0.0` fails loudly at the registry's package-existence check instead of silently re-registering a stale version.
- `6529f4b5e` — adds `workflow_dispatch` with a tag input, so a run that fails on nuget.org indexing lag is recoverable in CI rather than through the interactive device flow.

**Timing constraint.** A `release`-triggered workflow is read from the tree its tag points at, so this has to merge before the 6.5.0 release branch is cut. Otherwise 6.5.0 publishes by hand and the automation starts at 6.6.0.

**One correction to the original doc.** Publication does not require an organization **Owner**. That holds only for the interactive `login github` path; the registry's OIDC handler grants `io.github.<repository_owner>/*` directly from the token claim, with no org-role lookup and no secret. `MCP-REGISTRY.md` now scopes the Owner requirement to the manual fallback.

**Two things checked and deliberately left alone.** `--config` is `isRequired: true` in the manifest while the CLI treats it as optional — that is a client-prompt directive plus a documented narrowing, not a mismatch. And `description` is exactly 100 characters against the schema's `maxLength: 100`; the workflow's `validate` step catches any future overflow before anything is published.

The rest of the original metadata verified clean: the manifest passes the 2025-12-11 schema, the marker's placement and payload match the NuGet ownership rules including the boundary requirement, the packaged 6.4.0 README confirmed absent of the marker through the registry's own endpoint, and `dnx linq2db.cli@6.4.0 --yes -- --help` confirmed `dnx` resolves the per-RID pointer package, so `runtimeHint: dnx` is sound.

MaceWindu (approved):

MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

The CLI's MCP server is now published to the MCP Registry as `io.github.linq2db/linq2db.cli`, so MCP-aware clients can discover and install it without a hand-written configuration entry.


The linq2db CLI MCP server is now listed in the MCP Registry as `io.github.linq2db/linq2db.cli`.


---
<sub>Generated from commit `6529f4b`.</sub>

PR #5898, opened 2026-09-09
CI: run the docker-free Windows test legs on GitHub Actions

MaceWindu:
Phase 4a of the Azure → GitHub Actions test migration: the five Windows legs that need no docker container move to GitHub. The eight win-mssql legs stay on Azure.

| Leg | Setup | Arch |
|---|---|---|
| `g_SqlCE` | `sqlce.ps1` (MSI download) | x64 |
| `s_SQLite` | none | x64 |
| `r_Access_MDB` | none | x86 |
| `v_Access_ACE_Odbc` | `access.ace.ps1` | x86 |
| `x_Access_ACE_OleDb` | `access.ace.ps1` | x86 |

#5880 measured GitHub at 1.4–1.6x faster per leg and left Azure's critical path entirely Windows — build 23500's 13 Windows legs took 1h2m–1h12m each against a 67-min pole for GitHub's whole 12-leg Linux set. Thirteen legs over ten Azure slots is two waves; eight is one.

## This PR changes no existing run

`offload_windows_to_github` defaults to false and **nothing passes it yet**. Azure keeps running all thirteen Windows legs; the new GitHub job is reachable only by `workflow_dispatch`.

That is deliberate. `tests-comment.yml` runs the *default branch's* `tests.yml`, so between turning the flag on and merging the Windows job there is a window where those five legs run on neither CI, with nothing red to say so. The flip in `testing.yml` is a one-line follow-up PR once this is on master — the same order #5866 used for the Linux legs.

`default.yml` will never pass it: the release run keeps every leg on one CI. `build_x86_job` therefore stays on the Azure side.

## What is here

- `test-matrix.yml` — `gh_windows` on the five entries; `offload_windows_to_github` declared and forwarded; `gh_linux` and `gh_windows` documented in the property list (`gh_linux` only had an inline note in `test-jobs.yml`).
- `test-jobs.yml` — the same `not(and(…))` matrix skip the Linux job already carries.
- `Build/CI/run-provider-tests.ps1` — Windows counterpart of `run-provider-tests.sh`. Three differences the port could not copy: netfx is a TFM, the suites are apphost `.exe` rather than `dotnet <dll>`, and a leg may be x86.
- `tests.yml` — `prepare` emits a second matrix; `build` stages netfx x64 when a Windows leg is selected; new `build_x86` job; new `windows-tests` matrix job.

Two selection guards, both shown *firing* against mutated copies of the matrix rather than by reading: a `gh_windows` entry carrying an Azure template conditional key (whose selectors a stock YAML parser cannot see), and a selected Windows leg with no TFM enabled. Both would otherwise be a green run that tested nothing.

Artifacts are per TFM rather than Azure's single win-x86 one — `download-artifact` has no `itemPattern`, so one artifact would force every leg to take all four TFMs at once and lose the download-run-delete disk discipline.

## Unrelated CI blocker, fixed here

`Microsoft.SourceLink.GitHub` 10.0.301 -> 10.0.303, one line in `Directory.Packages.props`.

NuGet's vulnerability index picked up GHSA-23fw-v26w-5fgq after master's last build, so every restore in the repo now fails `NU1902` on the transitive `Microsoft.Build.Tasks.Git` 10.0.301 (flagged range 10.0.300-10.0.301). With audit warnings treated as errors that is a hard restore failure on every branch, not only this one - all six DB-free legs here died before compiling anything. 10.0.303 is the version the advisory names as patched for that band; verified against the restored graph, `Build.Tasks.Git`, `SourceLink.Common` and `SourceLink.GitHub` all resolve to it.
## Verification

Hand-dispatched against this branch:

| Surface | Result |
|---|---|
| `[access.all]` | 3 legs green. `r_Access_MDB` 28748 tests / 0 failures across 6 `.trx` |
| `[sqlce.all]` | green, 16115 tests / 0 failures across 6 `.trx` |
| `[sqlite.all]` | green. Windows leg is **netfx EF.Core only** (147 tests), main left to the Linux leg per `win_efcore_only` |
| `[oracle.all]` | green with every job skipped — control for the empty-matrix fix below |

Three defects were found and fixed during bring-up, each with both arms measured:

1. **GitHub evaluates `${{ … }}` anywhere in a workflow file**, including inside a Python heredoc comment. A comment quoting Azure's own `${{ if … }}` failed the whole workflow at dispatch with `Unrecognized named-value: 'if'`. The prefix is now assembled in Python.
2. **An empty matrix fails the run**, rather than producing no jobs — with no failed job to point at and no annotation on any check run, so every job reads green under a red run. Every matrix job is now gated on its own leg list being non-empty. `[oracle.all]` above is the red→green control.
3. **A native command's stdout joins its enclosing function's output stream**, so the suite's exit code came back as the last element of an array of log lines. Measured: the EF.Core suite never ran on any TFM that ran the main suite (4 of 6 expected `.trx`), and a main suite exiting 3 produced exit 0. Every native call now pipes to `Out-Host`. Against stub test apps carrying the real assembly names, pre-fix `main 3 → exit 0, efcore x0`; post-fix `main 3 → exit 3`, `main 0 → efcore x1`.

The x86 runtime step was verified with a control arm on a throwaway branch: without `setup-dotnet … architecture: x86`, `DOTNET_ROOT` stays `C:\Program Files\dotnet` and every x86 leg dies loading the x64 `hostfxr.dll` (`HRESULT: 0x800700C1`). With it, `DOTNET_ROOT: C:\Program Files\dotnet\x86` and 14484 net10 x86 tests pass.

One flake seen and not reproduced: `DataConnectionCommitTransactionAsync("Access.Jet.OleDb")` failed once with `Query must have at least one destination field`, then passed on `gh run rerun --failed` at the same commit. Across runs it alternates between Inconclusive-and-skipped and this failure. That re-run also confirms artifacts survive a partial re-run.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,5 @@
-# Provider tests on GitHub Actions - the Linux legs.
+# Provider tests on GitHub Actions - the Linux legs, and the Windows legs that need no docker
+# container (the win-mssql ones stay on Azure).
 #
 # Normally started by tests-comment.yml, when someone comments `/azp run test-<db>` on a PR. That path
 # calls this workflow rather than dispatching it, so `ref` can be refs/pull/<n>/merge - which
@@ -74,7 +75,10 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       linux: ${{ steps.select.outputs.linux }}
+      windows: ${{ steps.select.outputs.windows }}
       any: ${{ steps.select.outputs.any }}
+      windows_any: ${{ steps.select.outputs.windows_any }}
+      windows_x86: ${{ steps.select.outputs.windows_x86 }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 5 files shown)


MaceWindu:
## 📝 Release-notes draft

### Full release notes (wiki)

_(none)_


---
<sub>Generated from commit `f05995a`.</sub>

PR #5900, opened 2026-09-09
Fix silently truncated eager-loaded collections on Sybase

### LINKED ISSUE #5865: Sybase: element-form eager-loading preamble bypasses query validation and silently drops detail rows
### Summary

The element form of an eager-loading query — `.LoadWith(...).First()` — builds a detail-query
preamble that joins a derived table carrying `TOP`. Sybase mishandles that shape, so the eager-loaded
collection silently keeps only the first detail row. linq2db already has a provider flag describing
exactly this Sybase defect, but the preamble is not run through the validator that consumes it.

### Current behavior

The preamble emitted for `db.Parent.Where(p => p.ParentID == id).LoadWith(p => p.Children).First()`:

```sql
DECLARE @p Integer
SET     @p = 2

SELECT [m_1].[ParentID], [d].[ParentID], [d].[ChildID]
FROM ( SELECT TOP 1 [p].[ParentID] FROM [Parent] [p] WHERE [p].[ParentID] = @p ) [m_1]
     INNER JOIN [Child] [d] ON [m_1].[ParentID] = [d].[ParentID]
```

Against a parent with two children this returns `Rows Count: 1`. No exception is raised — the
collection is simply short. Parameters bind correctly, so this is not an argument-binding problem.

The flag that describes this defect, and the validation that should act on it:

- `Source/LinqToDB/Internal/SqlProvider/SqlProviderFlags.cs:87` — `IsJoinDerivedTableWithTakeInvalid`,
  documented as "Indicates that provider has issue with any JOIN to subquery which has TOP statement",
  with the remark "Currently use as workaround over Sybase bug"
- `Source/LinqToDB/Internal/DataProvider/Sybase/SybaseDataProvider.cs:56` — sets it to `true`
- `Source/LinqToDB/Internal/SqlQuery/Visitors/SqlQueryValidatorVisitor.cs:355` — consumes it and calls
  `SetInvalid(ErrorHelper.Sybase.Error_JoinToDerivedTableWithTakeInvalid)`

The preamble reaches Sybase without that check running, so the builder never falls back to another
eager-loading strategy.

### Expected behavior

The eager-load preamble goes through the same validation as the main query, so on a provider with
`IsJoinDerivedTableWithTakeInvalid` set the join-to-derived-table-with-TOP shape is rejected and a
different strategy is selected — rather than the query returning fewer rows than it should.

### Notes

Observed on CI build 23314, contexts `Sybase.Managed` and `Sybase.Managed.LinqService`, sync and
async, four failures on the same assertion.

Scope, measured in the same run: the queryable form (`LoadWithTest`, same `[DataSources]`) passes on
Sybase. Only the element form fails, consistent with `First()` being what introduces the `TOP`.

**Not verified:** whether a plain non-compiled `db.Parent.Where(...).LoadWith(p => p.Children).First()`
reproduces this. It likely does, since the preamble comes from the same builder, but it was not run —
Sybase was reached only through CI, not locally. Equally, "the preamble is not run through the
validator" is the explanation that best fits the observations; what is confirmed is that the flag is
set, the check exists, and the forbidden shape still reached Sybase. The construction path was not
traced to the point where validation should have been invoked.

`EagerLoadingStrategyKeyedQueryTests` already excludes `TestProvName.AllSybase` on around twenty
tests, so the keyed-query strategy is known-bad there. This path has neither an exclusion nor the
validation that would have caught it.

As a workaround, `CompileTests.ElementFormLoadWithTest` and `CompileTestsAsync.ElementFormLoadWithTest`
on the branch for #5842 carry `[DataSources(TestProvName.AllSybase)]` with a comment pointing here.

sdanyliv:
Fixes #5865.

## What was wrong

`db.Parent.Where(...).LoadWith(p => p.Children).First()` builds its detail preamble as
`parent.SelectDistinct().SelectMany(...)`, and the cloned parent still carries the `First()`'s `TOP 1`.
Sybase ASE applies a derived table's `TOP` to the **whole joined result**, so the eager-loaded
collection came back with one row and no error.

Confirmed at the database level with raw SQL against the container, with a control that fixes the
expected value empirically:

| shape | rows | correct |
|---|---|---|
| `FROM (SELECT TOP 1 …) m INNER JOIN CH c ON …` | **1** | 2 |
| control — same shape, `TOP` removed | 2 | 2 |
| `FROM (SELECT TOP 1 …) m` — no join | 1 | 1 |
| `FROM (SELECT TOP 1 …) m, CH c WHERE …` — comma join | **1** | 2 |
| `WHERE c.ParentId IN (SELECT TOP 1 …)` | — | `Incorrect syntax near the keyword 'TOP'` |

The plain, non-compiled form reproduces too — the issue lists that as unverified.

Two corrections to the issue's own account. The preamble **is** validated (`BuildPreambleQueryAttached`
→ `BuildQuery` → `SqlProviderHelper.IsValidQuery`); the *rule* did not describe the shape, because
`SqlQueryValidatorVisitor.VisitSqlJoinedTable` reads `element.Table.Source`, i.e. only the joined side.
And the exception a user sees is raised earlier still, from the build-time simulation in
`SelectManyBuilder` → `IsSupportedSubquery`, since Sybase has no APPLY support and the build runs twice.

## The fix

**`ResolveStrategy` remaps `Default` → `KeyedQuery`** when the query is limited and the provider
declares `IsJoinDerivedTableWithTakeInvalid` — beside the existing `CteUnion` → `KeyedQuery` remap. The
keyed strategy carries parent keys instead of joining, so the offending shape is never built. That is
what the issue asked for, and it turned out to be available: `KeyedQuery` works on Sybase, measured.
Its 37 blanket exclusions in `EagerLoadingStrategyKeyedQueryTests` sat beside `AllAccess` with no
comment and had never been tried; they are removed here — 76/76 green.

**The validator rule now covers the `FROM` side** and a multi-table `FROM`, and runs after the descent
under `IsValid`, so a source already refused for its own reason (an unsupported APPLY, `ORDER BY` in a
derived table) keeps reporting that reason. Queries no strategy can express now fail loudly instead of
returning short results.

Two approaches inside the optimizer were implemented and abandoned on measurement rather than taste:
tagging the parent with `DoNotRemove`, and guarding the merge in `IsMovingUpValid`. The second fires
and the bad shape still returns, because `OptimizeDistinct` then strips the `DISTINCT` off the
surviving wrapper — `IsLimitedToOneRecord` recurses into the limited inner query — after which the
wrapper is trivial and is inlined on a later pass. Holding the shape would take a second guard inside
`OptimizeDistinct`, i.e. two general simplifications specialised for one provider.

## Tests

`JoinFromLimited` / `CrossJoinFromLimited` cover the rule's two arms — a join, and a comma-`FROM` —
and are sensitive to the **defect**, not merely to its refusal: each reads two rows out of a
source limited to one. Proven by mutation: with the rule removed both fail on a row-count mismatch.
Each asserts the expected side is 2 rows, so the fixture fails loudly if the seed data ever drifts
into a state where a correct and a mangled plan agree.

`ElementFormLoadWithReturnsAllDetails` covers the plain element form; `LimitedFormLoadWithReturnsAllDetails`
is the control for the queryable form, which is correct on Sybase today and must stay so — it is what
makes the rule's depth-1 predicate a decision rather than an accident.

`Issue_SubQueryFilter1/2` and `Issue4596Test` declare the refusal. The first two are `EXISTS`-shaped,
where only emptiness matters and the defect is harmless; they returned correct data but never covered
the shape. `Issue4596Test` reaches `Default` because `KeyedQuery` declines it (`ComplexParentReference`)
and no strategy can express it on Sybase.

## Verification

- Full `Sybase.Managed` suite, before and after, diffed by test id: 8775/12 failures at the branch
  point (all pre-existing `ContainsTests`) → 8785/12, **diff empty**.
- Red arm observed for the right reason before the fix: `Has.Count.EqualTo(2)` → got 1, not an exception.
- 25/25 on `Sybase.Managed` + `SQLite.MS`, direct and `.LinqService`.
- `EagerLoadingStrategyKeyedQueryTests` on Sybase: 76/76.
- Control mutation: deleting the `IsJoinDerivedTableWithTakeInvalid` conjunct reddens all six SQLite
  cases; restoring it returns 8/8.
- `Release` builds clean on `net10.0` and `netstandard2.0`.

No single full-suite run contains both the fix and the un-gating — two attempts were lost to a known
Sybase stall unrelated to this branch (isolated: the affected fixture alone is 1007/1007, the newly
enabled one 76/76). CI is the first place all of it runs together.

Only `SybaseDataProvider` sets the flag, so no other provider's SQL can change.


MaceWindu:
Heads-up on an interaction with #5939 (6.5.1, `Fixes #5936`), since this PR removes the blanket `AllAccess` / `AllSybase` exclusions in `EagerLoadingStrategyKeyedQueryTests`.

#5939 stops `ProcessEagerLoadingKeyedQuery` taking the `Contains` rewrite when the child chain carries a `Take`/`Skip` — that rewrite flattens all parents into one result set, so the limit applied once to the whole batch. Those queries now take the correlated `SelectMany` + VALUES-JOIN path, which resolves to `CROSS APPLY` where supported and to `ROW_NUMBER() OVER (PARTITION BY fk …)` via `OptimizeApplyJoin` where it is not. Both measured — `SqlServer.2017` and `SQLite.Classic` respectively.

It adds three fixtures to that file, carrying the exclusions this PR deletes. On a provider with neither APPLY nor window functions, a per-parent limit is inexpressible under *any* strategy (`Default` included), so once both land those three would assert on exactly that case:

- `Select_KeyedQuery_AssociationDetailTakeIsPerParent`
- `Select_KeyedQuery_AssociationDetailSkipTakeIsPerParent`
- `Select_KeyedQuery_DetailTakeIsPerParent`

Whichever merges second will want to check them rather than assume the un-gating is uniform. No action needed here otherwise — flagging, not blocking.

Issue #5768, opened 2026-08-12
Add a non-WPF target so the driver can load on LINQPad 9 for macOS


bonza:
### Summary

`linq2db.LINQPad` 6.4.0 cannot be used on **LINQPad 9 for macOS**, because the driver assembly
has a hard dependency on WPF, which does not exist on non-Windows platforms.

The package installs without complaint (the `net8.0-windows7.0` output is published to a
platform-neutral `lib/net8.0` folder), so LINQPad offers the driver and lets a connection be
created — but the connection then fails permanently.

### Environment

| | |
|---|---|
| linq2db.LINQPad | 6.4.0 |
| LINQPad | 9.9.10 (macOS) |
| OS | macOS 26.5.2, Arm64 (Apple Silicon) |
| .NET runtime | 10.0.10 |
| Database | SQL Server (dynamic connection) |

### Steps to reproduce

1. On macOS, install the `linq2db.LINQPad` driver in LINQPad 9.
2. Add a connection → `Linq To DB` → `SqlServer`.
3. The connection appears in the tree but is marked with an error.

### Actual result

The connection node shows:

```
Error: Could not load file or assembly 'PresentationFramework, Version=8.0.0.0,
Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
```

(There is a second, separate failure reachable via **Add connection**, thrown from
`DCDriverLoader..ctor` before any driver code runs:

```
Could not load file or assembly 'System.Collections.Immutable, Version=10.0.0.0, ...'
The located assembly's manifest definition does not match the assembly reference.
```

That one is self-inflicted on the user side — the driver's dependency set pulls
`System.Collections.Immutable 10.0.3`, so the .NET 10 runtime must be installed.
Installing the .NET 10 runtime resolves it, and then the `PresentationFramework`
failure above is what remains.)

### Expected result

The driver loads and the connection works, as it does on Windows.

### Root cause

`Source/LinqToDB.LINQPad/LinqToDB.LINQPad.csproj`:

```xml
<TargetFrameworks>net472;net8.0-windows7.0</TargetFrameworks>
...
<UseWPF>true</UseWPF>
```

Inspecting the shipped `lib/net8.0/linq2db.LINQPad.dll` metadata confirms the dependency is real
and not merely a vestigial reference:

* **4** WPF assembly references — `PresentationFramework`, `PresentationCore`, `WindowsBase`, `System.Xaml`
[...]

PR #5942, opened 2026-09-18
Add .NET 11 support, drop .NET 8 and 9

MaceWindu:
Stage 0 of the managed-Access work: everything needed before [`LibRed.Ado`](https://www.nuget.org/packages/LibRed.Ado/) (net11.0-only) can be referenced at all. **Not for merge before .NET 11 RTM** — `global.json` pins an RC SDK.

Supported set becomes `net462`, `netstandard2.0`, `net10.0`, `net11.0`. .NET 11's release coincides with .NET 8 and 9 going EOL, so they go in the same change.

## Reducing the per-version edit cost

Adding a runtime used to mean editing a TFM list in ~10 project files, ~22 package rows, 25 matrix entries and a directory of ~27 CI configs. After this:

| Surface | Before | After |
|---|---|---|
| TFM lists | ~10 files | `Build/TargetFrameworks.props`, 2 properties |
| Package versions | 1 row per TFM per package (~66) | 2 rows per package (~24), via `$(BandLatest*)` |
| Test-matrix TFM flags | 3 flags x 25 entries | none — the flags were `true` everywhere |
| CI per-TFM steps | per-TFM rows | one `tfms` table row per engine |

The matrix-flag removal is behaviour-preserving and was checked, not assumed: the old rules over the old matrix and the new rules over the new one select identical legs, configs and per-leg TFM enablement across all 18 surfaces.

## Pre-release ecosystem workarounds

Each is commented at its site and marked for removal.

- **Npgsql EF Core 11** — `11.0.0-rc.1` hard-pins `Microsoft.EntityFrameworkCore 11.0.0-rc.1.26431.118`, never published ([npgsql/efcore.pg#3913](https://github.com/npgsql/efcore.pg/issues/3913)). NuGet treats a direct reference to the published build as a *downgrade*, not a conflict, so two pins plus a scoped `NoWarn` make it restore. Upstream reports `MissingMethod` at runtime under this workaround, so expect PostgreSQL EF11 noise.
- **MySQL EF Core 11** — does not exist (Pomelo stops at EF 9, the Microting fork at EF 10). The provider calls sit behind `EF_MYSQL`; the ADO driver has no gap and is referenced directly.
- **MA0001 / CA2016** — .NET 11 added `StringComparison` overloads for the char string searches and a `CancellationToken` overload for `TextWriter.WriteLineAsync`. Both analyzers report **real** findings on net11, but the argument is a compile error on every other TFM we target and `Meziantou.Polyfill` does not carry the overloads yet. Suppressed scoped to net11.0 only.

## Also here

- EF8/EF9 integrations replaced by EF11. Removing two shipped packages is consumer-facing and needs a release-notes line.
- Version-specific EF ifdefs become capability symbols (`EF10_OR_GREATER`, `EF_MYSQL`, `EF_FSHARP`), so a future EF version is a token in one project file.
- `FSharpTests.cs` was behind `#if EF8`, but the EF8 *test* project defined no constants — so it compiled into no assembly and never ran. That project was also the only consumer of the F# model project. Both repaired: the F# project is wired into the EF10 tests, which declare `EF_FSHARP`. The `ActiveIssue` gate stays.
- Roslyn floor moves 4.8 to 5.0 (lowest supported SDK's first band), and `Tests.Analyzers` follows off net8.0, retiring the .NET 8 runtime steps in CI.
- `CA2026` fixed rather than suppressed — `JsonDocument.Parse(...).RootElement` leaks a pooled buffer.

## Verification

- `dotnet build linq2db.slnx -c Debug` — 0 errors, 85 net11.0 outputs.
- `dotnet build linq2db.slnx -c Release` — 0 errors, 22 projects with net11.0 assemblies.
- `-p:RunApiAnalyzersDuringBuild=true` — no RS00xx, so the new `PublicAPI/net11.0` folders are correct.
- `third-party-notices.ps1 -Action check` — passes.

## Follow-up commit

Review items raised on the branch, plus the three red checks from the first CI run.

### Review items

- **Contributor docs.** `CONTRIBUTING.md`, `Build/Azure/README.md`, `Tests/EntityFrameworkCore/Tests/Providers.md` and two smaller READMEs. `Test.cmd` was not a doc problem — it still built net8.0/net9.0 and invoked the deleted EF8/EF9 test projects, so it was broken outright; its parameter list drops from five positions to four.
- **One test config per job, not one per job and TFM.** `Build/Azure/{netfx,net100,net110}` (68 files) fold into `Build/Azure/configs` (27). Each file holds a single TFM-agnostic `Azure.TestJob` section that `DataProviders.json` points every `<TFM>.Azure` configuration at, so a new .NET version costs **nothing** under `Build/Azure`. That also retires `run-provider-tests`' `--flag` / `-Flag` parameter, which existed only to name the bucket directory. `sqlserver.2005` is the one config that genuinely varies by framework, which needed `DisableRemoteContext` to survive `BasedOn` — the only `TestSettings` field missing from `SettingsReader.Merge`.
- **SAP HANA's generated config.** `hana2.sh` builds its config at run time and was still emitting `NET80` / `NET90` / `NET100` sections and **no `NET110`** — the net11 SAP HANA leg would have resolved zero providers and reported green. It now emits the same `Azure.TestJob` shape and needs no edit per .NET version either.
- **`EntityFrameworkCore.FSharp`.** Supersedes the *"wired into the EF10 tests"* line above: the fixture now runs on EF10 **and** EF11, over two model projects rather than one multi-targeted one — every `LinqToDB.EntityFrameworkCore.EFnn` publishes the same `PackageId`, so NuGet collapses them inside one restore graph and a TFM-conditional reference to two of them resolves to the wrong one (`NU1201`). The package's last release is **6.0.7, 2022-06-26**, built against EF Core 6; it registers an `IMemberTranslator` that calls `ISqlExpressionFactory.Convert(SqlExpression, Type, RelationalTypeMapping)`, removed in EF Core 10, so an option-typed member reaching the query pipeline throws `MissingMethodException`. Both `Issue4646` cases are `[ActiveIssue]`-gated on that; `TestLeftJoin` stays live because it never reaches a member translator.
- **Third-party notices.** `components.json` still named `net8.0` / `net9.0` and carried no `net11.0` entries, and those maps render straight into the generated notices — so every artifact's notices stated versions for frameworks the release no longer ships, and the packed-artifact gate would have failed on the missing entries. Refreshed from `-Action harvest`; `System.Net.ServerSentEvents` dropped as a net8.0/net9.0-only CLI transitive. `net8.0-windows7.0` stays: the LINQPad 5 driver still builds it.

### CI failures

| Check | Cause |
|---|---|
| Analyzer tests (both engines) | `Tests.Analyzers` moved to net10.0 but the verifiers still compiled snippets against `ReferenceAssemblies.Net.Net80`, so every test referencing linq2db hit `CS1705` |
| `Build and pack` — `MSB4018` on `linq2db.LINQPad.deps.json` | `LinqToDB.LINQPad.Pack.csproj` passed an **inert** `GenerateLpxArtifacts=false` global property on its `ProjectReference` — its only consumer is net472-only, while the reference pins net8.0-windows7.0 — which forked a second MSBuild instance of the same project and TFM. Both raced to write the deps file. Latent on master; this branch's shallower graph exposed it |
| `Lin s_SQLite` | `Issue4646TestEF`, the `EntityFrameworkCore.FSharp` translator above |

### Verification

- `dotnet test Tests/Tests.Analyzers -c Azure` — 173 passed, 0 failed.
- `dotnet build linq2db.slnx -c Release` — 0 errors.
- `SettingsReaderTests.SharedTestJobSection` — pins the shared-section resolution and the netfx `DisableRemoteContext` override.
- Matrix ↔ `Build/Azure/configs`: 27 config names, 27 files, no missing entries, no orphans; the 14 files `netfx` used to hold are exactly the 14 entries with `enable_fw_netfx: true`.
- `third-party-notices.ps1 -Action check` — passes.

### .NET 11 behaviour changes the net11 test leg surfaced

The net11 main suite had never actually run before this branch — the EF.Core suite failed first and the runner skips later TFMs — so its first green build produced 155 failures across 20 tests. Three distinct runtime/compiler changes, each measured on both runtimes rather than inferred:

| Change | Impact | Handled by |
|---|---|---|
| `double`→`decimal` converts exactly instead of rounding to 15 significant digits ([breaking change](https://learn.microsoft.com/dotnet/core/compatibility/core-libraries/11/decimal-biginteger-floating-point-conversions)) | 147 failures / 19 tests | `TestBase.AreEqualWithinDelta` — a **relative** 1e-12 tolerance, since one test asserts `2147483648.123m` (noise ~2e-7) while the rounding tests must still catch 1e-5 at magnitude 10 |
| `new[] { … }.Max()` binds to `MemoryExtensions` over an implicit `ReadOnlySpan` conversion | 8 failures / 1 test | Translator unwraps the span, exactly as `ExpressionBuildVisitor` already does for `Contains` (which moved in .NET 10). Only `Min` and `Max` moved; `Sum`, `Average`, `Count`, `Any`, `First`, `ElementAt` still bind to `Enumerable` |
| `Math.Round(double, digits)` changed at midpoints | 3 tests | Those tests now round the `decimal` instead — see below |

**`Math.Round` is a real divergence, not just a test problem.** `Math.Round(6.55d, 1)` returns `6.6` on .NET 10 and `6.5` on .NET 11; the `decimal` overload is unchanged at `6.6`. The nearest double to 6.55 is 6.5499999999999998, so .NET 11's answer is arguably the correct one. But linq2db translates `Math.Round` on a double to SQL `ROUND`, so **on .NET 11 a query mixing client-side and server-side rounding of doubles disagrees with itself at midpoints**. `Round4`, `Round4Sql` and `Round11` were changed to exercise the `decimal` overload rather than assert that divergence — the library behaviour is unchanged and this is not fixed here.

### Round 4 - triage of the first `[all]` matrix

Run `35396172098` (head `b77e890`) failed **712 cases across 8 legs**, from four causes. The four commits above address all of them.

**EF Core 11 has no MySQL provider** - 487 cases, `Lin p_MySQL`. `TestConfiguration.EFProviders` is EF-version-agnostic, so the EF11 leg offered MySQL configurations to `ContextTestBase.ProviderSetup`, whose MySQL arm sits behind `EF_MYSQL`; every case it reached failed with `ProviderSetup is not implemented for provider MySqlConnector.8.0`. `Tests/EntityFrameworkCore/Tests/Providers.md` already recorded that the net11.0 leg runs without MySQL - only the provider selection never followed.

**`double` -> `decimal` tolerance, second wave** - ~225 cases. `Round9`/`Round12` exclude SQLite, so the earlier green `[sqlite.all]` dispatch never reached them. Most of the remainder compare whole `LinqDataTypes` rows, and that type already compares `DateTimeValue` and `StringValue` leniently by design because it exists to compare provider round-trips; `MoneyValue` joins them, which is what closes the bulk without editing every call site. The rest take the tolerance where they assert.

**AdoNetCore.AseClient corrupts DECIMAL on .NET 11** - 94 cases on Sybase. Measured on both runtimes against the same container and query: `CAST(123.45 AS DECIMAL(10,2))` reads back as `123.45000000000000256981935710` from `GetValue`, `GetProviderSpecificValue`, `GetDecimal` and `GetString` alike, and as `123.45` on .NET 10. `GetDouble` is exact on both, and `MONEY` is unaffected. Wide values lose real precision rather than gaining noise - a `NUMERIC(38,10)` holding `12345678901234.5678` returns exactly that on .NET 10 and the double expansion on .NET 11 - so no read path preserves the value and no provider-side mitigation can recover it. Reported upstream as [DataAction/AdoNetCore.AseClient#240](https://github.com/DataAction/AdoNetCore.AseClient/issues/240).

Three of those failures are behavioural rather than comparison problems and are gated, each scoped to net11.0 and up so net462 and net10.0 keep running the test:

- `ToBit2` - the noise flips a client-side predicate, leaving the expected set empty; a tolerance cannot fix a predicate.
- `Issue42Test` - the round-tripped value makes linq2db declare `@MoneyValue Decimal(29, 28)`, the update does not take, and the re-read `BoolValue` is unchanged.
- `SybaseProviderSpecificReadMatrix` - it asserts the formatted value a provider returns, which is the corrupted string itself.

The two Informix failures with the same signature needed no gate: the tolerance covers them, verified against the local container on net11.0.

**YDB ROUND overflow** - 2 cases, and not a .NET 11 defect. `RoundScaled` scaled by `10^p` in the column's own declared type, so a `Decimal(6,2)` column overflowed both the product and the `10^p` literal itself. Fixed by widening the intermediate; `MathFunctionTests.Round13` pins it. The same fix ships standalone against master as #5943 for 6.5.1, so it will arrive here again at the next merge from master.

Local verification on this branch: the full `Sybase.Managed` suite on net11.0 is 8844 tests with no related failure, and YDB `MathFunctionTests` is 94/94.

### Round 5 - LINQPad driver off net8.0

The nuget driver was the last net8.0 target (`net472;net8.0-windows7.0`, packed into `lib/net8.0`). Its .NET build bound linq2db's netstandard2.0 asset, and its `System.Collections.Immutable` floor pin referenced `$(Net8LatestForNuget)`, which this branch no longer defines - the empty override fell back to 10.0.12, i.e. a reference .NET 8/9 hosts cannot satisfy (the #5768 failure).

- Driver builds `net10.0-windows7.0`, nuget ships `lib/net10.0` with a `net10.0` dependency group; the driver references in-box SCI 10.0.0.0 (the pin is gone - on net10.0 the package is framework-provided and pruned, NU1510).
- `Net.IBM.Data.Db2` (+ `-lnx`/`-osx`) 9.0.0.400 rows only served the net8 driver; collapsed to 10.0.0.200. That client makes `DB2Connection.ReleaseObjectPool()` obsolete-as-error with no replacement, so DB2/Informix `ClearAllPools` is a no-op in the nuget driver (the net472 lpx keeps it).
- Third-party notices: LINQPad graph keys retargeted to `net10.0-windows7.0` per `-Action harvest`; `System.Threading.AccessControl` dropped (in no restore graph any more). Harvest clean with CLI, NuGet and LINQPad graphs restored; `-Action check` passes.
- Local: Release build of both driver TFMs, `dotnet pack` inspected. Not verified: loading the driver in a real LINQPad on .NET 10.

## Reading a red run here

The `/azp run test-<db>` comment path executes the **default branch's** copy of `tests.yml`, which knows nothing about `net110`. Use `gh workflow run tests.yml --ref feature/net11-support` to exercise this branch's matrix.


ChrisJollyAU:
Couple of my comments, watch out for Min and Max i think it is on arrays. .Net 11 now adds MemoryExtension overloads for them which has the Span types which is preferred.

Also watch out for a numeric conversion change when casting/converting a double to a decimal. https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/rc1/libraries.md#breaking-changes So a double 58.6 will end up as something like a decimal 58.600000000000001421085471520 whereas the previous implementation had it as a 58.6 in decimal. I eventually implemented my own internal version based on VarDecFromR8

MaceWindu:
Provider tests are started by `workflow_dispatch` here, so they don't show up in this PR's check list:

- [`tests [all]`](https://github.com/linq2db/linq2db/actions/runs/35455297660) on `894491b` — in progress. 10 legs green so far; the only red one is `Lin w_DB2InformixDuckDB`, with exactly two failures: `Round4Sql("DB2")` and `Round4Sql("DB2.LinqService")`, both reported as *test passed but is marked with `[ActiveIssue]`*. Its gate is obsolete now that the test rounds a decimal rather than a double, so DB2's inexact `ROUND` over a `Float` is no longer on the path.
- [`tests [db2.all]`](https://github.com/linq2db/linq2db/actions/runs/35457817367) on `5ece0ee` — that leg again, with the gate removed.

PR #5944, opened 2026-09-19
Raise C# to 15 and enable runtime async on net11.0

MaceWindu:
Stage 1 of the .NET 11 track, on top of #5942. **Not for merge before .NET 11 RTM** — it inherits that branch's RC SDK pin.

Two independent switches, one commit each.

## C# 15

`LangVersion` 14 → 15. Nothing here uses a C# 15 feature except where an analyzer the bump enables demands one: `IDE0410` (use labeled jump statement) becomes applicable at C# 15 and is an error under this repo's analyzer severity. Three sites, all the same shape — a bool flag set in an inner loop, an inner `break`, then `if (flag) continue;` in the outer one:

| Site | Label | Flag dropped |
|---|---|---|
| `SqlTable.cs:80` | `nextSiblingColumn` | `alreadyPresent` |
| `JoinsOptimizer.cs:67` | `nextChildJoin` | `hasDependentSibling` |
| `ExpressionBuilder.EagerLoadUnion.cs:604` | `nextSlot` | `alreadyUsed` |

`Source/CodeGenerators` and `Tests/Tests.T4.Nugets` import no props file and carry hand-duplicated copies of the setting, so they move with it.

## Runtime async

Roslyn stops emitting async state machines on net11.0; the runtime drives suspension instead. Measured on the built library:

| TFM | `MethodImplAttributes.Async` | `AsyncStateMachineAttribute` |
|---|---:|---:|
| net11.0 | 254 | 0 |
| net10.0 | 0 | 254 |
| net462 | 0 | 255 |
| netstandard2.0 | 0 | 255 |

Facts worth carrying, each measured rather than read:

- **The opt-in is exactly `runtime-async=on`.** Bare `runtime-async` and `runtime-async=true` reach csc (verified on the real command line) and are silently ignored — no error, no warning. A green build is not evidence the feature is on.
- **It lives in a new root `Directory.Build.targets`, not in `Directory.Build.props`.** A props file is evaluated before the project body, so for a project that declares a single `TargetFramework` there — `Tests/Tests.SingleFile` does — a `$(TargetFramework)` condition never fires. The same condition in `.targets` does.
- **The condition is the `IsTargetFrameworkCompatible` form** already used by the package bands and the feature constants, so net12 and later need no edit here.
- **Async iterators are not converted** — `async IAsyncEnumerable<T>` keeps `AsyncIteratorStateMachineAttribute` — and `ExecutionContext` still flows across an await, so the four `AsyncLocal` scopes are unaffected.
- **To build without it: `-p:Features=strict`**, which overrides the conditioned assignment as a global property. MSBuild does not hash `$(Features)` into the `CoreCompile` cache, so that flip needs `-t:Rebuild` or it silently reuses the previous binary.

Runtime async is on for **every** net11.0 C# target including the test projects — the suite is deliberately the stress test. F# projects are unaffected (`Fsc` does not consume `$(Features)`); `Vbc` does receive the flag but `Tests/VisualBasic` contains no async code.

## Verification

- `dotnet build linq2db.slnx -c Release` — 0 errors, all four TFMs, analyzers and `TreatWarningsAsErrors` active.
- `dotnet build Examples/Examples.slnx -c Debug` — 0 errors, 0 warnings. (That solution's **Release** build is red on master too — 60 pre-existing analyzer findings, invisible because CI builds it Debug-only. Not touched here.)
- Lowering A/B, forced with `-t:Rebuild` in both directions: feature on → 254/0, `-p:Features=strict` → 0/254, on again → 254/0.

Provider matrix: not yet run at the time of opening.


MaceWindu:
Provider tests are started by `workflow_dispatch`, so they don't show up in this PR's check list:

- [`tests [all]`](https://github.com/linq2db/linq2db/actions/runs/35465948808) on `77800f9` — the first run with runtime async on. The control is #5942 at `5ece0ee`, whose own `[all]` run is green over the identical test corpus, so any delta here is attributable to this PR's two properties.

PR #5947, opened 2026-09-19
SelectDynamic and dynamic Pivot: runtime-determined column sets

MaceWindu:
A dynamic PIVOT API (#1475, #5708) and the runtime-column-set projection it is built on, which also answers discussion #4992.

`SelectDynamic` projects a static selector as usual and additionally materializes one column per entry of a **runtime** key set into `TResult`'s `[DynamicColumnsStore]` — the column set is decided when the query is built, while every expression stays statically typed.

```cs
db.Balances.SelectDynamic(
    x => new BalanceDto { Id = x.Id },
    columnNames,                                 // runtime
    (x, name) => Sql.Property<decimal>(x, name));
```

## Why the template is instantiated at the call site

The template is instantiated once per key in the extension method, not in a visitor. That is forced, not preferred: `ExposeExpressionVisitor` descends into quoted lambdas and rewrites `Sql.Property` while the name argument is still a free parameter, which throws before any builder runs. The alternative was a new branch in that visitor — a cross-cutting core change this avoids entirely.

## Query cache

Runtime keys are materialized to an array and passed as `[SqlQueryDependent] string[]`. A `List<T>` constant is stripped from the query cache key, after which the dependent comparison is skipped as well — which would let two different column sets share one compiled query.

## Implementation

`SelectDynamicBuilder` builds a `SqlGenericConstructorExpression` from the static selector and appends one `DynamicColumnInfo` assignment per key, reusing the existing store materialization in `EntityConstructorBase`. No new `IBuildContext`, no shared visitor changed, no `LinqServiceSerializer` work — the output is ordinary fields and columns.

Rejects a `TResult` without a store (the columns would otherwise be dropped silently), duplicate names, and names unusable as SQL identifiers.

## Tests

Store population, both guard paths, and three query-cache guards: the key set discriminates (with `List<T>` legs guarding the stripping trap), computed names discriminate despite identical SQL, and the value template discriminates.

## Dynamic Pivot

The reason `SelectDynamic` exists. The pivoted values are decided when the query is built, while the key, the FOR column and every cell expression stay statically typed.

```cs
var years = db.Sales.Select(x => x.Year).Distinct().ToList();   // runtime

db.Sales.Pivot(x => x.Category, x => x.Year, years,
    g => new SalesDto { Category = g.Key, Total = g.Sum(x => x.Amount) },
    PivotCell<Sales, int>.Sum(x => x.Amount, y => "AMT_" + y),
    PivotCell<Sales, int>.Max(x => x.Note,   y => "NOTE_" + y));
```

N cell templates crossed with M runtime values give N*M generated columns, so a report can carry several aggregates per pivoted value alongside ordinary static members. That is the shape the marker-projection API on #5708 cannot express, because its column count is fixed by the projection's member list.

- `PivotCell` declares one aggregate over one column plus the factory that names its generated columns. It never enters the expression tree - the extension method decomposes it at the call site, so the names are data and take part in the query cache key.
- Two result shapes: a user type carrying a dynamic-columns store, projected through a selector over the grouping, or the built-in `PivotRow<TKey>` when there is no type to declare.
- `Pivot` is sugar over `GroupBy` plus the `SelectDynamic` marker, lowering each cell to the conditional aggregation a portable PIVOT uses. No new builder, no AST node, no provider flag.
- `SelectDynamicCore` carries its cells as `LambdaExpression` so a pivot can mix cell types - `Sum` yields `decimal?`, `Count` yields `int`.

Guards: at least one cell, a name factory per cell once there is more than one, and no empty or duplicate generated names.

Not yet covered here: the native `PIVOT` fast path (it needs the AST from #5708) and dynamic `Unpivot`.

## Follow-up commits

**Cells without naming the source type.** `PivotCell<TSource, TFor>.Max(...)` is a static method reached through a generic type, so `TSource` has to be written out - impossible when the pivoted source is a join projected into an anonymous type, which is the shape of the production query on #5708. Two overloads take their cells as lambdas over a `PivotCellFactory<TSource, TFor>`, so the compiler supplies the type that cannot be spelled:

```cs
q.Pivot(x => x.e.Id, x => x.e.TheKey, activityIds,
    g => new ListDto { Id = g.Key, CoaName = g.Max(x => x.cm.Name) },
    c => c.Max(x => x.rir.IdeRub, a => "IDE_" + a),
    c => c.Max(x => x.rir.LibRub, a => "LIB_" + a));
```

**A generated name that a result member would shadow is now rejected.** Member resolution finds the declared member first, so a cell named `Key` on a `PivotRow<TKey>` silently read the grouping key. Guarded in the shared tail, so both `SelectDynamic` and `Pivot` are covered.

**Acceptance and justification tests.** The production query from #5708 - five tables, an inner join filtering the pivoted values against a configuration table, three left joins, a grouped anonymous projection, two cell templates crossed with a runtime activity list, static aggregates alongside - into both result shapes, direct and through `LinqService`. Its SQL shape is pinned too: no `PIVOT` keyword, one conditional aggregate per generated column, and a post-pivot filter over a generated column emitted as `HAVING`.

One test each for the cases this API was justified by. #4992 and #4248 are answered as claimed; **#2953 is solved** - `SelectDynamic` over `FromSql` selects the named columns into the store, where that issue reports only the mapped column being emitted.


MaceWindu:
Absorbed into #5708, which now carries every commit from this branch plus the unpivot work that sits on the same tree. The branch stays; review moves to #5708.

Issue #3929, opened 2023-01-26
Mapping to DateOnly[] column type doesn't work


VitaliyNec:
.Net 7
PostgreSQL 14.0 on x86_64-pc-linux-musl, compiled by gcc (Alpine 10.3.1_git20210424) 10.3.1 20210424, 64-bit

```cs
[Column] public DateOnly Date { get; set; }
```

 Work!

```cs
[Column] public DateOnly[] Dates { get; set; }
```

```
Cannot convert value 'System.DateTime[]: System.DateTime[]' to type 'System.DateOnly[]'

at LinqToDB.Common.ConvertBuilder.ConvertDefault(Object value, Type conversionType)
   at lambda_method158(Closure, DbDataReader)
   at LinqToDB.Expressions.ConvertFromDataReaderExpression.ColumnReader.GetValue(DbDataReader dataReader)
   at lambda_method137(Closure, IQueryRunner, DbDataReader)
   at LinqToDB.Linq.QueryRunner.Mapper`1.ReMapOnException(IDataContext context, IQueryRunner queryRunner, DbDataReader dataReader, ReaderMapperInfo& mapperInfo, Exception ex)
   at LinqToDB.Linq.QueryRunner.ExecuteQuery[T](Query query, IDataContext dataContext, Mapper`1 mapper, Expression expression, Object[] ps, Object[] preambles, Int32 queryNumber)+MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
```

MaceWindu:
Related to #1660

You can define required conversion in mapping schema as workaround

MaceWindu:
Let's keep it open till we not improve built-in mappings

PR #5956, opened 2026-09-22
LibRed managed Access provider

MaceWindu:
Stage 3 of the .NET 11 track, on top of #5942 (.NET 11 support) and #5944 (C# 15 + runtime async). **Do not merge before .NET 11 RTM** — `global.json` pins an exact prerelease SDK, inherited from #5942.

## What this adds

[LibRed](https://github.com/CirrusRedOrg/EntityFrameworkCore.Jet) is a managed, cross-platform Jet/ACE engine and ADO.NET provider. It gives Access a transport that needs no Microsoft driver install and is not Windows-only — today every Access provider resolves through OLE DB or ODBC, every Access CI leg is Windows with an ACE redistributable, three of the four are x86, and there is no Linux Access coverage at all.

`LibRed.Ado` is `net11.0`-only, which is why this sits behind the .NET 11 work.

## Shape

A third `AccessProvider` flavour, not a new provider family — the engine is Jet/ACE either way and only the ADO.NET transport differs, so the SQL builders, optimizer, convert visitor, member translators and every `ProviderName.Access`-keyed `Sql.*` registration are reused unchanged.

**One data provider**, `Access.LibRed`. `AccessVersion` selects only the member translator, and its Jet arm exists because Microsoft's JET driver has no `REPLACE` — LibRed has it against an `.mdb`, so a per-format split would have mistranslated. `Access.LibRed.Mdb` and `Access.LibRed.Accdb` are test configurations over that one provider.

The adapter is reflection-loaded like every other dynamic provider adapter, so the shipped `linq2db` package references no LibRed assembly and the portable target frameworks are untouched.

## Provider specifics

Each measured against the engine, not inferred:

| Area | Behaviour |
|---|---|
| Parameters | named-only, so the base `@name` builder applies and `IsParameterOrderDependent` is `false` |
| `IsDBNullAllowed` | overridden — LibRed implements neither `GetSchemaTable` nor `IDbColumnSchemaGenerator`, and the base calls the former unguarded, so every materialized `SELECT` would throw before its first row |
| GUID literals | emitted unbraced; the braced form the base schema emits parses and matches **zero rows** |
| `CHAR` padding | not trimmed — the reader reports CLR type names, so a fixed-width column cannot be told from a `VARCHAR` at read time |
| Object names | the database component is dropped, as `SqlCeSqlBuilder` does, since LibRed's grammar has no qualified-name production |
| `CREATE TABLE` | `COUNTER` type instead of an `IDENTITY` attribute, and a primary key without `CLUSTERED` |
| Schema | read with SQL from the engine's `INFORMATION_SCHEMA` pseudo-tables plus `MSysRelationships` for foreign-key column pairs; views come from `MSysObjects` filtered by query type, so no action query is ever executed |

## Tests

Two configurations, one per file format, listed in the **NET110 provider list only** — no conditional compilation in test code; a provider that cannot load on a target framework simply is not in that framework's list.

The committed `.mdb`/`.accdb` are created by Microsoft's own engine through ADOX, so the suite exercises LibRed against files the world actually has. Their schema and data are then written by LibRed through the normal create-data flow.

`AccessTypeTests` is Access's first `Types/` fixture — 16 storage types across all six Access providers, non-nullable and nullable, parameter and literal, every bulk-copy mode. Green on all four flavours reachable locally; the two Jet ones need an x86 run.

## State

Full suite on `Access.LibRed.Mdb`: **8134 tests, 60 failures**, down from 1440 on the first pass. The two largest reductions were defects in this branch rather than in LibRed — a concrete-type exception test that the remote transport's `RpcException` wrapper defeated (~1100), and `IDENTITY` vs `COUNTER` (~102).

Everything still red is either gated with `[ActiveIssue]` scoped to the LibRed configurations — which *asserts* the failure persists, so a fix upstream turns the gate into a failure demanding its own removal — or is one of ~37 remaining singletons.

Verified against `Access.Ace.OleDb` over the whole suite: **one failure, provider-independent and pre-existing**, so none of the shared-file changes moved the reference flavour.

## Not in scope

- **No extended dialect.** LibRed accepts window functions, `APPLY`, `FULL JOIN`, `OFFSET`/`FETCH`, `INTERSECT`/`EXCEPT` and `CASE`. None is enabled here; that is a separate change with its own baselines churn.
- ~~**No LINQPad support** — the driver targets `net472;net8.0-windows7.0`, which cannot load a net11.0-only dependency.~~ Superseded: see *Follow-up commits (2026-09-23) - LINQPad* below.
- **No CLI scaffold key or CI leg yet.**
- 23 upstream LibRed limitations are recorded and gated but **not yet filed** against EntityFrameworkCore.Jet; the gates carry no tracking link until they are.

## Follow-up commits (2026-09-22)

Two commits since the description above. They supersede its **State** section.

A clean full run over **both** configurations — 15 381 cases, direct and `LinqService` — reported **119 failures**, against an `Access.Ace.OleDb` control of **8 214 cases with exactly one**: `TestExpressionVisitorHops(10)`, which carries no provider in its case name and is pre-existing on this branch stack. So none of the shared-file edits moved the reference flavour, and the LibRed list is interpretable. All 119 reconcile into nine causes.

Run the two flavours in **separate hosts**: a first attempt with `Access.Ace.OleDb` in the same process died at ten minutes with a native `0xC0000005` inside `UnsafeNativeMethods.ICommandText.Execute`, taking the LibRed lanes with it. That is the access violation `AccessOleDbSchemaProvider.cs:52-55` already documents; run alone, the same control passed clean twice.

### `4e64debe2` — the GUID string translation was ours

`AccessMemberTranslator.GuidMemberTranslator` emits `LCase(Mid(CStr(g), 2, 36))`, where the `Mid` exists to strip the leading brace Access's `CStr` produces. **LibRed renders a GUID unbraced** (measured: `Len` 36), so the `Mid` ate the first hex digit instead — a silent wrong-results bug, and the exact mirror of the unbraced-literal decision already in this PR. **40 of the 119.**

The symptom is explained equally well by a 0-based `Mid`, and the two produce byte-identical output; `Mid('ABCDEF', 2, 3)` = `BCD` is the control that separates them. Fixed by a new `AccessLibRedMemberTranslator`, selected in `CreateMemberTranslator` by `Provider` rather than by `AccessVersion`.

Three `[ActiveIssue]` gates keyed to `AllAccess` were **passing** on LibRed — a red under the run-and-verify attribute — and are narrowed to `AllNativeAccess`, as is `TestDateOnly`'s, so LibRed falls onto the unscoped `#3929` gate whose message it actually matches.

### `5926952da` — the remaining dispositions

| Cause | Cases | Disposition |
|---|---|---|
| GUID `CStr`/`Mid` | 40 | fixed in the engine |
| over-broad `[ActiveIssue]` | 12 | un-gated — LibRed passes |
| sequential multi-column `SET` | 12 | gated, upstream |
| comma-joined multi-table `UPDATE` | 8 | gated, upstream |
| `ORDER BY` ordinal | 12 | excluded, matching the `SqlCe` exclusion already there |
| harness / capability | 12 | fixed test-side |
| `CVar`, 64-bit literal, `Byte` narrowing, GUID index key, `CHAR` padding | 22 | gated / asserted directly |

Twelve of the test-side fixes are not LibRed defects at all: opening `"BAD"` as a file database throws `FileNotFoundException`, which the invalid-connection-string assertions did not list; LibRed **does** support a second reader on one command, so it moves into `MARS_MultipleDataReadersOnSameCommand_Supported`; and `nchar(20)` padding is now asserted directly rather than gated.

Four new upstream defects were measured, each with a control that could have disagreed — the register is now **27 rows**, still unfiled:

- a multi-column `SET` is applied in order, so `SET [A] = [B], [B] = [A]` on `(100, 200)` yields `(200, 200)` where Access yields `(200, 100)`
- a bare integer in `ORDER BY` is inert rather than an ordinal — `ORDER BY 2` is the arm that rules out a mis-resolved ordinal
- `CVar(1)` returns a typed `Int32`, so the function is a no-op
- `-9223372036854775808` overflows the **literal** parser, while the same value bound as a parameter round-trips exactly

### Still open

- The post-fix failure count is **arithmetic, not a measurement** — each category was verified by its own filtered run, but no full suite has run since.
- TO-2's baselines were captured for all three configurations but **before** the GUID fix, so they are stale for exactly the tests that changed most. The cross-comparison is unrun.
- Phase D (CI leg) and Phase E (CLI scaffold) remain untouched.

## Follow-up commits (2026-09-23) — `LibRed.Ado` 11.0.0-alpha.3

Six commits. **alpha.3 published on 2026-09-22 and moves most of the surface this branch worked around**, so this section supersedes the *Provider specifics* table, the last two *Not in scope* bullets and the previous *Still open* list. Every verdict below is a re-measurement against the package, and where a behaviour changed the same battery was run against alpha.2 **and** alpha.3 on a copy of one database — so "fixed" and "regressed" are both diffs of two measurements, not readings of the release notes.

### `f480cb93a` — the bump, and the workarounds it retires

| Was | Now |
|---|---|
| `IsDBNullAllowed` overridden — `GetSchemaTable` unimplemented | both it and `IDbColumnSchemaGenerator` work, with correct `AllowDBNull`; override gone |
| `CHAR` padding not trimmed — reader reported CLR type names | `GetDataTypeName` reports the store type, so the real trim registration applies as on the other flavours |
| `COUNTER` type, primary key without `CLUSTERED` | both accepted; both `CREATE TABLE` overrides gone |
| explicit FK column lists in `Access.sql`, `TestBase.Identity.cs` | bare `REFERENCES` resolves to the parent's primary key — **both files are byte-identical to the merge base again** |
| five `CREATE Procedure` shapes skipped for LibRed | all five create; the `SKIP` regions and the `AccessProceduresTests` narrowing are gone |

That also empties ~35 `[ActiveIssue]` gates — `ORDER BY n` is an ordinal, comma-joined and derived-table `UPDATE`/`DELETE` parse, a multi-column `SET` is evaluated against the pre-update row, `DATEADD` returns a `DateTime`, date parts return `Int32`, the 64-bit minimum literal evaluates. Those files are reverted to the merge base rather than hand-edited.

**Two workarounds had become wrong, not merely unnecessary.** The braced GUID literal now matches and the unbraced one matches nothing — the exact reverse of alpha.2 — so the mapping-schema override was producing empty results; and `CStr` renders a GUID braced, so the base `Mid(CStr(g), 2, 36)` is correct again and `AccessLibRedMemberTranslator` is deleted.

### alpha.3 regressed two things

- **The reader contradicts itself on a computed column.** For `[x] + [x]` over a `CURRENCY` `SUM`, `GetFieldType` *and* `GetSchemaTable` both report `Int32`/`Long` while the value handed back is a `Decimal`, so a materializer compiled from the declared type fails the cast. alpha.2 reported `Decimal` and was consistent. 6 failures, probably 10.
- **A text column compared against a numeric literal** — `=`, `IN`, `NOT IN` over a `VARCHAR` — now coerces the text to a number and throws on any non-numeric value, where alpha.2 evaluated it and ACE accepts it. 4 failures.

Both are gated citing the A/B evidence. `TestZeroDate` stays gated but its mechanism was recorded from the wrong layer: 1899-12-30 round-trips through a parameter, a `#…#` literal and `DateSerial(…)`; the loss is in the typed read, where `GetValue` returns the date and `GetDateTime` returns `DateTime.MinValue`.

### `38be77098` — the schema provider reads `GetSchema`

D-4 read the engine's `INFORMATION_SCHEMA` pseudo-tables because alpha.2 implemented no ADO schema API. alpha.3 implements twenty collections, so the premise is gone. Measured side by side before the rewrite: `GetSchema` is a superset on every fact the provider needs except `CreateFormat`.

Views are three rows with their definitions instead of twelve `MSysObjects` candidates, a hand-written DAO `Flags` gate and one bind probe each — and the engine, not this repo, decides that a parameterised select is a procedure. View columns arrive typed. Identity is a flag rather than an inference from `DATA_TYPE = 'counter'`. Foreign keys carry their update/delete rules.

**Procedures are reported for the first time**, which retires the "LibRed reports no procedures" adjudication and un-gates both `Issue792Tests`. The safety property was measured before the code was written: `SchemaProviderBase` describes each procedure under `CommandBehavior.SchemaOnly`, alpha.2 ignored that and *ran* the statement, alpha.3 honours it — describing `Person_Insert`/`Update`/`Delete`/`AddIssue792Record` leaves the row counts untouched.

`AccessProceduresTests`'s three call helpers now name their parameters as the declarations do: **LibRed binds a stored query's parameters by name**, where OLE DB and ODBC bind positionally and ignore names, so one helper had carried a misspelling that never mattered.

### `637bf3460`, `ae3f09f96`, `31d564f7b`, `19ccea411` — Phase D and Phase E

- A **Linux-only CI leg**, `z_Access_LibRed` — the first Linux Access coverage. Its providers are declared under `NET110.Azure` rather than the shared TFM-independent `Azure.TestJob` section, because a `full_run` also runs the leg's net10.0 pass where `LibRed.Ado` is not referenced.
- A **CLI provider**, `AccessLibRed`, with the MCP tool entry. `LinqToDB.CLI` already targets `net10.0;net11.0`, so the reference is TFM-conditional and the arm is guarded to match rather than failing inside the reflection adapter.
- A **CLI scaffold key** across the five templates — 89 files, regenerating byte-identical.

**Generating real scaffold output found two defects nothing else does.** `GetSchema("Tables")` marks only the four catalog tables as `SYSTEM TABLE` and reports `MSysAccessStorage` and the `MSysNavPane*` family as ordinary tables, so the first model carried six `MSys` entities the other Access keys do not. And diffing the model against the OLE DB one showed every procedure parameter named `[@firstName]`: for a `PARAMETERS` clause **Access itself wrote**, LibRed reports the `[ ]` quoting as part of the name and re-emits the definition double-quoted, after which it no longer parses — so such a query can be neither described nor executed, and scaffolds as `ExecuteProc` returning `int` where OLE DB produces `QueryProc<…Result>`. A query written *through* LibRed round-trips, which is why the suite cannot see it: the suite's database is LibRed-written and only the CLI scaffolds an ACE-written one. The name is unquoted on our side; the result-set loss is upstream and the committed output shows it.

Three model differences are LibRed being **more accurate** and are kept: it detects identity where the OLE DB flavour hard-codes `IsIdentity = false` for #3149, reports `Binary(10)` where OLE DB coarsens to `VARBINARY(10)`, and types procedure parameters from the declaration rather than from a regex over the definition.

### State

Full suite over both configurations: **15 513 cases, 28 failures, 284 verified-still-failing skips**, against 119 on the previous run. The 28 are 10 alpha.3 regressions, 8 defects of these commits (since fixed and re-verified), 4 needing a re-gate on a changed mechanism, 2 that now pass and needed a pre-existing non-LibRed gate narrowed, 1 gRPC transport flake and 1 pre-existing provider-independent failure. After the fixes, procedures + schema provider + `Issue792` + create-data are **44/44** on both configurations, and the Microsoft flavours are green — `Access.Ace.OleDb` 84/0, `Access.Ace.Odbc` 98/0, each in its own host.

The P13 upstream register is re-measured: **14 of its 27 rows are fixed upstream** and must not be reported. The rest, plus the three new ones above, are collected in the corpus `findings.md` for triage rather than filed.

### Still open

- The post-fix count is again **arithmetic** — every cluster was re-run individually, but no full suite has run since the fixes.
- TO-2's baselines are stale and the cross-comparison is unrun. The expected delta set has shrunk: the GUID literal is back to the braced form and the `COUNTER`/no-`CLUSTERED` divergences are gone, leaving parameter naming.
- The two Jet flavours still rest on CI's x86 legs.

### Correction: the post-fix run, measured

The **State** block above was written before the fixes were re-run as a suite; its 28 is the pre-fix number.
Measured after them: **15 513 cases, 3 failures, 15 212 passed, 298 verified-still-failing skips**, in
17 m 36 s.

Two of the three were `Issue1925Test`, and the honest note is that it was in the *previous* run's failure
list too with no disposition written — the cluster breakdown above accounts for 26 of the 28, not 28.

It is not a LibRed defect. The test already expects Access to reject `LIKE '[0'`: the OLE DB arm asserts
`OleDbException` and the ODBC arm `OdbcException`. LibRed rejects the unterminated bracket as well, with
`ArgumentException("Invalid pattern string")`, so it fell through to the `else` branch that expects no
throw. Every other pattern in the test executes on LibRed, including `'%[[]0[-]9]%'`, so that one shape is
all that needed saying; `bcca00212` adds the arm, verified 7/7 across both LibRed configurations and
`Access.Ace.OleDb` in one run.

That leaves **`TestExpressionVisitorHops(10)` as the only failure** — no provider in its case name,
pre-existing on this branch stack.

One measurement note for anyone comparing durations: the first full run took 2 h 11 m and this one 17 m,
same filter and same providers. The difference is a cold `Tests/Linq` build, not the tests.

### TO-2: baselines captured and cross-compared

Supersedes the *TO-2 baselines are stale and the cross-comparison is unrun* line above.

Both LibRed configurations and `Access.Ace.OleDb` were captured over the full suite into an isolated
`BaselinesPath` under the worktree — 3 592 baseline files per LibRed config, one failure per run
(`TestExpressionVisitorHops(10)`).

**`Access.LibRed.Mdb` vs `Access.LibRed.Accdb`: 3 592 pairs, 0 differing, 0 unpaired.** One provider really
does serve both file formats identically, which until now was an argument rather than a measurement. It
also validated the comparison before the comparison that mattered — and caught two bugs in it: a
parameterised case name ends `(provider, arg).sql`, so pairing on `"(provider).sql"` left 862 files
unmatched on each side and read exactly like a coverage gap, and `Test-Path` treats a `[` in a case name as
a wildcard.

**`Access.LibRed.Mdb` vs `Access.Ace.OleDb`: 3 569 pairs, 2 393 identical, 1 176 differing, 23/19
unpaired.** The delta set this PR originally predicted did not materialise — parameter *names* are
identical, because both flavours use the base normalizer, and so is `DateValue`; the alpha.3 reverts
removed the rest. Three mechanisms account for everything:

| Files | Difference | Verdict |
|---|---|---|
| ~1 130 | the parameter-type comment in the baseline preamble: `DECLARE @ID  -- Int32` against `DECLARE @ID Integer -- Int32` | a fourth category, so a finding by TO-2's own rule — but trace-only |
| 16 | the database-qualified name dropped: `[Database\TestData].[Issue681Table]` → `[Issue681Table]` | expected, the `BuildObjectName` override |
| 3 | the GUID literal `'{…}'` against `{guid {…}}` | expected — OLE DB is the outlier here, that escape is its own mapping schema's |

The first was chased to a mechanism rather than left as a curiosity. Both Microsoft flavours **override**
`GetProviderTypeName` to read their provider enum (`OleDbParameter.OleDbType` / `OdbcParameter.OdbcType`);
`BasicSqlBuilder`'s base maps only six `DbType` values and returns `null` for the rest, so `Int32`,
`Boolean`, `DateTime`, `Guid`, `Int64`, `Int16`, `Double`, `Single` and `Byte` print blank. This PR
declined that override on the grounds that `LibRedParameter` exposes no provider type; re-probed on
alpha.3 that is still true — its only type member is `DbType` — so there is nothing to read and no override
to write. It affects the emitted comment, never the SQL sent.

The 42 unpaired files all trace to a decision already in this PR: the MARS supported/not-supported pair,
`CreateDatabase`'s ADOX gate, the `AllNativeAccess` narrowings, the `WITH OWNERACCESS` exclusions, and the
tests LibRed now passes where Access does not.

`baselines-reviewer` was not used: it reads the `linq2db.baselines` clone and this capture is a local
isolated directory, so the comparison ran through a purpose-built script that pairs by test identity, drops
the per-statement provider header, and buckets differences by the shape of the first differing line.

### TO-10: the net11.0-only package never loads on the net10.0 pass

The claim in `access.libred.json`'s comment — that declaring the providers under `NET110.Azure` rather
than the shared `Azure.TestJob` section keeps a `full_run`'s net10.0 pass from trying to load a
net11.0-only assembly — is now measured rather than argued. `Tests/Linq` was built `-c Azure`, so the
`AZURE` symbol appends `.Azure` to the TFM config name exactly as CI does; the pipeline's own
`Build/Azure/configs/access.libred.json` was staged as `UserDataProviders.json`; the test host was then
asked to list its cases on each TFM.

| TFM | cases listed | LibRed cases |
|---|---|---|
| net11.0 | 15 529 | 7 313 `Access.LibRed.Mdb` + 7 313 `Access.LibRed.Accdb` |
| net10.0 | 903 | **0** |

On net10.0 `NET100.Azure` resolves no providers at all, so every `[IncludeDataSources]` case drops out and
the listing collapses to the 903 provider-independent ones. The only `LibRed` text in that log is the
`UserDataProviders.json` banner the harness echoes at discovery — no test case carries it.

### `c44d367bb` — the CLI's LibRed payload in the third-party notices

The net11.0 CLI ships `LibRed.Ado`/`.Core`/`.Engine`/`.Sql` and their transitive
`Antlr4.Runtime.Standard`, none of which mapped to a component, so the notices verify step failed on all
seven `linq2db.cli` RID payloads with 35 unmapped DLLs — reddening both the GitHub `Build and pack` leg
and Azure build 23768. Entries taken from `-Action harvest`: ANTLR under BSD-3-Clause with a new
`texts/antlr4-bsd3.txt`, the four LibRed packages under the shared `mit.txt`, all scoped to the
`linq2db.cli` artifact and net11.0 only. `-Action check` passes, and `-Action verify` against a
cross-published `linux-x64` payload reports 156 entries, 101 components, 0 unmapped.

The other red leg is not this branch's doing: `tests [all]` failed at 0 s because `issue_comment` loads
the workflow from the **default** branch, whose `tests.yml` has no `DOTNET_SDK_11` — so only SDKs 8/9/10
install and this branch's `global.json` cannot resolve. The dispatch form is the documented trigger for
this stack, and [run 35801786703](https://github.com/linq2db/linq2db/actions/runs/35801786703) ran with
`surface='[all]'` - it **failed**, see *Still open (2026-09-23, after LINQPad)* below.

### `d5d75493f` - merge of #5944 (and through it #5942)

Brings the LINQPad driver's move to `net10.0-windows7.0`. Conflicts in `Directory.Packages.props` (the collapsed `Net.IBM.Data.Db2` rows next to `LibRed.Ado`) and the generated CLI notice; the notice was regenerated from the merged manifest, `third-party-notices.ps1 -Action check` passes.

### `3b299aace` - LibRed in the LINQPad driver; Access on every OS

- The driver offers `LibRed (managed)` without referencing the net11.0-only `LibRed.Ado`: `LibRedFactory.Instance` is reached through `LibRedProviderAdapter.ConnectionType.Assembly`. It works when the LINQPad query runtime is .NET 11+; the provider's troubleshoot note says so.
- Access is no longer Windows-only. Off Windows, OLE DB and ODBC are hidden (existing connections still load) and LibRed is the default. The net472 lpx does not list LibRed.
- ~~New `ProviderInfo.ProvisionOnlyWhenSelected` keeps `LibRed.Ado` out of the all-clients fan-out `OverrideDriverDependencies` uses when a connection's client is unknown, so a .NET 10 query runtime is never handed a net11.0-only package.~~ Replaced by `MinimumRuntime` in `a9cfd6787`.
- ~~Connection-string detection: OLE DB marker -> OLE DB, `Driver=`/`Dsn=` -> ODBC, otherwise LibRed. The positional `_providers[isOleDb ? 0 : 1]` became a lookup by name.~~ Removed in `a9cfd6787` / `f05176b05`: the provider is chosen explicitly.
- Verified: Release build of both driver TFMs; the generated `NuGetPackageVersions.LibRed_Ado` is `11.0.0-alpha.3`; the driver assembly references no LibRed assembly. **Not verified:** LibRed end-to-end inside LINQPad, or whether LINQPad accepts a .NET 11 preview query runtime - needs a manual check on Windows and macOS.

### `a9cfd6787` - explicit Access provider choice, LibRed gated on the query runtime

Testing `3b299aace` from the local feed on a .NET 10 query: a bare `Data Source=<file>` auto-selected LibRed and LINQPad prompted for a `LibRed.Ado` download that cannot succeed there. Access was the only database that picked its provider from the connection string, which also hid the selector.

- Access shows the provider selector like every other multi-provider database; OLE DB is the default on Windows, LibRed elsewhere.
- `ProviderInfo.MinimumRuntime` replaces `ProvisionOnlyWhenSelected`: `OverrideDriverDependencies` reads `DriverDependencyInfo.FrameworkVersion` and never provisions a client whose floor is above it, on the selected-provider path as well as the fallback ones. Hidden providers are left out of the fallback.
- A LibRed connection on a pre-.NET 11 query fails with the troubleshoot text instead of a failed restore.

### `f05176b05` - one connection string per Access provider, paired entries for merged schema

- OLE DB, ODBC and LibRed take one connection string.
- `OLE DB, with ODBC schema merged` / `ODBC, with OLE DB schema merged` take two, each labelled by its provider; queries run on the first, the second only feeds `MergedAccessSchemaProvider`. Both directions exist so no saved connection changes which provider runs its queries.
- `ProviderInfo.SecondaryName` carries the pairing. Stored settings are unchanged (`Provider` / `SecondaryProvider` / `SecondaryConnectionString`), so a saved connection with a second string loads as the matching paired entry; choosing a single entry clears the second string.
- Access connection-string detection is gone; the dialog rejects a paired entry without its second string. The provider-name registry skips paired entries, which repeat a name already registered.

Test builds on the local feed: `6.6.0-local.1` (`3b299aace`), `.2` (`a9cfd6787`), `.3` (`f05176b05`). Release build of both driver TFMs green at each.

### Still open (2026-09-23, after LINQPad)

- **The Linux LibRed leg has never passed.** [Run 35801786703](https://github.com/linq2db/linq2db/actions/runs/35801786703) (`c44d367bb`, `surface='[all]'`): `Lin z_Access_LibRed` fails 14 154 cases, all `FileNotFoundException` on `.../net11.0/main/x64/Database\TestData.LibRed.mdb` - the `DataProviders.json` connection strings use `\`, which is not a separator on Linux. Every earlier LibRed run was on Windows.
- `Lin b_ClickHouse` failed in the same run on Octonica's `Expected value: 0x82. Actual value: 0x90` - the ClickHouse 26.9 ZSTD default, fixed on master by #5962, which this stack does not have yet.
- LINQPad: `local.3` is not yet verified in LINQPad (Windows and macOS, .NET 10 and .NET 11 queries).
- Unchanged from above: the remaining upstream LibRed limitations (corpus `findings.md`) are unfiled; the Jet flavours rest on CI's x86 legs.


ChrisJollyAU:
@MaceWindu GetSchema is on its way for libred. and just to note i have a local build with linq2db working. currently its a separate provider: Total 9686, Passed 9150, Failed 387. And vast majority of that fail is MERGE and CTE (unimplemented). it was basically throw every test at it because none of the normal skips picked the new provider up

MaceWindu:
@ChrisJollyAU , you can crosscheck findings to see if something missing https://github.com/linq2db/agents/blob/master/plans/feature-access-libred/findings.md 
This pr will establish provider support on ODBC/OLEDB SQL level and then I will build another pr with advanced SQL features on it.

ChrisJollyAU:
@MaceWindu Just looking through that, I do want to add that the LibRedSqlMode is a efcore provider mode. it only changes what the efcore provider emits. the LibRed engine actually doesnt care, you can keep your sql that you emit as resticted or as extended as you want. it accepts it either way. And with alpha 3 coming up you shouldnt have to make a single change to that access.sql file

MaceWindu:
I noticed it's a bit dated file. Will upload fresh one soon

ChrisJollyAU:
Alpha 3 out now - should sort out a lot of your issues

PR #5750, opened 2026-08-08
Translate TimeSpan members and date differences as elapsed time

sdanyliv:
Makes `TimeSpan` a first-class value in a query: a duration column carries the unit it is stored in, its members translate, and the difference between two dates becomes a duration the server computes. Replaces #5739.

Closes #5334.
Closes #4308.

## The problem

A `TimeSpan` has no single storage form. A column holding one is a number, and nothing about that number says whether it counts seconds or ticks — so today a member access on it is dropped or answered from the raw stored value, and `end - start` either fails to translate or comes back as something that is not the elapsed time.

#4308 is that first half reported: a `TimeSpan?` stored as `Int64`, whose `.TotalMilliseconds` and `.TotalSeconds` vanished on the way into SQL and left `WHERE col < col` behind. It asks for "a way for each DataProvider to extract the TimeSpan Properties, in the way how a timespan is stored on this Provider", which is what the declaration below is. Support stays opt-in, so that model needs the unit stated — `HasDuration(DurationUnit.Millisecond)` beside the existing `HasDataType` — and the members then translate. Left undeclared, a `TimeSpan` column keeps whatever it means today rather than being reinterpreted as a tick count.

The same shape was reported twice before and closed without it: #3840 (``Property 'Int32 Milliseconds' is not defined for type 'System.Nullable`1[System.TimeSpan]'``, asking for `.Milliseconds` inside a `Sql.DateAdd`) and #3993, which adds *datetime2 is incompatible with bigint* — a date meeting a number where an interval was meant, which is what the shift node exists to stop. Both are covered now, on the same terms: state the unit.

## What this adds

**A unit on the declaration.** `[Duration(DurationUnit.Second)]`, or `HasDuration` in fluent mapping. The read and write conversions are *derived* from it rather than written by hand, so the same model works across providers whose storage types differ.

```csharp
[Column, Duration(DurationUnit.Second)] public TimeSpan Elapsed { get; set; }
```

Stating the unit and writing a converter as well is refused where it is stated: they are two answers to the same question, and the query would believe both — the value comes back through the converter while the SQL is built on the unit.

**Members that translate.** `Days`, `Hours`, `Minutes`, `Seconds`, `Milliseconds`, `Ticks` and the `Total*` family, on a declared column and on a computed difference alike — matching what the CLR answers, including truncation toward zero on a negative duration.

**Differences as durations.** `end - start` lowers to an exact elapsed time where the provider can express one, and is refused by name where it cannot, rather than silently answering something else.

[...]

PR #5959, opened 2026-09-22
Coerce ClickHouse date operands to DateTime64 before the epoch functions

### LINKED ISSUE #5955: ClickHouse datetime function regression in 6.5.0
### Description

In 6.5.0, some ClickHouse functions on DateTime values fail. This worked correctly in 6.4.0.

### Reproduction Steps

```
[Test]
public async Task Test()
{
        .UseClickHouse(CONNECTION_STRING, ClickHouseProvider.Octonica);

    await using var connection = new DataConnection(options);

    await using var tempTable = await connection.CreateTempTableAsync<Projection>();

    var toInsert1 = new[]
    {
        new Projection { DateTime = DateTime.UtcNow },
        new Projection { DateTime = DateTime.UtcNow.AddDays(-7) }
    };

    await tempTable.BulkCopyAsync(toInsert1);

    var query =
        from grp in tempTable.GroupBy(_ => 1)
        let minSampleDateTime = grp.Min(x => x.DateTime)
        let maxSampleDateTime = grp.Max(x => x.DateTime)
        let sampleDayCount = (int)(maxSampleDateTime - minSampleDateTime).TotalDays + 1
        select new
        {
            sampleDayCount
        };

    var result = await query.SingleAsync();
}

private sealed record Projection
{
    [Column(DataType = DataType.DateTime)]
    public DateTime DateTime { get; set; }
}
```

### Exception details

```
Octonica.ClickHouseClient.Exceptions.ClickHouseServerException: 'DB::Exception: A value of illegal type was provided as 1st argument 'value' to function 'toUnixTimestamp64Nano'. Expected: DateTime64, got: DateTime: In scope SELECT toInt32(toFloat64(intDiv(toUnixTimestamp64Nano(MAX(grp.DateTime)) - toUnixTimestamp64Nano(MIN(grp.DateTime)), toInt64(100))) / toFloat64(864000000000)) + 1 FROM Projection AS grp LIMIT 2'
```

### Regression?

Yes, worked in 6.4.0

### Linq To DB version

6.5.0

### .NET version

10.0.400

### Database

ClickHouse 26.4.1.2212

### Database provider

Octonica.ClickHouseClient 3.1.9

### Operating system

Windows

### Configuration

_No response_

### LINKED ISSUE #5960: GetDbDataType discards a SqlExtendedFunction's declared type for the schema default
### Summary

`QueryHelper.GetDbDataType` returns the mapping-schema default for a `SqlExtendedFunction` rather than the `DbDataType` the node was built with. `SqlFunction`, which represents the same thing, returns its declared type verbatim — so two nodes for the same expression answer the type question differently.

### Current behavior

`Source/LinqToDB/Internal/SqlQuery/QueryHelper.cs:697` — `GetDbDataTypeImpl` matches `SqlParameterizedExpressionBase { Type: var t } => t` at `:719`, which covers `SqlFunction` (`SqlFunction.cs:9`). `SqlExtendedFunction` derives from `SqlExpressionBase` instead (`SqlExtendedFunction.cs:12`), so it matches no arm and falls through to the catch-all `{ SystemType: var t } => new(t)` at `:735`, whose `DataType` is `Undefined`. The public wrapper at `:668` then sees `Undefined` and substitutes `mappingSchema.GetDbDataType(expr.SystemType)` — the node's own `DataType`, precision, scale and length are gone.

The node does carry a type. Every construction site passes one (`Source/LinqToDB/Internal/DataProvider/Translation/SqlExpressionFactoryExtensions.cs:357`, `Source/LinqToDB/Internal/SqlQuery/SqlExtensions.cs:123`, `Source/LinqToDB/Internal/Linq/Builder/SelectBuilder.cs:103`, `Source/LinqToDB/Internal/SqlQuery/Visitors/SelectQueryOptimizerVisitor.cs:1602`), the remote reader restores it (`Source/LinqToDB/Internal/Remote/LinqServiceSerializer.cs:3230`), and the visitor preserves it across rebuilds (`Source/LinqToDB/Internal/SqlQuery/Visitors/QueryElementVisitor.cs:275`).

Worked example, measured on ClickHouse: for a column declared `[Column(DataType = DataType.DateTime)]`, the field reports `DataType.DateTime` but `MAX(field)` reports `DateTime64(7)` — the provider's schema default for `DateTime`.

### Expected behavior

A `SqlExtendedFunction` reports the type it was built with, as `SqlFunction` does. Either give it its own arm, or hoist the `Type` read so both node families share one.

### Notes

The absence may be deliberate — the four interval nodes immediately above are omitted on purpose, with `QueryHelper.cs:710-714` recording why — but nothing records a reason here, and `SqlExtendedFunction` post-dates most of the switch (introduced by #5468, extended by #5750).

Calibration: what is confirmed is the mechanism above and the ClickHouse reading. What is **not** demonstrated is a user-visible wrong result. In the measured case the symptom is masked, because `Source/LinqToDB/Internal/DataProvider/Translation/AggregateFunctionsMemberTranslatorBase.cs:292` and `:364` type the node from `methodCall.Method.ReturnType` through the same mapping schema, so both routes happen to agree. The gap has teeth only where a caller passes a deliberately non-default type. No failing query was captured.

Surfaced while fixing #5955; it is why the ClickHouse coercion in #5959 could not be made conditional on the operand's reported type.

Related: #5758 proposes carrying the mapping on the expression instead of recovering it, which would subsume this. That is a 7.0.0 redesign; this arm is fixable independently.

MaceWindu:
Fixes #5955
Fixes #5960

## Problem

ClickHouse's `toUnixTimestamp64Nano` and `toUnixTimestamp64Milli` accept only a `DateTime64`. Every other date type is refused by name:

```
DB::Exception: A value of illegal type was provided as 1st argument 'value' to function
'toUnixTimestamp64Nano'. Expected: DateTime64, got: DateTime
```

#5750 gave ClickHouse an elapsed-interval lowering in 6.5.0 that passes its operands raw, so a date difference started failing at the server. The millisecond `DateAdd` and `DatePart` paths have the same shape and have failed the same way since before 6.5.0.

**The reported `[Column(DataType = DataType.DateTime)]` is only one of four ways such an operand arrives, and three of the four need no declaration at all** — they are manufactured by the provider's own translator. Measured against 26.5.7.64:

| operand | emitted by | server type | linq2db's type, before |
|---|---|---|---|
| declared `DataType.DateTime` / `Date` / `Date32` field | user mapping | `DateTime` / `Date` / `Date32` | as declared |
| `toDate32(col)`, i.e. `x.Col.Date` | `TranslateDateTimeTruncationToDate` | `Date32` | `Date32` |
| `now()` | `TranslateNow` / `TranslateServerNow` | `DateTime` | `DateTime64(7)` |
| `makeDateTime(y,m,d,H,M,S)` | `TranslateMakeDateTime` without milliseconds | `DateTime` | `DateTime64(7)` |
| `MAX(col)` / `MIN(col)` over any of them | `AggregateFunctionsMemberTranslatorBase` | the column's | `DateTime64(7)` |

So `(x.FinishedOn - x.FinishedOn.Date).TotalHours` and `(Sql.CurrentTimestamp - x.StartedOn).TotalHours` are broken on a default-mapped column, and the issue's own query is broken through the aggregate.

## Why the operands could not be told apart

The last two rows are type lies, and they are why the first version of this PR coerced every operand unconditionally: an in-process type test would have answered `DateTime64` for `now()`, for `makeDateTime` and for the aggregate in the reported query, and skipped exactly the cases that fail.

Both are fixed here rather than worked around, so the coercion can ask the operand what it is:

- **`MIN` / `MAX`** reported the mapping schema's default for the CLR return type instead of their argument's type, because `QueryHelper.GetDbDataTypeImpl` had no `SqlExtendedFunction` arm and fell through to the CLR-type catch-all. This needed no new mechanism: `SqlArgumentDomains.ForAggregate` already marks them `SqlArgumentDomain.Element` — *"the result is one of the argument's own values … so the argument's column describes it completely, width included"* — and `QueryHelper.GetColumnDescriptor` already follows that arm for the value-converter path. Only the typing path was missing it. That is #5960; `SUM` is deliberately excluded, since it answers in the argument's terms but can outgrow the width the argument is declared with.
- **`now(…)` and `makeDateTime`** typed their result from the mapping schema, which maps a CLR `DateTime` to `DateTime64(7)`, while the server answers `DateTime`. They now name what their SQL produces. All four `now(…)` overloads are covered, not just `TranslateNow` — `DateTime.UtcNow` and the zoned forms reach the same functions.

## Changes

- `QueryHelper.GetDbDataTypeImpl` gains the `SqlArgumentDomain.Element` arm, plus a `{ Type: var t } => t` arm so every other `SqlExtendedFunction` reports its declared type rather than the CLR fallback.
- The four ClickHouse `now(…)` translations and the no-millisecond branch of `TranslateMakeDateTime` declare `DataType.DateTime`.
- `ClickHouseExpressionExtensions.AsDateTime64` returns the operand unchanged when it already reports `DataType.DateTime64`, and otherwise casts it via `factory.Cast` — the builder renders that through the provider's own conversion table as `toDateTime64(x, 7)`, so the emitted SQL is the same string a hand-built function produced. Called from the four sites that passed an operand raw: `ElapsedTicks`, `LowerTemporalArithmetic`, `TranslateDateTimeDateAdd` (millisecond) and `TranslateDateTimeDatePart` (millisecond).
- `LowerTemporalArithmetic` coerces the shifted date too — a sub-second interval cannot be added to a date, ClickHouse answers `addNanoseconds cannot be used with Date32`.
- Corrects the millisecond date part before 1970: the epoch is negative there and the truncating remainder carried the sign, so `1969-01-01 00:00:00.500` answered `-500`. `toMillisecond` answers it in one call but needs ClickHouse 24.6, noted in the code for a later modernization.

`DateDiffBuilderClickHouse` already wraps its operands and is unchanged. `CommonTruncationToTime` moved onto the same helpers - see the follow-up commit below.

## Tests

Thirteen ClickHouse-scoped tests plus one core-layer test, each confirmed red against unfixed code with its expected failure rather than merely red:

- `IntervalTranslationTests.ClickHouse.cs` — the issue's grouped `Min`/`Max`/`TotalDays` shape, tick counts, and shift/difference over `DataType.DateTime` and `DataType.Date32` columns, plus the two default-mapping shapes above.
- `TheCoercionIsEmittedOnlyWhereTheOperandNeedsIt` pins the SQL on **both** sides: absent over a default-mapped column, present over `now()`, `makeDateTime` and a declared `DataType.DateTime` column. Dropping the coercion where it is not needed is invisible in a result assertion — every other test here passes with it applied to everything.
- `DateTimeFunctionsTests.cs` — `DateAdd`/`DatePart` millisecond over second-precision and `Date32` columns, an explicit `Sql.Convert` over a `DateAdd` result, and the pre-1970 millisecond.
- `SelectQueryOptimizationTests.MinMaxReportTheirArgumentsTypeRatherThanTheirOwn` pins the aggregate typing contract directly, provider-agnostically, with `SUM` as the control.

Verified locally: **2475 passed / 0 failed** over `IntervalTranslationTests` + `DateTimeFunctionsTests` + `DateTimeOffsetTests` + `AggregationTests` + `AggregationNullabilityTests` + `GroupByTests` + `SelectQueryOptimizationTests`, across ClickHouse.Octonica, SQLite.MS and DuckDB. The aggregate and group-by fixtures are there because the `QueryHelper` arm retypes every `MIN`/`MAX` on every provider.

Red arm, with both typing fixes stashed and everything else in place — 4 of 4 red, each for its own predicted cause:

| test | failure |
|---|---|
| `MinMaxReportTheirArgumentsTypeRatherThanTheirOwn` | reports `(String, NVarChar)`, wanted `(String, VarChar, "50")` |
| `ReportedDayCountOverASecondPrecisionColumn` | `toUnixTimestamp64Nano(maxOrNull(grp.StartedOn))` → `Code: 43 … got: DateTime` |
| `DifferenceAgainstServerNowAnswers` | `toUnixTimestamp64Nano(now())` → `Code: 43 … got: DateTime` |
| `TheCoercionIsEmittedOnlyWhereTheOperandNeedsIt` | the `now()` and `makeDateTime` coercions are gone |

Release builds clean on `netstandard2.0`, `net10.0` and `Tests/Linq` net10.0 — 0 warnings, 0 errors.

## Baselines

The coercion is now emitted only where the server needs it, so most of the churn the first version of this PR introduced is gone: of the 69 ClickHouse baseline files that use these functions, **50 emit their operands bare** and 19 wrap. The 19 are the declared `DataType.DateTime` / `Date32` columns, `now()`, `makeDateTime` and the `toDate32` paths. No other provider's SQL is affected by the ClickHouse changes; the `QueryHelper` arm is cross-cutting by nature, and CI's baselines PR is the authority on what it moves.

## Notes

The cast's own reduction is *not* what makes this conditional. A non-mandatory `SqlCastExpression` whose target equals its operand's reported type by construction still reaches the SQL builder, in both the member-translator and convert-visitor paths, so the type test is explicit in the helper. Filed with the evidence as #5963.

## Follow-up commit

`ed6c32b74` - no behaviour change beyond the two SQL shapes below.

- `ClickHouseDateTime` is now `ClickHouseExpressionExtensions`, `ISqlExpressionFactory` extension methods: `AsDateTime64`, `AsDate32`, `ToUnixTimestamp64Nano` / `ToUnixTimestamp64Milli` (coerce their operand), `Now(type, utc)`, `DateTimeType`. The four `now(...)` overloads, `makeDateTime`, both date truncations and every epoch call site go through them.
- `CommonTruncationToTime` (`TimeOfDay`) uses the coercing helpers: over a `DateTime64` column the operand is emitted bare instead of `toDateTime64(x, 7)`, and its `toDate32` is no longer typed `DateTime`.
- `Week` date part: `toDateTime64(x, 1)` typed as `long` becomes `AsDateTime64`, so `toISOWeek` over a `DateTime64` column takes it bare.
- `ElapsedTicks` reuses `TruncateDivide` instead of its own `intDiv`.

Baselines: ClickHouse `TimeOfDay` and `Week` SQL over `DateTime64` columns loses the wrap, so the bare/wrapped counts above shift accordingly. Verified locally on ClickHouse.Octonica net10.0: 63/63 over the PR's ClickHouse interval tests and the `DateTimeFunctionsTests` `Now` / `UtcNow` / `CurrentTimestamp` / `DatePartWeek` / `DatePartMillisecond` / `DateAddMillisecond` / `TimeOfDay` / `MakeDateTime` methods. Release builds clean on `netstandard2.0` and `net10.0`.

## Follow-up commit

`53f2e4a8a`: tests only, no product change. The branch is now stacked on #5967 (#5966, 6.5.1) and rebased onto it.

- The coarse date-operand tests (whole-second and date-only columns, `x.Date`, the server clock, `MIN`/`MAX` over a column, the millisecond `DateAdd`/`DatePart` shapes) run on `[DataSources]` instead of ClickHouse only, in the new `IntervalTranslationTests.Coarse.cs`. `IntervalTranslationTests.ClickHouse.cs` keeps only `TheCoercionIsEmittedOnlyWhereTheOperandNeedsIt`.
- Date-only columns are `DataType.Date`, and `Date32` on ClickHouse through a configuration-scoped `[Column]`.
- `DifferenceAgainstServerNowAnswers` stores a fixed date instead of `DateTime.UtcNow`. The changing parameter value failed the direct-vs-remote baseline check on every remote context, including the original ClickHouse-only version.
- The shift tests declare the fixture's existing refusals (`UnsupportedShiftProviders`, `ShiftRefusedWhileBuildingProviders`, `UnsupportedDifferenceProviders`).
- Failures found on other providers are gated on #5965 by provider and error text: SQL Server, Firebird and Oracle sub-day `DateAdd`/`DatePart` over date-only / `DATE` operands; Oracle's millisecond part at a `TIMESTAMP` precision other than 6; DuckDB's difference against `now()`.
- The widening also surfaced #5966: the remote server resolved columns without configuration-scoped attributes. It is fixed in #5967, which this PR now builds on.

Verified locally on net10.0: 177 cases across SQLite.MS, DuckDB, ClickHouse.Octonica, PostgreSQL 18, SQL Server 2022, Oracle 23 and Firebird 5, direct and remote: 152 passed, 20 gated (inconclusive), and 5 ClickHouse remote baseline mismatches caused by #5966. On the stacked branch those pass, and ClickHouse is 28/28. The `Tests/Linq` Release build is clean. The #5965 gates name whole provider families but were measured on one version each; CI covers the rest.

## Follow-up commit

`84d9b2103`: tests only. The branch is rebased onto #5967's new head, so the commits above are now `87c737361`, `cfb93d066`, `b2ae936d4` (was `ed6c32b74`) and `44da244fe` (was `53f2e4a8a`).

The full CI matrix found failures the local run did not reach:

- Declared refusals, using the fixture's existing lists: Informix and SQL Server 2014- (`UnsupportedDifferenceProviders`) on the five coarse difference tests, Access (`NoTickTotalProviders`) on the two `.Ticks` tests.
- Gated on #5965 (cases added there): DB2 `SQL0182N` and Sybase / YDB dropped milliseconds on a millisecond `DateAdd` over `DATE` / `Datetime`; Sybase answering `0` for `(x - x.Date).TotalHours`.
- YDB excluded from `DatePartMillisecondBeforeTheEpoch`: its `Timestamp` cannot hold a pre-1970 value and the driver overflows on insert.
- The Oracle and Firebird gates matched the Shouldly source expression, which is in the message only where the test sources are available, so they failed on CI with `[ActiveIssue] Expected ... but found`. They now match the expected value.

The `ClickHouse.MySql` failure of `ConfigurationScopedColumnAttributeAppliesOnTheServer` came from #5967's commit and is gated there.


alexngn22 (commented):
111

PR #5969, opened 2026-09-23
LibRed extended SQL dialect

MaceWindu:
Child of #5956. **Do not merge before .NET 11 RTM** — inherits #5942's pinned prerelease SDK. **Work in progress**: two items below are unfinished.

## What this adds

#5956 runs LibRed on the unchanged Access dialect. LibRed's engine accepts a good deal more than Jet SQL, and until now none of it was emitted — and none of the tests that exercise it ran on LibRed, because every dialect exclusion named `TestProvName.AllAccess`, which includes LibRed.

**Enabled for LibRed only**, each from what LibRed advertises (grammar `AccessSql.g4` and `docs/functions.md` at `v11.0.0-alpha.3`):

| Area | Now emitted |
|---|---|
| Paging | `OFFSET … ROWS FETCH NEXT … ROWS ONLY`, skip in sub-queries, `TOP @p` |
| Joins / set ops | `CROSS`/`OUTER APPLY`, `CROSS JOIN`, join without condition, `INTERSECT`/`EXCEPT` |
| Window functions | all ranking / value / offset functions, frames, `FILTER`, `IGNORE NULLS`, `NTH_VALUE … FROM LAST`, `DISTINCT` in window aggregates, variance / `CORR` / `COVAR_*` / `REGR_*` |
| Expressions | `CASE` instead of `IIF`, `^` as `BXOR` (power becomes `Power(x, y)`), `EXISTS`/`IN` kept instead of `COUNT(*) > 0` |
| Functions | `STRING_AGG` for aggregate `string.Join`/`Concat`, `LTrim`/`RTrim` with a character set, `StrReverse`, `'ms'` in `DatePart`/`DateAdd`/`DateDiff`, `REPLACE`-based LIKE escaping |
| Intervals | tick counts, bare differences and date shifts, counted with `DateDiff_Big` in milliseconds |
| Other | SQL comments |

**Not advertised, so LibRed is excluded again by name** (`AllNativeAccess, AllAccessLibRed`): MERGE and MERGE-lowered Upsert, `UPDATE`/`DELETE` with Take/Skip, `RETURNING`, `COLLATE`, `CAST`, `KEEP`, hypothetical-set aggregates, `MEDIAN`, windowed percentiles.

## Shape

New LibRed types beside the existing `AccessLibRedSqlBuilder`: `AccessLibRedSqlOptimizer`, `AccessLibRedSqlExpressionConvertVisitor`, `AccessLibRedMemberTranslator`, selected in `AccessDataProvider` by `Provider`. Shared Access files gain overridable switches only (`IsCaseExpressionSupported`, `IsCommentSupported`, `IsTickArithmeticSupported`, `CorrectExistsAndIn`, `DatePartName`, `MathMemberTranslator.Power`), each defaulting to today's behaviour.

Tests: `AllAccess` keeps its meaning. One attribute-context rewrite narrows 486 dialect-exclusion sites to `AllNativeAccess`; the re-exclusions above name LibRed explicitly, so a bare `AllAccess` in an exclusion now means "not evaluated for LibRed".

## State

Local full suite on `Access.LibRed.Mdb` (Windows): **1 227 failures before this branch, 2 at best, 10 at this head.** Engine defects in advertised features are gated with `[ActiveIssue]` and recorded in the plan's findings: `IIF` typed by the branch taken, `NTH_VALUE` ignoring the default frame.

## Unfinished

- **Boolean sort keys** (all Access flavours: Jet stores True as -1, so a boolean `ORDER BY` sorts True first). The lowering to a 1/0 key is in flight: with it at render time, native `Access.Ace.OleDb` fails `DistinctOrderByBoolean` (*ORDER BY clause conflicts with DISTINCT*); moved before optimization, it still does and no longer reaches window order keys. This is the source of the 10.
- **Native control**: the `Access.Ace.OleDb` full suite crashed twice with `0xC0000005` inside `ICommandText.Execute` at the same point (~2 120 results); not yet shown to be the pre-existing driver crash.

Not yet run: Release build, portable TFMs, CI, the `Access.LibRed.Accdb` configuration.


ChrisJollyAU:
@MaceWindu Your boolean sort key stuff is way to complex for whats needed. All you not is a NOT operator or unary minus. for example :  ORDER BY NOT [boolcol]

And that works in almost all places and where Access has converted a boolean to a smallint

MaceWindu:
@ChrisJollyAU I think I'm done for now with it. Will update on new releases and review/polish around .net 11 release.

Below is list of standing issues/findings.

---

# LibRed 11.0.0-alpha.3 — open issues found by the linq2db test suite


## A. Wrong results / type contract (silent until materialization)

### R1 — computed column declared one type, returned as another *(regression vs alpha.2)*

```sql
SELECT IIF([t3].[x] < 0, 9, [t3].[x] + 8), [t3].[x] + [t3].[x]
FROM (SELECT IIF([t2].[x] IS NULL, 0, [t2].[x]) AS [x]
      FROM (SELECT (SELECT SUM([MoneyValue]) FROM [LinqDataTypes]) AS [x]
            FROM [LinqDataTypes] [q]) [t2]) [t3]
```

```
alpha.2   Expr1 GetFieldType=Decimal   value Decimal   <- consistent
          Expr2 GetFieldType=Decimal   value Decimal
alpha.3   Expr1 GetFieldType=Int32     value Int32
          Expr2 GetFieldType=Int32     value Decimal   <- declared Int32, returns Decimal
```

`GetSchemaTable` agrees with `GetFieldType` (`DataType = Int32`, `DataTypeName = Long`), so every description of
`Expr2` says `Int32` while `GetValue` returns a `Decimal`. An ORM compiles its reader from the declared type and
fails with `Unable to cast object of type 'System.Decimal' to type 'System.Int32'`. `AllowDBNull` is correct —
nullability is not involved.

### R13 — `IIF` / `CASE` typed by the branch taken

```sql
SELECT TOP 1 1000 - CASE WHEN [t1].[Sum_1] IS NULL THEN 0 ELSE [t1].[Sum_1] END
FROM [Outer] [o]
     OUTER APPLY (SELECT SUM([i].[DecimalValue]) AS [Sum_1]
                  FROM [Inner] [i] WHERE [i].[Group] = [o].[Group]) [t1]
WHERE [o].[Id] = 2          -- a group with no Inner rows
```

The column is the aggregate's `Decimal` (same with `Double` / `Single`), but over the empty group the value comes
back as the `0` literal's `Int32` → `Unable to cast object of type 'System.Int32' to type 'System.Decimal'`.
`IIF(... IS NULL, 0, ...)` behaves identically. ACE unifies the branch types. This is the standard shape for a
coalesced aggregate over an empty sub-query.

### R14 — `NTH_VALUE` ignores the default window frame

```sql
NTH_VALUE([p].[Id], 2) OVER (ORDER BY [p].[Order] DESC)
```

With an `ORDER BY` and no frame, the default is `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`, so the first
row must get `NULL`. LibRed returns the partition's second value for it (evaluated over the whole partition).

### R6 — `GetDateTime` returns `DateTime.MinValue` for the Jet zero date

On a row whose stored serial is `0.0` (1899-12-30): `GetValue` returns `12/30/1899`, **`GetDateTime` returns
`01/01/0001`**. `IsDBNull` is `False`, `AllowDBNull` is `False`. The value itself round-trips correctly through a
parameter, a `#1899-12-30#` literal and `DateSerial(1899, 12, 30)` — only the typed accessor is wrong.

### R5 — `DATETIME` keeps sub-second precision

A parameter holding `09:44:34.653` reads back with the milliseconds. ACE truncates to the second, so the same file
round-trips differently through LibRed and through the Microsoft drivers.

> **My Comment**: I would say it is not an issue, but improvement

### R7 — `CVar` is a no-op

`SELECT CVar(1)` returns a typed `Int32` `1`; ACE returns a Variant that reads back as the string `"1"`.

---

## B. Errors on SQL that ACE executes

### R2 — text column compared to a numeric literal throws *(regression vs alpha.2)*

`VARCHAR` column `S` holding `''`, `'abc'`, `'11'`:

| Query | alpha.2 | alpha.3 |
|---|---|---|
| `SELECT COUNT(*) FROM t WHERE [S] = 11` | `1` | `InvalidCastException: Type mismatch: '' cannot be read as a number` |
| `… WHERE [S] IN (11, 18, 19)` | `1` | throws, same message |
| `… WHERE [S] NOT IN (11, 18, 19)` | `2` | throws — also on a row holding `'abc'` |
| control: `… WHERE [S] NOT IN ('11', '18')` | `2` | `2` |

### R3 — a `PARAMETERS` clause written by Access comes back double-quoted and no longer parses

Same parameterised stored `SELECT` in an ACE-written file and in a LibRed-written one (parameters declared
bracketed, `[@firstName]`, as Access writes them):

```
ACE-written     PARAMETER_NAME       = [@firstName]          <- brackets kept as part of the name
                PROCEDURE_DEFINITION = PARAMETERS [[@firstName]] TEXT(50), [[@lastName]] TEXT(50); SELECT ...
                describe with '@firstName'   -> SqlParseException: token recognition error at: ']'
                describe with '[@firstName]' -> SqlParseException: token recognition error at: '@['
LibRed-written  PARAMETER_NAME       = @firstName
                PROCEDURE_DEFINITION = PARAMETERS [@firstName] TEXT(50), ...; SELECT ...
                describe with '@firstName'           -> 6 columns
```

The read path seems to take the quoted token `[@firstName]` as the name, and the definition then quotes it again.
A parameterised stored `SELECT` in an ACE-written file therefore cannot be described or executed under any
spelling of the name.

### R4 — indexed `GUID` column vs a string literal

```sql
CREATE INDEX IX_G ON t([G]);
SELECT * FROM t WHERE [G] = '{6F9619FF-8B86-D011-B42D-00C04FC964FF}';
```

→ `NotSupportedException: Cannot encode GUID index key from String`. Both brace forms fail, and so does `INSERT`
of a string literal into the column. The message points at the index-key encoder rather than at the dialect.
Unchanged from alpha.2.

---

## C. Differences from ACE on SQL and errors

Each row was run through ACE OLE DB and LibRed on the same file. "Grammar" means LibRed's `AccessSql.g4` at
`v11.0.0-alpha.3`.

### R12 — `WITH OWNERACCESS OPTION` is rejected

```sql
SELECT [PersonID] FROM [Person] WHERE [PersonID] = 1 WITH OWNERACCESS OPTION
```

| ACE OLE DB | LibRed |
|---|---|
| executes, 1 row | `SqlParseException: mismatched input 'WITH' expecting {<EOF>, ';'}` |

Standard Access query syntax, and the one construct in the old R12 list that ACE accepts. A query written by Access with
the option set cannot be run.

### R8 — missing-object errors: inconsistent types, no error number

| Statement | ACE OLE DB | LibRed |
|---|---|---|
| `SELECT * FROM [NoSuchTable]` | `OleDbException`, ErrorCode `-2147217865` — *cannot find the input table or query* | `LibRed.Sql.Binding.SqlBindException` — *Table '…' does not exist.* — no number |
| `DROP TABLE [NoSuchTable]` | `OleDbException`, ErrorCode `-2147217865` — *Table '…' does not exist.* | `System.InvalidOperationException` — *DROP TABLE '…': no such table.* — no number |
| `DROP PROCEDURE [NoSuchProc]` | `OleDbException`, ErrorCode `-2147217865` — *Cannot find table or constraint.* | `System.InvalidOperationException` — *DROP PROCEDURE '…': no such procedure.* — no number |

ACE reports all three as one exception type with one error code (`0x80040E37`). LibRed uses two types, neither
carries `LibRedException.Number`, and a bare `InvalidOperationException` looks the same as a client-side bug, so
callers have to match on message text. Suggestion: `LibRedException` with a dedicated `Number` for "object not found".

### R9 — no external-database table reference *(documented non-goal — for information)*

| Form | ACE OLE DB | LibRed |
|---|---|---|
| `SELECT … FROM [Person] IN '<file>'` | executes | `SqlParseException: mismatched input 'IN'` |
| `SELECT … FROM [;DATABASE=<file>].[Person]` | executes | `SqlParseException: mismatched input '.'` |
| `SELECT … FROM [<file>].[Person]` | executes | `SqlParseException: mismatched input '.'` |

The grammar says `IN externaldatabase` is deliberately absent because LibRed has no linked-database subsystem. Listed
so the difference is on record, not as a request.

---

## D. Not defects (recorded so they are not re-reported)

> **My Comment**: This section could be ignored I think in most

- **A parameterised stored `SELECT` as a row source** (was R10). Behaves exactly like ACE:
  `SELECT [FirstName] FROM [Person_SelectByKey]` with `@id` supplied returns the row on both, and also works aliased,
  filtered and joined to a table. Without the parameter, both refuse: ACE with *No value given for one or more required
  parameters*, LibRed with *No value was supplied for parameter '@id'*. The alpha.2 `SqlParseException` is gone, so
  this is fixed.
- **Syntax neither ACE nor the LibRed grammar supports** (was R11 and most of R12). ACE rejects all of these too:

  | Construct | ACE OLE DB | LibRed |
  |---|---|---|
  | `WHERE ([PersonID] = 1) IS TRUE` | *Invalid use of IS operator* | grammar has only `IS [NOT] NULL` |
  | `WHERE [PersonID] IS NOT DISTINCT FROM 1` | *Syntax error (missing operator)* | same |
  | `CAST([PersonID] AS VARCHAR(10))` | `E_FAIL` | not in grammar |
  | `LIKE 'J%' ESCAPE '~'` | `E_FAIL` | not in grammar |
  | `SELECT TOP 1 WITH TIES …` | *reserved word … misspelled or missing* | not in grammar |
  | `NZ([MiddleName], '-')` | *Undefined function 'NZ'* (NZ exists only inside the Access application) | *Function NZ is not supported* |
  | `{ts '2021-01-01 00:00:00'}` | *Malformed GUID* (ODBC escapes are an ODBC-driver feature) | token error |
  | `SELECT * FROM (VALUES (1), (2)) AS [v]` | *Syntax error in FROM clause* | **works** — a LibRed extension (the grammar's `ValuesTerm`); `… UNION ALL VALUES (2)` works as well. Only the column-alias form `AS [v]([x])` is refused, and the grammar says names come from the leading query |
- **Stored-procedure parameters bind by name.** `CommandType.StoredProcedure` matches arguments by name and throws
  `No value was supplied for parameter '@MiddleName'` when the names differ. Microsoft's OLE DB / ODBC providers bind
  Access stored-query parameters by position and ignore the names, so code written against them often passes
  arbitrary names. Worth a line in the docs.
- A parameter declared bracketed (`[@id]`) keeps the `@` in `GetSchema("ProcedureParameters")`; one declared bare
  (`@id`) comes back as `id`.
- `ORDER BY <boolean>` sorts True first, because True is stored as -1. ACE does the same. A Yes/No column can't hold
  NULL, also as in ACE. linq2db handles both on its side.
- **An unterminated `[` in a `LIKE` pattern** (`… LIKE '[0'`) raises `ArgumentException: Invalid pattern string`. ACE
  rejects the same pattern, so the behaviour matches; only the exception type differs.
- **`CLng` returns a 32-bit `Int32`.** That is VBA's `Long`, which is also 32-bit, so this is not a narrowing bug.

- **Stored-procedure parameters bind by name.** `CommandType.StoredProcedure` resolves arguments by name and throws
  `No value was supplied for parameter '@MiddleName'` when names differ. Microsoft's OLE DB / ODBC providers bind
  Access stored-query parameters positionally and ignore names, so code written against them often passes
  arbitrary names. Worth documenting.
- A parameter declared bracketed (`[@id]`) keeps the `@` in `GetSchema("ProcedureParameters")`; one declared bare
  (`@id`) comes back as `id`.
- `ORDER BY <boolean>` sorts True first (True = -1), exactly as ACE does; a boolean column cannot hold NULL, also as
  ACE. linq2db handles both on its side.
- `LIKE '[0'` raises `ArgumentException: Invalid pattern string` — ACE rejects it too. `CLng` is 32-bit — VBA's
  `Long`.

## Fixed in alpha.3

`GetSchemaTable` / `IDbColumnSchemaGenerator` with correct `AllowDBNull`; `GetDataTypeName` returns store types;
`PRIMARY KEY CLUSTERED`; trailing `IDENTITY`; bare `REFERENCES <table>`; parameterised action-query procedures and
`CommandType.StoredProcedure`; `CommandBehavior.SchemaOnly`; `ORDER BY n` as an ordinal; comma-joined and
derived-table `UPDATE` / `DELETE`; `DATEADD` → `DateTime`; date parts → `Int32`; multi-column `SET` evaluated
against the pre-update row; GUID literals and `CStr(<guid>)` now braced as in ACE; a parameterised stored `SELECT` as
a row source (see D).

ChrisJollyAU:
Alread on it. Following are all fixed locally: R1, R2, R3, R4, R6.

Just a note on R2. I couldnt replicate `WHERE [S] IN (11, 18, 19)` passing in ACE oledb or odbc either. it was rejected by both. What i could find was that if those integers are a parameter ACE will coerce them to the string and it works - and thats one case where libred failed.

On parameters and such: do you prefer bound by name or by position?

R10 (parameterised stored select) like `SELECT [FirstName] FROM [Person_SelectByKey]` does actually work - although i think it broke because of the square brackets in the parameter name (R3 - just a different mechanism of hitting that issue)

R12 - LibRed has no concept of user security so not sure what i will do on that - maybe no-op it
on that note PRIMARY KEY CLUSTERED - the clustered does nothing in Access - i could not find a byte different - its in the grammar as access does accept it but it has no effect

R9 - maybe in the future - its in my want list but want to get the normal stuff good first

Reminder - ACE/VBA has type name that dont always align with c# - Long = Int32, Long long  (or bigint) = int64, Integer = Int16

I also saw somewhere you needed the CreateFormat from the GetSchema resaults - thats also done locally (and turned up a rare new data type as well)

Also, let me know if you want LibRed as .Net 10. Theres no technical reason for it not to be (and its original commit was on .net 10). It just moved to 11 with everything else with the efcore stuff altogether

MaceWindu:
- R10: I think it behaves correctly now, so no changes needed. LibRed provider use named parameters, so it should bind by name
- R12: It is more about accepting sql that MS providers accept, no-op if fine

Having .net 10 - it would be nice to have as it will be supported for two more years.

---

Regarding R2, you are rght. It doesn't work on ACE too.

### R2 — text column compared to an **integer parameter** throws

Re-measured 2026-09-24 against ACE on the same file. With a numeric *literal* all three engines refuse the
comparison, so that part is not a defect. With a numeric *parameter*, ACE converts the parameter to text and compares
strings, while LibRed tries to read every row's text as a number and fails on the first one that isn't (`''`,
`'abc'`).

Repro:

```sql
CREATE TABLE [R2Probe] ([Id] INT NOT NULL, [S] VARCHAR(10));
INSERT INTO [R2Probe] ([Id], [S]) VALUES (1, '');
INSERT INTO [R2Probe] ([Id], [S]) VALUES (2, 'abc');
INSERT INTO [R2Probe] ([Id], [S]) VALUES (3, '11');
```

```csharp
using var cmd = connection.CreateCommand();
cmd.CommandText = "SELECT [Id] FROM [R2Probe] WHERE [S] IN (@p1, @p2) ORDER BY [Id]";   // ODBC: IN (?, ?)

var p1 = cmd.CreateParameter(); p1.ParameterName = "@p1"; p1.DbType = DbType.Int32; p1.Value = 11; cmd.Parameters.Add(p1);
var p2 = cmd.CreateParameter(); p2.ParameterName = "@p2"; p2.DbType = DbType.Int32; p2.Value = 18; cmd.Parameters.Add(p2);

using var reader = cmd.ExecuteReader();   // ACE: returns Id 3. LibRed: throws.
```

| `WHERE` clause | ACE OLE DB | ACE ODBC | LibRed alpha.3 |
|---|---|---|---|
| `[S] = @p` (Int32 `11`) | `3` | `3` | `InvalidCastException: Type mismatch: '' cannot be read as a number.` |
| `[S] IN (@p1, @p2)` (Int32 `11`, `18`) | `3` | `3` | same |
| `[S] NOT IN (@p1, @p2)` (Int32 `11`, `18`) | `1, 2` | `1, 2` | same |
| `[S] = 11` (literal) | *Data type mismatch in criteria expression* | same | `InvalidCastException` (see note) |
| `[S] IN (11, 18)` / `NOT IN` (literals) | *Data type mismatch in criteria expression* | same | `InvalidCastException` |
| control: `[S] IN ('11', '18')` | `3` | `3` | `3` |

Expected: the parameter rows match ACE. Note on the literal rows: refusing is correct, only the exception type differs
(ACE's is a driver error, LibRed's an `InvalidCastException` raised while reading a row rather than a type error for the
statement).

ChrisJollyAU:
@MaceWindu I chase that NTH_VALUE R14 issue. Double checked the implementation and its fine. Feedback is: There's no frame bug. The report is misreading an ordering difference.

Values: NTH_VALUE gives NULL on every row for their data, as it should. The RESPECT NULLS and FROM FIRST RESPECT NULLS forms do too, and IGNORE NULLS correctly gives 5 on the Order 10 row. Only that IGNORE NULLS result puts a "second value" on a row, so the report's description is not what LibRed computes for the query they run.

Why the test fails: Issue1732NthValue has no ORDER BY, yet it asserts the rows arrive in the window's order: res[0].Id NULL, res[1].Id NULL, res[2].Id 6, res[3].Id 5. The window node (QueryExecutor.cs:1758) sorts internally to compute each value, but hands the rows back in input order: 5, 6, NULL, NULL. So res[0].Id is 5 where the test expects NULL. "Expected null but was 5" on the first row looks like the second value leaking in, but it's the row's Id, not its NTH_VALUE. SQL Server, PostgreSQL, SQLite and the rest pass only because they happen to return rows in the order they sorted them for the window. SQL doesn't promise that.

MaceWindu:
@ChrisJollyAU , you are right. Will fix test

